PR #335, opened 2012-12-04
Node.js - allow loadImage to support data url

TimNZ:
The overload for fabric.util.loadImage when run under node.js did not accomodate data:image url.

This could be optimised as src duplicates _src.

I need it to handle this scenario so that the data url is embedded in toSVG().
Perhaps some option to toggle when xlink:href should be original url or data url.

Driven by inkscape not supporting external href.

As it stands the changes still need tweaking due to above considerations.

<pre>
fabric.util.loadImage('https://www.google.co.nz/images/srpr/logo3w.png', function(img) {
    var oImg = new fabric.Image(img);
    oImg.toDataURL(function(e)
    {
        img.src = img._src = e;
        var oImg2 = new fabric.Image(img);
        canvas.add(oImg2);
        console.log(canvas.toSVG());
    });
})
</pre>

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -46,7 +46,7 @@
     var img = new Image();
   	if ( url && url.indexOf('data') == 0 )
   	{
-  		img.src = url;
+  		img.src = img._src = url;
   		callback(img);
   	}
   	else if ( url )


saran110313:
I'm receiving this error originating from fabric.js

Uncaught TypeError: Cannot read property 'apply' of undefined
function klass() {
(ln: 1690) --> this.initialize.apply(this, arguments);
}

asturur:
please, you posted same question with different line number on three threads.

this sounds more a stack overflow question. prepare a minimal jsfiddle that generate the error and post a issue or a stack overflow question. i sure that you understand that the error, outasode the context that is happening, is meaningless

saran110313:
Uncaught TypeError: url.indexOf is not a function

 if (url.indexOf('data') !== 0 && typeof crossOrigin !== 'undefined') {
        img.crossOrigin = crossOrigin;
      }
 this function error in fabric.js

PR #392, opened 2013-01-21
Bugfix align_guidelines for canvas.toDataURL()

Kienz:
Bugfix for canvas.toDataURL() - canvas.clearContext(canvas.contextTop) in after:render clears context if canvas.toDataURL() is called.

Observe before:render and clear contextTop.

### DIFF
--- a/lib/aligning_guidelines.js
+++ b/lib/aligning_guidelines.js
@@ -181,8 +181,11 @@ function initAligningGuidelines(canvas) {
     }
   });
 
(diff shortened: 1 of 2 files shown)


bulolo:
3.6.3 still happened

bulolo:
image

image

if i import and use the Guidelines to init
then i use canvas.toDataURL();
there is an error out.

when i delete the init Guidelines
image

canvas.toDataURL();  run ok.

PR #594, opened 2013-04-29
Link to font files in JSON export

xxorax:
This implements font-face client side for declare fonts added and store them for export, and add support for nodejs.

But, actualy most browsers don't support fontloader interface so there is no way to know if the font is loaded or not, this is because the font is often not applied to the text because the call to renderAll(), it's only when you re-render (when you click on the text object for example) that the font is applied...

The fontloader interface : http://www.w3.org/TR/2013/WD-css3-fonts-20130212/#font-load-event-examples


playground:
Has this been merged?

PR #725, opened 2013-06-27
Added the Textbox class

charlesschaefer:
This class is related with issue #187. It permits the creation of a text box, with or without a background image. 
The text isn't scaled when you scale the box object. It maintains its dimensions based in the fontSize only.

### DIFF
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -603,9 +603,8 @@
     /**
      * Applyes the coordinates and dimensions to the box image
      * @method _moveImageBox
-     * @param {CanvasRenderingContext2D} ctx Context to render on
      */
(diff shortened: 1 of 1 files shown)


fahadnabbasi:
Can anyone tell me how to get only the textbox code? I mean I have fabric.js included as a separate file and I want to include all other files related to textbox functionality. I looked into the code of the demo but it has fabric.js and textbox.js embeded in one file

luck2011:
@fahadnabbasi just click " Files changed 1" tab above and you'll see

devlfm:
I think it would be best if there was an Container that don't uses scales, then TextBox could extend this Container. This would be helpfull to extend to other situations.. Ex: I have a box with stroke, when I resize the box, the stroke scales too, and I don't want that..

PR #1171, opened 2014-02-15
Stop Event.js from messing with global Event objects.

### LINKED ISSUE #1170: Event.js seems to break any other framework
Since upgrading to the latest version of fabric.js, my Qooxdoo click handlers no longer work right. This also seems to be true for [Angular and Backbone](https://github.com/mudcube/Event.js/issues/10) and [jQuery](https://groups.google.com/forum/#!topic/fabricjs/79lggZ8ZVZ0), and apparently the only solution is to patch the fabric.js code to modify Event.js in the following way via a custom patch:

```
// Add custom *EventListener commands to HTMLElements (set false to prevent     funkiness).
root.modifyEventListener = false;

// Add bulk *EventListener commands on NodeLists from querySelectorAll and others  (set false to prevent funkiness).
root.modifySelectors = false;
```

miracle2k:
Fix #1170.

### DIFF
--- a/lib/event.js
+++ b/lib/event.js
@@ -27,10 +27,10 @@ if (typeof(eventjs) === "undefined") var eventjs = Event;
 (function(root) { "use strict";
 
 // Add custom *EventListener commands to HTMLElements (set false to prevent funkiness).
-root.modifyEventListener = true;
+root.modifyEventListener = false;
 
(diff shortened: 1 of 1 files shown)


mudcube:
Hey @kangax Yup, it was updated on Feb 18th 2014. Let me know if you have any problems.

PR #1397, opened 2014-06-18
allow browserify to run over fabric.js

### LINKED ISSUE #1368: Can't browserify Fabric.js
It superficially appears you should be able to browserify Fabric.js because it works as an NPM module but the browserify dependency management isn't smart enough to know that you can drop the node-canvas dependency in the browser.

Would be great to mention this in the docs around the npm section.

Currently only documented in a series of tweets.
https://twitter.com/fabricjs/status/421117110010806273

tmpvar:
These changes provide a shim to avoid trying to pull in `jsdom`, `node-canvas`, and `xmldom` when attempting to run fabric through browserify.  Since there is a check for `isLikelyNode` before performing operations that are specific to the platform these changes should work just like if you loaded fabric.js in the browser directly.

closes #1368 

EDIT: I was using an old browserify which didn't support the false property.  I've force pushed here to cleanup the pull, sorry for any confusion!

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,11 @@
     "HTML5",
     "object model"
   ],
+  "browser" : {
+    "canvas" : false,
+    "jsdom":   false,
+    "xmldom":  false
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/kangax/fabric.js"


gonsfx:
But you still have to install cairo to install fabric via npm - even though you don't need it because you're simply browserifying fabric for use in the browser?

data7of9:
cairo is a pain in the

unaffiliated:
https://www.npmjs.com/package/canvas-browserify

Issue #1270, opened 2014-04-11
Support percentage width / height values on the canvas


Siyfion:
Would it not be nice to also be able to set the canvas height or width to a percentage value? I know this means that internally an event would need to take place on window resize to update the width and height, but is it not worth it?

``` javascript
canvas.setWidth('100%');
canvas.setHeight('80%');
```
##

DesignSolu:
+1

rekrah:
+1

alanmastro:
+1

handersen:
+1

kangax:
Alrighty, marking this as feature.

engrsandeep:
+1 on this

engrsandeep:
Is there any deadline of this feature? eagerly waiting for this feature.... 
Thank you so much for consideration...

andycochran:
+1

andycochran:
Can't you simply style the canvas to make it responsive? 
````
#canvas {
  position: absolute;
  top: 0;
  right: 0;
  left: 0;
  width: 100% !important;
  height: 100% !important;
}
.canvas-container {
  position: relative;
  width: 100% !important;
  height: 0 !important;
  padding-bottom: 100%; // the canvas height as a percentage of its width
}
.upper-canvas {
  width: 100% !important;
  height: 100% !important;
}
````
Of course, this doesn't change the width of the canvas, it scales it. 
Thoughts?

atomicjeep:
The objects would need re-drawn & re-positioned as would the invisible interaction layer, plus we need to handle the proportions of the canvas and it's objects something like this (not tried) http://stackoverflow.com/questions/19937397/show-canvas-in-fullscreen-in-fabric-js/20008922#20008922

andycochran:
Just realized my solution wouldn't work on a StaticCanvas.

Bnaya:
Take a look on #1505

keremdemirer:
+1

naimulhaider:
+1

malrefaay:
+1

heastost:
+1

mkupiniak:
+1

vikaskallidantheyil:
+1

asturur:
i see we have lot of +1.
I'm going in the process to make the upper canvas retina enabled.
Once i see how the event position change with a css transform, and once i found a nice way to find the css transform for ie10+, then i think we can do this.

Bahaa-Addin:
[...]

PR #1505, opened 2014-07-18
Enhanced set dimensions

Bnaya:
Added ability to set css & backstore dimensions separately 
Introduced options param to the dimensions setters, options.backstoreOnly & options.cssOnly

When backstoreOnly set to true, only the value of the canvas element width & height will be changed.
When cssOnly set to true, only the css width & height of the involved DOM elements will be changed,
and you should include the unit of measure (px/%/em) with the passed value (eg '100px' '100%') and not just 100

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -172,8 +172,8 @@
   };
 
   var origSetWidth = fabric.StaticCanvas.prototype.setWidth;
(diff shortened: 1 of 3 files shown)


Bnaya:
#1270

lk77:
Is there any way to bypass this ? when i do canvas.setWidth(1920) it does not work and set a width of 3840

PR #1544, opened 2014-07-30
Add image blend filter

mgalang:
Implements several image blend modes (can add more later).


ignaty:
Demo would be really helpful to have.

njv:
Is there any way to get this working for image + canvas blends, i.e. not just image + image or image + color?

PR #1570, opened 2014-08-06
When user drags inactive object, make sure he won't scale it instead

pjanik:
To replicate the issue using current release of FabricJS:
1. Open http://fabricjs.com/.
2. Move your mouse pointer **exactly** to the corner of a red rectangle.
3. Do **not** select the rectangle, note that 'move' cursor is visible.
4. Drag the rectangle.

I would expect that object would be moved, but instead it's scaled, as exactly under the mouse pointer there is a control point (invisible at the beginning, until the object is selected).

The problem seems to be here:
https://github.com/kangax/fabric.js/blob/818ab118b30a9205a0e57620452b08bb8f5f18cc/src/mixins/canvas_events.mixin.js#L414-L415

`_findTargetCorner` checks whether the target is active (what makes sense). It's called twice, once in `_beforeTransform` and the second time in `_setupCurrentTransform`. However the target is activated in the meantime, in `_beforeTransform`. So the second call to `_findTargetCorner` actually returns corner and scaling is being activated. I've just moved target activation out of `_beforeTransform` and it seems to fix the problem for me. These functions are used only in this context, so hopefully it's not a risky change.

Regards,
Piotr

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -413,6 +413,11 @@
       if (target && target.selectable && !shouldGroup) {
         this._beforeTransform(e, target);
         this._setupCurrentTransform(e, target);
+        // activate object after transform setup to make sure that corners detection is valid
+        if (target !== this.getActiveGroup() && target !== this.getActiveObject()) {
+          this.deactivateAll();
+          this.setActiveObject(target, e);
+        }
       }
       // we must renderAll so that active image is placed on the top canvas
       shouldRender && this.renderAll();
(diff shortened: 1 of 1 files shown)


KJTsanaktsidis:
I agree with @pjanik; I would find it surprising if I could resize an object that I had not selected.

pjanik:
Yeap, I still feel the same after couple of months... This can be considered as a matter of taste though.

But the fact that I **can't** move small object (as I always resize it instead) is an obvious problem / bug to me and it's pretty annoying in practice.

asturur:
I was thinking again about this, i'm sure @pjanik is right.
An object that is smaller than twice the size of control is never draggabale.
And this is not normal - right.

asturur:
replaced by #2907

PR #1625, opened 2014-08-27
Use hiddenTextArea to generate inputs for IText - POC

vamo89:
I had problems with some inputs at iText that need two separated keystrokes, like discussed at https://github.com/kangax/fabric.js/issues/1175.
I'm not sure if I can solve the problem for Mac way of doing this, but for linux/windows I have a solution that may clean a lot of code and complexity from itext_key_behavior.mixin.js.

That's the idea: instead of listening some basic events like 'keypress' to handle input and mapping all user possible actions, we can let the textarea handle it and get it's input. In that way, the only thing that we've to care is keeping the caret at right position.

I made a POC to know if you have interest in changing the aproach. Chars like 'áéíâô' are now accepted. It works like an textarea. I didn't yet implement ctrl+left/right arrow, so in that case cursor become wrong-placed. 

If you have interest, I can continue development focusing on chrome/linux, and I hope some help to fix the behavior for others browsers/OS.
If you don't, I'll just create a new fabric class at my code and use this aproach, because it's easier for latin languages inputs.

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -12,9 +12,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     fabric.document.body.appendChild(this.hiddenTextarea);
 
     fabric.util.addListener(this.hiddenTextarea, 'keydown', this.onKeyDown.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'keypress', this.onKeyPress.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'copy', this.copy.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'paste', this.paste.bind(this));
+    fabric.util.addListener(this.hiddenTextarea, 'input', this.onInput.bind(this));
 
     if (!this._clickHandlerInitialized && this.canvas) {
       fabric.util.addListener(this.canvas.upperCanvasEl, 'click', this.onClick.bind(this));
@@ -26,13 +24,10 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    * @private
    */
   _keysMap: {
-    8:  'removeChars',
-    13: 'insertNewline',
     37: 'moveCursorLeft',
     38: 'moveCursorUp',
     39: 'moveCursorRight',
-    40: 'moveCursorDown',
-    46: 'forwardDelete'
+    40: 'moveCursorDown'
   },
 
   /**
@@ -40,12 +35,39 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    */
   _ctrlKeysMap: {
     65: 'selectAll',
-    88: 'cut'
+    // TODO: LINE LEFT AND LINE RIGHT
+    // 37: 'moveCursorLineLeft',
+    // 39: 'moveCursorLineRight'
   },
 
   onClick: function() {
     // No need to trigger click event here, focus is enough to have the keyboard appear on Android
     this.hiddenTextarea && this.hiddenTextarea.focus();
+  },
+
+  _doGetCaretPosition: function (textArea) {
+    var caretPos = 0;
+    // IE Support
+    if (document.selection) {
+      textArea.focus();
+      var sel = document.selection.createRange();
+
+      sel.moveStart('character', -textArea.value.length);
+
+      caretPos = sel.text.length;
+    }
+    // Firefox support
+    else if (textArea.selectionStart || textArea.selectionStart == '0')
+      caretPos = textArea.selectionStart;
+    return caretPos;
+  },
+
+  onInput: function (e) {
+    var cp = this._doGetCaretPosition(this.hiddenTextarea);
+    this.text = '';
+    this.insertChars(e.srcElement.value);
+    this.setSelectionStart(cp);
+    this.setSelectionEnd(cp);
   },
 
   /**


antoniojavierSC:
Is there any solutions about this?

Thanks

asturur:
do you have still problem using some characters? itext is on heavy restyling, talk now or never.

antoniojavierSC:
In a Spanish keyboard from a pc when you have to put an accent on a vowel is necessary to press the accent key and then the vowel, currently this works not directly editing the itext visually, seems not capture the key event accent. An example of Spanish words affected are: "jamón" and "fútbol" :)

This does not happen from an iPad, because when you want to put an accent on a vowel holding down the vowel key and character accentuated appears.

I searched the internet video where you can see what I mean in the way of accentuating in Spanish from pc:

https://www.youtube.com/watch?v=n98VcMz3t9U

Thanks!

asturur:
which keyboard do you have? i have an american keyboard i can input accent without anyproblem in firefox, and chrome.
Please provide browser version, keyboard type ( american , spanish... ), fabric version. thanks

antoniojavierSC:
I tried the demo directly with Google Chrome version 40.0.2214.91 (64-bit):

http://fabricjs.com/kitchensink/

The keyboard is the Spanish ISO

asturur:
i will put a modified version on fabricjs on a link, would you go and thest this, and report me some messages in the  console?

When i insert á or à i get number 224 and 225, plus a couple of false.
Would you try?

antoniojavierSC:
It is very strange, in the console's event appears all keys except the accent. If I press the accent key and then the vowel key, no event is captured in the console and nothing appears on the itext, but if I press the vowel key a second time does appear, but without the accent, obviously. For example if the vowel "a" appears 97 + false + false.

Are you testing with spanish ISO keyboard?

captura de pantalla 2015-01-25 a las 12 53 46

asturur:
no i have a english US keyboard.
The accent does not go trought, to get jsut accent i have to press accent AND space.
Because in my keyboard, and i think all US keybord single accent does not fire, keyboard wait for letter combination.

the same for ^ ~ ` ' "  that help produce   â ã à á ä  and maybe some more thing i does not know.
But here it works good, just the accent does not fire anything, not a single event, and after the second keypress a single event fires.
I do not know if this is handled at browser lever, driver level, OS level OR hardware level by keyboard. Here it jsut works.
So when you press accent you get nothing, and when you press the a you get 97?
If you press two times accent, do you get double 96?

antoniojavierSC:
No matter the time you press the accent key, it never produces anything for console :(

Nothing appears on console when you press a key that would have to produce â ã ä à á. I have been testing with OSX Yosemite and Ubuntu 14.04 with Firefox and Chrome.

asturur:
I should ask you to make more tries, if you go again on that page i linked you.
You will see 2 textboxes when you edit the text, the normal ktichensink has and the hidden one, not hidden, i also doubled the console.log passing all the event.
Does something happen when you press the accent or the letter? please try to describe as much accurate you can. thanks.

antoniojavierSC:
In the new textarea accents are displayed, but in itext it not appears. This is the console log, If you need to open an object let me know. Thanks

captura de pantalla 2015-01-25 a las 14 20 39

asturur:
keydown { target: <textarea>, key: "DeadGrave", charCode: 0, keyCode: 192 } fabric.js:22666
keydown { target: <textarea>, key: "à", charCode: 0, keyCode: 65 } fabric.js:22666
keypress { target: <textarea>, key: "à", charCode: 224, keyCode: 0 }

Don't you get even the keypress? no keypress no inserting in iText, it happens there.

antoniojavierSC:
Firefox console shows:

For "á":
keydown { target: <textarea>, key: "", charCode: 0, keyCode: 0 } fabric.js:22666

For "à":
keydown { target: <textarea>, key: "", charCode: 0, keyCode: 192 }

For "a":
keydown { target: <textarea>, key: "a", charCode: 0, keyCode: 65 } fabric.js:22666
keypress { target: <textarea>, key: "a", charCode: 97, keyCode: 0 }

Indeed if the accent key appears no keypress event

asturur:
ok don't know what to do.
@Kangax any idea how to handle it?

My worst nightmare is change the text insertion from event->itext object->sync on hidden textarea
to  event->hiddentextarea-> sync on itext object.
but looks like it would solve that.

kangax:
I don't see a way to solve this. On my OS X, holding "e", then choosing from "èéêëēėę" by clicking it, for example, doesn't trigger anything at all. So as far JS/browser is concerned, it only saw "e" being pressed and doesn't know that something was changed to, say, "é" afterwards.

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -12,9 +12,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     fabric.document.body.appendChild(this.hiddenTextarea);
 
     fabric.util.addListener(this.hiddenTextarea, 'keydown', this.onKeyDown.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'keypress', this.onKeyPress.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'copy', this.copy.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'paste', this.paste.bind(this));
```
kangax: Why does this remove copy and paste event handlers?

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -48,6 +45,31 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     this.hiddenTextarea && this.hiddenTextarea.focus();
   },
 
+  _doGetCaretPosition: function (textArea) {
```
kangax: Strange name. Why "doGet..." and not just "get..."? :)

src/mixins/itext_key_behavior.mixin.js
vamo89: Because the hiddenTextArea handle it without extra listeners

src/mixins/itext_key_behavior.mixin.js
vamo89: Strange indeed, was just the hurry to get it working =/

vamo89:
@asturur the 'event->hiddentextarea-> sync on itext object' was exactly what I did to fix that. Why is it your worst nightmare?
In that way, you just have to worry about the sync instead of re-implementing text-related functions like copy, paste, accent and special characters.

I used the demo as well. My keyboard is in Portuguese-BR.
When I press an accent key, console shows nothing.
Pressing a alone, I get two events: 'key 65 charcode 0' and 'key 97 charcode 97'
Pressing acute accent twice, I get one 'key 229'.
Pressing acute + a, I get one 'key 229' too.

Both of them appear at hiddenTextArea but doesn't at the kitchensink.
When I click at the kitchensink again, accents and accented letters desapear at hiddenTextArea.

asturur:
sorry @vamo89 i did not see the question. my nigthmare is to change some basic function and test everything from scratch.
for sure reading from textbox is the next idea, but there are lot of things to check, selecting by keys, with arrows..., copy, pasting... this pr was a pain :)

marcospassos:
@vamo89 Your PR is not working anymore using the latest version (something related to the `_clickHandlerInitialized`). I'm getting the following error:
```
Uncaught TypeError: Cannot read property 'bind' of undefined
```

Do you have an updated version of this path?

PR #1674, opened 2014-09-17
Refactor accessors

### LINKED ISSUE #1621: Fabric js is not csp unsafe-eval compliant
Due the use in the new Function notation, when preventing eval via csp header (unsafe-eval),
the library breaks.
https://developer.mozilla.org/en-US/docs/Web/Security/CSP/Using_Content_Security_Policy
The use in csp become more and more common.

I believe it will be possible to implement it using regular js without new Function,
is there a reason to keep it that way?  
https://github.com/kangax/fabric.js/blob/master/src/util/misc.js#L429-L434
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4048489-fabric-js-is-not-csp-unsafe-eval-compliant?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Bnaya:
Predefine the accessors methods as a mixing instead of creating them on runtime per class, while inherit only the needed methods for the class.
Fixes #1621

### DIFF
--- a/build.js
+++ b/build.js
@@ -144,6 +144,7 @@ var filesToInclude = [
   ifSpecifiedDependencyInclude('serialization', 'json', 'lib/json2.js'),
   ifSpecifiedInclude('gestures', 'lib/event.js'),
 
+  'src/mixins/accessors.mixin.js',
   'src/mixins/observable.mixin.js',
   'src/mixins/collection.mixin.js',
 
--- a/src/mixins/accessors.mixin.js
+++ b/src/mixins/accessors.mixin.js
@@ -1,0 +1,256 @@
+(function() {
+  fabric.accessors = {
+    getBackgroundColor: function() {
+      return this.get('backgroundColor');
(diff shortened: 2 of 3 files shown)


Bnaya:
@kangax can you take a look?

kangax:
@Bnaya sorry, I know this is important to you, but I'm not convinced I want all the extra code just to get rid of new Function. We can leave this PR hanging and see if anyone else comes across this issue in the future and is interested to see this as well.

XhmikosR:
@kangax please reconsider this one.

asturur:
accessors are optionals now, new function can be skipped entirely

PR #1806, opened 2014-11-03
Svg Export, do not export default values

asturur:
Fix for exporting just the non defalut values of elements attributes.


asturur:
I wanna finish this one day or another :)

PR #1832, opened 2014-11-12
User defined "selectable area" for objects

ghost:
This adds an optional {eventBounds} attribute to the fabric object that defines where selecting an object works.

IE: if you have a mask covering part of an object, it forces clicks to be on the object in order to interact with the object.

eventBounds = {x: val, y: val, width: val, height: val}


karthickvkumar:
Hi @ghost this i really cool feature. I was trying achieve like your demo. It would be helpful if u provide a gist or jsfiddle demo

Issue #1310, opened 2014-05-05
Starting IText inline editing scrolls to top of the page


ralfwahl:
When I click inside an IText field to start the inline editing, the whole document scrolls up to the start of the page, which is quite confusing for the user. I have this effect with fabric.js 1.4.4 and 1.4.5. It can be also seen  in the official IText demo http://fabricjs.com/test/misc/itext.html:
Take one of the last IText objects, e.g. "Lots of text (performance test)" and click there. The page will jump up, making the current inline editing move out of the visible screen.

This problem can be watched for example with Firefox 28/29 and Internet Explorer 11. Chrome 34 handles it better, stays at the current document scroll position.

When trying to debug the script, it seems that all the event handling starting from _onMouseUp works as expected. After leaving that fabric.js code some rerendering takes place that moves the document upward. 

Another effect is, that it works OK if I'm in fullscreen mode of one scrollable DIV containing the canvas (not the F11 overall fullscreen, just of the DIV node). In that case everything stays at place. Is there something happening on top of the document that makes some browsers scroll to that position?

Kind regards
Ralf

filiperoberto:
It's still scrolling for me, but now to bottom. It's weird because some times textarea is appended to body and sometimes to canvas-container. When it's appended to body there is no problem, but to canvas-container it scrolls to bottom.

FeaRCODE:
@filiperoberto @asturur I can also confirm in my configuration, the text JUMPS to the bottom of the page.

asturur:
in the current configuration the textbox is placed wher you clicked and then follow the cursor. How are you entering the textedit? clicking or by code?

FeaRCODE:
Clicking. I have an event that places the text box.

murilobd:
@filiperoberto I was with this same issue and fixed with that:
```javascript
var bodyRect = document.body.getBoundingClientRect(),
        elemRect = this.canvas.upperCanvasEl.getBoundingClientRect(),
        offsetRect = elemRect.top - bodyRect.top;
this.hiddenTextarea.style.cssText = 'position: fixed !important; top: ' + offsetRect + 'px !important; left: 0px; opacity: 0; width: 0px; height: 0px; z-index: -999;';
```

PR #1854, opened 2014-11-25
Fix text jump to top of page

asturur:
Following replies in #1310 this fix is the best all in all fix for the jumping problem.

zoomcase firefox and chrome not jumping, internet explorer i didn't bother to test.
I intentionally scrolled left and down, to make it as general as possible and used high browser zoom.

image

@slashgrin i opened parallel PR, sorry for not waiting your tests but i preferred to go on.

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -7,8 +7,8 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     this.hiddenTextarea = fabric.document.createElement('textarea');
 
     this.hiddenTextarea.setAttribute('autocapitalize', 'off');
-    this.hiddenTextarea.style.cssText = 'position: absolute; top: 0; left: -9999px';
-
+    this.hiddenTextarea.style.cssText = 'position: fixed; bottom: 20px; left: 0px; opacity: 0;'
+                                        + ' width: 0px; height: 0px; z-index: -999;';
     fabric.document.body.appendChild(this.hiddenTextarea);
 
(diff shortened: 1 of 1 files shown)


andycochran:
This is still a problem on mobile, where fixed position isn't supported. Shouldn't the textarea simply be positioned at the same Y value as the canvas element?

ajck:
@asturur @kangax - unfortunately itext on ipad/android tablet still a problem with on screen keyboard causing various issues including canvas scrolling out of the way when editing active or text hilighted with finger. I've also tried this fix: https://github.com/kangax/fabric.js/issues/1310#issuecomment-42272905

Tested on ipad mini with ios 9.1 and safari, also generic Android tablet with Android 4.1.1 and native browser plus latest Chrome and Opera. Also posted in group here: https://groups.google.com/forum/#!topic/fabricjs/Up2MBnbWcAU

Many thanks

asturur:
https://github.com/asturur/fabric.js/pull/6/files

could you try to apply this fix?
i was thinking of it but never found time to try it.

The fix could be improved detecting cursor position, but for now is a stard.
if it does not change anything is not right way to go.

ajck:
@asturur thanks for the suggestion. Tried that and unfortunately does not work - it makes a visible text field appear top right of the canvas that mirrors the IText you're editing, and it also stops the native on screen keyboard appear at all, on both iOS (9.1, tested on iPad mini) and Android (4.1.1, tested on latest Chrome and Opera).

asturur:
i ll check it better :)
but does it jump?

ajck:
It doesn't jump but with no keyboard appearing you can't type anything either :)

PR #1856, opened 2014-11-26
Fix requireJS compatibility issue

finfin:
last few lines in `src/amd/require.js` adds a  global **export** variable

however, when using in requirejs, the added global export will cause other module fails to load
cause the added global export will make other module thought they're in the wrong environment(server side)

I removed two lines and it works with my project


kangax:
Spoke with @jdalton about this and decided that we should adapt something similar to what lodash is doing — https://github.com/lodash/lodash/blob/master/lodash.src.js#L10992-L11020

PR #1875, opened 2014-12-03
Added clipBackground option

asturur:
Add an option to render the background before the canvas clipcontext.

I needed this option for a little tool based on fabric, i would like to see if it can be considered interesting.
I know i can achieve the same effect probably with some CSS background, but in this way the effect is exportable with canvas.toObject.

The effect is shown in the screens and is simple as it gets:
image

Using clipping masks and keeping the background function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 **Edge**
-
+- Add "clipBackground" a canvas option to render the background before the canvas clip context.
 - [BACK_INCOMPAT] `fabric.Collection#remove` doesn't return removed object -> returns `this` (chainable)
 
 - Add "mouse:over" and "mouse:out" canvas events (and corresponding "mouseover", "mouseout" object events)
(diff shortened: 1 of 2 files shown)


kangax:
@asturur Err... I followed your instructions and seeing this — https://www.dropbox.com/s/gzc6ro1ee3ijk9m/Screenshot%202015-01-12%2013.15.43.png?dl=0 Still not sure what exactly this feature adds and how it's different from current behavior.

asturur:
Because me, the genius, gave you this instructions but This morning, i needed the line of code from that example and i cutted from the file instead of copying it. 

Take this link: http://www.deltalink.it/andreab/fabric/editor/

scroll end of page, from blank product click the iphone. is already configured.
Then click an image from the box near there "available artwork"
As you see the image is clipped but the background not.

kangax:
Hm... I see it now! I can't immediately tell if this is already possible or not. Will give it more thought!

PR #1882, opened 2014-12-04
Make cleaner setCoords and geometry code

asturur:
@knagax This PR is mainly about code clean up.
3 methods have been shrunk. Code duplication has been removed and setCoords mad ligther.
Just 4 corners are calculated using the transformation with vpt the other are find "interpolating"
mtr before was a copy of mt , and then made correct in setCornerCoords, now is calculated correct and _setCornerCoords is made simpler.

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -302,35 +302,13 @@
      * @chainable
      */
     setCoords: function() {
-      var strokeWidth = this.strokeWidth,
-          theta = degreesToRadians(this.angle),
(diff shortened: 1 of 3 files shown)


kangax:
Is this good to go?

asturur:
I did not do all that test suite.
I do not have a mirrored copy of fabricjs.com
Trying in next days

kangax:
Ok

asturur:
This code looks to no harm to me, i did check trought all the tests and they look visually the same.,

PR #1902, opened 2014-12-18
Update object.class.js

logocomune:
Added the following  properties to object representation of an instance :
hasControls
hasBorders
hasRotatingPoint
lockMovementX
lockMovementY
lockRotation
lockScalingX
lockScalingY
lockUniScaling
lockScalingFlip


kangax:
We don't include those properties on purpose. You can always use "propertiesToInclude" argument for any extra ones you might need.

PR #1931, opened 2015-01-18
Fix Shadow svg output

asturur:
Make the shadow in the svg more similar to the canvas rendering, making it less dark.

canvas output
image

old svg output
image

new svg output
image


asturur:
i was fixing tests.....

2015-01-18 17:29 GMT+01:00 Juriy Zaytsev <notifications@github.com>:

PR #1933, opened 2015-01-19
Support HTTPS with different ports not just 443

earlyster:
We would like to make this change to support HTTPS running on different ports other than 443 since user may have 443 used for something else on their machine. 

Thanks,

Justin

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -22,7 +22,7 @@
     }
 
     // assign request handler based on protocol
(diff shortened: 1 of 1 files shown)


kangax:
Cool, but shouldn't it be indexOf() > -1?

kangax:
Will fail if index == 0 afaics

earlyster:
Trying to show someone how to use github forking and code at same time. Fixed condition to ===0 for https.

kangax:
Thanks! :)

PR #1954, opened 2015-01-30
Extend iText to accept accents

vamo89:
I'm accepting accents on fabric using this js I made.
It's another aproach for the same problem https://github.com/kangax/fabric.js/pull/1625.

Since itext is on heavy restyling, I hope you can find this useful.

It's not connected as a plugin yet.
Works only importing the new js and creating accentedTextCreator instead of iText.
Instead of: new fabric.IText(...
Do: new accentedTextCreator()(...

Change-Id: I31df92b058c0d8a7af6209618bcecfb99f4d2306

### DIFF
--- a/src/mixins/accented_text.js
+++ b/src/mixins/accented_text.js
@@ -1,0 +1,50 @@
+function accentedTextCreator() {
+    return fabric.util.createClass(fabric.IText, {
+        initHiddenTextarea: function() {
+            this.hiddenTextarea = fabric.document.createElement('textarea');
+
+            this.hiddenTextarea.setAttribute('autocapitalize', 'off');
+            this.hiddenTextarea.setAttribute('wrap', 'off');
+            this.hiddenTextarea.setAttribute('rows', '6');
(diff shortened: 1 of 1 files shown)


asturur:
@vamo89 are you sure some code is not missing? i can just see you reset this.text to ''. This means that somewhere your a putting this.text = this.hiddenArea.text. or am i wrong?

vamo89:
After the reset, the line <code>this.insertChars(e.srcElement.value); </code> does just that.

asturur:
i completely missed it :)

marcospassos:
This is a very critical bug. Any latin based keyboard works well using IText.

asturur:
This PR is not in sync with master OR anyway with current key use and fabric style of code.
The idea of using the hidden text area for interacting with ITEXT is for sure the next thing to do on text, but i would close this PR as is not usefull as it is.

marcospassos:
@vamo89 Victor, do you have the updated version of this path?

vamo89:
@marcospassos, I don't have an updated version. I still didn't updated fabric.js version, waiting for this fix. Fabric.js is useless for me without this. Sorry for late reply.

PR #1958, opened 2015-02-02
Fixes svg rendering rules

asturur:
This PR fixes some bugs of svg import:

Circle class:
Radius equal 0 is not an error, but the circle should not be rendered ( this is handled with normal if object.width and object.height both equal 0 early return from rendering );

Polygon and polyline, empty point should not give empty object, just not rendered object.
Same for odd point number. Firefox and chrome are rendering it, but they should not.

Rect: whilec canvas draw a 0 width OR 0 height stroke rectangle, svg says not. So simply set visible equal false to rect coming from svg elements with width or height equal to 0.

Those SVG where previously not rendered in fabric and were throwing errors.

image

Rendering by browsers:
image

image

image

image

### DIFF
--- a/src/shapes/polygon.class.js
+++ b/src/shapes/polygon.class.js
@@ -57,7 +57,7 @@
      */
     initialize: function(points, options) {
       options = options || { };
-      this.points = points;
+      this.points = points || [ ];
       this.callSuper('initialize', options);
       this._calcDimensions();
       if (!('top' in options)) {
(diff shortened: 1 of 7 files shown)


src/shapes/polygon.class.js
```diff
@@ -79,11 +79,11 @@
...
           maxY = max(points, 'y');
 
-      this.width = (maxX - minX) || 1;
-      this.height = (maxY - minY) || 1;
+      this.width = (maxX - minX) || 0;
+      this.height = (maxY - minY) || 0;
 
-      this.minX = minX,
-      this.minY = minY;
+      this.minX = minX || 0,
```
asturur: Accepting no points in polygon or polyline means that we have to have a fall back in case top and left are not specified.

src/shapes/polygon.class.js
```diff
@@ -154,7 +156,14 @@
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
     commonRender: function(ctx) {
-      var point;
+      var point, len = this.points.length;
+
+      if (!len || isNaN(this.points[len - 1].y)) {
```
asturur: this is a way to detect no points or odd points.

asturur:
@kangax when you have time have a look?

kangax:
Looks good, thanks @asturur

asturur:
?

PR #1960, opened 2015-02-04
Shadow Size fix

### LINKED ISSUE #1917: Shadows affected by the canvas zoom level
Hi everyone,

When the canvas zoom level is modified, the shadows of objects are affected. When zooming in, their offset is increased and when the canvas is zoomed out, the shadows sometimes disappear (as their offset eventually is set to zero, I presume). It seems to me that this is because the values for shadows offset remain absolute; that is, they do not take into account the changes on the canvas zoom level. 

The problem is specially evident when using the stroke property of fabric objects. I have placed an example to illustrate this at: http://jsfiddle.net/Lo9kxb0j/1

Cheers,
Gonzalo Gabriel

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/7501644-shadows-affected-by-the-canvas-zoom-level?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Fix the shadow and canvas zoom problem.

closes #1917, close #1917

i do not know where this
"this.canvas._currentMultiplier"
was coming from

image

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1065,12 +1065,13 @@
         return;
       }
 
-      var mult = (this.canvas && this.canvas._currentMultiplier) || 1;
+      var multX = (this.canvas && this.canvas.this.viewportTransform[0]) || 1,
+          multY = (this.canvas && this.canvas.this.viewportTransform[3]) || 1;
 
       ctx.shadowColor = this.shadow.color;
       ctx.shadowBlur = this.shadow.blur * mult * (this.scaleX + this.scaleY) / 2;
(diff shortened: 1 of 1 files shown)


asturur:
@kangax i finally tested it.
Now shadow is proportional to object scale AND canvas viewportTransform
I think you can pull it.

kangax:
Great. We should probably create a separate test page for canvas zoom, with different shapes, images, text, itext, shadows, patterns, etc.

Issue #1740, opened 2014-10-11
IText exitEditing, enterEditing does not work properly


kingkrausel:
In the latest version of chrome (38.0.2125.101 m) the IText methods exitEditing and enterEditing do not work properly in the following scenario:

```javascript
var text = new fabric.IText('text');
canvas.add(text);
//now click on the text and change it to '123'
//keep the IText selected as it is (in editing mode)
text.exitEditing();
text.enterEditing(); //now in editing mode, cursor on correct position
//now press the keys 4,5 and 6 on your keyboard
//text changes from '123' to '12356', i.e. the first key press is not processed correctly
```

So the first new input after entering editing mode gets lost. This was working for an older version of chrome, unfortunately, I can not tell which version it was. I used fabric 1.4.11 for the test.

LIDJungle:
Thanks for the fix! This one had me too for 15 minutes.

asturur:
@kangax if this is a bug we can solve it adding this.hiddenTextare.focus() on EnterEditing().
If this is not a bug and calling the focus is left to the user, we should close this issue.

asturur:
What do you think @kangax
Should the user call the setfocus by himself, if he want to enterEditing without using the mouse, or should we put this.hiddenTextarea.focus()  inside the enterEditing method?

kangax:
I think we should definitely set focus when calling `enterEditing` since `enterEditing` is supposed to be programmatic way of entering edit state (identical to user clicking the field).

PR #1961, opened 2015-02-05
Add forcedfocus on enterEditing

asturur:
As per #1740 discussion we force focus to mimic click on object

Someone should test if the android hack for onclcik event is removable.

close 1740#

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -15,11 +15,6 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     fabric.util.addListener(this.hiddenTextarea, 'keypress', this.onKeyPress.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'copy', this.copy.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'paste', this.paste.bind(this));
-
-    if (!this._clickHandlerInitialized && this.canvas) {
-      fabric.util.addListener(this.canvas.upperCanvasEl, 'click', this.onClick.bind(this));
-      this._clickHandlerInitialized = true;
-    }
   },
 
   /**
@@ -47,11 +42,6 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
   _ctrlKeysMap: {
     65: 'selectAll',
     88: 'cut'
-  },
-
-  onClick: function() {
-    // No need to trigger click event here, focus is enough to have the keyboard appear on Android
-    this.hiddenTextarea && this.hiddenTextarea.focus();
   },
 
   /**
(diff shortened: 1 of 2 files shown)


asturur:
@kangax check if is ok for you.

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -51,6 +51,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
 
   onClick: function() {
     // No need to trigger click event here, focus is enough to have the keyboard appear on Android
+    // 2015-02-05 with forced focus on enterEditing maybe we can remove this now.
```
kangax: Good point! I think we can remove this now.

asturur: You think it, or are you asking me to remove it?
for me remove. We have plenty of user ready to open another issue to tell us we were wrong.

kangax: How so? Don't we already call `enterEditing` when clicking field?

asturur: Yes i think you are right,
i do not even like that onClick event is on_key_behaviour instead that click_behaviour. I remove it.

asturur:
@kangax done.

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -15,11 +15,6 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     fabric.util.addListener(this.hiddenTextarea, 'keypress', this.onKeyPress.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'copy', this.copy.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'paste', this.paste.bind(this));
-
-    if (!this._clickHandlerInitialized && this.canvas) {
-      fabric.util.addListener(this.canvas.upperCanvasEl, 'click', this.onClick.bind(this));
-      this._clickHandlerInitialized = true;
```
kangax: Do we use `_clickHandlerInitialized` anywhere else?

asturur: Ok i did this PR very fast, but not so fast.
I made a search in fabric.js before deleting it.

kangax: Nice! Then we're all good :)

asturur:
Can we revert this? I want to test it separately from other things.
We have some click problem here and there.

asturur:
I tell you even why:
With enter edit we enter the editing.
If we click again, we put focus on the canvas and we do no go back to the hidden Textarea.

in my opionion focus should be in enterEdit AND in the onclick event.

Please revert or patch or i 'll do

kangax:
@asturur hm, you don't have access rights to revert?

asturur:
yes, but is just specular to merge.
So i thought it was your job.
I revert.

Issue #508, opened 2013-03-27
activeGroup object sort is incorrect


Kienz:
The zIndex of activeGroup objects is wrong.

objects before activeGroup selection:
<img src="https://s3-eu-west-1.amazonaws.com/kienzle.geschaeft/projects/fabricjs/objects_before_activeGroup.png"/>

objects in activeGroup selection:
<img src="https://s3-eu-west-1.amazonaws.com/kienzle.geschaeft/projects/fabricjs/objects_activeGroup.png" />

jagi:
Issue still exists 
![fabric1](https://cloud.githubusercontent.com/assets/1051229/6062505/9e4470c8-ad50-11e4-849b-2c563ee00220.png)
![fabric2](https://cloud.githubusercontent.com/assets/1051229/6062511/a1c0951a-ad50-11e4-9d02-39cb0dae00db.png)

The problem is in rendering order not placing group into `canvas._objects`.
Here are fixes to code that will make it work properly:

```js
fabric.Canvas.prototype.renderAll = function(allOnTop) {
  var canvasToDrawOn = this[(allOnTop === true && this.interactive) ? 'contextTop' : 'contextContainer'],
    activeGroup = this.getActiveGroup();

  if (this.contextTop && this.selection && !this._groupSelector) {
    this.clearContext(this.contextTop);
  }

  if (!allOnTop) {
    this.clearContext(canvasToDrawOn);
  }

  this.fire('before:render');

  if (this.clipTo) {
    fabric.util.clipContext(this, canvasToDrawOn);
  }

  this._renderBackground(canvasToDrawOn);
  this._renderObjects(canvasToDrawOn, activeGroup);
  // this._renderActiveGroup(canvasToDrawOn, activeGroup);

  if (this.clipTo) {
    canvasToDrawOn.restore();
  }

  this._renderOverlay(canvasToDrawOn);

  if (this.controlsAboveOverlay && this.interactive) {
    this.drawControls(canvasToDrawOn);
  }

  this.fire('after:render');

  return this;
};

fabric.Canvas.prototype._renderObjects = function(ctx, activeGroup) {
  var i, length;

  if (!activeGroup || this.preserveObjectStacking) {
    for (i = 0, length = this._objects.length; i < length; ++i) {
      this._draw(ctx, this._objects[i]);
    }
  } else {
    for (i = 0, length = this._objects.length; i < length; ++i) {
      if (this._objects[i] && !activeGroup.contains(this._objects[i])) {
[...]

PR #1962, opened 2015-02-05
Proper rendering order of activeGroup

jagi:
Resolves issue #508


asturur:
How many time will you render the activegroup in this way? each for every object contained ?

jagi:
Oh right, I've forgot about this. I will fix this in a moment.

PR #1963, opened 2015-02-05
Fix to activeGroup wrong selection order

jagi:
Fixes issue #508

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -725,7 +725,7 @@
       if (!object) {
         return;
       }
-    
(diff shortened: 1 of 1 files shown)


asturur:
@jagi with this PR you are ignoring the preserveObjectStacking option and setting it always to true.
look at #2130 for a fix of #508.
As you can see the code in renderAll is very complex for what it does and needs to be more simple and straightforward, something like this PR should be done, but for general improving and with a wider look on all functionalities. Thanks for having a look!

PR #1978, opened 2015-02-18
File is removed text.cufon is removed

boonkerz:


asturur:
i remember i deleted that line....

kangax:
Thanks

PR #1982, opened 2015-02-20
Fix aligning guidelines on zoom

sorich87:
Aligning guidelines are not displayed at the correct position when canvas is zoomed. This pull request fixes it.


asturur:
would not be better use viewporttransform [0] and [3] without adding the zoom factor?

is not viewporttransform the matrix that represent current canvas zoom level and panning?

i remember get zoom is internally using that matrix.
(i m from smartphone, otherwise i would check, consider it a note for later check :) )

sorich87:
It should be good now :grinning:

PR #1988, opened 2015-02-24
Fix base throwing error when strokeDashArray unitialized

KJTsanaktsidis:
Other places in the codebase guard the call to ctx.setLineDash,
so this should happen here too. Fixes https://github.com/kangax/fabric.js/issues/1986.


kangax:
Thanks!

PR #1989, opened 2015-02-24
Fixed race condition in the iText cursor flash animation

KJTsanaktsidis:
The animation that flashes the iText cursor can race its own
cancellation, meaning that you can have two _tick()'s and two
_tickComplete() animations at the same time. On iOS in particular,
most likely due to the lower frequency of timeouts, this can trap
the cursor in a state where its opacity is always <0.1 and is
essentially invisible.

This PR captures the abort state of _tick in such a way that
one can abort the existing _tick and then immediately call _tick()
again safely.

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -61,10 +61,13 @@
      * @private
      */
     _tick: function() {
-      if (this._abortCursorAnimation) {
-        return;
-      }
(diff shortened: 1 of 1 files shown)


kangax:
Ah interesting. Thanks for catching that (no pun intended) :)

P.S. I wonder why JSHint+JSCS didn't catch trailing comma...

PR #1990, opened 2015-02-24
Improve the way the IText hidden textarea is positioned.

### LINKED ISSUE #1985: Focus craziness on iOS with iText
In the current fabric release, typing text in an iText in Chrome causes the window to scroll to the top of the screen. In order to fix this, I cherry picked https://github.com/kangax/fabric.js/commit/b716a56e2cbbaa3c53fab9a3ed4fa9459b93045e into my copy of fabric.js. This fixed the focus issue on Chrome, but led to even worse things on iOS.

With the above cherry picked into my fabric, typing in iText on iOS causes the window to scroll to the bottom of the screen. This is because the textarea is `position: fixed;`, which on iOS allows the page to scroll behind the position fixed thing.

To fix this in our project, I have duck-punched window.fabric.iText.prototype.initHiddenTextarea to put the hidden textarea at the same vertical position as the iText. When the iText moves relative to the DOM for whatever reason, the position of the hidden textarea should be updated.

My monkey patch looks like this.
```javascript
//Apply violence to the stupid hidden input stuff to make iOS show the keyboard

(function(){
    var __initHiddenTextarea = window.fabric.IText.prototype.initHiddenTextarea;

    function initHiddenTextarea(){

        this.hiddenTextarea.style.cssText = 'position: absolute; opacity: 0; font-size: 1pt;' +
                                            ' left: -9999px; width: 8888px; height: 10px; z-index: -999;';
        if (this.canvas && this.canvas.wrapperEl) {
            this.canvas.wrapperEl.appendChild(this.hiddenTextarea);

            this.on('event:scaling', this.canvas.updateHiddenTextareaPosition.bind(this.canvas));
            this.on('event:moving', this.canvas.updateHiddenTextareaPosition.bind(this.canvas));
            this.canvas.updateHiddenTextareaPosition();
        } else {
        }

        //This was in the original initHiddenTextarea

        if (!this._clickHandlerInitialized && this.canvas) {
          fabric.util.addListener(this.canvas.upperCanvasEl, 'click', this.onClick.bind(this));
          this._clickHandlerInitialized = true;
        }
    }

    //We need to slap this on fabric, so that if we do something to change the position of the
    //textarea on the screen by e.g. resizing or zooming the canvas, we can move the hidden
    //textarea appropriately.
    function updateHiddenTextareaPosition() {
        //The text's bounding rectangle, IN CANVAS SPACE (not fabric logical coordinates)
        var itextObject = this.getActiveObject();
        if (itextObject && itextObject.isEditing) { //If not iText, will return false
            if (itextObject.hiddenTextarea.parentNode === this.wrapperEl) {
                var rect = itextObject.getBoundingRect();
                itextObject.hiddenTextarea.style.top = rect.top + 'px';
            }
        }
    }

    window.fabric.util.object.extend(window.fabric.IText.prototype, {
        initHiddenTextarea: initHiddenTextarea,
    });

    window.fabric.util.object.extend(window.fabric.Canvas.prototype, {
        updateHiddenTextareaPosition: updateHiddenTextareaPosition,
    });
})();
```

This seems to be working for me as expected. My questions are twofold.
- Are there any other circumstances in which `canvas.updateHiddenTextareaPosition()` will need to be called? Zooming comes to mind, but I'm not sure what the right events are to listen on.
- Is this an approach which the fabric maintainers agree with? If so, I'm happy to convert this into a proper PR

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8813717-focus-craziness-on-ios-with-itext?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KJTsanaktsidis:
Previously, it was positioned as position: fixed at the bottom of
the window. However, on iOS, this has the effect of scrolling the
page all the way to the bottom when the user types. This is because
position: fixed in iOS allows the page to scroll behind the fixed
element if there is a focused input in it.

Now, we put the textarea off to the left at the same height as the
top of the iText; there should be no scrolling. We also make some
attempt to update the position when the text moves on the screen.

Fixes #1985

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -18,10 +18,10 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       var updateHiddenTextareaPosition = function () {
         if (this.isEditing && this.canvas.getActiveObject() === this) {
           //The text's bounding rectangle, IN CANVAS SPACE (not fabric logical coordinates)
-          var rect = this.getBoundingRect();
+          var rect = this.getBoundingRect(),
 
           // Compute the scale transform between fabric coords and DOM coords
-          var canvasRect = this.canvas.lowerCanvasEl.getBoundingClientRect(),
+            canvasRect = this.canvas.lowerCanvasEl.getBoundingClientRect(),
             xScale = canvasRect.width / this.canvas.width,
             yScale = canvasRect.height / this.canvas.height;
 
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -19,10 +19,16 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
         if (this.isEditing && this.canvas.getActiveObject() === this) {
           //The text's bounding rectangle, IN CANVAS SPACE (not fabric logical coordinates)
           var rect = this.getBoundingRect();
-          this.hiddenTextarea.style.top = rect.top + 'px';
-          this.hiddenTextarea.style.left = rect.left + 'px';
-          this.hiddenTextarea.style.width = rect.width + 'px';
-          this.hiddenTextarea.style.height = rect.height + 'px';
+
+          // Compute the scale transform between fabric coords and DOM coords
+          var canvasRect = this.canvas.lowerCanvasEl.getBoundingClientRect();
+          var xScale = canvasRect.width / this.canvas.width;
+          var yScale = canvasRect.height / this.canvas.height;
+
+          this.hiddenTextarea.style.top = rect.top * xScale + 'px';
+          this.hiddenTextarea.style.left = rect.left * yScale + 'px';
+          this.hiddenTextarea.style.width = rect.width * xScale + 'px';
+          this.hiddenTextarea.style.height = rect.height * yScale + 'px';
         }
       }.bind(this);
 
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -21,9 +21,9 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
           var rect = this.getBoundingRect();
 
           // Compute the scale transform between fabric coords and DOM coords
-          var canvasRect = this.canvas.lowerCanvasEl.getBoundingClientRect();
-          var xScale = canvasRect.width / this.canvas.width;
-          var yScale = canvasRect.height / this.canvas.height;
+          var canvasRect = this.canvas.lowerCanvasEl.getBoundingClientRect(),
+            xScale = canvasRect.width / this.canvas.width,
+            yScale = canvasRect.height / this.canvas.height;
 
           this.hiddenTextarea.style.top = rect.top * xScale + 'px';
           this.hiddenTextarea.style.left = rect.left * yScale + 'px';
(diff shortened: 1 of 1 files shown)


asturur:
why width 8888px?
also could you try a situation in wich the page is scrolled to the right, if editing itext scroll.page to left? this would revert to old behavior.

Also i do not know if you can move or resize the itext while editing.

KJTsanaktsidis:
I found the hidden textarea needed a width, because otherwise the cursor could wind up quite far down the page while doing selections on the iText. As for why 8888px, just because it was smaller than 99999px.

You raise a very valid point about what happens during horizontal scrolling. I have not tested this scenario, so I will do so tomorrow. Unfortunately if you put the textarea within the page viewport, even if it is behind everything, it will still show the cursor on iOS. This seemed to be the best solution I found, but I'll have to do something about horizontal scrolling.

I'm more concerned about what happens when the canvas is resized/zoomed while editing text than what happens if the text is moved. I'm not sure what events to hook to fix this, what would you suggest?

asturur:
i do not know if canvas zooming trigger any event on objects ( fabric zoom ).

would not standard width of textbox be enough?
maybe 0 was to aggressive, it was a try to have ot on the page but to not click it by chance.

so bad fixed position is not going on ios! it looke like a problem gone :(

@kangax has an ipad

KJTsanaktsidis:
@asturur you are indeed correct; this solution had problems with horizontal scrolling.

I have now found a way to hide the cursor on iOS (you need to set the font-size to zero). Thus, we can position the hidden textarea directly behind the iText object without having the flashing cursor. I have updated my PR to do this.

I have tested this approach on iOS 7 & 8, Chrome & Firefox on OSX, and IE9 & 11. I saw no unexpected scrolling happening. In some browsers, typing when the textarea is not on the screen caused scrolling to the textarea to occur, but since the textarea now where the iText is, this behaviour is absolutely fine.

KJTsanaktsidis:
@asturur What do you think of this updated solution?

KJTsanaktsidis:
It occurs to me that the hidden textarea position should also be updated on canvas scale. I will include something to this effect tomorrow.

coulix:
This wont work in the future with keypress events since double-byte/multi-byte characters such as Korean/Chinese do not trigger keypress.

When I cherry pick this PR in my 1.4.13 tag I get a strange behavior where I need to first emit a keypress before the next keypress is captured. The cursor also seems to blink randomly.

I am trying to put a finger on this IOS/Chrome auto scroll behavior.

KJTsanaktsidis:
@coulix The cursor mis timing is fixed by #1989

KJTsanaktsidis:
@kangax @asturur is there any prospect of having this looked at & merged?

kangax:
@KJTsanaktsidis sorry to keep you waiting! Before merging, I really want to dig into this a little more and try it out myself to see how it works. Hopefully in the next few days!

asturur:
I was looking into it also, and was aking myself if it was possibile to position the hidden textarea using the mouse coorditanes of the user click event userd to enter editing.

Would it not be simpler? or am i wrong?

We are positioning the textarea in the same position of iText right?
Why we need to update it? there is way to move the itext without exiting and entering edit?

KJTsanaktsidis:
The idea of putting the hidden textarea in the position the mouse was clicked is probably not a bad one. One thing that worries me is what happens if you have a big iText, click the top, and type many lines, so that eventually you cannot fit the initial textarea location and the place where the cursor is on the same screen. Whenever you type into the iText in this situation, I think it will scroll you back to the top of the iText. Do we therefore need to update the position of the textarea when the cursor moves to always be behind the current cursor location?

KJTsanaktsidis:
I believe on an iPad this would also have the effect of keeping the current line you are typing on the screen, and scrolling the page behind it as you type lines.

asturur:
replaced with #2868

PR #1991, opened 2015-02-24
Add snap for rotation

sapics:


asturur:
My first question is, why per canvas and not per object?

sapics:
Thank you for you answer.
As you noticed, per object is better for general.
I close this request.

PR #1998, opened 2015-02-25
Revert "Add forcedfocus on enterEditing"

asturur:
Reverts kangax/fabric.js#1961

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -335,7 +335,6 @@
       this.isEditing = true;
 
       this.initHiddenTextarea();
-      this.hiddenTextarea.focus();
       this._updateTextarea();
       this._saveEditingProps();
       this._setEditingProps();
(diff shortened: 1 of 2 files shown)


kangax:
Do we need this on?

asturur:
absolutely.
This was here for a reson, and it is not related to android keyboard as in the comment.
EnteringEditing we put focus on the hiddenTextarea, and this was an improvement.
Removing this function that we are reverting, causes the focus  to disappear as soon as we click again the iText object in the editing mode.
Because it won t trigger another enterEdit, and the focus will go away from hidden Text Area.

src/mixins/itext_behavior.mixin.js
```diff
@@ -335,7 +335,6 @@
       this.isEditing = true;
 
       this.initHiddenTextarea();
-      this.hiddenTextarea.focus();
```
asturur: wait this has to stay.

kangax:
@asturur can you plz merge yourself once it's ready

PR #1999, opened 2015-02-26
Make toJSON/toSVG behave sensibly with an active selection group

KJTsanaktsidis:
Calling toJSON or toSVG on a canvas whilst there is an active selection group causes it to be destroyed & recreated. This causes a number of nasties to happen, as described in https://github.com/kangax/fabric.js/issues/1876. This PR changes the way these methods work to avoid needing to do this; instead of realising the result of the group's transform on its objects by destroying the group, the group gains a method for doing this explicitly, and toJSON/toSVG pass clones of selected objects into this method.

This is a version of @slashgrin's fiddle including this patch, demonstrating that the problem has gone:
http://jsfiddle.net/2yy2t5nw/6/

These fiddles verify that toJSON/toSVG seem to be doing the right thing:
http://jsfiddle.net/2yy2t5nw/7/
http://jsfiddle.net/zsc1p82w/8/

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -289,7 +289,7 @@
     */
     realizeTransform: function(object) {
       this._moveFlippedObject(object);
-      this._restoreObjectState(object);
+      this._setObjectPosition(object);
       return object;
     },
     /**
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -280,6 +280,19 @@
     },
 
     /**
+     * Realises the transform from this group onto the supplied object
+     * i.e. it tells you what would happen if the supplied object was in
+     * the group, and then the group was destroyed. It mutates the supplied
+     * object.
+     * @param {fabric.Object} object
+     * @return {fabric.Object} transformedObject
+    */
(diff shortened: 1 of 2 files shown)


KJTsanaktsidis:
It looks like calling clone while doing toJSON/toSVG is a bad idea for async objects. The solution seems to be to mutate and then restore the actual object being serialized.

KJTsanaktsidis:
@kangax @asturur I'd love some feedback on this PR- is what I've done here a good idea or does it need some improvement?

kangax:
@KJTsanaktsidis sorry, didn't have time to look through this yet. Will do soon!

KJTsanaktsidis:
It turns out that since the group calls `set()` on the object to apply the relevant transforms, when these transforms are unwound, `set()` should also be called. This ensures that if you are watching any properties by overriding `set`, you see that the group transform gets undone.

kangax:
I like the idea and the implementation! Docs look good too. Thanks @KJTsanaktsidis, merging.

kangax:
Btw, I don't remember if we have tests for this functionality; if we don't, would be awesome to add a few.

PR #2007, opened 2015-03-01
Itext selection on canvas

asturur:
This PR allows the user to select text moving mouse outside the iText object bounding box

@kangax i don't get if i have to de-register the event on mouseup or on exit editing or if is not necessary.
Is not clear to me.

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -342,27 +342,30 @@
 
       this._tick();
       this.fire('editing:entered');
-      if (this.canvas) {
-        var _this = this;
-        this.canvas.renderAll();
-        this.canvas.fire('text:editing:entered', { target: this });
-        this.canvas.on('mouse:move',  function(options) {
 
-          if (!_this.__isMousedown || !_this.isEditing) {
-            return;
-          }
-
-          var newSelectionStart = _this.getSelectionStartFromPointer(options.e);
-          if (newSelectionStart >= _this.__selectionStartOnMouseDown) {
-            _this.setSelectionStart(_this.__selectionStartOnMouseDown);
-            _this.setSelectionEnd(newSelectionStart);
-          }
-          else {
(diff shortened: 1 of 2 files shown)


src/mixins/itext_behavior.mixin.js
```diff
@@ -341,10 +341,28 @@
       this._setEditingProps();
 
       this._tick();
-      this.canvas && this.canvas.renderAll();
-
       this.fire('editing:entered');
-      this.canvas && this.canvas.fire('text:editing:entered', { target: this });
+      if (this.canvas) {
```
kangax: Hm, why not have it in a separate method?

asturur: you mean the "function(options) {}" ?

i just moved it from other section. i was enough happy that it was already working.

anyway an early return is something to do if !this.canvas

kangax: Yes, I thought about early return at first, coz it's simple and makes it easier to read afterwards. But then I saw you removed `initMousemoveHandler` and I realized that having all that code in a separate function is even better.

asturur:
@kangax i restored the method. i didn't the meaning of your comment at first.

kangax:
Great, thanks

asturur:
But should we not deregister the event? on exitEdit ? Does not the mouse:move stack one over the other? First the event was registered at element creation, now is on element enterEdit, so multiple times.

kangax:
Updated fabricjs.com with latest changes but it doesn't seem to work for me — 
http://fabricjs.com/test/misc/itext.html

kangax:
You're right we should. Or it'll be adding up infinitely.

asturur:
on kitchensing is fine
https://www.youtube.com/watch?v=vL8IbuWyBqo
For derigisted i can use this.canvas.off, but how do i get a pointer to that event?

kangax:
It works now! Maybe needed time to update on a server.

PR #2022, opened 2015-03-08
Fix canvas width and height when canvas zoomed

sorich87:


sorich87:
Actually this pull request is wrong, it's not the canvas width and height that broken but the setWidth and setHeight methods of objects. See http://jsfiddle.net/1vbxh3m7/.

I'll try to submit another one. Any pointer would be appreciated though.

PR #2043, opened 2015-03-16
Antialiased clipping in Chrome

### LINKED ISSUE #1915: Feature Request: Mask Smoothing for Chrome [$100]
(See also https://groups.google.com/forum/#!topic/fabricjs/k_fFQ6Iu9dI)

When clipping masks are applied in fabric.js, they render smoothly (anti-aliased) in most browsers, except for Chrome.

An example of how this plays out on a basic (non-fabric) canvas is shown here: http://jsfiddle.net/cXJMW/

This is apparently related to a Chrome browser issue report that's been open for a while now at https://code.google.com/p/chromium/issues/detail?id=124756

For example, applying the following circle shape as a clip to an image in fabric.js will produce this effect as well:
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" width="400px" height="400px" viewBox="0 0 400 400" enable-background="new 0 0 400 400"> <circle fill="#aaaaaa" fill-opacity=".3" cx="200" cy="200" r="200"/> </svg>

Varying the fill and fill-opacity parameters will mitigate some of the effect, but it seems that an actual solution will require additional operations on the canvas clipping operation itself.

A pertinent (non-fabric) question (with a suggested solution provided by someone) can be found here: http://stackoverflow.com/questions/9536257/how-to-anti-alias-clip-edges-in-html5-canvas-under-chrome-windows

@kangax has expressed that he would be fine pulling this into fabric.js as a feature if someone takes on the implementation.

I'm putting up a bounty for the implementation as a fabric.js feature, to fix this Chrome rendering issue.

The implementation must also work with the recently-implemented ResizeFilters (see https://github.com/kangax/fabric.js/issues/1741) and with masked transparent and non-transparent images.

A fast resolution would be greatly appreciated!

Thank you!
## <bountysource-plugin>

There is a **[$100 open bounty](https://www.bountysource.com/issues/7399630-feature-request-mask-smoothing-for-chrome-100?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

golive:
Perform canvas and object clipping using blending. This will be rendered with antialias in all browsers, including Chrome.
Here is the same example with [antialiased clipping](http://jsfiddle.net/2yeLoves/1/) vs [current Fabric.js](http://jsfiddle.net/5c5r8df0/). Use Chrome to see the difference.

It also allows to define 2 clipping masks for canvas:

  * `canvas.clipAllTo` will always clip everything, including the background
  * `canvas.clipTo` will clip the objects but not the background if `clipBackground == false` or `clipAllTo` is defined

This is useful for example on T-Shirt custom designing: you define the background color and a mask with the shape of the shirt, and another mask to clip the contents of the printing area.

I will add examples for every case.

Fixes #1915

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -342,9 +342,7 @@
     _onMouseDownInDrawingMode: function(e) {
       this._isCurrentlyDrawing = true;
       this.discardActiveObject(e).renderAll();
-      if (this.clipTo) {
-        fabric.util.clipContext(this, this.contextTop);
-      }
+
       var ivt = fabric.util.invertTransform(this.viewportTransform),
           pointer = fabric.util.transformPoint(this.getPointer(e, true), ivt);
       this.freeDrawingBrush.onMouseDown(pointer);
@@ -366,6 +364,7 @@
             pointer = fabric.util.transformPoint(this.getPointer(e, true), ivt);
         this.freeDrawingBrush.onMouseMove(pointer);
       }
+      fabric.util.clipContext(this.clipTo || this.clipAllTo, this.contextTop);
       this.setCursor(this.freeDrawingCursor);
       this.fire('mouse:move', { e: e });
 
@@ -381,9 +380,6 @@
      */
     _onMouseUpInDrawingMode: function(e) {
       this._isCurrentlyDrawing = false;
-      if (this.clipTo) {
-        this.contextTop.restore();
-      }
       this.freeDrawingBrush.onMouseUp();
       this.fire('mouse:up', { e: e });
 
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -226,14 +226,13 @@
       }
 
       ctx.save();
-      this.clipTo && fabric.util.clipContext(this, ctx);
 
       // the array is now sorted in order of highest first, so start from end
       for (var i = 0, len = this._objects.length; i < len; i++) {
         this._renderObject(this._objects[i], ctx);
       }
 
-      this.clipTo && ctx.restore();
+      fabric.util.clipContext(this.clipTo, ctx);
 
       ctx.restore();
     },
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -988,9 +988,8 @@
       }
       this._setOpacity(ctx);
       this._setShadow(ctx);
-      this.clipTo && fabric.util.clipContext(this, ctx);
       this._render(ctx, noTransform);
-      this.clipTo && ctx.restore();
+      fabric.util.clipContext(this.clipTo, ctx);
       this._removeShadow(ctx);
       this._restoreCompositeOperation(ctx);
 
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -475,9 +475,8 @@
       }
       this._setOpacity(ctx);
       this._setShadow(ctx);
-      this.clipTo && fabric.util.clipContext(this, ctx);
       this._render(ctx, noTransform);
-      this.clipTo && ctx.restore();
+      fabric.util.clipContext(this.clipTo, ctx);
       this._removeShadow(ctx);
       this._restoreCompositeOperation(ctx);
 
--- a/src/shapes/path_group.class.js
+++ b/src/shapes/path_group.class.js
@@ -86,11 +86,10 @@
       this.transform(ctx);
 
       this._setShadow(ctx);
-      this.clipTo && fabric.util.clipContext(this, ctx);
       for (var i = 0, l = this.paths.length; i < l; ++i) {
         this.paths[i].render(ctx, true);
       }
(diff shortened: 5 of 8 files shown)


asturur:
To me it looks good, also it includes the PR i proposed some time ago and that is marked as possible feature. (#1875)

Could you explain me the use of globalComposite operation?
Is this enabling the anti-aliasing in chrome?

golive:
Thank you. Yes, this is currently [the only way](https://medium.com/prezi-engineering/rendering-f857b6e09551) to perform clipping with antialiasing in Chrome. I don't think that Google will fix standard canvas `clip()` anytime soon, as this has been discussed for a while now ([2009](https://code.google.com/p/chromium/issues/detail?id=7508), [2012](https://code.google.com/p/chromium/issues/detail?id=124756)). Even [here](https://www.imperialviolet.org/2009/09/02/anti-aliased-clipping.html) is suggested that this is more like a feature than a bug.

With `globalCompositeOperation`, what you do is to tell to the canvas how to blend the subsequent paths in relation of what there is currently painted in the canvas. [Here](https://developer.mozilla.org/samples/canvas-tutorial/6_1_canvas_composite.html) you can see some examples.
The composition mode that I am using here is 'destination-in'. One benefit of this approach is that it works with images too.

asturur:
Yes i know the use og globalCompositeOperation, i was asking myself WHY it makes the antialiasing.

You should make an additional test, making a cliptest with holes, i m searching a shape but i cannot find one to send you for test.

Please could you try to clip with a shape that has an hole?

Anyway i feel like that:
ClipAllTo is too much, better to choose just clipBackground true or false.
Every object has is own clipTo property.
if clipTo is checked inside clipContext i would not check it also before calling clipContext

```
if (this.clipBackground) {
        fabric.util.clipContext(this.clipTo, canvasToDrawOn);
       }
```

I would pull it just for the code cleanup and clipbackground option, i m very unsure about that extra fill and that extra canvas save and restore operation.

Anyway kangax will check it.

golive:
I created the `clipAllTo` function to allow to define 2 clipping masks. This way, it is possible to achieve something like [this](http://jsfiddle.net/sL1jto7p/1/), where the background is dinamically defined. I know is not a common case, but it will add a lot of flexibility.

asturur:
i do not get the point, i will check it better as soon as i can.

jardakotesovec:
I suspect that with using destination-in it will clear everything that was created on canvas before the clipping. So its necessary to use scratch-canvas to use that.

For some reason fork/update on your jsfiddle did not work.. don't know why..   Here is updated code to show my point:
And it illustrates that you can see only triangle2. 

```
  var canvas = new fabric.Canvas('c');
  canvas.backgroundColor = 'gray';
  canvas.controlsAboveOverlay = true;
  
var triangle = new fabric.Triangle({
  width: 20, height: 30, fill: 'blue', left: 50, top: 50
});

var triangle2 = new fabric.Triangle({
  width: 20, height: 30, fill: 'red', left: 70, top: 50
});

  fabric.Image.fromURL('http://fabricjs.com/assets/pug_small.jpg', function(img) {
    img.scale(0.5).set({
      left: 100,
      top: 100,
      clipTo: function (ctx) {
        ctx.arc(0, 0, 180, 0, Math.PI * 2, true);
      }
    });
     canvas.add(triangle)
     canvas.add(img).setActiveObject(img);
     canvas.add(triangle2)
  });

```

jardakotesovec:
Here is nice [explanation](http://stackoverflow.com/a/13854185/1611208)

golive:
From your example, I can see that object clipping doesn't work as expected when you paint another object before the clipped one, you are right. I have to take a look on that.
But removing object clipping, global canvas clipping works ok. I think it is more related on how objects are painted and saving/restoring canvas state than the composition mode itself. No need to use another canvas.

asturur:
Other than that, you should consider path with holes.
Can you try path with holes and see what happens?
Because there is some difference that can be a bug or added featuer as you wish.

I try to explain:
SVGs has a property that is fill-rule.

in SVGs that property influence also clipping, so if you have a path with an hole, if the fill rule property is even-odd OR non-zero the clipping mask changes.

In canvas this property is not a parameter for clipping and is fixed to even-odd, while for filling is a parameter.

In other words in cavas a path with an hole in the center would clip always same way, but can fill in two ways. So clipping by filling would result in a non predictable result.

kangax:
This looks like a properly done PR, so I'm open to reviewing and merging it. But could you please resubmit with latest changes? Thanks.

asturur:
i will take a look at it if you do not mind. there was something wrong in my opinion mixing globalcomposite with clipping.

PR #2055, opened 2015-03-24
Track IText instances per-canvas, not globally.

slashgrin:
Previously instances of the IText shape were added to a globally-shared
array when they were created. There are two problems with this approach:

1) Interactions with one canvas affect others. I would never expect
   text in one canvas to exit edit mode just because I interacted with
   some otherwise-unrelated canvas.

2) Every IText instance leaks. There is no mechanism to clean up references
   to IText instances in the global array, so every such instance will
   hang around in memory forever, regardless of whether it is removed from
   the canvas or if the canvas itself is removed.

Discovered while profiling memory usage in Chrome.

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -9,6 +9,7 @@
      */
     initBehavior: function() {
       this.initAddedHandler();
+      this.initRemovedHandler();
       this.initCursorSelectionHandlers();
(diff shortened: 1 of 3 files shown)


asturur:
if you wanna check, tjere is something else to adjust about leaking. on enter editing i registered some mouse move event on the canvas, could you check how to deregister it from the canvas on exit editing?

PR #2062, opened 2015-03-27
Updated fabric.js, added mouseout on upperCanvas element

khawerrind:
There are **mouseover** & **mouseout** events for objects, which works fine. But there is a case where **mouseout** on an object does not trigger. Lets assume we have an overlay div on the main canvas and that div is placed right below the object. Now when you hover over the object **mouseover** gets triggered. But when you move your mouse to the overlay div the **mouseover** does not trigger. That is because these events are bind to **upper-canvas** and when mouse is on the overlay div the **mouseover** and **mouseout** does not gets trigger. 

Or lets forget about overlay div. Lets assume the object is on the left. And when you mouse out of the object to left in this [fiddle](https://jsfiddle.net/45vfbgw1/2/) you will notice the **mouseout** does not trigger. Because now mouse is out of the main canvas.

So i have added a **mouseout** event for **upper-canvas**. When this event gets trigger it will call the **mouseout** of the last **mouseover** object.

I have created 2 fiddles to explain the issue.

[Fiddle using the current fabirc.js:](https://jsfiddle.net/45vfbgw1/) In this fiddle you can see when you **mouseover** the big red box the **mouseover** event triggers. But when you move the mouse right below the object to the overlay div, the **mouseout** does not trigger.

[Fiddle with solution:](http://jsfiddle.net/wm69g3ed/)

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -9142,6 +9142,7 @@ fabric.PatternBrush = fabric.util.createClass(fabric.PencilBrush, /** @lends fab
       addListener(this.upperCanvasEl, 'mousedown', this._onMouseDown);
       addListener(this.upperCanvasEl, 'mousemove', this._onMouseMove);
       addListener(this.upperCanvasEl, 'mousewheel', this._onMouseWheel);
+      addListener(this.upperCanvasEl, 'mouseout', this._onMouseOut);
 
       // touch events
       addListener(this.upperCanvasEl, 'touchstart', this._onMouseDown);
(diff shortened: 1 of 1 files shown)


khawerrind:
These issues are related to my pull request. https://github.com/kangax/fabric.js/issues/1880

dnanusevski:
I have downloaded fabric build 1.5.0 and i am expiriancing same problems. 
Is there something that i am missing. I am using
 canvas.on('mouse:out', function(e) {e.target.set({ opacity: 1}); canvas.renderAll();}); 
Should i use different event handler ?

I should note that after i manualy change fill option of an element, then "mouse:out" is working as it should. Just like on working fiddle   http://jsfiddle.net/wm69g3ed/

dnanusevski:
I am still having isue when mause leaves throw bottom or right side of canvas.
I have downloaded last version of fabric.js and i am using
 canvas.on('mouse:over', function(e) {e.target.set({ opacity: 0.1}); canvas.renderAll();});
Am i missing something ?

PR #2064, opened 2015-03-30
Fix doc for setAngle

sapics:
Sorry, this PR is not affect to the feature...

1. src/shapes/object.class.js
In docs for fabric.Object and its extended class,
there are two definition for setAngle.
http://fabricjs.com/docs/fabric.Object.html#setAngle

2. HEADER.js
fabric.SHARED_ATTRIBUTES is used for parsing svg.
Therefore, it would be unnecessary for no parser module build.

3. src/mixins/canvas_events.mixin.js
Remove unused variable.


kangax:
Ah, nice. Thanks!

PR #2065, opened 2015-03-30
Render cursor on upper canvas

### LINKED ISSUE #2016: High CPU when IText element selected - bug?
Hi,

Creating this issue as per @kangax request in this post: https://groups.google.com/forum/#!topic/fabricjs/erhDXOq1EWQ

I'm seeing high CPU usage that remains high for no apparent reason even if there's no activity, on a browser process when there's a Fabric IText element on the page that's in edit mode, i.e. double clicked so flashing cursor appears, or double-clicked and dragged so a portion of text is hilighted. So long as the element remains in that mode (i.e user doesn't click elsewhere on the actual Canvas to deselect the IText element) and even if there is no other user activity or visible browser activity, CPU usage just sits high, computer fan comes on stronger and stronger. Tested in latest Chrome (CPU usage hovers around 50% just for that one browser window), latest Firefox (somewhere between 25 and 50%), latest Opera (40ish %) and unexpectedly a bit lower in IE 11 (about 15 - 25%). I'm running Windows 7, 32 bit, if that makes any difference...

You can try this in the kitchensink demo but it also applies on a very simple demo canvas with a single IText element with "hello world" in it.

Is this a bug? It should be :) I can understand high CPU usage potentially for moving objects around or other processing, but not just a selected IText element apparently doing nothing.
(Note this doesn't happen when it's single-selected so that handles are visible).

Kangax says in above linked thread that it's because entire canvas is re-rendered during cursor-tick. I am happy for cursor tick to be switched off (e.g. static but visible) if that helps!

Thanks,
Alex

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9228830-high-cpu-when-itext-element-selected-bug?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
This PR mainly avoid calling canvas.renderAll during iText cursor animation.

Both the tick and the render cursor or selection check if there is available contextTop and use that or standard ctx.

It looks very simple, i hope i did not introduce any bug
Can you think of some interference with other upper canvas drawing?
( overlay image? )

closes #2016

### DIFF
--- a/src/mixins/itext_click_behavior.mixin.js
+++ b/src/mixins/itext_click_behavior.mixin.js
@@ -198,7 +198,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
 
       if (this.flipX) {
         // when oject is horizontally flipped we reverse chars
-        this._textLines[i] = line.split('').reverse().join('');
+        this._textLines[i] = line.reverse().join('');
       }
 
       for (var j = 0, jlen = line.length; j < jlen; j++) {
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -476,16 +476,17 @@
      * @param {Object} boundaries
      */
     renderCursor: function(boundaries) {
-      var ctx = this.ctx;
+      var ctx = this.canvas.contextTop || this.ctx;
 
       ctx.save();
+      this.transform(ctx);
 
       var cursorLocation = this.get2DCursorLocation(),
           lineIndex = cursorLocation.lineIndex,
           charIndex = cursorLocation.charIndex,
           charHeight = this.getCurrentCharFontSize(lineIndex, charIndex),
           leftOffset = (lineIndex === 0 && charIndex === 0)
-                    ? this._getCachedLineOffset(lineIndex, this.text.split(this._reNewline))
+                    ? this._getCachedLineOffset(lineIndex)
                     : boundaries.leftOffset;
 
       ctx.fillStyle = this.getCurrentCharColor(lineIndex, charIndex);
@@ -506,9 +507,10 @@
      * @param {Object} boundaries Object with left/top/leftOffset/topOffset
      */
     renderSelection: function(chars, boundaries) {
-      var ctx = this.ctx;
+      var ctx = this.canvas.contextTop || this.ctx;
 
       ctx.save();
+      this.transform(ctx);
 
       ctx.fillStyle = this.selectionColor;
 
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -88,8 +88,16 @@
      * @private
      */
     _tick: function() {
+      this._currentTickState = this._animateCursor(this, 1, this.cursorDuration, '_onTickComplete');
+    },
 
-      var tickState, _this = this;
+    /**
+     * @private
+     */
+    _animateCursor: function(obj, targetOpacity, duration, completeMethod) {
+
+      var tickState;
+
       tickState = {
         isAborted: false,
         abort: function() {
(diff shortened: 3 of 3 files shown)


kangax:
Hmm... looks good to me! I don't think it's gonna interfere with `overlayImage` since (as far as I remember) we render everything on the bottom canvas; the top one is only for mouse/touch-based selection (and now cursor).

kangax:
Have you tested it on http://fabricjs.com/test/misc/itext.html?

asturur:
yes, flipped cursor detection is not working, not in fabricjs.com actually. So i do not think is fault of this commit.
Also i wanted to remove the extra renderAll when we are just moving the cursor, but is making the cursor disappear for one second.
So bad, cause is a shame to call renderAll when just moving cursor around.

ajck:
(I posted #2016)
1.) Just wondering, if this affects or is affected by, #1778 which I'm hoping to use too and which I believe is going in the 1.5 release(?), though it subclasses IText so hope no impact...

2.) (Sorry newbie question) How do I use this update ( #2016, or #1778 ) in my code?

Thanks

asturur:
Removede some code duplication
removed some other canvas.renderAll();
Teorically we could remove canvas.renderAll(); after keyDown event, but is not scope of this PR.
Also fixed a little bug with text flipped that i introduced in refactoring.
As always i found some new jscs difficulties.

src/mixins/itext_click_behavior.mixin.js
```diff
@@ -198,7 +198,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
 
       if (this.flipX) {
         // when oject is horizontally flipped we reverse chars
-        this._textLines[i] = line.split('').reverse().join('');
```
asturur: this was a bug causing error.
Anyway i found this reversing string useless.

src/shapes/itext.class.js
```diff
@@ -476,16 +476,17 @@
...
 
       ctx.save();
+      this.transform(ctx);
 
       var cursorLocation = this.get2DCursorLocation(),
           lineIndex = cursorLocation.lineIndex,
           charIndex = cursorLocation.charIndex,
           charHeight = this.getCurrentCharFontSize(lineIndex, charIndex),
           leftOffset = (lineIndex === 0 && charIndex === 0)
-                    ? this._getCachedLineOffset(lineIndex, this.text.split(this._reNewline))
```
asturur: This was part of iText rendering changes, forgot to remove it

ajck:
The high CPU issue is still there in 1.5.0 unfortunately, as per #2016 - it's whenever the cursor is being animated, fading in and out - I'm seeing my CPU usage jump to 40% from around 3%. Is there a way to stop the fade - i.e. just have a blink on/off every half second for example, or just a static cursor?

Many thanks,
Alex

asturur:
may i ask you what cpu are you using?
40% usage means high utilization of 1 core on 2 core machine, or more than 1 core using in a quad core machine or higher.

As of now there is no way to disable animation, what you can do is give to this.cursorDelay an high value, delay the start of the animation. Maybe in this way you will move the cursor before starting the animation. and the delay start again.

Can you post a screen of your canvas setup so we  can understand how many objects are being drawn?

ajck:
It's a dual Pentium T3400 at 2.16 GHz, laptop's a few years old.
Canvas is empty except for a single IText with 'hello world' in it. Nothing else happening on the page (e.g. scripts etc). High CPU only when cursor is animating, otherwise it's fine, around 3%.
I'm running in latest Google Chrome on Windows 7.

I saw cursorDelay and cursorDuration in the docs but thought they were read only?

What about the abortCursorAnimation() method? ("Aborts cursor animation and clears all timeouts") http://fabricjs.com/docs/fabric.js.html#line21943

asturur:
This function is just clearing some timeouts before starting again the animation, or interrupting the animation if you move the cursor with the keys.

You could override some function like _tick() to some not doing anything function, so the animation will simply not start.

I do not really have a ready solution nor time to think of it in the immediate future.
I hope some little hints can help you anyway

kangax:
I'm not sure why CPU would still jump to 40% if we're only rendering 1 tiny cursor (not entire canvas). Can someone profile and see what's going on? What's still being so computationally expensive during animation?

gordyr:
I haven't had a chance to look at the code but is he whole canvas being cleared each frame by any chance???

That would likely account for it.

We should just be clearing the drawn area.

Just a thought...

asturur:
the whole upperCanvas get cleared.
it can be optimized more for just the selection area.
but as of now we just make a context switch and call ctx.clear().

we are just avoiding to do renderAll()

take in account also that, if you are moving cursor, renderAll is called buy some function.

gordyr:
again, I haven't looked at the code but are we using requestanimationframe for the cursor rendering?

asturur:
no. fabircjs is not using request animation frame yet. not for cursor nor for generic rendering.

gordyr:
I would imagine that would be why then.

It will be rendering far more often than the Vsync rate.  Which is unnecessary. And will use a lot of cpu.

PR #2068, opened 2015-03-30
attemp to fix jscs crash

asturur:
Should not get this, like in the other build.

TypeError: Cannot read property 'type' of undefined
    at /home/travis/build/kangax/fabric.js/node_modules/jscs/lib/rules/require-multiple-var-decl.js:41:24
    at Array.forEach (native)
    at Object.JsFile.iterateNodesByType (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/js-file.js:406:42)
    at Object.consecutive [as _check] (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/rules/require-multiple-var-decl.js:37:10)
    at Object.module.exports.check (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/rules/require-multiple-var-decl.js:114:28)
    at null.<anonymous> (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/string-checker.js:133:18)
    at Array.forEach (native)
    at StringChecker._checkJsFile (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/string-checker.js:131:31)
    at StringChecker.checkString (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/string-checker.js:111:14)
    at null.<anonymous> (/home/travis/build/kangax/fabric.js/node_modules/jscs/lib/checker.js:43:21)

### DIFF
--- a/src/brushes/spray_brush.class.js
+++ b/src/brushes/spray_brush.class.js
@@ -123,10 +123,10 @@ fabric.SprayBrush = fabric.util.createClass( fabric.BaseBrush, /** @lends fabric
     this.canvas.renderAll();
   },
 
-    /**
-     * @private
-     * @param {Array} rects
-     */
+  /**
+   * @private
+   * @param {Array} rects
+   */
   _getOptimizedRects: function(rects) {
 
     // avoid creating duplicate rects at the same coordinates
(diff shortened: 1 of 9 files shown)


asturur:
Ok found out that var declaration under a switch case is crashing jscs, fortunately we have just one case in whole library.
Also i think that declaring variable outside the for loop is better than doing inside.
Fixed all the jscs wining about spaces and comments.
( who made those, and how sneaked past the jscs control is a mistery ).

PR #2069, opened 2015-03-31
Parse svg with 100% dimension or missing dimensions

### LINKED ISSUE #1711: Issue with control box
Just noticed this on http://fabricjs.com front page:

screenshot 2014-09-29 17 37 02

/cc @asturur 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4961181-issue-with-control-box?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1874: SVGs without Width and Height display at '0' size
To provide the most consistent behaviour between browsers, sometimes SVGs as images need to omit the `width` and `height` attributes. [See http://www.boogdesign.com/examples/svg/svg-as-image.html in Chrome and IE]

However there is inconsistent behaviour between browsers when adding such SVGs as Fabric Images to the Canvas. In Chrome, the SVGs appear to have a `naturalWidth` of `300` and `naturalHeight` of `150`. However in IE and FF, these values are `0`, and as such, when added to the canvas, aren't visible (however can be "found" by group selecting over the canvas.

http://jsbin.com/viluca/8/edit 

This isn't necessarily a _bug_ in the respective browsers or Fabric, however it would be _nice_ if Fabric handled it nicely.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/6710793-svgs-without-width-and-height-display-at-0-size?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
image

Of course is not working in Internet explorer, but i m too tired to understand why.

closes #1711 
closes #1874

### DIFF
--- a/src/shapes/path_group.class.js
+++ b/src/shapes/path_group.class.js
@@ -50,20 +50,45 @@
         this.paths[i].group = this;
       }
 
+      if (options.toBeParsed) {
+        this.parseDimensionsFromPaths(options);
+        delete options.toBeParsed;
+      }
       this.setOptions(options);
-
-      if (options.widthAttr) {
-        this.scaleX = options.widthAttr / options.width;
-      }
-      if (options.heightAttr) {
-        this.scaleY = options.heightAttr / options.height;
-      }
-
       this.setCoords();
 
       if (options.sourcePath) {
         this.setSourcePath(options.sourcePath);
       }
+    },
+
+    /**
+     * Calculate width and height based on paths contained
+     */
+    parseDimensionsFromPaths: function(options) {
+      var points, i, p, xC = [ ], yC = [ ], path, height, width,
+          m = this.transformMatrix;
+      for (var j = this.paths.length; j--; ) {
+        path = this.paths[j];
+        height = path.height + path.strokeWidth;
+        width = path.width + path.strokeWidth;
+        points = [
+          { x: path.left, y: path.top },
+          { x: path.left + width, y: path.top },
+          { x: path.left, y: path.top + height },
+          { x: path.left + width, y: path.top + height }
+        ];
+        for (i in points) {
+          p = points[i];
+          if (m) {
+            p = fabric.util.transformPoint(p, m, false);
+          }
+          xC.push(p.x);
+          yC.push(p.y);
+        }
+      }
+      options.width = Math.max.apply(null, xC);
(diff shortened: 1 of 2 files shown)


asturur:
It won't be easy.
Most browser will take naturlawidth and naturalheight at 150 / 300px.
I also tried to build dataurl from svg string and create an image from that and measuring the svg on the image.onload event. Same naturalwidth and naturaleight.

i could sniff all the object properties in a different place of the parser, but because of transformatrix coming from the SVG, the width and height of the object will be very complext to evaluate.

Still manageable ( taking all top , left , width, eight, multiply by transformmatrix, put all resulting coords in array and select min and max ) but will be huge code chunck.

I'll update you all.

asturur:
Works good togheter with #2086.

this is svg 171 with dimensions removed and parsed by the code.

image

asturur:
@kangax done

src/shapes/path_group.class.js
```diff
@@ -67,6 +63,35 @@
...
+        path = this.paths[j];
+        height = path.height + path.strokeWidth;
+        width = path.width + path.strokeWidth;
+        points = [
+          { x: path.left, y: path.top },
+          { x: path.left + width, y: path.top },
+          { x: path.left, y: path.top + height },
+          { x: path.left + width, y: path.top + height }
+        ];
+        for (i in points) {
```
kangax: if `points` is an array, it's better to iterate with `for(;;)` rather than `for..in`

kangax: other than that, LGTM! thanks a lot, as always

asturur:
@kangax i changed it

asturur:
@kangax what are you missing to pull this?

kangax:
Nothing, all good, thanks!

PR #2074, opened 2015-03-31
Fix hit detection on the active group after zooming.

rodovich:
The active group's coordinates don't get updated when the canvas's viewport transform changes. Therefore sometimes you can drag the group from outside of its visible bounds, and other times the mouse event will "miss" the group:

zoom-hit-testing

This PR fixes `setViewportTransform` to update the active group's coordinates (along with all the objects which are already updated there).


kangax:
Thanks!

PR #2083, opened 2015-04-04
Changes in rendering

asturur:
Moved the translate operation necessary for pathgroups ( SVGs ) in path group render.
Doing this we avoid calling an additional translate for every  object inside svg.

You can still see svgimport test with this build here:

www.deltalink.it/andreab/fabric/a.html
(left imported, center browser rendered, right is exported and browser rendered)

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -227,7 +227,7 @@
 
       ctx.save();
       this.clipTo && fabric.util.clipContext(this, ctx);
-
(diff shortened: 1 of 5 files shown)


asturur:
Same logi apply to this.transform, does not have to call parent group transform if we handle in the group render function.

asturur:
Group interaction can be tested here

www.deltalink.it/andreab/fabric/kitchensink.html

PR #2091, opened 2015-04-06
Shorter version of multiply transform matrix

asturur:
Shorter version of multiplyTransformMatrices.

Transform Matrices are always 6 elements over 9, so we can use a predefined multiplication schema without need to go around iterating over elements that are always 0 or 1.

Performance impact is more or less half.
i tried this test with kitchensink:
```
var a = [1, 2, 3, 4, 5, 6];
var b = [7, 8, 9, 10, 11, 12];
var c = [];
var start = new Date().getTime();
for(var i = 1; i < 1000000; i++) {
   c = fabric.util.multiplyTransformMatrices(a, b);
}
var end = new Date().getTime();
console.log(end - start);
console.log(c);
start = new Date().getTime();
for(var i = 1; i < 1000000; i++) {
   c = fabric.util.multiplyTransformMatricesFast(a, b);
}
end = new Date().getTime();
console.log(end - start);
console.log(c);
```
result is:
```
554  ms old version
Array [ 31, 46, 39, 58, 52, 76 ] 
258 ms new version
Array [ 31, 46, 39, 58, 52, 76 ]
```
Of course not a big deal, but faster, smaller, and free.
(chrome is 4 time slower than firefox in this test.


kangax:
Awesome!

PR #2093, opened 2015-04-08
avoid unnecesary save and restore per object

asturur:
Do not call save and restore before checking if we have to really draw borders or not.


asturur:
Diff is not  very  clear, i just inverted the border draw check for an early return instead of saving, checking if borders are to be draw, restore.

PR #2096, opened 2015-04-08
Changes in renderAll

asturur:
to be discussed and evaluated.
It just looks more clean, but maybe is just me.

What it does:

check if there is an activegroup and a not preserveObjectStacking.
in this case divide the objects in 2 groups, the normal and the one in activegroup.
Then renders the normal objects and later the activegroup it self.

Eliminates should render object, eliminates fastpath, eliminates the renderActiveGroup.
For sure it save some logic and processing, do not know if makes mess.

@kangax  @Kienz  if you want to have a look...

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -293,8 +293,8 @@
 
       target.hasBorders = target.transparentCorners = false;
 
-      this._draw(this.contextCache, target);
-
+      target.render(this.contextCache);
+      target._renderControls(this.contextCache);
       target.hasBorders = hasBorders;
       target.transparentCorners = transparentCorners;
 
@@ -1111,18 +1111,11 @@
     drawControls: function(ctx) {
       var activeGroup = this.getActiveGroup();
       if (activeGroup) {
-        this._drawGroupControls(ctx, activeGroup);
+        activeGroup._renderControls(ctx);
       }
       else {
         this._drawObjectsControls(ctx);
       }
-    },
-
-    /**
-     * @private
-     */
-    _drawGroupControls: function(ctx, activeGroup) {
-      activeGroup._renderControls(ctx);
     },
 
     /**
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -721,36 +721,6 @@
     },
 
     /**
-     * Given a context, renders an object on that context
-     * @param {CanvasRenderingContext2D} ctx Context to render object on
-     * @param {fabric.Object} object Object to render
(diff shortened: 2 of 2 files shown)


asturur:
Ok it works.
Even if it does not semplify as much as i wanted, is still a good cut to strange logic.

@kangax and @kienz, ready for your usual checkup

asturur:
Still i think we could extend the control up on the stack and save more code.

Active group controls are always on top, with preserveObjectStacking or NOT.
actually it is like that.
Also the internal control of every object in active group with preserveObjectStacking are anyway over the objects.

on the contrary Active object controls are on the same level of object level stack.

So we could semplify more the code and have alwyas controls up on the stack

we could avoid calling renderControls for every object and exit because it is not active.

Less code, more coherent.

asturur:
how is possible firefox is so faster?
image

asturur:
Purpose of this commit:
I did not like the logic of renderAll, it was for sure changed by commits and features in time, but to me it looked like messy now.

The logic was:
draw all objects one by one, for every object evaluate if we had to draw just the object or just the controls or both or even none of them.
Double check if object was null.

After that render active group that would sort the objects and draw them, checking if rendering just controls ( preserveObject stacking or not ) or control and object.

Every draw would save and restore the canvas and set the viewPort transformation.

Then after that there was a call to drawControls that was another function in charge of rendering just controls if upper overaly wanted to be used.

I changed in this way:
canvas transformation is done ONCE before starting the drawing of all objects.
all object are rendered first and then just the controls.
So in draw funcion nothing remains other than call to object.render, for this reason the  _draw function is deleted.

The drawControls function is used to draw controls in normal situation and in aboveOverlay situation.

There is a little difference now:
with preserveObjectStacking TRUE or FALSE the active group controls are always above the objects, even if the objects are down in the stack. On the contrary the single object control (activeObject) where following the stack pile. With those changes the controls of the single object are drawn on the same level of the activeGroup controls, and this look even logic to me.

Code cut is not small and testing with 1.4.13 i see also a small performance gain:

image


in random situation like this , with random paths and big selection i get for 4000 frames 9.5s vs 10.3s of the old code.
Not a big deal, i thing the main important factor is checking that nothing is broken.

asturur:
Add this screen as a remarks of why the controls should be rendered at the end always.
Otherwise controls get influced by globalCompositeOperation

image

asturur:
i close it, i will repropose it with better explanation for 1.5.2

PR #2102, opened 2015-04-09
Fix cursor AND selection

asturur:
Moved code of context selection and trasnformation to upper function.
@jafferhaider  i will not test it before this evening, if you want take some step further... you help me.

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -192,6 +192,7 @@
     _clearCache: function() {
       this.callSuper('_clearCache');
       this.__maxFontHeights = [ ];
+      this.__widthOfSpace = [ ];
     },
 
     /**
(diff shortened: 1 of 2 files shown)


asturur:
i'm working on that, i see is a big mess.
Anyway getting progresses

asturur:
@kanax:
Bugs left:
iText with styles and justify align, the style is wrong ( you can see on kitchensink)
Cursor position problem when abusing of spaces in justifymode
mixed use in the code of char = textline[i]  AND char = textline.split('')[i] which one should we use?
(https://github.com/asturur/fabric.js/blob/fix-cursor/src/shapes/itext.class.js#L803)

asturur:
@jafferhaider here link for test: http://www.deltalink.it/andreab/fabric/kitchensink.html

asturur:
I would add that  last line of text should not be justified ( like happens in normal text editor ). This should be a simple fix.

jafferhaider:
Typing is messed up. It doesn't let me enter a space when text align is set to justify.

asturur:
Can you share setup code to put in kitchensing?

asturur:
Typing was messed up even before, just in different way :)
http://www.deltalink.it/andreab/fabric/kitchensinkold.html
At least we can compare and take decision easier.

asturur:
we can still do differently, count effectively the number of spaces used and get the real widht of spaces.
This will cause different justification if too much spaces are used between words, but still better than hacking the text....

jafferhaider:
I'm sorry I don't understand what you mean by setup code. I just tested in the kitchensink link that you shared. I pasted in a paragraph of text, set it to justify and tried to type at the end of the paragraph.

PR #2120, opened 2015-04-16
Fix render preserve

asturur:
This fix make active group render ok with preserveObjectStacking.
Otherwise we should revert #2083.

It sound more reasonable to keep the render as it is and add this fix, expecially for big pathgroup of complex SVGs.


src/shapes/object.class.js
```diff
@@ -748,6 +748,9 @@
      * @param {Boolean} fromLeft When true, context is transformed to object's top/left corner. This is used when rendering text on Node
      */
     transform: function(ctx, fromLeft) {
+      if (this.group && this.canvas.preserveObjectStacking && this.group === this.canvas._activeGroup) {
```
inssein: This is breaking for me right now because this.cavas doesn't seem to be defined, which is odd because all groups should have a canvas.

inssein:
@asturur thanks for the quick fix, there is just a little mistake that I have mentioned as a line note. Also, I'm sure you see this already, but there is a code style failure (trailing whitespace).

asturur:
can you show me how you get this.canvas undefined?

not getting it.

inssein:
@asturur just emailed you back, but if anyone else is wondering, its because objects inside a group don't have access to this.canvas, but they can get it through this.group.canvas. Maybe this is a separate issue.

asturur:
No is just demostration that this fix is not 100% safe.
Reverting this AND #2083 till i have proper fix is better, but the problem happens just on .preserveObjectStacking = true, so maybe the fix will come before the issues.

asturur:
as found out from @sapics, i post the problem of this fix.
Or i found proper fix or better revert this and #2083.
i m on it.


ws000012

vizo:
Hey @asturur ,
i have the same problem like @inssein. Is this fixed yet?

vizo:
Ok i see in master is already working... thanks anyway ...;)

PR #2131, opened 2015-04-20
Added transform Matrix to group class for consistency

### LINKED ISSUE #1182: transformMatrix is not supported on Group
Trying to use the transformMatrix property on a Group does not have any effect.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1420103-transformmatrix-is-not-supported-on-group?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Also moved clip function after the transform, like in all the render method of other classes.

wait for screen for comparing.

closes #1182


kangax:
Can we have a test for this?

asturur:
How can we test this?
Is just a graphic rendering, nothing that i can check by code.
I can just check if pixel are transparent in some expected area with and without transformMatrix.
If we really need it i can try.

kangax:
Yeah, checking pixels is the best solution but it's also complex... ugh. We do need visual tests. I'll try to create a page soon! Alternatively, we can add 2 objects here so that we can at least test it manually — http://fabricjs.com/matrix-transformation/

PR #2136, opened 2015-04-21
Fix unit test for shadow

sapics:
Fix unit test for shadow by #2099.


asturur:
But why this test was not failing?

sapics:
I think so, too, why?
I am afraid that tests of qunit do not work in Travis CI.

sapics:
I have another screenshot which you might have interest.
About iText, I guess that you are fixing today?

ws000013

PR #2138, opened 2015-04-21
More robust fix for the "this.canvas" undefined situation

asturur:
Bug introduced with #2083 and partially fixed with #2120.

to be discussed, do not merge.

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -743,13 +743,23 @@
     },
 
     /**
+     * Get the canvas reference of the object
+     */
+    getCanvas: function() {
+      return this.group ? this.group.getCanvas() : this.canvas;
+    },
+
+    /**
      * Transforms context when rendering an object
      * @param {CanvasRenderingContext2D} ctx Context
      * @param {Boolean} fromLeft When true, context is transformed to object's top/left corner. This is used when rendering text on Node
      */
     transform: function(ctx, fromLeft) {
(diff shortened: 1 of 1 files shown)


sapics:
@asturur Thank you for quick fix!

Sorry, I know little about `transform`.


If the purpose is just avoiding undefined, getCanvas function would be better to

    return (this.group && this.group.getCanvas()) || this.canvas


I am afraid that, if we need to know **ctx** is come from **which canvas**, we need more complex function like below (sorry, i do not know which canvas element would be transformed, this is just a example).
    
    getCanvas(ctx.canvas);
    
    getCanvas: function(canvasEl) {
      if (this.canvas) {
        if((this.canvas.upperCanvasEl && this.canvas.upperCanvasEl === canvasEl)
           || (this.canvas.lowerCanvasEl && this.canvas.lowerCanvasEl === canvasEl)
           || (this.canvas.cacheCanvasEl && this.canvas.cacheCanvasEl === canvasEl)){
          return this.canvas
        }
      }
      return this.group && this.group.getCanvas(ctx)
      // or return (this.group && this.group.getCanvas()) || this.canvas
    }

asturur:
i do not need to know the canvas element.
I need just the canvas object to know the preserveObjectStacking parameter
I ended up doing another thing, simpler, that is ensure that every object on canvas HAS  is .canvas property set.

PR #2141, opened 2015-04-22
Make .canvas property always present

asturur:
i do not know if adding canvas to delegate properties, because this involves also reading.
anyway setting the canvas with ._set('canvas', value) automatically on object adding, both on canvas and on groups classes, should allow us to have canvas property spread to all objects inside ( in group and group of groups )

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -122,6 +122,7 @@
       if (object) {
         this._objects.push(object);
         object.group = this;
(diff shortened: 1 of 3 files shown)


asturur:
adding a test for canvas propagation in group tests.

asturur:
@kangax this should be ready.

kangax:
Perfect, thanks! Gotta add to change log.

PR #2144, opened 2015-04-23
fix group.toObject for _objects hasControls

sapics:
`hasControls` in _objects is always false by `_updateObjectCoords`.
To restore original `hasControls` at `_restoreObjectState`,
we need to save `__origHasControls` property.

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -211,8 +211,16 @@
      * @return {Object} object representation of an instance
      */
     toObject: function(propertiesToInclude) {
+      // `hasControls` in _objects is always false by `_updateObjectCoords`.
+      // To restore original `hasControls` at `_restoreObjectState`,
+      // we need to save `__origHasControls` property.
+      var objectsPropertiesToInclude = (propertiesToInclude || []).slice();
+      if (objectsPropertiesToInclude.indexOf('__origHasControls') === -1) {
+        objectsPropertiesToInclude.push('__origHasControls');
+      }
+
       return extend(this.callSuper('toObject', propertiesToInclude), {
-        objects: invoke(this._objects, 'toObject', propertiesToInclude)
+        objects: invoke(this._objects, 'toObject', objectsPropertiesToInclude)
       });
     },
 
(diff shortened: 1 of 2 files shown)


asturur:
Good find.
Thank you for finding all this issues.

With the aim of keeping the code compact, but readable, i was thinking if 

``` objects: invoke(this._objects, 'toObject', propertiesToInclude.push('__origHasControls')) ```

would be too lazy or not, whitout doing all the other things.
(cause  __origHasControls could be already there by user choice).

@kangax what do you think? the fix is good anyway, no?

And again, @sapics so many thanks for all this effort!!

**Comment by sapics** [comment]
Thank you for your compliment!

Your suggestion is much better readable than me. `propertiesToInclude` is optional and `Array.prototype.push` @return length of array, it looks like,
```
objects: invoke(this._objects, 'toObject', 
                 ((propertiesToInclude = propertiesToInclude || []).push('__origHasControls')
                  , propertiesToInclude))
```

The reason why I was used `indexOf` is just thinking about the situation like group->group->group->..., then, I wished to make one call to object property `__origHasControls`, it needs uniqueness of propertiesToInclude.

kangax:
Does this fix any particular issue or just a "good thing to do"?

asturur:
There is no issue now. It came out now that sapics start to dig in the toObject fromObject for the group class.
But effectively if you save toObject and you do not have __origHasControls set, you don't know what was the situation before saving.
Very minor issue of course.

sapics:
@kangax As you notice, I understand, it looks just a "good thing to do".

By reading object.class.js more carefully, I learn that `toObject` dose not copy interactivity properties like `hasControls` or `lockMovementX` and so on.
I understand that `toObject` and `fromObject` is designed to use always default prototype value for interactivity.

When `toObject` copy the interactivity properties in future or user add `hasControls` to propertiesToInclude, this PR would be important.
I guess that it looks very minor issue now, it might be better to close this PR.
I add unit test, which reproduce a bug. (At making unit test, I understand, it looks too minor situation (^_^:) )

P.S.
What I was afraid, `hasControls` is copied by `toObject`. At that time, hasControls of grouped object is always false. Therefore, when `group.destroy` after `fromObject`, the objects' hasControls are always false because `__origHasControls` is false or undefined(at #2101, I would fix).

PR #2150, opened 2015-04-24
Change the firing order of mouse:out/mouse:over events.

Incuuu:
Issue https://jsfiddle.net/mcnq7z2d/3/


kangax:
What's wrong with current order (over &rarr; out)?

PR #2151, opened 2015-04-27
[#93301636] Take the lack of stroke scaling into account in geometry calculations

KJTsanaktsidis:
These functions are returning the wrong values because they assume that the
stroke is affected by the scale.


KJTsanaktsidis:
Wrong darn banch

KJTsanaktsidis:
Sorry guys, this was mean to be an internal pull request on our own fork- wasn't supposed to end up here!

PR #2153, opened 2015-04-27
I text glitches fix

asturur:
Solves bad file shifting when deleting multiple characters across newlines. (old bug)
fixes  style disappearing if you add empty lines and then delete them (regression)
Solves the deleting of first word of itext. (old bug)

~~I noticed that in current version the double click and triple click are not working anymore on chrome browser. Firefox is ok.~~ it works

~~Current bug: style leaks on empty chars:
Go on kitchensink, delete the line between  "dolor" and "sit Amet"m then press spacebar.
The style of "Amet" leaks on the space between "sit" and "Amet"~~

Some other fixes possible, do not merge, leave here for inspection.


asturur:
to test: 
http://fabricjs.com/kitchensink/?load_fabric_from=http://www.deltalink.it/andreab/fabric/fabric.js

sapics:
@asturur Thank you for many fixes!

I have found other bug.
I push key [Delete] between [o] and [l], then, [o] becomes green from black.
http://giphy.com/gifs/3oEdv4LKyddQWRXisU

asturur:
ahahah so ashamed!
thanks i ll fix it.

asturur:
removed the automatic advance of cursor on newlines.
For me it does not feel natural ( also because we use it on one way and not going back ).

We can still put it back.

sapics:
I confirmed fixed! Thanks!
I would like to confirm about keydown behavior.

1. When cursor is between [ipsum]'s [u] and [m].
Press [keydown].
Then the cursor is between [dolor]'s [o] and [r].
For me, the cursor will be end of [dolor] looks natural.
http://giphy.com/gifs/l41lQovMQh7TQszE4

2. When the cursor is at top of [iorem] or between [iorem]'s [i] and [o].
Press [keydown].
Then the cursor is end of line.
For me, the cursor will be next line top looks natural.
http://giphy.com/gifs/l41lQuvlcYBv5C3io

asturur:
Fixed cursor movement (the error came because before the function was working on the text , now is working on textlines that do not include newlines and so we had a character less)
@kangax I think we are good to merge now.
@sapics thanks for keeping an eye next to mine

PR #2155, opened 2015-04-28
Remove old strokeWidth calculation

### LINKED ISSUE #1929: Wrong calculation of bounding rect
Hi, I was using function `getBoundingRect` on the object to detect its width and height but it wasn't real value. Having image of size 200x200 pixels, from `getBoudingRect()` I get width 200.5 and the same with height. After investigation I got to conclusion that it's because of `setCoords` function takes `this.strokeWidth` into calculations even though `this.stroke` property is undefined, so there is no stroke rendered. I had to set `strokeWidth` to 0 and now calculations are made correctly. Maybe it's not big deal but I would propose two solutions. The default value for `strokeWidth` property should be `0` or `setCoords` function should count `strokeWidth` only when stroke is visibile:

```js
setCoords: function() {
  var strokeWidth = typeof this.stroke === 'string' ? this.strokeWidth : 0,
  /* ... */
}
```

The ideal solution would be to do both.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7701514-wrong-calculation-of-bounding-rect?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Need to fix all the tests...

closes #1929

### DIFF
--- a/src/mixins/object_origin.mixin.js
+++ b/src/mixins/object_origin.mixin.js
@@ -13,21 +13,20 @@
      */
     translateToCenterPoint: function(point, originX, originY) {
       var cx = point.x,
-          cy = point.y,
-          strokeWidth = this.stroke ? this.strokeWidth : 0;
+          cy = point.y;
 
       if (originX === 'left') {
-        cx = point.x + (this.getWidth() + strokeWidth * this.scaleX) / 2;
+        cx = point.x + (this.getWidth() + this.strokeWidth * this.scaleX) / 2;
       }
       else if (originX === 'right') {
-        cx = point.x - (this.getWidth() + strokeWidth * this.scaleX) / 2;
+        cx = point.x - (this.getWidth() + this.strokeWidth * this.scaleX) / 2;
       }
 
       if (originY === 'top') {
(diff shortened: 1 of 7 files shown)


asturur:
@kangax, after a bit of thinking i realized that strokeWidth = 1 and stroke = none is the good default. Is the same also in SVGs, so if we do not respect that all the svg import will fail or whe should change all the "fromElement" functions to add strokeWidth = 1 if not differently specified.

The only place where i switched to strokeWidth = 0 is the group class. In this case stroke is not rendered and is useless to have it to 1 making the group width and height calculation unexpected ( one object with width 101 and group should have bounding rect  101 wide, not 101 + 1).

I fixed the test with less changes possible, if you agree i do not have anything more to add you can pull.

kangax:
@asturur please add to changelog

PR #2160, opened 2015-04-28
Don't import elements declared within `<metadata>`

paulkaplan:
Some SVG files declare elements inside of the metadata block. Don’t import them.


kangax:
Thanks!

PR #2163, opened 2015-04-30
node.js segmentation fault when image couldn't be loaded

### LINKED ISSUE #1355: Segmentation Fault when accessing image from site with htpasswd - using Node.js
Lost a bit of time debugging this. An incorrect URL for an image when using loadFromJSON with Node JS will throw an error you can trace, an url on a site using htpasswd gives Segmentation Fault D:

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2370895-segmentation-fault-when-accessing-image-from-site-with-htpasswd-using-node-js?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kibao:
Don't create image Buffer based on `data` when `data` is empty.
It happens when response status code of accessed image is other than 200.

Should also fix #1355

Example of crashing code:
```
var fabric = require('fabric').fabric;

fabric.util.loadImage('http://www.google.com/non-existing', function() {
  console.log('Loaded');
});
```

### DIFF
--- a/src/node.js
+++ b/src/node.js
@@ -71,10 +71,12 @@
 
   fabric.util.loadImage = function(url, callback, context) {
     function createImageAndCallBack(data) {
-      img.src = new Buffer(data, 'binary');
-      // preserving original url, which seems to be lost in node-canvas
-      img._src = url;
-      callback && callback.call(context, img);
+      if (data) {
+        img.src = new Buffer(data, 'binary');
+        // preserving original url, which seems to be lost in node-canvas
+        img._src = url;
+        callback && callback.call(context, img);
+      }
     }
     var img = new Image();
     if (url && (url instanceof Buffer || url.indexOf('data') === 0)) {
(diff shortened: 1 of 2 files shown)


kangax:
Thanks! I think we still need to invoke callback if there's no data (just not passing anything).

kibao:
Sure, I also though about it, just wanted to do it in another PR. But I will add it here as you wish.

@kangax Should I also update [``requestFs``](https://github.com/kangax/fabric.js/blob/39e5f1210a6eb40bcd3d69e40ce2718341dc2af6/src/node.js#L64) function, so it won't throw an error, but will invoke callback ?

kangax:
@kibao Hm, no I don't think so. Let's leave that one as is.

kangax:
@kibao do you mind also adding a unit test (the example you made in the description would be good)

PR #2164, opened 2015-05-02
Fix image resize filter from object

asturur:
Note on changes on image initialize:

We are calling initElement, that calls setElement, that is calling initConfig and applyFilters.
So the part after this is just duplicated.

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -88,18 +88,10 @@
      */
     initialize: function(element, options) {
       options || (options = { });
-
       this.filters = [ ];
       this.resizeFilters = [ ];
       this.callSuper('initialize', options);
-
       this._initElement(element, options);
-      this._initConfig(options);
-
-      if (options.filters) {
-        this.filters = options.filters;
-        this.applyFilters();
-      }
     },
 
     /**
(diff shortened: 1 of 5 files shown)


asturur:
couple of test needed

src/shapes/image.class.js
```diff
@@ -88,18 +88,10 @@
...
       this.filters = [ ];
       this.resizeFilters = [ ];
       this.callSuper('initialize', options);
-
       this._initElement(element, options);
-      this._initConfig(options);
-
-      if (options.filters) {
-        this.filters = options.filters;
-        this.applyFilters();
```
kangax: How would the filters be applied now during image initialization?

kangax: Oh, I see your explanation. Sorry, got it!

kangax:
After tests, I'm ready to merge then!

asturur:
@kangax any help on why the test is failing? it gives no clue. The row is:
```
ok(filterFromObj instanceof fabric.Image.filters.Resize);
```

asturur:
I commented the part of the test that i added and does not run under node.

asturur:
@kangax it is finally done.

kangax:
+Changelog

PR #2166, opened 2015-05-02
Remove build.js from .npmignore.

m14t:
I would like a way to have a reproducible custom version of Fabric.js in my project.  Your git tags (thanks for moving to SemVer!) and `build.js` file go a long way to making this possible, but the `build.js` file being in the `.npmignore` list excludes it from npm installations which prevents me from being able to do this during my build process.

If there's no conflict, it would be great if we could include the `build.js` file in npm installations.

Thanks!

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,4 +6,3 @@ dist/fabric.min.js
 dist/fabric.min.js.gz
 .DS_Store
 HEADER.js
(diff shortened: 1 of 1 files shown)


kangax:
Hm, I guess I'm ok with that if it'll help you and others!

PR #2168, opened 2015-05-03
Only renderAll() while moving objects if renderOnAddRemove is true

AdamMerrifield:
Currently all methods that move objects to different indexes check the
existence of the renderAll method which i believe is supposed to be
checking if renderOnAddRemove is true

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1368,7 +1368,8 @@
     sendToBack: function (object) {
       removeFromArray(this._objects, object);
       this._objects.unshift(object);
-      return this.renderAll && this.renderAll();
+      this.renderOnAddRemove && this.renderAll();
+      return this;
     },
 
     /**
@@ -1380,7 +1381,8 @@
     bringToFront: function (object) {
       removeFromArray(this._objects, object);
       this._objects.push(object);
-      return this.renderAll && this.renderAll();
+      this.renderOnAddRemove && this.renderAll();
+      return this;
     },
 
     /**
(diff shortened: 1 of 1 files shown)


asturur:
I'm not  sure renderOnAddRemove has this description:

```
     * Indicates whether {@link fabric.Collection.add}, {@link fabric.Collection.insertAt} and {@link fabric.Collection.remove} should also re-render canvas.
     * Disabling this option could give a great performance boost when adding/removing a lot of objects to/from canvas at once
     * (followed by a manual rendering after addition/deletion)
     * @type Boolean
     * @default
```

It refer to add, insertAt and remove.
Is not related to moving to another index.

AdamMerrifield:
You may be right. I just assumed it was an error because why check for `this.renderAll` before invoking it? When would `renderAll` not be present? It's not as if renderAll is a module that can be removed. It just didn't make sense to me.

asturur:
Nor to me.
Maybe, instead of not present could be set equal to false on purpose:
```
originalRenderAll = this.renderAll();
this.renderAll = false
... later...
this.renderAll = originalRenderAll.
```
i'm guessing, because i do not the original intention of that code.

AdamMerrifield:
I see what you're saying. I did a quick grep on all the files for `renderAll =` & `renderAll=` and didn't come up with any results. It could just be legacy though. Either way it should be changed. 

Maybe @kangax will have some better insight as to if it should use `RenderOnAddRemove` or just add an optional parameter to each of these.

kangax:
Hm, don't know 100% off hand but I think it might be related to static canvas which might not have renderAll. But again, need to check closer (I'm not by computer now)

asturur:
i do not think so, renderAll is a method of the staticCanvas.

kangax:
Then we probably don't need it

PR #2173, opened 2015-05-05
Avoid restoring properties manually that are handled by canvas save stack

### LINKED ISSUE #2103: too much canvas restoring in object render
I was checking object.render:

we do a startup ctx.save, and an ending ctx.restore.

Before ctx.restore we
```
   _removeShadow: function(ctx) {
      if (!this.shadow) {
        return;
      }

      ctx.shadowColor = '';
      ctx.shadowBlur = ctx.shadowOffsetX = ctx.shadowOffsetY = 0;
    },
```
and
```
    _restoreCompositeOperation: function (ctx) {
      if (this.globalCompositeOperation && this._prevGlobalCompositeOperation) {
        ctx.globalCompositeOperation = this._prevGlobalCompositeOperation;
      }
    }
```

But those property are reset by the normal ctx.restore that is one line below.
So why we execute them?

Also the clipContext method automatic save and restore could be avoided by saving just when is necessary, by looking at the code, we call it mainly in object rendering where save and restore specific for clipContext is not necessary at all, because is always done before and after clipping another save and restore with no drawing in the middle.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10920054-too-much-canvas-restoring-in-object-render?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
to be tested with some visual test.

closes #2103

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -979,8 +979,6 @@
       this.clipTo && fabric.util.clipContext(this, ctx);
       this._render(ctx, noTransform);
       this.clipTo && ctx.restore();
-      this._removeShadow(ctx);
-      this._restoreCompositeOperation(ctx);
 
       ctx.restore();
     },
@@ -1102,9 +1100,6 @@
         ctx.fill();
       }
       ctx.restore();
-      if (this.shadow && !this.shadow.affectStroke) {
-        this._removeShadow(ctx);
-      }
     },
 
     /**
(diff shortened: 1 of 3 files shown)


asturur:
@kangax to me looks working

image

One property frome object ( shadow or compositing ) does not leak to the next one in stack.

For the cliptTo function i will do a separate PR

PR #2176, opened 2015-05-06
IText 'getSelectionStyles' Bug Fix

ArtisanCodesmith:
This fixes a bug where if you select a portion of text within an IText object and select another element without clearing the text selection it still returns the old selection's styles when getSelectionStyles is called.

For instance, if you have a bold button which tells whether or not the currently selected text is bold, if you highlight text, make it bold, click off the object and click back on it - it will still say the selected text is bold when there is no selected text.

If you have trouble replicating the bug it give me a shout,

Thanks!! :)

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -20639,6 +20639,10 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
      */
     getSelectionStyles: function(startIndex, endIndex) {
 
+      if (! this.getSelectedText().length) {
+        return { };
+      }
+
       if (arguments.length === 2) {
         var styles = [ ];
         for (var i = startIndex; i < endIndex; i++) {


asturur:
The function is supposed to work also without any selection.
The only fix could be avoid returning a style if no argument is passed and if  we are not in edit mode.
How are you using this function?
Anyway to submit a pr you have to modify the partial files, not fabric.js file.
Just itext.class.js in your case.

asturur:
Also documentation of this function is a little bit misleading i see.

ArtisanCodesmith:
Ahh I see, I'll make a new pull request.

The problem is:

Say you have a bold button which is only green if the selected text is bold.

If you select text and then call this function - the button will turn green as expected.

If you then highlight other text that's not bold - the button will no longer be green as expected.

But if you highlight text and it turns green - then you deselect the element entirely and then reselect the element (without highlighting text) it will return the style you highlighted earlier when not even in edit mode if that makes sense?

If I realized there was an edit mode boolean I would have used that instead of 'this.getSelectedText().length' :)

asturur:
function has two parameters and behaves differently for two of them.

if you send 0 parameters and select 5 chars you are getting anyway an invalid result.


can i see how you call it in your code?

ArtisanCodesmith:
The function (with my fix) works really well, but otherwise returns a late selection styles object where it doesn't exist, here is the function:

    /**
     * Updates the font style buttons.
     *
     */
    function updateFontStyleButtons()
    {
        // Get the active Object
        var activeObject = canvas.getActiveObject();

        // There is no active object
        if (! activeObject || activeObject.type != 'textbox') {
            jQ('.fa-bold').parent().removeClass('btn-success').addClass('btn-light');
            jQ('.fa-italic').parent().removeClass('btn-success').addClass('btn-light');
            jQ('.fa-underline').parent().removeClass('btn-success').addClass('btn-light');

            return;
        }

        // Get the selected text style
        var selectionStyles = activeObject.getSelectionStyles();

        // Turn on the bold button
        if (selectionStyles.fontWeight && selectionStyles.fontWeight.trim() == 'bold' ||
            activeObject.fontWeight && activeObject.fontWeight.indexOf('bold') > -1) {
            jQ('.fa-bold').parent().removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the bold button
        else {
            jQ('.fa-bold').parent().removeClass('btn-success').addClass('btn-light');
        }

        // Turn on the italic button
        if (selectionStyles.fontStyle && selectionStyles.fontStyle.trim() == 'italic' ||
            activeObject.fontStyle && activeObject.fontStyle.indexOf('italic') > -1) {
            jQ('.fa-italic').parent().removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the italic button
        else {
            jQ('.fa-italic').parent().removeClass('btn-success').addClass('btn-light');
        }

        // Turn on the underline button
        if (selectionStyles.textDecoration && selectionStyles.textDecoration.trim() == 'underline' ||
            activeObject.textDecoration && activeObject.textDecoration.indexOf('underline') > -1) {
            jQ('.fa-underline').parent().removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the underline button
        else {
            jQ('.fa-underline').parent().removeClass('btn-success').addClass('btn-light');
        }

        // Turn on the align left button
        if (activeObject.textAlign && activeObject.textAlign.indexOf('left') > -1) {
            jQ('.btn-align-left').removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the align left button
        else {
            jQ('.btn-align-left').removeClass('btn-success').addClass('btn-light');
        }

        // Turn on the align right button
        if (activeObject.textAlign && activeObject.textAlign.indexOf('right') > -1) {
            jQ('.btn-align-right').removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the align right button
        else {
            jQ('.btn-align-right').removeClass('btn-success').addClass('btn-light');
        }

        // Turn on the align center button
        if (activeObject.textAlign && activeObject.textAlign.indexOf('center') > -1) {
            jQ('.btn-align-center').removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the align center button
        else {
            jQ('.btn-align-center').removeClass('btn-success').addClass('btn-light');
        }

        // Turn on the align justify button
        if (activeObject.textAlign && activeObject.textAlign.indexOf('justify') > -1) {
            jQ('.btn-align-justify').removeClass('btn-light').addClass('btn-success');
        } 

        // Turn off the align justify button
        else {
            jQ('.btn-align-justify').removeClass('btn-success').addClass('btn-light');
        }
    }

The problem is when I do:

    // Get the selected text style
    var selectionStyles = activeObject.getSelectionStyles();

After I've deselected the textbox entirely and then reselect it - it gives the the old selection styles even though they don't exist anymore and nothing is selected, which is why I thought it wise in the function to check whether anything is actually selected before we give them non-existent styles for non-existent text.

I'm starting to think it's a genuine bug but from your reaction the wrong place to fix it! :)

asturur:
the only problem is that the function is meant to give back a selection style, and selection can be also more than one character.

so without any selection getSelectionStyle(1,3) has to return the style

without any selection getSelectionStyle(3) has to return the style

Without selection in editmode, getSelectionStyle() return style of current cursor position and is ok.

with selection of multiple chars in edit mode, getSelectionStyle() will return the style of first character selected.

in your case the problem is that you rely on a ghost value of "selectionStart".maybe the bug is that and if any parameter are passed are we are not editing we should return empty OR return the style of first  character.... i wanted to point out that is not an immediate solution.

ArtisanCodesmith:
I just checked now and realized there is a boolean at 'fabric.IText.prototype.isEditing' which would solve the problem entirely as I can just use that to check whether a style needs to be retrieved or not,

Thanks for your time! :)

asturur:
Maybe the solution could be also here:

https://github.com/kangax/fabric.js/blob/master/src/mixins/itext_behavior.mixin.js#L477

```diff
- this.selectionEnd = this.selectionStart;
+ this.selectionEnd = this.selectionStart = 0;
```

And keeping the function as it is.

ArtisanCodesmith:
I removed my tweak to the function and tried your proposed solution and it worked perfectly so I'm glad I raised this now.

Works like a charm! :)

asturur:
:+1 !

ArtisanCodesmith:
Hoorah!! :tada: :)

PR #2178, opened 2015-05-08
Fix stroke width calculation

### LINKED ISSUE #2177: mouse scaling bug
A scaling by mouse might have a bug.

The scaled size of Line is twice larger than a expected size. The strokeWidth of object might cause this size difference.

http://jsfiddle.net/sapics/tt5zjjot/1/

ezgif com-gif-maker 1

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13389170-mouse-scaling-bug?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sapics:
From #1929, strokeWidth calculation dose not depends on stroke.

closes #2177

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -499,10 +499,13 @@
      */
     _setObjectScale: function(localMouse, transform, lockScalingX, lockScalingY, by, lockScalingFlip) {
       var target = transform.target, forbidScalingX = false, forbidScalingY = false,
-          strokeWidth = target.stroke ? target.strokeWidth : 0;
+          vLine = target.type === 'line' && target.width === 0,
+          hLine = target.type === 'line' && target.height === 0,
+          strokeWidthX = hLine ? 0 : target.strokeWidth,
+          strokeWidthY = vLine ? 0 : target.strokeWidth;
 
-      transform.newScaleX = localMouse.x / (target.width + strokeWidth / 2);
-      transform.newScaleY = localMouse.y / (target.height + strokeWidth / 2);
+      transform.newScaleX = localMouse.x / (target.width + strokeWidthX);
+      transform.newScaleY = localMouse.y / (target.height + strokeWidthY);
 
       if (lockScalingFlip && transform.newScaleX <= 0 && transform.newScaleX < target.scaleX) {
         forbidScalingX = true;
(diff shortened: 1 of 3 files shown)


asturur:
I definitely missed this.
Inspecting for #2177 i noticed that, and i was about to fix this.

sapics:
@asturur i found it from #2177, too!

asturur:
Sapics if you remove the half part of strokewidth, and you scale will FULL strokewidth in both _setObjectScale  and _scaleObjectEqually, you will notice that #2177 is fixed.
For the third function you modified check if apply.

Obviously if we are doing something with width, we cannot just use half of stroke.
For vLine and hLine still some tweak is necessary.

Could you make that? if i do this we will have conflicting PR

sapics:
Absolutely!
I have not understood about definition of localMouse yet, thus, I cannot judge HALF or FULL strokeWidth is right.
It would be better to close this PR to avoid conflict.

asturur:
I think this PR is good, let's just finish it.
Exactly where you changed the strokeWidth definition, if you remove the use of strokeWidth / 2 and use strokeWidth, the object scaling is fixed 99% of time.

localMouse  depends on the translate ( s) function, there are at least 3, and the fix is ugly as shit for them.
the fix would be:
```
          vLine = target.type === 'line' && target.width === 0,
          hLine = target.type === 'line' && target.height === 0,
          strokeWidthX = hLine ? 0 : target.strokeWidth,
          strokeWidthY = vLine ? 0 : target.strokeWidth;
```
and use strokeWidthX and strokeWidthY for the x and y instead of general strokeWidth.

If you try to resize this in current kitchensink on fabricJS.com:
```
canvas.add(new fabric.Rect({
 width:50, height:50,
      fill: 'red',
      stroke: 'blue',
      strokeWidth: 20
}))
```

You will notice that is wrong.
So the scaling function we have is just wrong, no one notice because large strokeWidth is not so popular.

asturur:
@sapics thanks!

Can you also rebase the commit to have them nice?

sapics:
@asturur I am sorry, I misunderstood you wrote.
I create jsfiddle to check this fix.
PREVIOUS: http://jsfiddle.net/sapics/tt5zjjot/8/
THIS PR: http://jsfiddle.net/sapics/tt5zjjot/7/

sapics:
Sure!!

asturur:
To fix the "translate effect" that still remain for just horizontal or vertical lines i would wait.

We should spam this code 
```
          vLine = target.type === 'line' && target.width === 0,
          hLine = target.type === 'line' && target.height === 0,
          strokeWidthX = hLine ? 0 : target.strokeWidth,
          strokeWidthY = vLine ? 0 : target.strokeWidth;
```
All around the functions, while i want to do something more clean.
@kangax i think you can merge this after a quick look.

sapics:
Is it better or not?
```javascript
vLine = target.type === 'line' && target.width === 0,
hLine = target.type === 'line' && target.height === 0,
sLine = vLine || hLine,
capped = target.strokeLineCap === 'round' || target.strokeLineCap === 'square',
strokeWidthX = ((capped && hLine) || !sLine) ? target.strokeWidth : 0,
strokeWidthY = ((capped && vLine) || !sLine) ? target.strokeWidth : 0;
```

sapics:
@asturur Thank you! I have fixed as you coded and rebase!

kangax:
Awesome, thanks. Don't forget changelog please.

PR #2181, opened 2015-05-09
fix for null shadow

sapics:
This PR fix error `Uncaught TypeError: Cannot read property 'affectStroke' of null`

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1111,7 +1111,7 @@
         return;
       }
 
-      if (!this.shadow.affectStroke) {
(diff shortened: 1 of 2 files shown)


asturur:
ok this is my fresh error of yesterday.
so dumb.

i m sorry i did not use stroke on my tests. will be more carefull.

asturur:
i removed it because inside the remove shadow function we do the same check again.

asturur:
@sapics sorry i have to submit 2 PRs both conflicting with this.
I copy your fix in my PR for text and for object class, otherwise is complicated to merge

sapics:
@asturur I see! Is it better to close?

PR #2182, opened 2015-05-09
Fix restore stroke gradient from object

asturur:
The initGradient method was missing the stroke object

includes @sapics fix #2181 

closes #2181


kangax:
+changelog

PR #2184, opened 2015-05-09
Improve Object selection - Fix Line shape toObject

farnabaz:
Let canvas select a selectable objects that exists under a none-selectable Object by only processing selectable objects

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -790,6 +790,7 @@
      */
     _checkTarget: function(e, obj, pointer) {
       if (obj &&
+          obj.selectable &&
           obj.visible &&
           obj.evented &&
           this.containsPoint(e, obj)){
--- a/src/shapes/line.class.js
+++ b/src/shapes/line.class.js
@@ -202,7 +202,12 @@
      * @return {Object} object representation of an instance
      */
     toObject: function(propertiesToInclude) {
-      return extend(this.callSuper('toObject', propertiesToInclude), this.calcLinePoints());
+      return extend(this.callSuper('toObject', propertiesToInclude), {
+        x1: this.x1,
+        x2: this.x2,
+        y1: this.y1,
+        y2: this.y2
+      });
     },
 
     /**


sapics:
@farnabaz Is there a example (like jsfiddle) which causes problems by calling `calcLinePoints` in `toObject`?

farnabaz:
@sapics here is example for first commit http://jsfiddle.net/o48qeepu/
and for second commit, i'll try to create a proper example, but to explain the problem: this issue occurs when I'm trying to create socket base canvas app using fabric, and when I send `toJSON()` value to another user over socket, created Line on receiver canvas is in different position from sender

sapics:
@farnabaz Thank you for jsfiddle and explanation!
For the second commit, it reverts a part of #1714. It would be necessary to demonstrate a bug.

asturur:
i understand the bug, i will put up a test to demonstrate it.

asturur:
Regarding to line object:
http://jsfiddle.net/xvzms0j6/34/
Please try to make a jsfiddle with bug, from this jsfiddle it looks like is working good.

sapics:
@asturur Thank you for jsfiddle!
I also confirmed that working well, with moving, scaling, rotation.

farnabaz:
I try to create jsfiddle to show bug, but in all attempts I fail
here is my work (fabric.js on socket.io), the idea is to observe `object:added` `object:modified` .. and send these events over socket and create/update object in client side, this has server & client, and only server could draw 

server: http://nodejs.moror.ir/server.html
client: http://nodejs.moror.ir/

- open server & client in two window
- press `L` and drag to draw line

as you see in client side line position is different from server side, but when iill change `Line.toObject` method lines are in same position

asturur:
I see.
I'm sorry but we still need a simple test case to replicate the bug and accept the PR or propose another solution.
We need to replicate it in a simple js script or in a nodejs setup where we can see the code.

http://jsfiddle.net/xvzms0j6/34/

Can you not make it happen in that jsfiddle?

farnabaz:
look at this jsfiddle: http://jsfiddle.net/xvzms0j6/36/

farnabaz:
@asturur Thank you, It works so well
but I think this is a little confusing, because `set` function works well for other shapes

asturur:
I think is a coincidence.
toObject and fromObject are created for other reasons.
But as you noticed, what you where doing  is just passing the value of properties as they are, without any modification ( so instaed of calcLinePoints, just the four points ).

So you can create your own method "toSimpleObject", to do just what you need. 
So passing each property or a subset of them, to an object, that later you use with "set" method.

PR #2186, opened 2015-05-11
Add alternative to get elementById for ie9

### LINKED ISSUE #1518: Functions not working in ie9
I tried those functions in ie9 ( well ie10 set as ie9 emulation ).

**addSvgTransform**
```
  /**
   * Add a <g> element that envelop all SCG elements and makes the viewbox transformMatrix descend on all elements
   */
  function addSvgTransform(doc, matrix) {
    matrix[3] = matrix[0] = (matrix[0] > matrix[3] ? matrix[3] : matrix[0]);
    if (!(matrix[0] !== 1 || matrix[3] !== 1 || matrix[4] !== 0 || matrix[5] !== 0)) return;
    // default is to preserve aspect ratio
    // preserveAspectRatio attribute to be implemented
    var el = document.createElement('g');
    while (doc.firstChild != null) {
      var node = doc.firstChild;
      el.appendChild(node);
    }
    el.setAttribute('transform','matrix(' + matrix[0] + ' ' + matrix[1] + ' ' + matrix[2] + ' ' + matrix[3] + ' ' + matrix[4] + ' ' + matrix[5] + ')');
    doc.appendChild(el);
  }
```
Failing point:
el is a document.element.
ie9 doesn't wan't to append to this kind of node any other node from our SVG and it doesn't want to do the contrary , appending this node to the Svg doc directly.
our doc doesn't have a createElement method.
Purpouse of this function is to add a 'G' element between root of the doc and all other elements

**parseUseDirectives**
```
  function parseUseDirectives(doc) {
    var nodelist = doc.getElementsByTagName('use');
    while (nodelist.length) {
      var el = nodelist[0],
          xlink = el.getAttribute('xlink:href').substr(1),
          x = el.getAttribute('x') || 0,
          y = el.getAttribute('y') || 0,
          el2 = doc.getElementById(xlink).cloneNode(true),
          currentTrans = (el.getAttribute('transform') || '') + ' translate(' + x + ', ' + y + ')',
          parentNode;

      for (var j = 0, attrs = el.attributes, l = attrs.length; j < l; j++) {
        var attr = attrs.item(j);
        if (attr.nodeName === 'x' || attr.nodeName === 'y' || attr.nodeName === 'xlink:href') continue;

        if (attr.nodeName === 'transform') {
          currentTrans = currentTrans + ' ' + attr.nodeValue;
        }
        else {
          el2.setAttribute(attr.nodeName, attr.nodeValue);
        }
      }

      el2.setAttribute('transform', currentTrans);
      el2.removeAttribute('id');
      parentNode = el.parentNode;
      parentNode.replaceChild(el2, el);
    }
  }
```
This function is failing because  doc doesn't have a doc.getElementById method.

Those function should be fixed to have support for ie9, i wrote them, they got merged, now i noticed they do not work on ie9, i tried to fix but i'm not so cross browser expert.

Any ideas, help?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3238983-functions-not-working-in-ie9?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Not tested yet,
can someone forsee some problem from this implmentation?
At the end of parsing, should i clear that big object that i just populated?

closes #1518

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -367,19 +367,36 @@
     return selector.length === 0;
   }
 
+  function elementById(doc, id, elementsById) {
+    var el = doc.getElementById(id);
+    if (el) {
+      return el;
+    }
+    if (elementsById[id]) {
+      return elementsById[id];
+    }
+    var node, i, idAttr, nodelist = doc.getElementsByTagName('*');
+    for (i = 0; i < nodelist.length; i++) {
+      node = nodelist[i];
+      idAttr = node.getAttribute('id');
+      elementsById[idAttr] = node;
+    }
+    return elementsById[idAttr];
+  }
+
   /**
    * @private
    */
   function parseUseDirectives(doc) {
-    var nodelist = doc.getElementsByTagName('use');
-    while (nodelist.length) {
-      var el = nodelist[0],
+    var nodelist = doc.getElementsByTagName('use'), elementsById = { }, i = 0;
+    while (nodelist.length && i < nodelist.length) {
+      var el = nodelist[i],
           xlink = el.getAttribute('xlink:href').substr(1),
           x = el.getAttribute('x') || 0,
           y = el.getAttribute('y') || 0,
-          el2 = doc.getElementById(xlink).cloneNode(true),
+          el2 = elementById(doc, xlink, elementsById).cloneNode(true),
           currentTrans = (el2.getAttribute('transform') || '') + ' translate(' + x + ', ' + y + ')',
-          parentNode;
+          parentNode, oldLength = nodelist.length;
 
       for (var j = 0, attrs = el.attributes, l = attrs.length; j < l; j++) {
         var attr = attrs.item(j);
@@ -400,6 +417,10 @@
       el2.removeAttribute('id');
       parentNode = el.parentNode;
       parentNode.replaceChild(el2, el);
+      // some browsers do not shorten nodelist after replaceChild
+      if (nodelist.length === oldLength) {
+        i++;
+      }
     }
   }
 
--- a/src/parser.js
+++ b/src/parser.js
@@ -367,34 +367,36 @@
     return selector.length === 0;
   }
 
-  function elementById(doc, id, elementsById) {
-    var el = doc.getElementById(id);
+  /**
+   * @private
+   * to support IE8 missing getElementById on SVGdocument
+   */
+  function elementById(doc, id) {
+    var el;
+    doc.getElementById && (el = doc.getElementById(id));
     if (el) {
       return el;
-    }
-    if (elementsById[id]) {
-      return elementsById[id];
     }
     var node, i, idAttr, nodelist = doc.getElementsByTagName('*');
     for (i = 0; i < nodelist.length; i++) {
       node = nodelist[i];
-      idAttr = node.getAttribute('id');
-      elementsById[idAttr] = node;
+      if (idAttr === node.getAttribute('id')) {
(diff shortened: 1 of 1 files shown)


asturur:
anyone subscribed and with real internet explorer 9 ( emulated is not enough ) wants to test?

src/parser.js
```diff
@@ -367,6 +368,23 @@
     return selector.length === 0;
   }
 
+  function elementById(doc, id) {
+    var el = doc.getElementById(xlink);
```
kangax: xlink -> id, right?

asturur: lazy copy paste.

kangax:
Yeah, I don't think we should cache things like that. If an element with id was replaced by another element with the same id, we'll still be returning an old one (cached in a "hash"). We would need to invalidate this hash with mutation observers, etc. and it's just not worth the hassle, obviously. The actual finding of an element (via attribute) is good though!

asturur:
I did not mean to cache them, just retrieve where getElementById fails.
I want the object to last just for duration of the parsing process of ONE svg, and then trash it.

I just did it in the wrong place....

kangax:
Then maybe just return an element as soon as you find it, instead of creating an object?

I'm not sure how gEBI should work with repeated id's though — does it return 1st found or last? This might be against specification, but need to check.

asturur:
There should be one unique id, i don't think we are in the position we can also care for specification-violating SVGs.
What i wanted was to be sure that the object is garbage collected after svg parsing, and with this last commit should be done.

I decided to create the object because if i do not, when we have 10 ```<use>``` tags,  i do not want to cycle and do getAttribute for every of them. I pay once the load of cycling all elements and then i have all of them.

I also thought of exiting on first find, but then again i would have a partial object and there would be chances i would cycle over the same objects again for the next id.

What do you think?

asturur:
I have to to test:
```document.querySelector('#'+ id)[0]```
So strange i did not use it, maybe some difficulty i forgot.

**Comment by kangax** [comment]
I'm not sure querySelector works with SVG. As far as iteration, I wouldn't worry about performance in this case. It's a one time operation and it's only in older IE so it's a rare case, you know...

**Comment by asturur** [comment]
Ok big mistake from me.
This function is already ok with ie9.
The one failing is ie8.
So i made it ie8 safe. I do not like the code, but if it is acceptable at least we can still load SVG with use tag in ie8.

Ie8 problem other than not using ```getElemenById``` on SVG is that using ```replaceChild``` was not shortening the node list.


image

**Comment by asturur** [comment]
@kangax if the code is not too hacky for you we can pull.

**Comment by kangax** [comment]
I still don't really like having that hash. Don't think we should worry about performance in this case. So can you please just get rid of it and return element right away. Also let's leave a comment that this is for IE8 so that when we deprecate it, it's easier to clean all these places.

**Comment by asturur** [comment]
:+1:  done

**Comment by asturur** [comment]
Just a note:

I did not write
```
    if (doc.getElementById) {
      return doc.getElementById(id);
    }
```
just because some browser without ID declaration, supporting doc.getElementById, fail to recognize id as ID attribute if not explicitly stated somewhere in the document.

PR #2191, opened 2015-05-12
Fix insertNewlineStyleObject failing, when lineIndex is not present in style object

### LINKED ISSUE #2189: [IText] insertNewlineStyleObject fails if lineIndex is not existing in style object
I'm using fabric on node.js. I create IText objects on the server and can't use insertChars() since it fires renderAll() at a time I don't need it. So I add newline chars to the text directly and set insertNewlineStyleObject() manually. However, this fails if the line after which the new line style should be created has no current styles.

In browser usage this error is prevented because insertStyleObjects() checks the styles for the line and just creates an empty style object for the line before calling insertNewlineStyleObject(). But since insertNewlineStyleObject() is not a private function I think this check hast to be done in there again.

See http://jsfiddle.net/cygvj607/1/ for an example (the fiddle shows two fixes, the error can be seen in the console on clicking "broken function")

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14434470-itext-insertnewlinestyleobject-fails-if-lineindex-is-not-existing-in-style-object?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

milan-w:
currentCharStyle was assigned the value of this.styles[lineIndex][charIndex - 1]. This fails if there is no lineIndex object in this.styles. This assigns an empty object to currentCharStyle by default and only assigns the value if it's present.

Closes #2189

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -560,8 +560,8 @@
       var currentCharStyle = { },
           newLineStyles = { };
 
-      if(this.styles[lineIndex] && this.styles[lineIndex][charIndex - 1]) {
-          currentCharStyle = this.styles[lineIndex][charIndex - 1];
+      if (this.styles[lineIndex] && this.styles[lineIndex][charIndex - 1]) {
+        currentCharStyle = this.styles[lineIndex][charIndex - 1];
       }
 
       // if there's nothing after cursor,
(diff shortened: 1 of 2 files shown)


asturur:
To me it looks ok! Thanks.

PR #2203, opened 2015-05-17
Set default value of strokeDashArray so that SVG is valid

sorich87:
When `strokeDashArray` is not set on an object, it's rendered in the SVG as `stroke-dasharray: ;`, which breaks some parsers.

This PR fixes it so that the default value `none` is used.


kangax:
Thanks! Could you please fix unit test as well.

sorich87:
Done.

asturur:
There was an old pr that i never finished to export just NON default values into svgs.
While this is a simple fix to merge and fix asap if it breaks some parsers, with mind term thinking we should avoid to export all the properties that are default svg value ( black color, strokewidth = 1, opacity = 1.... and so no) 
@sorich87 wich parser where you using?

sorich87:
It was PHP xml parser, and I think librsvg too.

PR #2204, opened 2015-05-18
Add a bool to force cache clear on next render

### LINKED ISSUE #2201: i-text border selection when font weight changed for selected styles
This issue exist even in the kitchen sink demo. To replicate do the following steps

#1) Select some text [inside the i-text]
#2) Change font size large or small
#3) again select the i-text
#4) The border selection broken now.

Please see the screen shot
image

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15240517-i-text-border-selection-when-font-weight-changed-for-selected-styles?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Issue #2201 made me found out the bug i was searching for.
Clearing cache is not enough, because dimension of text where not recalculated if cache clearing was not detected by ```_shouldClearCache()``` method.

Adding a boolean that force cache clearing solves this problems, also avoids to clear cache too many time for one rendering

closes #2201

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -579,7 +579,6 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
 
     this._removeExtraneousStyles();
 
-    this._clearCache();
     this.canvas && this.canvas.renderAll();
 
     this.setCoords();
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -582,7 +582,7 @@
         }
         this.styles[lineIndex + 1] = newLineStyles;
       }
-      this._clearCache();
+      this._forceClearCache = true;
     },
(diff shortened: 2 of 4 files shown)


src/mixins/itext_key_behavior.mixin.js
```diff
@@ -579,7 +579,6 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
 
     this._removeExtraneousStyles();
 
-    this._clearCache();
```
asturur: this gets removed.
Removing char is changing text property so no need to enforce cache clearing.

sarath-prohaktiv:
Thanks asturur  
It is working now

PR #2206, opened 2015-05-18
dimensions code refactor

### LINKED ISSUE #2187: Changing line stroke width shifts it to the right
Hi,
I'm trying to make a line thicker, as per my understanding I have to change the strokeWidth of the line.
If I change the strokeWidth property thel ine actually gets bigger but also shifts to the right.
I prepared a jsfiddle, try fiddleing around with the strokeWidth value and you will se that the origin point of the line moves right together with the strokeWidth. This is the fiddle: http://jsfiddle.net/3mqgp08m/1/ , the rectangle is there just to give you a point to compare with the line starting point. 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14416186-changing-line-stroke-width-shifts-it-to-the-right?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Fixes handling of strokewidth limit cases.

I did this to move the logic of calculating how big is an object and put under a function ```_getUntransformedDimensions()``` to reuse when necessary.

Lot of functions like:
translateToCenterPoint
translateToOriginPoint
getPointByOrigin

relies on a wrong calculation of the object unscaled dimensions, assuming that
```width * scaleX + strokewidth * scaleX```
is the solution for everything.

While this is not true for
- horizontal lines with big strokewidth,
- object with strokewidth bigger than width
- and it won't work in situation where the object inside has a complex transform

For example you can try to paste this code in the current kitchensink and try to resize the resulting object:

```
// clear canvas
canvas.clear();

// add red rectangle
canvas.add(new fabric.Line([100,100,100,200],{
      fill: 'red',
      stroke: 'red',
      strokeWidth: 500,
    }));
```

You will see that the object moves while you resize it.
This pr fixes that behaviour.

And make easier handle objects where there is a transformMatrix or a skew

closes #2187

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -509,13 +509,10 @@
      */
     _setObjectScale: function(localMouse, transform, lockScalingX, lockScalingY, by, lockScalingFlip) {
       var target = transform.target, forbidScalingX = false, forbidScalingY = false,
-          vLine = target.type === 'line' && target.width === 0,
-          hLine = target.type === 'line' && target.height === 0,
-          strokeWidthX = hLine ? 0 : target.strokeWidth,
-          strokeWidthY = vLine ? 0 : target.strokeWidth;
+          dim = target._getNonTransformedDimensions();
 
-      transform.newScaleX = localMouse.x / (target.width + strokeWidthX);
-      transform.newScaleY = localMouse.y / (target.height + strokeWidthY);
+      transform.newScaleX = localMouse.x / dim.x;
+      transform.newScaleY = localMouse.y / dim.y;
 
       if (lockScalingFlip && transform.newScaleX <= 0 && transform.newScaleX < target.scaleX) {
         forbidScalingX = true;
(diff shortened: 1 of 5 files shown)


kangax:
I love this cleanup and simplification! One thing — untransformed sounds strange to my ear. I would prefer nontransformed or even better withoutTransform.

kangax:
Actually, no, getNonTransformedDimensions and getTransformedDimensions would be best, I think.

asturur:
Done.

i will make more changes in little steps.

kangax:
This is not ready yet? More changes coming?

asturur:
no no, different PRs with different cleaning ( setCoords is the next one )

kangax:
+changelog

PR #2218, opened 2015-05-21
remove console.log

sapics:
remove unnecessary console.log


asturur:
My error.

kangax:
I fixed it and made it so that `console.log` are caught in the future

PR #2237, opened 2015-05-27
Textbox Implementation - from Jaffer Haider

inssein:
I hope this is not frowned upon, but I had been waiting for @jafferhaider's textbox implementation to be merged for a long time as I was falling behind on fabricjs master. I took the liberty of taking his work, and merging it into fabricjs master.

The only difference between this and @jafferhaider's work is the textbox_key_behavior.js in which he was using "callSuper", which wasn't working for me as I had extended the textbox class in my own project.


inssein:
Forgot to mention the relevant PR's that this work is based on: https://github.com/kangax/fabric.js/pull/2225 and https://github.com/kangax/fabric.js/pull/1778.

inssein:
There are currently a few bugs with this implementation:
* get2DCursorLocation returns incorrect positions when mixing line breaks and wraps (I have a solution in place, needs more testing before I commit)
* Style does not seem to carry over when it auto wraps
* Removing characters seem to destroy the style of the current line (all the way to \n).

I don't know if the last issue is with this implementation alone, it happens on the itext demo on the fabric.js website, but that does not include @asturur's recent changes.

inssein:
Ah crap, looks like I didn't squash the commits when I rebased. My Apologies.

kangax:
Thanks @inssein, I was waiting for a clean PR to look through the changes. Will do that as soon as I can and comment!

src/shapes/textbox.class.js
```diff
@@ -0,0 +1,321 @@
...
+     * Gets lines of text to render in the Textbox. This function calculates
+     * text wrapping on the fly everytime it is called.
+     * @returns {Array} Array of lines in the Textbox.
+     * @override
+     */
+    _splitTextIntoLines: function () {
+      this.ctx.save();
+      this._setTextStyles(this.ctx);
+
+      lines = this._wrapText(this.ctx, this.text);
```
kangax: Is this an undeclared var?

src/shapes/textbox.class.js
```diff
@@ -0,0 +1,321 @@
...
+     * the amount that the group's increases. This is to maintain the effective
+     * scale of the Textbox at 1, so that font-size values make sense. Otherwise
+     * the same font-size value would result in different actual size depending
+     * on the value of the scale.
+     * @param {String} key
+     * @param {Any} value
+     */
+    setOnGroup: function (key, value) {
+      if (key === 'scaleX') {
+        this.set(key, Math.abs(1 / value));
```
kangax: Can we do `this.set('scaleX')` here? Would be a little clearer to read, IMO.

src/shapes/textbox.class.js
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        // adjust selection start
+        selectionStart -= lineLength;
+
+        // keep track of removed characters
+        removed += lineLength;
+
+        // if there was a newline, we need to account for the newline character
+        if (this.text[removed] == '\n') {
```
kangax: == &rarr; ===

src/shapes/textbox.class.js
```diff
@@ -0,0 +1,321 @@
...
+          lineIndex: numLines - 1,
+          charIndex: this._textLines[numLines - 1].length
+        };
+      }
+
+      for (var i = 0; i < numLines; i++) {
+        lineLength = this._textLines[i].length;
+
+        // case: we are in the current line
+        if (selectionStart < lineLength + (this.text[removed + lineLength] == '\n' ? 1 : 0)) {
```
kangax: == &rarr; ===

src/shapes/textbox.class.js
```diff
@@ -0,0 +1,321 @@
...
+      mt: false,
+      mr: true,
+      mb: false,
+      mtr: true
+    };
+  };
+  /**
+   * Contains all fabric.Textbox objects that have been created
+   * @static
+   * @memberof fabric.Textbox
```
kangax: memberof &rarr; memberOf

kangax:
@inssein @jafferhaider left few (minor) comments; other than that, I think we're ready to merge this in.

inssein:
@kangax thanks for the quick review - i will make those changes asap (probably tomorrow)

kangax:
@inssein @jafferhaider just to make it clear, since we're following [semver](http://semver.org/) now, this will have to be part of 1.6.0 release, since new features must bump minor version. We were planning to do 1.5.1 before that but I guess we might as well just merge this and go for 1.6.0. Probably in the next few days, once @asturur is back, since I need to coordinate on few active tasks with him.

inssein:
@kangax sounds good to me

inssein:
@kangax done.

kangax:
LGTM /cc @asturur @jafferhaider

sapics:
@inssein Thank you for your work!

setOnGroup is necessary for keep scaling of textbox.
So, i guess that we need to add setOnGroup to group.class.js
```javascript
    /**
     * @private
     */
    _set: function(key, value) {
      var i = this._objects.length;
      if (this.delegatedProperties[key] || key === 'canvas') {
        while (i--) {
          this._objects[i].set(key, value);
        }
      }
      else {
        while (i--) {
          this._objects[i].setOnGroup(key, value);
        }
      }
      this.callSuper('_set', key, value);
    },
```
and to object.class.js as
```javascript
    /**
     * This callback function is called by the parent group of an object every
     * time a non-delegated property changes on the group. It is passed the key
     * and value as parameters. Not adding in this function's signature to avoid
     * Travis build error about unused variables.
     */
    setOnGroup: function() {
      // implemented by sub-classes, as needed.
    },
```

inssein:
@sapics ah I see what I missed, as it only comes up when a textbox is grouped. That said, I don't think that's a good way of solving that problem...

@kangax @asturur any suggestions on how else to do it?

inssein:
@sapics what exactly are you trying to do? I feel like I should be removing the code for setOnGroup as it doesn't make sense. Why would you want to reverse the effects of scale on the group?

I am thinking that perhaps that code is better served by extending the group shape on an application level.

sapics:
Personally, I want to reverse the effects of scale on the `selection` group.

sapics:
In your textbox.class.js, there is a setOnGroup function.
If you are not necessary, i think that it would be better to remeve setOnGroup function from textbox.class.js.

jafferhaider:
I added setOnGroup as a general purpose way for objects to respond to properties set on the Group, didn't want to bog down the set() function with additional logic (@kangax and I discussed this in the very first PR). In Textbox it is needed to reverse the value of scale which is set by group ... because if we let the group scale size of font on the Textbox ... the value of fontSize becomes irrelevant.

@kangax I was working on fixing some selection bugs that had cropped up after recent commits in Fabricjs master

inssein:
@jafferhaider @sapics ok I have committed it for now, perhaps I can address it in a later commit if I figure out a better way.

sapics:
@jafferhaider Thank you for your comment and great job!
I am now using your pr, in my project, it is so great!!

I guess that we need to remove __oldScaleX when ungrouping, is it right?

jafferhaider:
Thanks @sapics. Yep, `__oldScaleX` should be removed when ungrouping. 

@inssein I see that there are some changes missing in your PR that were made to existing FabricJS files (e.g. some itext mixins). Please review changes here: https://github.com/kangax/fabric.js/compare/master...PosterMyWall:TextboxPullRequest3 ... we should add them to this PR.

asturur:
@jafferhaider what selectiin bugs? do you speak about text selection bugs or general selection bugs?

jafferhaider:
@asturur actually in this PR @inssein has fixed the selection issue but that results in a cursor placement bug. Create a Textbox with a few lines of text. As you go down while clicking on each line, notice how the cursor renders increasingly to the right of where you click.

The root cause is the difference in selectionStart and selectEnd values because IText assumes that every line will end with a newline char, whereas that is not the case in Textbox. I'd managed to fix both the selection and cursor placement bug in my original code but they cropped up again with recent IText changes. :(

No general bugs, this is specific to Textbox.

inssein:
@jafferhaider yeah I noticed that too, maybe I can look at that tonight.

Bugs Remaining:
* Cursor placement issue when moving between lines
* Style does not seem to carry over when it auto wraps
* Wrapping doesn't work properly when there is more than one style in a line

inssein:
As the above post mentions, there one bug left. The bug is that when the width of the textbox changes, the styles do not adjust accordingly.

I have spent a couple of hours on this already and I cannot seem to find a way to do this. There may be a way if we were storing styles as indices of ``` this.text ```, and not a 2d array of lineIndex/charIndex (but even this would be pretty hard, as line wraps don't offset this.text with characters like a new line does).

The main issue here is that ``` _wrapText ``` could add or delete lines depending on the changes of ``` _dimensionAffectingProps ```, but I have yet to find a way to translate those changes into the style object. I don't think its possible without other changes to how textbox works.

inssein:
@jafferhaider totally missed your earlier comment regarding changes I missed. I just reviewed your changes, and it seems like the only ones I missed were in itext_behavior.mixin.js, but I have no idea what that does. I wanted to keep the changes to strictly textbox related, and if they are a bugfix for itext, then they probably belong in a separate PR.

jafferhaider:
@inssein There's one change in `itext_behavior.mixin.js` that used to be needed for Textbox ... that's the line that calls `this._splitTextIntoLines();`. This function allows Text sub-classes to define their own line splitting function. In the Textbox's case, this function implements line wrapping.

Other changes are IText bug fixes ... you can leave them out. But you should make the change in line 523
https://github.com/kangax/fabric.js/compare/master...PosterMyWall:TextboxPullRequest3 (scroll down to itext_behavior.mixin.js).

Btw, nice job on fixing the cursor bugs.

jafferhaider:
Are we waiting for Styles support in Textbox?

inssein:
@jafferhaider regarding your previous comment, still not sure what you are talking about because I do have the splitTextIntoLines() in my code (https://github.com/inssein/fabric.js/blob/textbox/src/mixins/itext_behavior.mixin.js). 

@kangax are we waiting for all bugs to be fixed before this gets merged? The biggest issue left is getting styles to conform to changes in wrapping (adjusting width, etc).

inssein:
Ok I have figured out how to maintain styles while wrapping, but its a fairly extensive change. I just need to now figure out how to maintain style while inserting text, especially when it auto wraps. After that commit, everything should be in a merge-able state.

inssein:
I still have a few bugs left, but I got a lot done today. All the changes I have made are about how textbox reads the style object. The style's object is now directly correlated to line breaks, and not _textLines as that is affected by line wraps as well.

src/shapes/itext.class.js
```diff
@@ -917,14 +911,64 @@
...
+    _getLineStyle: function(lineIndex) {
+      return this.styles[lineIndex];
+    },
+
+    /**
+     * @param {Number} lineIndex
+     * @param {Object} syle
+     * @private
+     */
+    _setLineStyle: function(lineIndex, syle) {
```
sapics: syle -> style :)

sapics:
@inssein Thank you for great jobs!

I create http://jsfiddle.net/sapics/51u08n39/111/ for checking with all!
In the jsfiddle, justified text style looks strange.

asturur:
Style with justify looks strange generally. It is a little bit bugged in the library.

sapics:
@asturur Thank you for your information!

I add http://jsfiddle.net/sapics/51u08n39/113/ to check IText behavior, because this PR includes some fixes for IText.

inssein:
@sapics @asturur I will be working again on this tomorrow, and hopefully I should be able to get it to an acceptable state.

Notes for future: I think the way we store styles, and the text is not going to scale (especially if we start considering advanced functionality like lists, subscripts, superscripts, etc.). I think it would be best represented in some sort of XML structure (ie: HTML, Adobe TLF, etc).

asturur:
i do not know how much elaborate itext and so texybox should be.

we already have lots of things to fix and add for it.

yes style does not scale good with functionalities.

pleae if you are fixing any itext or text functionality, do it separately, open a PR that demonstrate the bug and fixes it. so we can integrate fixes apart from textbox.

jafferhaider:
Thanks @inssein. Looking forward to taking the final code for a spin.

inssein:
@asturur the only Itext fix I have made is here: https://github.com/kangax/fabric.js/pull/2237/files#diff-72c0bde139ef2f0654ef458619a1b618R611. What would you like me to do about that, remove it from here and include it in a separate PR?

As of now, I only have one bug that remaining that I know of. When the textbox is made so small that the a word doesn't fit, it splits the word and add's spaces (from the initial implementation).

It needs to either:
1) If it add's spaces, it also needs to add the styles for it, which would be pretty complicated because it needs to be removed when textbox is made bigger.
2) Limit the textbox so that it cannot go smaller than the longest word.

inssein:
@kangax @asturur the build failed because of the addition of jshint to travis (https://github.com/kangax/fabric.js/commit/750a4a5e4644e3fc9820993c1eb4448ead4a6c8e). I fixed one of them which was in itext since I was working on it, but the other ones should be fixed separately right?

kangax:
Yep, we have a separate PR for jshint warnings. I had no idea we didn't have it as part of CI which is bad and so I added it earlier today.

inssein:
@jafferhaider @asturur @sapics I think I am going to make a change to the initial implementation in order to fix the last bug, let me know what you guys think.

__The Problem:__ When the longest word doesn't fit, it breaks up the words, which not only isn't right (since it doesn't hyphenate them), but introduces styling issues because now there are additional spaces due to the new "fake" words.

__My Solution:__ Change the textbox implementation so that a textbox cannot be smaller than the biggest word. This would introduce a new private property (dynamicMinWidth?) that dictates the minimum size of the textbox. I see that textbox already has a minWidth, but it would be changed so that getMinWidth() returns Math.max(dynamicMinWidth, minWidth). This would introduce some new issues as well (textbox width changing when style changes), but IMO it's a better solution.

inssein:
Alright, I committed the solution I mentioned above..

jafferhaider:
@inssein not a fan of this solution since it has the potential of introducing usability issues. The user might try to resize their Textbox and not realize that it is the long word that is preventing them from doing it. Also, the behavior in the original version is consistent with most common textbox implementations ... have a look at what Flash's textbox does ... the HTML Textbox wraps the word as well (without hyphens).

I think you should focus on low-level code for a fix, rather that limiting the user.

inssein:
@jafferhaider fair enough. I can make it so that its an option, as I prefer it the way it's currently working. The only thing is I don't have a solution for the styles issue when it breaks up words.

I will probably be working on this mid next week, see if I can think of a solution until then.

@kangax @asturur I think what I have done up to now is an acceptable solution, and it is bug free as far as I am aware (probably more will be found as people start using it).

inssein:
Sorry guys, I finally integrated this with my own application of fabric.js and there are some problems. I will be working those out over the next few days.

The bugs are mostly with alignment (if there are two big words wrapped in 2 lines, the first line has a space at the end). Removing the space adds a lot of additional bugs and that is what I am dealing with right now, among other things.

inssein:
I believe I have fixed all bugs in my own implementation, but not all the code is ported back yet. There is only one bug left which pretty much involves deleting the textbox_key_behaviour file as it is no longer necessary.

The only other thing missing now is the feature to allow it to wrap down to a single character.

jafferhaider:
Great job @inssein ... looking forward to seeing the code.

inssein:
Alright, I am now done everything and this is ready to be merged.

@jafferhaider I tried quickly implementing the feature that would allow you to wrap text all the way down to a single character, but I was having a lot of styling issues with it. I am sure it's possible, but I don't have anymore time right now. 

@kangax I finally set my phpstorm javascript code style settings to match what fabric.js and auto formatted all the changes I made. The only unfortunate part about this is that I auto-formatted all of itext_behavior.mixin.js, which added a lot of additional changes (JSCS doesn't check for space after "function"). I checked what the correct setting should be regarding that, but it's pretty inconsistent. We should either add it to "requireSpaceAfterKeywords" in jscs.json, or add a check that a space doesn't exist.

kangax:
I'm in favor of space after parens but not before, like 'function getSomething() { }'

Please feel free to send a PR for jscs change! Thanks.

inssein:
@kangax ah ok then, I will have to reformat all my changes so there is no space before parenthesis, because every function has it.

i'll see if that can be ensured via JSCS and submit a pull request.

Other than that, when do you see this being merged?

kangax:
Honestly as soon as possible but I'm going through a rough patch now — tons of stuff at work and my laptop just broke yesterday!

inssein:
@kangax damn that sucks. Hope nothing major is wrong with the laptop.

Anyways, I reverted the code style changes. I also created a separate branch for the jscs change, but there is a lot of code failing the check. I will create a separate PR at a more appropriate time later on, its probably good to get some of these other PR's in first.

The jscs change in question is ```"disallowSpaceAfterKeywords": ["function"]```.

droplab:
Hi guys, thanks for all your work on a great library. I'm having trouble trying to integrate this Textbox class into an existing custom build I have (text,gestures,interactions). 

Uncaught TypeError: Cannot read property 'prototype' of null

	function callSuper(methodName) {
		var fn = this.constructor.superclass.prototype[methodName];
		return (arguments.length > 1) ? fn.apply(this, slice.call(arguments, 1)) : fn.call(this);
	}


Any suggestions?

nokoz:
Sorry, but how can I use this method? I connect the jquery library, fabric library and textbox class. As a result, I have the error: "Uncaught TypeError: Cannot read property 'fontFamily' of null"

inssein:
@nokoz You have to checkout this branch and build it.

@droplab very hard to help you with that much detail. The error happens earlier.

@kangax @asturur any updates? I have been using it for a while and I think there are some improvements that could be made, but it would be better if this was reviewed and merged first.

kangax:
Aaaand merged! Thanks to everyone involved. I'm sure we'll need to fix a thing or two in a future but this looks like a great start. If someone is willing to "add" it to fabricjs.com, that would be awesome. Perhaps, as an option in kitchensink or a separate demo.

PR #2238, opened 2015-05-27
image.class.js not properly calling _initConfig when initializing

inssein:
There was a small bug introduced in https://github.com/kangax/fabric.js/commit/ed5b1b66b05fe2f2433cce8da6340d2ebd27dc3b which didn't carry initialization options through.


asturur:
:ok: !

PR #2243, opened 2015-05-29
Make IText support accent-based languages

### LINKED ISSUE #2051: [IText] IText is unusable for accent based languages
Currently, the IText input binding strategy has some issues and one of most critical one is that IText is completely unusable for accent based languages (such as Portuguese, Spanish, Japanese, Chinese, etc) due the lack of support to [dead keys](http://en.wikipedia.org/wiki/Dead_key).

**The problem**
IText uses a hidden input for capturing user's input, through the [`keydown` event](https://github.com/kangax/fabric.js/blob/4ad253ffba4f98b3161d46c6a3f86b9d403a16c7/src/mixins/itext_key_behavior.mixin.js#L160). The problem is that `keydown` events originated from dead keys cannot be converted into characters, because the event does not provide the character's code but a fixed key code (229) instead.

Examples:

Key | Key Code | Expected Result
----- | ------------- | --------
~ | `229` | ~
' `+` c | `229`, `229` | ç
~ `+` a | `229`, `229` | ã
" `+` u | `229`, `229` | ü
~ `+` ~ | `229`, `229` | ~~
~ `+` ~ `+` ~ `+` a | `229`, `229`, `229`, `229` | ~~ã
~ `+` [space]| `229`, `229` | ~ [trimmed, no space]

Notice that pressing `a` after a dead key produces another dead key event. So, it's very hard to handle it in the way it's currently architected.

Holding the key of a given character to choose one of its variants is OS specific and almost no user of accented keyboards is familiar with this, so it's not a good alternative IMHO. I've been trying to fix it in the last days, but seems like we should change the way it is currently done, once we'll have to reimplement basic browsers/OS features bringing a lot of unnecessary code overhead.

Unfortunately, we've been working on a big startup project in the last months without noticing this issue and maybe we'll have to stop it immediately, at least until discover a workaround.

Related issues: #1175 and #1954

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9855309-itext-itext-is-unusable-for-accent-based-languages?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jcppman:
Hi there, thanks for the awesome library!

Purpose
-----------

This PR is for making IText support accent-based languages IMEs (will fix #2051 ). This feature was on milestone v1.5.1 but removed, not sure if the way I implement this is the desired way, might cause some duplicated works if you guys already have a plan about how to implement this, but I guess it's worth trying :) 

Currently we use this modified version in our desktop webapp projects and it works fine/correctly in both windows and OSX, haven't test it on other platforms yet.

Concerns & Changes
------------

Firstly, the original `keypress` event handler is kept for normal input scenarios, because in order to fully support the accent-based languages a dirty scan must be done to detect the actual changes (detail is explained below). We can skip that if the incoming input action is not accent-based.

This implementation use the fact that input actions in accent-based languages usually
fire `keydown` with keyCode `229` and following event `input`,
records the latest `keyCode` and handle accent-based input separately. Another scenario is that only an `input` event is fired (e.g: user select a character from options) but it will only happens after one or more regular `keydown`, `input` event sequences, in this case the `latestKeyCode` is still `229` so it is covered.

In accent-based input methods, there's no predictable results according to a given keystroke, the length might be `n` characters longer/shorter or even unchanged, the content might be entirely re-written or changed partially, so we need to scan `hiddenTextArea.value` and `this.text` from both sides to extract the 'changed parts' and the range, then apply it on IText.

one major change except the input handling part is:
`hiddenTextArea.value`, `hiddenTextArea.selectionStart` and `hiddenTextArea.selectionEnd` should not be touched during these input actions, otherwise the input process & state of IME will be intercepted. So an additional parameter `noTextAreaUpdate` is added to several methods to determine if the opperation is triggered by accent-based languages input.

One last thing
------------------

To make it works fluently, one additional change has to be done: move the `hiddenTextArea` closer to the belonging `IText`, but since #1990 is already WIP (and almost done), I'll leave this unchanged for now and fix problems caused by the merging in the future (#1990 will break this PR in some edge cases).

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -478,6 +478,8 @@
       this.hiddenTextarea && this.canvas && this.hiddenTextarea.parentNode.removeChild(this.hiddenTextarea);
       this.hiddenTextarea = null;
 
+      this.latestKeyCode = null;
+
       this.abortCursorAnimation();
       this._restoreEditingProps();
       this._currentCursorOpacity = 0;
@@ -502,12 +504,12 @@
     /**
      * @private
      */
-    _removeCharsFromTo: function(start, end) {
+    _removeCharsFromTo: function(start, end, noTextareaUpdate) {
       while (end !== start) {
         this._removeSingleCharAndStyle(start + 1);
         end--;
       }
-      this.setSelectionStart(start);
+      this.setSelectionStart(start, noTextareaUpdate);
     },
 
     _removeSingleCharAndStyle: function(index) {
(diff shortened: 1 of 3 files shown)


asturur:
so you would like to have a mixed input scenario, using onkeypress an oninput, instead of migrating everything to oninput?

jcppman:
My reason of keeping `onKeypress` was that `keydown` and `input` events don't give us the actual keyCode (when type in 'A' and 'a', `e.keyCode`s are the same 'A'), so I started with the current solution. But now it seems like with some extra efforts they can be merged

To avoiding dirty scanning for regular input, here are two ways I can think of:

- use `keypress` to record the actual keyCode, maybe a new property `this.latestChar` or something
- use `selectionStart` to locate the incoming char

Which way do you prefer?

jcppman:
So this is kinda like a trade off, a more elegant, universal way or a more straight forward way. :)

jcppman:
Turns out that the unification is the only way.

Previous solution is not robustic enough due to different event sequences accent-based IMEs may trigger. For instance in Firefox it won't trigger any "deadkey" when user is using IME.

PR #2256, opened 2015-06-02
Fix style parsing

### LINKED ISSUE #2241: Trim issue in parseStyleString while parseSVGdocument
Safari 5 fails on "parseSVGdocument" while it tries to "parseStyleString".

You except always to have at least two elements while splitting the "chunk" but for some reason in Safari 5 it returns just a single empty string, not sure why was to lazy to fully trace it down.

Following simple check worked for me on Line: [3089] (https://github.com/kangax/fabric.js/blob/master/dist/fabric.js#L3089)
```
  function parseStyleString(style, oStyle) {
    var attr, value;
    style.replace(/;$/, '').split(';').forEach(function (chunk) {
      var pair = chunk.split(':');

      if ( pair.length > 1 ) {

        attr = normalizeAttr(pair[0].trim().toLowerCase());
        value = normalizeValue(attr, pair[1].trim());

        oStyle[attr] = value;
      }
    });
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/18398560-trim-issue-in-parsestylestring-while-parsesvgdocument?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
closes #2241

Changing regex should be fine instead of testing pair length.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -275,7 +275,7 @@
    */
   function parseStyleString(style, oStyle) {
     var attr, value;
-    style.replace(/;$/, '').split(';').forEach(function (chunk) {
+    style.replace(/;\s*$/, '').split(';').forEach(function (chunk) {
       var pair = chunk.split(':');
 
       attr = normalizeAttr(pair[0].trim().toLowerCase());
--- a/test/unit/parser.js
+++ b/test/unit/parser.js
@@ -146,6 +146,17 @@
     deepEqual(fabric.parseStyleAttribute(element), expectedObject);
   });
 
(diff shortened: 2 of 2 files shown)


kangax:
Can we add a unit test?

asturur:
@kangax done.

Issue #2234, opened 2015-05-27
rotated i-text , correct cursor only on origins : Top Left


sarath-prohaktiv:
Reference URL :http://fabricjs.com/kitchensink/
Steps to produce the issue
#1) Select i-text ["foo lorembar
baz lorem
quux"]

#2) Rotate the text
#3) set Origin X, Origin Y as center.
#4) Try to click on third line.
#5) Unable to move cursor to that position

PR #2257, opened 2015-06-03
Clean origins

asturur:
Cleaning of object origins code.
easier to add functions to switch from topleft, topcenter, bottomright and so on.
Necessary but not sufficietn to fix the pointer of rotated itext that is not top left (#2234)

### DIFF
--- a/src/mixins/object_origin.mixin.js
+++ b/src/mixins/object_origin.mixin.js
@@ -1,8 +1,41 @@
 (function() {
 
-  var degreesToRadians = fabric.util.degreesToRadians;
+  var degreesToRadians = fabric.util.degreesToRadians,
+      originXOffset = {
+        left: -0.5,
+        center: 0,
+        right: 0.5
+      },
+      originYOffset = {
+        top: -0.5,
+        center: 0,
+        bottom: 0.5
+      };
 
   fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prototype */ {
+
+    /**
+     * Translates the coordinates from origin to center coordinates (based on the object's dimensions)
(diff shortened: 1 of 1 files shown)


asturur:
wait one test missing.

asturur:
Merged similar code, avoid rotating point if the angle is 0. General code reduction.
I noticed that those functions does not have tests, so we should add some.

src/mixins/object_origin.mixin.js
```diff
@@ -1,40 +1,55 @@
 (function() {
 
-  var degreesToRadians = fabric.util.degreesToRadians;
+  var degreesToRadians = fabric.util.degreesToRadians,
+      originXOffset = {
+        left: -0.5,
```
kangax: These are the only things that look confusing to me. Why -0.5, 0 and 0.5?

asturur: because we are always working with half width and half height to translate
the points.

i writing tests for all those functions to be sure they give same result as
old one.
 Il 06/giu/2015 05:07, "Juriy Zaytsev" <notifications@github.com> ha
scritto:

asturur:
@kangax they give same results in tests but are shorter, also they do not rotate point if there is no angle.

kangax:
@asturur you mean shorter in code? So you think this is OK to merge?

asturur:
i think yes.
i separated from test to permit a painless revert.

if you think code looks better and is worth ro skip rotation on  angle 0, merge it.

asturur:
Note: i forgot 1 small optimization ( double calculation of sin and cos ) and to change this.getWidth() with this.getTransformedDimensions().

PR #2258, opened 2015-06-03
Line with arrow heads

bbbalabhaskar:
Added arrows to line if provided value arrows in options of line, arrows will get added basing on the count.
If value is more than 1 or less than -1,arrows will get added on both sides.
If -1 is provided, to top of line and bottom side of line if 1.
If no value or 0 is provided line will render as it is.

### DIFF
--- a/src/shapes/line.class.js
+++ b/src/shapes/line.class.js
@@ -74,7 +74,13 @@
       this.set('y1', points[1]);
       this.set('x2', points[2]);
       this.set('y2', points[3]);
-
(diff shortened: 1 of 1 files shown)


kangax:
Please look at the guidelines for contributing new features — https://github.com/kangax/fabric.js/wiki/How-to-contribute-to-Fabric#for-any-new-features-please-account-for

andrevlins:
That's a great feature, I need this but can't make it work.

PR #2261, opened 2015-06-05
Added canvas.renderOnStackMove and canvas.renderOnObjectCenter

matanhershberg:
Setting these to false will disable automatic renderAll() calls for
corresponding methods in order to increase performance when performing
many
operations at once. Specifically, moving objects in the stack of drawn
objects, and centering objects on the canvas.

Applies to:
Canvas.moveTo
Canvas.sendToBack
Canvas.sendBackwards
canvas.bringForwards
Canvas.bringToFront

Canvas.centerObject
Canvas.centerObjectH
Canvas.centerObjectV

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -105,6 +105,24 @@
     renderOnAddRemove: true,
 
     /**
+     * Indicates whether {@link fabric.StaticCanvas#moveTo}, {@link fabric.StaticCanvas#sendToBack}, {@link fabric.StaticCanvas#sendBackwards}, {@link fabric.StaticCanvas#bringForward}, and {@link fabric.StaticCanvas#bringToFront} should also re-render canvas.
+     * Disabling this option could give a great performance boost when moving a lot of objects in the stack of drawn objects at once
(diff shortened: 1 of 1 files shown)


kangax:
I'd rather we have one property to avoid having (to remember) so many. Don't think such granularity is needed. We can have, for example, `renderOnChange` which would fire control `renderAll` during all kind of changes. Alternatively, we can support specifying array of events would would control that. For example, `renderOnlyOn: "object:added"` or `renderOnlyOn: "object:rotated"`, etc.

matanhershberg:
I agree, I was thinking myself that having just one property that would disable automatic rendering globally would be the most ideal, unfortunately I have only been using this library for a couple months, and as I'm still not familiar with many of the features I was worried about breaking something.

Also I couldn't get the test to run, and I wasn't sure if there is a way to test that the renderAll function was called if it needed to, so I ended up testing manually, which I'm guessing might be problematic if we want to implement that for the entire library.

What do you think?

asturur:
this is gonna be enabled on fabric 2.0, with the name automaticRender ( we can still change it ) but the concept is the one express by kangax. one property for all the events.

PR #2265, opened 2015-06-05
Fabric.loadSVGFromString Fails on Empty Path

timkendall:


timkendall:
Realized I built this specifically for Node.js

kangax:
There seems to be few irrelevant things in this PR so I'm gonna close it for now. Please resubmit if still relevant.

PR #2267, opened 2015-06-07
Add 100+ test to object origin mixin

asturur:
To be sure to do not mess up them in the future.

### DIFF
--- a/test/unit/object_origin.js
+++ b/test/unit/object_origin.js
@@ -1,0 +1,351 @@
+(function() {
+  
+  var rectOptions = {
+    'originX':                  'left',
+    'originY':                  'top',
+    'left':                     35,
+    'top':                      45,
+    'width':                    20,
+    'height':                   40,
+    'fill':                     'rgb(0,0,0)',
+    'strokeWidth':              2,
+    'angle':                    0,
+    'scaleX':                   2,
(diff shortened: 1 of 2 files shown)


asturur:
@kangax for me this is done. After pulling this we can pull also the #2257 if still mergeable.

test/unit/object_origin.js
```diff
@@ -0,0 +1,351 @@
...
+    deepEqual(rect.top, 56.47152872702093);
+    equal(rect.originX, 'left');
+
+    p = rect.adjustPosition('center');
+    deepEqual(rect.left, 67.76608177155967);
+    deepEqual(rect.top, 67.94305745404185);
+    equal(rect.originX, 'center');
+
+    p = rect.adjustPosition('right');
+    deepEqual(rect.left, 84.1491226573395);
```
kangax: I'm a little unsure about such precise fractions. Usually they always cause problems. But maybe OK in this case.

asturur: i was thinking same.
first time that this create some error in tests we will add toFixed(NUM_FRACTION_DIGIT) to every coordinate  of every point of the test.

kangax: Alright, let's see how it goes

PR #2272, opened 2015-06-10
fix paste of clipboard content

### LINKED ISSUE #2270: [IText] paste is broken when content is plaintext
Tried on Chrome 43 and Firefox 38 and Safari 8.

It works correctly if content is copied from IText, otherwise nothing happens and an error is thrown: `Uncaught TypeError: Cannot read property 'length' of undefined.`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/21318835-itext-paste-is-broken-when-content-is-plaintext?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jcppman:
closes #2270 

use `copiedStyle` only if the pasted content source is `copiedText`.

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -114,7 +114,8 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    */
   paste: function(e) {
     var copiedText = null,
-        clipboardData = this._getClipboardData(e);
+        clipboardData = this._getClipboardData(e),
+        useCopiedStyle;
 
     // Check for backward compatibility with old browsers
     if (clipboardData) {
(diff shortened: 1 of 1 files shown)


asturur:
Please, i cannot reproduce it on kitchensink.
Could you please tell me where did you found the bug 
version of fabricjs ( number or if you pulled github ) and browser used. ( and version )

jcppman:
Hi, I'm using the master branch, latest commit: 2d38f9eb22de519d302187013104af896cc76ad2

I tried on on:

- Chrome 43
- Firefox 38
- Safari 8

To reproduce the bug, make sure you don't copy anything from IText before, otherwise even if you copy anything from other places which doesn't set `this.copiedStyles`,  previous value is still there. 

So it means that in current master branch, in order to make an IText work, user has to copy a value from itself frist.

asturur:
ok i ll check it as priority and see where the problem came from.

jcppman:
The problem comes from `_insertStyles` cuz in `insertChars` we pass a `true` to use `copiedStyle` but in this scenario it is `undefined`.

asturur:
@kangax this is fine, you should merge this.

asturur:
need to resubmit with additional fix

PR #2273, opened 2015-06-10
Differing sizes for control points on Object

jasond-s:
I have completed a preliminary feature addition for this in the related branch. I want to use the pull as a platform for conversation as much as review.

I have a need for changing the sizes independantly for the control points on an active object. 

There are obviously some ways I could have done this externally to the library, but on my inspection it seemed quite easy to implement the feature in the mixin for the controls itself.

To achieve this I have:

1. Added a `cornerSizeOverrides` property to `Object`.
2. Where the `cornerSize` is used we now have to look for overrides.
3. Removed all assuptions that the corner bounding box hypotenuse internal angle is 45deg.
4. Removed all assumptions that the width and height of each point are always equal.

This has effected:
- The private `_drawControl` function. 
- The private `_setCornerCoords` function.

Obviously this is a non trivial change and I have tried to maintain exisiting optimisations where they were present. Adding a new property to `Object` I also accept is quite a big change. But it does seem like the most succinct way of achieving my requirement.

Please let me know your thoughts. If this is not acceptable then I will need to implement this another way asap. The code detailed is 'working for me', but I might need some more insight as this is the first time I have really delved into the source code.

Thanks.

### DIFF
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -1,6 +1,7 @@
 (function() {
 
   var degreesToRadians = fabric.util.degreesToRadians,
+      radiansToDegrees = fabric.util.radiansToDegrees,
       //jscs:disable requireCamelCaseOrUpperCaseIdentifiers
       isVML = function() { return typeof G_vmlCanvasManager !== 'undefined'; };
   //jscs:enable requireCamelCaseOrUpperCaseIdentifiers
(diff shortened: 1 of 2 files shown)


tigerswithguitars:
Here is a quick example of how the override might be used...

```js
var obj = new fabric.Rect({width: 50, height: 30, left: 100, top: 100});

obj.cornerSizeOverrides = {
    mtWidth: 40,
    mbWidth: 40,
    mlHeight: 20,
    mrHeight: 20
};
```

or possibly with a getter for calculated values

```js
var obj = new fabric.Rect({width: 60, height: 30, left: 100, top: 100});

obj.cornerSizeOverrides = {
    get mtWidth() {
        return obj.getWidth() - 20;
    },
    get mbWidth() {
         return obj.getWidth() - 20;
    }
};
```

asturur:
the object ```this.cornerSizeOverrides``` would be something like:

```
{
 trWidth: x,
 trHeight: y,
 tlWidth: j,
 tlHeight: z,
 blWidth: q,
.... and so on...
} 
```
?

This looks like a particular application requirement, and would be easily implemented with an ovveride of the functions.

For my personal point of view fabric is already full of very particular features that if i would add something on controls, this would be custom functionality to corners.

Other than my personal opinion, i would not define the cornerSizeOverrides as null, because doing this, as soon as you set just one of them, you would calculate the standard ```newTheta``` and ```cornerHypotenuse``` for every of the other 8 controls.
I would make the object ready with trWidth ( and others ) equal to 0 and i would test for:

```
if (size[point + 'Width'] || size[point + 'Height']) {
  ... special dimension code ....
}
```

tigerswithguitars:
Thanks for your input. I appreciate that the catalyst for this was a fairly specific application requirement, but as I was doing it as an override of the functions, it felt like sometime maybe someone else might appreciate. 

As for your initial value suggestion, this is a much cleaner way of doing it. I'll add that to the pull... 

Out of interest, what custom functionality are you thinking of adding to the corners?

kangax:
Hey @tigerswithguitars, I also think this is a bit too application specific to be included in core Fabric. We're trying really hard to avoid featuritis with our already large code base. I am, of course, open to leaving this as a feature request and seeing if anyone else needs it. If this ever becomes a popular request, we can certainly revisit it. Cheers!

tigerswithguitars:
@kangax thanks for your reply. I fully understand you don't want the library to become bloated. I'll write something up about it, then, if you want to get rid of the pull in the future because it goes dead, there will be some info online if someone goes looking for the feature :+1:

PR #2277, opened 2015-06-11
character correction

sapics:
This pr dose not affect to the feature.
Just a character correction and for better minify by removing `var`.


asturur:
:+1:

sapics:
Grazie per :+1:

PR #2278, opened 2015-06-11
fix undefined function call

sapics:
just a function name correction.


asturur:
What the hell?
I tested everything before committing, how could i use the function in that way?

i go to bury myself.

Issue #2287, opened 2015-06-13
Error in regex in parseAttributes


dgodard:
There is a problem with regex in Parser.js in parseAttributes line 763 :
```javascript
// if there's a parent container (`g` or `a` or `symbol` node), parse its attributes recursively upwards
      if (element.parentNode && /^symbol|[g|a]$/i.test(element.parentNode.nodeName)) {
```

`^symbol|[g|a]$` match also for "svg" element node name because it tests if nodeName starts with "symbol"  or contains "g" or "a" char at the end

It needs parenthenses : It must be `^(symbol|[g|a])$` for an exact match "symbol" "a" or "g".

kangax:
Good catch, thanks! This looks like a regression, since we added symbol recently and I missed the simplest mistake of grouping and anchors :P

If you could send a PR, that would be great. Or one of us will get to it eventually. Also, a minimal failing test case would be perfect!

asturur:
@dgodard  what svg was failing with this error?

PR #2292, opened 2015-06-15
Fix regex in parseAttribues (issue #2287)

dgodard:
Regex fix in parseAttributes as described in issue #2287.
^symbol|[g|a]$ match also for "svg" element node name because it tests if nodeName starts with "symbol" or contains "g" or "a" char at the end

It needs parenthenses : It must be ^(symbol|[g|a])$ for an exact match "symbol" "a" or "g".


src/parser.js
```diff
@@ -761,7 +761,7 @@
         svgUid = element.getAttribute('svgUid');
       }
       // if there's a parent container (`g` or `a` or `symbol` node), parse its attributes recursively upwards
-      if (element.parentNode && /^symbol|[g|a]$/i.test(element.parentNode.nodeName)) {
+      if (element.parentNode && /^(symbol|[g|a])$/i.test(element.parentNode.nodeName)) {
```
asturur: Why not ^(symbol|g|a)$ ? without [ ] ?

dgodard: You're right, it will test exact match : "symbol" "g" or "a"

PR #2294, opened 2015-06-16
fix setDimensions css unit

Julyyq:
When invoke `setDimensions` method with using options `{cssOnly: true}`, the `cssValue` variable is not added postfix "px" unit, so the`cssOnly: true` option is not working.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -543,10 +543,10 @@
 
         if (!options.cssOnly) {
           this._setBackstoreDimension(prop, dimensions[prop]);
-          cssValue += 'px';
         }
 
         if (!options.backstoreOnly) {
+          cssValue += 'px';
           this._setCssDimension(prop, cssValue);
         }
       }
--- a/test/unit/canvas_static.js
+++ b/test/unit/canvas_static.js
@@ -1078,6 +1078,22 @@
     equal(canvas.getHeight(), 500, 'Should be as the backstore only value');
   });
 
+  test("setDimension css only", function() {
+    canvas.setDimensions({ width: 123 });
+    canvas.setDimensions({ width: 500 }, { cssOnly: true });
+    
+    equal(canvas.lowerCanvasEl.style.width, 500 + 'px', 'Should be as the css only value');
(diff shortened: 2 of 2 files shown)


kangax:
Don't understand the issue here

Julyyq:
@kangax See the following code snippet:

    if (!options.cssOnly) {
      this._setBackstoreDimension(prop, dimensions[prop]);
      cssValue += 'px';
    }
    if (!options.backstoreOnly) {
      this._setCssDimension(prop, cssValue);
    }

If set `cssOnly: true`, the line code `cssValue += 'px';` will not be called, and the `this._setCssDimension(prop, cssValue);` method pass the cssValue variable without unit "px".

kangax:
Strange, how did it even work before? I remember people saying that it works (I personally haven't used it)

kangax:
Also, a test would be nice for this!

Julyyq:
@kangax Unit test-case is done, definitely it's a bug so I don't open a new branch for pull request. Hope it doesn't matter.

inssein:
@Julyyq there may be an invalid assumption on what that does. Check out the tests that are broken because of the change.

Julyyq:
@inssein I clone the master branch without any code modifications, run `npm test`, and occurs several errors. It's not about my changes.

sapics:
@Julyyq check travis ci result in https://travis-ci.org/kangax/fabric.js/builds/67474320.

Julyyq:
@sapics Thank you, 100% becomes 100%px.

Julyyq:
@kangax `cssOnly: true` worked before only for percentage values like 100%, and test case only tests percentage cases. Now it only supports "px" within my fix, without "%" or "em".  But I saw `Sets dimensions (width, height) of this canvas instance. when options.cssOnly flag active you should also supply the unit of measure (px/%/em)` in the api document. How to handle this? Should only support "px" or make some extra coding to support "em" and "%"?

asturur:
Before writing additional code, check if the function "parseUnit" that i think is in fabric.util, is ok for the purpouse.

PR #2316, opened 2015-07-01
IE9 bugfix for loading SVGs from URL that contain <text> nodes

mbrzuzy:


src/shapes/text.class.js
```diff
@@ -1038,8 +1038,25 @@
...
+      if ('firstChild' in element && element.firstChild !== null) {
+        if ('data' in element.firstChild && element.firstChild.data !== null) {
+          textContent = element.firstChild.data;
+        }
+      }
+    } else {
+      textContent = element.textContent;
+    }
+
+    textContent = element.textContent.replace(/^\s+|\s+$|\n+/g, '').replace(/\s+/g, ' ');
```
asturur: you are accessing element.textContent here anyway. So the problem is still here.

asturur:
Also textContent, altough not standard, takes all the content of the element/ node.
I think that firstChild.data takes the text of the first textNode of the element, discarding near tspans.
am i wrong?

src/shapes/text.class.js
mbrzuzy: My mistake, I missed that when porting changes over.

mbrzuzy:
Yes, you're right asturur. Although this fix no longer breaks fabric completely in IE9, tspans should still be supported. When I get the chance I'll amend my commit.

src/shapes/text.class.js
kangax: Btw, we should probably use trim here — https://github.com/kangax/fabric.js/blob/f0fcd770bd489bb6c2322f99462c23a41b7ca6b9/src/util/lang_string.js#L10

asturur:
here is missing somethig:
We have to concatenate all childs not jsut firstChild.

PR #2332, opened 2015-07-13
Fix boundingbox in group

asturur:
still coding.

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -126,6 +126,7 @@
         left: objectLeft - center.x,
         top: objectTop - center.y
       });
+      object.setCoords();
     },
 
     /**
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -471,7 +471,7 @@
 
       for ( ; i < iLen; ++i) {
         o = this._objects[i];
-        o.oCoords || o.setCoords();
+        o.setCoords();
         for (j = 0; j < jLen; j++) {
           prop = props[j];
           aX.push(o.oCoords[prop].x);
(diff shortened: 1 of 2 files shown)


src/shapes/group.class.js
```diff
@@ -383,7 +380,6 @@
       object.hasControls = object.__origHasControls;
       delete object.__origHasControls;
       object.set('active', false);
-      object.setCoords();
```
asturur: there is one setCoords() 4 rows above.

src/shapes/group.class.js
```diff
@@ -475,7 +471,7 @@
 
       for ( ; i < iLen; ++i) {
         o = this._objects[i];
-        o.setCoords();
+        o.oCoords || o.setCoords();
```
asturur: do not recalculate if i already have.

src/shapes/group.class.js
```diff
@@ -123,12 +123,9 @@
           center = this.getCenterPoint();
 
       object.set({
-        originalLeft: objectLeft,
-        originalTop: objectTop,
         left: objectLeft - center.x,
         top: objectTop - center.y
       });
-      object.setCoords();
```
asturur: no reason to make setCoords relative to center of group.
oCoords are used for mouse interactions and object interactions ( collision )

src/shapes/group.class.js
```diff
@@ -123,12 +123,9 @@
           center = this.getCenterPoint();
 
       object.set({
-        originalLeft: objectLeft,
```
asturur: .originalLeft/.originalRight property not used in all library. Why do we store it?

PR #2340, opened 2015-07-15
Fix itextcopypaste

### LINKED ISSUE #1399: Bug in Itext component and solution
image1
image2

hello every one 

I am working on iText in fabric.js

bug is related to Itext style copy/paste with multiple line.
steps to repopulate the bug are 

1. add new text with Itext (in two line) (See attached image1)

2. Change the color of last character in first line and first character in second line.

3. Select and copy both characters whose color has been changed.

4. Paste the characters in second line. The first character retains its color property but the send character in new line does not

solution: 

Now today I finally find the solutions for this bug.. please update this bud in new version on fabric

actually the  problem is if u copy and paste multiline at that time new line style is not copy and paste working correctly

solution :

capture1
capture

1. add if else condition (show attachment image capture.png) on insertCharStyleObject function 

2. create one variable outside the Itext function ( see attachment image capture1.png)

3. then test

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2706833-bug-in-itext-component-and-solution?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
closes #1399 

fix a bug introduce with latest commit in wich style is never used if browser support clipboard.
Moved copiedText and copiedStyle from this.copiedStyle to fabric.copiedStyle to allow copy -> paste of style between different iTexts on canvas ( or different canvases ).
Made a rough check to see if text in clipboard equals to fabric.copiedText so we do not carry on wrong styles around with pasting
Modified insertChar function in a way it inserts one character at time and updating textlines accordingly.
Even if it is not efficent as before, pasting multiple lines or character at the edge of lines looks always correct.
Removed the comment "// TODO: support multiple style insertion if _chars.length > 1" now inserting happens just on _chars.length = 1;

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -522,23 +522,44 @@
     },
 
     /**
-     * Inserts a character where cursor is (replacing selection if one exists)
+     * Inserts characters where cursor is (replacing selection if one exists)
      * @param {String} _chars Characters to insert
+     * @param {Boolean} useCopiedStyle use fabric.copiedTextStyle
      */
     insertChars: function(_chars, useCopiedStyle) {
+      var style;
+
       if (this.selectionEnd - this.selectionStart > 1) {
         this._removeCharsFromTo(this.selectionStart, this.selectionEnd);
         this.setSelectionEnd(this.selectionStart);
       }
+
+      for (var i = 0, len = _chars.length; i < len; i++) {
+        if (useCopiedStyle) {
+          style = fabric.copiedTextStyle[i];
+        }
+        this.insertChar(_chars[i], i < len - 1, style);
+      }
+    },
+
+    /**
+     * Inserts a character where cursor is
+     * @param {String} _char Characters to insert
+     * @param {Boolean} skipUpdate trigger rendering and updates at the end of text insert
+     * @param {Object} styleObject Style to be inserted for the new char
+     */
+    insertChar: function(_char, skipUpdate, styleObject) {
       var isEndOfLine = this.text[this.selectionStart] === '\n';
       this.text = this.text.slice(0, this.selectionStart) +
-        _chars + this.text.slice(this.selectionEnd);
-      this.insertStyleObjects(_chars, isEndOfLine, useCopiedStyle);
-      this.setSelectionStart(this.selectionStart + _chars.length);
+        _char + this.text.slice(this.selectionEnd);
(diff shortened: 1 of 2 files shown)


asturur:
you can try here:

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -102,8 +102,8 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       clipboardData.setData('text', selectedText);
     }
 
-    this.copiedText = selectedText;
-    this.copiedStyles = this.getSelectionStyles(
+    fabric.copiedText = selectedText;
+    fabric.copiedStyle = this.getSelectionStyles(
```
kangax: Good idea! I would just call it `copiedTextStyle` so that it's clear what kind of style these are and in case we add something else in the future.

src/mixins/itext_behavior.mixin.js
```diff
@@ -524,24 +524,40 @@
...
+        var style;
+        for (var i = 0, len = _chars.length; i < len; i++) {
+          if (useCopiedStyle) {
+            style = fabric.copiedStyle[i];
+          }
+          this.insertChars(_chars[i], useCopiedStyle, i < len - 1, style);
+        }
+      }
+      else {
+        var isEndOfLine = this.text[this.selectionStart] === '\n';
```
kangax: It seems we should move this whole block out of here into a separate method. Perhaps, `insertChar`?

src/mixins/itext_behavior.mixin.js
```diff
@@ -522,23 +522,44 @@
     },
 
     /**
-     * Inserts a character where cursor is (replacing selection if one exists)
+     * Inserts characters where cursor is (replacing selection if one exists)
      * @param {String} _chars Characters to insert
+     * @param {Boolean} useCopiedStyle use fabric.copiedStyle
```
kangax: `copiedStyle` &rarr; `copiedTextStyle` :D

asturur: where jscs and lint cannot see....

kangax:
+changelog

Issue #532, opened 2013-04-03
activeGroup removeObject (shiftKey) did not work if perPixelTargetFind = true


Kienz:
<p>In the following jsfiddle you can see the behavior of activeGroup with <code>perPixelTargetFind = true</code> (left canvas) and activeGroup with <code>perPixelTargetFind = false</code> (right canvas).</p>

<p>Select all objects and try to deselect one of the group objects with shiftKey. On the left canvas it doesn't work.<br><a href="http://jsfiddle.net/Kienz/62AXX/">http://jsfiddle.net/Kienz/62AXX/</a></p>

<p>The problem is that absolute pointer coordinates are used for target with relative coordinates (group object). I tried to normalizePointer (_normalizePointer) =&gt; this worked only for objects with positive relative left/top parameters.</p>

<p>The other thing i tried is to use whole group object for check if target is transparent:<br><code>var isTransparent = this._isTargetTransparent(activeGroup &amp;&amp; this.containsPoint(e, activeGroup) ? possibleTargets[j].group : possibleTargets[j], pointer);</code><br><a href="https://github.com/kangax/fabric.js/blob/master/src/canvas.class.js#L749">https://github.com/kangax/fabric.js/blob/master/src/canvas.class.js#L749</a></p>

<p>This works but maybe there is some performance problem for large groups (all objects have to be renderd on cacheContext).</p>

asturur:
Actually i cannot deselect in any grouped object. with shift. not with perPixelTarget, not normal.

asturur:
@kangax @kienz i just tested, i cannot  deselect any selection created active group using shift key.
I really do not know if this is a new issue

kangax:
@asturur @Kienz looks like a regression (it worked before)

asturur:
https://github.com/kangax/fabric.js/commit/9e1a93db5a37d814fd9110e4fdddda21aae37047#diff-9dc132ed9a0787568f773999ba577e2bR813

asturur:
No restoring that line those not change situation

PR #2342, opened 2015-07-15
Fix shiftclick activegroup

### LINKED ISSUE #2170: Shift click can't remove objects from group selection
The kitchensink demo doesn't allow objects to be removed from the selection using shift click even
though there is code in the canvas_grouping.mixin that is apparently intending to support this inside _updateActivteGroup .

I had a look and _shouldClearSelection is now always clearing the entire selection if the target is contained in the group selection. Is removing individual objects a feature that's meant to be supported?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/13112594-shift-click-can-t-remove-objects-from-group-selection?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
helps #532
close #2170

make possible again too shift click to remove object from activeGroup.
Does not fix ( never worked good ) if group is scaled, rotated.
_normalizePointer requires major rework.

This is anyway a small fix that restore the simple functionality.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -284,10 +284,11 @@
           lt;
 
       if (isObjectInGroup) {
-        lt = new fabric.Point(activeGroup.left, activeGroup.top);
-        lt = fabric.util.transformPoint(lt, this.viewportTransform, true);
+        lt = fabric.util.transformPoint(activeGroup.getCenterPoint(), this.viewportTransform, true);
         x -= lt.x;
         y -= lt.y;
+        x /= activeGroup.scaleX;
+        y /= activeGroup.scaleY;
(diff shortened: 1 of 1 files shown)


asturur:
added fix to find target when active group is scaled.
The unlocking object from scaled activegroup triggers others bug that are not purpouse of this PR.

kangax:
+changelog

PR #2344, opened 2015-07-15
Allow to create path with empty path

### LINKED ISSUE #2264: Fabric.loadSVGFromString Fails on Empty Path
Fabric: `1.5.0`
Running Fabric in Node.js `0.12.4`

 When parsing an SVG with a `<path>` that has an empty or no `d` attribute Fabric fails to load the SVG and throws `'path' argument is required`. This behavior is not ideal for our use case as we are dynamically rendering complex SVGs that may have empty paths. Perhaps this could be an option in Fabric to ignore empty paths?

Failure example: `<path fill="#ccc"/>`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/20665551-fabric-loadsvgfromstring-fails-on-empty-path?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
closes #2264

fixed relevant test.

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -78,7 +78,7 @@
       this.setOptions(options);
 
       if (!path) {
-        throw new Error('`path` argument is required');
+        path = [ ];
       }
 
       var fromArray = _toString.call(path) === '[object Array]';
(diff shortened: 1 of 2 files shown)


asturur:
How do i say contrary of "ok" for Qunit?

kangax:
+changelog

PR #2311, opened 2015-06-26
Clenaer PR for regex in parser

asturur:
Cleaner PR with just issue relate changes + test

closes #2292
closes #2287

PR #2345, opened 2015-07-15
Restore parsing svg elements as a parent in caso of use tag plus svg tag

### LINKED ISSUE #2341: SVG rendering incorrectly
I can convert my svg fabric in js.
I find it a problem between version 1.4 and 1.5

Example SVG : 
http://jsfiddle.net/pj7773aa/

Test with Google Chrome

Against it by a regression with version 1.5 in this example fabricjs.

http://fabricjs.com/kitchensink/

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/24374027-svg-rendering-incorrectly?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
fixes parsing regressions from #2341 derived from #2311
closes #2341

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -481,7 +481,7 @@
     if (toBeParsed) {
       return parsedDim;
     }
-    
+
     if (missingViewBox) {
       parsedDim.width = parseUnit(widthAttr);
       parsedDim.height = parseUnit(heightAttr);
@@ -502,7 +502,7 @@
     else {
       parsedDim.width = viewBoxWidth;
       parsedDim.height = viewBoxHeight;
-	  }
+    }
 
     // default is to preserve aspect ratio
     // preserveAspectRatio attribute to be implemented
--- a/src/parser.js
+++ b/src/parser.js
@@ -15,6 +15,11 @@
       parseUnit = fabric.util.parseUnit,
       multiplyTransformMatrices = fabric.util.multiplyTransformMatrices,
 
+      reAllowedSVGTagNames = /^(path|circle|polygon|polyline|ellipse|rect|line|image|text)$/i,
+      reViewBoxTagNames = /^(symbol|image|marker|pattern|view|svg)$/i,
+      reNotAllowedAncestors = /^(?:pattern|defs|symbol|metadata)$/i,
+      reAllowedParents = /^(symbol|g|a|svg)$/i,
+
       attributesMap = {
         cx:                   'left',
         x:                    'left',
@@ -398,10 +403,23 @@
           y = el.getAttribute('y') || 0,
           el2 = elementById(doc, xlink).cloneNode(true),
           currentTrans = (el2.getAttribute('transform') || '') + ' translate(' + x + ', ' + y + ')',
-          parentNode, oldLength = nodelist.length;
+          parentNode, oldLength = nodelist.length, attr, j, attrs, l;
 
-      for (var j = 0, attrs = el.attributes, l = attrs.length; j < l; j++) {
-        var attr = attrs.item(j);
+      applyViewboxTransform(el2);
+      if (/^svg$/i.test(el2.nodeName)) {
+        var el3 = el2.ownerDocument.createElement('g');
+        for (j = 0, attrs = el2.attributes, l = attrs.length; j < l; j++) {
+          attr = attrs.item(j);
+          el3.setAttribute(attr.nodeName, attr.nodeValue);
+        }
+        while (el2.firstChild != null) {
+          el3.appendChild(el2.firstChild);
+        }
+        el2 = el3;
+      }
+
+      for (j = 0, attrs = el.attributes, l = attrs.length; j < l; j++) {
+        attr = attrs.item(j);
         if (attr.nodeName === 'x' || attr.nodeName === 'y' || attr.nodeName === 'xlink:href') {
           continue;
         }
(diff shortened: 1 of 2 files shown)


asturur:
ouch we added test for this :(

asturur:
i checked the specs and svg element is a contanier element like any other else ( g, a , glyph ... )
If we do not want to parse the attributes on top svg element and make descend on contained elements ( but we should see a failing test case to worry about it ) we can still check if svg is the topmost element and decide to do not parse attributes from it.

@dgodard do you have any svg problem with that?

asturur:
no wait we need different solution.
Fifrefox and chrome does not parse attributes from SVG tag.
So the regex is correct, but we need to catch those attributes in some way, because we need them to descend on the child elements if we are referring an svg with the USE tag.

asturur:
an overkill but a working solution.
if we are using SVG as referenced by use tag, we change the svg with a "g" element and we pass all attributes over and all childs under.
@kangax any idea?

asturur:
Need proper inspection because we should also support new viewbox from inner svgs.
I will check it another day.

src/parser.js
```diff
@@ -479,7 +491,7 @@
                   (minX * scaleX) + ' ' +
                   (minY * scaleY) + ') ';
 
-    if (element.tagName === 'svg') {
+    if (/^svg$/i.test(element.tagName)) {
```
kangax: This is so that it catches SVG and svg?

src/parser.js
```diff
@@ -564,7 +577,7 @@
       var elements = descendants.filter(function(el) {
         reViewBoxTagNames.test(el.tagName) && addVBTransform(el, 0, 0);
         return reAllowedSVGTagNames.test(el.tagName) &&
-              !hasAncestorWithNodeName(el, /^(?:pattern|defs|symbol|metadata)$/); // http://www.w3.org/TR/SVG/struct.html#DefsElement
+              !hasAncestorWithNodeName(el, reNotAllowedAncestors); // http://www.w3.org/TR/SVG/struct.html#DefsElement
```
kangax: Good consistency, nice catch!

kangax:
@asturur definitely need some unit tests for this. Could you please add a couple? Thanks!

asturur:
i want to improve some functionality of viewbox, very rare but i prefer to make them work.

i need to create some test svg that we will use also for testing.

do not parse till me signal

asturur:
we should support this:

image

```xml
<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0" y="0" width="980" height="950" >
    <defs>
        <!-- Definition Motif remplissage : un seul par porte -->
        <svg id="Imposte" width="465" height="130" >
            <g fill="#00FF00" stroke="none" >
                <rect x="10" y="10" width="15" height="15"/>
                <rect x="50" y="50" width="15" height="15"/>
                <rect x="50" y="10" width="15" height="15"/>
                <rect x="90" y="10" width="15" height="15"/>
            </g>
        </svg>
        <svg id="Imposte2" width="465" height="130" viewBox="15 15 465 130" >
            <g fill="#0000FF" stroke="none" >
                <rect x="10" y="10" width="15" height="15"/>
                <rect x="50" y="50" width="15" height="15"/>
                <rect x="50" y="10" width="15" height="15"/>
                <rect x="90" y="10" width="15" height="15"/>
            </g>
        </svg>
        <svg id="Imposte3" width="465" height="130" viewBox="0 0 232 65" >
            <g fill="#FF0000" stroke="none" >
                <rect x="10" y="10" width="15" height="15"/>
                <rect x="50" y="50" width="15" height="15"/>
                <rect x="50" y="10" width="15" height="15"/>
                <rect x="90" y="10" width="15" height="15"/>
            </g>
        </svg>
    </defs>
    <g transform="translate(0,10)">
        <use xlink:href="#Imposte" />
        <use xlink:href="#Imposte2" />
        <use xlink:href="#Imposte3" />
        <use xlink:href="#Imposte" transform="translate(100,40)"/>
        <use xlink:href="#Imposte2" transform="translate(100,80)"/>
        <use xlink:href="#Imposte3" transform="translate(100,80)"/>
    </g>
</svg>
```

asturur:
original
image

parsed
image

more than this we cannot go, i explain you why:

Normally SVG defined by width and height, and what is outside it, and what has negative coords, is cutted out.
When an element estabilish a new viewport, what is outside is outside.
In the original test.svg i posted here the blu squares looks rectangular but they are simply offscreen and they are cut away.

I can implement simple viewbox clipping for the MAIN svg element but not yet for the others because we loose group information.

asturur:
Actually we are not applying viewBox scaling factor if width or height are not specified.
I do not find any spec and i do not see coherent behaving between browsers to understand what we should do.

asturur:
ok removed useless parameters from addVBtransform.

i need to find out a simple test for it.
i should extend the current test, adding a viewbox and a transform attribute to check.

asturur:
i lost all my changes???

asturur:
@kangax give a review to code changes, i'll prepare the tests.

asturur:
under node js :

```descendants = doc.selectNodes('//*[name(.)!="svg"]');```

it means that we discard inner svg in descendat
so wi will not rund the ciclying over it:
```
var elements = descendants.filter(function(el) {
        addVBTransform(el);
        return reAllowedSVGTagNames.test(el.tagName) &&
              !hasAncestorWithNodeName(el, reNotAllowedAncestors); 
      });
```

so we are not applying eventual viewboxtransform over them.

Any chance this has been fixed in node or we can change with:
```descendants = doc.selectNodes('//*[name(.)!="x"]');```

asturur:
@kangax You can merge this.
www.deltalink.it/andreab/fabric/a.html

all svgs still parsing good.

PR #2348, opened 2015-07-17
Fix parsing pathgroup dimensions from inner paths

asturur:
now:
image

we cannot get the shape dimensions, declared as 100% we are not able to parse the dimensions correctly.

after fix ( i colored white space to show that the padding is actually a defined shape )
image

### DIFF
--- a/src/shapes/path_group.class.js
+++ b/src/shapes/path_group.class.js
@@ -67,7 +67,7 @@
      */
     parseDimensionsFromPaths: function(options) {
       var points, p, xC = [ ], yC = [ ], path, height, width,
-          m = this.transformMatrix;
+          m;
       for (var j = this.paths.length; j--;) {
         path = this.paths[j];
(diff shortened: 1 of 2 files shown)


kangax:
Nice. Do we have that shape in tests?

asturur:
no but we can do a simple test if i found a pathgroup loading test.

asturur:
I want to clarify something.
PathGroup is essentially a structure made to contain SVGs.
Or at least now is that.

Path inside pathgroup are not supposed and not supported to have own scaleX, scaleY, angle. Only transformMatrix that normally comes from parsing SVG transform(s) property.

I dream a version in wich group and pathgroup become one class and svg are handled as a normal group and can be degrouped on the fly. But we are still far away.

asturur:
@kangax added tests for the parsing dimension and parsing dimensions with transform Matrix. You can merge.

PR #2354, opened 2015-07-21
Fix text glitche

asturur:
What this commit do:

- Resolve text decoration issue for some textAlign values
- Remove a couple of canvas.save and canvas.restore couples
- Solve a regression: when text is justified decoration and backgroundcolor are not.
- do not justify text if a line contains just one word


asturur:
@kangax ready to merge for me.
misc fixes nothing special, getLineWidth got more complicated but is necessary.

kangax:
Great, thanks! Just one minor thing in a comment: "I have to" -> "we have to" or "it has to be" coz it sounds strange otherwise

PR #2359, opened 2015-07-23
some code cleaning

asturur:
Removed use of 
getBoundingRectWidth, getBoundingRectHeight, deprecated since 1.0.4 ( we cannot delete them till 2.0 right? )

Simplified the _getNonTransformedDimensions() method

removed the parameter 'shouldTransform' to getCurrentDimensions

removed some duplicate code

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -1,5 +1,14 @@
 (function() {
 
+  function getCoords(oCoords) {
+    return [
+      new fabric.Point(oCoords.tl.x, oCoords.tl.y),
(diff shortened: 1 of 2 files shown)


src/mixins/object_geometry.mixin.js
```diff
@@ -222,6 +216,7 @@
      * @return {Number} width value
      */
     getWidth: function() {
+      //needs to be changed
```
kangax: why does it need to be changed?

asturur:
need to become this.getTransformedDimensions().x

this.width * this.scaleX does not take in account skew and does not take in account border.

i did not change now because all tests on width and height will change by strokewidth.

PR #2370, opened 2015-07-29
Remove duplicate code and fix justify style issue

asturur:
A little bit of cleaning plus bugfix:

IText justified without style breakdown.

image

Fixes case in wich spaceWidth is infinite.

The justified spaces are not calculated right as of now, but I cannot fix this bug now.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -471,29 +471,28 @@
       top -= this.fontSize * this._fontSizeFraction;
 
       // short-circuit
-      if (this.textAlign !== 'justify') {
+      var lineWidth = this._getLineWidth(ctx, lineIndex);
+      if (this.textAlign !== 'justify' || this.width < lineWidth) {
         this._renderChars(method, ctx, line, left, top, lineIndex);
         return;
       }
 
-      var lineWidth = this._getLineWidth(ctx, lineIndex),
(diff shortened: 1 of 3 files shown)


asturur:
@kangax ready to merge!

src/shapes/text.class.js
```diff
@@ -471,29 +471,28 @@
...
-            widthDiff = totalWidth - wordsWidth,
-            numSpaces = words.length - 1,
-            spaceWidth = widthDiff / numSpaces,
-            leftOffset = 0;
-
-        for (var i = 0, len = words.length; i < len; i++) {
-          this._renderChars(method, ctx, words[i], left + leftOffset, top, lineIndex);
-          leftOffset += ctx.measureText(words[i]).width + spaceWidth;
+      // stretch the line
+      var words = line.split(/\s+/),
```
kangax: Hm, don't we have a separate method for this now?

asturur: if you mean splitting words, no i don t think so.
i can check.
anyway that code needs a review.

measuring width of words without calculating styles is very wrong and produce bad ( rare ) result.

asturur: no we do not have any method for splitting line in words.

PR #2389, opened 2015-08-06
Removed unnecessary call to ctx.measureText

asturur:
just better than nothing.


asturur:
@kangax nothing else to add for me.

PR #2395, opened 2015-08-08
Fix itext paste performance

asturur:
Triggering a hiddenTextarea update twice every char inserted is not a good idea, it slows down everything.

We will trigger it just on last character inserted.

(setSelectionStart and setSelectionEnd they both call this._updateTextarea function that update both cursor position on textarea and the text)

This does not fix the textbox performance issue.


asturur:
@kangax ready to merge.

Issue #2396, opened 2015-08-08
IText Clone object, pass reference to style object instead of cloning it.


rijint:
1) Add an IText with some initial text eg:-sample, delete all characters(sample) using delete key/backspace key  type new content (eg:-mytext).
    
2) Take clone of edited IText
3) Update color of single character the changes reflected to parent also
    change the color by selecting "my" in mytext

asturur:
please post a fiddle, help us checking the bugs faster.

rijint:
![untitled](https://cloud.githubusercontent.com/assets/7856892/9150021/8e3cb4ec-3de2-11e5-804e-ec2198533cad.png)
screenshot

asturur:
Can you show us how you clone the itext? otherways is not possible to help you.

rijint:
var obj = canvas1.getActiveObject();

            obj.clone(function (res) {
                res.set({
                    left: 200,
                    top: 100
                });
                canvas1.add(clone);
            });

rihanrahul:
Here is the fiddle... http://jsfiddle.net/rihanrahul/6zmu0zce/

Please follow the steps mentioned above to reproduce the issue.

asturur:
changing the util.extend function with something like this solve the problem 

I m really not able to identify potential drawbacks here,
need @kangax help here. :)

```
  function extend(destination, source) {
    // JScript DontEnum bug is not taken care of
    for (var property in source) {
      var _source = source[property];
      if( Object.prototype.toString.call(_source) === '[object Array]' ) {
        destination[property] = extend([ ], _source);
      }
      else if (_source && typeof _source === 'object'){
        destination[property] = extend({ }, _source );
      }
      else {
        destination[property] = _source;
      }
    }
    return destination;
  }
```

PR #2397, opened 2015-08-08
Some sort of deep cloning of objects

asturur:
Possible fix for #2396.

Not really sure if it can break something.

### DIFF
--- a/src/util/lang_object.js
+++ b/src/util/lang_object.js
@@ -10,7 +10,16 @@
   function extend(destination, source) {
     // JScript DontEnum bug is not taken care of
     for (var property in source) {
-      destination[property] = source[property];
+      var _source = source[property];
+      if (Object.prototype.toString.call(_source) === '[object Array]' ) {
+        destination[property] = extend([ ], _source);
+      }
+      else if (_source && typeof _source === 'object'){
+        destination[property] = extend({ }, _source);
+      }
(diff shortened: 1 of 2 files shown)


kangax:
Technically, this is a back-incompatible change in behavior so... this could break things in some applications, yes; we'd need to bump version. I'm not sure about changing extend to be deep clone because deep clone comes with its own problems like taking care of circular references. It's also a tad slower, which would make entire Fabric slower. lodash, for example, provides `clone` and a separate `cloneDeep` (https://lodash.com/docs#cloneDeep). One option for us would be to introduce 3rd argument — `isDeep (true | false)`, default being `false`.

kangax:
You could see lodash taking care of circular references and other things like arguments object (which has a [[Class]] of "[object Arguments]") or host objects — https://github.com/lodash/lodash/blob/3.10.1/lodash.src.js#L1817-L1871

asturur:
lets have as a reminder. all itext cloning and objects with transform matrix are cloned with references instead of values.
also polygon's  array of points and path commands.
group's list of objects. all referenced.

at least in toObject method of object.class,
transformatrix should be extend([], this.transformMatrix).

kangax:
> transformatrix should be extend([], this.transformMatrix).

we can just do that for now

PR #2405, opened 2015-08-10
Changes in render ALL

asturur:
### Current rendering stack and possible improvements.

First I will describe the current renderAll situation.
We start with a _renderBackground.
renderBackgrond calls _draw(backgroundImage).
_draw function has this scheme: 
    save context
    apply viewportTransform
    check if the object should be rendered and actually render it
    (this means object !== activeGroup ||  this.preserveObjectStacking)
    restore context
    renderObjectControls ( exit if not !this.active or render the controls )
_renderObjects ( rendering of all objects )
    render all the objects in stack, or exclude the one in activeGroup depending on preserveObjectStacking and checking if active group contains them
Every object is drawn with _draw, so again save context, apply viewportTransform, again check if the object is in activeGroup,
restore context, renderControls.
_renderActiveGroup (rendering objects of activeGroup) with the
_draw procedure already described.
_renderOverlay mirrored from  _renderBackgroundImage
_renderControls if controlsAboveOverlay, make a global call for rendering controls.

- What '_draw' does? wax on and wax off the viewport transform for every object, wrong in my opinion. and check if an object should be rendered... But all object should be rendered now or later.

- What 'shouldRenderObject' does? basically obey to preserveObjectStacking that decide if we draw all the objects together or if we render before the normal objects and later the activeGroup.
Because objects in activeGroup are still present on canvas, in the objects array, so we need to be careful to do not draw them twice.

All the checks are about being in the activeGroup, so useless for activeGroup itself, useless for background and overlay images.

Proposed changes in this PR.
background image and overlay image are rendered with their render method.
ViewportTransform get applied once for renderAll, with one save and restore context.

Objects are divided into two groups immediately, and then we render the groups one after the other without additional checks.

Because of this changes take over _draw functionality, the _draw function disappear.

All the controls in the canvas are drawn with canvas.drawcontrols.

What actually changes in the rendering stack? controls are always on top of the object pile, (not above overaly ) regardless if the object is on top.
But consider that we have this bug as of now, and this new situation would solve it.
The bug is that when controls are drawn during the stack, they are present on canvas when the objects after the 'active index' are rendered, so any eventual global composite operataion get influenced by them

So in my opinion moving the control in the top of the stack is a good move, also code reduction is noticiable and we should see **minor** performance improvements.

Example of bug:
image

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -749,36 +749,6 @@
     },
 
     /**
-     * Given a context, renders an object on that context
-     * @param {CanvasRenderingContext2D} ctx Context to render object on
-     * @param {fabric.Object} object Object to render
-     * @private
-     */
-    _draw: function (ctx, object) {
-      if (!object) {
-        return;
-      }
-
-      ctx.save();
-      var v = this.viewportTransform;
-      ctx.transform(v[0], v[1], v[2], v[3], v[4], v[5]);
-      if (this._shouldRenderObject(object)) {
-        object.render(ctx);
-      }
-      ctx.restore();
-      if (!this.controlsAboveOverlay) {
-        object._renderControls(ctx);
-      }
-    },
-
-    _shouldRenderObject: function(object) {
-      if (!object) {
-        return false;
-      }
-      return (object !== this.getActiveGroup() || !this.preserveObjectStacking);
-    },
-
-    /**
      * @private
      * @param {fabric.Object} obj Object that was added
      */
@@ -855,7 +825,7 @@
      */
     renderAll: function (allOnTop) {
       var canvasToDrawOn = this[(allOnTop === true && this.interactive) ? 'contextTop' : 'contextContainer'],
-          activeGroup = this.getActiveGroup();
+          activeGroup = this.getActiveGroup(), objsToRender = [ ], object, activeGroupObjects = [ ];
 
       if (this.contextTop && this.selection && !this._groupSelector) {
         this.clearContext(this.contextTop);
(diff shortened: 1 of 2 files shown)


asturur:
minification failing also here. Start to think is not my fault.

asturur:
@kangax have a look, i was holding this back since April.

asturur:
any time for some comments?

src/static_canvas.class.js
```diff
@@ -871,16 +841,34 @@
         fabric.util.clipContext(this, canvasToDrawOn);
       }
 
-      this._renderBackground(canvasToDrawOn);
-      this._renderObjects(canvasToDrawOn, activeGroup);
-      this._renderActiveGroup(canvasToDrawOn, activeGroup);
+      if (activeGroup && !this.preserveObjectStacking) {
```
kangax: I'm concerned that we're making code worse here. `_renderBackground`, `_renderObjects`, `_renderActiveGroup` are easy to read and understand because they're all on the same level of abstraction and are named well. You can then look into each of them for implementation details. This is the kind of stuff we should be doing MORE in our code but here I see that we're going the opposite way and introducing a relatively big chunk of some sort of procedure without naming it.

Yes — sometimes we DO need to "unroll" things, such as for performance. Is this why we're doing it here? Is there a way to keep same "style" as before but only change implementation?

src/static_canvas.class.js
```diff
@@ -893,46 +881,32 @@
...
-        for (i = 0, length = this._objects.length; i < length; ++i) {
-          if (this._objects[i] && !activeGroup.contains(this._objects[i])) {
-            this._draw(ctx, this._objects[i]);
-          }
-        }
+    _renderObjects: function(ctx, objects) {
+      for (var i = 0, length = objects.length; i < length; ++i) {
+        objects[i] && objects[i].render(ctx);
       }
     },
```
kangax: This is a nice change! The less conditions the better.

src/static_canvas.class.js
```diff
@@ -871,16 +841,34 @@
...
+      }
+      else {
+        objsToRender = this._objects;
+      }
 
+      //apply viewport transform once for all rendering process
+      canvasToDrawOn.setTransform.apply(canvasToDrawOn, this.viewportTransform);
+      this._renderBackground(canvasToDrawOn);
+      this._renderObjects(canvasToDrawOn, objsToRender);
+      !this.preserveObjectStacking && this._renderObjects(canvasToDrawOn, [activeGroup]);
```
kangax: Also could be `this.preserveObjectStacking || this._renderObjects(canvasToDrawOn, [activeGroup]);`

src/static_canvas.class.js
```diff
@@ -941,20 +915,7 @@
...
-        ctx.fillRect(
-          this.backgroundColor.offsetX || 0,
-          this.backgroundColor.offsetY || 0,
-          this.width,
-          this.height);
-      }
-      if (this.backgroundImage) {
-        this._draw(ctx, this.backgroundImage);
-      }
+      this._renderBackgroundOrOverlay(ctx, 'background');
```
kangax: This (and a method below) is a nice removal of duplicated logic. Good stuff!

src/static_canvas.class.js
```diff
@@ -893,46 +881,32 @@
...
     },
 
     /**
      * @private
      * @param {CanvasRenderingContext2D} ctx Context to render on
-     * @param {fabric.Group} activeGroup
      */
-    _renderActiveGroup: function(ctx, activeGroup) {
+    _renderBackgroundOrOverlay: function(ctx, property) {
+      if (this[property + 'Color']) {
```
kangax: I wonder if we should perhaps save `property + 'Color'` to a var (to avoid constant concatenation) since this is an often-called function, IIRC

asturur: i can make a resubmit solving your comments if you are interested in a
merge. as of now is not mergeable.
Il 12/ott/2015 19:24, "Juriy Zaytsev" <notifications@github.com> ha scritto:

kangax: That would be great! And thanks so much for this effort and your patience ;)

src/static_canvas.class.js
asturur: cool!

PR #2406, opened 2015-08-10
Bug in getElementById polyfill

### LINKED ISSUE #2402: Undefined result in getElementById method(svg parser)
Hi Guys!
While using fabric 1.5.0 on node I faced with rendering issue for some SVGs. After small investigation (after I tried 1.6.0 rc) I've found that there is fix I need in 1.6.0 but it has incorrect value for idAttr property here: https://github.com/kangax/fabric.js/blob/master/src/parser.js#L385
it's empty, probably should use id agument.

asturur:
thanks to @bambinnesss to point it out.

closes #2402

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -382,10 +382,10 @@
     if (el) {
       return el;
     }
-    var node, i, idAttr, nodelist = doc.getElementsByTagName('*');
+    var node, i, nodelist = doc.getElementsByTagName('*');
     for (i = 0; i < nodelist.length; i++) {
       node = nodelist[i];
-      if (idAttr === node.getAttribute('id')) {
+      if (id === node.getAttribute('id')) {
         return node;
       }
     }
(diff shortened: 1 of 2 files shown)


kangax:
Hm, can we add a failing test case as well?

asturur:
in node i think yes.
i should check an existing svg load test, but being it a function, how i can call it in the test? is not sort of 'private' ?

kangax:
We can probably test it through `parseSVGDocument`?

asturur:
yes i need a bit of time.
My linux node box went up with some npm package and now when i run node
test i get message that with js-dom 4.x i need io.js.


2015-08-11 2:50 GMT+02:00 Juriy Zaytsev <notifications@github.com>:

kangax:
yeah, that's why we have it locked on 1.x — https://github.com/kangax/fabric.js/blob/master/package.json#L42

asturur:
@kangax Added failing test case for the function, ready to merge.

PR #2407, opened 2015-08-11
Deep clone transformMatrix, strokeDashArray and styles properties

### LINKED ISSUE #2396: IText Clone object, pass reference to style object instead of cloning it.

asturur:
This will be hardly noticed by anyone.
But still is a bug.

closes #2396.

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -1121,16 +1121,16 @@
      * @return {Object} object representation of an instance
      */
     toObject: function(propertiesToInclude) {
-      var newStyle = { }, i, j, row;
+      var clonedStyles = { }, i, j, row;
       for (i in this.styles) {
         row = this.styles[i];
-        newStyle[i] = { };
+        clonedStyles[i] = { };
         for (j in row) {
-          newStyle[i][j] = clone(row[j]);
+          clonedStyles[i][j] = clone(row[j]);
         }
       }
       return fabric.util.object.extend(this.callSuper('toObject', propertiesToInclude), {
-        styles: newStyle
+        styles: clonedStyles
       });
     }
   });
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -1121,8 +1121,16 @@
      * @return {Object} object representation of an instance
      */
     toObject: function(propertiesToInclude) {
+      var newStyle = { }, i, j, row;
+      for (i in this.styles) {
+        row = this.styles[i];
+        newStyle[i] = { };
+        for (j in row) {
+          newStyle[i][j] = clone(row[j]);
+        }
+      }
       return fabric.util.object.extend(this.callSuper('toObject', propertiesToInclude), {
-        styles: clone(this.styles)
+        styles: newStyle
       });
     }
   });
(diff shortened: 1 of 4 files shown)


asturur:
Added clone for the iText style.
@kangax what do you think?

asturur:
Added failing tests for the modified properties.
ready to merge for me.

src/shapes/itext.class.js
```diff
@@ -1121,8 +1121,16 @@
      * @return {Object} object representation of an instance
      */
     toObject: function(propertiesToInclude) {
+      var newStyle = { }, i, j, row;
```
kangax: `newStyle` &rarr; `clonedStyles`

src/shapes/object.class.js
```diff
@@ -783,7 +783,7 @@
             fill:                     (this.fill && this.fill.toObject) ? this.fill.toObject() : this.fill,
             stroke:                   (this.stroke && this.stroke.toObject) ? this.stroke.toObject() : this.stroke,
             strokeWidth:              toFixed(this.strokeWidth, NUM_FRACTION_DIGITS),
-            strokeDashArray:          this.strokeDashArray,
+            strokeDashArray:          this.strokeDashArray ? extend([ ], this.strokeDashArray) : this.strokeDashArray,
```
kangax: So we're cloning array because... something might modify it somewhere else?

asturur:
if it happens with styles it will happen with array also.

kangax:
But `styles` has objects, array has primitives; can't modify primitives' references

asturur:
I do not understand what  you mean, but the bug is there.
Go to kitchensink, execute this code:

```
// clear canvas
canvas.clear();

// remove currently selected object
canvas.remove(canvas.getActiveObject());

var transform = [2,0,0,2,4,4];
var strokedash = [3, 3];
// add red rectangle
var rect = new fabric.Rect({
  width: 50,
  height: 50,
  left: 50,
  top: 50,
  fill: 'rgb(255,0,0)',
  transformMatrix: transform,
  strokeDashArray: strokedash,
  stroke: 'black',
  strokeWidth: 2
});

// add green, half-transparent circle
canvas.add(rect);
canvas.add(rect.clone().setFill('blue'));
canvas.add(rect.clone().setFill('green'));
```

obtain this
image

then run this code

```
canvas.getActiveObject().transformMatrix[0] = 1;
canvas.getActiveObject().strokeDashArray[1] = 8;
canvas.renderAll();
```

and you get this: ( all object get modified )
image

kangax:
@asturur not sure I understand the bug here... in your last example, you're changing values of transformMatrix and seeing result rendered, no? Isn't this expected behavior?

asturur:
i m changing the transformatrix of one rect, and all three are changing.
because in transformMatrix we have reference of the cloned object instead of a new array.

kangax:
Oh, I didn't realize it's cloned. Of course, you're right.

kangax:
One last thing — we don't want to use extend for arrays since it does for..in iteration and that kind of iteration doesn't just iterate over indices (0..n) but also anything in Array.prototype, etc. Just do transformMatrix.concat() to create a clone of array.

asturur:
@kanga done.

kangax:
+changelog

PR #2410, opened 2015-08-11
add flood fill color

kuafucode:
The flood fill color draws 'background' color or pattern only overlaps with background image instead of the whole canvas like backgroundColor does.
Its very useful when you want to change the color or pattern of the real design area, like a T-shirt background image with transparent background in a canvas.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -62,6 +62,14 @@
     backgroundImage: null,
 
     /**
+     * Flood fill color within background image of canvas instance.
+     * Should be set via {@link fabric.StaticCanvas#setFloodFillColor}.
+     * @type {(String|fabric.Pattern)}
+     * @default
+     */
+    floodFillColor: '',
+
+    /**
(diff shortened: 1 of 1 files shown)


kangax:
Could you please make a small demo to see this in action?

kuafucode:
https://jsfiddle.net/74eLgwuw/

asturur:
I would add the function to image.class.js, to have a flood fill color or pattern for images rather than just for canvas background. Then you can use that image for canvas background.

Or the user can still do it using some image layered by and  (rect + image ) using the specific composite operation.

kuafucode:
In our case, flood fill is only an attribute, but not an item in the design, same as background image. Means we don't want to save it as an object with size, position and etc in the output JSON.

kuafucode:
Oh yea, that's a better idea. Would you like to implement it?

EddieOne:
If I draw a circle, can I flood fill it with this?

PR #2414, opened 2015-08-12
Offset stop

asturur:
get Offset stop as 0 if not presente in the color stop tag.


asturur:
Added failing test.

/cc: @ParOxUs

Issue #2388, opened 2015-08-05
[Textbox] Very slow pasting


baubbelayponge:
I'm trying out the new textbox feature in the 1.6.0-rc.1 version.
Unfortunately, we can't use it in production yet as it is extremely slow when pasting more than one line at a time.

See here :
http://jsfiddle.net/xmfw65qg/

Try editing the textbox and pasting :
"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus auctor blandit mattis. Morbi eu euismod nibh, a auctor sem. Maecenas faucibus dolor eget neque pulvinar lacinia. Cum sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Sed iaculis, neque id pharetra placerat, enim lacus ornare neque, vitae rutrum massa neque vitae massa. Cras quis egestas arcu. Nam nec dolor lacus. Phasellus pretium ac magna eget pharetra."

Using an older version I found floating on another fiddle, I can see it used to be much faster.
Please try here :
http://jsfiddle.net/xmfw65qg/1/

Any ideas?
Thanks!

asturur:
I think the problem is that i changed the paste text from block pasting to character pasting, also for textbox we need to recalculate style map and line split every character.
1) or the process is so slow 
2) or there is some bug in the code that make useless calculation.

Before it was not  working properly, style pasting was never working as expected.

asturur:
![image](https://cloud.githubusercontent.com/assets/1194048/9104076/21ff4cb8-3c07-11e5-8486-8c99ea941fba.png)

added some profiling.
anyway something to be solved.

asturur:
This is test case with iText, not slow as textbox but a bit slow anyway.

i do not feel like making a switch "if there is no style, paste a the text all in single operation", because style object get very ruined after pasting with multiple newlines and i do not know what happens on inserting style later.

iText profiling 
![image](https://cloud.githubusercontent.com/assets/1194048/9105457/42b9c954-3c18-11e5-9e98-5fe43fd25a08.png)

asturur:
```diff
    insertChar: function(_char, skipUpdate, styleObject) {
      var isEndOfLine = this.text[this.selectionStart] === '\n';
      this.text = this.text.slice(0, this.selectionStart) +
        _char + this.text.slice(this.selectionEnd);
      this._textLines = this._splitTextIntoLines();
      this.insertStyleObjects(_char, isEndOfLine, styleObject);
[...]

PR #2423, opened 2015-08-17
Textbox optimization on wraptext

asturur:
I may be wrong, but as of now we are measuring the width of a word with measureText multiple times.

We measure once for sure and the we measure it as many times as many words are in front of it, plus one.

in a sentence like: "the lazy dog jumped over the frog" with a textbox large enough to keep all sentence:

First we measure all line to see if it fits, like a shortcut, that really is not, because the functions called from this._measureText have a per char based logic.
So we are iterating over all char. ( caching the widths or not, this is not the point )

If the line does not fit we proceed as follow:

We measure "the",
Then we add to the line.
Then we measure lazy, we check if line is empty (is not) we measure "the lazy"
Then we add to the line.
Then we measure dog, we check if line is empty (is not) we measure "the lazy dog"
Then we add to the line.
Then we measure jumped, we check if line is empty (is not) we measure "the lazy dog jumped"
and so on....

Because measureText is a performance killer, we should avoid this logic.

This PR is an attempt to fix that behavior.

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -645,16 +645,18 @@
         charHeight = this.fontSize;
       }
 
-      shouldStroke = (shouldStroke || this.stroke) && method === 'strokeText';
-      shouldFill = (shouldFill || this.fill) && method === 'fillText';
+      shouldStroke = shouldStroke || method === 'strokeText';
+      shouldFill = shouldFill || method === 'fillText';
 
       decl && ctx.save();
 
       charWidth = this._applyCharStylesGetWidth(ctx, _char, lineIndex, i, decl || {});
       textDecoration = textDecoration || this.textDecoration;
 
-      if (shouldFill || shouldStroke) {
-        ctx[method](_char, left, top);
+      shouldFill && ctx.fillText(_char, left, top);
+      shouldStroke && ctx.strokeText(_char, left, top);
+
+      if (textDecoration || textDecoration !== '') {
         offset = this._fontSizeFraction * lineHeight / this.lineHeight;
         this._renderCharDecoration(ctx, textDecoration, left, top, offset, charWidth, charHeight);
       }
@@ -814,14 +816,16 @@
      * @param {Object} [decl]
      */
     _applyCharStylesGetWidth: function(ctx, _char, lineIndex, charIndex, decl) {
-      var styleDeclaration = decl || this._getStyleDeclaration(lineIndex, charIndex, true);
+      var charDecl = this._getStyleDeclaration(lineIndex, charIndex),
+          styleDeclaration = decl || clone(charDecl), width;
 
       this._applyFontStyles(styleDeclaration);
 
       var cacheProp = this._getCacheProp(_char, styleDeclaration);
 
-      // short-circuit if no styles
-      if (this.isEmptyStyles() && this._charWidthsCache[cacheProp] && this.caching) {
+      // short-circuit if no styles for this char
+      // global style from object is always applyed and handled by save and restore
+      if (!charDecl && this._charWidthsCache[cacheProp] && this.caching) {
         return this._charWidthsCache[cacheProp];
       }
 
@@ -844,12 +848,9 @@
       ctx.font = this._getFontDeclaration.call(styleDeclaration);
       this._setShadow.call(styleDeclaration, ctx);
 
-      if (!this.caching) {
-        return ctx.measureText(_char).width;
-      }
-
-      if (!this._charWidthsCache[cacheProp]) {
-        this._charWidthsCache[cacheProp] = ctx.measureText(_char).width;
+      if (!this.caching || !this._charWidthsCache[cacheProp]) {
+        width = ctx.measureText(_char).width;
+        this.caching && (this._charWidthsCache[cacheProp] = width);
       }
 
       return this._charWidthsCache[cacheProp];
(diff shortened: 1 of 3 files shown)


asturur:
/cc: @inssein @jafferhaider @kangax
What do you think?

src/shapes/textbox.class.js
```diff
@@ -259,45 +259,33 @@
      * to.
      */
     _wrapLine: function(ctx, text, lineIndex) {
-      var maxWidth  = this.width,
-          lineWidth = this._measureText(ctx, text, lineIndex, 0);
```
asturur: There is no reason to measure the length of the line when measuring single words has more or less same cost, and we are likely doing it anyway.

asturur:
helps also #2388,
does not solve it completely, but speed improvement is noticeable.

asturur:
@kangax ready to merge for me.

asturur:
Need to check for first word bigger than line as first word, and empty lines.
But this should be done without over use of this,_measureText.

src/shapes/itext.class.js
```diff
@@ -989,15 +989,6 @@
      * @private
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
-    _getHeightOfCharAt: function(ctx, lineIndex, charIndex) {
```
asturur: unused method

src/shapes/itext.class.js
```diff
@@ -980,7 +980,7 @@
      * @private
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
-    _getHeightOfChar: function(ctx, _char, lineIndex, charIndex) {
```
asturur: unused variable

src/shapes/itext.class.js
```diff
@@ -636,12 +636,12 @@
           offset = this._fontSizeFraction * lineHeight / this.lineHeight;
 
       if (decl) {
-        var shouldStroke = decl.stroke || this.stroke,
-            shouldFill = decl.fill || this.fill;
+        var shouldStroke = (decl.stroke || this.stroke) && method === 'strokeText',
+            shouldFill = (decl.fill || this.fill) && method === 'fillText';
```
asturur: We where both stroking and filling when calling method = 'strokeText' or 'fillText'

src/shapes/itext.class.js
```diff
@@ -631,38 +631,37 @@
...
+
+      decl && ctx.save();
+
+      charWidth = this._applyCharStylesGetWidth(ctx, _char, lineIndex, i, decl || {});
+      textDecoration = textDecoration || this.textDecoration;
+
+      if (shouldFill || shouldStroke) {
+        ctx[method](_char, left, top);
+        offset = this._fontSizeFraction * lineHeight / this.lineHeight;
+        this._renderCharDecoration(ctx, textDecoration, left, top, offset, charWidth, charHeight);
```
asturur: No nice way to avoid renderChar two times.
Actually we are rendering two times, one in renderStroke and one in renderFill.

inssein:
I gave it a quick review this morning, and it looks good, however I won't be able to test it until Wednesday. Thanks for the hard work @asturur .

asturur:
the built library is online here:
It contains also some iText code cleaning.

asturur:
@kangax ready to merge for me.

src/shapes/itext.class.js
```diff
@@ -831,14 +816,16 @@
...
 
       this._applyFontStyles(styleDeclaration);
 
       var cacheProp = this._getCacheProp(_char, styleDeclaration);
 
-      // short-circuit if no styles
-      if (this.isEmptyStyles() && this._charWidthsCache[cacheProp] && this.caching) {
+      // short-circuit if no styles for this char
+      // global style from object is always applyed and handled by save and restore
+      if (!charDecl && this._charWidthsCache[cacheProp] && this.caching) {
```
asturur: Shortcircuit was not used if some style was present in the Itext Object. We need to apply style only if there is a special style object for the char.

src/shapes/textbox.class.js
```diff
@@ -234,17 +234,8 @@
...
-          width += this._applyCharStylesGetWidth(ctx, text[i], lineIndex, i, decl);
-          ctx.restore();
-        }
-        else {
-          // @note: we intentionally pass in an empty style declaration, because if we pass in nothing, it will
-          // retry fetching style declaration
-          width += this._applyCharStylesGetWidth(ctx, text[i], lineIndex, i, {});
-        }
+      for (var i = 0, len = text.length; i < len; i++) {
+        width += this._getWidthOfChar(ctx, text[i], lineIndex, i + charOffset);
```
asturur: Use standard method for char width calculation.

asturur:
added some more tweaks. textBox pasting is nearly fast as before now.

jafferhaider:
I just played around with Textbox in the latest build, it is completely broken for me. What's going on?

asturur:
i made a mistake please.check latest open pr.
one character typo, don t even know how it went in

jafferhaider:
Heh ok, no worries. I'll try again when that is merged. Thanks for your hard work!

asturur:
is just one char, swap 0 with i, and you are good to go.
Check also the proposed word-wrapping if you have time, it is posted in the
related issue.

2015-08-21 13:50 GMT+02:00 Jaffer Haider <notifications@github.com>:

fahadnabbasi:
excellent improvement, I was trying to find a solution for quite some time now. Thanks @asturur

arsen2010:
@asturur , @kangax  Thank you for your hard work, nice and sharp.

PR #2427, opened 2015-08-18
Add support for caching parsed SVG files instead of making an XHR request for each

AdamMerrifield:
I realized that the `svgCache` methods were all no operation. I need svg caching for another project and thought I'd add a pull request of my additions.

I debated on not having a `setRequesting` method and just adding an optional parameter to the `has` method for `addToRequesting` which would default to `false`.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -617,38 +617,95 @@
    * Used for caching SVG documents (loaded via `fabric.Canvas#loadSVGFromURL`)
    * @namespace
    */
-  var svgCache = {
+  var svgCache = (function() {
+    var cache = {},
+        requesting = {};
 
-    /**
-     * @param {String} name
-     * @param {Function} callback
-     */
-    has: function (name, callback) {
-      callback(false);
-    },
+    return {
 
-    get: function () {
-      /* NOOP */
-    },
+      /**
+       * @param {String} name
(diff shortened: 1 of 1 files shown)


src/parser.js
```diff
@@ -617,38 +617,95 @@
...
+      if (result instanceof fabric.constructor) {
+        eachObjCallback(result);
+      }
     });
 
-    return ({ objects: objects, options: options });
+    function addBuiltObject(builtObj, objectIndex) {
+      builtObjects[objectIndex] = builtObj;
+
+      var builtObjectsLength = Object.keys(builtObjects).length;
```
AdamMerrifield: Also, I couldn't think of a better way to maintain the order of the objects, so I'm using `Object.keys` to tell when all of them are done. If we decided to keep it this way we may need to add an `Object.keys` polyfill into `src/util/lang_object.js`.

asturur:
Long time passed, i m sorry to not have review this before.
Today i decided to remove this feature completely from fabric since the caching logic is something to be left to developer of application and not to library.
This is the main reason.
Another one, even if very minor, is that i can imagine a server script that return a random svg from a fixed url. In that case we would need to give the dev a way to disable cache also.

PR #2434, opened 2015-08-21
Bugfix

asturur:
sorry i do not know how it went in this.


asturur:
For the second change, i realized that something was wrong with
stroke='' , fill='', _skipFillStrokeCheck = true and some styled letters.
The rendering was black instead of transparent.

PR #2445, opened 2015-08-24
attempt to word break for textbox

asturur:
image

It looks working good.
Please test are welcome.

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -699,9 +699,9 @@
 
       var decorationWeight = charHeight / 15,
           positions = {
-            'underline': top + charHeight / 10,
+            underline: top + charHeight / 10,
             'line-through': top - charHeight * (this._fontSizeFraction + this._fontSizeMult - 1) + decorationWeight,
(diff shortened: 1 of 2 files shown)


asturur:
Someone should get an idea about how to handle text like
"   Lorem ipsum dolor   "
wit trailing spaces or consecutive spaces.

asturur:
@jafferhaider @inssein

inssein:
@asturur will be checking it out tomorrow, good work :)

onassar:
Dang I'm not set up with support for separate classes (eg. `textbox.class.js`)
Right now, just have everything included in one `fabric.js` file
Will try and separate things so I can test properly

asturur:
check how to download and build a branch

jafferhaider:
@asturur works great! 

A few comments:
1. I'd suggest that breakWords be `true` by default since that is more inline which how I imagine people will use Textbox ... to limit text within a box.
2. In other textbox implementations (e.g. the textarea tag in a browser), when you start typing a long word somewhere in the middle of a line, the textbox wraps the long word onto the next line in an attempt to make it fit all in one line. I think the Textbox class should do the same.

PR #2453, opened 2015-08-30
correct value of e.button

### LINKED ISSUE #2420: __onMouseDown expects middle button instead of left
In canvas_events.mixin.js, in __onMouseDown, there is the following line:
`var isLeftClick  = 'which' in e ? e.which === 1 : e.button === 1;`
According to https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button, e.button being 1 usually means middle click. This probably hasn't been noticed because most browsers support (the apparently non-standard) e.which.

asturur:
as per https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button

closes #2420


asturur:
What is is problem with extra quotes?

PR #2479, opened 2015-09-15
After rendering a gesture set the origin back to its initial value

### LINKED ISSUE #2478: Gestures change origin to center,center
I am currently using the canvas.toJSON() function to export my canvas to a server side rendering system and I have noticed that after performing a gesture (zoom or rotate) the origin is always changes from top,left to center,center.

I believe this is caused by the _setOriginToCenter() function call within __gesturesRenderer() in the canvas_gestures mixin.

From my limited testing it seems that the origin can be reset by calling _setCenterToOrigin() just prior to the call to renderAll().

nibynool:
After completing gesture rendering reset the origin values to prevent unexpected behaviors.

Fixes #2478 

(This is my first PR, so please be gentle if I got anything wrong)

### DIFF
--- a/src/mixins/canvas_gestures.mixin.js
+++ b/src/mixins/canvas_gestures.mixin.js
@@ -62,6 +62,8 @@
         this._rotateObjectByAngle(self.rotation, e);
       }
 
+      this._setCenterToOrigin(t.target);
+
       this.renderAll();
       t.action = 'drag';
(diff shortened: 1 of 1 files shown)


asturur:
wait to merge, it will conflict we skew pr.

kangax:
Thanks!

PR #2482, opened 2015-09-17
Skew

asturur:
Skew functionality with no merge conflict.
Still have to put in all the tests.


asturur:
cmon....
"Expected points but got points"  what he expect me to do?
Is point some kind of reserved word?
```
Expected points but got points, at src/util/misc.js :
   105 |    /**
   106 |     * Returns coordinates of points's bounding rectangle (left, top, width, height)
   107 |     * @param {Array} points, 4 points array
------------------------------^
   108 |     * @return {Object} Object with left, top, width, height properties
   109 |     */
```

asturur:
@kangax ready to merge!

src/util/misc.js
```diff
@@ -457,21 +482,44 @@
...
     },
 
     /**
+     * Decompone standard 2x2 matrix into transform componentes
+     * @static
+     * @memberOf fabric.util
+     * @param  {Array} a transformMatrix
+     * @return {Object} Components of transform
+     */
+    qrDecompone: function(a) {
```
kangax: What's decompone?

asturur: Is that function to normalize the control box
Mathematically is calle QR decomposition, and it's done on linear transform matrices.
Is not directly necessary for skewing, but is necessary to decompone the transformation, also is an handy tool for expert users that need to do something with transformMatrix.

We will use it later, it was already in the source.

kangax: Hm, I don't think there's such word as decompone so let's change to decompose?

asturur: mmm it is italian. i will put it in english.

PR #2487, opened 2015-09-19
Fix object.to data url

### LINKED ISSUE #2474: .toDataURL({ multiplier: 2 }) on single object goes wrong. The image data contains both original object and multiplied object
Using toDataURL on an object like a text, textBox or iText  and using the multiply option is causing a problem. You get an image with the text at original scale and the text at multiplied scale. It's both in the same image.
The text at original scale should not be there. 
This bug is introduced since version 1.5.0 Older versions don't have this problem

  { 
var text = new fabric.Text("A very fancy italic text",{
        top: 10,
        left: 10,
        width: 100,
        height: 100,
        textAlign: 'center',
        fontFamily: 'Impact',
        stroke: '#c3bfbf',
        strokeWidth: 3
    });

    canvas.add(textbox);

    $("#btn-save").click(function () {
        var src = textbox.toDataURL({ multiplier: 2 });
        var img = document.getElementById("image");
        img.src = src;
    });

}
image

Thanks

asturur:
If toDataUrl is method of statiCanvas is better to render everything on the lower canvas and remove checks for upperCanvas and contextTop.

The real bug is somewhere else:

```
    renderAll: function (allOnTop) {
      var canvasToDrawOn = this[(allOnTop === true && this.interactive) ? 'contextTop' : 'contextContainer'],
          activeGroup = this.getActiveGroup();

      if (this.contextTop && this.selection && !this._groupSelector) {
      }

      if (!allOnTop) {
        this.clearContext(canvasToDrawOn);
      }
```

we were calling renderAll with allOnTop = true on staticCanvas.
So canvasToDrawOn  was contextContainer.
But later we were cleaning the context just on allOnTop = false.
So both copies of object, scaled and unscaled were on the canvas.
The unscaled one coming from canvas.add() in the object.toDataURL.

really this allOnTop i think is something from the past, now we are not rendering anything on contextTop other than controls sometimes or selectors. i see no reasons to draw all on top.
But not time to check it now.

closes #2474

### DIFF
--- a/src/mixins/canvas_dataurl_exporter.mixin.js
+++ b/src/mixins/canvas_dataurl_exporter.mixin.js
@@ -57,7 +57,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
    */
   __toDataURL: function(format, quality, cropping) {
 
-    this.renderAll(true);
+    this.renderAll();
 
     var canvasEl = this.upperCanvasEl || this.lowerCanvasEl,
         croppedCanvasEl = this.__getCroppedCanvas(canvasEl, cropping);
@@ -70,9 +70,6 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
     var data = (fabric.StaticCanvas.supports('toDataURLWithQuality'))
               ? (croppedCanvasEl || canvasEl).toDataURL('image/' + format, quality)
               : (croppedCanvasEl || canvasEl).toDataURL('image/' + format);
-
-    this.contextTop && this.clearContext(this.contextTop);
-    this.renderAll();
 
     if (croppedCanvasEl) {
       croppedCanvasEl = null;
@@ -119,7 +116,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         activeObject = this.getActiveObject(),
         activeGroup = this.getActiveGroup(),
 
-        ctx = this.contextTop || this.contextContainer;
+        ctx = this.contextContainer;
 
     if (multiplier > 1) {
       this.setWidth(scaledWidth).setHeight(scaledHeight);
@@ -152,8 +149,6 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
     else if (activeObject && this.deactivateAll) {
       this.deactivateAll();
     }
-
-    this.renderAll(true);
 
     var data = this.__toDataURL(format, quality, cropping);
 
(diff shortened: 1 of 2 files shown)


asturur:
I add, after this change nothing in the library is calling renderAll(true)

kangax:
> I add, after this change nothing in the library is calling renderAll(true)

Perhaps it's time to remove it then?

asturur:
just waiting for you to ask.

kangax:
:) let's do it! the less old, unnecessary cruft the better

asturur:
@kangax ready to merge

kangax:
Gotta update changelog with back-incompat note

taveras:
It looks like this change has affected rendering of the `fabric.Canvas.toDataURL` method. The data URL returned by the method is always an empty image.

[This line](https://github.com/kangax/fabric.js/blob/69d0071536e3bbaaa14d9a236ecb6f130180e475/src/mixins/canvas_dataurl_exporter.mixin.js#L62) short-circuits the canvas element used to render to the upper canvas, if the upper canvas exists.

It does exist, but it is empty, so is renders the empty canvas as the fabric.Canvas's data URL.

Previously, `fabric.Canvas.renderAll` was called with the `allOnTop` flag set to true. This forced all of the content to be rendered onto the upper canvas, before the native `toDataURL` method was called.

asturur:
@taveras thank you for noticing. i resubmitted the fix.

taveras:
@asturur Awesome! I also had another strange observation:

When calling `fabric.Canvas.toDataURL`, it appears to always output an image that is two-times the size of the canvas, but cropped to the dimensions of the canvas.

If I call `fabric.Canvas.setZoom(0.5)` just before I call `fabric.Canvas.toDataURL`, it renders the canvas as expected.

I realize that this may require a new issue be created to be understood and resolved, and I'll do so if you feel it necessary.

asturur:
What do you mean, if it is cropped, how do you know how large was the image before?
Can you show the bug somehow?

taveras:
I only meant cropped in the sense that it was clipped to the normal canvas size.
Funnily enough, though, I believe the first issue may have even been with my own local build.
When I grab [this](https://cdn.rawgit.com/kangax/fabric.js/5d752d2b4d30d163d3650ddb3b5a627e43c55dd8/dist/fabric.min.js) fabric.min.js, the original issue I noted about a transparent Data URL image is gone.

It is [this commit](https://github.com/asturur/fabric.js/commit/520e8e61796ba4d563398d0efc61490451a2c44c) that seems to make the image in the outputted Data URL be the canvas scaled 2x, but clipped to the original dimensions. 

Here is an example of the effect of your most recent commit:

[Before](http://www.dev.bytedept.com/fabric-bug/old.html)
[After](http://www.dev.bytedept.com/fabric-bug/new.html)

again, the original reasoning for this change, was that I was getting a transparent image when using `fabric.Canvas.toDataURL`, but it turns out that my local build of fabric.js was different than the build in the 'dist' folder on GitHub.

PR #2502, opened 2015-09-27
Fix itext onInput with selected text

### LINKED ISSUE #2451: IText/Textbox - The first character is ignored when you type after selecting all text
Steps for recreating:
1. Type some text into IText or Textbox
2. Use keyboard shortcut to select all (Cmd+A)
3. Type some more text. The first character that you type will not appear in the text. Text will start updating when the 2nd character is entered.

This is a recent issue, haven't come across it in the older builds. Any ideas @asturur?

asturur:
Remove cancelOnInput logic, a normal e.preventDefault + stopPropagation is enough.
Take in account already selected text when deciding how much text we are inserting during input event

closes  #2451

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "canvas": "1.2.x",
-    "jsdom": "1.1.x",
+    "jsdom": "7.x.x",
     "xmldom": "0.1.x"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   },
   "dependencies": {
     "canvas": "1.2.x",
-    "jsdom": "7.x.x",
+    "jsdom": "6.x.x",
     "xmldom": "0.1.x"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


asturur:
unified removeStyleObject for itext and textbox

asturur:
@kangax those are failing for node4. Should i do something?

kangax:
@asturur yeah, Fabric is failing on Node 4.x for some reason. 

@linusU any idea what this is? (https://travis-ci.org/kangax/fabric.js/jobs/82422214#L584) I thought this was supposed to be fixed in 1.2.9 but apparently something is still wrong? Would appreciate if you could give some pointers!

asturur:
should we not update also jsdom? i read sonewhere that to use node4 you beed jsdom higher than 6.3

asturur:
anyway this PR would be ready like it is if you do not see anything strange

LinusU:
@kangax Hmm, that is very strange, here is a recent build from us that uses the same version of node which works. https://travis-ci.org/Automattic/node-canvas/jobs/82633575

It's strange that all libraries gives the same error as well...

asturur:
We are using a old version of jsdom. may it be the problem?

PR #2510, opened 2015-09-29
Update

dgodard:


asturur:
id attribute and regex expression are already fixed.
Did you want to contribute something else?

PR #2529, opened 2015-10-07
Fix to `animateColor` when transitioning to/from opacity of 0

mattkahl:
Addresses failure case:

``` js
animatedObject = canvas.getActiveObject();
animateColorOnChange = function(value){
  animatedObject.stroke = value;
  canvas.renderAll();
};
fabric.util.animateColor('green', 'rgb(0,0,0,0)', 1000, {onChange: animateColorOnChange});
```

This would result in a slowly fading green stroke that ended up as an opaque and not-quite black stroke (final `stroke` value of `rgba(0,1,0,1.1102230246251565e-16)`).

### DIFF
--- a/src/util/animate_color.js
+++ b/src/util/animate_color.js
@@ -4,11 +4,15 @@
   //         http://www.bitstorm.org/jquery/color-animation/jquery.animate-colors.js
   function calculateColor(begin, end, pos) {
     var color = 'rgba('
-        + parseInt((begin[0] + pos * (end[0] - begin[0])), 10) + ','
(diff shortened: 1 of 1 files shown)


kangax:
Could you please also add a unit test for this?

mattkahl:
Sure, I'll make some time to do that.

Sorry about the initial omission. I noticed their weren't any existing tests to modify and wasn't motivated enough to create some at the time.

kangax:
ping @mattkahl

mattkahl:
Failed to make a unit test for this. Closing because it's probably been fixed in the last four years.

asturur:
i m not 100% sure it has been fixed.

PR #2538, opened 2015-10-08
fixing the "fabric.Object.getSvgFilter is not a function"

sadiqevani:
Hi,

This pull request fixes an issue on a really specific case. When i wanted to export the whole canvas into an SVG if i had interactive text elements in my canvas and changed them, when i tried to convert it into an SVG context, it said "fabric.Object.getSvgFilter is not a function", i think this issue comes since the "this" object is not bound to the actual object, but to the overall system.

Case step by step:

Add a iText object
Edit iText with some text (after creating a dummy).
Export to SVG.

Probably there is a better way to do this? @kangax up to you.

### DIFF
--- a/src/mixins/object.svg_export.js
+++ b/src/mixins/object.svg_export.js
@@ -23,7 +23,7 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
         opacity = typeof this.opacity !== 'undefined' ? this.opacity : '1',
 
         visibility = this.visible ? '' : ' visibility: hidden;',
-        filter = this.getSvgFilter();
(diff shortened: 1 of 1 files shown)


asturur:
This error has been fixed somewhere:
https://github.com/kangax/fabric.js/blob/master/src/mixins/itext.svg_export.js#L93

When we export text elements we manually tell the object what is getSvgFilter().

Can you try downloading and building latest source? i think the issue is already solved.

sadiqevani:
@asturur sweet :) will look into it.

Thank you

asturur:
did you have time to check if the problem is solved? if yes, please close the PR.

PR #2545, opened 2015-10-12
RenderAll() changes

asturur:
### Current rendering stack and possible improvements.

First I will describe the current renderAll situation.
We start with a _renderBackground.
renderBackgrond calls _draw(backgroundImage).
_draw function has this scheme: 
    save context
    apply viewportTransform
    check if the object should be rendered and actually render it
    (this means object !== activeGroup ||  this.preserveObjectStacking)
    restore context
    renderObjectControls ( exit if not !this.active or render the controls )
_renderObjects ( rendering of all objects )
    render all the objects in stack, or exclude the one in activeGroup depending on preserveObjectStacking and checking if active group contains them
Every object is drawn with _draw, so again save context, apply viewportTransform, again check if the object is in activeGroup,
restore context, renderControls.
_renderActiveGroup (rendering objects of activeGroup) with the
_draw procedure already described.
_renderOverlay mirrored from  _renderBackgroundImage
_renderControls if controlsAboveOverlay, make a global call for rendering controls.

- What '_draw' does? wax on and wax off the viewport transform for every object, wrong in my opinion. and check if an object should be rendered... But all object should be rendered now or later.

- What 'shouldRenderObject' does? basically obey to preserveObjectStacking that decide if we draw all the objects together or if we render before the normal objects and later the activeGroup.
Because objects in activeGroup are still present on canvas, in the objects array, so we need to be careful to do not draw them twice.

All the checks are about being in the activeGroup, so useless for activeGroup itself, useless for background and overlay images.

Proposed changes in this PR.
background image and overlay image are rendered with their render method.
ViewportTransform get applied once for renderAll, with one save and restore context.

Objects are divided into two groups immediately, and then we render the groups one after the other without additional checks.

Because of this changes take over _draw functionality, the _draw function disappear.

All the controls in the canvas are drawn with canvas.drawcontrols.

What actually changes in the rendering stack? controls are always on top of the object pile, (not above overaly ) regardless if the object is on top.
But consider that we have this bug as of now, and this new situation would solve it.
The bug is that when controls are drawn during the stack, they are present on canvas when the objects after the 'active index' are rendered, so any eventual global composite operataion get influenced by them

So in my opinion moving the control in the top of the stack is a good move, also code reduction is noticiable and we should see **minor** performance improvements.

Example of bug:
image

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -749,36 +749,6 @@
     },
 
     /**
-     * Given a context, renders an object on that context
-     * @param {CanvasRenderingContext2D} ctx Context to render object on
-     * @param {fabric.Object} object Object to render
-     * @private
-     */
-    _draw: function (ctx, object) {
-      if (!object) {
-        return;
-      }
-
-      ctx.save();
-      var v = this.viewportTransform;
-      ctx.transform(v[0], v[1], v[2], v[3], v[4], v[5]);
-      if (this._shouldRenderObject(object)) {
-        object.render(ctx);
-      }
-      ctx.restore();
-      if (!this.controlsAboveOverlay) {
-        object._renderControls(ctx);
-      }
-    },
-
-    _shouldRenderObject: function(object) {
-      if (!object) {
-        return false;
-      }
-      return (object !== this.getActiveGroup() || !this.preserveObjectStacking);
-    },
-
-    /**
      * @private
      * @param {fabric.Object} obj Object that was added
      */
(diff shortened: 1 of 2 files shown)


src/static_canvas.class.js
```diff
@@ -868,10 +864,17 @@
       if (this.clipTo) {
         fabric.util.clipContext(this, canvasToDrawOn);
       }
+      
+      objsToRender = this._chooseObjectsToRender();
```
asturur: Following your advice i grouped the big block of code and gave it a clear ( i hope ) name.

src/static_canvas.class.js
```diff
@@ -891,46 +894,35 @@
...
 
     /**
      * @private
      * @param {CanvasRenderingContext2D} ctx Context to render on
-     * @param {fabric.Group} activeGroup
+     * @param {string} property 'background' or 'overlay'
      */
-    _renderActiveGroup: function(ctx, activeGroup) {
+    _renderBackgroundOrOverlay: function(ctx, property) {
+      var object = this[property + 'Color'];
```
asturur: Added a temp variable do reduce accessing object properties.

src/static_canvas.class.js
```diff
@@ -869,9 +865,16 @@
         fabric.util.clipContext(this, canvasToDrawOn);
       }
 
+      objsToRender = this._chooseObjectsToRender();
```
asturur: Following your advice i grouped the big block of code and gave it a clear ( i hope ) name.

PR #2547, opened 2015-10-13
upgrade jsdom

inssein:
Just upgraded jsdom to 6.5.x. Tested against node v4.2.0 and canvas 1.2.10. Not sure what this means for previous versions of node.

### DIFF
--- a/src/mixins/canvas_dataurl_exporter.mixin.js
+++ b/src/mixins/canvas_dataurl_exporter.mixin.js
@@ -85,20 +85,21 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
 
     var croppedCanvasEl,
         croppedCtx,
+        croppedNodeCanvasEl,
         shouldCrop = 'left' in cropping ||
                      'top' in cropping ||
                      'width' in cropping ||
                      'height' in cropping;
 
     if (shouldCrop) {
-
       croppedCanvasEl = fabric.util.createCanvasElement();
+      croppedNodeCanvasEl = fabric.isLikelyNode ? canvasEl._nodeCanvas : canvasEl;
       croppedCtx = croppedCanvasEl.getContext('2d');
 
       croppedCanvasEl.width = cropping.width || this.width;
       croppedCanvasEl.height = cropping.height || this.height;
 
-      croppedCtx.drawImage(canvasEl, -cropping.left || 0, -cropping.top || 0);
+      croppedCtx.drawImage(croppedNodeCanvasEl, -cropping.left || 0, -cropping.top || 0);
     }
 
     return croppedCanvasEl;
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ node_js:
   - "0.12"
   - "0.11"
   - "0.10"
+addons:
+  apt:
+    packages:
+      - g++-4.8
 script: 'npm run build && npm run test && npm run lint && npm run lint_tests'
 before_install:
   - sudo apt-get install -qq libgif-dev libpng-dev libjpeg8-dev libpango1.0-dev libcairo2-dev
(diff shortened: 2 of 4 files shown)


inssein:
Not sure why its failing yet. I just built it with node v4.2.1 and it seems to work fine. I did have to move to GCC 4.8 when I initially upgraded to 4.2.0, but I don't think that is the problem.

inssein:
@kangax the problem is GCC:

```This version of node/NAN/v8 requires a C++11 compiler```

**Comment by inssein** [comment]
Seems like all the tests are failing on ```Image or Canvas expected```, which at this point seems jsdom related. I think the canvas element jsdom returns is not the same.

**Comment by inssein** [comment]
@kangax should I be pursuing this further? The path that I am currently headed down doesn't allow for a backwards compatible fabric.js due to the changes mentioned above.

There could be some potentially backwards compatible ways of moving forward, such as trying to use a version of jsdom prior to 6.3. (I haven't confirmed that jsdom 6.3 works on node > 4)

**Comment by kangax** [comment]
@inssein I'm not entirely sure (need to look into it more when I get a chance) but fwiw we've done some back-incompat changes and are technically on the verge of 2.0 (since we follow semver now). Perhaps we can use that opportunity to update JSDom.

**Comment by inssein** [comment]
@kangax sounds good. I just made node 4+ pass on travis. The fix was a lot easier than I thought.

Not sure how we can support older versions of node yet.

**Comment by inssein** [comment]
@kangax is it okay if i rebuild the library and commit the build files in this PR? How else do most people make PR and use it as a part of a node app?

**Comment by kangax** [comment]
> How else do most people make PR and use it as a part of a node app?
IIRC, you can refer to a commit in package.json

**Comment by inssein** [comment]
@CombatCode I can try this again on Monday, but it seems like you missed a step perhaps? The error  is pretty much saying "fabric.window" wasn't defined properly, which is not possible if jsdom is properly installed.

**Comment by CombatCode** [comment]
There is a [README](https://github.com/tmpvar/jsdom/blob/6.5.1/README.md) for jsdom in ver. 6.5.1.
Fabric.window is based on ``fabric.document.parentwindow`` but they are not using anymore ``parentwindow`` in favor of ``defaultView``

**Comment by inssein** [comment]
If you look at my commit, that's exactly what I'm using.

**Comment by CombatCode** [comment]
ahh shit, that's true, I have lost a bit time ;) but after installing package via npm, the ``dist/`` folder is in use with old code https://github.com/kangax/fabric.js/blob/master/dist/fabric.js#L23

Anyway, thanks, your commit was very helpfull for me

File: src/mixins/canvas_dataurl_exporter.mixin.js [hunk 1]
```diff
@@ -85,20 +85,21 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
...
+        croppedNodeCanvasEl,
         shouldCrop = 'left' in cropping ||
                      'top' in cropping ||
                      'width' in cropping ||
                      'height' in cropping;
 
     if (shouldCrop) {
-
       croppedCanvasEl = fabric.util.createCanvasElement();
+      croppedNodeCanvasEl = fabric.isLikelyNode ? canvasEl._nodeCanvas : canvasEl;
```
asturur: this looks like a big change.
or am i wrong? looks like that canvasEl is no more the same between browser and nodecanvas

inssein: It's not a big change. Canvas was never the same between browser and node. The previous version of jsdom just returned the canvas when you did ```fabric.document.createElement('canvas')```, but in the new version of jsdom, they create a wrapper around the canvas returned, but have it accessible via ```_nodeCanvas```. (https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/level2/html.js#L1472)

asturur: i was thinking about other places where we use the canvasEl directly. But i didn't found yet.
Other places we use just the context. Maybe in (StaticCanvas / Canvas).getElement() we should make the same check and return.

Also we use 
in other places like:
https://github.com/kangax/fabric.js/blob/master/dist/fabric.js#L780
https://github.com/kangax/fabric.js/blob/master/dist/fabric.js#L6246
https://github.com/kangax/fabric.js/blob/master/dist/fabric.js#L8052

Some of them returns the element directly.

asturur:
replaced by #2872

PR #2559, opened 2015-10-16
Reset viewport transform at the end of render all

### LINKED ISSUE #2558: Problems when canvas zoom < 1 with new renderAll
Hi,
I was trying out the new renderAll by @asturur, it looks awesome and it makes more sense than the preciding one, but I found a problem when I set a zoom < 1, to the canvas. I create a [fiddle](http://jsfiddle.net/kyh0Lwyt/1/), as you can see it looks like the objects get re-drawn on every mouse:move, which is correct, but they never get removed. The behaving is quite strange, so I suggest you look at the fiddle to have a better understanding of the problem, just try dragging the objects around....

asturur:
closes #2558 

At the end of renderAll we were not resetting the viewportTransform to neutral.
That is affecting next cycle of renderAll before the first setTransformMatrix.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -887,7 +887,7 @@
       }
 
       this.fire('after:render');
-
+      canvasToDrawOn.setTransform.apply(canvasToDrawOn, [1, 0, 0, 1, 0, 0]);
       return this;
     },
 


kangax:
Any way to make a test for it?

asturur:
checking pixel values?

asturur:
Please just merge it, i'm gonna make a test later i promise.
I wanna fix this and merge the activegroup i finally fixed.

PR #2565, opened 2015-10-20
Add two helper functions to help with changing attributes on grouped objects

VictorHagsand:
Thought this might be useful since its currently a bit tricky to make changes to a single or all selected objects and get the recalculated correctly. Its possible to add and remove objects and get everything recalculated, so this is a compliment to that.

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -188,6 +188,49 @@
     },
 
     /**
+     * Sets a property to an object in a group; Then recalculates group's dimension, position.
+     * @param {Object} object
+     * @param {String} prop Property to set
+     * @param {String} value Value to set
+     * @return {fabric.Group} thisArg
+     * @chainable
+     */
+    setObjectWithUpdate: function(object, prop, value) {
+        this._moveFlippedObject(object);
+        this._restoreObjectsState();
+        # since _restoreObjectsState set objects inactive
+        this.forEachObject(this._setObjectActive, this);
+        
+        object.set(prop, value);
+        this._calcBounds();
+        this._updateObjectsCoords();
+        
+        return this;
+    },
+    
+    /**
+     * Sets a property to all objects in a group; Then recalculates group's dimension, position.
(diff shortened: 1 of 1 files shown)


src/shapes/group.class.js
```diff
@@ -188,6 +188,49 @@
...
+    
+    /**
+     * Sets a property to all objects in a group; Then recalculates group's dimension, position.
+     * @param {String} prop Property to set
+     * @param {String} value Value to set
+     * @return {fabric.Group} thisArg
+     * @chainable
+     */
+    setObjectsWithUpdate: function(prop, value) {
+        this._moveFlippedObject(object);
```
kangax: Which `object`?

asturur:
I have some doubts on this.
Other than _moveFlippedObject is deprecated, i do not see why someone should recalculate the object when changing "fill" for example.
We have .realizeTransform that is chainable and someone could just call group.objects[x].setColor().realizeTransform() for example.

kangax:
I think this kind of API is useful but the implementation could use some tweaking. @asturur, like you say, we should use `realizeTransform` and possibly _only_ when changing certain set of properties (for perf. reasons)

VictorHagsand:
Can you point me at realizeTransform() in the docs or is that a undocumented function as many others? The docs are in a pretty bad state sadly and its really hard to find stuff : /

asturur:
Yes docs are in bad state.

Please check 
https://github.com/asturur/fabric.js/blob/fix-group-transform/src/shapes/group.class.js#L327

moveFlippedObject is phasing out, realizeTransform is now the old "resetObjectPosition".
But we had it before as the combination of _moveFlippedObject and resetObjectPosition.

VictorHagsand:
Thank you : )

kangax:
Andrea, could you please add this to docs.

asturur:
When pr is merged and is stable, i will check even docs. Now is not the moment. We are changing lot of stuff.

PR #2568, opened 2015-10-21
Update shadow.class.js

### LINKED ISSUE #2566: Stroke shadow is lost when cloned
A shape with a stroke shadow will lose the shadow when cloned:

http://jsfiddle.net/19xvxjb1/2/

image

asturur:
add affectStroke to toObject.

@kangax do you think that we should modify also _toString method?

we need test for this, before you say that :)

closes #2566

### DIFF
--- a/src/shadow.class.js
+++ b/src/shadow.class.js
@@ -151,21 +151,13 @@
         };
       }
       var obj = { }, proto = fabric.Shadow.prototype;
-      if (this.color !== proto.color) {
-        obj.color = this.color;
-      }
-      if (this.blur !== proto.blur) {
-        obj.blur = this.blur;
-      }
-      if (this.offsetX !== proto.offsetX) {
-        obj.offsetX = this.offsetX;
-      }
-      if (this.offsetY !== proto.offsetY) {
-        obj.offsetY = this.offsetY;
-      }
-      if (this.affectStroke !== proto.affectStroke) {
-        obj.affectStroke = this.affectStroke;
-      }
+
+      ['color', 'blur', 'offsetX', 'offsetY', 'affectStroke'].forEach(function(prop) {
+        if (this[prop] !== proto[prop]) {
+          obj[prop] = this[prop];
+        }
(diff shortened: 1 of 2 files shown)


src/shadow.class.js
```diff
@@ -162,6 +163,9 @@
       if (this.offsetY !== proto.offsetY) {
         obj.offsetY = this.offsetY;
       }
+      if (this.affectStroke !== proto.affectStroke) {
+        obj.affectStroke = this.affectStroke;
+      }
       return obj;
```
kangax: I think it's time to replace it with this :)

```js
['color', 'blur', 'offsetX', 'offsetY', 'affectStroke'].forEach(function(prop) {
  if (this[prop] !== proto[prop]) {
    obj[prop] = this[prop];
  }
}, this);
```

asturur: i do not understand why just this class has this feature.

kangax: That's a very good question! What happens if we remove this chunk completely?

asturur: That we have to remove some tests as well, and that we have to wait for 2.0 i think.
I would merge it and make a checklist for 2.0 to remove some old things.

kangax:
@asturur actually let's wait with this until we merge promises branch; otherwise we'll have tons of conflicts

asturur:
Basically promises will conflict with any open PR.
I would add them anyway, just i have to retype all when you will merge.
Let's make a label "approved" that you stick on open PR that cannot be merged soon.
So when there is to do some retyping to solve conflicts i know i'm doing on the final version of code.

What do you think?

kangax:
Good idea, please add that label to this one!

asturur:
I think you are the only one that can create labels.

kangax:
Hm, you don't have "New label" button here? https://github.com/kangax/fabric.js/labels

asturur:
@kangax merge this for 1.6.0 release

PR #2578, opened 2015-10-23
Pass along the dimensions

inssein:
The two _scaleObjectBy's expect the dimensions to be passed along, but the only caller doesn't pass in along. This PR is based on assumption.


asturur:
this is probably right, i had same problem in texbox override, but i missed this.

i will check later if it works and report.

asturur:
the function that has to receive dimension, expects transformed dimensions.

PR #2580, opened 2015-10-24
fix bug with retina scaling

asturur:
Using setTransform during renderAll cycle kills any other transformations added to canvas before rendereAll ( retina scaling for example ).

Using save and restore method should be safer.


asturur:
@kangax ready to merge

Issue #1520, opened 2014-07-23
[SVG IMPORT] Shapes with transformMatrix have misplaced and/or stretched controls.


saeschdivara:
I tried to load an svg which was created by fabric.js (toSVG). Loading was possible but the selection area of the canvas objects are totally wrong (as can be seen in the screenshot)
![screen shot 2014-07-23 at 10 13 56](https://cloud.githubusercontent.com/assets/4942808/3670275/a38b6500-1241-11e4-94c8-22283e101047.png)

The fabric.js version is 1.4.9

In the fiddle you cannot even select the objects: http://jsfiddle.net/saskyrar/L8cmw/2/

Here is the svg:

``` xml
<?xml version="1.0" encoding="UTF-8" standalone="no" ?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="800" height="700" style="background-color: rgb(255, 255, 255)" xml:space="preserve">
    <desc>Created with Fabric.js 1.4.5</desc>
    <defs></defs>

    <circle
        cx="0"
        cy="0"
        r="75"
        style="stroke: none; stroke-width: 1; stroke-dasharray: ; stroke-linecap: butt; stroke-linejoin: miter; stroke-miterlimit: 10; fill: blue; opacity: 1;"
        transform="translate(521 203)" id="2b0a303e-857e-9431-e2c9-d67a59759422"/>

    <rect
        x="-150"
        y="-100"
        rx="0"
        ry="0"
        width="300"
        height="200"
        style="stroke: none; stroke-width: 1; stroke-dasharray: ; stroke-linecap: butt; stroke-linejoin: miter; stroke-miterlimit: 10; fill: blue; opacity: 1;"
        transform="translate(185 229)" id="09e6bb6e-8970-57b2-bec4-bcb77787efc5"/>
</svg>
```
##

defue:
I've created a simple demonstration of the bug using the latest FabricJS version (1.4.13):
http://jsfiddle.net/4793dtua/

In the example if you click on red rectangle you will see that selection box is misplaced.

Hope it helps to debug the issue.

asturur:
The bug is very well know, i recently change some function with this topic in mind, so i hope i can draw the correct bounding box.

[...]

PR #2582, opened 2015-10-25
[WIP] Import ungrouped svg

asturur:
early pr to comment code or other.
#1520


src/shapes/line.class.js
```diff
@@ -235,9 +224,6 @@
       var markup = this._createBaseSVGMarkup(),
           p = { x1: this.x1, x2: this.x2, y1: this.y1, y2: this.y2 };
 
-      if (!(this.group && this.group.type === 'path-group')) {
-        p = this.calcLinePoints();
```
asturur: note: this should be kept

asturur:
it looks like i have to do something to gradientTransform.
something that allows me to unwind the change of rendering the objects around center.

asturur:
image

It woorks!!!

oh yeahhhhhh

asturur:
Some micro differencies to be solved.

New Method

image


Old Method

image

that will be hard to spot.
Accuracy test needed... :(

asturur:
i improved it a bit:
this is scale 50X

image

opened in inkscape:

image

It looks correct.
I moved the pathOffset of half strokeWidth that was missing.
But:

this is a particular of the window cleaning:

current rendering:
image

old rendering of 1.5 and 1.6rc1
image

inkscape:
image

asturur:
it looks like there are some math rounding error here and there in all the process.
I do not really know how to solve it.
On idea could be keeping pathgroup rendering and  transformMatrices for groupe svg, and use the new code when ungrouping them
I'm really confused as of now.

asturur:
@kangax i need you to read this text with attention:
At the end i opted for less radical changes.
Pathgroups are still there.

What happens?
obejcts get parsed from SVG.
TransformMatrix get created, get multiplied for the translation the object has from the canvas 0,0. And we keep this transformMatrix.
Object get left, top, angle, skew, scale modified to respect the new transformMatrix.

At this point object can be alone in canvas or grouped in pathgroups.
They can be also grouped in normal groups with small aproximation error ( very small ) that we will investigate in future.

During rendering ( any rendering ) transformMatrix is now an alternative to angle, skew, scale, left and top. And is no more an addition to normal transformation.

at the end we have ungrouped svg with very minor position displacement but good controls, and grouped svg in pathgroup.

waiting for your comments

deenfoxx:
I cloned your branch and build a new dist/fabric.js with all modules.

It works a LOT BETTER than it did. All elements are contained within their bounding boxes, but alignment is still way off. Do I need to do anything different than before?

    fabric.loadSVGFromURL("/images/ticket-sample.svg", function(objects, options) {
        canvas.add.apply(canvas, objects);
        canvas.renderAll();
    });

import-ungrouped
original-svg

Referring to this... https://github.com/kangax/fabric.js/issues/1281#issuecomment-157150540

asturur:
i will update the code. i left text behind for now, as you may notice text need an additional translation of width /2 to the left.


remember also tspan are not supported yet, so multiline text will not work.

deenfoxx:
Okay. Do you know if there is something I could ask our graphic developers to do (or not do) in Illustrator/Photoshop/InDesign to avoid the non-functional parts... aside from no multi-line? I am going to have to programatically convert all the SVG text elements into fabric iText anyway, so at that point I can introduce multiline, I am assuming.

Yes, I noticed the 45 degree text but it seemed as weird to me as the rest of the misalignment. I'm glad you know what's going on.

asturur:
can you give me that svg? so i will have something to test "TEXT" with.
The problem of newlines is that tspan are really for everything out there, colors, font-size, angle, rotation.

There is no multiline in svg.
What you have is an element that says 

```xml
<tspan x="0" y="0">LINE 1</tspan>
<tspan x="0" y="20">LINE 2</tspan>
```
So you have 2 separate text objects, one under the other.

From this simple situation you have to add color and font-size change support, that we can support with iText.

asturur:
Also avoid:

ClipPaths
Patterns
Gliphs
masks

We do not import them yet.

deenfoxx:
I emailed the SVG since GitHut wouldn't let me attach it. The plain text SVG source without the embedded image is here... https://github.com/kangax/fabric.js/issues/1281#issuecomment-157150540

Thank you.

deenfoxx:
Guess what? I figured it out... I set the object.originX and object.originY to 'center' before adding the text elements to canvas and they are now aligned correctly (except for the rotated 90 degrees showing up as 45 degrees issue). Did this while I was converting to IText. I also found out I had to remove the transformMatrix in order for it to be correctly editable... otherwise the edit cursor and selection were below and to the right.

    fabric.loadSVGFromURL("/images/ticket-sample.svg", function(objects, options) {
        canvas.renderAll();
    }, function(item, object) {
        if (object.type == 'text') {
            var itext = new fabric.IText('text');
            for (var attr in object) {
                if (object.hasOwnProperty(attr) && attr != 'type' && attr != 'transformMatrix') {
                    console.log(attr + ': '+ object[attr]);
                    itext[attr] = object[attr];
                }
            }
            object = itext;
            object.originX = 'center';
            object.originY = 'center';
        }
        canvas.add(object);
        if (object.type == 'image') {
            canvas.sendToBack(object);
        }
    });

I had to use the sendToBack(object) on my image because it was not adding the objects in the correct order. I know this will not work for other potential SVG imports, but it was a quick fix while I was working on the text.

import-ungrouped-better

deenfoxx:
I have found a bug in this WIP that was inadvertently created. I tested and it does not occur in 1.5 or 1.6-rc1 builds.

To duplicate, create any two objects (I tested with Text, IText and Rect). Group at least two objects and rotate 360 degrees. You will see the selection boxes misalign when they reach some individual threshold. If you exit the group selection at this time, it will snap the objects into the place of the misaligned selection boxes and will be a bit screwy if you try selecting them again (until they reach the normal angle range that doesn't exhibit the bug).

bug-in-fabric-fix

    var text1 = new fabric.Text('Hello', {
        fill: 'purple',
        borderColor: 'yellow',
        cornerColor: 'yellow',
        borderWidth: 3,
        transparentCorners: false
    });
    var text2 = new fabric.IText('World', {
        fill: 'blue',
        borderColor: 'white',
        cornerColor: 'white',
        borderWidth: 3,
        transparentCorners: false
    });
    var rect = new fabric.Rect({
        fill: 'red',
        width: 100,
        height: 50,
        borderColor: 'green',
        cornerColor: 'green',
        borderWidth: 3,
        transparentCorners: false

    });
    canvas.add(text1, rect, text2);
    canvas.renderAll();

asturur:
i think we already parse scientific notation.
If grouped the svg looks fine, it means we parse it. we may have some rounding error when decomponing. that yes.

i have to finish some tasks and then i will improvw this pr, it will stay WIP until is not accurate

asturur:
@deenfoxx  i introduced a bug in the  code for rotation.
i was adding 180 degrees to angle BEFORE changing radians in degree. So definitively a mess.

I changed a little bit code ,did you try the latest?

deenfoxx:
I will grab the latest soon. Been on holiday. :-) Thank you.

asturur:
solved with #3969

PR #2615, opened 2015-11-10
Update IText hidden text area - IOS 8 Safari bug 

mrahla:
Fix keyboard and page scrolling in IOS 8 safari


asturur:
$(document).scrollTop ?
is not that jQuery?

mrahla:
Fixed it to native

PR #2624, opened 2015-11-12
change default strokewidth to 0  for images

### LINKED ISSUE #2590: Change default image strokeWidth to 0
I propose this change to avoid fuzzyness of images in standard setups.

Also i would add a comment in the property of fabric.Image.prototype.strokeWidth that using even numbers produce better results.

Update of tests is necessary of course.

@kangax, put approved label or trash the issue.

asturur:
As discussed in #2590 this gives better rendering results.

closes #2590

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -65,6 +65,14 @@
     meetOrSlice: 'meet',
 
     /**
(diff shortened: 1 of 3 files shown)


asturur:
merge this as pre 1.6.0 release

PR #2625, opened 2015-11-13
Fix multi touch scaling speed

### LINKED ISSUE #2586: 1.6.0-rc1 multitouch not working with gesture
Multitouch support seems to be broken with 1.6.0-rc1 (works with 1.5)

Tested a build with NODE build.js modules=ALL based on the current version from https://github.com/kangax/fabric.js (downloaded as zip 27 oktober 2015)

Everything seems fine except multitouch support is not working.

Is this a know issue?
Any help with this is really appriciated. 
Thanks in advance.
----
/* build: `node build.js modules=ALL minifier=uglifyjs` */

/*! Fabric.js Copyright 2008-2015, Printio (Juriy Zaytsev, Maxim Chernyak) */
 

var fabric = fabric || { version: "1.6.0-rc.1" };

if (typeof exports !== 'undefined') {

  exports.fabric = fabric;

}

### LINKED ISSUE #2597: Multitouch zooming is buggy on Android
There are problems appear when I try to resize object with 2 fingers. An object goes smaller and I cannot do anything with it.

Open [this page](http://fabricjs.com/touch-events/) on Android browser (not Chrome) and try to zoom with 
2 fingers.

I have this problem on Android 5, don't know does it appear on android 4 or 6.

Let me show you some screenshots. They just reproduce what do I see on my Android.

The original dog: 

And that's what I get after I touch an object with 2 fingers:

Any thoughts? As quick and dirty solution can you tell me how to turn off multitouch?

asturur:
we need to use transformed dimensions to include effect of skew. but we need to divide by current scale and use just the initial scale plus the scaling factor of gesture.

Closes #2586
closes #2597

### DIFF
--- a/src/mixins/canvas_gestures.mixin.js
+++ b/src/mixins/canvas_gestures.mixin.js
@@ -136,7 +136,7 @@
       var constraintPosition = target.translateToOriginPoint(target.getCenterPoint(), t.originX, t.originY),
           dim = target._getTransformedDimensions();
 
-      this._setObjectScale(new fabric.Point(dim.x * s, dim.y * s),
+      this._setObjectScale(new fabric.Point(t.scaleX * dim.x * s / target.scaleX, t.scaleY * dim.y * s / target.scaleY),
         t, lockScalingX, lockScalingY, null, target.get('lockScalingFlip'), dim);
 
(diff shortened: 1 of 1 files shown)


asturur:
it works good.
@kangax mergw this as pre 1.6.0 release. othewise multi touch scaling is broken.

PR #2627, opened 2015-11-16
Clean polygon class

asturur:
Remove _applyPointOffset that i introduced while ago and is totally bad.

Mainly because it change the original point array and makes difficult to add additional points to the polygon.

Need to fix SVG method.

### DIFF
--- a/src/shapes/polygon.class.js
+++ b/src/shapes/polygon.class.js
@@ -66,6 +66,8 @@
       if (!('left' in options)) {
         this.left = this.minX;
       }
+      this.offsetX = this.minX + this.width / 2;
+      this.offsetY = this.minY + this.height / 2;
     },
 
     /**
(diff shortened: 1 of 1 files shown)


asturur:
Also there was a never triggered bug:
Points where changed in first render. But if we initialized the polygon and then executed the toSVG without going for a render, this would output a wrong polygon.

asturur:
i noticed two unfixed shadow test. my fault.

kangax:
> i noticed two unfixed shadow test. my fault.
so... no merge yet?

asturur:
i uploaded the fix in some pr. maybe we just need to relaunch the tests.
Il 03/dic/2015 19:35, "Juriy Zaytsev" <notifications@github.com> ha scritto:

asturur:
we have 3 failing tests:

this from opacity pr

Module: fabric.Group Test: set objects' value should be set properly at Object.<anonymous> (/home/travis/build/kangax/fabric.js/test/unit/group.js:130:23) Actual value: 1 Expected value: 0.12345 

those failing in nodejs but not in browser.

Module: fabric.Group Test: test group transformMatrix at Object.<anonymous> (/home/travis/build/kangax/fabric.js/test/unit/group.js:524:11) Actual value: true Expected value: false 


Module: fabric.Group Test: test group transformMatrix at Object.<anonymous> (/home/travis/build/kangax/fabric.js/test/unit/group.js:528:11) Actual value: true Expected value: false

asturur:
any thought on this?

PR #2634, opened 2015-11-18
General code clean up for convolute filter

asturur:
we do not need to create a temporary canvas to call its createImageData, when normal canvas we are filtering can provide same method.
Removed 2 unused variables ( w and sw, h and sh where copies of each other and used in read only)


asturur:
this is not solving the slowness issue. but is still nice.
@kangax you can merge if you like it too.

PR #2642, opened 2015-11-22
Svg viewport transform fixes

asturur:
Currently the property svgViewportTransformation does not produce correct output if set to false.
The option should ignore viewportTransform but it produce also a smaller svg.

This PR also separate the width and heigth from the viewbox and centralize the viewport handling in the svgHeader where the viewbox parameter is.

current BUG on fabricjs.com kitchensink viewport set to [1.5, 0, 0, 2, -100, 100]
image

svg output is correct.

Now we set canvas.svgViewportTransformation = false;
and svg output is:
image

instead of correct output that should be this:
image

This because the function is dividing width and heigth of the svg to zoom back to 1.

Now also is possible to export svg desired size:
```canvas.toSVG({width: '500mm', height: '500mm'});```
```canvas.toSVG({width: 300, height: 1500});```

this will produce svg of desired dimension without changing aspect ratio.


asturur:
fixed also shadow tests.

asturur:
@kangax ready to merge

PR #2645, opened 2015-11-23
Fix importing paths with `rotate` transforms with anchor coordinates.

rodovich:
This adds support for the `rotate(angle, x, y)` syntax for rotating around a given point
(http://www.w3.org/TR/SVG/coords.html#TransformAttribute). Tested by the fixture in https://github.com/kangax/fabricjs.com/pull/101.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -130,12 +130,16 @@
    */
   fabric.parseTransformAttribute = (function() {
     function rotateMatrix(matrix, args) {
-      var angle = args[0];
+      var angle = args[0],
+          x = (args.length === 3) ? args[1] : 0,
+          y = (args.length === 3) ? args[2] : 0;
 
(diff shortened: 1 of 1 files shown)


src/parser.js
```diff
@@ -130,12 +130,16 @@
    */
   fabric.parseTransformAttribute = (function() {
     function rotateMatrix(matrix, args) {
-      var angle = args[0];
+      var angle = args[0],
+          x = (args.length === 3) ? args[1] : 0,
+          y = (args.length === 3) ? args[2] : 0;
 
       matrix[0] = Math.cos(angle);
```
asturur: matrix[0] = Math.cos(angle);
       matrix[1] = Math.sin(angle);
       matrix[2] = -matrix[1];
       matrix[3] = matrix[0];

non sense optimization?

PR #2649, opened 2015-11-25
fix reverse chars  when oject is horizontally flipped

Helloooooooooo:
fix reverse chars  when object is horizontally flipped


kangax:
Can we also make a test for this?

asturur:
i think we should not flip characters at all.
editing when flipping is very difficult and flippes chars feel so innatural.

personally when i flip something i wanna flip it like a picture and that's it.

asturur:
i tried this fix and i'm still more convinced we should not flip chars when flipping text as an object.

PR #2671, opened 2015-12-07
Fix shadow to SVG

asturur:
modify the offset and filter box to take in account of rotation.


asturur:
it was not ready, the shadow test was not finished yet. i ll update it now

PR #2672, opened 2015-12-08
Shadow to svg fixed test

asturur:
Using NUM_FRACTION_DIGITS for consistency and to fix the test.
(avoid 8.1412543e-16)

### DIFF
--- a/src/shadow.class.js
+++ b/src/shadow.class.js
@@ -3,7 +3,8 @@
   'use strict';
 
   var fabric = global.fabric || (global.fabric = { }),
-      toFixed = fabric.util.toFixed;
+      toFixed = fabric.util.toFixed,
+      NUM_FRACTION_DIGIT = fabric.Object.NUM_FRACTION_DIGITS;
 
   if (fabric.Shadow) {
     fabric.warn('fabric.Shadow is already defined.');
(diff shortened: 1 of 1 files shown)


src/shadow.class.js
```diff
@@ -111,23 +111,24 @@
...
+      var fBoxX = 40, fBoxY = 40, NUM_FRACTION_DIGITS = fabric.Object.NUM_FRACTION_DIGITS,
           offset = fabric.util.rotateVector({x: this.offsetX, y: this.offsetY},
             fabric.util.degreesToRadians(-object.angle));
 
       if (object.width && object.height) {
         //http://www.w3.org/TR/SVG/filters.html#FilterEffectsRegion
         // we add some extra space to filter box to contain the blur ( 20 )
-        fBoxX = toFixed((Math.abs(offset.x) + this.blur) / object.width, 2) * 100 + 20;
-        fBoxY = toFixed((Math.abs(offset.y) + this.blur) / object.height, 2) * 100 + 20;
+        fBoxX = toFixed((Math.abs(offset.x) + this.blur) / object.width, NUM_FRACTION_DIGITS) * 100 + 20;
```
kangax: Let's store 20 in a BLUR or something like that, for clarity

asturur:
@kangax done
still i cannot figure out thos 2 failing tests.
They fail on node only, on browsers pass.
Nothing has changed dipendency wise from when they start to fail.

PR #2689, opened 2015-12-16
use atan2 instead of atan and manual checks

asturur:
I forgot to commit those changes from long ago

### DIFF
--- a/src/util/misc.js
+++ b/src/util/misc.js
@@ -524,16 +524,13 @@
      * @return {Object} Components of transform
      */
     qrDecompose: function(a) {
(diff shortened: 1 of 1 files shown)


kangax:
Does this fix something or purely optimization?

kangax:
@asturur does this fix something or purely optimization?

asturur:
is just optimization. so i do not have to check for angle. i have to check
if it can improve precision of  then ungrouped load.
Il 24/dic/2015 04:15, "Juriy Zaytsev" <notifications@github.com> ha scritto:

PR #2703, opened 2015-12-22
Fill opacity and stroke opacity

asturur:
Attempt to fix and add tests


asturur:
tests were not going... i clean them later.

PR #2708, opened 2015-12-24
Add active instance cleaning to dispose.

### LINKED ISSUE #2701: `dispose` should clear `activeInstance`
If it doesn't (current state), the last instance is always leaked even after disposed and removed.

asturur:
closes #2701.

Still i do not get what this activeInstace does.
It is a reference to last initialized canvas.

But what is its purpouse?

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -47,8 +47,6 @@
       this._initStatic(el, options);
       this._initInteractive();
       this._createCacheCanvas();
-
-      fabric.Canvas.activeInstance = this;
     },
 
     /**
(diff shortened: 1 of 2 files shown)


kangax:
Yep, last initialized one. Now that I think about it, it perhaps does seem unnecessary...

asturur:
To me it looks unnecessary.
Maybe it was a try to do not create a global variable to access the canvas?

I would just trash it.

kangax:
I agree. I'm all for removing unnecessary crap.

asturur:
@kangax you can merge

PR #2711, opened 2015-12-26
Fix test fill-opacity and stroke-opacity

asturur:


asturur:
@kangax this is ok merge it.
i will work on the other failing test but i need more time.

PR #2726, opened 2016-01-07
Update Qunit to 0.7.x latest and instabul 0.4.x

asturur:
if we are interested in testing with latest versions of packages.


asturur:
Trying to get rid of 
[execsync v1.0.2] Attempting to compile native extensions.
[execSync v1.0.2] Native code compile failed!!

using sync-exec instead of execSync.
@kangax What do you think?

PR #2733, opened 2016-01-12
Fix scrolling on focus on IText on iOS 8/9 Safari

QasimK:
Fix insane random scrolling up & down when trying to edit an IText.
The cause is a bug with iOS 8 & 9 and use of fixed position elements.

Fixes [#1985](https://github.com/kangax/fabric.js/issues/1985).

References for bug:
- https://stackoverflow.com/questions/29001977/safari-in-ios8-is-scrolling-screen-when-fixed-elements-get-focus
- https://meta.discourse.org/t/dealing-with-ios-8-mobile-safari-bugs/24101

Note an [alternative PR #2615](https://github.com/kangax/fabric.js/pull/2615) for this issue.

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -4,12 +4,17 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    * Initializes hidden textarea (needed to bring up keyboard in iOS)
    */
   initHiddenTextarea: function() {
+    var text = this.canvas.getActiveObject();
+    // Position in the middle because we don't know where the cursor is
+    var left = text.oCoords.ml.x;
+    var top = text.oCoords.ml.y;
+
     this.hiddenTextarea = fabric.document.createElement('textarea');
 
     this.hiddenTextarea.setAttribute('autocapitalize', 'off');
-    this.hiddenTextarea.style.cssText = 'position: fixed; bottom: 20px; left: 0px; opacity: 0;'
-                                        + ' width: 0px; height: 0px; z-index: -999;';
-    fabric.document.body.appendChild(this.hiddenTextarea);
+    this.hiddenTextarea.style.cssText = 'position: absolute; left: ' + left + 'px; top: ' + top + 'px;'
+                                        + 'opacity: 0; width: 0; height: 0; z-index: -999;';
+    this.canvas.contextContainer.canvas.parentNode.appendChild(this.hiddenTextarea);
 
     fabric.util.addListener(this.hiddenTextarea, 'keydown', this.onKeyDown.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'input', this.onInput.bind(this));
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -4,17 +4,17 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    * Initializes hidden textarea (needed to bring up keyboard in iOS)
    */
   initHiddenTextarea: function() {
-    var text = this.canvas.getActiveObject();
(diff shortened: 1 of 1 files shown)


src/mixins/itext_key_behavior.mixin.js
```diff
@@ -4,12 +4,17 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    * Initializes hidden textarea (needed to bring up keyboard in iOS)
    */
   initHiddenTextarea: function() {
+    var text = this.canvas.getActiveObject();
```
asturur: Why you have to cache this.canvas.getActiveObject(); when text at the end is 'this'?

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -4,12 +4,17 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
...
+    // Position in the middle because we don't know where the cursor is
+    var left = text.oCoords.ml.x;
```
asturur: instead of text.oCoords.ml.x use:
var rect = this.getBoundingRect(), left = rect.left, top = rect.top + rect.height;
In this way you have always medium left corner, also when the text is rotated or flipped.

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -4,12 +4,17 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
...
+
     this.hiddenTextarea = fabric.document.createElement('textarea');
 
     this.hiddenTextarea.setAttribute('autocapitalize', 'off');
-    this.hiddenTextarea.style.cssText = 'position: fixed; bottom: 20px; left: 0px; opacity: 0;'
-                                        + ' width: 0px; height: 0px; z-index: -999;';
-    fabric.document.body.appendChild(this.hiddenTextarea);
+    this.hiddenTextarea.style.cssText = 'position: absolute; left: ' + left + 'px; top: ' + top + 'px;'
+                                        + 'opacity: 0; width: 0; height: 0; z-index: -999;';
+    this.canvas.contextContainer.canvas.parentNode.appendChild(this.hiddenTextarea);
```
asturur: this.canvas.contextContainer.canvas = this.canvas.lowerCanvasEl

asturur:
I left some comments. I made the same fix but i missed the part of the "fixed" positioning, every iphone user was reporting as no working infact.

If you could make up a fiddle with this fixed ( + comments ) version so we can ask to test to some iphone user, that would be super.

QasimK:
Thanks for your quick feedback @asturur!

I have created a [jsfiddle with fabric 1.5](https://jsfiddle.net/aj10zkqs/4/) with demonstrates the bug on my iPhone 6 with iOS 9.

I have created [another jsfiddle with the bug-fix applied](https://jsfiddle.net/d9htbp93/5/) with demonstrates much better behaviour.

There are two notes:
- For some reason the original jsfiddle's fabric.js has the line `fabric.util.addListener(this.hiddenTextarea, 'keypress', this.onKeyPress.bind(this));` instead of `fabric.util.addListener(this.hiddenTextarea, 'input', this.onInput.bind(this));`. I don't envision any impact - but it does affect typing.
- The jsfiddle page is not adapted for mobiles (e.g. no `width=device-width, initial-scale=1.0, user-scalable=no`) meaning you get poor zooming in and out when you try to edit the IText which can be (mostly) mitigated by tapping twice slowly to edit the text field.

asturur:
i'm gonna check this with android, only phone i have.

asturur:
i'm gettin hands on an iphone today, i will check it and ask for merge if everything is fine.

QasimK:
@asturur Did you get a chance to have look?

asturur:
I put the effort of this PR inside this one #2861 thanks to @QasimK that found the tricky part, not is just a matter of refinement.
@QasimK i would be happy if you could test the other solution also.

PR #2741, opened 2016-01-18
Added object property to disable object from group selections.

Stjaertfena:
Enhanced how group selections are made. Before an object could only be either 'selectable' or not, meaning point-and-click based and group-based selection worked the same way. I now added a bit of granularity to the selection model so an item can now be 'selectable' using only point-and-click based interaction and cannot be part of a group selection (e.g. if 'groupSelectable' is set to true).

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -573,6 +573,14 @@
     selectable:               true,
 
     /**
+     * When set to `false`, an object can not be selected for modification using group-based selection.
+     * But events still fire on it.
+     * @type Boolean
+     * @default
+     */
+    groupSelectable:          true,
+
+    /**
      * When set to `false`, an object can not be a target of events. All events propagate through it. Introduced in v1.3.4
      * @type Boolean
      * @default
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -574,7 +574,6 @@
 
     /**
      * When set to `false`, an object can not be selected for modification using group-based selection.
-     * But events still fire on it.
      * @type Boolean
      * @default
      */
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -77,19 +77,25 @@
      * @private
      */
     _createActiveGroup: function(target, e) {
+      if (!this._activeObject) {
+        // Group cannot be created if no existing active object is present.
+        return;
+      }
 
-      if (this._activeObject && target !== this._activeObject) {
+      var groupSelectable = (this._activeObject.groupSelectable && target.groupSelectable),
+          isNewSelectionObject = (target !== this._activeObject),
+          group;
 
-        var group = this._createGroup(target);
+      if (isNewSelectionObject && groupSelectable) {
+        group = this._createGroup(target);
         group.addWithUpdate();
 
         this.setActiveGroup(group);
         this._activeObject = null;
 
         this.fire('selection:created', { target: group, e: e });
+        target.set('active', true);
       }
-
-      target.set('active', true);
     },
 
     /**
@@ -150,7 +156,7 @@
       for (var i = this._objects.length; i--; ) {
         currentObject = this._objects[i];
 
-        if (!currentObject || !currentObject.selectable || !currentObject.visible) {
+        if (!currentObject || !currentObject.selectable || !currentObject.groupSelectable || !currentObject.visible) {
           continue;
         }
 


src/shapes/object.class.js
```diff
@@ -573,6 +573,14 @@
     selectable:               true,
 
     /**
+     * When set to `false`, an object can not be selected for modification using group-based selection.
+     * But events still fire on it.
+     * @type Boolean
```
asturur: what do you mean by "but events still fire on it"? If there is some conflict between .selectable and .evented should be solved anyway regardless of this new feature

Stjaertfena: Sorry, I incorrectly commited that comment, it has no meaning. It is now removed.

Stjaertfena:
@asturur Are you happy for this to be merged now or is there anything else that you want me to change?

asturur:
i do not merge, @kangax does.
I'm not really sure about usefullness and application of another option that go togheter with evented and selectable adding complexity for developers. Some may find this options usefull anyway.

Stjaertfena:
@kangax Do you have any updates on this one? Will it be merged or not?

kangax:
So... with this flag on, you'd be able to select multiple objects by shift-clicking but not by typical draw-box-around selection?

kangax:
@Stjaertfena could you explain your use case for this? I don't remember getting requests for this kind of granular functionality, so I'll mark it as possible feature for now.

Stjaertfena:
@kangax So the use case is: Lets say you have a number of objects drawn to a canvas, with all default values set, you are able to select the objects using point-and-click, and you are also able to select a group of items either by shift clicking or by drawing a selection box using the cursor. That is how it works today.

With the new object option "groupSelectable", which defaults to true, you get the same interaction experience as described above. When the flag is set to false, those objects can never be part of a selection group, whether you are shift-clicking or drawing a selection box.

In my current application I have this scenario where I want to be able to have different objects drawn to the same canvas, but some objects should never be able to be part of a group selection (whether shift-clicking or drawing a selection box). I guess you can consider this pull request a feature request. ;)

Let me know if you require any additional information.

dhakan:
+1 for this feature

Stjaertfena:
@kangax Have you made up your mind about this feature request? =)

kangax:
> but some objects should never be able to be part of a group selection (whether shift-clicking or drawing a selection box)

So... why not just make them unselectable? http://fabricjs.com/customization/

Stjaertfena:
If you refer to setting the object property "selectable" to false, it disables all selections on that particular object, e.g. point-and-click, shift clicking as well as drawing a selection box. What I need, and what I'm suggesting, is adding a property to only disable group selection (e.g. selections done by either drawing a box or shift clicking). However, I still want to be able to select objects individually, but never part of a group selection.

I believe this is not possible in the current implementation, or am I wrong?

asturur:
I do not want to be harsh neither bad. But to me it looks like something to
be implemented at application level, i would prefer, if not exist already,
to fire a "beforeselection" event other than adding a new parameter. Just
my opinion.
Il 01 feb 2016 09:32, "Alexis" <notifications@github.com> ha scritto:

Stjaertfena:
I guess we could, however it is a fairly small change to fabric and we though it might be something that others would like to use. It's your call, if it does not get merged we'll just stick to our own fork...

asturur:
Would not an event, to check a property, custom or not, be more flexible and less specific?

Onbeforeselection .... Return false... Do not select or something like that.

What do you think?

Stjaertfena:
Hmm, maybe... But wouldn't that go against the current implementation logic, where you have the "selection" property on the canvas and the "selectable" property on the object, if this particular scenario would be handled in an event callback? Also, in the event based solution you would have to know how the selection was made, e.g. using point-and-click, shift-click, or drawing of selection box.

I'm not sure if the event solution is the way to go, maybe because I've already implemented an alternate solution that does the job. =P

dhakan:
Did you make up your mind regarding this yet, @asturur and @kangax ? I don't really see how an onbeforeselection event could work in this case. Care to explain? 

Example: If you have a football field with 11 players represented as fabric.Object, and you want to multiselect all of these players at once, but not include the ball(say a fabric.Circle) in the selection being made, the only way this would be possible is to have the football be non-selectable. But what if I want to be able to select the football alone?

ncou:
Hi, this feature could be usefull if you want to "lock" an object, this could prevent a selection from the group but still be possible to select the object (while selectable=false prevent all the click on the object) to "unlock" it if necessary.

+1 for this function :) :+1:

PR #2754, opened 2016-01-26
Fallback to styleElement.text for IE9

toddbc:
IE9 doesn't provide a `textContent` property on this object.  This caused errors when trying to import SVG files with `<style>` inside them.


kangax:
Could you please add a comment for why this is needed as well?

toddbc:
Ah, good point - done.

PR #2770, opened 2016-02-04
pass image/img to callback

asturur:
Inspired from a stack overflow question, i was thinking this is a relatively free feature that can come handy to set some image properties after loading it.

we can still refer to this[property] in the callback if we do not like this.

What do you think?


kangax:
Sure, fine with me :)

kangax:
Just need to update docs and stuff

PR #2788, opened 2016-02-12
Fix typo: Cario should be Cairo

negomi:
Link is correct but the text is incorrect, which could cause confusion.


kangax:
Thanks!

PR #2794, opened 2016-02-15
Add support for native getters and setters in createClass

### LINKED ISSUE #2786: util.createClass to support (native) getters and setters
Our customers are of a special kind: they always manage to break the canvas.

A week ago, a customer managed to set the scale of a text object to 0, which caused massive rendering issues (division by zero) on our backend and caused it to crash. Yesterday, another customer managed to nullify a value that should have been a String. These are only two examples.

Therefore, I want to implement input sanitation on the frontend and backend by using getters and setters for values like these. They would not be traditional getters and setters `setScaleX(val)` but would use the property descriptor `scaleX = val`, as we would otherwise need to refactor a large amount of the codebase. 

As the frontend and backend share the same Fabric implementation, it makes sense for me to do this either in our implementation, or in Fabric itself, which I prefer. I want to add support to the `createClass` function in Fabric itself to be able to do this. @kangax Do you think this would be accepted as pull request provided the code is clean, performant and won't introduce breaking changes?

The current flow would still be supported and would be left completely untouched:

``` javascript
fabric.MyImageClass = fabric.util.createClass(fabric.Image, {
    type: 'MyImageClass',
    originX: 'center',
    offsetBackX: 0,
    imageId: null
});
```

But then it would also support:

``` javascript
fabric.MyImageClass = fabric.util.createClass(fabric.Image, {
    get type() {
        return 'MyImageClass'
    },

    get originX() {
        return this._originX;
    },

    set originX(val) {
        if (/left|center|foobar/.test(val)) {
            this._originX = val;
        }
    },

    get offsetBackX() {
        return this._offsetBackX;
    },

    set offsetBackX(val) {
        //cast to 32bit integer
        this._offsetBackX = val | 0;
    },

    get imageId() {
        return this._imageId;
    },

    set imageId(val) {
        if (typeof val === 'string') {
            this._imageId = val;
        }
    },

    initialize: function() {
        this.originX = 'center';
        this.offsetBackX = 0;
        this.imageId = null; //invalid, won't be set
    }
});
```

StephanBijzitter:
Patch resolves issue #2786;
Branch is based off the v1.5.0 tag

### DIFF
--- a/src/util/lang_class.js
+++ b/src/util/lang_class.js
@@ -50,7 +50,8 @@
             var descriptor = getOwnPropertyDescriptor(source, property);
             if (descriptor && (descriptor.get || descriptor.set)) {
               Object.defineProperty(klass.prototype, property, descriptor);
-            } else {
+            }
+            else {
               klass.prototype[property] = source[property];
             }
           }
--- a/src/util/lang_class.js
+++ b/src/util/lang_class.js
@@ -2,14 +2,18 @@
 
   var slice = Array.prototype.slice, emptyFunction = function() { },
 
+      makePropertyDescriptor = function(value) {
+        return {
+          value: value,
+          writable: true,
+          enumerable: true,
+          configurable: true
+        };
+      },
+
       getOwnPropertyDescriptor = (function() {
         function getOwnPropertyDescriptorPolyfill(obj, prop) {
-          return (prop in obj) ? {
-            value: obj[prop],
-            writable: true,
-            enumerable: true,
-            configurable: true
-          } : undefined;
+          return (prop in obj) ? makePropertyDescriptor(obj[prop]) : undefined;
         }
 
         return Object.getOwnPropertyDescriptor || getOwnPropertyDescriptorPolyfill;
@@ -48,12 +52,12 @@
           }
           else {
             var descriptor = getOwnPropertyDescriptor(source, property);
-            if (descriptor && (descriptor.get || descriptor.set)) {
-              Object.defineProperty(klass.prototype, property, descriptor);
+
+            if (!descriptor || !(descriptor.get || descriptor.set)) {
+              descriptor = makePropertyDescriptor(source[property]);
             }
-            else {
-              klass.prototype[property] = source[property];
-            }
+
+            Object.defineProperty(klass.prototype, property, descriptor);
           }
 
           if (IS_DONTENUM_BUGGY) {
--- a/src/util/lang_class.js
+++ b/src/util/lang_class.js
@@ -34,7 +34,12 @@
             })(property);
           }
           else {
-            klass.prototype[property] = source[property];
+            var descriptor = Object.getOwnPropertyDescriptor(source, property);
+            if (descriptor && (descriptor.get || descriptor.set)) {
+              Object.defineProperty(klass.prototype, property, descriptor);
+            } else {
+              klass.prototype[property] = source[property];
+            }
           }
 
           if (IS_DONTENUM_BUGGY) {
--- a/src/util/lang_class.js
+++ b/src/util/lang_class.js
@@ -2,6 +2,19 @@
 
   var slice = Array.prototype.slice, emptyFunction = function() { },
 
+      getOwnPropertyDescriptor = (function() {
+        function getOwnPropertyDescriptorPolyfill(obj, prop) {
+          return (prop in obj) ? {
+            value: obj[prop],
+            writable: true,
+            enumerable: true,
+            configurable: true
+          } : undefined;
+        }
+
+        return Object.getOwnPropertyDescriptor || getOwnPropertyDescriptorPolyfill;
+      })(),
+
       IS_DONTENUM_BUGGY = (function() {
         for (var p in { toString: 1 }) {
           if (p === 'toString') {
@@ -34,7 +47,7 @@
             })(property);
           }
           else {
-            var descriptor = Object.getOwnPropertyDescriptor(source, property);
+            var descriptor = getOwnPropertyDescriptor(source, property);
             if (descriptor && (descriptor.get || descriptor.set)) {
               Object.defineProperty(klass.prototype, property, descriptor);
             } else {
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 *.iml
+.idea
 /node_modules/
 /npm-debug.log
 before_commit


StephanBijzitter:
Right, I fixed the scoping issue. However, this may introduce backward-incompatibility.

I tested this locally and everything seemed fine to me. Will wait for the regression testing to be completed and report back.

asturur:
I have a question.
We have an issue open about switching to ES6 and using babel for ES5 browsers.

This native setters and getters are something that will be taken care of by ES6 syntax?

I m not expert about this topic (es6 and classes in general)

StephanBijzitter:
I'm not sure I understand your question.

If you are asking whether the ES6 transpiling will render this pull request useless, then my answer will be a firm no.

Let's take the following example:
```javascript
fabric.util.createClass(fabric.Object, {
    type: 'MyObject',

    get name() {
        return 'Stephan'
    },

    set name(val) {
        //don't do anything, it's constant!
    },

    get age() {
        return this._age;
    },

    set age(val) {
        if (val < this._age) {
            throw new TypeError('Beauty products can help you make you look younger, but your age will not decrease!');
        }

        this._age = val;
    },

    initialize: function() {
        this.age = 10;
    }
});
```

Fabric.js (before this pull request), will treat this code as follows:
```javascript
fabric.util.createClass(fabric.Object, {
    type: 'MyObject',
    name: 'Stephan',
    age: 10
});
```

Now, this is a problem for us. The name of the object can now be changed, but we don't want that, it should be a constant. Next to that, we lose the ability to check the  proposed new age before setting it, allowing people to become younger.

This can be solved in two different ways:

1. We can use property descriptors to define properties, allowing native accessors;
2. We can use separate accessor functions.

For the first option, you can see the first code block in this message. The most important advantage is that it does not require any changes in an application that switches from example 2, to example 1.

The second option, would look like this:
```javascript
fabric.util.createClass(fabric.Object, {
    type: 'MyObject',
    _name: 'Stephan',
    _age: 10,

    getName: function() {
        return this._name;
    },

    setName: function() {
        //don't do anything, it's constant!
    },

    getAge: function() {
        return this._age;
    },

    setAge: function(val) {
        if (val < this._age) {
            throw new TypeError('Beauty products can help you make you look younger, but your age will not decrease!');
        }

        this._age = val;
    }
});
```
This works. However, it would require the entire codebase to be updated, too! Any reference to `myObject.age` would need to be changed to `myObject.setAge` or `myObject.getAge`!

In our opinion, the first option is a much cleaner solution. The fact that it is built-in JavaScript behaviour makes it even better, as even IE9 supports it! For browsers that do not support it, we can still easily polyfill back to the old behaviour (as is added in this PR).

I would be the first one to cheer if Fabric starts using ES6 and transpiles to ES4 or ES5, and I would certainly be able and willing to help! However, this would have no effect on the above mentioned problem. ES6 would simply allow for much more efficient development.

StephanBijzitter:
Regression tests have been completed and we had no issues with compatibility. However, that may just have been luck. I can certainly imagine other projects depending on Fabric _might_ find this change to break their applications.

So this would be safe for a v2.0.0 release, but v1.6.0 could be a bit of a gamble.

asturur:
I'm so sorry i cannot contribute because i still do not get it.
But i have to understand soon or later also this getter and setters and the different ways we create them.

src/util/lang_class.js
```diff
@@ -2,6 +2,23 @@
 
   var slice = Array.prototype.slice, emptyFunction = function() { },
 
+      makePropertyDescriptor = function(value) {
```
StephanBijzitter: The `makePropertyDescriptor ` creates a descriptor object for properties that were added to an object without descriptor.

Let's take the following example:
```javascript
var source = {};
source.foo = 'bar';
var descriptor = Object.getOwnPropertyDescriptor(source, 'foo');
```
Here, the `descriptor` object is `undefined`, because the property `foo` was not set using a descriptor. Because `Object.defineProperty` requires a descriptor, we need to construct one ourselves. You can read more about the individual arguments of a descriptor in [this](http://patrickdelancy.com/2012/09/property-descriptors-in-javascript/) slightly outdated blog, which I Googled.

We can write the exact same code (behaviour-wise) like this:
```javascript
var source = {};
Object.defineProperty(source, 'foo', {
  value: 'bar', //initial value
  writable: true, //allow source.foo = 'foobar'
  enumerable: true, //('foo' in source) will be true
  configurable: true //allow the descriptor to be changed later
})
```
The advantage to using descriptors can now be seen, as we're able to completely control how this property must behave on an engine level. If we want the value of the property to be constant, we would just give the `writable` argument a value of `false`.

src/util/lang_class.js
```diff
@@ -2,6 +2,23 @@
...
+        return {
+          value: value,
+          writable: true,
+          enumerable: true,
+          configurable: true
+        };
+      },
+
+      getOwnPropertyDescriptor = (function() {
```
StephanBijzitter: The `getOwnPropertyDescriptor` is partially supported in IE8, although I could not find information regarding what is actually implemented and what is not implemented. Thankfully, IE9 fully supports this method.

For the sake of supporting older browsers, we can drop some functionality in favour of compatibility. Any property will be treated as if it were declared as follows:
```javascript
var source = {};
source.foo = 'bar';
```
This inevitably means that the descriptor adds no direct benefit if polyfilled, but we're still using it to advocate consistency.

src/util/lang_class.js
```diff
@@ -34,7 +51,13 @@
...
           }
           else {
-            klass.prototype[property] = source[property];
+            var descriptor = getOwnPropertyDescriptor(source, property);
+
+            if (!descriptor || !(descriptor.get || descriptor.set)) {
+              descriptor = makePropertyDescriptor(source[property]);
+            }
+
+            Object.defineProperty(klass.prototype, property, descriptor);
```
StephanBijzitter: Finally, we're going to use the awesome new functionality to set every property using a descriptor.

The first block of code uses the browser's version of `getOwnPropertyDescriptor` if available, but would otherwise use the polyfill.

Then, in the second block, we're going to test the descriptor.
* If there is no descriptor (it is `undefined`);
* Or if there is a descriptor, but there is neither a getter, nor a setter in the descriptor, we can safely assert that the property is meant to be declared as a regular variable as in `source.foo = 'bar'`.

If either of these options are met, we're going to create a descriptor for the property using the default arguments. This would allow other developers to update Fabric without having to re-write all their custom class declarations.

Lastly, in the third block, we're properly defining the property with its intended descriptor.

src/util/lang_class.js
```diff
@@ -2,6 +2,23 @@
...
   var slice = Array.prototype.slice, emptyFunction = function() { },
 
+      makePropertyDescriptor = function(value) {
+        return {
+          value: value,
+          writable: true,
+          enumerable: true,
+          configurable: true
+        };
+      },
```
StephanBijzitter: The `makePropertyDescriptor ` creates a descriptor object for properties that were added to an object without descriptor.

Let's take the following example:
```javascript
var source = {};
source.foo = 'bar';
var descriptor = Object.getOwnPropertyDescriptor(source, 'foo');
```
Here, the `descriptor` object is `undefined`, because the property `foo` was not set using a descriptor. Because `Object.defineProperty` requires a descriptor, we need to construct one ourselves. You can read more about the individual arguments of a descriptor in [this](http://patrickdelancy.com/2012/09/property-descriptors-in-javascript/) slightly outdated blog, which I Googled.

We can write the exact same code (behaviour-wise) like this:
```javascript
var source = {};
Object.defineProperty(source, 'foo', {
  value: 'bar', //initial value
  writable: true, //allow source.foo = 'foobar'
  enumerable: true, //('foo' in source) will be true
  configurable: true //allow the descriptor to be changed later
})
```
The advantage to using descriptors can now be seen, as we're able to completely control how this property must behave on an engine level. If we want the value of the property to be constant, we would just give the `writable` argument a value of `false`.

src/util/lang_class.js
```diff
@@ -2,6 +2,23 @@
...
+
+      getOwnPropertyDescriptor = (function() {
+        function getOwnPropertyDescriptorPolyfill(obj, prop) {
+          return (prop in obj) ? makePropertyDescriptor(obj[prop]) : undefined;
+        }
+
+        return Object.getOwnPropertyDescriptor || getOwnPropertyDescriptorPolyfill;
+      })(),
```
StephanBijzitter: The `getOwnPropertyDescriptor` is partially supported in IE8, although I could not find information regarding what is actually implemented and what is not implemented. Thankfully, IE9 fully supports this method.

For the sake of supporting older browsers, we can drop some functionality in favour of compatibility. Any property will be treated as if it were declared as follows:
```javascript
var source = {};
source.foo = 'bar';
```
This inevitably means that the descriptor adds no direct benefit if polyfilled, but we're still using it to advocate consistency.

src/util/lang_class.js
StephanBijzitter: Finally, we're going to use the awesome new functionality to set every property using a descriptor.

The first block of code uses the browser's version of `getOwnPropertyDescriptor` if available, but would otherwise use the polyfill.

Then, in the second block, we're going to test the descriptor.
* If there is no descriptor (it is `undefined`);
* Or if there is a descriptor, but there is neither a getter, nor a setter in the descriptor, we can safely assert that the property is meant to be declared as a regular variable as in `source.foo = 'bar'`.

If either of these options are met, we're going to create a descriptor for the property using the default arguments. This would allow other developers to update Fabric without having to re-write all their custom class declarations.

Lastly, in the third block, we're properly defining the property with its intended descriptor.

PR #2806, opened 2016-02-24
Fix retina scaling shadows and data url

### LINKED ISSUE #2744: toDataUrl exports are automatically scaled up and cropped on 1.6RC
When I use toDataUrl on a canvas, it exports the image scaled up and cropped. 
It has the same behaviour even if I set multiplier, size or top/left position. 

Reverting back to 1.5 fixed the issue.

### LINKED ISSUE #2801: Text shadows rendered incorrectly, off by pixels
Hey There,

Since it looks like this is a active issue queue, here is another bug with shadows. See the below SVG. The shadow renders in the incorrect x/y offset position inside fabric.js when compared with browser/svg viewer. I know this is an extreme example of 100 pixel X offset, but it illustrates there is some bug with your shadow pixel calculations. Think I have some other bugs I found as well. I will add them when I have a better handle on them. 

```
<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" width="300" height="300" style="background-color: rgba(0,0,255,0.3)" viewBox="0 0 300 300" xml:space="preserve">
<desc>Created with Fabric.js 1.6.0-rc.1</desc>
<defs></defs>
<filter id="SVGID_3" y="-20%" height="140%" x="-33%" width="166%" >
	<feGaussianBlur in="SourceAlpha" stdDeviation="0"></feGaussianBlur>
	<feOffset dx="100" dy="0" result="oBlur" ></feOffset>
	<feFlood flood-color="rgb(0, 0, 0)"/>
	<feComposite in2="oBlur" operator="in" />
	<feMerge>
		<feMergeNode></feMergeNode>
		<feMergeNode in="SourceGraphic"></feMergeNode>
	</feMerge>
</filter>
	<g transform="translate(392.5 110.17)">
		<text font-family="Times New Roman" font-size="195" font-weight="normal" style="stroke: none; stroke-width: 0; stroke-dasharray: none; stroke-linecap: butt; stroke-linejoin: miter; stroke-miterlimit: 10; fill: #00ff00; fill-rule: nonzero; opacity: 1;filter: url(#SVGID_3);" >
			<tspan x="-392.5" y="61.42" fill="#00ff00">Some text</tspan>
		</text>
	</g>
</svg>
```

asturur:
Do not pull untill confirmation screenshots.

closes #2801
closes #2744


asturur:
canvas to dataUrl:
image

original canvas:
image

looks like that if i'm setting fabric.devicePixelRation = 2, the dataUrl export effectively produce a double size canvas and shadows are correct.
@kangax you can merge for me.

keanass:
what about node js the problem is still there. when put shadow to bigger image image and export the shadow is doubling : 

beforescale
before scale image with shadow in browser : 
 


after scale in browser : 
beforetransform


after send json to server and generate with dataurl : 

aftertrans


any solution for this ??

PR #2820, opened 2016-02-29
fix(itext): fix bug that IME can not be inputted

### LINKED ISSUE #2473: Version 1.6.0-rc.1.  Cannot input Chinese character in iText in Chrome, windows 7
Test version 1.6.0-rc.1 on windows 7.  iText can not input Chinese characters in Chrome. 

Tested in firfox 39.0, IE10, Chrome 45.0.2454.85 
This issue only happens on Chrome. 
Whatever key is pressed,  the key displays in iText as English Character.  No Chinese input interface is activated .  

Before this I tried 1.5.0 version,  the Chinese input interface was activated,  but can not input Chinese characters as well.

abruzzihraig:
Fix #2473 

Fabric.js does not support IME input, it is looks weird especially in Chinese input. 

The reason is fabric just listen the event `oninput`, and calculate a diff between the last two text, then inserts those new chars. The key is It doesn't have a reducing logic here. 

For example, when you type Chinese text with any IME, the previous result may looks like `xxxnanrenwangxxx`,  but the next you will got `xxx男人王xxx` after entered a return button. Note the length of the chars reduce from 16 to 9, but there is no reducing logic in onInput handler.

So I use the events `compositionstart` and `compositionend` to check if users are using IME and prevent any insertion before they ended IME typing.

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -99,14 +99,14 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
   /**
    * Composition start
    */
-  onCompositionStart: function(e) {
+  onCompositionStart: function() {
     this.inCompositionMode = true;
   },
 
   /**
    * Composition end
    */
-  onCompositionEnd: function(e) {
+  onCompositionEnd: function() {
     this.inCompositionMode = false;
   },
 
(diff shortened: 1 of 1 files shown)


kangax:
LGTM

asturur:
Looks good also to me. It could avoid to switch all the text input as planned. Can we pull after 1.6.0 release? I would be scared of another 0 day bug after release.

asturur:
@abruzzihraig can you fix the lint errors? ( variable e declared but not used in the events  you added ).

asturur:
Proven to be usefull also for managing accents on some configurations.

asturur:
I leave a note: check if thos compositioncheck is also usefull when using arrows on some japanes input methods

astw:
Is this issue include 1.6.0 release?  I still can see this issue in http://fabricjs.com/kitchensink
Or the demo is not updated?

asturur:
something brougth the issue back. i hope on the weekend i can clear something out

PR #2826, opened 2016-03-10
Added the overlay blending method

khawerrind:
Overlay
Multiplies or screens the colors, depending on the base color. Patterns or colors overlay the existing pixels while preserving the highlights and shadows of the base color. The base color is not replaced, but mixed with the blend color to reflect the lightness or darkness of the original color.

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -19347,6 +19347,11 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
             data[i + 1] = Math.max(g, tg);
             data[i + 2] = Math.max(b, tb);
             break;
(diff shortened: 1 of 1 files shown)


milan-w:
You only patched the distribution file. This would be overwritten with every new build of fabric.
If you want this feature to be permanent, you have to edit the blend filter class. 
https://github.com/khawerrind/fabric.js/blob/patch-2/src/filters/blend_filter.class.js

Also it seems that the Blend filter is currently untested. If you know how to write a meaningful test for your overlay filter and the blend method in general, I think it would be very much appreciated.  
You can find the current tests here: https://github.com/khawerrind/fabric.js/blob/patch-2/test/unit/image_filters.js

PR #2842, opened 2016-03-17
Do not draw control borders in normal group with active objects

asturur:
Currently addWithUpdate set objects to active by default ( probably because it has been created for shift click and active group ).

Since we changed the rendering queque, objects inside group that are marked as active get borders drawn too, that is wrong.

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1075,7 +1075,8 @@
      * @param {Boolean} [noTransform] When true, context is not transformed
      */
     _renderControls: function(ctx, noTransform) {
-      if (!this.active || noTransform) {
+      if (!this.active || noTransform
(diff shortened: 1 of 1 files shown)


asturur:
Before deciding to patch there i considered that:
We could add an __origiHasBorders in par with __origHasControls to skip drawing border, but then we should discern when creating group if the group would be an activeGroup or a normal one.

Also we could check if the object should be set to active or not by addWithUpdate, but this was more complicated.

Also the starting check in renderControls ``` if (!this.active || ) ``` overcome any   ``` hasControls ``` or  ``` hasBorders ``` so having a similar check for  ``` my object is in not in activeGroup ``` makes sense to me.

@kangax if i do not get any negative comment from you or @kienz i will merge this.

PR #2853, opened 2016-03-26
Textboxspaces

### LINKED ISSUE #2829: SVG Export for Textbox loses style on the 2nd line, and other issues
## Version
1.6.0

## Test Case
http://jsfiddle.net/jtweku/cdjth9k4/2/

there are 3 issues all with the second textbox, 
1. the second textbox lost the formatting after the first line when exporting to SVG.
2. space should not be there on the first beginning of the second line.
3. did you notice that the last character ('x') on the second line is not Bold, I use selectAll when applying style, unless Im not doing it right.  

## Expected Behavior
I have a fix for the svg export, remove the commented line of the code to see the correct behaviour 
_setSVGTextLineChars in the IText class should use the _getStyleDeclaration (because it is overridden in the Textbox class)

asturur:
Check properly if the iText or TextBox object has the style set for a particular line.
This is a temporary solution untill we can schedule a full svg export cleaning.

Infact big text chunks with same style are exporte as single character tspans.

closes #2829

### DIFF
--- a/src/mixins/itext.svg_export.js
+++ b/src/mixins/itext.svg_export.js
@@ -9,7 +9,7 @@
      * @private
      */
     _setSVGTextLineText: function(lineIndex, textSpans, height, textLeftOffset, textTopOffset, textBgRects) {
-      if (!this.styles[lineIndex]) {
+      if (!this.hasStyleOnLine(lineIndex)) {
         fabric.Text.prototype._setSVGTextLineText.call(this,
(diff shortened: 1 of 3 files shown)


asturur:
@kangax, @kienz it is ok for you to include a temporary solution like this one?

The svg export of styled text needs a deep rework and this solution, yet not optimal fixes a current bug with textbox.

PR #2868, opened 2016-04-05
Hidden text area

### LINKED ISSUE #1985: Focus craziness on iOS with iText

### LINKED ISSUE #2740: setSelectionStyles behaviour not same on all characters selected
Once a text selection is made, I cannot change the font size for the entire itext object.

// clear canvas
canvas.clear();
var t = new fabric.IText("Hello world !", {
  top: 100,
  left: 100,
  fontSize: 12,
});

canvas.add(t);
canvas.setActiveObject(t);
t.selectAll();
var style = { };
style['fontSize'] = 84;
style['textBaseline'] = 'bottom';
t.setSelectionStyles(style);
canvas.renderAll();
canvas.discardActiveObject();
canvas.setActiveObject(t);
t.setFontSize(48);
canvas.renderAll();

### LINKED ISSUE #2773: Pasting after 'cut'-ing text in i-text pastes previous data from system clipboard
On the [kitchen sink demo](http://fabricjs.com/kitchensink/), cmd+x and a subsequent cmd+v always pastes text cut/copied previously to the system clipboard.

This happens on osx 10.11.3 on chrome, safari, and ff.

### LINKED ISSUE #2823: Bug Texbox in Explorer/Edge for selected text
## Version
1.6.0-rc.1

## Test Case
http://jsfiddle.net/2o46nnka/20/

Fails in Edge and partly in Chrome

## Steps to reproduce
Use Edge 
1. Manually select a part of the text  (ex ou in Your) and key a replacement text 12
The text is messed up in Edge not replacing the selected text.

2. Second test select - behaviour in Edge depends on the selection is made from left to right or right to left.
Select a part of the word (ex ou in Your) and select by marking from right to left  then key 12
Result is You12r

Test once again but select from left to right.
The result is now Y1ou2r

3. Chrome
Seems to work in Firefox and Chrome. BUT if you reduce the box there are also some mess up in Chrome.
Reproduce in this fiddle where the Width parameter is omitted
http://jsfiddle.net/y6rzoxu2/1/
Select ou and key 12
Result in Chrome is Yo12 and not Y12r
Seems to relate to the cursor position. 

Try again http://jsfiddle.net/y6rzoxu2/1/  - position the cursor before Y and press BACK.
Y disappears !
Somehow Textbox loose track of the cursor position.

## Expected Behaviour
The selected text should be replaced with the text keyed.

## Actual Behaviour
The text entered is mixed with the text to be replaced. Seems to be depending on the direction of the selection.

### LINKED ISSUE #2839: IText copy not working in IE and Edge
Reproduce: try to copy text with IE and Edge from an IText textarea on http://fabricjs.com/test/misc/itext.html

I tried to catch the copy event, but no matter how I tried with Angular (ng-copy), jQuery ($('...').on('copy', ...), and plain JavaScript addEventListener, I couldn't catch copy events in IE and Edge. Does anyone in which way IE and Edge handle copy event differently than other browsers?

ps. paste seems to work, though

### LINKED ISSUE #2859: Itext with style but no fill and no stroke
## Version
1.6.0

## Test Case
http://jsfiddle.net/Da7SP/39/

## Steps to reproduce
Draw some iText with no fill and no stroke properties, put fill and stroke in the style instead
## Expected Behavior
Text should render
## Actual Behavior
Text does not render

This happen because we have a check on stroke and fill before going trought rendering.
There is a ambiguous property in Itext "_skipFillStrokeCheck" that is default to false.

Instead of setting it to true to force rendering of this kind of text, it would be better to add a proper check before rendering:

```
if ((!this.stroke || this.strokeWidth === 0) && this.isEmptyStyles()) {
        return;
      }
if ((!this.fill) && this.isEmptyStyles()) {
        return;
      }
```

and remove _skipFillStrokeCheck that is of no clear use to me.

Normal text class has no isEmptyStyles functions.

We could do in two way:
1) check for a lazy !this.styles instead of isEmptyStyles()
2) add a isEmptyStyles function that return true for the Text class

@kangax and @kienz if you have 2 minutes i need your opinion on this.

asturur:
Improves iphone focus crazyness on edit
improve japanese input method editor
improve copy cut paste behaviour of internet explorer
improve cut for all browsers ( cut was cutting but not copying )
fixes Shadow not rendered from style object (since we add shadow scaling with canvas)

closes #2823
closes #1985
closes #2839 
closes #2773
closes #2740 
closes #2859

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -337,7 +337,7 @@
      * @return {fabric.IText} thisArg
      * @chainable
      */
-    enterEditing: function() {
+    enterEditing: function(e) {
       if (this.isEditing || !this.editable) {
         return;
       }
@@ -348,7 +348,7 @@
 
       this.isEditing = true;
 
-      this.initHiddenTextarea();
+      this.initHiddenTextarea(e);
       this.hiddenTextarea.focus();
       this._updateTextarea();
       this._saveEditingProps();
@@ -420,13 +420,37 @@
      * @private
      */
     _updateTextarea: function() {
-      if (!this.hiddenTextarea) {
+      if (!this.hiddenTextarea || this.inCompositionMode) {
         return;
       }
 
       this.hiddenTextarea.value = this.text;
       this.hiddenTextarea.selectionStart = this.selectionStart;
       this.hiddenTextarea.selectionEnd = this.selectionEnd;
+      if (this.selectionStart === this.selectionEnd) {
+        var p = this._calcTextareaPosition();
+        this.hiddenTextarea.style.left = p.x + 'px';
+        this.hiddenTextarea.style.top = p.y + 'px';
+      }
+    },
+
+    /**
+     * @private
+     */
+    _calcTextareaPosition: function() {
+      var chars = this.text.split(''),
+          boundaries = this._getCursorBoundaries(chars, 'cursor'),
+          cursorLocation = this.get2DCursorLocation(),
+          lineIndex = cursorLocation.lineIndex,
+          charIndex = cursorLocation.charIndex,
+          charHeight = this.getCurrentCharFontSize(lineIndex, charIndex),
+          leftOffset = (lineIndex === 0 && charIndex === 0)
+                    ? this._getLineLeftOffset(this._getLineWidth(this.ctx, lineIndex))
+                    : boundaries.leftOffset,
+          m = this.calcTransformMatrix(),
+          p = { x: boundaries.left + leftOffset, y: boundaries.top + boundaries.topOffset + charHeight};
+      this.hiddenTextarea.style.fontSize = charHeight + 'px';
+      return fabric.util.transformPoint(p, m);
     },
 
     /**
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -448,7 +448,7 @@
                     ? this._getLineLeftOffset(this._getLineWidth(this.ctx, lineIndex))
                     : boundaries.leftOffset,
           m = this.calcTransformMatrix(),
-          p = { x: boundaries.left + leftOffset, y: boundaries.top + boundaries.topOffset + charHeight};
+          p = { x: boundaries.left + leftOffset, y: boundaries.top + boundaries.topOffset + charHeight };
       this.hiddenTextarea.style.fontSize = charHeight + 'px';
       return fabric.util.transformPoint(p, m);
     },
--- a/src/mixins/itext_click_behavior.mixin.js
+++ b/src/mixins/itext_click_behavior.mixin.js
@@ -129,7 +129,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       }
 
       if (this.__lastSelected && !this.__corner) {
-        this.enterEditing();
+        this.enterEditing(options.e);
         this.initDelayedCursor(true);
       }
       this.selected = true;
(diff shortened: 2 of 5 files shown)


asturur:
summary:

Added keyup events, moved "CTRL + Key" combination on keyup so i cancel the event if the textarea consumes the event with other ( real copy and cut )
Added Ctrl+C for copy.
Added compositionUpdate event.
SetShadow was expecting some parameters, we add them at style declaration
Moving the textArea position when editing
managing japanese IME ( and maybe others )

PR #2872, opened 2016-04-07
Update dependencies to allow node 4, node 5 and 0.10, 0.11 and 0.12

asturur:
This should allow everyone to solve npm issues.
If travis can npm fabric, everyone should be able to do it.


asturur:
@inssein what do you think of this? you were experimenting with jsdom 6.5.0 and over, but that was killing previous nodejs version ( that i would do on a 2.0.0 release but not now ).

inssein:
@asturur I guess this works, since jsdom 3.0 doesn't seem to break our uses cases of it. That said, I would much rather try to use the latest jsdom version (its 8.3, but this PR is using 3.0).

It would be pretty simple to use the latest version of jsdom, probably need a few more lines from my previous PR and whatever has changed since.

asturur:
Newer jsdom is faster they say, but REQUIRES node 4.
Jsdom 3 is the latest that can handle nodejs 0.12.

inssein:
If we are going to be releasing a 2.0, I don't see why that would be a problem. Current LTS is 4.4.2, with stable at 5.10.1. Anybody who still needs node 0.12 can stay at v1.3. of fabric.

PR #2875, opened 2016-04-08
Canvas dispose unwrap the lowercanvas element

### LINKED ISSUE #2864: canvas.dispose() does not clear the html element created for initialization
After posting a [question on SO](http://stackoverflow.com/questions/36357627/objects-arent-selectable-after-disposing-and-creating-fabric-canvas-over-the-sa), I've reflected a bit on it and now I think it looks more like an actual issue with fabric.js, so I'm posting it here

## Version
1.5.0, 1.6.0-rc.1 (I've found [this issue](https://github.com/kangax/fabric.js/pull/2708), so gave it a try too)

## Test Case
http://codepen.io/aush/pen/dMVNdz - 1.5.0
http://codepen.io/aush/pen/dMVvYK - 1.6.0-rc.1

## Steps to reproduce
```javascript
let fabricCanvas = new fabric.Canvas('c');
fabricCanvas.add(new fabric.Rect({ left: 10, top: 10, fill: 'red', width: 20, height: 20 }));
fabricCanvas.dispose();
fabricCanvas = new fabric.Canvas('c');
fabricCanvas.add(new fabric.Rect({ left: 100, top: 100, fill: 'red', width: 20, height: 20 }));
```

## Expected Behavior
The rectangle should be selectable

## Actual Behavior
It isn't

The actual behavior probably means there is some state mutated by Fabric which isn't restored on dispose, which implies a leak.

I've got this problem not out of curiosity, but working on actual React/Redux app where I want to keep a state of my canvas in my Redux store (serialized to JSON) and applying it on render (deserialize JSON to a new Fabric.Canvas (but the same html canvas element, because React...)).

asturur:
needs proper testing for both static and normal canvas

closes #2864

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1244,6 +1244,25 @@
     },
 
     /**
+     * Clears a canvas element and removes all event listeners
+     * @return {fabric.Canvas} thisArg
+     * @chainable
+     */
+    dispose: function () {
+      this.callSuper('dispose');
+      var wrapper = this.wrapperEl;
+      this.removeListeners();
+      wrapper.removeChild(this.upperCanvasEl);
+      wrapper.removeChild(this.lowerCanvasEl);
+      delete this.upperCanvasEl;
+      if (wrapper.parentNode) {
+        wrapper.parentNode.replaceChild(this.lowerCanvasEl, this.wrapperEl);
+      }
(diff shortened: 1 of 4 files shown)


asturur:
added tests to element surrounding canvas

PR #2890, opened 2016-04-13
Allow a not stateful canvas to fire object:modified at end of transform.

### LINKED ISSUE #1927: lockRotation and object:modified event
Hi, I've noticed strange behavior when trying to rotate object that has `lockRotation` set to `true`. At first it performs `object:rotating` event on object even that it's not rotating. And what is even more unwanted when rotation is finished it fires `object:modified` but nothing has changed. When I investigated what properties has changed using this code:

```js
fabric.Object.prototype.hasStateChanged = function() {
  this.stateProperties.forEach(function(prop) {
    if (this.get(prop) !== this.originalState[prop]) {
      console.log('hasStateChanged(' + prop + ')')
      console.log('    current:  ' + this.get(prop));
      console.log('    original: ' + this.originalState[prop]);
    }
  }.bind(this));

  return this.stateProperties.some(function(prop) {
    return this.get(prop) !== this.originalState[prop];
  }, this);
};
```

it appeared to be left, top, originX, originY properties. I understand that it's because of changing origin point to `center`, `center` while rotating. But origin change shouldn't happen at all because rotation had been locked. When I set `lockRotation` to `false` it's detecting change of `angle` property and those four previously mentioned. But it should only show `angle`. My conclusion is that it shouldn't detect properties change if those properties were modified only as a side effect of transformation. Probably the same situation has place if it goes about scaling.

```
// From console
hasStateChanged(top) fabric.js:12846
    current:  196.06459398155516 fabric.js:12847
    original: 71.06459398155516 fabric.js:12848
hasStateChanged(left) fabric.js:12846
    current:  627.9755859375 fabric.js:12847
    original: 559.55517578125 fabric.js:12848
hasStateChanged(originX) fabric.js:12846
    current:  center fabric.js:12847
    original: left fabric.js:12848
hasStateChanged(originY) fabric.js:12846
    current:  center fabric.js:12847
    original: top 
```

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7676390-lockrotation-and-object-modified-event?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2362: Stateful: false doesn't fire object:modified
(cross posting from google groups)

Hi guys, quick question for what I suspect is a perfectly understood situation:

I've created a fiddle https://jsfiddle.net/sibbaldl/f6q3agtf/1/  that demonstrates a simple situation whereby, if a canvas is initialised with "stateful:false" the "object:modified" event won't fire when moving the red square around and releasing the mouse button. The event will fire if stateful:true.
It took me a while to realise the cause of this in my code, because even with stateful set to false other events such as scale,rotate,move will fire.

Just wanted to know if this is expected behaviour.

Cheers!

Lee

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25002104-stateful-false-doesn-t-fire-object-modified?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Need some manual testing.

The problem is this can fire if the object is dragged around and left in same place it was before dragging.

Closes #2362
closes #1927

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -652,18 +652,18 @@
      */
     _setObjectScale: function(localMouse, transform, lockScalingX, lockScalingY, by, lockScalingFlip, _dim) {
       var target = transform.target, forbidScalingX = false, forbidScalingY = false, scaled = false,
-          changeX, changeY;
+          changeX, changeY, scaleX, scaleY;
 
-      transform.newScaleX = localMouse.x * target.scaleX / _dim.x;
-      transform.newScaleY = localMouse.y * target.scaleY / _dim.y;
-      changeX = target.scaleX !== transform.newScaleX;
-      changeY = target.scaleY !== transform.newScaleY;
+      scaleX = localMouse.x * target.scaleX / _dim.x;
+      scaleY = localMouse.y * target.scaleY / _dim.y;
+      changeX = target.scaleX !== scaleX;
+      changeY = target.scaleY !== scaleY;
 
-      if (lockScalingFlip && transform.newScaleX <= 0 && transform.newScaleX < target.scaleX) {
+      if (lockScalingFlip && scaleX <= 0 && scaleX < target.scaleX) {
         forbidScalingX = true;
       }
 
-      if (lockScalingFlip && transform.newScaleY <= 0 && transform.newScaleY < target.scaleY) {
+      if (lockScalingFlip && scaleY <= 0 && scaleY < target.scaleY) {
         forbidScalingY = true;
       }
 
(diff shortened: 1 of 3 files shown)


asturur:
needs major rework.
Actually rotating, skew, scale moving fire even when the object is locked on those actions.

I will add a property to current transform to detect if something happened during transform.
target.Moving is good for that.

asturur:
I added some more booleans to our boolean salad that was in the scaling and skewing events.
Now we should be able to understand when a transformation modified the object or not and fire the proper event and object:modified.

A this point stateful could be defaulted to false, since saving and checking the object stateProperties is needed just for advanced features. But being a breaking change i will not change it now.

PR #2894, opened 2016-04-15
transform context active group

### LINKED ISSUE #532: activeGroup removeObject (shiftKey) did not work if perPixelTargetFind = true

asturur:
Back from when the issues were only 3 digits long.

If we transform canvas with the activeGroup matrix, we can properly detect the transparency of the pixels.

closes #532

@Kienz

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -312,8 +312,7 @@
 
       if (shouldTransform) {
         ctx.save();
-        var m = target.group.calcTransformMatrix();
-        ctx.transform.apply(ctx, m);
+        ctx.transform.apply(ctx, target.group.calcTransformMatrix());
       }
       target.render(ctx);
       target._renderControls(ctx);
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -304,20 +304,30 @@
      */
     isTargetTransparent: function (target, x, y) {
       var hasBorders = target.hasBorders,
-          transparentCorners = target.transparentCorners;
-
+          transparentCorners = target.transparentCorners,
+          ctx = this.contextCache,
+          shouldTransform = target.group && target.group === this.getActiveGroup();
       target.hasBorders = target.transparentCorners = false;
 
-      target.render(this.contextCache);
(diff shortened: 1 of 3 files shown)


asturur:
Added correct checking for perPixelTargetFind when hovering an activeGroup ( before it was just containsPoint )

asturur:
need  proper test in canvas unit.

PR #2908, opened 2016-04-26
Use sendToBack, bringToFront for multiple selections

### LINKED ISSUE #1301: Make possible to use stack operation with multiple selections
I was trying to keep a single object on top of everything on my project , and noticed that the group selection does not respect the "sendToFront()" of my object. After some experimenting i discovered (and you can do it in the kitchensink of fabric) that if you select 2 object as group with your mouse and type in the console "canvas.getActiveGroup().sendToBack()" the two objects duplicate with an offset and by selecting and moving one of them you actually move the cloned one too.
This bug aside , is it possible to have a specific object always on top of everything? As of right now i captured the events for everything on fabric (object:selected, moved,rotated, etc) and force my desired object on top using bringToFront() but it doesn't work with group. As i understand groups have other order of appearance than Objects.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1913807-group-selection-clones-itself-when-trying-to-execute-sendtoback-on-the-group?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
This allows to pass an active group to sendToBack and bringToFront and move all the objects mantaining theyr stack order.

to see a visual effect during selection you should use `canvas.preserveObjectStacking = true`;

Some additional work is required for kitchensink demo on farbcjs.com and bringFoward and sendBackwards methods.

closes #1301


asturur:
needs test

PR #2918, opened 2016-04-28
make IText fires object:modified on exiting:edit if text changes.

### LINKED ISSUE #2308: Modifying text does not dispatch object:modified event
I just realized that the itext never dispatches an object:modified event when the text is altered. What would be an appropriate place to fix this?

Currently, I am thinking of putting it inside exitEditing() (itext_behavior.mixin.js). Any other suggestions?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/23059761-modifying-text-does-not-dispatch-object-modified-event?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
@inssein reading your issue about text not fire object:modified in during editing, i made this draft.

Is a simple change that may add the missing feature.
The point is that object:modified fires normally at the end of transformation, and i felt like that end of editing was a nicer place for it.
Do you have some suggestion

i would add: 
deleting the `_textBeforeEditing `property at end of function. 
Do not fire if there is no `_textBeforeEditing `set on exitEditing.

@Kienz, @kangax any comment?

made to close #2308


src/mixins/itext_behavior.mixin.js
```diff
@@ -508,7 +509,11 @@
       this._currentCursorOpacity = 0;
 
       this.fire('editing:exited');
-      this.canvas && this.canvas.fire('text:editing:exited', { target: this });
+      isTextChanged && this.fire('modified');
+      if (this.canvas) {
+        this.canvas.fire('text:editing:exited', { target: this });
+        isTextChanged && this.canvas.fire('object:modified', { target: this });
```
inssein: this line will make it so object:modified gets fired twice if ```isTextChanged && this.canvas```

asturur: As any other event, it fires from the object and from the canvas. dev has the choice to register his listener wherever he prefers.

inssein: Ahh sorry, I thought I read ```canvas._fire```, which fires both canvas level and object level events.

asturur:
i lost the branch i will reopen the pr with tests added.

PR #2925, opened 2016-05-01
make Modifier keys customizable

### LINKED ISSUE #2680: Make it possible to select modifyer key
Currently the user are forced to use shift as the modifier key when selecting multiple objects as an example. It would be nice if this could be set in the canvas, for example canvas.set(modifierKey, "CRTL") or something.

asturur:
@VictorHagsand i made up this based on #2680.
No added code, just properties. It may be cool.

closes #2680

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -84,6 +84,7 @@
 
     /**
      * Indicates which key enable centered Transfrom
+     * values: altKey, shiftKey, ctrlKey
      * @since 1.6.2
      * @type String
      * @default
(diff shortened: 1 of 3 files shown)


VictorHagsand:
Very cool, thank you!
Saves us a bunch of overrides : )

asturur:
yes you have to override 5 functions to change all keys.

VictorHagsand:
Yupp, did that already before i made this suggestion here, thought that it might be useful for other people as a standard feature. I assume this will get merged later?

asturur:
i have to be sure it works good. did not try yet.

PR #2926, opened 2016-05-01
Image stroke

asturur:
I will post screenshot to check i did not broke anything, but looks like that we have some mess in the stroke rendering under certain conditions.

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -167,8 +167,13 @@
       if (!this.stroke || this.strokeWidth === 0) {
         return;
       }
+      var w = this.width / 2, h = this.height / 2;
       ctx.beginPath();
-      ctx.strokeRect(-this.width / 2, -this.height / 2, this.width, this.height);
+      ctx.moveTo(-w, -h);
+      ctx.lineTo(w, -h);
+      ctx.lineTo(w, h);
+      ctx.lineTo(-w, h);
+      ctx.lineTo(-w, -h);
       ctx.closePath();
     },
 
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -164,12 +164,12 @@
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
     _stroke: function(ctx) {
-      ctx.save();
-      this._setStrokeStyles(ctx);
+      if (!this.stroke || this.strokeWidth === 0) {
+        return;
+      }
       ctx.beginPath();
       ctx.strokeRect(-this.width / 2, -this.height / 2, this.width, this.height);
       ctx.closePath();
-      ctx.restore();
     },
 
     /**
@@ -410,6 +410,7 @@
                                      imageMargins.height
                                     );
 
+      this._stroke(ctx);
       this._renderStroke(ctx);
     },
 
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1194,6 +1194,11 @@
         var g = this.stroke.gradientTransform;
         ctx.transform.apply(ctx, g);
       }
+      if (this.stroke.toLive) {
+        ctx.translate(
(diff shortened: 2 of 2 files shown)


src/shapes/image.class.js
```diff
@@ -164,12 +164,12 @@
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
     _stroke: function(ctx) {
-      ctx.save();
-      this._setStrokeStyles(ctx);
```
asturur: _stroke is called by the _render function, _setStrokeStyle has been already done.

src/shapes/image.class.js
```diff
@@ -410,6 +410,7 @@
                                      imageMargins.height
                                     );
 
+      this._stroke(ctx);
```
asturur: Image class is the only one with the internal _stroke method. We can call it directly from its render method

src/shapes/object.class.js
```diff
@@ -1185,20 +1185,16 @@
         }
         if (supportsLineDash) {
           ctx.setLineDash(this.strokeDashArray);
-          this._stroke && this._stroke(ctx);
```
asturur: this has been done by the _render itselft

src/shapes/object.class.js
```diff
@@ -1185,20 +1185,16 @@
...
-        if (this.stroke.gradientTransform) {
-          var g = this.stroke.gradientTransform;
-          ctx.transform.apply(ctx, g);
-        }
-        this._stroke ? this._stroke(ctx) : ctx.stroke();
+      if (this.stroke.gradientTransform) {
+        var g = this.stroke.gradientTransform;
+        ctx.transform.apply(ctx, g);
       }
+      ctx.stroke();
```
asturur: we have to stroke anyway, so out of the IF

src/shapes/image.class.js
```diff
@@ -164,12 +164,17 @@
...
+        return;
+      }
+      var w = this.width / 2, h = this.height / 2;
       ctx.beginPath();
-      ctx.strokeRect(-this.width / 2, -this.height / 2, this.width, this.height);
+      ctx.moveTo(-w, -h);
+      ctx.lineTo(w, -h);
+      ctx.lineTo(w, h);
+      ctx.lineTo(-w, h);
+      ctx.lineTo(-w, -h);
```
asturur: This looks like an un-optimization, but in my opinion it simplify the rendering flow for the dev.

asturur:
The point is that every shape do its path in the _render function and then we call .fill(); and .stroke();

Image was the only one that was doing differently, it was drawing the eventual border inside the renderStroke with strokeRect function and then not doing the ctx.stroke();

I think that getting 4 lineTo instead of strokeRect is slower, but we get a cleaner pipeline.

asturur:
Also look this code:

```
// add red rectangle
var rect = new fabric.Rect({
  width: 50,
  height: 50,
  left: 50,
  top: 50,
});
rect.setGradient('fill', { type:'linear',
                        x1: 0, 
                        y1: 0, 
                        x2: 0, 
                        y2: rect.height,
                        colorStops: { 0 : 'rgba(0,255,0,1)', 1 : 'rgba(0,0,255,1)'} });

// add line
var line = new fabric.Line([10,10,10,60], {strokeWidth: 50});
line.setGradient('stroke', { type:'linear',
                        x1: 0, 
                        y1: 0, 
                        x2: 0, 
                        y2: rect.height,
                        colorStops: { 0 : 'rgba(0,255,0,1)', 1 : 'rgba(0,0,255,1)'} });
canvas.add(rect);
canvas.add(line);
```

look results:

image

left rectangle, right a chubby line. We are not translating the gradient for stroke ( and so patterns!!!)

asturur:
new SVG IMPORT 
image

real SVG

image


current SVG IMPORT

image

PR #2932, opened 2016-05-03
Control customization part 1 - Set dashed line to controls and borders

asturur:
Give controls and borders chance to be dashed trought

added:
- cornerDashArray
- borderDashArray
- borderScaleFactor influences both border and controls now
- changed default corner size to 13:
(a default size of 12 is always out of focus when borders are on focus and viceversa, at least now they are same. difference is not noticeable because the 1st and 12th pixel where always across 2 different pixels)


asturur:
example
image

PR #2933, opened 2016-05-03
Control customization part 2 - Allow for separate corner filling and cornerStroke

asturur:
- add CornerStrokeColor property
- default behaviour with transparentCorners does not change ( for people loading json created before or overriding classes with just cornerColor and transparentCorner = true );

### DIFF
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -362,7 +362,7 @@
       var size = this.cornerSize;
       isVML() || this.transparentCorners || ctx.clearRect(left, top, size, size);
       ctx[methodName](left, top, size, size);
-      if (!this.transparentCorners) {
+      if (!this.transparentCorners && this.cornerStrokeColor) {
         ctx.strokeRect(left, top, size, size);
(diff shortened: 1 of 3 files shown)


asturur:
image

taqimustafa:
Is it possible to have the option to change the rotation handle position, The default position is top, but if we want to show it at the bottom of the Object ?

asturur:
```javascript
       if (this.hasRotatingPoint) {
         this._drawControl('mtr', ctx, methodName,
           left + width / 2,
           top + height + this.rotatingPointOffset);
       }
```

you should override the function that has this snippet with another that is same content but with the hasRotatingPoint part changed like i did

taqimustafa:
Also can we make the borders dashed ? something similar to the attached image ? 
screenshot_35

taqimustafa:
i found a way by adding `ctx.setLineDash([3,2]);` in drawBorders function, but i was hoping if there is anyway we can do that directly ?

asturur:
http://fabricjs.com/fabric-changelog

check borderDashArray

PR #2935, opened 2016-05-03
Asturur patch 1

### LINKED ISSUE #2934: Error in createSVGFontFacesMarkup function - char is a reserved keyword
Hello,

createSVGFontFacesMarkup uses a variable named "char", that's a reserved keyword and yui compressor in sencha touch doesn't allow to compress the app source. 

`var markup = '', fontList = { }, obj, fontFamily,
style, row, rowIndex, char, charIndex,
fontPaths = fabric.fontPaths;`

The variable name should be changed to something that's not reserved. 
http://www.w3schools.com/js/js_reserved.asp

Thanx

asturur:
rename an internal variable, fixed a bug.. var row was not initialized.

closes #2934 by @haloween

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1031,7 +1031,7 @@
      */
     createSVGFontFacesMarkup: function(objects) {
       var markup = '', fontList = { }, obj, fontFamily,
-          style, row, rowIndex, char, charIndex,
+          style, row, rowIndex, _char, charIndex,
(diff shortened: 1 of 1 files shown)


haloween:
fontFamily = _char.fontFamily; ?

asturur:
Yes. for the same reason i did the first error, here was the second.
Thanks god test would stop me this time, just you came first.

PR #2942, opened 2016-05-05
Control cusomization part 3 - allow for rounded corners

asturur:
Rounded corner have some inconsistency:
They work better on solid color and with a little bit of padding.
Needs some thinkering


asturur:
image

PR #2943, opened 2016-05-05
fix itext with caching false

### LINKED ISSUE #2937: Text dissapear from a canvas when caching property is false
## Version
1.6.1

## Test Case
http://jsfiddle.net/Da7SP/70/

## Steps to reproduce
Create an i-text object and set the "caching" property to false
## Expected Behavior
Text dissapears from text box
## Actual Behavior
Text renders correctly

asturur:
closes #2937


asturur:
We left a dead condition in wich we were returning a stored value that was calculated if caching was disabled but stored if caching was enabled.

PR #2944, opened 2016-05-05
Fix i-text caching

m2broth:
Fix i-text char width caching

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -190,6 +190,7 @@
     _clearCache: function() {
       this.callSuper('_clearCache');
       this.__widthOfSpace = [ ];
+      this._charWidthsCache = [ ];
     },
 
     /**


asturur:
This is not correct.
The cache we cleare with clearCache is the cache for the itext lineWidth and lineHeight.

Clearing the characterCache there would be the same as disabling caching completely every text or font change.

If you have will and time, it was an idea waiting to someone to do it, to move the character cache outside the fabric.Itext instance and move it to fabric.

Infact when you have 10 iText on the canvas, all character "A" of Arial 40px are big the same, but we calculate the cache for every of them and store the cache for any of them

If you wanto to submit a pr for this change, that would be nice.

Maybe creating a cache at fabric level, and just put a reference of the cache in the class is enough.

m2broth:
Thanks, I like your solution about separate cache class

m2broth:
One moment.

The lineWidths cache array depends on charWidthsCache. When we change font family in one i-text instance we should to clear cached widths for this instance to correctly calculate i-text object width

m2broth:
And if charWidthsCache has been changed we should to change the lineWidths cache.

asturur:
charsWidth cache does not contains width of "A", "B".
Is a complex object that contains charswidth[propcache] = xxx

propcache is:
```javascript
    _getCacheProp: function(_char, styleDeclaration) {
      return _char +
             styleDeclaration.fontFamily +
             styleDeclaration.fontSize +
             styleDeclaration.fontWeight +
             styleDeclaration.fontStyle +
             styleDeclaration.shadow;
    },
```

So really the cache has so many keys and gets never deleted. That is why would be nice to have at library level and not per objectlevel.
(i think that carrying shadow in this declaration is not usefull anyway, shadow is under the char and does not contribute to its width as a rect shadow does not contribute to rect's width)

asturur:
lineWidths cache get deleted on every change, it just avoid us to poll
charWidthsCache every time.
When we drag around or scale an object we use lineWithsCache, when we
change text, or fontSize or style we delete the lineWidths cache and
recalculate the width with the new charWidths.

2016-05-06 6:52 GMT+02:00 m2broth <notifications@github.com>:

PR #2980, opened 2016-05-18
Apply viewport transform when calculating isTargetTransparent()

rockmacaca:
This fixes a bug with how an interactive canvas with perPixelTargetFind behaves after being panned/zoomed.

## Version
1.6.2

## Test Case
http://jsfiddle.net/d10y74q7/5/

## Steps to reproduce
1. Turn perPixelTargetFind on for a Canvas instance
2. Modify the viewport of the canvas (pan or zoom)
3. Attempt to select objects of the interactive canvas

## Expected Behavior
Cursor should change to the move cursor when over opaque pixels, and clicking & dragging should move items.

## Actual Behavior
Objects can only be interacted with when the XY coordinates of the pointer are coincidentally over opaque pixels in the canvas with an unmodified viewport.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -346,11 +346,14 @@
 
       target.hasBorders = target.transparentCorners = false;
 
+      ctx.save();
+      ctx.transform.apply(ctx, this.viewportTransform);
       if (shouldTransform) {
-        ctx.save();
         ctx.transform.apply(ctx, target.group.calcTransformMatrix());
       }
       target.render(ctx);
+      ctx.restore();
+
       target.active && target._renderControls(ctx);
 
       target.hasBorders = hasBorders;
@@ -358,7 +361,6 @@
 
       var isTransparent = fabric.util.isTransparent(
         ctx, x, y, this.targetFindTolerance);
-      shouldTransform && ctx.restore();
 
       this.clearContext(ctx);
 


rockmacaca:
Ok, I just discovered that the fix is still incomplete with my commit as the resize handles are still not selectable after panning/zooming. I'll see if I can figure out what's wrong.

asturur:
i'm guessing if is better to transform the coordinate of mouse or the context from a performance point of view.
I added that shouldTransform because i did not want to save restore and transform on every mouse move when isTargetTransparent is used.

rockmacaca:
I suppose the only optimization that could be made there is to check that viewportTransform hasn't deviated from its default setting before we start saving or restoring the context. However it should be expected that there be some performance hit when we have perPixelTargetFind turned on.

What do you suggest should be done to get this resolved?

rockmacaca:
Keep in mind, this transform is already being done when _renderControls(ctx) is called. If you wanted, we could just have _renderControls() ignore the viewport and have the caller take care of that, like it currently is in renderAll() calling render():

```javascript
    renderAll: function () {
      ...
      canvasToDrawOn.save();
      objsToRender = this._chooseObjectsToRender();
      //apply viewport transform once for all rendering process
      canvasToDrawOn.transform.apply(canvasToDrawOn, this.viewportTransform);
      this._renderObjects(canvasToDrawOn, objsToRender);
      ...
    }
```

asturur:
no renderControls can't be touched easily. Jsut guessing if is lighter to tranform a context or a mouse coordinates.

rockmacaca:
Ok. I'm not sure transforming mouse coordinates would be a proper fix though. I updated the jsfiddle to better illustrate the issue. The problem with transforming mouse coordinates is that when your transformed position leaves the original viewport, it is seen as transparent since the cache canvas has not rendered that area.

asturur:
I'm getting a mac next week. waiting for it so i test this + retina screen and i will merge or ask you for tweak.

rockmacaca:
Any updates?

By the way, you can set the device pixel ratio with Chrome's dev tools or just by setting the browser's zoom level before the page loads. I've tested that to all work fine.

asturur:
got mac yesterday. I know i can test with device pixel ration ( also with browser zoom sadly, that is the problem ). I need some days to get back on the issues and prs.

rockmacaca:
Bumping. Have you had time to test this recently?

PR #2986, opened 2016-05-19
Fix IText rendering to take viewportTransform into account

fenderle:
When having a viewportTransform on the canvas IText fails to correctly render the cursor and the hidden area since it doesn't take the viewportTransform into account.

This patch fixes this, but I'm not entirely sure if I took the right approach.

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -430,8 +430,8 @@
       this.hiddenTextarea.selectionEnd = this.selectionEnd;
       if (this.selectionStart === this.selectionEnd) {
         var p = this._calcTextareaPosition();
-        this.hiddenTextarea.style.left = (p.x * this.canvas.viewportTransform[0]) + 'px';
-        this.hiddenTextarea.style.top = (p.y * this.canvas.viewportTransform[3]) + 'px';
+        this.hiddenTextarea.style.left = p.x + 'px';
+        this.hiddenTextarea.style.top = p.y + 'px';
       }
     },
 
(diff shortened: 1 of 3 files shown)


src/shapes/itext.class.js
```diff
@@ -497,7 +497,7 @@
       ctx.fillRect(
         boundaries.left + leftOffset,
         boundaries.top + boundaries.topOffset,
-        this.cursorWidth / this.scaleX,
+        this.cursorWidth / this.scaleX / this.canvas.viewportTransform[0],
```
asturur: Before fillRect, calculate a scale factor that take in account scaleX and this.canvas.getZoom();
var scaleFactor = this.scaleX;
if (this.canvas) {
  scaleFactor *= this.canvas.getZoom();
}

and then use that scaleFactor to divide the cursorWidth.

src/mixins/itext_behavior.mixin.js
```diff
@@ -430,8 +430,8 @@
       this.hiddenTextarea.selectionEnd = this.selectionEnd;
       if (this.selectionStart === this.selectionEnd) {
         var p = this._calcTextareaPosition();
```
asturur: modify _calcTextareaPosition to take in account the viewporttransform.

before returning the point, use fabric.util.transformPoint(p, this.getViewportTransform());

so then we can use this modified function to take in account also the other case.

asturur:
it misses translation but it was sometihng in the list of things to do. I think it misses also retina control.

fenderle:
I applied your comments - hopefully I understood them correctly...

asturur:
I'm getting a mac next week. waiting for it so i test this + retina screen and i will merge or ask you for tweak.

asturur:
@fenderle i m working on a project where i need this to be fixed asap. Do you mind if i take over your work and finish it? i have to do it today.

PR #2987, opened 2016-05-19
Fix _isTargetTransparent to honor viewportTransform

fenderle:
When a viewportTransform is set the transparency detection does not work since it does not respect the viewportTransform when rendering the target onto the cache canvas.

This patch fixes this, but I'm not entirely sure if I took the right approach.


asturur:
duplicate of #2980.
i have to give a look to both. I'm waiting a mac to try them properly because we have to check the retina scaling also.

PR #2989, opened 2016-05-20
Object caching

### LINKED ISSUE #318: Using pre-rendering techniques [$25]
Currently the _draw implementation of static canvas will always do a full render of each object - no matter what has been modified. The HTML5 Canvas best practices suggest to render an object only once if you have a lot of render passes and then just draw the bitmap representation if needed - this is equal to what flash does using the cacheAsBitmap property and it's display list .

Without caching:

```
// canvas, context are defined
function render() {
  drawMario(context);
  requestAnimationFrame(render);
}
```

Using caching:

```
var m_canvas = document.createElement('canvas');
m_canvas.width = 64;
m_canvas.height = 64;
var m_context = m_canvas.getContext(‘2d’);
drawMario(m_context);

function render() {
  context.drawImage(m_canvas, 0, 0);
  requestAnimationFrame(render);
}
```

For Fabric this would mean that the objects manage a property to tell if a redraw is really needed - this could be bound to the stateProperties. Then the static_canvas will use this property to call render or not. After rendering the bitmap will be cached and reused on the next calls. 

Maybe I just did not inspect the code good enough, is fabric maybe already doing this?

See:

http://www.html5rocks.com/en/tutorials/canvas/performance/
## <bountysource-plugin>

There is a **[$25 open bounty](https://www.bountysource.com/issues/55981-using-pre-rendering-techniques-25?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #814: Feature request - Object caching
Fabric.js is in my opinion very well written with excellent overall performance.

However there are occasions where the performance is drastically reduced due to performance limitations of the html5 canvas API and not down to fabric itself.

Two of these occasions are:

- When manipulating a text object using a very complex font or with a textShadow applied
- When manipulating objects with shadows turned on (particularly with a large blur radius)

Having gotten quite familiar with the source code over the last few weeks I can see a very clear way of rectifying these performance issues while dramatically increasing overall manipulation performance across the board.

My proposal is to cache objects (on demand, not continuously) as off screen canvas's so that during manipulations (moving, rotating etc) the object is drawn to the canvas using drawImage() rather then being completely redrawn.

The flow as I see it would go as follows:

- when mousedown occurs on an object an offscreen canvas is temporarily created

- the temporary canvas is set to the size and width of the object and the object is rendered to this canvas instead of the main lower one

- during manipulations the the _render() is halted and drawImage(), using the temporary canvas as the source element, is called instead.

- on mouseup the temporary canvas is nulled to avoid creating excessive in memory canvases.

I have been hacking away at the fabric source code and have implemented this myself.  For text objects that use extremely complex fonts or ones with text shadows etc, the performance increase is from about 10fps up to 60fps.  The same goes for image objects with shadows set.  The visual result is identical to rendering the objects each frame, just much faster.

Everything works beautifully for moving and rotating however my implementation currently leaves an issue with scaling.  Since we are dealing with a canvas element the quality changes as you scale, then on mouseup is re-rendered correctly.  This is mostly only an issue when scaling up, but for text it is also heavily noticable when scaling down.

In order to get around this some sort of throttle or timer could be used to rebuild the cached canvas at the new size in the background.  Hopefully leaving the visual difference almost imperceptible.  So far I haven't managed to get around to attacking this yet.

Implementing the above is actually relatively simple, but with my lack of knowledge of Fabric's inner workings (particularly around the scaling/rendering) means that my code is perhaps more complex than it needs to be, hence my leaving it as a feature request rather than making a pull request.

Also... Another huge benefit of using the above method mean it solves a large bug in Chrome on Windows....

Currently in Chrome on windows when rotating a text object the letter jiggle continuously.  I assume this is down to Chrome not properly implementing sub-pixel rendering on canvas text objects.  Using the above caching method this jiggling completely disappears.

I would be interested to hear your thoughts Kangax.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/869428-feature-request-object-caching?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
Still lot of work to do, basically working.

Object get rendered from canvas cache copy if objectCaching is true.
For usability purpouse iText automatically disable and rebuild cache on edit enter - exit
Speed improvements: noticeable just with big loads ( 200 complicated iText, you go from heavy shutter to smooth)
Cache copy does not include shadow and it will not for now.

TODO:
fix group,
decide what to do with automatic cache clearing, if use object statefull properties ( a subset of them ) to autoclear cache.
implement scaling
possible tests about the caching canvas dimensions

DONE:
pathgroups
fix polygon
do not cache images.

closes #814 
closes #318

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -341,7 +341,8 @@
       if (this.isEditing || !this.editable) {
         return;
       }
-
+      this._originalCaching = this.objectCaching;
+      this.objectCaching = false;
       if (this.canvas) {
         this.exitEditingOnOthers(this.canvas);
       }
@@ -507,7 +508,8 @@
       this.abortCursorAnimation();
       this._restoreEditingProps();
       this._currentCursorOpacity = 0;
-
+      this.objectCaching = this._originalCaching;
+      this.objectCaching && this.refreshCache();
       this.fire('editing:exited');
       isTextChanged && this.fire('modified');
       if (this.canvas) {
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -89,6 +89,14 @@
     _lastScaleY: 1,
 
     /**
+     * enable object caching
+     * since 1.6.3
+     * @type Boolean
+     * @default
+     */
+    objectCaching: false,
+
+    /**
      * Constructor
      * @param {HTMLImageElement | String} element Image element
      * @param {Object} [options] Options object
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -826,7 +826,11 @@
       this.cacheContext.scale(ratio, ratio);
     },
 
-    refreshCache: function() {
+    /**
+     * redraw the cache and update the cacheCanvas by request
+     * @param {Boolean} [noTransform] for pathgroup rendering purpouse, leave it to false.
+     */
+    refreshCache: function(noTransform) {
       var ctx = this.cacheContext,
           dim = this._getNonTransformedDimensions(),
           width = dim.x, height = dim.y;
@@ -837,7 +841,7 @@
         ctx.clearRect(-width / 2, -height / 2, width, height);
       }
       ctx.globalAlpha = 1;
-      this._cachingNeeds && this._cachingNeeds(ctx);
+      this._cachingNeeds && this._cachingNeeds(ctx, noTransform);
       this._draw(ctx);
       this.isCacheDirty = false;
     },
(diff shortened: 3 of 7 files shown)


asturur:
@plainview

PR #2729, opened 2016-01-10
Events in grouped object

asturur:
Preliminary cleaning and test.

First i remove some duplicate calls of getPointer(e, true); We are passing event e, and in every function searching for the pointer from the event. While we can search pointer once and then just pass reference to it.

Then a nice logic to discern when we should go deep in the groups and when we should stop at first object is still needed.

PR #2997, opened 2016-05-22
Target in group

### LINKED ISSUE #485: Support events for objects in group [$35]
http://jsfiddle.net/citykid/NSD4P/

the fiddle creates 3 rect objects

component
  body
  port

events for the inner rects, body and port do not fire.

<bountysource-plugin>
---
There is a **[$35 open bounty](https://www.bountysource.com/issues/138341-support-events-for-objects-in-group-35?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2998: Multiselection with editing itext
## Version
1.6.2

## Test Case
no fiddle needed

## Steps to reproduce
go on kitchensink, select on Itext, go on edit mode. With shift click select another object, group is created. destroy active group.
## Expected Behavior
Itext has controls
## Actual Behavior
iText lost controls

When creating activeGroup iText has to exit editing mode.

### LINKED ISSUE #3005: Polygon perPixelTargetFind
## Version
1.6.2

## Test Case
https://jsfiddle.net/5t4mfu7r/6/

I create 2 Polygons, one is transparent, one isn't (the transparent one is left to the other). Both have "perPixelTargetFind" true. 

The Polygon should just be selected, when the click is located inside the object

Both get selected if you click outside of it (but inside its bounding box)

asturur:
updated version of #2729

closes #485
closes #2998 
closes #3005

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -947,7 +947,7 @@
       }
 
       var pointer = this.getPointer(e, true);
-      this.targets = [ ]
+      this.targets = [ ];
 
       if (this._isLastRenderedObject(pointer)) {
         return this.lastRenderedWithControls;
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -295,7 +295,7 @@
      * Checks if point is contained within an area of given object
      * @param {Event} e Event object
      * @param {fabric.Object} target Object to test against
-     * @param {Object} point x,y object of point coordinates we want to check.
+     * @param {Object} [point] x,y object of point coordinates we want to check.
      * @return {Boolean} true if point is contained within an area of given object
      */
     containsPoint: function (e, target, point) {
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -946,20 +946,24 @@
         return;
       }
 
-      var pointer = this.getPointer(e, true);
-      this.targets = [ ];
-
-      if (this._isLastRenderedObject(pointer)) {
-        return this.lastRenderedWithControls;
-      }
-
       // first check current group (if one exists)
+      // avtive group does not check sub targets like normal groups.
+      // if active group just exits.
       var activeGroup = this.getActiveGroup();
-      if (!skipGroup && this._checkTarget(pointer, activeGroup)) {
+      if (activeGroup && !skipGroup && this._checkTarget(pointer, activeGroup)) {
         return activeGroup;
       }
 
-      var target = this._searchPossibleTargets(this._objects, pointer);
+	  
+      var pointer = this.getPointer(e, true),
+          objects = this._objects;
+      this.targets = [ ];
+
+      if (this._isLastRenderedObject(pointer)) {
+        objects = [this.lastRenderedWithControls];
+      }
+
+      var target = this._searchPossibleTargets(objects, pointer);
       this._fireOverOutEvents(target, e);
       return target;
     },
@@ -1020,7 +1024,7 @@
           target = objects[i];
           if (target.type === 'group' && target.subTargetCheck) {
             normalizedPointer = this._normalizePointer(target, pointer);
-            subTarget = this._searchPossibleTargets(normalizedPointer, target._objects);
+            subTarget = this._searchPossibleTargets(target._objects, normalizedPointer);
             subTarget && this.targets.push(subTarget);
           }
           break;
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -954,7 +954,6 @@
         return activeGroup;
       }
 
-	  
       var pointer = this.getPointer(e, true),
           objects = this._objects;
       this.targets = [ ];
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -295,12 +295,19 @@
      * Checks if point is contained within an area of given object
      * @param {Event} e Event object
      * @param {fabric.Object} target Object to test against
+     * @param {Object} point x,y object of point coordinates we want to check.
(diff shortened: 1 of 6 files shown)


asturur:
closes #485

asturur:
Needs:
- TESTS
- add mouseover mouseout for subtargets (maybe separate PR)
- regression checks
- decide if passing also subtargets to events on canvas (maybe yes don't want to spam too much events)
- check strange behaviour with ITEXT ( it gets editables in active group)

asturur:
mouseover mouseout for subtargets will be made in a separate PR.
This pr has mouse up, mouse down, mouse move.

There should be no problem implementing something like that, but becase now we use _hoveredTarget and we must change to a chain of hoverd targets, i need to take some time to choose best way.

src/canvas.class.js
```diff
@@ -1007,22 +1012,23 @@
...
+      var target, i = objects.length, normalizedPointer, subTarget;
       // Do not check for currently grouped objects, since we check the parent group itself.
       // untill we call this function specifically to search inside the activeGroup
       while (i--) {
-        if ((!this._objects[i].group || skipGroup) && this._checkTarget(e, this._objects[i], pointer)){
-          this.relatedTarget = this._objects[i];
-          target = this._objects[i];
+        if (this._checkTarget(pointer, objects[i])) {
+          target = objects[i];
+          if (target.type === 'group' && target.subTargetCheck) {
```
samuelhorwitz: In order to better support subclassing, this would make more sense as `target instanceof fabric.Group`

src/canvas.class.js
```diff
@@ -1007,22 +1012,23 @@
...
+            normalizedPointer = this._normalizePointer(target, pointer);
+            subTarget = this._searchPossibleTargets(target._objects, normalizedPointer);
+            subTarget && this.targets.push(subTarget);
```
samuelhorwitz: What's the reason for moving over to this `subTarget` thing as opposed to just pushing each `target` onto the `this.targets` array? The old way from the old PR guaranteed that you would have a `targets` array that was in order of grouping depth with index 0 being the "real" target and the final index being the deepest non-group object with a collection of groups filling in the space between.

asturur: There is no really a reason. I approached the change with small steps, and i wanted to leave the code as it was.
For now i do not have all the events for subtargets and i did not want to have a duplicate target between main target and targets array. I'm still reviewing it, so any comment is appreciated.

samuelhorwitz: Cool :) Like I said in the other PR, I'm working off your changes to implement grouped object selection and it's definitely useful for me in that case to have `this.targets` be ordered. But I've made that change on my fork, so once the dust is settled it can be re-reviewed with all of that in mind. Just was wondering if there was anything else I was missing

src/mixins/canvas_events.mixin.js
```diff
@@ -285,23 +285,37 @@
...
+    /**
+     * Handle event firing for target and subtargets
+     * @param {Event} e event from mouse
+     * @param {String} eventType event to fire (up, down or move)
+     * @param {fabric.Object} targetObj receiving event
+     */
+    _handleEvent: function(e, eventType, targetObj) {
+      var target = targetObj || this.findTarget(e),
+          targets = this.targets,
+          options = { e: e, target: target, subTargets: targets };
```
asturur: regarding other comment: they are still ordered anyway and fired in order, and the array is passed to canvas events like options.subtargets
options = { e: e, target: target, subTargets: targets };
So in subtargets you have the ordered targets but not the first one that is target.

samuelhorwitz: Ah you know what. I think they are reversed order. So index 0 is the deepest object. That combined with the top level object missing made me just not look too closely and assume order was no longer guaranteed. My bad.

samuelhorwitz:
Does Fabric expect nested grouping? `addWithUpdate` inside of a nested group breaks. I'm looking into that now on my branch also but is that a regression or a known state of affairs with regards to groups?

asturur:
Groups can be nested. If not we fix it.
are you able to reproduce the bug with a small fiddle?

samuelhorwitz:
Yeah, here is a code pen.

http://codepen.io/anon/pen/bpXyQM

Sorry for cluttering this issue, it came up with my work off this PR but then I realized it was actually going back to Fabric before any of these changes.

asturur:
Nesting group will be topic of another PR. I need to keep changes small, independent, revertable, understandable.

Does the subtargeting work in group of groups?

samuelhorwitz:
Yes, subtargeting works fine in groups of groups

NikitaPirat:
Hi, 
thanks a lot for your work! Could you say when will be the next release? I hope this fix will be there.

Thanks

helia19s:
the subTargetCheck feature only works for Groups not pathgroup. I tried to fix it my self but the problem is the path position in a group is incorrect when we have scale and angle . will it get fixed in next releases ?

asturur:
no. pathgroups are completely different, are not meant for group/ungrouping, they are `harder` to handle due to transform matrix.
There is a PR open, blocked from some rounding issue in positioning that aims at removing path-groups to have just groups.
I'm not expanding any path-group feature since the idea is to remove them.

PR #3006, opened 2016-05-23
remove jsdom require if building in certain condition (webpack)

asturur:


born2net:
what was the solution as I am building using webpack and need to not install the whole jsom / canvas?
tx
Sean

asturur:
put fabric.js in a dep folder and tell webpack is a dependency.

born2net:
ok tx, I hope the fabric team will remove the optionalDeps

PR #3009, opened 2016-05-25
Grouped object selection/manipulation

samuelhorwitz:
This pull requests allows for objects inside of groups to be selected without ungrouping.

It is currently in a working state, but with some roadblocks and is not currently presumed fit for final review, but instead is currently more a request for comments.

Tests also still need to be written/updated.

It also branches off of the yet unmerged pull request #2997 (this is necessary due to that PR containing the groundwork for this functionality). I will rebase this branch once that PR is merged into master.

The UI used is very straightforward and workable with precedent from the application Sketch. It relies on double clicking. Double clicking drills down to the object that was double clicked on as well as making it so everything at that level of nesting and above can now be freely selected with single clicks. Leaving this click context (by selecting a different higher level element) resets the single click stuff. This is easier played with than reading about; try it out and use double clicking to select grouped and nested group objects.

@asturur I'm going to move all my #2997-off-topic questions here to stop cluttering your thread :)

# Shortcomings

There is one major shortcoming. The selection/active group setting of non-contiguous objects and objects already within permanent groups. It currently fails and bugs out for obvious reasons. I'm going to start work on this, however I'm still trying to think of the best route. Suggestions are welcome.

One question I have is is the point of `preserveObjectStacking`? It seems like it would always be desirable to have it set to true. Why would someone ever have it false? Having it false also makes the above shortcoming even more complicated.

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -87,6 +87,9 @@
      * @return {Boolean} true if point is inside the object
      */
     containsPoint: function(point) {
+      if (!this.oCoords) {
+        this.setCoords();
+      }
       var lines = this._getImageLines(this.oCoords),
           xPoints = this._findCrossPoints(point, lines);
 
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -292,10 +292,17 @@
      * @chainable
      */
     setCoords: function() {
+      this._setCoords();
+      this._setCoords(true);
+      return this;
+    },
+
+    _setCoords: function(normalized) {
       var theta = degreesToRadians(this.angle),
           vpt = this.getViewportTransform(),
           dim = this._calculateCurrentDimensions(),
-          currentWidth = dim.x, currentHeight = dim.y;
+          currentWidth = dim.x, currentHeight = dim.y,
+          oCoords;
 
       // If width is negative, make postive. Fixes path selection issue
       if (currentWidth < 0) {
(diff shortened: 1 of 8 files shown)


samuelhorwitz:
Here is a good test Javascript snippet.

```
var rect = new fabric.Rect({
    top : 100,
    left : 100,
    width : 60,
    height : 70,
    fill : 'red'
});

var rect2 = new fabric.Rect({
    top : 200,
    left : 200,
    width : 60,
    height : 70,
    fill : 'blue'
});

var rect3 = new fabric.Rect({
    top : 300,
    left : 300,
    width : 60,
    height : 70,
    fill : 'cyan'
});

var rect4 = new fabric.Rect({
    top : 400,
    left : 400,
    width : 60,
    height : 70,
    fill : 'pink'
});

var rect5 = new fabric.Rect({
    top : 500,
    left : 500,
    width : 60,
    height : 70,
    fill : '#ff00ff'
});

var group2 = new fabric.Group([rect3, rect4], {subTargetCheck: true});

var group = new fabric.Group([rect, rect2, group2], {subTargetCheck: true});

canvas.add(group);
```

src/mixins/object_geometry.mixin.js
```diff
@@ -341,10 +351,15 @@
         mtr: mtr
       };
 
-      // set coordinates of the draggable boxes in the corners used to scale/rotate the image
-      this._setCornerCoords && this._setCornerCoords();
+      if (normalized) {
+        this.oCoordsNormal = oCoords;
```
samuelhorwitz: One of the major changes I had to do was store normalized `oCoords` as well as group adjusted ones. Perhaps we don't need the group adjusted ones? I tried to take it out and it broke and so I didn't look too hard into it. But I could have been missing something.

src/shapes/group.class.js
```diff
@@ -146,25 +136,68 @@
...
+    /**
+     * Inserts an object into collection at specified index; Then recalculates group's dimension, position.
+     * @param {Object} object Object to insert
+     * @param {Number} index Index to insert object at
+     * @param {Boolean} nonSplicing When `true`, no splicing (shifting) of objects occurs
+     * @return {fabric.Group} thisArg
+     * @chainable
+     */
+    insertWithUpdate: function(object, index, nonSplicing) {
+      if (this.group) {
```
samuelhorwitz: All this parent group checking and index maintaining is a workaround for the issue I described here: https://github.com/kangax/fabric.js/pull/2997#issuecomment-221310801

This solves that problem but probably in an inelegant way. It can be removed when a better fix is found.

asturur:
I will open an issue about that and discuss how the library should behave, what functionality we want to give to library and what are to be left at application level. I'm going to check the code anyway

asturur:
Please reopen the pull request as pull request on my branch and not on master, so we see the differences better.

samuelhorwitz:
Done #3014

PR #3016, opened 2016-05-29
update area clearing during rendering

asturur:
Meaning of this pull request is:
Compact code of initDelayedCursor that was mainly abortCursor + start New animation,
Clear area of text and a little padding, i noticed that when rendering a selection, a small border was not cleared
Animate just the cursor and not also the selection (this is the main thing)


asturur:
on enter edit we where adding some event to mouse up that we were not deregistering on editing:exited.

asturur:
needs unit testing for the event handler removal plus better checking for more events registered as this one.

PR #3017, opened 2016-05-30
Add additional methods for centering objects

### LINKED ISSUE #3015: Canvas zoom/position bug [$5]
## Version
1.6.2

## Test Case
https://jsfiddle.net/qkj8o6eh/1/

## Expected Behavior
When I initially set the zoom of my canvas (along with the canvas size and background image size)  to 1, i add a rectangle. This rectangle acts as a marker for the center of our canvas. When we scale the canvas (and also the background image and canvas size respectfully) I would expect the center of the canvas to remain the same, regardless of the zoom level. This expected result would be the two rectangles centered on the canvas.

## Actual Behavior
Instead, what we see is the position of the second rectangle differs when we zoom the canvas to something other than the initial value "1".

## NOTES:
Keep in mind, this might also effect other align functions, (i.e. `centerObjectH`, etc.) other than `object.center()`.

https://api.bountysource.com/badge/issue?issue_id=34629617

asturur:
Just a draft,
needs unit tests, same method at object level.

closes #3015

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -975,32 +975,28 @@
     },
 
     /**
-     * Centers object horizontally.
+     * Centers object horizontally in the canvas
      * You might need to call `setCoords` on an object after centering, to update controls area.
      * @param {fabric.Object} object Object to center horizontally
      * @return {fabric.Canvas} thisArg
      */
     centerObjectH: function (object) {
-      this._centerObject(object, new fabric.Point(this.getCenter().left, object.getCenterPoint().y));
-      this.renderAll();
-      return this;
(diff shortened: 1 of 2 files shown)


FeaRCODE:
Will this make it in 1.6.3?

asturur:
yes, trying to finish it with tests.

FeaRCODE:
@asturur Great! Will this also fix the problem I was having when centering the object with it's left and top? I think there was an issue with the coords in general.

asturur:
What do you mean? can yuo point me to the problem?

asturur:
it looks like i m not able to write tests for this PR. I need holyday.

PR #3019, opened 2016-05-30
introduce a parameter to choose behaviour of background and overlay

### LINKED ISSUE #2946: Zoom and pan using canvas.setZoom() should affect canvas background image
## Version
1.6.1

## Test Case
1.6.1 (Actual):  https://jsfiddle.net/mL58b3rr/10/
1.6.0-rc.1 (Expected) : https://jsfiddle.net/1c6n1bry/3/
(They are both the exact same code, the only difference is the library`s version)

## Steps to reproduce
Try Zooming in and out

## Expected Behavior
Prior to version 1.6.0 (1.6.0-rc1), when setting the canvas zoom, through canvas.setZoom(), would also set the zoom of the background image. 1.6.0 and on, only the elements are being zoomed / panned

## Actual Behavior
the canvas' Background image is static and don't change on canvas.setZoom(), in versions 1.6.0+

### LINKED ISSUE #3007: v 1.6.2 toDataURL with multiplier does not scale background image
## Version
1.6.2

## Test Case
http://jsfiddle.net/jp8rL7pg/

## Steps to reproduce
click "render preview" button

## Expected Behavior
Canvas background image should be scaled up.
Image with red rectange on blue background.

## Actual Behavior
Background image is not scaled. I see green canvas background color.

asturur:
closes #2946
closes #3007

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -938,12 +938,12 @@
       }
       object = this[property + 'Image'];
       if (object) {
-        if (this.[property + 'Vpt']) {
+        if (this[property + 'Vpt']) {
           ctx.save();
           ctx.transform.apply(ctx, this.viewportTransform);
         }
         object.render(ctx);
-        this.[property + 'Vpt'] && ctx.restore();
+        this[property + 'Vpt'] && ctx.restore();
       }
     },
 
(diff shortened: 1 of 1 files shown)


FeaRCODE:
@asturur Does this allow movement of the backgroundImage with Pan? For instance, if I pan to the right, all the contents of the canvas (including the backgroundImage) will move to the right? If not, how hard would this be to implement?

asturur:
yes this was make to zoom and pan the background. viewportTransform represent both zoom and pan togheter.

FeaRCODE:
That is awesome! I cant wait till 1.6.3 :D :tada:

PR #3023, opened 2016-06-01
Remove reference to lastRenderedWithControls when that object is deleted

plainview:
This fixes an issue where a user could remove a textbox from the canvas, and then click in the space where the textbox was, and it would re-appear.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1361,6 +1361,17 @@
         this._objects[i]._renderControls(ctx);
         this.lastRenderedWithControls = this._objects[i];
       }
+    },
+
+    /**
+     * @private
+     * @param {fabric.Object} obj Object that was removed
+     */
+    _onObjectRemoved: function(obj) {
+      if (obj === this.lastRenderedWithControls) {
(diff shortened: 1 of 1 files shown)


src/canvas.class.js
```diff
@@ -1361,6 +1361,17 @@
...
+
+    /**
+     * @private
+     * @param {fabric.Object} obj Object that was removed
+     */
+    _onObjectRemoved: function(obj) {
+      if (obj === this.lastRenderedWithControls) {
+        delete this.lastRenderedWithControls;
+      }
+      this.callSuper('_onObjectRemoved', obj);
```
asturur: would it not fit in StaticCanvas more than Canvas? StaticCanvas does not have interactivity layer but still can render controls.
Other than that it look goods to me.

plainview: `this.lastRenderedWithControls` is defined in the canvas.class.js.  Should I move all uses of the flag to StaticCanvas?

asturur: no you are right, it does not belong in StaticCanvas. is it fine like it is.

PR #3048, opened 2016-06-15
Added travis ci for node 6

amilajack:


asturur:
If i add 6 i remove 5. I do not want running so much instances. is 5 still supported / used / make sense?

amilajack:
I have seen a lot of repos test only node 5. If i were to keep it minimal, I would test only against 4 and 6.

Issue #3074, opened 2016-07-01
Group object ignore own lockMovementX/Y = false properties


geekduck:
## Version

1.6.2
## Test Case

https://jsfiddle.net/geekduck/0u62cwy5/2/
## Steps to reproduce
1. Create two objects (lockMovementX/Y = true) .
2. Create group object(lockMovementX/Y = false) with 1. objects.
## Expected Behavior

When group object has own lockMovementX/Y=false properties,  should moving.
## Actual Behavior

Group object is locked.

geekduck:
Here.
https://github.com/kangax/fabric.js/blob/master/src/shapes/group.class.js#L499
```
if (typeof this[prop] !== 'undefined') {
```

neopheus:
like this : https://jsfiddle.net/0u62cwy5/3/ ?

asturur:
incline to remove both those lockProperties and delegated properties.
if you want the group to do not move, set the group props.
For activeGroup ( multiselection ) i will find another way.

asturur:
will take longer than expecting.
I m removing the lock properties and delegated properties from Group. This is a breaking change and so i ll do for 2.0. ( already did )
I will introduce a ActiveSelection class that will do this quirks for the Group.

Group is a container and will do the container.
ActiveSelection/ActiveGroup is a more evolved container and will get some automatism for setting  props and locking.

fredguth:
@asturur  Could you further elaborate on this?  i didn't understand.

I had a set of items with lockMovementX/Y=true.
before grouping, I set items lockMovementX/Y= false
I create a group from items, I expected group to be unlocked, but it was locked.

I then further set group.lockMovementX/Y = false
Still, my group is locked.

Didn't understand why.

asturur:
can you elaborate a fiddle to show it? are you talking of  version 1 or 2?

fredguth:
Never mind... I thought I was in the latest version, but wasn't. Sorry.

PR #3075, opened 2016-07-01
Group.get return own lockProperties.

geekduck:
#3074

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -496,7 +496,7 @@
      */
     get: function(prop) {
       if (prop in _lockProperties) {
-        if (this[prop]) {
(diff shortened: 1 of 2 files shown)


asturur:
does this fixes the issue from the group moving?

geekduck:
yes.
group object ignore all own lock properties(moving, rotate, scale).
```
  var _lockProperties = {
    lockMovementX:  true, // #3074 moving
    lockMovementY:  true, // #3074 moving
    lockRotation:   true,
    lockScalingX:   true,
    lockScalingY:   true,
    lockUniScaling: true
};
```

asturur:
hard to tell. hard to tell.
what were the real intensions of this code?
maybe if one object is locked inside, even if the group is unlocked, it should not move. i will think of this trying to understand what to do

PR #3077, opened 2016-07-02
FindTarget patch 1

blucobalto:
the activegroup was not working on transparency check because the pointer was not defined when calling findTarget on the activeGroup


asturur:
thanks!

PR #3100, opened 2016-07-12
Fix getPointer for canvas in scrolled elements

specialkk:
This restores some of the changes that #407 made to fix the issue that has since reappeared. I've also fixed the documentation for the signature on the Canvas getPointer method (the second one was undocumented, the third was unnecessary).


asturur:
could you make an example of what case is failing with current code and which one is fixed now?

rockmacaca:
Sure, I'll try to find some time to make one, it's rather simple. I think all that is needed is a canvas with a large top margin inside an element with `overflow: auto`.

asturur:
any time for showing the failing cases?

rockmacaca:
http://jsfiddle.net/n7qonqcg/2/

I don't have access to my usual setup to build fabric with the new changes, but here is a failing case. As it turns out, I can only reproduce it for overflowing children in a fixed position container.

In the fiddle, dragging the picture so the cursor is outside the bounds of any elements that are causing the container to overflow (the blue spacer div or the orange canvas container) will calculate an incorrect pointer position.

specialkk:
It looks like this got fixed at some point with the changes for 2.0. It's still an issue for the latest build in 1.x, but I think it's safe to close this.

PR #3101, opened 2016-07-13
getSrc: Return the src property on an Image if no element is present

rockmacaca:
This allows toSVG to generate an xml string without having to load the
images. Especially helpful in NodeJS environments.

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -224,7 +224,7 @@
       });
 
       var object = extend(this.callSuper('toObject', propertiesToInclude), {
-        src: element ? element.src || element._src : '',
+        src: element ? element.src || element._src : this.src || '',
         filters: filters,
         resizeFilters: resizeFilters,
         crossOrigin: this.crossOrigin,
@@ -299,6 +299,9 @@
     getSrc: function() {
       if (this.getElement()) {
         return this.getElement().src || this.getElement()._src;
+      }
+      else {
+        return this.src;
       }
     },
 


asturur:
it looks like that src property is available just by chance when we load back an image instance from an object as a side effect. is not a standard available property.

rockmacaca:
Hmm, I thought it was intended. Is there any reason people shouldn't rely on that property being there?

`toSVG` really shouldn't have to rely on an image being loaded if its `src`, `width`, and `height` are all defined on the object. Currently I have to override `fabric.Image.prototype.getSvgSrc` to get `toSVG` to work the way I want to in Node.js. Would a good compromise be just modifying `getSvgSrc` to return the `src` property on the object if `this.getSrc()` is undefined?

asturur:
Your change is fine, just 90% of the time this.src is not there. I checked the source code and that .src does not get set by anything. toObject method put the src value inside the src property. FromObject method reload the image from .src and then pull all the object as options, and so just in that case .src gets populated.

Do you understand what i mean?

rockmacaca:
Yeah, I completely understand. Your point also got me to foresee some potential trouble for myself when exporting an element-less `Image` as a plain object, so I also updated `toObject` to grab `this.src` if no other options are available.

asturur:
Still i do not get who is setting 'this.src' in the code. in my opinion no one.
if you do var img = new fabric.Image(imageElement);  img.src is undefined

asturur:
I m ok with merging this if you add the default value of .src to the class and relative tests.

asturur:
https://github.com/kangax/fabric.js/pull/3173

PR #2771, opened 2016-02-05
do not force rect visibility true if dimensions are > 0

asturur:
this fixes correct parsing of display / visibility attribute for rects.

failing svg

```
[...]

PR #3106, opened 2016-07-15
Fix to Rect imported from svg is invisible when display = "none"

zntus:
i imported a svg file contain this line
`<rect x="12" y="0" display="none" width="56.7" height="56.7"/>`
display attriubute was 'none'.  but the rect was visible.
so i fixed it.


asturur:
can you help me understanding what is wrong on #2771?
i noticed this bug a while a go and tryied to fix it, but:

if a rect has display block, but is in a group with display none,  it should not be visible.
if a rect is in a group with visible false and the rect is visible true it should be visible.

https://www.w3.org/TR/SVG/painting.html#VisibilityControl

zntus:
@asturur i didn't know about #2771. i'm closing this request

asturur:
for now i merge this, still better to have this fix as i do not have time to investigate #2771.

PR #3128, opened 2016-07-27
Add custom draw control functions

vudduu:
Add custom draw functions for the corner-resize-controls and rotation-control.

example in how to use:

```
canvas.on('object:selected', (event) => {
  const { target } = event;
  target.set({
    cornerStyle: 'custom',
    drawCustomControlShape: {
      bl: function drawCustomBorder(ctx, left, top, size) {
        // use ctx to draw corner shape for bl
      },
      br: function drawCustomBorder(ctx, left, top, size) {
        // use ctx to draw corner shape for br
      },
      tl: function drawCustomBorder(ctx, left, top, size) {
        // use ctx to draw corner shape for tl
      },
      tr: function drawCustomBorder(ctx, left, top, size) {
        // use ctx to draw corner shape for tr
      },
      mtr: function drawCustomBorder(ctx, left, top, size) {
        // use ctx to draw corner shape for mtr
      }
    }
  });
});
```

### DIFF
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -393,8 +393,8 @@
         case 'custom':
           if (this.drawCustomControlShape[control]) {
             this.drawCustomControlShape[control].call(this, ctx, left, top);
-            break;
           }
+          break;
         default:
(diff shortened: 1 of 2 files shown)


neopheus:
it's a good idea :)
👍

PR #3145, opened 2016-08-07
Added missing object:added on docs

jrbasso:
`object:added` is also fired by canvas. It is proved on the test cases: https://github.com/kangax/fabric.js/blob/1.6.3/test/unit/canvas_static.js#L269-L279


asturur:
Thanks!
jdocs contribution are very well welcome.

Issue #2930, opened 2016-05-03
object.borderScaleFactor


asturur:
We have a property:
borderScaleFactor that is the inverse of the thickness of the object control box border.
Influences borders but not the borders of controls when drawn in transparent mode.

@kangax @kienz, someone of you remember why has been introduced?

I'm inclined towards making it the border tickness and not the inverse of and apply it to controls borderds. ( actually a borderScaleFactor of 2 produces a border of 0.5 pixel, counter intuitive)

OR

remove it

Any suggestion?

kangax:
Probably best to remove since it's so specific...

asturur:
Yes i see an utility now in this mood of controls customization, but at least used as a multiplier and not a divider.

PR #3154, opened 2016-08-10
Change borderScaleFactor behaviour (#2930)

sdvg:
See discussion in https://github.com/kangax/fabric.js/issues/2930

I think a multiplication makes much more sense here to cover all use cases. My use case for example is to render a 2px border.

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1183,7 +1183,7 @@
 
       ctx.save();
       ctx.translate(options.translateX, options.translateY);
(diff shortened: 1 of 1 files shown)


sdvg:
I just figured out that I could just have used `0.5` 🙈
Anyway, I think a multiplier is easier to grasp :)

asturur:
I agree.
I was keeping this feature for 2.0 because is a back incompatible change. But i also think that this feature is so unpopular that no one will complain.
closes #2930

Issue #3095, opened 2016-07-10
make small object on top of active one selectable with preserve object stacking true


thachnv:
## Version
1.6.3

## Test Case
http://jsfiddle.net/ecvctazk/

## Steps to reproduce
Try to select the inner object

## Expected Behavior
You should be able to select the inner object

## Actual Behavior
You always got the background object selected

thachnv:
This problem does not happen in <= 1.6.2 version

asturur:
I have some difficulties trying to accomodate any selecting situatuon. That involves object selected is under another, but is bigger, or smaller than the covering one. Or when the selecte object has just one handle under another object.

If you think to have a logic that fix all necessity please illustrate it and help me.

virror:
This does not really seem like the same issue that you just closed for me, but here are my 5 cents anyways. This order should be the most logical one:
-Widgets should always have priority.
-Smaller objects should get selected before larger ones
Not sure the z-index should affect selecting other than if no one of the other conditions are true. Might add an option though that always selects by z-order, but even there, widgets should always have priority.

butch2k:
I would say the first object to get the click event should keep it. If clicked within its visible boundaries an object get selected. Same logic applies to handles an handle should get the click event as if the object. An handle in front of an object should get the click and keep it. 





On Sun, Jul 10, 2016 at 11:02 PM +0200, "Andrea Bogazzi" <notifications@github.com> wrote:












I have some difficulties trying to accomodate any selecting situatuon. That involves object selected is under another, but is bigger, or smaller than the covering one. Or when the selecte object has just one handle under another object.



If you think to have a logic that fix all necessity please illustrate it and help me.

virror:
BTw, my issue occurs on earlier version as well, not just 1.6.2 as for the OP. In our application we changed how selection is handled and smaller objects are always selected regardless of whats above it, like this:
Fabric.Canvas.prototype._searchPossibleTargets = (objects, pointer) ->
	target = undefined

	for object in objects when @_checkTarget(pointer, object)
		if not target?
			target = object
		else if object.height * object.width < target.height * target.width
			target = object

	@relatedTarget = target

This is of course not a complete solution, but i think the concept is much more logical for most cases. To this it also needs to be added a fix for always selecting widgets first, but i have no clue on how that stuff works.

asturur:
-Widgets should always have priority.  Done with #3111

neopheus:
[...]

PR #3157, opened 2016-08-10
preserveObjectStacking Problem #3095

neopheus:
Solution for problem : 
make small object on top of active one selectable with preserve object stacking true #3095


asturur:
i want to have chance to select the active object during preserve object stacking. if you move a small active object behind a big other object and accidentally do mouse up, you cannot recover it anymore.

PR #3163, opened 2016-08-11
replaced 'or' with 'and' in order to enforce 'uniScaleTransform' flag

sirrodgepodge:
1.6.2

The "uniScaleTransform" flag is not actually enforced if false (the default) because of an accidental "or" where an "and" should be

Issue here:
https://github.com/kangax/fabric.js/issues/3162

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -677,7 +677,7 @@
      * @return {Boolean} true if the scaling occurred
      */
     _onScale: function(e, transform, x, y) {
-      if ((e[this.uniScaleKey] || this.uniScaleTransform) && !transform.target.get('lockUniScaling')) {
+      if ((this.uniScaleTransform && e[this.uniScaleKey]) && !transform.target.get('lockUniScaling')) {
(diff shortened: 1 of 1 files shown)


asturur:
There was an older issue about this and the precedence of who would enforce what.
Because we could not decide what was the best option here, if the uniscaleKey should overcome the uniscaleTransform flag or not, in a recent reformatting i made the keys ( once fixed at shift ) customizable.
At this point if i want the uniscaleTransform flag to win over anythingelse i just bind the uniScaleKey to  null or to something that does not exist in the event and i obtain the same effect of swapping the || with &&.

I mean, with the current codebase you can have:
Always uniform - flag to false and key to non existent
Always not uniform - flag to true
normal scaling and key for not uniform - flag false and key to shift, alt, ctrl, super

What is missing is the opportunity to have the key acting as a switch over the flag.

sirrodgepodge:
ahhh I see so the flag is intended to set all scaling to be unproportional, not just on shift key, but you would like to still by default have the shift key allow a user to switch into "unproportional mode".

I now get it

PR #3165, opened 2016-08-15
preserveObjectStacking Problem #3095

neopheus:
Solution for problem : 
make small object on top of active one selectable with preserve object stacking true #3095 
+ with backward compatibility


saadqbal:
Issue:   You cannot resize the object from its controls when its below another object. Please see the jsfiddle for reference:
[https://jsfiddle.net/ecvctazk/35/](https://jsfiddle.net/ecvctazk/35/)

neopheus:
Example with new code :

https://jsfiddle.net/ecvctazk/37/

saadqbal:
Bring the Black box to the edge of the red box and then try to resize the red box from the right handles. It won't resize

saadqbal:
Try resizing the red box with the handles that are over the black box. 
[https://jsfiddle.net/ecvctazk/39/](https://jsfiddle.net/ecvctazk/39/)

neopheus:
Like this : https://jsfiddle.net/ecvctazk/48/

neopheus:
With the new build : https://jsfiddle.net/ecvctazk/49/

saadqbal:
Yup. thats exactly what I wanted

asturur:
i do not feel like adding ANOTHER flag. 
I prefer to find a way to select an object, when there is some race condition, basing on area, distance from center, or whatever may help us. but no other flags.

neopheus:
I agree with you

But in this case, as there has been a major change functioning. Adding a backward compatibility option seems an appropriate solution

neopheus:
jsfiddle refresh : https://jsfiddle.net/ecvctazk/53/

neopheus:
new pull request : https://github.com/kangax/fabric.js/pull/3178
can you close this pull ?
Thank

PR #3181, opened 2016-08-20
Spray chunk height

neopheus:
sprayChunk - 'width' should probably not be assigned to 'height'


asturur:
Not sure about spray chunk. Never used it. maybe is supposed to be always square? What behaviour changes you notice, or what bug you notice to make you think it needs a fix?

neopheus:
I do test

It's ok with width : https://jsfiddle.net/ecvctazk/55/

So I close ;)

PR #3185, opened 2016-08-21
Json2 update

neopheus:
update json2.js


asturur:
no reason to update builds or json library. json is excluded in all builds and kept just for compatibility with old browsers.

PR #3188, opened 2016-08-21
Lerp - lock t between 0 and 1

neopheus:
Lerp - lock t between 0 and 1


asturur:
Can you add 2 little test in the point unit test were we actually check that lerp is locked between 0 and 1 when passing higher or lower numbers?
I think i do not care to bump to 2.0 to merge this.

neopheus:
no problem.

I'm taking care of it

neopheus:
Unit Test ok ;)

PR #3195, opened 2016-08-23
Make object fire deselect event

asturur:
There is no current way to make some action during object deselection that will intervene after the deactivating the activeObject/ activeGroup.
We have selection:cleared that had no reference to object deselected.
To be similar to other events we mirror the event on single object.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1285,7 +1285,8 @@
       if (this.getActiveObject() === obj) {
         this.fire('before:selection:cleared', { target: obj });
         this._discardActiveObject();
-        this.fire('selection:cleared');
+        this.fire('selection:cleared', { target: obj });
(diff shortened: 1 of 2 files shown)


colesimmons:
Can you attach this as a canvas event?

e.x. `canvas.on('object:deselected', e => {});`?

If so, is it in production yet? This is great - exactly what I need.

colesimmons:
Okay, but it is available for use as an object listener, like `rect.on('deselected', () => {});` ?

asturur:
yes

2016-09-19 18:55 GMT+02:00 Cole Simmons <notifications@github.com>:

colesimmons:
Oh, fantastic! Just what I needed. http://fabricjs.com/fabric-intro-part-2#events probably needs to be updated to reflect the change.

PR #3197, opened 2016-08-23
Add Dblclick Event

neopheus:
Add Dblclick Event
Example : 
canvas.on('mouse:dblclick', function(event){
   // ADD YOUR CODE
});
or
rect.on('object:dblclick', function(event){
   // ADD YOUR CODE
});

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -50,6 +50,8 @@
       addListener(this.upperCanvasEl, 'touchstart', this._onMouseDown);
       addListener(this.upperCanvasEl, 'touchmove', this._onMouseMove);
 
+      addListener(this.upperCanvasEl, 'dblclick', this._onDoubleClick);
+
       if (typeof eventjs !== 'undefined' && 'add' in eventjs) {
         eventjs.add(this.upperCanvasEl, 'gesture', this._onGesture);
         eventjs.add(this.upperCanvasEl, 'drag', this._onDrag);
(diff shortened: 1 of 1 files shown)


asturur:
i do not think dblclick is a standard event. is it?

neopheus:
this event is managed by event.js

so I update with :

`if (typeof eventjs !== 'undefined' && 'add' in eventjs) {`

neopheus:
it's ok

new version ;)

Fidel90:
`dblclick` seems to be a standard event: https://developer.mozilla.org/en-US/docs/Web/Events/dblclick

asturur:
i ll check it. need to check itext and other stuff behaving.

Fidel90:
I would totally appreciate if `dblclick` finds its way into fabric. I just want to suggest that if you want to keep event naming consistent it should be `mousedblclick` for objects (`mouse:dblclick` on canvas would be ok).

PS: As I'm writing this I'm wondering if there is already simple `mouse:click` / `mouseclick` available? I know I can detect a click via watching mousedown/up but same would apply for `dblclick` and others ;-)

asturur:
as of now doubleclick is simulated with a succession of down up down up.
i see in adding  simple click an added complexity in managing effect of mouse up and moise down that are built in in fabric ( destroying active selection ) also double click will not stop the other events to fire so is something i have to evaluate carefully. as soon as i can finish preparing 1.6.4 update the immediate priorities are selection of overlapping objects, zoomed canvas and group subtargeting,  clean up textspacing and cleanup text in general. i m not gonna watch doubleclick immediately. there is a working implementation of double click in itext if you need it.

PR #3204, opened 2016-08-25
Update Wheel event

neopheus:
Demo with new code : https://jsfiddle.net/bzd6888b/3/

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -44,6 +44,7 @@
       addListener(this.upperCanvasEl, 'mousedown', this._onMouseDown);
       addListener(this.upperCanvasEl, 'mousemove', this._onMouseMove);
       addListener(this.upperCanvasEl, 'mouseout', this._onMouseOut);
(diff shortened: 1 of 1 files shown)


asturur:
What was the problem with old code? i want to rely on event.js as little as possible.

neopheus:
With the old code the event does not work (test On Firefox 47 Windows + Chrome 52)

Example with bug : https://jsfiddle.net/bzd6888b/5/

asturur:
image

it looks like the wheel event is standard. Why rely on event.js?

neopheus:
Actually in the code : https://github.com/kangax/fabric.js/blob/fd6038d50d4b32f2b43448fef1b8b4b3ba996432/src/mixins/canvas_events.mixin.js#L46

We use "mousewheel" event, in the Mozilla documentation : https://developer.mozilla.org/en-US/docs/Web/Events/mousewheel

neopheus:
New version Mouse Wheel Event without event.js

asturur:
this i will check carefully. it looks strange that we have mousewheel broken and no one complains.

neopheus:
yes it's strange
With the new code, we don't use Event.js ;)

PR #3210, opened 2016-08-26
bring callback toapplyfilters

### LINKED ISSUE #365: Filters don't apply after load canvas from JSON
When you save and load the canvas in JSON, the filters on the objects don't be apply until you call an async renderAll on the canvas, by refreshing it with canvas.renderAll() or by clicking inside the canvas.

See an exemple :
http://www.xorax.info/labs/misc/canvas/fabricjs-issue-loadAndFilter.php

Apply the greyscale filter, click save to save the JSON in the input, and
- click to Load button to load the JSON in the input and click on the canvas to refresh it
- or click to Load and refresh button to load and call Load and refresh after 500ms

It seems there is an async break somewhere.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/107471-filters-don-t-apply-after-load-canvas-from-json?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

asturur:
to solve load from json issues and apply filters, both normal and dynamic resize.
Also take in consideration retina when scaling down the image so that crispness is preserved.
Added a parameter to do not scale over a factor of 0.5
closes #365


neopheus:
great :)

src/filters/resize_filter.class.js
```diff
@@ -62,6 +62,9 @@
      * @param {Number} scaleY
      */
     applyTo: function(canvasEl, scaleX, scaleY) {
+      if (scaleX === 1 && scaleY === 1) {
+        return;
+      }
```
asturur: make filter exit if we want have a bigger image than original.
The resize filter is meant for downscaling and too slow to be used as upscaler. There are no added details

src/shapes/image.class.js
```diff
@@ -121,15 +122,27 @@
...
       this._initConfig(options);
 
+      if (this.resizeFilters.length === 0) {
+        _callback = callback;
+      } else {
+        _this = this;
+        _callback = function() {
+          _this.applyFilters(callback, _this.resizeFilters, _this._filteredEl || _this._originalElement, true);
+        }
+      }
```
asturur: if there are resize filters in the restore from object/json, let's fire the callback after we restore also those.

src/shapes/image.class.js
```diff
@@ -296,7 +319,7 @@
      * @return {String} Source of an image
      */
     getSrc: function() {
-      var element = this.getElement();
+      var element = this._originalElement;
```
asturur: _element could be a replacement from filtering where src is no more meaningfull

goodcode:
Hi there,

It seems like this commit changes something in a major way - the way we serialize and send the canvas as svg using toSVG method, when there are filters applied. 
In the following scenario: you add an image and then you apply some filters on it, here is what happens
1. previously it used to apply the filters on the image and send a new image as data (base 64)
2. now it sends a URL to the original image

Is this the intended behavior? If yes, then how should one work with filters?

asturur:
no, is not intended behaviour, i did not touch the export toSVG function but maybe the getSrc function changes are implied. Probably we do not have the test for it.

Please open an issue with the latest version, a small clear fiddle with the related code and i will fix it asap.

goodcode:
Thanks a lot for your cooperation!
There is a link to the newly opened issue:
https://github.com/kangax/fabric.js/issues/3298

PR #3229, opened 2016-09-03
switch to eslint

asturur:
not working for now.
converted configuration files with tool.


asturur:
Main things to fix are:
one-var declaration, remove extra space between square brackets,  fix some indent case.

After all this taking down jsdocs that are most of the failures ( now disabled ).

asturur:
@neopheus  if you have will to help, this is just about fixing spaces and indent. If no sure about some rules ask.
I know this is a boring job

neopheus:
no pb
i do it ;)

asturur:
pick up some files, let me.know wich, i ll work on other, so we do not steps on our feets

asturur:
fixed util/misc.js

asturur:
removed space in square bracket in all files

asturur:
disabled one-var, down to 79 failures

asturur:
Some comparison with stricit equal probably killed some functionalities. if i cannot find it i will revert changes.

PR #3254, opened 2016-09-11
fixed selection, added test

### LINKED ISSUE #3095: make small object on top of active one selectable with preserve object stacking true

asturur:
I change again the logic of selection. I hope this time everyone find it feasible.

1) clicked activegroup? return activeGroup
2) clicked activeObject on a corner? return activeObject
3) clicked activeObject and preserveObjectStacking is false? return activeObject.
clicked activeObject? ok take note. Examine targets.....
4) target is activeObject again? return activeObject.
5) target is an object that is over activeObject? check if the alternativeSelectionKey is pressed.
key is pressed? if yes return activeObject, normally return the target clicked.

So there is a solution for selecting activeObjects under another object but that require a keypress ( configurable ). In normal situation the topmost object is returned.

close #3095

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -118,12 +118,22 @@
 
     /**
      * Indicates which key enable multiple click selection
-     * values: altKey, shiftKey, ctrlKey
+     * values: altKey, shiftKey, ctrlKey, cmdKey
      * @since 1.6.2
      * @type String
      * @default
      */
     selectionKey:           'shiftKey',
+
+    /**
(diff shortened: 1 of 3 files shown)


asturur:
@neopheus

PR #3262, opened 2016-09-14
allow for deep state check and save

asturur:
SaveState and hasStateChanged work on a simple equality and reference storage.
So storing a fill color is ok, storying an array of values ( transformMatrix, textDecoration ) is not.

This Pr enable for deeper checks, Fixes a bug in wich additional state properties where not check during hasStateChanged and extend the merge ( and clone ) util with deep option.


eschnabel:
I have updated to 1.6.5 (and now 1.6.6) and we have an object that extends Line and then includes a shadow. When the shadow is added this new deep state clone stuff fails in a stack overflow and breaks everything for us. If I change the deep = true to deep = false, it works. Is there some way for us to configure this outside of fabric? Is there something that would cause this endless loop?

Rad1calDreamer:
+1 to previous

asturur:
did not get this notice. Giving a look to this. sorry i released 1.6.7 just 2 days ago.

asturur:
I have a question, are you using stateful = true?
Since some version stateful true is sort of useless unless you are using it for your custom code.
So as an immediate solution try to start the canvas with stateful = false and  i m looking this now.

Rad1calDreamer:
i use stateful:false in my app

Rad1calDreamer:
```javascript
this.canvas.forEachObject(function(obj) {
            if (!obj.id) {
               var _obj2 = obj.toObject(['x1', 'y1', 'x2', 'y2', 'objID', 'selectable']),
                  _obj;
               if (obj.type === 'path') {
                  _obj = new fabric.Path.fromObject(_obj2, function(foo) {
                        _tmpCanvas.add(foo);
                     }
                  );
               } else {
                  _obj =
                     new fabric [fabric.util.string.camelize(fabric.util.string.capitalize(obj.type))].fromObject(_obj2);
                  _tmpCanvas.add(_obj);
               }

            }
         });
```

that code. If some of objects have shadow -> extend function gooing to maximull stack call

PR #3277, opened 2016-09-19
fixes Issue 3256 - line's style was gone when pushing line down

### LINKED ISSUE #3256: Textbox - line's style is gone when pushing line down
## Version
1.6.4

## Test Case
http://jsfiddle.net/cf513q09/2/

## Steps to reproduce
1. Open the fiddle. The first line of text is already bolded.
2. Position a cursor before first letter
3. Hit enter

## Expected Behavior
The text should be pushed to the second line and it should preserve the styling.

## Actual Behavior
The text was pushed to the second line, but the styles are lost.

## Additional informations
I have noticed, that adding one simple condition to the `if` statement of `fabric.Textbox.prototype._generateStyleMap` seems to fix the issue. Take a look at my solution below (marked with a comment `// belfz's fix: added '&& i > 0' here`):
```
fabric.Textbox.prototype._generateStyleMap = function() {
          var realLineCount     = 0,
              realLineCharCount = 0,
              charCount         = 0,
              map               = {};

          for (var i = 0; i < this._textLines.length; i++) {
            // belfz's fix: added '&& i > 0' here
            if (this.text[charCount] === '\n' && i > 0) {
              realLineCharCount = 0;
              charCount++;
              realLineCount++;
            }
            else if (this.text[charCount] === ' ') {
              // this case deals with space's that are removed from end of lines when wrapping
              realLineCharCount++;
              charCount++;
            }

            map[i] = { line: realLineCount, offset: realLineCharCount };

            charCount += this._textLines[i].length;
            realLineCharCount += this._textLines[i].length;
          }

          return map;
        };
```

belfz:
It fixes the bug mentioned [here](https://github.com/kangax/fabric.js/issues/3256).
close #3256

### DIFF
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -118,7 +118,7 @@
           map               = {};
 
       for (var i = 0; i < this._textLines.length; i++) {
-        if (this.text[charCount] === '\n') {
+        if (this.text[charCount] === '\n' && i > 0) {
           realLineCharCount = 0;
           charCount++;
           realLineCount++;


asturur:
can you shed some light on why the bug was happening? i did not want / had time to investigate becuase i do not really like touching that function.

belfz:
The logic of these algorithms are also very complicated for me, but as far as I remember: if you hit enter as the first character of the first line of text (using Textbox), the generated style map would incorrectly point to line index 1 (when asked for value under key 0 of the returned map).

Please also see the fiddle attached to my original bug report. I also remember, that while investigating this problem a week ago, the incorrect output of `_generateStyleMap` influenced the way the fabric.Textbox.prototype.shiftLineStyles worked.

asturur:
if i cannot find strange bugs i pull it in that i m ready to release 1.6.5

belfz:
Thank you. I wanted to prove my point via unit tests, but it turned out that are none for Textbox component.

asturur:
you can write them all.
I know that is why i m always inclined to do not accept features big or small without coming UT.

PR #3286, opened 2016-09-26
Switch the ocoords system for groups

asturur:
This PR aims at setting oCoords of grouped objects without the zoom of canvas.
In this way the zoom is not applied once for nesting level and normalizing the pointer is easier.


samael205:
Hi @asturur , this commit solves the problem of the coordinates of the groups when selecting programmatically.

I explain what I do:

What I do is that when the load canvas, select all objects on the canvas, creating a group. From that group, to obtain the boundingBox, and go zooming in or zoom out so that the elements conform to the width and height of the canvas.

Currently only I can do this, if pre-select all items with the mouse

Regards!

asturur:
yes but it needs more changes and controls. is not ready yet. i m finishing object caching before.

samael205:
Ok, I will be patient

samael205:
@asturur Hi, I have a another question referenced by this PR.

Is it possible that the top and the left of the copy in groups are also badly calculated?

PR #3287, opened 2016-09-26
Fixing hidden text positioning issue

codewp:
This pull request fixing an issue related to false positioning of hidden textarea.
When document has elements otherwise canvas fabricjs can not position hidden textarea in right position and this pull request will fix it.


asturur:
this has been reverted because was breaking IE selection. the hidden textarea cannot be child of canvas.

codewp:
I created this pull request to fixing an issue related to calculating position of hidden text area that used in IText of fabricjs.

You can view issue in below jsfiddle code, when user wants to edit text layer inside canvas, fabricjs can not calculate right position of hidden textarea so browser scrolls up to hidden text area position.
_calcTextareaPosition function can not calculate right position of hidden text area.

https://jsfiddle.net/ajrtr1fc/

Please let me know how can I fix this issue.

PR #3309, opened 2016-10-02
use native stroke dash for selection area

asturur:


eschnabel:
When updating to 1.6.5 from 1.6.4 a piece of code that creates dashed line with an added shadow suddenly goes into an endless loop inside of fabric. Cannot debug and unfortunately cannot provide code. When I remove the setShadow(shadow) call from that element, it works as before, just without the shadow which we need. Has this been tested with shadows?

asturur:
can you open a fiddle that show your issue?

PR #3314, opened 2016-10-02
changed to dataurl export

### LINKED ISSUE #3152: Issue on resize canvas, and then generate data url
Hello,

I have an issue when I resize the canvas and then call `toDataURL` on iPhone and iPad.

I saw that when we call `setWith()` and `setHeight()`, fabric call the function `setDimensions()`, which call `_initRetinaDisplay()` function to apply a pixel ratio on the canvas (x2 on iOS).

Then, when I call toDataURL, fabric call again the function `_initRetinaDisplay()`, so the width and height are mutiplied twice (x4 in total). So, I have an error message in the console: "Canvas area exceeds the maximum limit (width * height > 16777216).

asturur:
close #3152

-reverted the retinaScaling as multiplier change
-removed cropped logic and used viewportTransform
-use this.interactive = false to avoid draw border of active objects

The actual implementation was effectively producing a double size canvas for retina enabled devices, creating problems to low memory devices as ipads.

### DIFF
--- a/src/mixins/canvas_dataurl_exporter.mixin.js
+++ b/src/mixins/canvas_dataurl_exporter.mixin.js
@@ -1,217 +1,124 @@
-fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.StaticCanvas.prototype */ {
+(function () {
 
-  /**
-   * Exports canvas element to a dataurl image. Note that when multiplier is used, cropping is scaled appropriately
-   * @param {Object} [options] Options object
-   * @param {String} [options.format=png] The format of the output image. Either "jpeg" or "png"
-   * @param {Number} [options.quality=1] Quality level (0..1). Only used for jpeg.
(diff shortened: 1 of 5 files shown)


src/mixins/canvas_dataurl_exporter.mixin.js
```diff
@@ -1,217 +1,124 @@
...
+
+    /**
+     * @private
+     */
+    __toDataURLWithMultiplier: function(format, quality, cropping, multiplier) {
+
+      var origWidth = this.getWidth(),
+          origHeight = this.getHeight(),
+          scaledWidth = (cropping.width || this.getWidth()) * multiplier,
+          scaledHeight = (cropping.width || this.getHeight()) * multiplier,
```
thoepfner: For `scaledHeight`, this should be `cropping.height` (not `cropping.width`).

PR #3317, opened 2016-10-03
Objectcaching

asturur:
- [X] basic caching system
- [x] adapt stateful system to work on different sets of properties
- [x] working cache for text
- [x] caching for groups
- [ ] evaluate performance of caching images  --> not doing it now
- [x] evalualte benefits of having 2 kinds of caching, one involving scaling and one not
- [ ] write tests for changed statefull functions


asturur:
Cannot run Unit Tests under node with caching on, something to do with creation of additional canvases with jsdom. I do not think we need cache under node, if we are not interacting with objects with mouse or rendering animations, cache is generally slower. So for now is disabled under node. I will write specific tests anyway just to be sure i can track changes.

dist/fabric.js
```diff
@@ -6087,7 +6087,7 @@ fabric.Pattern = fabric.util.createClass(/** @lends fabric.Pattern.prototype */
      * @type Boolean
      * @default
      */
-    stateful: true,
+    stateful: false,
```
asturur: default stateful to false. stateful was used in automatic way to detect a change in object modified at end of transformation. Now we detect it properly with the result of actions. Stateful is no more necessary if not for user needs that will activate it. ( detecting changes in fill, stroke and so on)

samael205:
which it is the basic essence of this functionality?

asturur:
having complex object cached on additional canvases. it won t help for rectangles, circles and simple paths, but will help fps/general smoothness in situation where you want to draw lot of complex groups, or styled text.

samael205:
additional canvases?, Hidden canvases? o auxiliar canvases

samael205:
For example, for complex brushes like spray, that render thousands of points to draw the effect?

asturur:
did not try with brushes yet.

samael205:
Ok, when you release, i will try.

asturur:
@samael205  i pushing this change in.
It helps a lot with spray brush, under certain conditions.
I m gonna add a page on the website to explain what this do and how, what are the side effects and how a developer should use it.

samael205:
@asturur Thanks for your Hard work. I will try tomorrow.

samael205:
@asturur, I'm going to work on a totally different project in the probably  threee next months.
I will continue reading the news so I do not miss anything.
See you in a few months

PR #3324, opened 2016-10-04
Fixing an issue in applying filters when filter type is Tint.

codewp:
This pull request will fix an issue in applyFilters function of fabric.Image.
This issue occurred when I used canvas.toJSON function as you know this function convert canvas to json object, after that when I tried to loading that object I saw that there is a problem for layers that have filters of Tint type, because when canvas.toJSON function used it converts Tint filters to object with toObject function of Tint class, so in that object there is not applyTo function, and this issue occurs in fabric.Image applyFilters function.

In applyFilters function of fabric.Image class it use `filter.applyTo(canvasEl, scaleX, `scaleY);` this code to applying filter but when filter is type of `Tint` object because it has not applyTo function this issue will occurs.

For more information I debugged codes of fabric to finding this issue you can see some of screen shots of debugging process.
fabric-filters-2

filters before using toJSON function of canvas.

fabric-filters

filters after using toJSON function of canvas.

Regards.


asturur:
This is not the fix.
Did you try changing in the fabric source code the filter type from 'Tint' to 'tint'? this should fix.

codewp:
As I said filter is not in Tint type and it is a plain object with color and type, so it has not applyTo function so it cause error, for fixing it first I change that plain object to Tint class type that has applyTo function and after that use it's applyTo function.

One another solution for it can be editing toObject function of filters.Tint class because it returns an object like this { color : '#ffffff', type : 'Tint' } and returning right object of Tint class that has it's functions too.

Please let me know your opinion.

asturur:
no, all the object output the filter in that way. there is no need for the function. imageFromObject is taking care of restoring functions also for filter.
I guess that all filters are failing then.
If only Tint is failing, then there must be a problem somewhere else.

codewp:
OK, excuse me this problem occurred because I used my own implementation to adding layers to canvas instead of using loadFromJSON function to loading layers.

Issue #3326, opened 2016-10-05
[$200] Introduce cross-browser CI testing


jedateach:
Travis CI is only running tests on Node.js. This is not sufficient for fast-paced cross-browser CI testing, and quickly ensuring changes are not breaking across all supported browsers.

I would like to suggest we get Fabric.js running on saucelabs open-source offering: https://saucelabs.com/open-source

Here's instructions for hooking into Travis:
https://docs.travis-ci.com/user/gui-and-headless-browsers/

I have managed to run Fabric.js tests using the testem test runner:
![screen shot 2016-10-04 at 9 33 01 pm](https://cloud.githubusercontent.com/assets/1356335/19101830/cb8f33e0-8b27-11e6-933a-200f2e2670e5.png)
As you can see, there are failing tests that need to be looked into.

Testem is a great tool for test-driven-development across multiple browsers at the same time.

We can set up a similar testem setup for hooking into saucelabs, like Ember.js:
https://github.com/emberjs/ember.js/blob/master/testem.dist.json

<svg xmlns="http://www.w3.org/2000/svg" width="168" height="20"><linearGradient id="b" x2="0" y2="100%"><stop offset="0" stop-color="#bbb" stop-opacity=".1"/><stop offset="1" stop-opacity=".1"/></linearGradient><mask id="a"><rect width="168" height="20" rx="3" fill="#fff"/></mask><g mask="url(#a)"><path fill="#555" d="M0 0h87v20H0z"/><path fill="#129e5e" d="M87 0h81v20H87z"/><path fill="url(#b)" d="M0 0h168v20H0z"/></g><g fill="#fff" text-anchor="middle" font-family="DejaVu Sans,Verdana,Geneva,sans-serif" font-size="11"><text x="43.5" y="15" fill="#010101" fill-opacity=".3">bountysource</text><text x="43.5" y="14">bountysource</text><text x="126.5" y="15" fill="#010101" fill-opacity=".3">$200 bounty</text><text x="126.5" y="14">$200 bounty</text></g></svg>

asturur:
i would like this much.  i will get a look into this now

jedateach:
That PR may help.
It isn't configured to be the `npm test` script, but could be set that way at some stage.

asturur:
i m gonna read it carefully and try to activate it on travis. for now i opened the opensauce account

asturur:
So:
- testem would be a new dev dependency.
[...]

PR #3331, opened 2016-10-07
Pr/saucelabs

jedateach:
CI build is working smoothly now.

I've configured it to test all the existing node versions, plus PhantomJS and some browsers that came from the ember config, but I would recommend we add more of the available browsers from various platforms: https://saucelabs.com/platforms - especially mobile browsers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ matrix:
     - node_js: "5"
     - node_js: "4"
     - node_js: "0.12"
+    - node_js: "0.11"
+    - node_js: "0.10"
     - env: LAUNCHER=PhantomJS
   # TODO: fix issues and remove allowed failures
   allow_failures:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,47 @@
 language: node_js
 node_js:
   - "6"
-  - "5"
-  - "4"
-  - "0.12"
-  - "0.11"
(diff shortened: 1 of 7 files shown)


jedateach:
I did change the credentials before PR, but none of the saucelabs launchers appear to be working - I think there is something up with the authentication. 

Here you can see saucelabs runs are working with my account:
https://travis-ci.org/jedateach/fabric.js/builds/165947729

asturur:
very nice!

jedateach:
I think this is failing because it is a pull request.

This:
https://docs.travis-ci.com/user/sauce-connect/

mentions to set up this:
https://docs.travis-ci.com/user/jwt

jedateach:
Good to go, but I think there is a concurrency limit, I've discussed in #3326

asturur:
we need to do some changes to the test page so that the fixtures have the same relative path as fabricjs.com

jedateach:
OK. Do what you need to.

jedateach:
@asturur Just dropping in to say I haven't forgotten about this. Been rather busy lately.

Is there anything I can do to help this? Is the website code open sourced? Doing so might help to keep the website up-to-date, and make improvements like the one needed here.

asturur:
yes website is on another reposotory. i did additional cleaning of tests. i ll try to recap what is missing

PR #3332, opened 2016-10-09
change how itext handle removal handlers

### LINKED ISSUE #3320: JS Error when calling discardActiveObject()
## Version
1.6.5

## Test Case
http://jsfiddle.net/ftxjwfh9/2/

## Steps to reproduce
Simple add an iText to canvas, remove it, add a new one and then call discardActiveObject()
This started happening after update to 1.6.5, works fine with 1.6.4

## Expected Behavior

No JS Error

## Actual Behavior

JS Error on the console
```
VM1140 fabric.js:23496 Uncaught TypeError: Cannot read property '_iTextInstances' of undefined
```

I believe it's not deregistering to events correctly when removed

asturur:
close #3320 

Make the handlers for adding and removing itext instances unique to the canvas so that they can be removed when there are no more itext instances

Properly remove canvas reference to the itext object when removed from canvas

Added same logic to canvas clear that was missing

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -33,18 +33,14 @@
     initAddedHandler: function() {
       var _this = this;
       this.on('added', function() {
-        if (this.canvas && !this.canvas._hasITextHandlers) {
-          this.canvas._hasITextHandlers = true;
-          this._initCanvasHandlers();
-        }
-
-        // Track IText instances per-canvas. Only register in this array once added
-        // to a canvas; we don't want to leak a reference to the instance forever
-        // simply because it existed at some point.
-        // (Might be added to a collection, but not on a canvas.)
-        if (_this.canvas) {
-          _this.canvas._iTextInstances = _this.canvas._iTextInstances || [];
-          _this.canvas._iTextInstances.push(_this);
+        var canvas = _this.canvas;
+        if (canvas) {
+          if (!canvas._hasITextHandlers) {
+            canvas._hasITextHandlers = true;
+            _this._initCanvasHandlers(canvas);
+          }
+          canvas._iTextInstances = canvas._iTextInstances || [];
+          canvas._iTextInstances.push(_this);
         }
       });
     },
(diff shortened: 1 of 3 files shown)


samael205:
Hello, This merge fix the bug?
Thanks for your hard work.

asturur:
it should

samael205:
Ok, how can i apply these changes, i downloaded again the library, But the bug persists. Iit is because it has not been compiled yet to the minified version?

asturur:
i think i ll build try it good and release 1.6.6 hotfix so we are all set

samael205:
Again many thanks, its very important for me, i work a lot with this type of element

samael205:
Compiled and tested, it seems to work well

samael205:
Ok, I tested more thoroughly, and now when you delete a itext element type, and select another element, the object event: selected does not jump.

asturur:
i have to test deeply.
In the old code the events were not removed.
And every element carried a canvas reference after it was removed from canvas.
Also the iTextInstances array was not cleared at all on canvas.clear

Can you explain me carefully what you expected after the itext removal and what happened? is it possible to make a fiddle about it ( with old code is ok also, i will run on new code )

samael205:
The fiddle: https://fiddle.jshell.net/samael205/ojs1wm9c/6/

In this fiddle works fine, but on my code does not work

After i remove iText from the canvas, i expect that the event 'object:selected' jumps when i select other object.

currently not jumps the event.

samael205:
Hello Asturur, i want to fix this issue, because i really need this functionality.

Yesterday, i compiled this version with nodejs and uglify using node build.js modules=ALL.

When the compilation ends i get the fabric.js and the fabric.min.js. Then upload the library to my project, and I found that when an instance i-text is deleted, do not receive the error '_iTextInstances' of undefined.

But with these solution, when i delete any instance i-text, the 'object:selected' not jumps when you try to select other objects. I dont know where is the problem, but it is necessary to fix.

inovek:
Did I understand something wrong, or do I've the same issue like the one mentioned in this topic here?

When I delete ITEXT objects from my canvas and click somewhere on the cancas without selection I get the following error message:

facricjs_i-text_deletion_issue
 
Tried then to copy the code changes of the **static_canvas.class**', but when I click into an empty space I get the same result. (I think that to copy the static_canvas.class part into my build isn't enough, bad/sad that I don't know how to compile these src folders you use)

If this isn't the same issue, I will open a new issue. (please mention it)

samael205:
The two problems arise from the same merge.
Mine is generated after this update (the here mentioned by me)
The @inovek yours, is generated at a rate of [#3320](https://github.com/kangax/fabric.js/issues/3320)

asturur:
replicate issue, i m working on it.

samael205:
Many Thanks!

asturur:
https://github.com/kangax/fabric.js/pull/3345  this should fix.

PR #3335, opened 2016-10-10
FIX for toDataURL cropping

thoepfner:
The new toDataURL code in 1.6.5 uses `cropping.width` to calculate `scaledHeight`. It should use `cropping.height` .

### DIFF
--- a/src/mixins/canvas_dataurl_exporter.mixin.js
+++ b/src/mixins/canvas_dataurl_exporter.mixin.js
@@ -58,7 +58,7 @@
       var origWidth = this.getWidth(),
           origHeight = this.getHeight(),
           scaledWidth = (cropping.width || this.getWidth()) * multiplier,
-          scaledHeight = (cropping.width || this.getHeight()) * multiplier,
+          scaledHeight = (cropping.height || this.getHeight()) * multiplier,
           zoom = this.getZoom(),
           newZoom = zoom * multiplier,
           vp = this.viewportTransform,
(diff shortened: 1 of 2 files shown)


asturur:
Thank you!!!
That means we are missing a test.
Would you try to add it?

The test should load a statiCanvas, then dataurl export it with cropping, reload dataurl in image and verify dimensions.
Maybe there are some examples in our test files.
Would you try to make one?

thoepfner:
I'll try to make a test case, but I cannot get the existing fabric.js tests up and running on my machine. I'll need some time to track that down.

asturur:
what machine do you have?

2016-10-10 16:21 GMT+02:00 Timo Höpfner <notifications@github.com>:

thoepfner:
MacBook Pro, OS X 10.10, installed cario with MacPorts. 

When running `testem`, I'm getting tons of `Uncaught ReferenceError: QUnit is not defined at http://localhost:7357/...` for Chrome and `Cannot read property 'navigator' of undefined` for node. 

I'll try on a different machine tonight.

asturur:
do not run testem, run just ```npm run test```  we have an onging try for other testing, but npm run test ( after npm run build ) is the way to go.

thoepfner:
Added unit test for cropping. 
I should have started this pull request from a topic branch. Shall I create a new one?

asturur:
no no, we need fixes, not issues topics :D

PR #3341, opened 2016-10-11
Contrast & saturation

asturur:
I added those 2 filters we are missing. they are widely used i do not see why we should not have them.


asturur:
Ok i have to fix up jsdocs....

Issue #3384, opened 2016-10-28
[Feature] Event mouse:in


samael205:
Hi asturur,
I want to use in fabric the event "mouse: in" to know when I'm inside the canvas.
Currently works when the mouse leaves the canvas with mouse: out, but do not jump when you walk into the canvas, you Could implement a "mouse: in" to jump when you walk with the mouse on the canvas?.

regards

asturur:
there is a mouse:over event, now i do not remember if it fires everytime the cursor is over the shape or just the first time

http://fabricjs.com/docs/fabric.Canvas.html

does mouse:over work?

samael205:
the mouse:over fires every time when the cursor is over a shape, but doesn't when the cursor is outer the canvas and enter inside canvas

asturur:
ok this could be a nice addition. if we mouse:out on exit canvas, maybe we can mouse:over on canvas enter. i will take a look.

asturur:
closed via #3389  @samael205 get a check if it is ok.

samael205:
Ok, i will try soon, Many Thanks Guys @stefanhayden  && @asturur

samael205:
Hi, works fine!.
Thank you so much guys.
Regards!

PR #3389, opened 2016-10-31
Trigger mouse over on canvas enter

stefanhayden:
As talked about in #3384 this adds the existing `mouse:over` event to fire when the mouse enters the canvas.

If the mouse enters the canvas and there is no target the fire the event. We need to do a check if there is a target, other wise the event will fire 2 times on entering the canvas with an object on the edge. once with null and once with the target,

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -148,10 +148,8 @@
      */
     _onMouseEnter: function(e) {
       if (!this.findTarget(e)) {
-        var target = this._hoveredTarget || null;
-        this.fire('mouse:over', { target: target, e: e });
+        this.fire('mouse:over', { target: null, e: e });
         this._hoveredTarget = null;
-        target && target.fire('mouseenter', { e: e });
       }
     },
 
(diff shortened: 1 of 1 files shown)


src/mixins/canvas_events.mixin.js
```diff
@@ -141,6 +144,19 @@
...
     /**
      * @private
+     * @param {Event} e Event object fired on mouseenter
+     */
+    _onMouseEnter: function(e) {
+      if (!this.findTarget(e)) {
+        var target = this._hoveredTarget || null;
+        this.fire('mouse:over', { target: target, e: e });
+        this._hoveredTarget = null;
+        target && target.fire('mouseenter', { e: e });
```
asturur: i think ```mouseenter``` is our ```mouseover```.

But beside that, if you do not find any target, ```this.findTarget(e)``` why do you want to fire the event for the last hoverdTarget ?

asturur:
Thanks @stefanhayden i left a comment on the code, i m not sure of something.

stefanhayden:
you were right. So we only fire this event if there is no target. we also set the _hoveredTarget  to null or mouse:out reports undefined instead of null for the target.

asturur:
nice thanks!

PR #3406, opened 2016-11-09
fixes #3405 - deleting large chunks of text caused a noticeable lag

### LINKED ISSUE #3405: Deleting large chunks of text causes a significant lag.
## Version
1.6.6

## Test Case
http://jsfiddle.net/2qa39t6g/2/

## Steps to reproduce
1. Click on the text and make sure the cursor is displayed.
2. Hit ctrl+a (cmd+a on macOS) to select all the text.
3. Hit backspace on your keyboard.

## Expected Behavior
The text should be removed immediately.

## Actual Behavior
There is a noticeable lag when removing large chunks of text.
This is caused by a time-expensive call to `this._splitTextIntoLines()` inside `fabric.IText.prototype._removeSingleCharAndStyle` (which is called for every character eligible for removal).

I will send a Pull Request shortly.

belfz:
This is a fix for [issue 3405](https://github.com/kangax/fabric.js/issues/3405).
It also fixes the `tests.mustache` file ([issue](https://github.com/kangax/fabric.js/issues/3404)), because I added unit tests and had to run them.

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -592,21 +592,24 @@
      */
     _removeCharsFromTo: function(start, end) {
       while (end !== start) {
-        this._removeSingleCharAndStyle(start + 1);
+        this._removeSingleCharAndStyle(start + 1, Math.abs(end - start) === 1);
         end--;
       }
       this.selectionStart = start;
       this.selectionEnd = start;
     },
 
-    _removeSingleCharAndStyle: function(index) {
+    _removeSingleCharAndStyle: function(index, shouldRewrapLines) {
+      shouldRewrapLines = typeof shouldRewrapLines === 'undefined' ? true : shouldRewrapLines;
       var isBeginningOfLine = this.text[index - 1] === '\n',
           indexStyle        = isBeginningOfLine ? index : index - 1;
       this.removeStyleObject(isBeginningOfLine, indexStyle);
       this.text = this.text.slice(0, index - 1) +
         this.text.slice(index);
 
-      this._textLines = this._splitTextIntoLines();
+      if (shouldRewrapLines) {
+        this._textLines = this._splitTextIntoLines();
+      }
     },
 
     /**
--- a/tests.mustache
+++ b/tests.mustache
@@ -2,7 +2,7 @@
 <html>
 <head>
 <title>Test'em</title>
-<script src="/bower_components/qunit/qunit/qunit.js"></script>
+<script src="/node_modules/qunitjs/qunit/qunit.js"></script>
 <script src="/testem.js"></script>
 {{#serve_files}}<script src="{{{src}}}"{{#attrs}} {{&.}}{{/attrs}}></script>{{/serve_files}}
 <link rel="stylesheet" href="/bower_components/qunit/qunit/qunit.css"/>
--- a/test/unit/itext.js
+++ b/test/unit/itext.js
@@ -475,6 +475,21 @@
 
   });
 
+  test('removeChars - should call _splitTextIntoLines() only once when removing multiple chars', function () {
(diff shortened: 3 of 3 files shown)


src/mixins/itext_behavior.mixin.js
```diff
@@ -592,21 +592,24 @@
...
+        this._removeSingleCharAndStyle(start + 1, Math.abs(end - start) === 1);
         end--;
       }
       this.selectionStart = start;
       this.selectionEnd = start;
     },
 
-    _removeSingleCharAndStyle: function(index) {
+    _removeSingleCharAndStyle: function(index, shouldRewrapLines) {
+      shouldRewrapLines = typeof shouldRewrapLines === 'undefined' ? true : shouldRewrapLines;
```
belfz: `shouldRewrapLines` is `true` by default.

tests.mustache
```diff
@@ -2,7 +2,7 @@
 <html>
 <head>
 <title>Test'em</title>
-<script src="/bower_components/qunit/qunit/qunit.js"></script>
+<script src="/node_modules/qunitjs/qunit/qunit.js"></script>
```
belfz: Tests are actually fired now. However, some tests seem to fail (CLI report reports 1 failed test, while Chrome browser's result mentions 2 failed tests). They failed before and are not caused by my changes. Please see [this issue](https://github.com/kangax/fabric.js/issues/3404).

asturur:
I have feelings this will break styles.
Can you try this code with style around text? i guess that when you delete chars that spans across lines, style will be broken.
Actually the old code was faster, this splitTextIntoLines was introduced by me both in the paste and delete code.

belfz:
@asturur thanks - I am using this fix in the project I am working on and so far it seems to be working fine.

Characters are being removed one by one, in a loop. There is no need to fire the line-breaking algorithm after each character (`this._splitTextIntoLines()`) - it can be fired only once, at the very end of the loop.

Also, inside `_splitTextIntoLines` method, there is a call to `_generateStyleMap` method, which regenerates the styles map (based on the state of the text).

I have prepared a [test fiddle here](http://jsfiddle.net/LsLno527/3/). You can play around and verify that the fix doesn't break the styles (unless I am missing some case).

asturur:
thanks. i ll give it a look. from my memories is the getcursorLocation, necessary for style shift, that needa that line refactor. textbox will be slower than itext i guess

asturur:
Ok here is the problem.
Your logic works with Texbox but with not IText.

take this fiddle:
http://jsfiddle.net/LsLno527/8/

image

delete text like showed in picture and style will be messed up.
If you manage to make it work pretty i m more than happy to push it in.

belfz:
Thank you @asturur. That's true, it doesn't seem to work correctly with the instances of `IText` (in the project I am working on I use `Textbox`). I might try to look for the solution to make it work in `IText`, but in the meantime let's put this pull request on hold.

asturur:
a speed up could be gained trying to call the slow function for every deleted line.
In textbox with `\n` you will get the same problem.

I guess you can inspect the char deleted, check if it is a `\n` and in that case rewrap lines.

Not really sure, i do not remember everything but i think that before giving up on doing this operation every character, i tried all the first level solutions.

belfz:
Closing this pull request, because the proposed solution breaks the styles in the text.

asturur:
Can't you give a shot at detecting line deletion and call the `expensive` function just in that case?

belfz:
@asturur I gave it a shot in [this updated fiddle](http://jsfiddle.net/LsLno527/10/)

I just changed that `if` statement in `_removeSingleCharAndStyle`:
```
if (isBeginningOfLine) {
    this._textLines = this._splitTextIntoLines();
}
```

Unfortunately, it still doesn't work correctly.

PR #3408, opened 2016-11-10
Word widths on justified line of text were calculated incorrectly.

belfz:
The width of words was incorrectly calculated for justified line of text. Please see the comments to the code changes I committed.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -500,7 +500,7 @@
       // stretch the line
       var words = line.split(/\s+/),
           charOffset = 0,
-          wordsWidth = this._getWidthOfWords(ctx, words.join(''), lineIndex, 0),
+          wordsWidth = this._getWidthOfWords(ctx, words.join(' '), lineIndex, 0),
           widthDiff = this.width - wordsWidth,
           numSpaces = words.length - 1,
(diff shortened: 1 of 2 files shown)


src/shapes/text.class.js
```diff
@@ -500,7 +500,7 @@
       // stretch the line
       var words = line.split(/\s+/),
           charOffset = 0,
-          wordsWidth = this._getWidthOfWords(ctx, words.join(''), lineIndex, 0),
+          wordsWidth = this._getWidthOfWords(ctx, words.join(' '), lineIndex, 0),
```
belfz: Hence, the solution was to preserve the space characters.

asturur:
can you correct the lint error?

I m going to try this code, it looks good.
I have been very busy and i m behind with issue tracker one week.
As soon as i catch up i merge this i think.

itext needs a full rewrite in my opinion.

belfz:
Thank you @asturur! I was about to fix the lint error this morning, but I saw that you already did that and merged the fix. Appreciate that!

asturur:
i wanted to ship in 1.6.7
 i guess 1.7.0 will take longer or will be bugged. so better to have as much as possible in this release

PR #3419, opened 2016-11-14
Modified npmignore for custom builds through npm

devnix:
Right now, if you install Fabric through `npm install`you can't do a build with gestures, which is quite frustrant and completly unnecesary!


DevNIX:
Doing this way allows our projects to keep always up to date from the npm channel, while building our custom version every time. For example:

### package.json
```json
{
  "dependencies": {
    "fabric": "fabric",
  },
  "scripts": {
    "preinstall": "cd node_modules/fabric && nodejs build.js modules=ALL exclude=json",
    "install": "bower install && gulp"
  }
}
```

brayn003:
+1
Can we please merge this? It's a pain downloading custom builds from site everytime

chipbell4:
What's the status on this? This is a big blocker for projects for me as well...

asturur:
i have no idea. are there any other packages that offer built js and source files inside?

chipbell4:
Actually, it appears that jquery does this. [Check out their `.npmignore`](https://github.com/jquery/jquery/blob/master/.npmignore). I'm guessing this also in the spirit of custom builds. A quick `npm install` confirms that their `src` directory is installed.

asturur:
i just do not want to push up twice the thing. either a do a postinstall script that makes the custom build, but that would force you to have uglifyjs installed

i m really undecided. The only point of this looks like that for who is building an app, cloning the repo and doing the custom build is too much of an hassle.

am i right?

DevNIX:
For me, it was a big blocker. I just had to clone my repo just to make a custom build.

I feel the current way like the opposite desired effect :smile: 

PS: A way to build without having installed uglifyjs could be installing it as a local dependency, and using it programatically, even migrate the build system to Gulp.

asturur:
i think this looks merged already or no diffs?

devnix:
Seems like there are no diffs! I guess this should be closed 🤦‍♂ 😂

PR #3420, opened 2016-11-14
Switch to different coordinates for groups

asturur:
make the coords of internal group objects indipendent from canvas viewport.


samael205:
Thanks @asturur!!!

PR #3479, opened 2016-12-01
Update shiftKey & altKey values during __onMouseMove()

stefanhayden:
Currently the values of `shiftKey` & `altKey` are set on mouseDown and only the cached values are referenced during mouseMove events. This means if you start a transform and then want to hold `alt` to `centerTransform` you can't until you stop transforming and hold `alt` before starting the transform.

This change updates the values during mouseMove which lets you start a transform and add the key modifier on the fly.

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -638,6 +638,8 @@
 
       transform.reset = false;
       transform.target.isMoving = true;
+      transform.shiftKey = e.shiftKey;
+      transform.altKey = e[this.centeredKey];
 
       this._beforeScaleTransform(e, transform);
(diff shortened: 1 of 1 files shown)


asturur:
I m guessing if that hardCoded shiftKey is a missed `uniscaleKey`.

stefanhayden:
hmm. I got that shiftKey reference from here: https://github.com/kangax/fabric.js/blob/52013a94d18a6c6ce6ead426176ce8709c6c7ccb/src/canvas.class.js#L625

should that also use the uniScaleKey?

asturur:
not before a full read of what is there.
I remember i spent some time naming the keys insted of having them fixed. Why that shift is still fixed.

PR #3497, opened 2016-12-05
fix includeDefaultValues

### LINKED ISSUE #3491: includeDefaultValues on text and path object (minor priority)
Hi,

I don't want to export default values, but it seems there is still some default values staying if i check the generated JSON. these are the minor bugs/strange behavior i detected.

1)
If i export a "path-group" object, the default properties are correctly filtered out, but in the "PATH" object there is still in the default values (like scaleX/Y or skewX/Y), it looks like the sub object "path" is not affected but the value of the "includeDefaultValues" parameters if included in a "path-group" object.

2)
For the text object, the propertie "charSpacing" is exported even if i don't change it (default = 0). It's a new function perhaps the attribute name is not present in the list to remove used by the includeDefaultValues function. Or is that normal ?

3)
The canvas "background" attribute isn't filtered by the "includeDefaultValues" even when the value is the default value "".

4)
For the image object the "crossOrignin" (default value = "") is exported event if it's not changed is this normal ?

5)
The filters[] and resizeFilters[] are exported even if empty. Idem for the canvas.objects[] when it's empty.
=> If it's normal to keep them ? if yes, is there a way to manually add these attributes to the remove default list attributes ? (i tried to play with the "statesProperties" attribute but without success).

PS : I don't make a fiddle but i could create one or attach a json export if necessary.

Thank you for your help. Keep up the good work.

asturur:
Remove defaultValues from children of groups during group toObject.
Remove canvas background export is empty string to be choerent with overlay and backgroundImage and overlayImage

closes #3491


asturur:
@ncou have a look. and please confirm text charspacing with a fiddle.

PR #3501, opened 2016-12-06
Ensure clone invokes callback in all cases

fabiancook:
> Clones an instance, some objects are async, so using callback method will work for every object. 

This should mean that `callback` is invoked no matter what, however if `this.constructor.fromObject` is not a function then it should never be invoked. 

This solves the issue by invoking `callback` if the value is just being returned. 

`setTimeout` is used to ensure `callback` is invoked in an *async* fashion

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1471,10 +1471,15 @@
      * @return {fabric.Object} clone of an instance
      */
     clone: function(callback, propertiesToInclude) {
-      if (this.constructor.fromObject) {
+      if (this.constructor.fromObject instanceof Function) {
         return this.constructor.fromObject(this.toObject(propertiesToInclude), callback);
       }
-      return new fabric.Object(this.toObject(propertiesToInclude));
+      var cloned = new fabric.Object(this.toObject(propertiesToInclude)); 
+      if(!(callback instanceof Function)) {
+        return cloned;
+      }
+      setTimeout(callback.bind(null, cloned), 0);
+      return cloned;
     },
 
     /**


asturur:
Every subclass needs to have a fromObject method. The fabric Object class is a base class not meant for real use, i do not see a need for this change.
mind to explain further?

fabiancook:
More for consistency, matching what the documentation says, noticed it by chance

No issues if it won't be merged.

asturur:
i guess docs need an update then

fabiancook:
I think the better option would be to throw if the object doesn't have a `fromObject` method? 

If the function is never going to hit that use case, it probably should be pruned.

asturur:
So all the subclasses have a fromObject method, or they can't be cloned.
This Object.class is just a main class to derive subclasses.
If you create a fabricObject subclass without a fromObject method, you did a wrong thing or you did your own personal subclass and you will take care of not cloning it.
There are no subclasses without that method as of now.

fabricjs code is far from perfect, but is enough big and complicated. I would not add other checks or additions if not to fix actual bugs that show up in real use case.

asturur:
just to add, your consideration are correct about code correctness and error throwing and preChecks before calling a function.
I just want to keep it simple.

fabiancook:
No issues,

Please close this pull request if you don't believe it should be in there. 

My thoughts are just that either the code should be removed completely, or to invoke the callback if it was passed.

asturur:
We could correct with 
```
     clone: function(callback, propertiesToInclude) {
       if (this.constructor.fromObject) {
          return this.constructor.fromObject(this.toObject(propertiesToInclude), callback);
       }
       callback && callback(null);
     },
```

this would keep overhead to minimum and save from unexpected behaviour.

fabiancook:
That would work

Just curious, what's wrong with using `instanceof`?

asturur:
nothing.
The docs says that you have to pass a callback function.
Either you pass it ( and is a function ) or you do not pass it so it does not exist.
I just do not want consider other creative alternatives.
If i have to add code to check that devs passes correct types i have to add lot of things i do not want to add.
I do not think there are concerns with performances using instanceOf, typeof or other tricky edge cases.

PR #3510, opened 2016-12-07
Fixed object:modified issues when moving and resizing objects in OSX …

dennisrjohn:
…Sierra.

As discussed in issue 3469.
https://github.com/kangax/fabric.js/issues/3469


asturur:
can you explain the reasons of this changes and why it should affect sierra only?

PR #3528, opened 2016-12-13
Added missing semicolons to canvas.class.js

Newman101:
Hi. 

Looks like there were couple of semicolons missing in canvas.class.js. 

These have been now amended. 

Thank you.


asturur:
thanks

PR #3554, opened 2016-12-20
Improved patterns [WIP]

### LINKED ISSUE #935: Load from JSON is not working with Patterns in the latest version of Fabric.js
When am trying to load the JSON data onto the canvas, every shape is getting loaded but the patterns that are filled in the shapes are loaded only after I click on the shape.

Could some one please explain why this is happening? I even checked this:

canvas2.loadFromJSON(json,canvas2.renderAll.bind(canvas2));
Also, the same problem is with the clone() function: the object is duplicated. But the pattern filled in the object is shown only after I click on the object.

I am using version 1.3.7.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1048517-load-from-json-is-not-working-with-patterns-in-the-latest-version-of-fabric-js?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F23217&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #3499: Pattern streched on scaled SVG (feature request ?)
Hi,

I have a SVG with default width/height = 32pix (it's a square).
I scale this object, and I once the scale is finish i want to apply a pattern image.

The pattern seems to be first applied on the 32x32 size svg then stretched after (the result is a big pattern stretched on all the square).

I think the normal behaviour should be to put the pattern on the shape with the scaled width and height.

I don't know if this could be a possible feature request, somethink somethink like a parameter "applyAfterScale = true" on the pattern object to always have a pattern scale = 1.0 even if the object si scaled X times.

Not sure if i am enought clear.

Here is a fiddle with on the left the scaled square, and on the right a scare without scale only width/height
http://jsfiddle.net/QT3pa/154/

What are you thinking about this ?

asturur:
Add ability to loadfromJSON to wait for patterns load
Add additional controls to patterns ( like angle, scale factors, and possibiloty to do not stretch with object )

close #935
close #3499


Ramyakutralam:
Hi , i m using fabric 2.2.3. I 'm trying to apply pattern for the selected objects like rect and circle. but the problem is while scaling the rectange image is also being stretched. Is there any way i can restrict it?

Thanks in advance

PR #3596, opened 2016-12-29
fix space for cache canvas for thin lines

### LINKED ISSUE #3590: Freedrawn lines sometimes rendered wrong
I have a problem which sometimes also occurs on the demo page.
When drawing a freehand, the line is flattened at some point (see blue one in image). This line supposed to be like the other two, with a little curve.

Is this something of the canvas element itself or something in fabric?

I am using firefox.

rendererror

asturur:
fixes #3590


asturur:
need a better test for text. then i'll merge.

PR #3598, opened 2016-12-30
WIP Patternload

### LINKED ISSUE #935: Load from JSON is not working with Patterns in the latest version of Fabric.js

asturur:
close #935

create a fromObject function that can wait for patterns in stroke and fill to be loaded.
Make all objects able to be loaded with a callback


mariotacke:
This is a breaking change from 1.7.2 to 1.7.3. We had the following code:

```
fabric.Image.filters.Thresholding.fromObject = function (object) {
  return new fabric.Image.filters.Thresholding(object);
};
```

This no longer works. Is the required callback documented anywhere? I was not able to see this change via the release changelog.

asturur:
oh well, i think you are right. for custom saved json this may be a problem. you can also return the instance for normal restoring.
I should have switched to 1.8.0 for this change at least.

i will.come back with more example and docs

asturur:
for example the base filter fromObject will serve the purpose of any filter restore unless you have some special code.

So in your case if the Thresholding filter is of type `Tresholding`

`fabric.Image.filters.Thresholding.fromObject = fabric.Image.filters.BaseFilter.fromObject;`

would be enough.

I will add this notice in the changelog, not doing it was a fault.

mariotacke:
I appreciate the transparency! Will adjust our code to your suggestion!

PR #3615, opened 2017-01-09
Color.class update

neopheus:
New Unit Test
JSDoc Update
New toHexa() function


src/color.class.js
```diff
@@ -185,6 +185,29 @@
...
+      g = (g.length === 1) ? ('0' + g) : g;
+
+      b = source[2].toString(16);
+      b = (b.length === 1) ? ('0' + b) : b;
+
+      a = source[3] * 255;
+      a = a.toString(16);
+      a = (a.length === 1) ? ('0' + a) : a;
+
+      return r.toUpperCase() + g.toUpperCase() + b.toUpperCase() + a.toUpperCase();
```
asturur: all this could be `return this.toHex() + a.toUpperCase()` so we do not need to duplicate toHex code again.

asturur:
Thanks for the change, could you address my comment?

src/color.class.js
neopheus: It's do :)

PR #3656, opened 2017-01-28
Added an additional check for the blur event on hiddenTextarea

CristiScorteanu:
When having a Textbox in the canvas with some text in it and `canvas.deactivateAllWithDispatch()` is called the `this.hiddenTextarea` might be null sometimes.

This check is also made on the following line so it made sens to also do it here.
Hope this helps


asturur:
would you put those statements under one if?

```
if (this.hiddenTextarea) {
 statement 1
statement 2
this.hiddenTextarea = null;
}
```

CristiScorteanu:
Thanks for the tip. I added it.

PR #3676, opened 2017-02-08
text rewriting - reorganization

asturur:
Text will get all property of itext and textbox.
IText will be just an interactive layer
Textbox will disappear

aim of the change:
mantain same functionality
better code readibility
better text dimension cache ( cache is stored at 40px dimensions per font family per character )
support grapheme of 2 bytes ( and maybe more )
put underline before text, overline and line-trought after char.

later:
rework input system to fully support composition


asturur:
before merge list:

verify svg output
verify justified svg output
verify 24 failing tests
decide and implement methods for programmatic text insertion
add test for new methods

after merge:
add methods for style manipulation ( cleaning, flattening if possible )

asturur:
i merge this as it is. Tests pass.
Some tests disabled since are no more pertinent.
There is one bug: pasting with styled text on multi lines create some defect.

PR #3690, opened 2017-02-13
passivefalseeverywhere

### LINKED ISSUE #3687: Chrome update 55 in Android - fabric js zoom breaking
After updating to the latest version of chrome 55 in Android (mine is HTC 816, but the issue is reflecting in other devices as well), pinch zoom functionality is broken in the latest version of fabricjs.

Whats new in chrome update 55 related to touch events is here - https://developers.google.com/web/updates/2016/11/nic55

A similar issue was there in the jquery panzoom library after the latest chrome 55 update - https://github.com/timmywil/jquery.panzoom/issues/310

On zooming an object on canvas, the entire browser responds to the event and zooms instead. 
@kangax  @asturur any clue about this issue?

asturur:
close #3687

Add { passive: false } to every touch event.
Add relevant code to bring down the option object to the actual addEventListener function
Add check to make sure old browser do not get the option object ( that would equal to `useCapture` true )

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -49,8 +49,8 @@
       addListener(this.upperCanvasEl, 'contextmenu', this._onContextMenu);
 
       // touch events
-      addListener(this.upperCanvasEl, 'touchstart', this._onMouseDown);
-      addListener(this.upperCanvasEl, 'touchmove', this._onMouseMove);
+      addListener(this.upperCanvasEl, 'touchstart', this._onMouseDown, { passive: false });
+      addListener(this.upperCanvasEl, 'touchmove', this._onMouseMove, { passive: false });
 
       if (typeof eventjs !== 'undefined' && 'add' in eventjs) {
         eventjs.add(this.upperCanvasEl, 'gesture', this._onGesture);
@@ -199,7 +199,7 @@
     _onMouseDown: function (e) {
       this.__onMouseDown(e);
 
-      addListener(fabric.document, 'touchend', this._onMouseUp);
+      addListener(fabric.document, 'touchend', this._onMouseUp, { passive: false });
       addListener(fabric.document, 'touchmove', this._onMouseMove, { passive: false });
 
       removeListener(this.upperCanvasEl, 'mousemove', this._onMouseMove);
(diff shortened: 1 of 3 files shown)


src/mixins/canvas_events.mixin.js
```diff
@@ -228,8 +228,8 @@
       removeListener(fabric.document, 'mousemove', this._onMouseMove);
       removeListener(fabric.document, 'touchmove', this._onMouseMove);
 
-      addListener(this.upperCanvasEl, 'mousemove', this._onMouseMove);
-      addListener(this.upperCanvasEl, 'touchmove', this._onMouseMove);
+      addListener(this.upperCanvasEl, 'mousemove', this._onMouseMove, { passive: false });
```
Rycochet: Only needed on touch events, so this one is redundant :-)

asturur: right!
copy paste speed error

samuelkilada:
Are we sure this is fixed? Chrome seems to be scrolling the window now when you use touch and drag. I confirmed this only started happening after updating to the latest version of Chrome on an Android phone. I just tested with fabricJS 1.7.6.

asturur:
I have nothing scrollable, so i cannot really confirm. i fixed reading the specs from the mdn/ and chrome website.

i ll get a scrollable device soon

samuelkilada:
Thank you. FYI, if you open this demo on and Android phone and zoom in a bit until the window can scroll up and down, you can see the error: http://fabricjs.com/touch-events

For some reason, this doesn't happen in the kitchen sink demo!

asturur:
please check the version of that demo. i guess is based on fabric 1.5

samuelkilada:
Interesting! Kitchen sink uses http://fabricjs.com/lib/fabric.js, which is currently 1.7.6 and uses all modules except gestures and a couple others!

`/* build: `node build.js modules=ALL exclude=json,gestures minifier=uglifyjs` */`

The broken touch events demo uses http://fabricjs.com/lib/fabric_with_gestures.js, but yes it is 1.5.0, which explains it. I wonder if I should not use the gestures module since in the kitchen sink  dmeo it seems to avoid the error without that module and works fine on touch. :S

asturur:
gesture module is for pinch and rotate i guess. i will update that demo
page with the modern build as soon i update the website for 1.7.7

samuelkilada:
Well, I must say I just made a really stupid mistake. In thinking that I was applying the updated version of Fabric I was actually not. 1.7.6 does indeed fix the problem Chrome introduced. I think the touch events demo running on 1.5.0 is what threw me off. Thanks very much for the quick responses.

PR #3709, opened 2017-02-17
Keep default values for fill and stroke

Kienz:
If ```includeDefaultValues = false``` the fill and stroke values are overwritten by undefined if no values are defined.

See:
http://jsfiddle.net/Kienz/0mcb5k58/

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1814,8 +1814,17 @@
     object = clone(object, true);
     if (forceAsync) {
       fabric.util.enlivenPatterns([object.fill, object.stroke], function(patterns) {
-        object.fill = patterns[0];
-        object.stroke = patterns[1];
+        // If fill or stroke is undefined use prototype property defaults
+        delete object.fill;
+        delete object.stroke;
+
+        if (patterns[0]) {
+          object.fill = patterns[0];
+        }
+        if (patterns[1]) {
+          object.stroke = patterns[1];
+        }
+
         var instance = extraParam ? new klass(object[extraParam], object) : new klass(object);
         callback && callback(instance);
       });


asturur:
Hey @Kienz we did some recent change to includeDefaultValues  since @ncou found out that most of the time many defaults values were there.
What exactly did you experience?

Kienz:
Hey @asturur, you can see it in the jsfiddle:
1. Press button "saveToJson"
2. Press button "loadFromJson"
3. The drawn object is empty, but you should see a small "o" and a small "e" (like if you reload the jsfiddle)

I also recognize that fabric.Object.prototype.includeDefaultValues is lost while calling ```canvas.toJSON()```. You can see it in the jsfiddle http://jsfiddle.net/Kienz/9z2m7Ly6/.
Only if both includeDefaultValues properties are ```true``` the object is correct after restored from json.
If ```fabric.Canvas.prototype.includeDefaultValues = false``` and ```fabric.Object.prototype.includeDefaultValues = true``` the object is incorrect after restored from json.

asturur:
i ll take a deep look.
@ncou did you experience anything like that?

Kienz:
I think we should check ```patterns[0]``` and ```patterns[1]``` in ```fabric.Object._fromObject``` to be not undefined.

```
    if (forceAsync) {
      fabric.util.enlivenPatterns([object.fill, object.stroke], function(patterns) {
        if (typeof patterns[0] !== 'undefined') {
          object.fill = patterns[0];
        }

        if (typeof patterns[1] !== 'undefined') {
          object.stroke = patterns[1];
        }

        var instance = extraParam ? new klass(object[extraParam], object) : new klass(object);
        callback && callback(instance);
      });
    }
```

Otherwise the prototype chain for ```object.fill``` and ```object.stroke``` is destroyed.

asturur:
that is completely true. if stroke is red from prototype, includedefaultvalue false, pattern is not a pattern because the value was omitted during export,we break the prototype value.

would you mind opening a pr with the fix and merge it? if you have no will for tests i ll add them later

asturur:
use squash and merge button i have to cherry pick the fix in the 1.x branch.

PR #3724, opened 2017-02-20
Update fabric.js

by12:


asturur:
i see it now. i was seraching it in the path class while is pathgroup. you have to submit the change to the PathGroup.class.js not to fabric.js file

PR #3729, opened 2017-02-22
Add the functionality of draw controls with custom methods

vudduu:
Added custom draw functions for resize-controls and rotation-control.
It will enable users to draw, animate or customize the controls with their own logic.

Next is an small sample of how to use it.
**[JSFIDDLE DEMO](https://jsfiddle.net/vudduu/7Lfjx90t/)**


neopheus:
Jsfiddle Update : https://jsfiddle.net/7Lfjx90t/1/

asturur:
There are resons why we still do not have this.

First of all this implementation is good.

It open some problems:

if i set type to custom and i want to change just one control, i m forced to get the rect for the others.
With custom drawing per controls may come need to define custom function per control 

So while a single custom type for every corner could be ok, going in the specification of each corner seems to much for me now.

I m open to merge this ( for the 2.0 branch that is the current master ) if we skip the per corner specification ( either rect, circle or custom for all the corners ) and if we add a small example on how to draw a corner that is not a circle and not a rect ( either a cross or an icon or whatever you feel like )

vudduu:
Hi @asturur thank you for your help.

A custom draw method per control opens multiple features while animating corners. I'm using it in another project to have a different icon on each corner, animate size on click event and animate color on hover event.
[DEMO CUSTOM COLOR ON OVER OR WHEN CLICKING A CORNER](https://jsfiddle.net/vudduu/7Lfjx90t/3/)

asturur:
i know, but still who wants to go radically different can go with the override of some functionality.

Defining a custom control for each corner in this way is not good.
Either the control type ( circle, rect, custom ) gets per corner, or the drawing function gets per object.

I prefer to stick to the per object model.
To draw the actual different corner can be done with current code since this._corner keep the reference to the current hovered corner ( i m actually using it ) i prefer to define activeCornerColor and hoveredCornerColor properties.

As usually here is matter of personal preferences, not really about working or good code.

Let's think about it i'm fairly open to it, i will also discuss with some people that are using fabric on a daily basis and may have opinion on it.

vudduu:
My mistake, yes we need just one draw method. I will update the new changes.

asturur:
There are many little things to fix here and there.
For example fabric recognize the target over just on the outer part.
Fabric does not rerender on hover, and it should not do it automatically, so maybe hoverTargetColor is not really an option out of the box.

vudduu:
I changed the logic to use the drawCustomControlShape method declaration per object.
[DEMO DRAW A TRIANGLE FOR EACH CORNER](https://jsfiddle.net/vudduu/7Lfjx90t/)

asturur:
@ncou read here for custom controls. we are evaluating what to do here

ncou:
hi,

nice feature. I see it's to define one function to render ALL the controls.
It's a usefull feature to avoid to override the _drawControls() function, but i think it's could more flexible to define a control type ( circle, rect, custom ) sets per corner, so this way you can customize only one control rendering.

The possibility to define the drawing controls function for the object is interesting, but i could bet if you add this feature very soon someone will ask a feature to customize only one control and not all the controls.

In any case as i said it's a usefull feature, and thank you for your contribution.

iongion:
@ncou This is so awesome, as I see, it acts as an iterator/decorator, you are passed the control, so you can decide with an `if`, if it should be drawn or not, or draw all the same but one .. etc

The problem appears, what if I want to also draw my own guides (such as live indicators of width and height..etc)

What would be an option to solve both needs using the decorator pattern as above ?
Currently, overriding the `_drawControls` is a bit obscure ... so reaching a place where controls drawing can be hijacked at a higher level would be nice.

ibrcic:
Sorry for resurrecting such an old issue, but can someone point me in direction of documentation for custom controls? I see that PR is merged, but still can't figure out how to actually create custom controls?

Fiddles have normal controls.

asturur:
there is a whole ongoing project for customizing controls. As soon as i know when i can release it i ll share api details for feedback.

ibrcic:
Thanks! Until then should I just override drawControls() or should I somehow use this PR implementation? Essentially I just want to change rotation control icon.

asturur:
i would do what is simpler. the new api is substantially different and i
trying to understand some things yet.

PR #3742, opened 2017-02-24
Fix problem on fabric.Line controls and line.getWidth()

vudduu:
There is a bug when creating a fabric.Line + strokeWidth.
- There is a padding in the control.
- getWidth returns width + strokeWidth.

[DEMO MASTER BRANCH](http://jsfiddle.net/vudduu/nj29m5qu/)
[DEMO WITH FIX](http://jsfiddle.net/vudduu/qehwu2mj/)

### DIFF
--- a/src/shapes/line.class.js
+++ b/src/shapes/line.class.js
@@ -167,10 +167,11 @@
         //  Line coords are distances from left-top of canvas to origin of line.
         //  To render line in a path-group, we need to translate them to
         //  distances from center of path-group to center of line.
-        var cp = this.getCenterPoint();
+        var cp = this.getCenterPoint(),
+            offset = this.strokeWidth / 2;
         ctx.translate(
-          cp.x - this.strokeWidth / 2,
-          cp.y - this.strokeWidth / 2
+          cp.x - (this.strokeLineCap === 'butt' && this.height === 0 ? 0 : offset),
+          cp.y - (this.strokeLineCap === 'butt' && this.width === 0 ? 0 : offset)
         );
       }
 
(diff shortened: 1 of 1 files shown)


asturur:
it has to be like that.
we cannot change this or a lot of things break.

asturur:
There is an error,
that is this:

image

line has its own getDimensions function.
What is wrong is that we check for dim.x and dim.y while in that if we should check for this.width and this.height.

This should be fixed.
Would you do that?

vudduu:
I made the changes.
Now it adds the changes in the Line getDimensions function instead of the Objects one.

asturur:
i m working a bit on it. there are problems.here and there importing svgs with lines.

i ll push up changes asap

asturur:
To me the changes i committed to your branch looks like are fixing all cases ( horizontal line, vertical line, svg import )

asturur:
```xml
<svg width="120" height="120" viewBox="0 0 120 120"
    xmlns="http://www.w3.org/2000/svg">

  <line x1="15" y1="15" x2="100" y2="15"
      stroke-width="30" stroke="black" stroke-linecap="round" />
    <line x1="0" y1="47" x2="115" y2="47"
      stroke-width="30" stroke="black" stroke-linecap="butt" />
      <line x1="8" y1="79" x2="115" y2="150"
      stroke-width="30" stroke="red" stroke-linecap="butt" />
</svg>
```

asturur:
Please check them, i will merge them once you figure out if they solve what you noticed wrong.

vudduu:
Last changes solved the problem except for **stroke-linecap="round"**
[DEMO LAST CHANGES - PROBLEM WITH ROUNDED LINES](http://jsfiddle.net/vudduu/qq37kd76/)

asturur:
what is that fiddle? i see a snake growing. where is the problem?

asturur:
if you refer to the fact that getwidth is bigger than width, that is because getWidth should be named differently. i will just remove that method.

asturur:
getWidth return the actual width, including scaling, skewing. is just a wrong named method.

it was a shortcut to get width * scaleX, now is a shortcut for getNonTransformedDimensions().x and soon removed.

we should merge this PR i think

PR #3743, opened 2017-02-25
Remove some code from ITEXT, do not pollute styles if not necessary

asturur:


src/mixins/itext_behavior.mixin.js
```diff
@@ -466,9 +466,7 @@
           lineIndex = cursorLocation.lineIndex,
           charIndex = cursorLocation.charIndex,
           charHeight = this.getCurrentCharFontSize(lineIndex, charIndex),
-          leftOffset = (lineIndex === 0 && charIndex === 0)
-                    ? this._getLineLeftOffset(this._getLineWidth(this.ctx, lineIndex))
-                    : boundaries.leftOffset,
+          leftOffset = boundaries.leftOffset,
```
asturur: boundaries.leftOffset and this._getLineLeftOffset(this._getLineWidth(this.ctx, lineIndex)) carry the same value. no need to make a special case.

src/mixins/itext_behavior.mixin.js
```diff
@@ -672,10 +670,6 @@
 
       this.shiftLineStyles(lineIndex, +1);
 
-      if (!this.styles[lineIndex + 1]) {
-        this.styles[lineIndex + 1] = {};
-      }
-
```
asturur: no need to create an empty style object for a line just because maybe we will put something into it.

src/mixins/itext_behavior.mixin.js
```diff
@@ -685,21 +679,24 @@
...
+          var numIndex = parseInt(index, 10);
+          if (numIndex >= charIndex) {
+            somethingAdded = true;
+            newLineStyles[numIndex - charIndex] = this.styles[lineIndex][index];
             // remove lines from the previous line since they're on a new line now
             delete this.styles[lineIndex][index];
           }
         }
-        this.styles[lineIndex + 1] = newLineStyles;
+        somethingAdded && (this.styles[lineIndex + 1] = newLineStyles);
```
asturur: create the new line style only if necessary

src/mixins/textbox_behavior.mixin.js
```diff
@@ -93,24 +91,10 @@
...
-        if (numericLine > lineIndex) {
-          this.styles[numericLine + offset] = clonedStyles[numericLine];
-
-          if (!clonedStyles[numericLine - offset]) {
-            delete this.styles[numericLine];
-          }
-        }
-      }
-      //TODO: evaluate if delete old style lines with offset -1
+      fabric.IText.prototype.shiftLineStyles.call(this, lineIndex, offset);
```
asturur: the prototype function does exactly the same thing

PR #3754, opened 2017-02-28
Fix textarea focus bug when canvas is in an element that stops events from reaching the body

kghandi:
When using bootbox to create a modal dialog, the dialog has a tabindex=-1 to prevent focusing on items outside of the dialog. Therefore the hiddenTextArea must be inside the dialog to be able to get focus.

See example: https://jsfiddle.net/kghandi/jhd6xn4x/

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -24940,7 +24940,8 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     var style = this._calcTextareaPosition();
     this.hiddenTextarea.style.cssText = 'position: absolute; top: ' + style.top + '; left: ' + style.left + ';'
                                         + ' opacity: 0; width: 0px; height: 0px; z-index: -999;';
-    fabric.document.body.appendChild(this.hiddenTextarea);
+    
+    this.canvas.upperCanvasEl.parentNode.appendChild(this.hiddenTextarea);
 
     fabric.util.addListener(this.hiddenTextarea, 'keydown', this.onKeyDown.bind(this));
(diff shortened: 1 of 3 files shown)


asturur:
can you verify that internet explorer 10 and 11 can still copy paste good?
There was a problem for which we were keeping the textarea on the body.

asturur:
Other than that calcTextArea position is written to consider body position not canvas container

kghandi:
I just create a 2nd fiddle to point to my patched version so we can test.

kangax version: https://jsfiddle.net/kghandi/jhd6xn4x/
kghandi version: https://jsfiddle.net/kghandi/vcujr8no/

As far as I can tell copy/paste is working correctly in Chrome 56, Safari 10, Edge, and IE 11. I don't currently have IE 10 handy.

You are correct about the position offset being different. I can see this in the fiddle. But since the element is hidden I am not sure that this is significant.

asturur:
is significant since you start to use composition features ( chinese, japanese, other languages )

asturur:
you have to restore the fabricjs file to the original,
put a check before this.canvas to really check that the Itext is on some canvas.
Infact 4 tests are failing.
UpperCanvasElement is not said to be available since you could be using an Itext on an StaticCanvas, use lowerCanvasEl, the parent is the same.

kghandi:
Thank you for the feedback. I think all the issues you pointed to are fixed.

asturur:
check #3137 i think i have to revert this before other investigations

philipstratford:
I'm having exactly this problem with a Bootstrap modal, but implementing the changes in this now-reverted PR don't resolve the problem for me. In fact, even the fiddle for the patched version linked above by @kghandi doesn't work for me - I'm still unable to type text in the text area in the fiddle's modal.

PR #3774, opened 2017-03-09
Update path.class.js

by12:
Bug Fixed Request. Can you please fixed a small bug in path class.

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -684,8 +684,6 @@
             break;
 
           case 'C': // bezierCurveTo, absolute
-            x = current[5];
-            y = current[6];
             controlX = current[3];
             controlY = current[4];
             bounds = fabric.util.getBoundsOfCurve(x, y,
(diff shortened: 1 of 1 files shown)


asturur:
Sorry i missed this PR.
What the bug fixes? i would be super happy to merge it if i know what does it change.
Any screenshot available?

by12:
Hi,
    Only the problem here is you are using to time the same x and y. But normally for the Bezier Curve we need the  fabric.util.getBoundsOfCurve( x, y,  current[1],  current[2],  current[3],  current[4],  current[5],  current[6]);

But in your case you have 
x = current[5];
y = current[6];
controlX = current[3];
controlY = current[4];

and then fabric.util.getBoundsOfCurve(x,  y,  current[1],  current[2] ,  controlX,  controlY,  x,  y);
so its mean you always have the same start and end points.   current[5],  current[6],
  
fabric.util.getBoundsOfCurve(current[5],  current[6],  current[1],  current[2],  current[3],  current[4],  current[5],  current[6]);

which is making problems

asturur:
ohhh this could be the error i was searching for so long!!!!!

any idea why on the left screenshot the curve do not touch the left side of the bbox?

by12:
Hi,
   Screen shots were from quick and dirty modifications. space on left side was some thing else not because of fabric. There are new screen shot with fabric 1.7.8.
untitled
 
So everything looked fine now. 
Have a nice day

PR #3788, opened 2017-03-17
fix(dependencies): move optional deps to peer

cogell:
purpose: currently, downstream users of fabric.js needed to either:
1. install cairo
2. run `npm install --no-optional`

or those downstream users faced a failing npm install that blocked most
CI usage.

With this change, downstream users of frabric.js can `npm install` and
continue without any errors. If the downstream users require use of
node-canvas then they can explicitly install those deps (as npm will
spit out an unmet peer dependency warning after `npm install`)


asturur:
as of now with canvas prebuilt this is no more true, is just a dependency like another.

PR #3844, opened 2017-04-11
fixing callSuper

sirrodgepodge:
At present callSuper can go into an infinite loop if it is called above the bottom level in the prototype chain, this fixes that.

This fix allows fixing the IText class to run `callSuper('onDeselect')`.  

Also placing an empty function on the `fabric.Object` class in case there is no `onDeselect` defined on the parent.

thank you for your help @asturur

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -18,6 +18,7 @@
     onDeselect: function() {
       this.isEditing && this.exitEditing();
       this.selected = false;
+      this.callSuper('onDeselect');
     },
 
     /**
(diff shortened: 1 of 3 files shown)


asturur:
what the hell, once o renamed `currentContext` with `_this` it started an infinite loop

sirrodgepodge:
hmmmm yea I can't see any reason that this would happen. I actually tried making this change in editor fabric just to make sure I wasn't crazy.  That leaves either the build process for this project or some edge case failure as culprits.

asturur:
canvas prebuilt for node linux x64 has been updated and locked down something.

Issue #3822, opened 2017-04-03
Fabric hosted in Android WebView misses touch events sporadically


daveespo:
## Version
1.7.9 (the current version being hosted in the demos)

## Test Case

When loading the following URL on the Chrome (or system) Web Browser on Android (Marshmallow, Nougat, etc), the ability to use a stylus to capture handwriting is responsive and accurate

http://fabricjs.com/freedrawing

However, loading the same URL in a Cordova or other Android app that hosts an android.webkit.WebView, the same implementation of FabricJS loses glyphs when writing with a stylus

This is a sample Android app that loads the same URL:

AndroidManifest.xml
````
<?xml version="1.0" encoding="utf-8"?>
<manifest package="xyz.web"
          xmlns:android="http://schemas.android.com/apk/res/android">

	<uses-permission android:name="android.permission.INTERNET" />
	<application
		android:allowBackup="true"
		android:icon="@mipmap/ic_launcher"
		android:label="@string/app_name"
		android:roundIcon="@mipmap/ic_launcher_round"
		android:supportsRtl="true"
		android:theme="@style/AppTheme">
		<activity android:name=".MainActivity">
			<intent-filter>
				<action android:name="android.intent.action.MAIN"/>

				<category android:name="android.intent.category.LAUNCHER"/>
			</intent-filter>
		</activity>

	</application>

</manifest>
````

activity_main.xml
````
<?xml version="1.0" encoding="utf-8"?>
<LinearLayout
	xmlns:android="http://schemas.android.com/apk/res/android"
	xmlns:app="http://schemas.android.com/apk/res-auto"
	xmlns:tools="http://schemas.android.com/tools"
	android:layout_width="match_parent"
	android:layout_height="match_parent"
	tools:context="xyz.web.MainActivity">

	<WebView
		android:id="@+id/web_view"
		android:layout_width="match_parent"
		android:layout_height="match_parent"
		tools:layout_editor_absoluteX="8dp"
		tools:layout_editor_absoluteY="8dp"/>

</LinearLayout>
````

MainActivity.java
````
package xyz.web;

import android.app.Activity;
import android.support.v7.app.AppCompatActivity;
import android.os.Bundle;
import android.view.Window;
import android.webkit.WebChromeClient;
import android.webkit.WebView;
import android.webkit.WebViewClient;
import android.widget.Toast;

public class MainActivity extends AppCompatActivity {

	private WebView webView;
	@Override
[...]

PR #3860, opened 2017-04-21
Fixes touchmove event being missed when using on Android/Chrome

homerlex:
This fixes the behavior reported in #3822.  When using Chrome on Android writing/drawing was not "smooth" and parts of drawing would be dropped.

With this in place free drawing is much smoother and no glyphs are missed.

I'm interested in hearing if 
- there is a different solution
- this might cause any ill-effects (I saw none testing with various devices and browsers)
- should any other event handlers preventDefault()

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -204,6 +204,7 @@
      * @param {Event} e Event object fired on mousedown
      */
     _onMouseDown: function (e) {
+      e.preventDefault();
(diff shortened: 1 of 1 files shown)


asturur:
i wish i could help. I really have no idea what we could achieve with this.
I see that preventing the default will probably stop some scrolling start behaviour.
I know this can be obtained also by some css property on the canvas.

Do you have any profiling with this prevent on and off? what is the browser not executing to get smoother action?

PR #3879, opened 2017-04-24
Cursor disappears after text entry

stefanhayden:
calling `this.abortCursorAnimation()` means the cursor does not get restarted. `this.initDelayedCursor();` both aborts and starts it again.


asturur:
thanks

PR #3896, opened 2017-04-28
Fix rendering justified text with line breaks and last line

pavelKozlov:
Here is a linked issue: https://github.com/kangax/fabric.js/issues/3444.

The problem is that text with textAlign == 'justify' also stretches lines with line breaks and the last line in text, but these lines should be rendered as left aligned.

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -792,15 +792,15 @@
      * @param {Number} left
      * @param {Number} top
      * @param {Number} lineIndex
-     * @param {Boolean} withBreak Detects if line contains line break at the end or it's the last line in text
+     * @param {Boolean} isLastLine Detects if the line is last in the text
      */
-    _renderTextLine: function(method, ctx, line, left, top, lineIndex, withBreak) {
+    _renderTextLine: function(method, ctx, line, left, top, lineIndex, isLastLine) {
       // to "cancel" this.fontSize subtraction in fabric.Text#_renderTextLine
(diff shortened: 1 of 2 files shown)


asturur:
I see some problem here: in Itext all lines are with break, so justify will stop working?

pavelKozlov:
Are you sure? I tested it on class that extends fabric.Textbox in project that I'm working on and it works like a charm.
I get line breaks using `this.text` split by `_reNewline` and compare it to `_textLines`.

asturur:
does it work on simple.ITEXT class?

pavelKozlov:
I think, if text with text align justify has every line contains line break, it should be render as left-aligned.
That's how it acts in text editor such as Word or Google Docs.

asturur:
yes but this is out it always worked here. i do not want to change this since application depends from it. i m ok with last line not being justified but if we insert a sort of paragraph meaning to the hard newline is too much change.

pavelKozlov:
But that's how justify works. What if use decide to add text with a few paragraphs?
Anyway, it's up to you to decide.
Last line is also improvement.
Please, let me know if I need fix to work only for last line.

pavelKozlov:
@asturur, I have changed code to left align only last line in justified text.

asturur:
sorry to let pass all this time.
I realize now we do not have svg export functionality for this.
You can add or i will add as soon as i can

jilster:
See https://github.com/kangax/fabric.js/pull/4441

PR #3953, opened 2017-05-23
hoverCursor for selectable: false object

andrea11:
Allow to display custom cursor also for unselectable items (useful for example if you want to suggest a possible event on mouseclick)


asturur:
yes this is right. it was an error to insert that additional condition.

Maybe we just need some more cursors.

PR #3980, opened 2017-06-06
Inheritance breaking minor bug regarding textbox behavior objecting scaling override

samuelhorwitz:
Hey all, locally I have a very heavily extended version of Fabric (utilizing `klass` subclassing). I was noticing some strangeness regarding `this` bindings when attempting to subclass and override certain scaling operations and traced it back to this. It seems to be an oversite that doesn't manifest unless overriding the scaling logic in a subclass.

### DIFF
--- a/src/mixins/textbox_behavior.mixin.js
+++ b/src/mixins/textbox_behavior.mixin.js
@@ -18,7 +18,7 @@
       }
     }
     else {
-      return setObjectScaleOverridden.call(fabric.Canvas.prototype, localMouse, transform,
+      return setObjectScaleOverridden.call(this, localMouse, transform,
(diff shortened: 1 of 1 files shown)


asturur:
yeah that could be a solution.
There are another 2 places like that.

I have a question, overriding with the prototype or with the instance, retain the same results? What are the values of the properties of the object in the context of function call?

If i set this.fill to my instance and then i call the function using fabric.Object.prototype, this.fill what value has in that function execution?

Other than that, please be carefull of fabric 2.0 it brings lot of improvements on the text side, but has very low compatibility with text  custom classes working on 1.x

What version did you extend?

PR #3983, opened 2017-06-07
stop double binding

### LINKED ISSUE #3971: Calling canvas.initialize() more than once leaks window "resize" event listeners, and thus entire canvases
## Version
1.7.6

## Test Case
http://jsfiddle.net/vcubzdgo/1/

## Steps to reproduce
Load http://jsfiddle.net/vcubzdgo/1/ on iPad

## Expected Behavior
Page finishes loading.

## Actual Behavior
Safari web content is jettisoned because of high memory use.

This was reported as a WebKit issue (https://bugs.webkit.org/show_bug.cgi?id=172728) and the analysis we did suggests a bug in fabric.

The fiddle is creating canvases in a loop like this:

      canvas = new fabric.Canvas();
      canvas.initialize(canvasElement, {height:2000, width: 2000})

This pattern causes _initEventListeners() to get called twice for each canvas object, so _bindEvents() is called twice. On the first call, _onResize is bound:

`this._onResize = this._onResize.bind(this); `

and then registered as an event listener on the window:

`addListener(fabric.window, 'resize', this._onResize);`

On the second call, the binding and registration happens again, which will register a second event listener on 'resize' on the window object. However, disposal only removes the second one:
`removeListener(fabric.window, 'resize', this._onResize);`

so the code leaks a registered event listener on the window object. This listener keeps the entire canvas object alive, which in turn keeps its <canvas> elements and their large backing stores alive.

The fix should be to make it harmless to call canvas.initialize() more than once, by saving some state about whether event listeners have been registered.

asturur:
close #3971 

@smfr what do you think about this approach?

1) remove listener before binding ( first time is useless i know )
2) also avoid rebind more than once.

I do not like the flagged solutions, but is the minimum impact solution i can think of.
Another solution would be keeping a reference to the first bind function to window.resize and remove that listener only, but that would not save leaks from double initialize on the same canvas element ( who know why they should happen ) that would mean we would have 2 mouseDown, 2 mouseUp and so on.


asturur:
```[ [Function: bound _onMouseDown],
  [Function: bound _onMouseMove],
  [Function: bound _onMouseUp],
  [Function: bound _onResize],
  [Function: bound _onGesture],
  [Function: bound _onDrag],
  [Function: bound _onShake],
  [Function: bound _onLongPress],
  [Function: bound _onOrientationChange],
  [Function: bound _onMouseWheel],
  [Function: bound _onMouseOut],
  [Function: bound _onMouseEnter],
  [Function: bound _onContextMenu] ]
```
  
after 2 initializes:

```
 [ [Function: bound bound bound _onMouseDown],
  [Function: bound bound bound _onMouseMove],
  [Function: bound bound bound _onMouseUp],
  [Function: bound bound bound _onResize],
  [Function: bound bound bound _onGesture],
  [Function: bound bound bound _onDrag],
  [Function: bound bound bound _onShake],
  [Function: bound bound bound _onLongPress],
  [Function: bound bound bound _onOrientationChange],
  [Function: bound bound bound _onMouseWheel],
  [Function: bound bound bound _onMouseOut],
  [Function: bound bound bound _onMouseEnter],
  [Function: bound bound bound _onContextMenu] ]
```

fixing this already looks good to me

asturur:
tests are stalling. :(

PR #3996, opened 2017-06-11
Force async

asturur:
Remove the force async logic from fromObject.
All fromObject and clone work in async way.

### DIFF
--- a/src/elements_parser.js
+++ b/src/elements_parser.js
@@ -41,18 +41,7 @@ fabric.ElementsParser.prototype.createObject = function(el, index) {
 };
 
 fabric.ElementsParser.prototype._createObject = function(klass, el, index) {
-  if (klass.async) {
-    klass.fromElement(el, this.createCallback(index, el), this.options);
-  }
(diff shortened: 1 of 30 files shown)


sfentress:
Hi @asturur , I've been updating an old project from 1.x to the current version, and the dropping of `object.async` was never mentioned in any change logs or notes that I could find. It took me a long time to track this change down.

It might have been worth keeping `async = true`, since all code now acts as it did when async existed and was true. All my code along the lines the following broke.

```
if (klass.async) {
      activeObject.clone(function (clonedObject) {
        clone = clonedObject;
      }
    } else {
      var clone = activeObject.clone();
    }
```

But I guess three years on this is a moot point.

asturur:
When i got to update fabric from 1 to 2 i had still lot to learn around backward compatibility and possible breakage patterns. Not that now i m a genius, but 3 years ago was far worse.

sfentress:
No problem! And my message above was not particularly helpful either.

PR #4003, opened 2017-06-13
Parser RegEx update

stefanhayden:
better regex and attribute validation as well as public access to the regex

- centralize regex creation
- adds `'linearGradient', 'radialGradient', 'stop'` to `svgValidTagNames` since fabric does parse them for gradients
- expose regex to public
- adds `getSvgAttributes`. While fabric instances (circle, rect, ect..) have their own list of acceptable params these parsed elements for gradients do not have any other place to list the params that fabric parses.

this will allow a user to takes the regexes and fully detect which features from a loaded svg were not parsed by fabric.


asturur:
@stefanhayden i forgot about this.
I m going to merge it , but we should improve the fabric.util.getSvgAttributes to get the list for any shape.

PR #4006, opened 2017-06-14
Update rgba regex to account for alpha

### LINKED ISSUE #4004: canvas.toSVG fails for rgba colors
## Version
1.7.6

## Test Case
http://jsfiddle.net/7qebwqfL/

## Steps to reproduce
Create a shape on the canvas with an rgba fill. Call canvas.toSVG.

## Expected Behavior
SVG fill colors will be the same as the they were in the canvas.

## Actual Behavior
SVG fill color is rgb(0, 0, 0)

## Related stack overflow question
https://stackoverflow.com/questions/44533511/how-do-i-make-fabric-js-output-svg-including-alpha-color

nickvans:
Previous regex would fail for an rgba string with a non integer alpha value. "rgba(255, 128, 64, .5)"

closes #4004

### DIFF
--- a/src/color.class.js
+++ b/src/color.class.js
@@ -280,7 +280,7 @@
    * @memberOf fabric.Color
    */
    // eslint-disable-next-line max-len
-  fabric.Color.reRGBa = /^rgba?\(\s*(\d{1,3}(?:\.\d+)?\%?)\s*,\s*(\d{1,3}(?:\.\d+)?\%?)\s*,\s*(\d{1,3}(?:\.\d+)?\%?)\s*(?:\s*,\s*(\d+(?:\.\d+)?)\s*)?\)$/;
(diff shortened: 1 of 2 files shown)


asturur:
in this file
https://github.com/nickvans/fabric.js/blob/3422c01a6811f756cdeb4ebc37fde6821e9c7995/test/unit/color.js#L176

add a test like the one i underlined that parse a color with opacity .x instead of 0.x and verify that the value is read correctly.

And we are done

asturur:
managed to do it

nickvans:
Thanks. I was trying to follow your instructions (I haven't contributed to a github project before) and when I followed the URL you gave me it told me I couldn't make changes unless I was in a branch. I made the change in my branch but couldn't figure out how to commit it so it would be added to the pull request. Thanks again for the help.

PR #4070, opened 2017-07-04
mouse event

InStrive:
Mouse events, draw lines but does not trigger a selection object, is how to define;How to selectively to learn I want to use the interface.
thanks in this!


asturur:
@InStrive  can you open an issue with better explanation? thanks.

InStrive:
@asturur sorry! i'm beginner,i found it,that's [canvas.isDrawingMode = !canvas.isDrawingMode],drawing and 
select split two part.
but a new question is last step and next step in canvas drawing or add object,What should  i do?
forgive my English is so bad.

asturur:
that is not the place to ask. This is a pull request.

You should learn some basic and explain what are you doing, otherwise is hard to help.

PR #4076, opened 2017-07-05
Remove activeGroup functionalities from Group, create ActiveSelection class

asturur:
Pro:
Simplifications:
getActiveObject, setActiveObject, getActiveObjects return an array of selected objects.
getActiveGroup, setActiveGroup are removed.
deactivateAllWithDispatch is removed, use discardActiveGroup.
deactivateAll is removed, use _discardActiveGroup.
Added activeSelection => group, single high level method.
removed all the dancing of `originalHasControls, hasControls` relative to group.
removed the weird logic of rendering controls in groups.
built in possibility to cancel a deselect, making onDeselect return true.
possibility to do group => activeSelection

missing:
restoring the lock system, an object locked in the active selection makes the activeselection locked

Cons:
Api changes
everyone confused on what is going on
possible regressions
riot
i eventually get killed


ozooner:
I hope you won't get killed any time soon :smiley: 

In case someone happens to read this, few corrections:

> deactivateAllWithDispatch is removed, use discardActiveGroup.

use **discardActiveObject**

> deactivateAll is removed, use _discardActiveGroup.

use **_discardActiveObject**

PR #4086, opened 2017-07-11
Improve performance of copyGLTo2D in Firefox & potentially other browers

plainview:
Why:

* Copying data to 2d canvas from GL context dominates filtering time when using webgl
* Chrome is happy to copy data from GL with ctx.drawImage in sub-millisecond time
* Firefox performs drawImage much more slowly but can do putImageData quite fast instead
* Using putImageData results in ~ 2-3x speedup on Firefox 54 in the fabricjs.com filtering demo on my Macbook Pro

This change addresses the need by:

* Adding a runtime check for which copy func performs better
* Swaps drawImage for putImageData if that func is detected as being faster

Also submitted for review:

* Log minifier error output to console if build.js minifyCmd fails (helps identify syntax errors while using build:watch)

### DIFF
--- a/src/filters/webgl_backend.class.js
+++ b/src/filters/webgl_backend.class.js
@@ -60,6 +60,55 @@
       this.createWebGLCanvas(width, height);
       // eslint-disable-next-line
       this.squareVertices = new Float32Array([0, 0, 0, 1, 1, 0, 1, 1]);
+      this.chooseFastestCopyGLTo2DMethod(width, height);
+    },
+
+    /**
+     * Pick a method to copy data from GL context to 2d canvas.  In some browsers using
+     * putImageData is faster than drawImage for that specific operation.
+     */
+    chooseFastestCopyGLTo2DMethod: function(width, height) {
+      var canMeasurePerf = typeof window.performance !== 'undefined';
+      var canUseImageData;
+      try {
+        new ImageData(1, 1);
+        canUseImageData = true;
+      }
+      catch (e) {
+        canUseImageData = false;
+      }
+      // eslint-disable-next-line no-undef
+      var canUseArrayBuffer = typeof ArrayBuffer !== 'undefined';
+      // eslint-disable-next-line no-undef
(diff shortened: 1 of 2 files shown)


src/filters/webgl_backend.class.js
```diff
@@ -304,3 +336,44 @@
...
+  var width = targetCanvas.width;
+  var height = targetCanvas.height;
+  var numBytes = width * height * 4;
+  // eslint-disable-next-line no-undef
+  var u8 = new Uint8Array(this.imageBuffer, 0, numBytes);
+  // eslint-disable-next-line no-undef
+  var u8Clamped = new Uint8ClampedArray(this.imageBuffer, 0, numBytes);
+  gl.readPixels(0, 0, width, height, gl.RGBA, gl.UNSIGNED_BYTE, u8);
+  var imgData = new ImageData(u8Clamped, width);
+  ctx.putImageData(imgData, 0, 0, 0, 0, width, height);
```
asturur: the ```ctx.putImageData(imgData, 0, 0)``` version is not enough?

asturur:
Just a question to recap:

`this.imageBuffer` is created to stay and gives us the abilty to share between glReadPixels and putImageData

We create 2 differents array view because glReadPixels does not like clampedData and putImageData does not like normal uint8 values?

Or because providing the right data type speed up something?

plainview:
> We create 2 differents array view because glReadPixels does not like clampedData and putImageData does not like normal uint8 values?

Yes, exactly.  `new ImageData()` expects a clamped array and throws a TypeError if a Uint8Array is passed instead.  `readPixels` writes to a TypedArray that must match the GLenum passed in; in this case we pass `gl.UNSIGNED_BYTE`, which only matches with the Uint8Array type.

src/filters/webgl_backend.class.js
plainview: This is a good question, i will try tomorrow and report back.

asturur: thanks!

plainview:
Yup, `putImageData(data, dx, dy)` works absolutely fine!

PR #4093, opened 2017-07-12
Use generic setter for sourcePath

jparkrr:
Not completely sure if this is the right fix, but I'm pretty sure this is broken.

Thanks for all your amazing work @asturur

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -920,7 +920,7 @@
         delete object.path;
 
         path.setOptions(object);
-        path.setSourcePath(pathUrl);
+        path.set('sourcePath', pathUrl);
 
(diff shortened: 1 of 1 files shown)


asturur:
There is not setSourcePath function in the class? i recently moved away the named setter/getters, i may have break something

asturur:
would you mind changing in `this.set('sourcePath', pathUrl)`  since sourcePath is something particular, i would prefer it stays wrapped in a getter in case we need to do something with it later. ( and since it was in the getter )

jparkrr:
@asturur Thanks, updated

asturur:
i ll merge this tomorrow after i verified why no test was failing for this error

PR #4096, opened 2017-07-13
Add missing double quotes

0s1r1s:
The 'width' attribute didn't close inside toSvg function


asturur:
thanks! i have to check why tests did not fail and i  ll merge it

asturur:
added a broken test that this PR should Fix. passing should be 4458

PR #4097, opened 2017-07-13
Font cache: no lower case of fontWeight

jparkrr:
Shouldn't ever be a string, right?


asturur:
Well i prefer to be a string so that we do not end with cache for 

normal_bold
normal_BOLD
normal_Normal
and so on...

if there is an errorr where a 400 in font weight throws out because cannot go to string, let's fix that.

maybe  (fontWeight + '').toLowercase?  No idea on top of my head which is the right nice fix.

jparkrr:
right, it was the number that breaks it

Issue #3213, opened 2016-08-29
[Feature] - Subscript/Supscript


terrancesnyder:
All,

The company I work for deals heavily with graphic rendering and text layout of printed products and digital ones. The fabric library is fantastic, but we've run into a desire to scale and we'd like to open a bounty on fabricjs core to allow richer dynamic iText editing that support sub-script and sup-script.

We're open to many possible solutions but at the end of the day the solution needs to accomplish the following:

The ability to change a selection of character text to sub-script or sup-script. For example rendering 1.00 as 1<sup>00</sup>.
The solution needs to use the standard measureText(...) functions as we have overridden them to ensure kerning support via openType.js. Your solution need to be concerned with kerning support.
Here are some helpful links to get you started:

An example "poor guy/gals" superscript/subscript

https://forum.processing.org/two/discussion/6367/primitive-superscript-subscript-text-rendering
A simple video on what subscript and supscript means

https://www.youtube.com/watch?v=4boHA5wo5e0
Validation of work:

Modification of the http://fabricjs.com/test/misc/itext.html page with a button that shows "sub-script" and "sup-script" and the ability to select character ranges and subscript and sup-script them.

asturur:
would like to write a bullet point list of feature instead of a video, with some images?

i kind of understood what we need to do, but i would like to know how it should parametrized.

i cannot assure this can go in the library, i think text needs a full rewrite and i m hesitant to write features now.

denim2x:
I'd like to help out with this - say a plugin named 'super-text' or something like that

asturur:
I m refactoring text completely so that it can do more simple things that combined will give also this functionality.
I will not modify the itext demo, since is just a show case of styles.
I would like understand who sub script and super script should be parametrized.
Text will get in style a dx, dy, angle property that will be used to displace characters and also support TSPANS from svg.
There will be an extra method ( in default build or as an extension ) to just use a set of this parameter to create a subScript/superScript ( a particular dY a particular fontSize in relation to the previous character probably or from the main fontSize)
All those things should be probably discussed, not for the bounty or this particular feature but just to make something that is the best for the general dev.
Fabric lies in a dangerous place between an editor sdk and a canvas lib, while i favor the canvas lib, many people seek for the editor sdk. And i have to balance the things so that we do not get a mix of disjointed methods.

terrancesnyder:
+1 anything we can help with here @asturur ?

poffringa:
[...]

PR #4099, opened 2017-07-13
Feature/subscript supscript

zilker:
Basic sub/superscript support #3213 
Example: https://codepen.io/zilker/full/eRxXRB

### DIFF
--- a/src/mixins/itext.svg_export.js
+++ b/src/mixins/itext.svg_export.js
@@ -108,7 +108,8 @@
           charBox, style,
           boxWidth = 0,
           line = this._textLines[lineIndex],
-          timeToRender;
+          timeToRender,
+          realTextTopOffset;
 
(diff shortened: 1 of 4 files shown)


asturur:
Did not had time to read it fully yet.

My first impression is that we should try to bring in something that is deltaY for the text position ( in the style ) and let setSubscript and setSuperscript a method that manipulate the style of char X reading the char X - 1.

That is my opinion at least.

That would put the super/sub logic in 2 methods and do minimal changes to the render code.

zilker:
Hi asturur.
I based my solution on native baseline-shift SVG and future CSS property. I think it will be helpful in future for supporting this property in export/import to SVG or something like this. That's why it in style. My solution works like you write - I manipulate "fontSize" by "baselineShiftSettings.size" setting, dY by "baselineShiftSettings. superPosition" and "baselineShiftSettings. subPosition".

PR #4118, opened 2017-07-22
Add a test for group fromObject

asturur:
Since my friend @stefanhayden is too lazy to do so 😄


stefanhayden:
I was going to!!!! But thanks!

PR #4121, opened 2017-07-22
Fixed typo in createImageData

dragondgold:


asturur:
good just in time for beta4.
Since resize filter has been fast patched to release webgl, please if you are using it and notice something unoptimal from api point of view, please tell

dragondgold:
I just have a question regarding the API, it's the first time I'm using a resize filter. I see there's `applyFilters()` and `applyResizeFilters()`. Should I call both? Any particular order? I would expect `applyFilters()` to apply all kind of filters it doesn't matter if it's a resize filter.

PR #4128, opened 2017-07-23
Update comments

neopheus:
Update JS DOC


asturur:
thanks, i write fast, i break things.

PR #4129, opened 2017-07-23
Update Event.js - Variable name

neopheus:
Update Event.js - Variable name


asturur:
this is an external lib, we should not modify it.

neopheus:
no problem 
I send a pr : https://github.com/mudcube/Event.js/pull/20

asturur:
yes!

PR #4133, opened 2017-07-24
Update build.js

neopheus:
Add alignment and centering to build.js

### DIFF
--- a/build.js
+++ b/build.js
@@ -125,6 +125,8 @@ var filesToInclude = [
   'HEADER.js',
 
   ifSpecifiedInclude('gestures', 'lib/event.js'),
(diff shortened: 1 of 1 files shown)


asturur:
Those are not official features, and are not fully functional i think. I do not want to have them in the build or i have to support them also.

neopheus:
I understand. It's just to be able to create a build with these elements easily

asturur:
we may work on it, maybe on the website.
I think align guideline and centering are not in good state now, let's move this later in time

asturur:
i need to get 2.0 out in decent condition. I cannot add more things to it before release.

neopheus:
no problem ;)

PR #4137, opened 2017-07-25
Some modifications

neopheus:
JSDoc
unused var
unnecessary semicolon


src/parser.js
```diff
@@ -88,7 +88,7 @@
       }
     }
     else if (attr === 'visible') {
-      value = (value === 'none' || value === 'hidden') ? false : true;
+      value = (!(value === 'none' || value === 'hidden'));
```
asturur: this can be value !== 'none' && value !== 'hidden'

asturur:
thanks for the clean up. can you address that smal comment if possible?

neopheus:
it's do ;)

neopheus:
new modifications

asturur:
please write when you are done.

neopheus:
it's done for me now

src/util/lang_string.js
```diff
@@ -43,12 +43,12 @@
...
    * @memberOf fabric.util.string
-   * @param {String} string String to escape
+   * @param {String} textstring String to escape
    * @return {Array} array containing the graphemes
    */
   function graphemeSplit(textstring) {
-    var i = 0, graphemes = [];
-    for (var i = 0, chr; i < textstring.length; i++) {
+    var i = 0, chr, graphemes = [];
+    for (i = 0, chr; i < textstring.length; i++) {
```
asturur: if chr is declared outside, inside the for braket is not needed.
Also please, since you are on this thing, rename chr to _chr, because in the ES4 specs chr was a reserved word, and some compilers complain about using it.

asturur:
I do not thin the `for ( var i )` changes do any boost to the code.
But i have no preferences for one or the other.

Do you have any reference that may bring to  be inclined to one or the other form?

PR #4177, opened 2017-08-07
'deltaY', superscript, subscript etc.

denim2x:
- added `deltaY` to `_styleProperties`;
- updated `getSvgSpanStyles` for `deltaY` -> `baseline-shift`;
- updated `_renderChar` to honor `deltaY` by adjusting `top`;
- added functionality for 'superscript' and 'subscript';
- fixed some bugs 🐞

### DIFF
--- a/src/mixins/object.svg_export.js
+++ b/src/mixins/object.svg_export.js
@@ -60,7 +60,6 @@
 
     /**
      * Returns styles-string for svg-export
-     * @param {Boolean} skipShadow a boolean to skip shadow filter output
      * @return {String}
      */
     getSvgSpanStyles: function(style) {
@@ -71,7 +70,8 @@
           fontWeight = style.fontWeight ? 'font-weight: ' + style.fontWeight + '; ' : '',
           fill = style.fill ? getSvgColorString('fill', style.fill) : '',
           stroke = style.stroke ? getSvgColorString('stroke', style.stroke) : '',
-          textDecoration = this.getSvgTextDecoration(style);
+          textDecoration = this.getSvgTextDecoration(style),
+          deltaY = style.deltaY ? 'baseline-shift: ' + (-style.deltaY) + '; ' : '';
 
       return [
         stroke,
@@ -82,6 +82,7 @@
         fontWeight,
         textDecoration,
         fill,
+        deltaY,
       ].join('');
     },
 
--- a/src/mixins/object.svg_export.js
+++ b/src/mixins/object.svg_export.js
@@ -60,6 +60,7 @@
 
     /**
      * Returns styles-string for svg-export
+     * @param {Object} style the object from which to retrieve style properties
      * @return {String}
      */
     getSvgSpanStyles: function(style) {
@@ -86,6 +87,11 @@
       ].join('');
     },
 
+    /**
+     * Returns text-decoration property for svg-export
+     * @param {Object} style the object from which to retrieve style properties
+     * @return {String}
+     */
     getSvgTextDecoration: function(style) {
       if ('overline' in style || 'underline' in style || 'linethrough' in style) {
         return 'text-decoration: ' + (style.overline ? 'overline ' : '') +
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -45,6 +45,7 @@
     'charSpacing',
     'styles'
   );
+
   /**
    * Text class
    * @class fabric.Text
@@ -57,7 +58,7 @@
 
     /**
      * Properties which when set cause object to change dimensions
-     * @type Object
+     * @type Array
      * @private
      */
     _dimensionAffectingProps: [
@@ -169,6 +170,26 @@
     lineHeight:           1.16,
 
     /**
+     * Superscript schema object (based on https://tr.im/subscript_superscript)
+     * @type Object
+     * @default
+     */
+    superscript: {
+      size:      0.60, // fontSize factor
+      baseline:  0.67  // baseline-shift factor (upwards)
+    },
+
+    /**
+     * Subscript schema object (based on https://tr.im/subscript_superscript)
+     * @type Object
+     * @default
+     */
+    subscript: {
+      size:      0.62, // fontSize factor
+      baseline: -0.25  // baseline-shift factor (downwards)
+    },
+
+    /**
      * Background color of text lines
      * @type String
      * @default
@@ -235,8 +256,8 @@
     charSpacing:             0,
 
     /**
-     * Object containing character styles
-     * (where top-level properties corresponds to line number and 2nd-level properties -- to char number in a line)
+     * Object containing character styles - top-level properties -> line numbers,
+     * 2nd-level properties - charater numbers
      * @type Object
      * @default
      */
@@ -253,7 +274,7 @@
     _measuringContext: null,
 
     /**
-     * Array of properties that define a style unit.
+     * Array of properties that define a style unit (of 'styles').
      * @type {Array}
      * @default
      */
@@ -443,16 +464,9 @@
      */
     _extendStyles: function(index, styles) {
       var loc = this.get2DCursorLocation(index);
-
-      if (!this._getLineStyle(loc.lineIndex)) {
-        this._setLineStyle(loc.lineIndex, {});
-      }
-
-      if (!this._getStyleDeclaration(loc.lineIndex, loc.charIndex)) {
-        this._setStyleDeclaration(loc.lineIndex, loc.charIndex, {});
-      }
-
-      fabric.util.object.extend(this._getStyleDeclaration(loc.lineIndex, loc.charIndex), styles);
+      var decl = this._getStyleDeclaration(loc.lineIndex, loc.charIndex) || {};
+      fabric.util.object.extend(decl, styles);
+      this._setStyleDeclaration(loc.lineIndex, loc.charIndex, decl);
     },
 
     /**
@@ -738,7 +752,9 @@
      * @private
      */
     _setStyleDeclaration: function(lineIndex, charIndex, style) {
-      this.styles[lineIndex][charIndex] = style;
+      var decl = this._getLineStyle(lineIndex) || {};
+      decl[charIndex] = style;
+      this._setLineStyle(lineIndex, decl);
     },
 
     /**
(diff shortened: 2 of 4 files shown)


src/mixins/itext.svg_export.js
```diff
@@ -131,7 +131,7 @@
           timeToRender = this._hasStyleChanged(actualStyle, nextStyle);
         }
         if (timeToRender) {
-          style = this._getStyleDeclaration(lineIndex, i) || { };
+          style = this.getStyleDeclaration(lineIndex, i) || { };
```
asturur: we cannot rename this as part of the PR. _getStyleDeclaration will stay named like this for now.

asturur:
a couple of suggestions, leave refactors of properties array and renaming out. Is easier to grasp meaningfull changes.
If you have some ideas about them open a PR later.

src/shapes/text.class.js
```diff
@@ -169,6 +160,13 @@
     lineHeight:           1.16,
 
     /**
+     * deltaY
+     * @type Number
+     * @default
+     */
+    deltaY:               0,
```
asturur: i m uncertain about this. We should not offer a global deltaY property to the developer.

There is a probably a way to manage style without having the parent property set.

src/shapes/text.class.js
```diff
@@ -908,7 +913,7 @@
       var box = {
         width: width,
         left: 0,
-        height: charStyle.fontSize,
+        height: style.fontSize,
         kernedWidth: kernedWidth,
       };
```
asturur: would be nice to have the deltay in the graphmebox information, even if for now we are not using it in the grapheme box context.

src/shapes/text.class.js
```diff
@@ -919,32 +924,37 @@
...
+     * @return {Number}
      */
     getHeightOfLine: function(lineIndex) {
       if (this.__lineHeights[lineIndex]) {
         return this.__lineHeights[lineIndex];
       }
 
       var line = this._textLines[lineIndex],
-          maxHeight = this.getHeightOfChar(lineIndex, 0);
+          maxHeight = Number.NEGATIVE_INFINITY,
```
asturur: In this way you are loosing the maxHeight of the first line.
Other than that, text can be in an empty state situation, where there is no text. I would not start from negative_infinity with the risk of getting it as result.

denim2x: the idea was that we need to take `dy` into account - and the only way was to start with `-Infinity`

asturur:
I'm unsure about the lineHieght changes.
Moving up a char, should not make the line taller.
We should just need to worried about the first and last line to do not go out of the bounding boxes, but setting the proper values of dy to do not make the text look bad is a developer task, not the library to allow everything to look good. is just too hard.

As a first step would be nice just to see  the letters moved up and and down in the render, even if they overlap with others, and the same effect respected in toSVG.

That would be enough small to be readable/reviewable and would be enough working to be mergeable.

asturur:
image

this is a screen from google docs. They limit the displacement of apix and pedix to little so that they do not cross line and lineHeight stays the same.

Since we are implementing this property as a tool to get apix and pedix, is ok leave lineHeight calculation out.

Since later i want to introduce per char rotation as svg, i think the measurement logic must be revised useing the boxes of each char. Do not worry now.

asturur:
image

they actually nearly overlap

denim2x:
then may I calculate `deltaY` to fit `lineHeight`?

asturur:
you can go by trying, do not need to get it perfect.
Over the UPPERCASE letters there is like 0.15 space of font size or 0.12. It looks like they are using a 0.5 or 0.4 multiplier.

src/shapes/text.class.js
```diff
@@ -1109,6 +1106,17 @@
...
+      if (decl && decl.deltaY) {
+        var height = this.getHeightOfChar(lineIndex, charIndex);
+        var deltaY = decl.deltaY;
+        if (deltaY <= 0) {
+          deltaY = height - Math.min(height - deltaY, lineHeight * (1 - LINE_PADDING));
+        } else {
+          deltaY = Math.max(deltaY, lineHeight * LINE_PADDING);
+        }
+        top += deltaY;
+      }
```
asturur: what is this doing? why we cannot have just top += deltay? i do not understand the math purpose.

denim2x: well the math ensures there's no overlap; there could be some flag (e.g. `allowOverlap`) for disabling this behavior (defaults to `false`)

asturur:
no let the overlap free.
we need to be free with rendering and svg overlaps too if it needs it.

The method to set sub or super script can take in consideration overlap, but rendering just excutes the values that find in style.

denim2x:
🔗   [plunker](http://plnkr.co/edit/IfyiId5A83qxEP3LWjQq?p=preview)

src/shapes/text.class.js
```diff
@@ -1092,7 +1106,7 @@
      * @param {Number} top Top coordinate
      * @param {Number} lineHeight Height of the line
      */
-    _renderChar: function(method, ctx, lineIndex, charIndex, _char, left, top) {
+    _renderChar: function(method, ctx, lineIndex, charIndex, _char, left, top, lineHeight) {
```
asturur: do we still need that lineHeight? looks like we do not need now.

src/shapes/text.class.js
```diff
@@ -1203,16 +1284,34 @@
...
+     * @param {Number} char the character number
+     * @param {String} key the property name
+     * @param {Any} value the value
+     * @returns 'value'
+     */
+    setPropertyAt: function(line, char, key, value) {
+      var decl = this._getStyleDeclaration(line, char) || {};
+      decl[key] = value;
+      this._setStyleDeclaration(line, char, decl);
+      return value;
```
asturur: I do not think we should return a value with the setter. We can return this and make it chainable

src/shapes/text.class.js
asturur: i think we can just return deltay: deltay in the box object, is more than fine for now.

asturur:
I think is everything ok. I left a pair of comments and i have some comment to write about the scale function that is too much general for what we have to do. Let me think a bit about it.

src/mixins/object.svg_export.js
```diff
@@ -60,7 +60,6 @@
 
     /**
      * Returns styles-string for svg-export
-     * @param {Boolean} skipShadow a boolean to skip shadow filter output
```
asturur: would you mind correcting the JSDOC here?
* @param {Object} style object to getthe style-string from

denim2x: done

asturur:
I would prefere something like this for bot sup and sub script

```javascript
   setSuperscript: function(line, char) {
         var currentFontSize = this.getPropertyAt(line, char, 'fontSize');
         setPropertyAt(line, char, 'fontSize', currentFontSize * this.superscript.fontSize);
         setPropertyAt(line, char, 'deltaY',  -currentFontSize * this.superscript.baseline);
         return this;
   },
```

Is shorter than the general one and we do not have any general case to apply this for.
Any reason why you think the scale function is necessary?


I m really trying to just save space and code to mantain. Fabric is a big library full of features that maybe should not be introduced at all at the time.

I'm sorry if it looks like i'm spoon feeding you.

denim2x:
fine by me

asturur:
image

take care of this ^^^^

if you cloned the repository, you should have a command line
`npm run lint` that would allow you to check this beforehand.

Also try to add a small  test for setSubscript and setSuperScript so that it verifies that after calling the method the style is mutated.

i can provide a snippet for you.

I would merge this then.

After this is merged we would need some more work to displace the cursor that i believe does not follow the superscript. Probably it automatically gets the size but not the baseline.

Do not worry for now do not mix it here.

asturur:
text example
```javascript
  test('text superScript', function() {
    var text = new fabric.Text('xxx', { fontSize: 30 });
    text.styles = { 0: { 0: { fontSize: 20, fill: 'blue' }, 1:  { fill: 'blue' }}};
    ok(typeof text.setSuperScript === 'function');
    text.setSuperScript(0, 0);
    text.setSuperScript(0, 1);
    equal(text.styles[0][0].baseline, undefined, 'the baseline is not set');
    equal(text.styles[0][0].fontSize,....., 'the fontSize has been.... add description');
    equal(text.styles[0][0].baseline, ...., 'baseline is now a portion of ...');
// then verify for the char 1 that has no fontSize in the style
    equal(text.styles[0][1].baseline, undefined, 'the baseline is not set');
    equal(text.styles[0][1].fontSize,....., 'the fontSize has been.... add description');
    equal(text.styles[0][1].baseline, ...., 'baseline is now a portion of ...');
  });
```

asturur:
I was also thinking that while the fontsize is multiplied the baseline is added. This is ok for 1 usage of superscript.

The functionality is generic enough so that i can set superscript twice, but i need to add the baseline to the current one.
Since deltaY is not a property of the text prototype and probably should not be, i think we need a special handling when we get is as undefined to interpret as 0.

denim2x:
well currently 'deltaY' is calculated from the current 'fontSize' (not 'deltaY') - I never retrieve 'deltaY' in my code

denim2x:
when do we need 'deltaY == 0'? I always check for the *presence* of 'deltaY' in my code - when rendering and converting to SVG

asturur:
I made the ipothesis of calling setSuperScript twice.

The font mulitplication works. Because it gets multiplies by some 0.x twice. So the char gets smaller and smaller.

The second time you call it, baseline would be erased and recalculated by the current fontSize that is the one of the first call of setSuperScript, while thinking of making the apix of the apix, you would need to displace the character UP by what it was before plus a portion of the currentFontsize.

This bring in the need of knowing the current baseline. the current baseline may be undefined.

denim2x:
I'll check it

asturur:
https://github.com/kangax/fabric.js/pull/4177#issuecomment-321477262

in this comment i was adding it in that small function example

denim2x:
well what do you know - it works! I just had to do `dy + size * ...`

asturur:
i think we are over engineering the solution.
all the schema thing is way overcomplicated for me, the plain function is the best in my opinion.

we need tests to pass. are you able to run tests locally on your end?

denim2x:
first off - when you said `scaleChar` should go - well, it's gone;
secondly, well - there's absolutely *no point* in writing the *same* code *twice* (*DRY* anyone?);
and last (but *not* least), have you checked the plunker I posted?

asturur:
what you say is true in general, in this case is more than necessary. in
this case the general code is bigger than twice the simple one and it
involves modifying our class creation code.

other than that being an open project mantained by volunteers, getting a
grasp of what a function does at first glance is very important.

i ll show you later what i mean.

asturur:
no i did not open the plunker. i saw a link with no introduction no
explanations and does not work from mobile.

asturur:
so the feature is working, tests do not pass but i think i can fix them.
I m going to merge this branch, fix the tests in a PR.

We should try to fix the cursor look that is not in the right position during deltaY displacement.

denim2x:
ok

PR #4184, opened 2017-08-09
Allow multiple keys to enable multiple click selection

brianchitester:
Changed selectionKey to an array in order to allow multiple keys to enable multi click selection.  I thought this would be useful since, in my application that uses fabric.js, I would like to make it so that holding either shift or ctrl will enable multi selection.  This is consistent with the behavior in other editors like Google Slides.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -518,7 +518,7 @@
         selectionKeyPressed = this.selectionKeys.some(function (selectionKey) {
           return e[selectionKey];
         });
-      } else if (e[selectionKeys.toString()]){
+      } else if (e[this.selectionKeys.toString()]){
         selectionKeyPressed = true;
       }
 
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -518,7 +518,8 @@
         selectionKeyPressed = this.selectionKeys.some(function (selectionKey) {
           return e[selectionKey];
         });
-      } else if (e[this.selectionKeys.toString()]){
+      } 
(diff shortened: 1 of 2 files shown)


src/canvas.class.js
```diff
@@ -512,6 +512,14 @@
     _shouldClearSelection: function (e, target) {
       var activeObjects = this.getActiveObjects(),
           activeObject = this._activeObject;
+
+      var selectionKeyPressed = false;
+      this.selectionKeys.forEach(function (selectionKey) {
+        if (e[selectionKey]){
+          selectionKeyPressed = true;
+        }
+      });
```
asturur: anyway this could be done with  this.selectionKeys.some(function(key) { return e[selectionKey]; }) ?  is still ES5 and i think has the same performance of forEach.

asturur:
The problem of this change is that it breaks the api, again.
Whoever is changing the value of that key is getting a failure after this update.

brianchitester:
I'll push an update soon that allows for backwards compatibility.

asturur:
yeah let me also think about it. i m not sure i want to push it in. I need to recap all the key combination and figure out if setting this for more keys can cause problems.

I m not a fan of too many features

brianchitester:
Sure, thanks for considering ✌️

asturur:
I m going to merge this, i need to clean it a bit, but talking with some people we decided that this feature is indeed requested by users because of the differences on how selection works in different software/OS.

so... thanks!

PR #4190, opened 2017-08-12
Simplify setSubscript, setSuperscript

asturur:
There are some reasons behind having the code in this simpler form:

1) providing that we are not duplicating code or violating any normal coding good practice, this code is smaller.

2) The functionality behind `*fontSize` was a singular case that would have need its own doc, explanation

3) all the fabric functions for style set char by char, introducing a `many lines, many chars` would have make the other methods impair. Also we have setSelectionStyles to do this and can be used in the same way.

4) i believe that for who comes tomorrow this function _setScript is very easy to read.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -176,7 +176,7 @@
      */
     superscript: {
       size:      0.60, // fontSize factor
-      baseline:  0.67  // baseline-shift factor (upwards)
+      baseline:  -0.67  // baseline-shift factor (upwards)
     },
 
     /**
@@ -186,7 +186,7 @@
      */
     subscript: {
       size:      0.62, // fontSize factor
-      baseline: -0.25  // baseline-shift factor (downwards)
+      baseline: 0.25  // baseline-shift factor (downwards)
     },
 
     /**
@@ -1161,7 +1161,7 @@
      */
     _setScript: function(line, char, schema) {
       var fontSize = this.getValueOfPropertyAt(line, char, 'fontSize'),
-          baseline = this.getValueOfPropertyAt(line, char, 'deltaY') + fontSize * -schema.baseline;
+          baseline = this.getValueOfPropertyAt(line, char, 'deltaY') + fontSize * schema.baseline;
       this.setPropertyAt(line, char, 'deltaY', baseline);
       this.setPropertyAt(line, char, 'fontSize', fontSize * schema.size);
       return this;
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1151,11 +1151,20 @@
       return this._setScript(line, char, this.subscript);
     },
 
+    /**
+     * Turns the character into an 'inferior figure' (aka. 'subscript')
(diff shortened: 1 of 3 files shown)


asturur:
@denim2x this is how i prefer the code to be.

If you want to discuss about why/ho or if you see something wrong in this approach, or you have strong opinions on the good in your version, i m very open to talk about it.

src/shapes/text.class.js
```diff
@@ -1153,7 +1148,14 @@
      * @returns {Object} this
      */
     setSubscript: function(line, char) {
-      return this.subscript.apply(this, line, char);
+      return this._setScript(line, char, this.subscript);
+    },
+
+    _setScript: function(line, char, schema) {
+      var fontSize = this.getValueOfPropertyAt(line, char, 'fontSize'),
+          baseline = this.getValueOfPropertyAt(line, char, 'deltaY') + fontSize * schema.baseline;
```
denim2x: that's supposed to be `... + fontSize * -schema.baseline` (follows the `baseline-shift` attribute from SVG)

src/shapes/text.class.js
```diff
@@ -1153,7 +1148,14 @@
      * @returns {Object} this
      */
     setSubscript: function(line, char) {
-      return this.subscript.apply(this, line, char);
+      return this._setScript(line, char, this.subscript);
+    },
+
+    _setScript: function(line, char, schema) {
```
denim2x: there are some missing checks here

src/shapes/text.class.js
```diff
@@ -1153,7 +1148,14 @@
...
+      var fontSize = this.getValueOfPropertyAt(line, char, 'fontSize'),
+          baseline = this.getValueOfPropertyAt(line, char, 'deltaY') + fontSize * schema.baseline;
+      this.setPropertyAt(line, char, 'deltaY', baseline);
+      this.setPropertyAt(line, char, 'fontSize', fontSize * schema.size);
```
denim2x: `return this;`

src/shapes/text.class.js
denim2x: that's supposed to be ... + fontSize * -schema.baseline (follows the baseline-shift attribute from SVG)

src/shapes/text.class.js
denim2x: some checks are missing here

denim2x:
... the rest looks ok

denim2x:
after swapping of signs we should either rename `deltaY` to `baseline` - or change the schemas' `baseline` to `deltaY` - correct?

asturur:
Names are fine. We do not mirror any name other that for comfort.
Subscript/superScript can have any name in the style as long as it moves the thing in the right direction.
Having to swap the sign of schema multiplier is an added notion that a dev should remember, that is why i moved the signs.

Fabric will not have support for baseline has people knows it, 
Fabric supports just baseline `alphabetic` and i do not want to give impression that they can use other values giving a a property with the same name in the style.

denim2x:
fine

PR #4202, opened 2017-08-14
Normalize Api for getSelectionStyles, setSelectionStyles

asturur:
- Moved style related method to a different file
- getSelectionStyles and setSelectionStyles are available for Text now.
- before it could be called with 0, 1 or 2 arguments. Now or 0 or 2.
- being a `selectionStyle` no selection, no style. Calling  getSelectionStyles(2,2) will return an empty array.
- before the function could return a single object or an array of objects. now just array.


src/mixins/text_style.mixin.js
```diff
@@ -0,0 +1,312 @@
...
+          };
+        }
+        selectionStart -= lines[i].length + 1;
+      }
+      return {
+        lineIndex: i - 1,
+        charIndex: lines[i - 1].length < selectionStart ? lines[i - 1].length : selectionStart
+      };
+    },
+
```
asturur: From here down start the modified methods

src/mixins/text_style.mixin.js
```diff
@@ -0,0 +1,312 @@
...
+        endIndex = this.selectionEnd || startIndex;
+      }
+      for (var i = startIndex; i < endIndex; i++) {
+        this._extendStyles(i, styles);
+      }
+      /* not included in _extendStyles to avoid clearing cache more than once */
+      this._forceClearCache = true;
+      return this;
+    },
+
```
asturur: end of modified methods

PR #4205, opened 2017-08-15
'deltaY': preliminary feature-complete implementation

denim2x:
upcoming incremental commits

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -176,7 +176,7 @@
      */
     superscript: {
       size:      0.60, // fontSize factor
-      baseline:  0.67  // baseline-shift factor (upwards)
+      baseline:  -0.67  // baseline-shift factor (upwards)
     },
 
     /**
@@ -186,7 +186,7 @@
      */
     subscript: {
       size:      0.62, // fontSize factor
-      baseline: -0.25  // baseline-shift factor (downwards)
+      baseline: 0.25  // baseline-shift factor (downwards)
     },
 
     /**
@@ -855,11 +855,6 @@
      * @return {Number} fontSize of the character
      */
     getHeightOfChar: function(line, char) {
-      var size = this.getValueOfPropertyAt(line, char, '*fontSize');
-      if (typeof size === 'number') {
-        return size;
-      }
-
       return this.getValueOfPropertyAt(line, char, 'fontSize');
     },
 
@@ -1143,7 +1138,7 @@
      * @returns {Object} this
      */
     setSuperscript: function(line, char) {
-      return this.superscript.apply(this, line, char);
+      return this._setScript(line, char, this.superscript);
     },
 
     /**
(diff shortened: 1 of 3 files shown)


asturur:
Hello cristian,
would be better if you comment my open pr about the deltay changes. so we can merge that and you can continue with the changes on top of that.

Do not mix with master yet, or later will be harder to merge the branches for me.

src/shapes/text.class.js
```diff
@@ -1152,22 +1156,42 @@
...
+     * @returns {Object} thisArg
+     * @chainable
      */
     _setScript: function(line, char, schema) {
+      if (!this._checkLocation(line, char) || typeof schema !== 'object') {
+        return this;
+      }
+      if (typeof schema.size !== 'number' || typeof schema.baseline !== 'number') {
+        return this;
+      }
```
asturur: We do not take care of this kind of checks.
If the dev provide a line/char that does not exist, he will set the style for that numbers.
There may be reasons to do that.

src/shapes/text.class.js
```diff
@@ -927,9 +931,9 @@
       var box = {
         width: width,
         left: 0,
-        height: style.fontSize,
+        height: style.fontSize + Math.abs(style.deltaY),
```
asturur: the height of the box is the fontSize. We should not move from this assumptions.

asturur:
@denim2x i do not understand the purposes of this changes.
On what part of the feature are you working on?

src/shapes/text.class.js
```diff
@@ -175,8 +175,8 @@
      * @default
      */
     superscript: {
-      size:      0.60, // fontSize factor
-      baseline:  -0.67  // baseline-shift factor (upwards)
+      size:      0.83, // fontSize factor
+      baseline: -0.36  // baseline-shift factor (upwards)
```
asturur: Any reason why the old numbers do not look good anymore?

src/shapes/text.class.js
```diff
@@ -737,12 +737,16 @@
...
     getCompleteStyleDeclaration: function(lineIndex, charIndex) {
       var style = this._getStyleDeclaration(lineIndex, charIndex) || { },
-          styleObject = { }, prop;
-      for (var i = 0; i < this._styleProperties.length; i++) {
-        prop = this._styleProperties[i];
-        styleObject[prop] = typeof style[prop] === 'undefined' ? this[prop] : style[prop];
+          decl = { }, slate = this._styleProperties;
+      for (var i = slate.length - 1; i >= 0; --i) {
+        var prop = slate[i];
+        if (style[prop] != null) {
```
asturur: Why the reversed for loop and why changing an undefined check with a null?

src/shapes/text.class.js
asturur: the source you documented:
image

has references to behaviours of other apps.

src/shapes/text.class.js
```diff
@@ -950,8 +954,8 @@
 
       var line = this._textLines[lineIndex], maxHeight = 0;
       for (var i = 0, len = line.length; i < len; i++) {
-        maxHeight = Math.max(this.getHeightOfChar(lineIndex, i), maxHeight);
-        // TODO: Take 'deltaY' into account to avoid overlaps (may require 'allowOverlaps' property)
+        var box = this._getGraphemeBox(null, lineIndex, i);
+        maxHeight = Math.max(maxHeight, box.height);
```
asturur: i would not change getHeightOfLine, i know devs will be able to set values that overlap, but i do not want to change basic behaviours for a new feature that has a marginal impact on overall library usage.

src/shapes/text.class.js
denim2x: well, about that... when order doesn't matter, a reversed loop is usually *faster* in JS;
as for `null`, well - it's shorter for one - but it's also *equivalent* (i.e. `null == undefined`)

src/shapes/text.class.js
denim2x: well then I'll have to come up with a different method - say `getExtentOfLine`

asturur:
@denim2x 
I m stating clearly as much as possible.
I do not want to change any measurement procedure in fabricjs to introduce this feature.
I simply cannot support it.

From a wrong consideration now may come an handfull of bugs that then i have to fix or revert changes.

I do not want to change any code that is unrelated to the functionality for some supposed performance improvement that is not backed by solid benchmarks.

The functionality is as good as it is if we fix the cursor rendering.

Unless the owner of the bounty comes with some different requirements and in that case i have to weight in my decisions.

PR #4216, opened 2017-08-17
enhanced `_renderTextDecoration`, `renderCursor`; tests

denim2x:
new properties: `downsizing` and `baselineShift`; `_renderTextDecoration` and `renderCursor` take `deltaY` into account; updated tests;
the result:

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -170,26 +170,6 @@
     lineHeight:           1.16,
 
     /**
-     * Superscript schema object (based on https://tr.im/subscript_superscript)
-     * @type Object
-     * @default
-     */
-    superscript: {
-      size:      0.60, // fontSize factor
-      baseline:  -0.67  // baseline-shift factor (upwards)
-    },
-
-    /**
-     * Subscript schema object (based on https://tr.im/subscript_superscript)
-     * @type Object
-     * @default
-     */
-    subscript: {
-      size:      0.62, // fontSize factor
-      baseline: 0.25  // baseline-shift factor (downwards)
-    },
-
-    /**
      * Background color of text lines
      * @type String
      * @default
@@ -296,6 +276,22 @@
      * contains characters bounding boxes
      */
     __charBounds: [],
+    
+    /**
+     * The scaling factors to be used upon applying super/subscript styles; when
+     * single-valued then same value applies to both.
+     * @type {Number} [<superscript>, [<subscript>]]
+     * @default
+     */
+    downsizing: [0.6],
+    
+    /**
+     * Amounts relative to current fontSize by which the baseline will be shifted 
+     * upon applying super/subscript styles; the default value has minimum overlap.
+     * @type {Array} [<superscript>, <subscript>]
+     * @default
+     */
+    baselineShift: [-0.35, 0.11],
 
     /**
      * Constructor
@@ -304,8 +300,26 @@
      * @return {fabric.Text} thisArg
      */
     initialize: function(text, options) {
-      this.styles = options ? (options.styles || { }) : { };
       this.text = text;
+      if (options) {
+        this.styles = options.styles || { };
+        var keys = ['size', 'baseline'];
+        ['downsizing', 'baselineShift'].forEach(function(key, i) {
+          var array = this[key].slice();
+          ['super', 'sub'].forEach(function(key, j) {
+            var schema = options[key + 'script'];
+            if (schema) {
+              var value = schema[keys[i]];
+              if (value) {
+                array[j] = value;
+              }
+            }
+          });
+          this[key] = array;
+        }, this);
+      } else {
+        this.styles = { };
+      }
       this.__skipDimension = true;
       this.callSuper('initialize', options);
       this.__skipDimension = false;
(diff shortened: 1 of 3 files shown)


asturur:
I do not want this schema logic nor modify how fabric calculate the object sizes.
If there are various ways of setting subscript/superscript to  imitate some app behaviour, we can suggest that in the comments, leaving one by default and explaining the others:

```diff
    /**
     * Superscript schema object (based on https://tr.im/subscript_superscript)
     * Word seems to use size: x and baseline: y. Css seems to use size: h and baseline: w ...
     * @type Object
     * @default
     */
    superscript: {
      size:      0.60, // fontSize factor
      baseline:  -0.67  // baseline-shift factor (upwards)
    },
```

I m not going to pull the feature in if it gets more complicated of what we did so far.

You can still publish a complete override for the functionality or an extension that you think is better and have it linked here in all the docs, but the core library cannot take all this stuff.

denim2x:
Does this mean we could have custom sizing methods in a separate extension as well?

asturur:
Sure you can build a file in your own repository or in this, taking example from /lib/aligning_guidelines.js  or from any class that extend a prototype ( textbox, iText ) that provide a new function.

I still think that you are over engineering a solution
Who ever uses fabric will have some customization needs that the library cannot provide for everyone.
Any dev will have to customize the behaviour a bit and in that case is better to have a simple library that provide the basic functionalities rather than having a bit of everything.

For a developer that prefer the superscript of a particular app, setting two values in an object is completely fine.

denim2x:
ok - thank you for clarifying

src/shapes/text.class.js
```diff
@@ -913,6 +924,9 @@
      * @param {String} [prevGrapheme] character preceding the one to be measured
      */
     _getGraphemeBox: function(grapheme, lineIndex, charIndex, prevGrapheme, skipLeft) {
+      if (grapheme == null) {
+        grapheme = this._textLines[lineIndex][charIndex];
+      }
```
asturur: where did you find this failing?

src/shapes/text.class.js
```diff
@@ -1132,38 +1145,65 @@
...
+
+      var size = value.size;
+      if (typeof size !== 'number') {
+        return this;
+      }
+
+      var baseline = value.baseline;
+      if (typeof baseline !== 'number') {
+        return this;
+      }
```
asturur: those things should not fail at all. We provide the class with a base schema, the developer set another.
If it set it wrong with wrong types or missing a piece, he will get error and correct.

Is not something that the user of the final product built with fabricjs has access to.

denim2x: well the user could do `text.schema.sub = 'foo'` - we need to handle it correctly

src/shapes/text.class.js
denim2x: it's meant as a convenience; should we take it out?

asturur: fabric is approaching 30k lines of code. i think it reached 300k of minified size. If not something that the user can break or that some browser needs in particular, i do not put it inside.

asturur:
There is a general thing about having the superScript and subScript options separated.
```javascript
new Text('some text...', {
  schema: {
    sub: {
      a: 1,
      b: 1
    },
   sup: {
      a: 1,
      b: 1
   },
  }
});
```

is worse than

```javascript
new Text('some text...', {
    superscript: {
      a: 1,
      b: 1
    },
   subscript: {
      a: 1,
      b: 1
   },
});
```

having one level of nesting less is generally easier to manage when writing code.
What is the reason behind changing the superscript/subscript code that we finished discussing over a week ago?
What are the points that make this solution an improvement over the older one?

denim2x:
it's designed to accommodate for future schema attributes (besides super/subscript) - e.g. paragraph margins, kerning, maybe ligatures etc.

src/shapes/text.class.js
asturur: the user is a developer. if it does it wrong, it will do it good after understanding whit it does not work. A nice jsdoc over the property will inform him what to use as schema or a sub/sup/superscript/subscript.

denim2x:
so now `options.superscript` and `options.subscript` get turned into `schema` if present; fixed `_getGraphemeBox` and `_setScript`

asturur:
@denim2x you continue to change the code around the schema with different solutions, all good and nice,but i just want the shortest, that was the one we already have.

I want you to take the bounty since you spent time over the feature.
I want the code to be small and i do not want to create additional other half features that are side effects of this.
Please make the cursor render good and also textDecoration as you did. ( verify the linetrough ).
Leave all the rest unchanged please.

SuperScript/subScript managment is good as it is, unless you find a bug.
If you find a bug open a PR that explain and fix just the bug.

additionaly this change here:
image

to which i m completely neutral about since change just the name of variable, for the reviewer is pure noise.
Distract from what he is concentrated about and makes the line count:
image

grow without any reason.

Contributing to an OS project is as hard as you are finding ( i guess ) because every project has its own rules and history and you have to stick at them or is hard to contribute.

If i haven't been clear on what i wanted in the last 2 weeks i have something to improve as moderator here, but i think i have been explicit, sometime pedant, on what i wanted from  the code style of the solution.

PR #4225, opened 2017-08-18
Better mouse support when lockScalingFlip is true.

### LINKED ISSUE #4209: [Feature request] Better mouse support when lockScalingFlip is true
Hi,

I have set the parameter 'lockScalingFlip' to false in my app, but there is something not really "user friendly" when you change this parameter to true.
The scaling is stopped if the mouse cursor go over the oposite edge of the object (i suppose it should lanch the flip, but in this case with the parameter lock=true, it stop the current action, so the stretching is stopped :( ).

Use case : 
http://jsfiddle.net/eA3xH/175/

If you select the bottom right corner of the rect and press you mouse button to stretch the rect, and move the cursor on the left, you can see the stretching is stopped if your mouse go over the left edge of the shape.

Not really sure if i am clear enough.
Let me know if this behavior could be better handled in the futur version.

Have a good day.

ncou:
close #4209

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -841,14 +841,16 @@
 
       if (lockScalingFlip && scaleX <= 0 && scaleX < target.scaleX) {
         forbidScalingX = true;
+        localMouse.x = 0;
       }
 
       if (lockScalingFlip && scaleY <= 0 && scaleY < target.scaleY) {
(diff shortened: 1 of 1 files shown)


asturur:
I think i will add some tests.

PR #4240, opened 2017-08-24
Enlarged hiddenTextarea font size to prevent iOS's input zoom

dambrogia:
The Problem
====
My canvas is zooming in on mobile (iOS) when I edit an `iText` object.

A good explanation of what is happening and why this is happening can be explained by the following:

> Whether you agree or disagree, Apple decided that form input on iOS needed to be a comfortable reading size, set at a minimum of 16px. If your form field font-size is less than that, they will automatically zoom into the field until the text appears at that size.

The `hiddenTextarea` within the `iText` object is assigned a `font-size` of `1px` which is causing this to zoom action happen. 

Obviously these picture aren't what are happening with the canvas itself, but a great representation of the action that is causing my headache.

----

##### Pics and quote are taken from [this link](https://uxcellence.com/2014/fix-ios-input-zoom).

lame_zoom

----

The solution
====

Increase the `font-size` attribute on the `hiddenTextarea` within the `iText` object from `1px` to  `16px`

Although the `textarea` element isn't visible, the zooming action still occurs. By setting the hidden textarea to a larger size the zooming action doesn't occur and we are all happy-coders.

----

no_zoom

----

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -13,7 +13,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     this.hiddenTextarea.setAttribute('wrap', 'off');
     var style = this._calcTextareaPosition();
     this.hiddenTextarea.style.cssText = 'position: absolute; top: ' + style.top +
-    '; left: ' + style.left + '; z-index: -999; opacity: 0; width: 1px; height: 1px; font-size: 1px;' +
+    '; left: ' + style.left + '; z-index: -999; opacity: 0; width: 1px; height: 1px; font-size: 16px;' +
     ' line-height: 1px; paddingｰtop: ' + style.fontSize + ';';
     fabric.document.body.appendChild(this.hiddenTextarea);
 


asturur:
Hello thanks for the nice pr with all explanations.

Some sick browsers will display the cursor for the hidden textarea because they are unable to hide it.

If this solves the zooming that is good, but instead of setting it at 16 fixed we should provide a configuration parameter, so that developers can choose between a big flashing cursor and a jumping page.

can you do that?

asturur:
you can try with the overflow hidden that i never tried, but setting the position offscreen will:

1) break composition
2) make the window on android/ios jump everywhere.

I'm ok with no additional configuration if overflow work.
Please try:
Android, Ie11,10, edge, chrome and firefox and safari if you have a chance.

asturur:
the reason why the hidden textarea is in place is this:

image

asturur:
Any news on experiments?

dambrogia:
No new news, sorry for the delay. Most likely I will have some time at the end of next week. But I'll keep you posted!

asturur:
Any news on this? i can take it over but i love to be helped.

dambrogia:
Hey @asturur ! 

Sorry for the delay, I actually did the testing last week but never got around to replying!

I tested both previously mentioned fixes I spoke about earlier:

- Setting `top` and `left` to `-100vh` and `-100vw` led to the input not being out of the view port which is a major dealbreaker.
- Setting the style to `height: 0; width: 0; overflow: hidden;` didn't change the view at all (I was pretty confident this would work too 🙁 )  I still saw the blinking blue cursor that is default within safari and chrome browsers on iOS.

The most successful approach I found was targeting the textarea within necessary browsers or viewports with CSS and applying styles as they were needed.

The bad news is - you must use `! important` in your CSS rules to overwrite any inline styles injected through javascript (like `font-weight`)

The good new is - adjusting the `font-weight` property of the text area to `16px` or greater omitted the auto zoom of the text area.

[This stack overflow question](https://stackoverflow.com/questions/30102792/css-media-query-target-only-ios-devices#answer-30103052) has some great references to target iOS devices withinin your CSS.

Specifically,
- [This link](http://stephen.io/mediaqueries/)
- [And this link](https://css-tricks.com/snippets/css/media-queries-for-standard-devices/) 

Lastly, a pure javascript implementation can be done by editing the style of the `textarea` within the `iText` object.

**This is sudo code** - but it should get anyone who is interested started in the right direction.

    var textLayer = new fabric.iText('my text here');

    textLayer.hiddenTextArea.style.fontSize = 16px;

    canvas.renderAll();




Thank you _very_ much for your attentiveness to this issue and for the awesome repo, it's much appreciated!

I will close my issue and cancel my PR, since multiple resolutions have been found and can be implemented on a circumstantial basis.

Cheers! 🍻

Issue #4235, opened 2017-08-22
Load image from JSON made in 1.7 into 2.0 not the same


blobinabottle:
## Version
2.0b6

## Test Case
http://jsfiddle.net/fmgXt/849/ Fabric 2.0
http://jsfiddle.net/fmgXt/848/ Fabric 1.7.15

## Steps to reproduce
load a JSON made in Fabric 1.7 in Fabric 2.0 

## Expected Behavior
Images should load the same way.
I guess it's because of the change in width/height behavior?

## Actual Behavior
The image in Fabric 2.0 is cropped and bigger. Can figure out a way to make it similar to Fabric 1.7.15

asturur:
good point.
The changes in width/height influence the way the image is displayed in a different way.

for me now is a mess going in the json, could you eliminate what is not image from your json fiddle?

Also do you think you can explain me why / how you ended up modifying the image width/height?

blobinabottle:
Cleaner version of json here:
fab 1.7.15 http://jsfiddle.net/fmgXt/848/
fab 2.0b6 http://jsfiddle.net/fmgXt/851/
And you're right, I think we made a mistake on our side, in some cases we change the widht/height by setting all the properties at once with set{height, width, etc} (it's when user upload big images we resized them this way... so it's our mistake).

asturur:
i think anyway flagging the new json of fabricjs as version 2 must be done.
So that everyone can see if the JSON loading back has a version, and if not can apply some migration scripts.
Breaking changes in json are are, the one for text are handled automatically fabric and are very simple.

In your case is just needed to reset image width/height reading from element and apply a scale factor. let me study the thing better but i think is fairly simple.

jilster:
I encountered the same thing. In my application it is very important that all objects have exactly the same strokeWidth and shadowOffset. Because those are influenced by the scale of the object, i initially made a simple function that kept the scaleX and scaleY at 1 and changed the dimensions instead doing something like:

```javascript
object.width = object.width * object.scaleX;
object.height = object.height * object.scaleY;
object.scaleX = 1;
object.scaleY = 1;
```
This kept the objects nice and uniform, but since version 2, the images were not displayed correctly. I fixed that using another function that reapplies scale to images, and uses it to scale the strokeWidth and shadowOffset accordingly. To do this i used the image's natural dimensions. Something like:
```javascript
object.width = imageNaturalWidth;
object.height = imageNaturalHeight;
[...]

PR #4251, opened 2017-08-27
Output version

asturur:
Currently undecided about this.
The code change is small, the data added to each fabricjs object dump is like 15 bytes.

related to #4235


asturur:
Should add to gradient, patterns and shadows to be more versatile

asturur:
Waiting for some people to comment, then i'll merge.

PR #4254, opened 2017-08-28
adds , {backstoreOnly: true} for toDataURL export

### LINKED ISSUE #4252: fabricjs call to canvas.toDataURL() resizes the dimensions of the canvas
I've posted this behavior on stackoverflow here: https://stackoverflow.com/questions/45908772/fabricjs-call-to-canvas-todataurl-resizes-the-dimensions-of-the-canvas

and per @asturur answer I'm creating an issue for the bug.

I'm using fabricjs 1.7.17 and have seen this issue below on IE Edge, Chrome and Firefox.

where the canvas is being set with a width and height in HTML of 768px.

Within my fabricjs canvas init I've got:
```
editor.canvas = new fabric.Canvas('editor')
editor.canvas.setDimensions({
  width: 1080,
  height: 1080,
  allowTouchScrolling: false,
  backgroundColor: 'rgba(55, 55, 55, .33)',
  rotationCursor: 'url(/static/img/rotate.cur) 10 10, crosshair'
}, {backstoreOnly: true})
```

The problem is when you call the editor.canvas.toDataURL() method the size of the goes from 768px to 1080px

If you call the browser's native canvas.toDataURL() method there isn't a jump in size.

Anyone bump into this and have a possible fix?

Here's a simplified jsfiddle: https://jsfiddle.net/m8dhmbtu/23/

Click on the "Native Canvas.toDataURL()" and you'll get the results to console.log and there will be no change to the displayed canvas in the preview window.

Click on the "Fabricjs canvas.toDataURL()" and you'll also get the results (the same correct png data) to console.log however; you'll also get a jump in the size of the in the preview window...this is what I'm trying to prevent.

asturur:
close #4252

Makes the canvas resize do not resize also the css dimensions of the canvas.

### DIFF
--- a/src/mixins/canvas_dataurl_exporter.mixin.js
+++ b/src/mixins/canvas_dataurl_exporter.mixin.js
@@ -73,7 +73,7 @@
       // setting interactive to false avoid exporting controls
       this.interactive = false;
       if (origWidth !== scaledWidth || origHeight !== scaledHeight) {
-        this.setDimensions({ width: scaledWidth, height: scaledHeight });
+        this.setDimensions({ width: scaledWidth, height: scaledHeight }, {backstoreOnly: true});
       }
       // call a renderAll to force sync update. This will cancel the scheduled requestRenderAll
       // from setDimensions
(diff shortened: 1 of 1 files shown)


asturur:
@joemanfoo i started the PR from your branch.
If you want to finish it would be nice to do something like:
```
var needsResize = origWidth !== scaledWidth || origHeight !== scaledHeight
```
Then use needsResize to conditionally resize both the time or just not do it.

Also add a this.renderAll() after the second if ( newly added ) since the requestRenderAll created from setDimensions does not make really sense for an export.

PR #4277, opened 2017-09-01
fix object.getObjectOpacity() docs

dersphere:
Seems like a copy&paste leftover.


asturur:
yes totally copy paste

asturur:
thankyou

PR #4283, opened 2017-09-03
Move the mouse over/ mouse out event fire in the mouseMove events only

asturur:
I moved the function out for 2 reasons:

1) is not right to get an event fired as a side effect of running findTarget
2) mouse enter, mouse leave belong to mouse move and not also to mousedown, mouseup, mousewheel.


asturur:
@ncou this could influence that bug about event firing.

PR #4295, opened 2017-09-07
Compare distance between cursor with absolute value.

asturur:
When cursor is after the end of line, we cannot use the calculation as if the cursor was between 2 chars.
The cursor is after both of them and so we need to compare the distance in absolute value.

@blobinabottle that should fix the cursor to next line.
(not tried yet)


asturur:
Ok the Math.abs trick did not work.
If i explicitly check for  the difference to be negative it works as expected.

PR #4302, opened 2017-09-13
Check fabric.window for ontouchstart

### LINKED ISSUE #2664: isTouchSupported detection.
When Chrome (Canary, 49.0.2578.0) is set to mobile emulation (Android or iPhone), `"ontouchstart" in fabric.document.documentElement;` will always return false.

Which disables touch event binding, `mouse:down` event does nothing.

jordan-acosta:
Fixes #2664

### DIFF
--- a/HEADER.js
+++ b/HEADER.js
@@ -26,7 +26,7 @@ else {
  * True when in environment that supports touch events
  * @type boolean
  */
-fabric.isTouchSupported = 'ontouchstart' in fabric.document.documentElement;
+fabric.isTouchSupported = 'ontouchstart' in fabric.window;
 
 /**
  * True when in environment that's probably Node.js


jordan-acosta:
This fix worked for me, tested in Chrome OSX & iOS Safari. Can test on more browsers, devices tonight.

Having trouble running tests on OSX, without this change. Will post tonight with more details if problem persists. Never contributed to this project before, so I may just have botched setup.

asturur:
https://github.com/Modernizr/Modernizr/blob/master/feature-detects/touchevents.js#L40

Can you please try firefox, 52 or higher what does it do?

Also unsure, if touchstart is a thing or we should check pointer events?

jordan-acosta:
`isTouchSupported` gets set to false in Chrome 60, FF 55, and true when each is switched into mobile dev mode.

Checking for `ontouchstart` is the most reliable approach. That line in Modernizr is pretty old, and the `DocumentTouch` API is obsolete: https://developer.mozilla.org/en-US/docs/Web/API/DocumentTouch

You raise a good point, though. I'll find time tonight to test across Chrome, FF, Safari, IE & Edge on my Macbook, iPad, Desktop & Android devices. I'll also look into more robust detection options.

The recommended approach these days is to support touch/pointer events in response to interactions -- in addition to mouse/keyboard -- instead of switching between mouse and touch modes. That may be beyond the scope of this fix, but I can see that being useful for the project I'm using Fabric for.

asturur:
i m available to pair on this, as i said somewhere else in the issues. I took a touch device with the purpose of fixing things.

asturur:
let's get this in.

jordan-acosta:
@asturur , sorry for going missing yesterday, putting out a fire at my day job. Thanks for merging this in!

I'm using Fabric to build a fairly sophisticated vector drawing app for mobile, so rock-solid touch support is a requirement for me.

asturur:
Please let me know how event handling works.
Not using it on touch devices ( yet ) makes hard for me build features or fix bugs.

The idea would be searching a better replacement for eventjs if exist, and add a gesture supports.
Did you do some work in this area?

jordan-acosta:
I'll let you know. Looks good so far.

I have done work in this area. I began my career doing mobile and cross-platform sites, though I've since moved on to doing full-stack work (mostly Golang these days.) I'm working on a side project to get back in touch with my front-end roots. While the project itself is proprietary, I'm using it as an excuse to contribute more to open source.

Also, I was initially just using Fabric to prototype, but it's performance has impressed me. The work you guys put into caching and off-canvas rendering really payed off. I want to see just how far I can push Fabric on mobile devices.

PR #4303, opened 2017-09-13
Paint order

stefanhayden:
This PR updates Fabric to respect the paint-order attribute in SVG2 as documented here: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/paint-order

This will both read paint-order from an SVG and output it to an svg from fabric.

Internally fabric will call this `paintFirst` as fabric only has stroke or fill to draw first and does not implement the more complicated markers. Setting `paintFirst` to `fill` or `stroke` will draw that first and will get drawn over by what is painted second.

this did break some tests that were not expecting to see `paintFirst` come out of toObject and I've add 1 tests for loading in an svg with the value, checking the value is set correctly, then exporting to SVG and then also checking for the correct output.


stefanhayden:
I see I have some cleanup to still do on this.

asturur:
And that requireOwnCache is too strict.
I think you need to force cache if you have inverted stroke + shadow.

Issue #4336, opened 2017-09-22
symmetric behavior for selection:created - selection:cleared


iongion:
This is just a question, for consistency reason, `selection:cleared` works whenever an object has been de-selected, then, isn't it expected to also have `selection:created` even for a single object ? even if it collides with `object:selected` for this case, shouldn't both be triggered ?

Now, `selection:created` is only for multiple objects, with their equivalent `selection:cleared`.
With the current life-cycle, one has to both intercept `object:selected`, for single object and `selection:created` for multiple to be able to handle a unified selection creation.

asturur:
I agree with you on this. This was a main concern also on 1.x

as of now you have:

each object can fire  selected/deselected

canvas fire object:selected ( 1 object ), selection created (2 objects ) selection updated ( 3 + ) selection:cleared and before:selection:cleared.

Before there was also the activeGroup/activeObject  difference.

To be honest object:selected could be completely replaced by selection:created or at least i could explicitly deprecate object:selected and fire selection:created with it.
The point is how much damage would occur to current applications?
While for now listening to object:selected and selection:created with the same listener is not a big deal

asturur:
https://github.com/kangax/fabric.js/pull/4340/files

This change is minimal and could bring a bit of coherence.

Need to check failing tests and side effects

asturur:
i m not closing the issue until i finished testing the new code but looks good to me.

asturur:
Readded old properties to the selection:updated and created to do not break existing apps

PR #4340, opened 2017-09-22
Normalize selection events

asturur:
This is a proposal for #4336 
Side effects to be evaluated.


asturur:
I extended the concept a bit.
Ideally now the selections are fired by one function only.
This function examines the old selection, the new selection and by difference tells if something has been updated, cleared or created.

asturur:
Also the onDeselect and onSelect functions needs to be evaluated in the multi selection cases.

PR #4356, opened 2017-09-28
isTargetTransparent fixes && cachedCanvas

stefanhayden:
fixes problems with isTargetTransparent targeting. also nulls out cachedCanvas when not being used to make it easier to check if a proper cachedCanvas exists

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -480,8 +480,9 @@
       var ctx = this.contextCache,
           originalColor = target.selectionBackgroundColor;
 
-      target.hasBorders = target.transparentCorners = false;
       target.selectionBackgroundColor = '';
+
+      this.clearContext(ctx);
 
       ctx.save();
       ctx.transform.apply(ctx, this.viewportTransform);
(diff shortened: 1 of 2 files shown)


src/canvas.class.js
```diff
@@ -480,9 +480,10 @@
       var ctx = this.contextCache,
           originalColor = target.selectionBackgroundColor;
 
-      target.hasBorders = target.transparentCorners = false;
       target.selectionBackgroundColor = '';
 
+      this.clearContext(ctx);
```
stefanhayden: moving this higher in the function makes debugging the result of this function much easier.

src/canvas.class.js
```diff
@@ -480,9 +480,10 @@
       var ctx = this.contextCache,
           originalColor = target.selectionBackgroundColor;
 
-      target.hasBorders = target.transparentCorners = false;
```
stefanhayden: these did not actually do what was intended.

src/canvas.class.js
```diff
@@ -491,15 +492,15 @@
       target === this._activeObject && target._renderControls(ctx, {
         hasBorders: false,
         transparentCorners: false
+      }, {
+        hasBorders: false,
```
stefanhayden: this is to set the child properties.

src/shapes/object.class.js
```diff
@@ -981,6 +981,7 @@
         this.drawCacheOnCanvas(ctx);
       }
       else {
+        this._cacheCanvas = null;
```
stefanhayden: if we don't set this to null it's hard to know if the _cacheCanvas is up to date or not. I could imagine some kind of performance in memory by not keeping extra canvases in memory.

asturur:
thanks @stefanhayden

PR #4369, opened 2017-10-02
Added a justify-left option for text alignment

jilster:
textAlign: 'justify' gives the default style for justified text, textAlign: 'justify-left' aligns the lines with breaks and the final line to the left.

Replaces https://github.com/kangax/fabric.js/pull/3896, which was based on 1.x

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -148,7 +148,7 @@
     linethrough:       false,
 
     /**
-     * Text alignment. Possible values: "left", "center", "right" or "justify".
+     * Text alignment. Possible values: "left", "center", "right", "justify" or "justify-left".
      * @type String
      * @default
      */
@@ -334,8 +334,8 @@
       this._splitText();
       this._clearCache();
       this.width = this.calcTextWidth() || this.cursorWidth || MIN_TEXT_WIDTH;
-      if (this.textAlign === 'justify') {
-        // once text is misured we need to make space fatter to make justified text.
+      if (this.textAlign.indexOf('justify') !== -1) {
+        // once text is measured we need to make space fatter to make justified text.
         this.enlargeSpaces();
       }
       this.height = this.calcTextHeight();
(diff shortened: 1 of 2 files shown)


jilster:
Hm, apparently i nested to deeply in enlargeSpaces. I don't know how to fix this though, i needed the extra if statement for it to work

blobinabottle:
Would be awesome to get this. Will make justify useful :-)

asturur:
I see i see... so now is `useless` ... :suspect:

blobinabottle:
😇 Let say it has potential

asturur:
Jokes apart, i see in the code you are sort of introducing the concept of paragraph in textbox. that iText does not have.

So you are detecting the hard new lines in textbox and skip justification for them.

I have some thought about that.

1) line that do not wrap will never justify?. How IText will behave?
2) would be better to call it `justify-wrapped`  because this is just justifying the wrapped lines?
In that case  Itext do not wrap will not justify but the name would sort of suggest it by itself

jilster:
@asturur Thanks for preventing me from breaking our beloved fabric.js :astonished:
@blobinabottle Thanks for still believing in me :blush:

I forgot to check how IText would behave, and i'll admit that using this feature on IText broke it. You are right, it should be called justify-wrapped because now there are 2 options that are more logical:

1. justify - justify all lines
2. justify-wrapped - justify only wrapped lines

In enlargeSpaces i changed the if statement to:
``if (this.textAlign === 'justify' || !this.hasOwnProperty('_textLineBreaks') || this._textLineBreaks.indexOf(i) === -1) {``

Here are the scenarios i tested after my second commit:

**Textbox**
Lines are wrapped by _WrapText, the _textLineBreaks property is set here.
If textAlign is 'justify', all lines are justified.
If textAlign is 'justify-wrapped', lines except the ones in _textLineBreaks are justified.

**IText & Text**
Lines are not wrapped.
If textAlign is 'justify', all lines are justified.
If textAlign is 'justify-wrapped', all lines are still justified.

Is this ok? I don't see an easy way to determine where the user using a linebreak manually, because IText and Text insert newlines after each line in their version of _splitTextIntoLines. This means we can only justify all lines, or none of the lines.

asturur:
Need to download the code and give it a shot, i guess with the current code there is anyway a method to figure out what lines are hard breaked by `\n` without collecting them at wrap time. We have a styleMap feature that should provide this information anyway.

I'll try to find time tomorrow.

jilster:
That would be great, thanks

jilster:
I guess there was no time to be found haha. Is there anything i can do? Like looking into the styleMap for example?

asturur:
Yes i m very loaded with things to do at work and in private life. Is a couple of weeks i m not working that much on fabricjs.
I'll be back

jilster:
Ok, no rush

jilster:
Continued in https://github.com/kangax/fabric.js/pull/4441

PR #4394, opened 2017-10-19
object_stacking fix

stefanhayden:
when calling moveTo on an activeSelection the code should move the objects in the canvas._objects stack instead of moving the objects in the activeObject._objects stack. This already works correctly for moveForward / moveBack / ect.

### DIFF
--- a/src/mixins/object_stacking.mixin.js
+++ b/src/mixins/object_stacking.mixin.js
@@ -69,7 +69,7 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
    * @chainable
    */
   moveTo: function(index) {
-    if (this.group) {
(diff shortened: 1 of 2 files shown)


asturur:
i agree with this fix.
Is there any change you can think of?

So the point of that code line before the group is to be able to reorder objects inside the group.

If i have a multi selection and want to reorder object inside there, i do not really care to do inside the active selection itself since their stack order is in the canvas.

If this change would not happen, i would reorder them in the active Selection stack getting 0 results in render and loosing the order as soon as the active selection is disbanded.

In case i m working with preserveObjectStacking = false, the active selection is rendered as an objet, so i would see the order swap but i would lost the effect as soon as it is disbanded.

I think is good.
Would you mind adding a UT?

asturur:
after just 3 tries....

PR #4424, opened 2017-10-28
Disable http/https requires for browser builds.

thheller:
See https://github.com/kangax/fabric.js/issues/4423

`webpack -p` build just including `fabric` goes from `334kb` to `300kb`.


asturur:
thanks!
i m not yet good with webpack to know that this thing help.

PR #4433, opened 2017-10-31
Downgrading webGL precision if target hardware does not support highest

ozooner:
So there are multiple ways to go about fixing the precision change, this is the one I came up with and like the most. If you think there's more elegant way of doing it - let me know and I can refactor and test it.

### DIFF
--- a/src/filters/base_filter.class.js
+++ b/src/filters/base_filter.class.js
@@ -20,8 +20,6 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
    */
   type: 'BaseFilter',
 
-  webGlPrecision: 'highp',
-
   vertexSource: 'attribute vec2 aPosition;\n' +
     'attribute vec2 aTexCoord;\n' +
     'varying vec2 vTexCoord;\n' +
@@ -68,8 +66,8 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
     if (!this.vertexSource || !this.fragmentSource) {
       return;
     }
-    if(this.webGlPrecision !== 'highp'){
-      fragmentSource = fragmentSource.replace(/precision highp float/g, 'precision ' + this.webGlPrecision + ' float');
+    if(fabric.webGlPrecision !== 'highp'){
+      fragmentSource = fragmentSource.replace(/precision highp float/g, 'precision ' + fabric.webGlPrecision + ' float');
     }
 
     var vertexShader = gl.createShader(gl.VERTEX_SHADER);
--- a/src/filters/base_filter.class.js
+++ b/src/filters/base_filter.class.js
@@ -20,6 +20,8 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
    */
   type: 'BaseFilter',
 
+  webGlPrecision: 'highp',
+
   vertexSource: 'attribute vec2 aPosition;\n' +
     'attribute vec2 aTexCoord;\n' +
     'varying vec2 vTexCoord;\n' +
@@ -66,6 +68,10 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
     if (!this.vertexSource || !this.fragmentSource) {
       return;
     }
+    if(this.webGlPrecision !== 'highp'){
+      fragmentSource = fragmentSource.replace(/precision highp float/g, 'precision ' + this.webGlPrecision + ' float');
+    }
+
     var vertexShader = gl.createShader(gl.VERTEX_SHADER);
     gl.shaderSource(vertexShader, vertexSource || this.vertexSource);
     gl.compileShader(vertexShader);
--- a/src/filters/webgl_backend.class.js
+++ b/src/filters/webgl_backend.class.js
@@ -2,6 +2,24 @@
 
   'use strict';
 
+
+  /**
+   * Tests if webgl supports certain precision
+   * @param {WebGL} Canvas WebGL context to test on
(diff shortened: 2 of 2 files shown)


src/filters/base_filter.class.js
```diff
@@ -20,6 +20,8 @@ fabric.Image.filters.BaseFilter = fabric.util.createClass(/** @lends fabric.Imag
    */
   type: 'BaseFilter',
 
+  webGlPrecision: 'highp',
```
asturur: can we make this fabric global property? as we have fabric.webGlprecision. texture size is also there

src/filters/webgl_backend.class.js
```diff
@@ -19,6 +37,19 @@
     if (gl) {
       fabric.maxTextureSize = gl.getParameter(gl.MAX_TEXTURE_SIZE);
       isSupported = fabric.maxTextureSize >= tileSize;
+      var precision;
+      var precisions = ['highp', 'mediump', 'lowp'];
```
asturur: are there outside implementation that works but do not support lowp? how do they work at all?

ozooner: I don't know actually - I just wrote it to be as safe as possible for any scenario.

src/filters/base_filter.class.js
ozooner: on it

asturur:
Do you have any webgl skill that go beyond the complete beginner that i am?
I would like someone to take a look at the current webgl implementation built for fabric.
There is also a lanzcos resize filter in an open PR that change something in the backend itself ( like switching between textures of different sizes )

src/filters/webgl_backend.class.js
```diff
@@ -2,6 +2,24 @@
...
+    var fragmentSource = 'precision ' + precision + ' float;\nvoid main(){}';
+    var fragmentShader = gl.createShader(gl.FRAGMENT_SHADER);
+    gl.shaderSource(fragmentShader, fragmentSource);
+    gl.compileShader(fragmentShader);        
+    if (!gl.getShaderParameter(fragmentShader, gl.COMPILE_STATUS)) {
+      return false;
+    }
+    return true;
+  }
+
```
asturur: unsure about the position of this function.
fabric.isWebglSupported  can be overridden by a dev with something more fancy.

testPrecision is here with a jsdoc, but will not be catched is not attached to any object, and can't be used from outside or from an overridden isWebglSupported. can you think of something better than calling it fabric.testPrecision?

does not block merging

ozooner: Sorry, I don't follow. Do you want to move testPrecision to fabric object as well?
Like all developers, I am not good at naming things :D

ozooner:
I changed a bit, now I am no longer handling a scenario where none of the precision are supported - I think if it passes the isWebGlSupported test, at least one precision should be supported as well.

When it comes to WebGL skill, then my entire contact with webgl is from reading your source...

asturur:
oh ok, so consider this is my first webgl experience before reusing this code. i have been helped lot from @plainview ( fifty fifty maybe ? )

src/filters/webgl_backend.class.js
ozooner: Maybe I should move the testprecision function inside isWebGlSupported function?

asturur: no i would leave it here. i was just wondering if it makes sense to have a non accessible function that can't be reused if a dev change the implementation of `isWebglSupported`. not a problem at all.

asturur:
everything looks good. i ll wait some minute and merge, just in case you spot something else.

ozooner:
Maybe I should move the testprecision function inside isWebGlSupported function since it's not used anywhere else as you say?

asturur:
that is not a big deal. there is added value here, is fine as it is.

asturur:
Do you want a fabricjs sticker? do you work with other devs that would stick it on some laptop/hardware?

ozooner:
Absolutely!

plainview:
Hey @ozooner did you consider already the gl.getShaderPrecisionFormat() method?  Was it unsuitable for this use-case?

https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext/getShaderPrecisionFormat

asturur:
what is the performance of that mali gpu anyway? what it can actually filter and at what speed?

ozooner:
@plainview - so this is what the hardware in question returns for that function:
image
As you can see, I could use that function to detect if returnObject.precision === 0. But if I go down that path, I don't know if there could be any hardware that would return 0 for `highp`, but still actually work? What do you think, should I use that function to detect precision support or existing one?

@asturur Amazon is using this GPU, because it is highly optimized for h264 decoding. For fabric I am just using tint filter to paint SVG's once, so it is not a problem for me. I am also using animations - certain animations, such as rotate/scaleX/scaleY work around 20 fps, but animating position properties (left/top) is rather laggy

plainview:
According to pyalot a precision of `0` always indicates that a float type is not supported: http://codeflow.org/entries/2013/feb/22/how-to-write-portable-webgl/#how-to-query-fragment-shader-precision

It will also always be 0 for integer types but I don't believe that is something we need to worry about here.  I am no expert on this though.

asturur:
let's keep what we have.
I'm wondering if using highp make sense for texture that goes maximum to 8k size ( texel size 1/8000 ) and for texture color that at the end are 1/255.

 This if fine and introduce the added knowledge that this thing must be checked.

i would like to understand what those numbers mean before using them.

@ozooner if you send me an address i ll send some stickers as soon as i cumulate some shipping ( a whopping 5 request for now, not worth a travel to post office ).

asturur:
image

this new travis thing is making me crazy.

ozooner:
E-mailed my address

PR #4437, opened 2017-11-04
Proposal for different lineJustify.

asturur:
@jilster @blobinabottle
replaces #4369 

This is my idea to implement it, giving us 3 different options.

To be honest i fast draft it, i did not event tested it.

### DIFF
--- a/src/mixins/itext.svg_export.js
+++ b/src/mixins/itext.svg_export.js
@@ -118,8 +118,11 @@
         charBox = this.__charBounds[lineIndex][i];
         if (boxWidth === 0) {
           textLeftOffset += charBox.kernedWidth - charBox.width;
+          boxWidth += charBox.width;
         }
-        boxWidth += charBox.kernedWidth;
+        else {
+          boxWidth += charBox.kernedWidth;
+        }
         if (isJustify && !timeToRender) {
           if (this._reSpaceAndTab.test(line[i])) {
             timeToRender = true;
--- a/src/mixins/itext.svg_export.js
+++ b/src/mixins/itext.svg_export.js
@@ -119,7 +119,7 @@
           textLeftOffset += charBox.kernedWidth - charBox.width;
         }
         boxWidth += charBox.kernedWidth;
-        if (this.textAlign === 'justify' && !timeToRender) {
+        if (this.textAlign.indexOf('justify') !== -1 && !timeToRender) {
           if (this._reSpaceAndTab.test(line[i])) {
             timeToRender = true;
           }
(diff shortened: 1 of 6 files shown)


src/mixins/itext.svg_export.js
```diff
@@ -119,7 +119,7 @@
           textLeftOffset += charBox.kernedWidth - charBox.width;
         }
         boxWidth += charBox.kernedWidth;
-        if (this.textAlign === 'justify' && !timeToRender) {
+        if (this.textAlign.indexOf('justify') !== -1 && !timeToRender) {
```
asturur: this was not replaced in the whole code in the other PR,

src/shapes/text.class.js
```diff
@@ -348,6 +348,9 @@
     enlargeSpaces: function() {
       var diffSpace, currentLineWidth, numberOfSpaces, accumulatedSpace, line, charBound, spaces;
       for (var i = 0, len = this._textLines.length; i < len; i++) {
+        if (i === len - 1 || this.isEndOfWrapping(i)) {
```
asturur: ok here needs a this.textAlign !== 'justify' && ( ..... )

asturur:
@jilster since you started the whole thing, i would be happy if yo accelerate the thing you could test this branch and open a PR over it in order fix what is not really working.
If you have time.

jilster:
Thanks, i will do that!

jilster:
See https://github.com/kangax/fabric.js/pull/4441

src/shapes/text.class.js
asturur: i still think this change is needed.
Since we are giving the user

justify-left, jsutify-right, justify-center,  this thing here should change that we do not want the old plain `justify`.

Who built an app over 2.0b7 or earlier, does not want to see old JSON saved with plain justify loaded differently just because we think is better to have last line not justified.

In this way we give an option to everyone at expense of littled added complexity, and we do not break old stuff.

jilster: yes, i agree. backwards compatibility is important

jilster:
i don't know if you were notified when i replied on a code review comment, but just to be sure: i am happy to do some last fixes if needed.

asturur:
i spent the weekend on fabricjs basically.
i cleared the top things, i want to try this now to be sure is working as expected/intended by me.
I also try to add some tests if possible.

asturur:
image

example textselection we forgot it.

asturur:
image

text decoration not working at all...

jilster:
very nice! 👍

blobinabottle:
Thank you guys! great job!

PR #4441, opened 2017-11-06
Proposal for different lineJustify small fixes

jilster:
Worked like a charm after changing styleMap to _styleMap. I tested with Textbox, Text, IText and export to SVG

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -148,7 +148,8 @@
     linethrough:       false,
 
     /**
-     * Text alignment. Possible values: "left", "center", "right" or "justify".
+     * Text alignment. Possible values: "left", "center", "right", "justify",
+     * "justify-center" or "justify-right".
      * @type String
      * @default
      */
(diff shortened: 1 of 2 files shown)


src/shapes/text.class.js
```diff
@@ -148,7 +148,8 @@
     linethrough:       false,
 
     /**
-     * Text alignment. Possible values: "left", "center", "right" or "justify".
+     * Text alignment. Possible values: "left", "center", "right", "justify",
+     * "justify-center" or "justify-right".
```
asturur: and justify-left too.

asturur:
aren't we missing some pieces? i committed more on the other side.

asturur:
ohhh this is fixing mine :D sorry i 'm busy doing other stuff.
Let's put this in and continue

src/shapes/text.class.js
jilster: I now realise what you meant here. I thought we were going to overwrite the behaviour of "justify", but we are keeping that one intact and are adding "justify-left". Are you going to change this in https://github.com/kangax/fabric.js/pull/4437 or should i make another PR?

redlive:
Is there a way to make an alignment described above per line of text as well? When I say the line I mean text split by `\n`

PR #4465, opened 2017-11-17
removed array reference for _dimensionsAffectingProps

### LINKED ISSUE #4464: Text and Textbox shared the same reference to _dimensionAffectingProps
## Version
1.7.x & 2.0 beta

## Test Case
http://jsfiddle.net/zrkwzzk1/2/

## Steps to reproduce
follow the fiddle ;)

## Expected Behavior
properties should not be affected across classes

## Actual Behavior
properties are affected across classes

I'm afraid that's a big bug affecting a lot of classes and properties because of inheritance. Array and Object should never be define as class attribute because of the nature of inheritance like fabric does (it's the same issue with Backbone.js extend for those familiar with Backbone) but always created during the `initialize` call.

Text is quite affected by Textbox on `_dimensionAffectingProps` which is used to compute `shouldClear` and affect `render`.

Using hundreds of objects it has severe performance penalty. I have a personal use case with nearly ~100% performance difference, because `shouldClear` is involved in `saveState` call which loop over every objects: this lead to a multiple nested loop.

Unfortunately maybe other bugs and perf issues are hidden in the hood of this shared reference issue. All classes should be reviewed.

asturur:
Clone the array for textbox and stop adding property at initialize
@JSteunou

close #4464

### DIFF
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -58,6 +58,13 @@
     noScaleCache: false,
 
     /**
+     * Properties which when set cause object to change dimensions
+     * @type Object
+     * @private
+     */
+    _dimensionAffectingProps: fabric.Text.prototype._dimensionAffectingProps.slice(0).push('width'),
+
+    /**
      * Constructor. Some scaling related property values are forced. Visibility
      * of controls is also fixed; only the rotation and width controls are
      * made available.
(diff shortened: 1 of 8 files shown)


src/shapes/textbox.class.js
```diff
@@ -66,11 +73,8 @@
      * @return {fabric.Textbox} thisArg
      */
     initialize: function(text, options) {
-
       this.callSuper('initialize', text, options);
       this.ctx = this.objectCaching ? this._cacheContext : fabric.util.createCanvasElement().getContext('2d');
```
asturur: this line should be go away anyway, not scope of this PR.

PR #4466, opened 2017-11-17
removed this.ctx from Textbox.

asturur:
Main purpose is to remove that weird canvas creation at object initialize

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -255,7 +255,6 @@
      */
     _render: function(ctx) {
       this.callSuper('_render', ctx);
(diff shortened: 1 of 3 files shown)


src/shapes/itext.class.js
```diff
@@ -255,7 +255,6 @@
      */
     _render: function(ctx) {
       this.callSuper('_render', ctx);
-      this.ctx = ctx;
     },
```
JSteunou: _render does nothing except calling super, can be removed I guess

asturur: yeah....

PR #4475, opened 2017-11-20
removed array reference in shapes classes

JSteunou:
_dimensionsAffectingProps cacheProperties & stateProperties
backport of #4465 + one missing on image class


JSteunou:
@asturur I had a missing one on image class, do you need a PR on v2 for it?

PR #4478, opened 2017-11-20
Remove push on shared array

JSteunou:
follow up for #4465


asturur:
thanks, i prefer to have those properties there for more visibility

PR #4492, opened 2017-11-23
Fix drawing of a point

### LINKED ISSUE #4480: Freehand drawing prevents drawing of point
## Version
2.0

## Test Case
https://jsfiddle.net/q6hw8m3n/

## Steps to reproduce
Configure canvas with `drawingMode` enabled. Click on the canvas.

## Expected Behavior
A single point to have been drawn on the canvas.

## Actual Behavior
A single point flickers briefly into view then disappears. 

This actually worked in V1 https://jsfiddle.net/oz31LqLq/1/ so I'd suggest this as a regression. Funnily enough I have vague recollections of this being an issue at one point in a previous V1 build but cannot find a corresponding issue, but may have missed it?

asturur:
close #4480

I did not take in consideration that not having more than 2 points could bring to  undefined sign for multiplication.

### DIFF
--- a/src/brushes/pencil_brush.class.js
+++ b/src/brushes/pencil_brush.class.js
@@ -145,7 +145,7 @@
       var path = [], i, width = this.width / 1000,
           p1 = new fabric.Point(points[0].x, points[0].y),
           p2 = new fabric.Point(points[1].x, points[1].y),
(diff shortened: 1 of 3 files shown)


asturur:
needs tests!

PR #4504, opened 2017-11-27
Prevent calling expensive fire events logic when not needed

JSteunou:


asturur:
this is not correct.
If the selection / deselection return false, it just means the selection/deselection process did not happen.

During a selection, with concurrent first deselection of the previous object, the deselection may happen and the new selection not ( for some constrain applied by the developer ).
We need anyway to fire the selection events.

if you need to select something without firing events you can use `_setActiveObject` and `_discardActiveObject` directly.

In which case this logic was expensive in your use case?
Is expensive on itself or for the events attached?

JSteunou:
I cannot find example of a concurrent 1st deselection that would make a *non happening selection without fire events* missing some events. A 1st deselection would trigger its own events, shouldnot it?

asturur:
example put 2 objects on the canvas, one of them add the function `onSelect = () => { return true; }';
Select the other and try to select the one with the onSelect handler.

The code called is setActiveObject, that internally calls _discardActiveObject and then _setActiveObject.

the _discardActiveObject returns true, while the _setActiveObject fails. at this point you would not fire anything but having the selection changed.

JSteunou:
The order is more `setActiveObject` which calls `_setActiveObject` which calls `_discardActiveObject` then `onSelect` but I see what you mean. The selection it canceled but not the deselection. I though the all chain of actions would be canceled I was wrong, sorry :pray:

Issue #4000, opened 2017-06-12
Feature request: ellipsis text


andrea11:
Hi,
I would like to integrate ellipsis and breakword text in textbox, in order to prevent text overflow (I use the second one also to preserve textbox original size when I edit the text). Maybe this feature will be more suitable in IText, but I have no idea how to implement in it. So I started to write a part of code (with the help of google) for Textbox. I 
```javascript
fabric.Textbox.prototype._wrapLine = function(_line, lineIndex, desiredWidth) {
      var lineWidth        = 0,
          graphemeLines    = [],
          line             = [],
          // spaces in different languges?
          words            = _line.split(this._reSpaceAndTab),
          word             = '',
          offset           = 0,
          infix            = ' ',
          wordWidth        = 0,
          infixWidth       = 0,
          largestWordWidth = 0,
          lineJustStarted = true,
          additionalSpace = this._getWidthOfCharSpacing(),
          textWidth = this._measureWord(_line, lineIndex, offset);

      if (this.ellipsis && textWidth > desiredWidth && !this.isEditing) {
        var ellipsis = '...';
        var length = _line.length;
        while (textWidth >= desiredWidth && length-- > 0) {
          _line = _line.substring(0, length);
          textWidth = this._measureWord(_line, lineIndex, offset);
        }
        graphemeLines.push([_line + ellipsis]);
        return graphemeLines;
      } else {
        for (var i = 0; i < words.length; i++) {
          // i would avoid resplitting the graphemes
          word = fabric.util.string.graphemeSplit(words[i]);
          wordWidth = this._measureWord(word, lineIndex, offset);
          offset += word.length;

          lineWidth += infixWidth;

          if (this.breakWords && wordWidth > desiredWidth || this.ellipsis && this.isEditing) {
              line.push(infix);
              // var word = word.split('');
              while (word.length) {
                  letterWidth = this._measureWord(word[0], lineIndex, offset);
                  if (lineWidth + letterWidth > desiredWidth) {
                      graphemeLines.push(line);
                      line = [];
                      lineWidth = 0;
                  }
                  line.push(word.shift());
                  offset++;
                  lineWidth += letterWidth;
              }
              word = '';
          } else {
              lineWidth += wordWidth - additionalSpace;
          }

          if (lineWidth >= desiredWidth && !lineJustStarted) {
            graphemeLines.push(line);
            line = [];
            lineWidth = wordWidth;
            lineJustStarted = true;
          }

          if (!lineJustStarted) {
[...]

PR #4513, opened 2017-11-28
Allow cutted words in a Textbox shape

### LINKED ISSUE #2376: Entering text in a Textbox object expands it when there aren't newlines

JSteunou:
It's a way to mimic html/css world *white-space normal or pre-wrap* property, because sometime you want the text to fit exactly in the `width` you set no matter what.

Example here http://jsfiddle.net/Da7SP/736/ one with `minWidth` one without, both with the option activated and a last one without it.

this would close #2376

### DIFF
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -301,6 +301,7 @@
       for (var i = 0; i < words.length; i++) {
         // i would avoid resplitting the graphemes
         word = fabric.util.string.graphemeSplit(words[i]);
+        cutIndex = word.length - 1;
         wordWidth = this._measureWord(word, lineIndex, offset);
 
         lineWidth += infixWidth + wordWidth - additionalSpace;
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -354,7 +354,7 @@
 
       i && graphemeLines.push(line);
 
-      if (largestWordWidth > this.dynamicMinWidth) {
+      if (!this.allowCuttedWords && largestWordWidth > this.dynamicMinWidth) {
         this.dynamicMinWidth = largestWordWidth - additionalSpace;
       }
 
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -41,6 +41,13 @@
     dynamicMinWidth: 2,
 
     /**
+     * Allow cutted words to fit in the textbox width
+     * pretty much like css white-space normal or pre-wrap
+     * @type {Boolean}
+     */
+    allowCuttedWords: false,
+
+    /**
      * Cached array of text wrapping.
      * @type Array
      */
@@ -288,12 +295,13 @@
           infixWidth       = 0,
           largestWordWidth = 0,
           lineJustStarted = true,
+          cutIndex = 0,
+          cuttedWord = [],
           additionalSpace = this._getWidthOfCharSpacing();
       for (var i = 0; i < words.length; i++) {
         // i would avoid resplitting the graphemes
         word = fabric.util.string.graphemeSplit(words[i]);
         wordWidth = this._measureWord(word, lineIndex, offset);
-        offset += word.length;
 
         lineWidth += infixWidth + wordWidth - additionalSpace;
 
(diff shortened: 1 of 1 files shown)


asturur:
oh this has an open issue around the tracker.
With a similar implementation that i provided some time ago ( before text rewrite so no more usable ).

This has to wait 2.0 release we can discuss it meantime to see perf issue/testing/ problem that may arise.

JSteunou:
Perf are not bad, it's just like having more tiny words. Like the other PR I will need guidance for testing.

I found these related issue  #2376 & #4000

lsiann:
The last letter of a cutted word can not be selected.Maybe need a fix somewhere else

sentoro:
Does anyone going to fix this issue? the text wrap still not working for me in 2.0

asturur:
What is to be fixed actually?

blobinabottle:
hey @asturur any idea why the bug mentioned? The graphemeLines returned by _wrapLine seems correct. Where would you look to try to fix it?

I know it's not in your roadmap, but this functionality could help us a lot :-)

asturur:
We can revisit it. I do not remember.
I wonder if it makes sense to break works in this way, without a breaking sign in the middle.

blobinabottle:
The use for us is more about using asian language (without spaces apparently) in textbox and being able to resize them. For the moment it's impossible (because considered as a long non-breaking word).

asturur:
Apparently.
Maybe they use other dividers, while fabricjs split just with space, like an illitterate barbarian.
That is something to fix anyway, if we did not do already, including the other 2 or 3 unicode spaces that are out there.

blobinabottle:
http://www.fileformat.info/info/unicode/category/Zs/list.htm I guess this is an helping list. I see U+3000 as ideographic space

blobinabottle:
But this seems way more complicated as Japanese doesn't have space at all. There are libraries to detect words:
https://stackoverflow.com/questions/49367783/js-split-for-japanese-string

After trying in several softwares they just consider Japanese as a big long word:
https://cl.ly/e2600c23a296 (result in Illustrator, ideally it should be segmented as two "words"  '公開' and '文書' )

So I think the approach of this PR is probably the easiest. It's working whatever the language you're using.

asturur:
what we can do is making the splitWords function a part of fabric.util as much as grapheme split so that devs can override it when they need to support other languages.

asturur:
Is also still unclear if the wordbreaking should trigger just for words longer than width or always to fill up spaces. I can foresee people coming with good reasons for the opposite feature we will choose.

blobinabottle:
That doesn't make sense. If you try any other softwares it behave as this PR... 
And you can still use the option allowCuttedWords for the ones unhappy about it. No?

asturur:
i think you misunderstood me.
Should you split each word that is bigger than width, or split each word to fit width, when the option is active?

blobinabottle:
Ah but then that's something else; "hyphenation" I think. And I guess you could only make it for English (or languages where you know how to cut words). 
But you can only find this options in "advanced" softwares (illustrator, indesign, photoshop). Sketch Figma and Framer are doing as this PR and they have no option about hyphenation.

blobinabottle:
Google Slides and keynote are doing the same (no hyphenation): breaking word longer than the desired width.

blobinabottle:
What I can tell is that this PR is working pretty well with styles.
toSVG is perfect as well. 

Only issues I saw:
- cursor and selection position (we have 1 extra char calculated by line created by cutted word)
- if you resize big texttblock (font size > 40px) to a veiry small width, cutting word will stop working.

I can offer to beta-test this PR as much as possible  (if this can help you :-D)

asturur:
well i want to merge clippath first, and give a way a new render method for
higher res. I also need interaction tests in general.

Publishing features on top of features and leaving bugs open is frustrating
for me and for library user in general.

asturur:
I do not remember now, i need to read on the computer. I do remember that measuring text in order to split it, instead of after splitting it, was a problem.

I ll be back in the evening and try to point you in some direction

JSteunou:
FYI I'm not a big fan of how I did this. It should have been written as a recursive functional code, but the old code is imperative so I kept imperative code. But for cases like this one, imperative is not easy to read.

asturur:
you should propose better code if you have it. readibility can be addressed with comments.

I care it works, i remember my first patch could split a long word in many lines.

PR #4519, opened 2017-11-30
Reinsert text shortcut for some cases

### LINKED ISSUE #4514: V2 is slower at rendering simple text
v1 has a bypass for empty styles that v2 has not causing lines is twice longer to render. When using fabric use a log of text shapes this is a 100% perf hit compared to v1.

Temporary solution: using cache on those shapes (which improve perf a lot but add a little blur when scaling up and down)

Backport solution: porting the v1 check and logic into v2 like

```js
      if (this.isEmptyStyles()) {
        return this._renderCharsFast(method, ctx, line, left, top);
      }
```

asturur:
Add a faster rendering in case we do not have nor style nor alignement for a particular line

close #4514


asturur:
we can skip the join since we have already the text ready in another array

PR #4570, opened 2017-12-19
Added added/removed events to fabric.Group #4569

### LINKED ISSUE #4569: Events firing on groups
## Version
2.0

## Test Case
http://jsfiddle.net/wferris722/ajug6Ltw/1/

## Steps to reproduce
Create grouped objects and attach object:added and object:removed events to group and added / removed to objects. Add / remove object from group

## Expected Behavior
object:removed/object:added fired on group object
removed/added fired on object

## Actual Behavior
No events fired on either group or object

wferris722:
Resolves #4569 by adding required code for firing events to _onObjectAdded and _onObjectRemoved

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -18,6 +18,8 @@
    * @mixes fabric.Collection
    * @tutorial {@link http://fabricjs.com/fabric-intro-part-3#groups}
    * @see {@link fabric.Group#initialize} for constructor definition
+   * @fires object:added
+   * @fires object:removed
    */
   fabric.Group = fabric.util.createClass(fabric.Object, fabric.Collection, /** @lends fabric.Group.prototype */ {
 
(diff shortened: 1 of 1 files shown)


src/shapes/group.class.js
```diff
@@ -98,8 +100,19 @@
...
     },
 
+    /**
+     * @private
+     */
+    _objectCleanup: function(){
+      while (this._objects.length){
+        this.remove(this._objects[0]);
+      }
+    },
```
asturur: this and the automatic event binding are not part of the PR description.
Also this has a problem:
if you put a group inside a group, when you remove the group from the outside group, this will loose all objects.

asturur:
i'm not sure about this change.
For sure the automatic event binding on group initialize is something we do not want

wferris722:
You're right, I'll modify the PR to limit it to _onObjectAdded and _onObjectRemoved, and get rid of the automatic event binding. 

I was thinking of the added/removed event as being added/removed from the canvas rather than the object that contained it, which was wrong. It also causes the problem you mentioned above.

Issue #4516, opened 2017-11-29
Problem building a fabric app with webpack and babel


pvanliefland:
## Version
2.0 rc3

## Test Case

Not applicable

## Steps to reproduce

Use webpack with babel-loader to transpile a project that include fabric.

## Expected Behavior

No ES6 code is included in the build.

## Actual Behavior

Non-transpiled ES6 code from jsdom is included in the build (and crashes the app in older browsers).

-----------------------------------------------------------------------------

I think the problem lies in this commit: https://github.com/kangax/fabric.js/commit/71b36f6228c4d56d3e0d41942cb2ce4e3bfb1f90#diff-f38f28902dcbc587be375ebd446c5341.

If you inspect what lies in `jsdom/lib/jsdom/living/generated/utils`, you can see that there is some ES6 code:

```javascript
module.exports.mixin = function mixin(target, source) {
  const keys = Object.getOwnPropertyNames(source);
  for (let i = 0; i < keys.length; ++i) {
    if (keys[i] in target) {
      continue;
    }

    Object.defineProperty(target, keys[i], Object.getOwnPropertyDescriptor(source, keys[i]));
  }
};
```

Problem is, when you build your app with webpack, you typically don't transpile code from `node_modules` - you expect it to be free from ES6 code.

asturur:
the problem is that when you build your app with webpack you do not expect JSDOM to finish inside. JSDOM is node only, or if it run in the browser is not something you generally want in.

requiring that function in that way is confusing webpack i presume since it runs on node and so `isLikelyNode` is on.

Do you have an idea how to solve it?

asturur:
summoning @stefanhayden a notorius webpack expert

stefanhayden:
yeah I'm really no webpack expert.

I can think of a few ways to solve this. 

One would be for fabric to have a web and node build.

Perhaps the easiest way is to ignore the jsdom module using `new webpack.IgnorePlugin(/jsdom/)`. http://remarkablemark.org/blog/2017/02/25/webpack-ignore-module/

I've not tried it myself but is seem to be perfect for this situation. If it does work as advertised it might be worth updating some docs on how to do this.

pvanliefland:
The web build and node build is probably the best solution.

[...]

PR #4576, opened 2017-12-20
Ignore jsdom utils for browsers build

### LINKED ISSUE #4575: jsdom utils is imported even for browsers at build
Some part of jsdom are included for nodejs but it has some side effect on browser build https://github.com/kangax/fabric.js/blob/master/HEADER.js#L21

1. useless code is included
2. it is not ES5 compatible and generally not transpile by babel as it is inside node_modules

JSteunou:
Close #4575 & #4516

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
     "canvas": false,
     "fs":      false,
     "jsdom":   false,
+    "jsdom/lib/jsdom/living/generated/utils": false,
+    "jsdom/lib/jsdom/utils": false,
     "http":   false,
     "https":   false,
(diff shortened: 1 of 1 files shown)


JSteunou:
FYI lib like node-uuid is this feature to alias some modules between nodejs & browsers  https://github.com/kelektiv/node-uuid/blob/master/package.json#L24

Might give you some ideas ;)

asturur:
oh that feature is cool for polyfills.
Thanks this is good.

stefanhayden:
yeah this looks good to me. thanks!

PR #4579, opened 2017-12-22
Fix color toHexa()

jparkrr:
Need to strip out the decimals.

### DIFF
--- a/src/color.class.js
+++ b/src/color.class.js
@@ -191,7 +191,7 @@
     toHexa: function() {
       var source = this.getSource(), a;
 
-      a = source[3] * 255;
+      a = Math.round(source[3] * 255);
       a = a.toString(16);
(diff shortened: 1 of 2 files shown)


jparkrr:
@asturur did you see this one? Currently `color.toHexa()` can give things like `FFFFFF19.8`

asturur:
i saw it, but between wekeend and christmas i did not have time to check it

src/color.class.js
```diff
@@ -192,7 +192,7 @@
       var source = this.getSource(), a;
 
       a = source[3] * 255;
-      a = a.toString(16);
```
asturur: can't we use math.Round on source[3] * 255 instead of using a split string?

jparkrr:
@asturur Great idea on rounding! Updated

Sorry if I rushed you, take your time. Hope you had a good Christmas!

PR #4595, opened 2017-12-31
Introduced .npmrc file to prevent package-lock files being created

jedateach:
I assume that because there is no package-lock file in the project, then we don't want one to ever be committed. This will prevent anyone creating one during `npm install`

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false


asturur:
ok we can merge this straight away.

Shrinkwrapping is a pain, and we do not really need it here.

PR #4597, opened 2017-12-31
Testing lint as a separate stage

asturur:
@jedateach  i m testing stages.

Please feel free to leave any comment/criticism

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,11 @@ matrix:
     - node_js: "6"
     - node_js: "4"
 
-script: 'npm run build && testem ci --port 8080 -f testem.json -l $LAUNCHER'
+script: 'npm run lint && npm run lint_tests && npm run build && testem ci --port 8080 -f testem.json -l $LAUNCHER'
 
-jobs:
-  include:
-    - stage: Linting
-      script: 'npm run lint && npm run lint_tests'
+install:
+  - if [ "$LAUNCHER" != "Node" ]; then npm install --no-optional; fi
+  - if [ "$LAUNCHER" == "Node" ]; then npm install; fi
 
 before_install:
   - if [ "$LAUNCHER" == "Node" ]; then sudo apt-get update -qq; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,6 @@ jobs:
     - stage: Linting and Building
       before_install: skip
       env: STEP=BUILD
-      cache:
-        directories: skip
       install: npm install uglify-js@3.1.x
       script: 'npm run build'
     - stage: test
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,6 @@ env:
   global:
     - LAUNCHER=Node
 
-# matrix:
-#   include:
-#     - node_js: "9"
-#     - node_js: "8"
-#     - node_js: "6"
-#     - node_js: "4"
-
 jobs:
   include:
     - stage: lint_build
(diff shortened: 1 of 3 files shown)


asturur:
image

Travis is like this from yesterday, wondering what is going on, i rarely see it soo busy.

asturur:
ok definitely wrong way to do it.

jedateach:
I think you could safely call the stage 'lint'.

build.js
```diff
@@ -23,7 +23,7 @@ var noStrict = 'no-strict' in buildArgsAsObject;
 var noSVGExport = 'no-svg-export' in buildArgsAsObject;
 var requirejs = 'requirejs' in buildArgsAsObject ? 'requirejs' : false;
 var sourceMap = 'sourcemap' in buildArgsAsObject;
-
+var buildFast = 'fast' in buildArgsAsObject;
```
jedateach: I have no issue with a build fast step, but want to note that I don't think the regular build is slow.
Because we can have a 'build' stage, we only need to run once anyway.

asturur: so the `fast`build is just to skip the uglify and multiple file creation and uglify-js installation too.
Also i will use it in build:watch so that when i updated files when developing, the test website is update a couple of seconds faster.

all the tests runs on the non minified build, so there is no need to minify it.
Is important to test the minification in the lint stage because uglify can fail sometimes with something that lint does not catch.

asturur: Also each stage start a fresh container, so i have to build each stage.
I do not think caching the dist folder ( that change each PR  and so gets updated by travis ) is a good thing, so for now we are stuck with a build for each stage and job, so i think the faster the better

.travis.yml
```diff
@@ -2,23 +2,41 @@ language: node_js
...
   include:
-    - env: LAUNCHER=Firefox
-    - env: LAUNCHER=Chrome
-    - node_js: "9"
-    - node_js: "8"
-    - node_js: "6"
-    - node_js: "4"
+    - stage: lint_build
+      before_install: skip
+      install: npm install uglify-js@3.1.x eslint@4.7.x
```
jedateach: Doing these specific package installs will make things harder to maintain. If you ever change a pacakge version, you need to change it in multiple places.

I think we should leave `npm install` to run completely.
To help improve `npm install` speed We can cache node_modules folder, as per these instructions:
https://docs.travis-ci.com/user/caching
https://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies

Travis also supports yarn now, which might add some speed improvement:
https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

asturur:
i completely agree.  i m testing out things.

Caching the npm folder  is the first thing i thought off when i started digging in docs.

On the side of mantainability i rarely touch the dev deps and i think i could eventually mantain them manually if i do not manage to run it faster.

Since node-canvas has a build process step i m worried it will invalidate the cache each build and will start to reupload it each build.

It requires some more dedicate time to test it out.

I was also thinking to leave in devdeps only the devdeps required to run continuos integration tests and add a npm command to install the one i use locally to develop: onchange, node-qunit, instanbul

jedateach:
I've had a better read of the stage docs.
That's disappointing we can't share build files without using s3. Perhaps something to consider eventually.

I've realized we've improved things by splitting out linting into a stage so linting is run once, but this could in theory be run in parallel with the tests. On closer inspection, test runs are not really parallel - it is probably based on Travis load / open source policy for simultaneous jobs. I'm liking the separate linting stage for now.

The cache setup step is surprisingly slow. In fact in my little bit of testing, Travis seems really slow.

asturur:
I prefer lint non in parallel just as an attitude to travis.
If lint is broken there is no point in running the test, i prefer to pick up one instance, use it for 1 minute, discover i forgot to run lint before pushing, and fix it.

i m trying to figure out if smaller caches helps with timings.

PR #4618, opened 2018-01-11
Add CrossOrigin managment to fabric.Pattern

### LINKED ISSUE #4612: Force anonymous cross origin on every load?
I'm getting sick of having to specify {"crossOrigin", "anonymous"} for every web resource.

I also have a problem where setting a pattern on a shape is fine until you save it to json and then decide to try restore it. Which doesn't let you specify via an attribute or just luck the crossOrigin attribute.

Is there a way of setting it on the prototype or can someone help work out how to get it working on loadFromJSON when a shape or font has a pattern? As I'm always ending up with a dirty canvas.

asturur:
closes #4612

Add crossOrigin option to fabric.Pattern in order to be able to deserialize image based pattern that requires crossorigin.


asturur:
Needs tests.

PR #4626, opened 2018-01-17
Add patternTransform toObject saving

### LINKED ISSUE #4625: patternTransform not saving via canvas.toJSON
patternTransform not saving via canvas.toJSON.
Any way to change that?

DangerD256:
closes #4625

### DIFF
--- a/src/pattern.class.js
+++ b/src/pattern.class.js
@@ -95,7 +95,7 @@
         repeat: this.repeat,
         offsetX: toFixed(this.offsetX, NUM_FRACTION_DIGITS),
         offsetY: toFixed(this.offsetY, NUM_FRACTION_DIGITS),
-        patternTransform:  this.patternTransform
+        patternTransform: this.patternTransform ? this.patternTransform.concat() : null
       };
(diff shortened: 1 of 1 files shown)


src/pattern.class.js
```diff
@@ -95,6 +95,7 @@
         repeat: this.repeat,
         offsetX: toFixed(this.offsetX, NUM_FRACTION_DIGITS),
         offsetY: toFixed(this.offsetY, NUM_FRACTION_DIGITS),
+        patternTransform:  this.patternTransform
```
asturur: patternTransform: this. patternTransform ? this. patternTransform.concat() : null,

this allow to avoid sending out a reference to the real matrix that could be changed by other function later.
A sort of shallow protection on mutations.

You do not expect your object to to influence rendering once exported.

asturur:
@DangerD256 do you think you can make this correction?

src/pattern.class.js
FLKQ: great version ( at this version you broke video scalling ;))

asturur: ?

PR #4631, opened 2018-01-18
Update use of getBoundingRect in Guides Example

jrking4:
Aligning Guidelines Example broken. Update to use getBoundingRect() function

### DIFF
--- a/lib/aligning_guidelines.js
+++ b/lib/aligning_guidelines.js
@@ -85,8 +85,9 @@ function initAligningGuidelines(canvas) {
       var objectCenter = canvasObjects[i].getCenterPoint(),
           objectLeft = objectCenter.x,
           objectTop = objectCenter.y,
-          objectHeight = canvasObjects[i].getBoundingRect().height / viewportTransform[3],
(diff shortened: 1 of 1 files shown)


asturur:
thanks!

do guidelines work correctly? i do not check them since long

jrking4:
Yes they work well in 2.0. Was just throwing a cant find getBoundingRectHeight() function error. So I updated the functions.

lib/aligning_guidelines.js
```diff
@@ -84,8 +85,8 @@ function initAligningGuidelines(canvas) {
       var objectCenter = canvasObjects[i].getCenterPoint(),
           objectLeft = objectCenter.x,
           objectTop = objectCenter.y,
-          objectHeight = canvasObjects[i].getBoundingRectHeight() / viewportTransform[3],
-          objectWidth = canvasObjects[i].getBoundingRectWidth() / viewportTransform[0];
+          objectHeight = canvasObjects[i].getBoundingRect().height / viewportTransform[3],
+          objectWidth = canvasObjects[i].getBoundingRect().width / viewportTransform[0];
```
asturur: please use the same logic here, call boundingRect once, and then read width/height.

PR #4667, opened 2018-02-02
toDataURL callback for jpeg

nobuttsseb:
Problem: in nodejs there was no way to get **jpeg** data back from canvas as fabric did not provide a way to pass through a callback.

Solution: just added in a callback argument in the functions from fabric.toDataURL to canvas.toDataURL


asturur:
JPEG todataurl is async on nodecanvas?

nobuttsseb:
Yup, it is async. The jpegStream in canvas has a stream where the callback with the base64 data triggers when the stream ends.

asturur:
how jsdom solves this? any idea?

nobuttsseb:
No idea, sorry.

nobuttsseb:
The same code diff needs to apply to 1.x branch. I'm just updating my fork.

brandondong:
Hi, I'm running into this issue as well. What's the status of this PR? Are there any known workarounds for this? It looks like node-canvas recently merged a fix: https://github.com/Automattic/node-canvas/pull/1194.

asturur:
So if we could backport the nodecanvas fix in the 1.x branch that would be awesome, otherway we need to update to nodecanvas 2, and fork jsdom 9.x in order to support nodecanvas 2 or fork jsdom latest in order to support nodecanvas 2 and also drop support for node 4 and so bump fabric version to 3.0

asturur:
this has been fixed with nodecanvas 2

PR #4668, opened 2018-02-02
FIX firing selection:cleared when group is changing to active selection.

jgprogram:
There is problem when we changing group to activeSelection because canvas fire selection:cleared. 

This commit silently unselect group before removing it from canvas and it should prevnet trigering before:selection:cleared and selection:cleared events in canvas _onObjectRemoved.

Other way to resolve it could be fire event selection:created after set canvas._activeObject = activeSelection;

### DIFF
--- a/src/shapes/active_selection.class.js
+++ b/src/shapes/active_selection.class.js
@@ -76,6 +76,7 @@
       canvas.add(newGroup);
       canvas._activeObject = newGroup;
       newGroup.setCoords();
+      canvas.trigger('selection:updated', { target: newGroup });
       return newGroup;
     },
 
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -432,6 +432,7 @@
       var activeSelection = new fabric.ActiveSelection([]);
       activeSelection.set(options);
       activeSelection.type = 'activeSelection';
+      canvas._activeObject = activeSelection;
       canvas.remove(this);
       objects.forEach(function(object) {
         object.group = activeSelection;
@@ -440,8 +441,8 @@
       });
       activeSelection.canvas = canvas;
       activeSelection._objects = objects;
-      canvas._activeObject = activeSelection;
       activeSelection.setCoords();
+      canvas.trigger('selection:updated', { target: activeSelection });
       return activeSelection;
     },
 


asturur:
Wait, i think fabric should just fire selection:modified that is the proper event in this case.
What do you think?

jgprogram:
I am agree with You it's the best approach that method should only propagate selection:modified event. In this case we should update ActiveSelecetion.toGroup() too, beacuse there is the same problem.

If we want to be perfect imho we should fix object:remove, object.add events too. As You can see now those methods triggers that events because toGroup and to toActiveSelection is using canvas.add() and canvas.remove() and that propagate these events.

I will look deeper that problem and I will try prepare new commit.

asturur:
Let's share your findings and ideas, before using time writing code.
Let's decide what is best and then code it.

jgprogram:
## Problem (how it works now)
Now when we invoke activeSelection.toGroup() and group.toActiveSelection() canvas triggering not needed events like before.selection.cleared, selection.cleared, object:added, object:removed etc. The reason of that is using canvas.remove(), canvas.add() etc. As we know these methods are from Collection mixin where after work they are triggering eg. _onObjectRemove or _onObjectAdded. 

## How it should work
ActiveSelection.toGroup() and Group.toActiveSelection() shouldn’t propagate event when it is manipulating with objects in canvas. These methods should propagate only one event selection:modified after work.

## How we can achieve it

**Proposal 1**
It’s naive and messy way but I think it should work. We can use methods canvas.remove(obj), canvas.add(obj) etc. without changes but before invocation we must unset event handlers _onObjectRemove, _onObjectAdded etc. and after operation set handler back. 

Example when we want remove object from canvas:
```
let onObjRemove = canvas._onObjectRemove;
canvas._onObjectRemove = null;
canvas.remove(obj);
canvas._onObjectRemove = onObjRemove;
```

Similarly we can do it for canvas.add(obj) and other operations.

**Proposal 2**
In this way we can replace canvas.remove(obj), canvas.add(obj) by raw operations.

Example when we want remove object from canvas we can:
```
var objects = canvas.getObjects();
index = objects.indexOf(groupToRemove);
if (index !== -1) {
    objects.splice(index, 1);
}
```

Example when we want add object on canvas:
```
var objects = canvas.getObjects();
objects.push.apply(newActiveSelection);
```

And after all whe can fire event selection:modified.

Do You have any other proposal?

asturur:
ok i will read this long post today or tomorrow and we can move on.

asturur:
So i think that effectively, canvas added and canvas removed, are really too much here, and a selection:updated is all we need.

let me see what i think is best to do that, without creating other bugs

asturur:
I would start by avoiding firing too much selection events, and think later of add/remove idea.
In the toActiveSelection, probably all we need to do is move 
canvas._activeObject = activeSelection before the group removal.

And before returning we could manually fire the selection:modified, passed the right options to calculate manually. ( Maybe there is a method for it already ).

I would fix this simple case, and then move to toGroup use case.
What do you think?

jgprogram:
I have created new commit based on Your advises. I added event triggering before returning from `toGroup` and `toActiveSelection` methods. I've run it in test environment and now events work fine.

So, what do You think about add/remove events?

PR #4688, opened 2018-02-08
change way for mousemove when freedraw,O(n) => O(1) 

boomyao:


asturur:
We have a problem here,
_render needs to be able to redraw everything if needed.
If you want to draw just one segment at time, look at how the other brushes works ( circleBrush i believe is a good example)

boomyao:
i see, i add _curPointIdx at "_reset" func record the next point index and do away with clear canvas context  mousemove, i try to use circleBrush, it's still work.

asturur:
@boomyao did you understand my comment, and can you update the PR?

asturur:
Where is the old code was actually good, we just need to understand each other.

asturur:
What we need to do is split the render method in 2 functions, one able to render a single segment, another able to render everything, for compliance with the library.
I probably failed in comunicating this to you. lets reopen this PR and fix it. I'll help

boomyao:
sorry, my english is so poor.
i add a new pr #4743

asturur:
I see the new PR, but this was better.

boomyao:
no......  fine, i will back to this PR.

boomyao:
hi! i have updated at #4743

Issue #4693, opened 2018-02-08
multiplier in toDataURL is not working


loserford:
## Version
2.0

It should be caused by the change of this line
```
multiplier = options.multiplier || options.enableRetinaScaling ? 1 : 1 / this.getRetinaScaling(),
```

Actually () is missing after ||
```
multiplier = options.multiplier || (options.enableRetinaScaling ? 1 : 1 / this.getRetinaScaling()),
```

Xylaw:
I can confirm that

asturur:
What is not working to be clear?

PR #4698, opened 2018-02-09
#4693: multiplier in toDataURL is not working

loserford:
It's quite a trivial bug.


asturur:
i think the bug is even bigger.
Because that change was supposed to enable/disable retina scaling.
How did test pass...

PR #4699, opened 2018-02-09
do not mutate passed object to fromObject

stefanhayden:
if this is not shallow cloned we end up mutating the pass object which we do not want to do.

I would like to add a test for this but I can't get the tests to run on my machine.


asturur:
we need a small test to be sure we do not break it again.
Let me see if i can fast copy one.

PR #4712, opened 2018-02-12
Show defaultCursor for none selectable objects

Kienz:
If none selectable objects hovered show defaultCursor instead of hoverCursor.

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -826,6 +826,7 @@
       }
 
       var hoverCursor = target.hoverCursor || this.hoverCursor,
+          defaultCursor = target.defaultCursor || this.defaultCursor,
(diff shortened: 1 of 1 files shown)


asturur:
We already discussed this once, and i think we have to make it straight in the docs and to the users.
So the name hoverCursor suggest me that is a cursor that appears on hover.
Otherwise would be called `beforeSelectCursor` or `hoverCursorIfSelectable`

So the point is that we have to decide and state clearly if:
the developer that wants to show a `not-allowed` cursor on hovering an object has to set it to the object defaultCursor or to the object hoverCursor.

Do you understand what i mean? My comment may look like a pun, but is not. I'm fine with any approach as long is consistent along the library/docs.

asturur:
Stefan please share some ideas.

PR #4743, opened 2018-02-19
add drawDot func to pencilbrush

### LINKED ISSUE #4680: freedrawing slowly on low-profile phone.
```
for (i = 1, len = this._points.length; i < len; i++) {
        // we pick the point between pi + 1 & pi + 2 as the
        // end point and p1 as our control point.
        var midPoint = p1.midPointFrom(p2);
        ctx.quadraticCurveTo(p1.x, p1.y, midPoint.x, midPoint.y);

        p1 = this._points[i];
        p2 = this._points[i + 1];
      }
```

this is the way to free draw at mousemove event, but i think it's a bad way to draw on canvas.

boomyao:
add simplePencil_brush that _render not needs to redraw everything, suitable for low-profile phone.

close #4680


asturur:
Ok so this is what i meant.
- _render stays untouched like in your PR
- drawDot ( i renamed _drawSegment ) is just a small pieces of _render that can draw just a part
- _render uses drawDot for code reuse.

In this way i achieved exactly your goal, but in addition i minimized code changes. It make easeir to see what changed.

boomyao:
cool ! so beautiful code you do.

asturur:
Do not worry is just practice.
But for contributing to an open project is important to minimize changes, it makes easier for the reviewer to understand what is going on, and also makes easier to understand if something is going to break with the change.

In this way the capture logic is unchanged, and we should not be worried. The drawing logic still needs to be tested manually but it looks good to me.

Please confirm me that it draws as intended and that performances are improved on slow phones, and i ll merge it.

boomyao:
@asturur, maybe this PR should be close.
if don't exec ctx.beginPath() everytime, ctx.stroke() will render the path many time, then the path is bold, so i think it's the reason to draw all points when mousemove.

asturur:
no i think we can fix it. let's try

asturur:
i think we are good with this latest change.
Thanks @boomyao

boomyao:
@asturur Great ! ! Thank you for give me a lesson!

PR #4756, opened 2018-02-21
cleanup function to be more readable

stefanhayden:
This is not exactly fabric's style and a lot of this is personal preference but I find these changes make the function more readable and I believe would have made finding the bug we previously fixed a little easier.

### DIFF
--- a/src/mixins/text_style.mixin.js
+++ b/src/mixins/text_style.mixin.js
@@ -63,7 +63,7 @@
       if (!this.styles || !property || property === '') {
         return false;
       }
-      var obj = this.styles, stylesCount = 0, letterCount, style,
+      var obj = this.styles, stylesCount = 0, letterCount, stylePropertyValue,
           allStyleObjectPropertiesMatch = true, graphemeCount = 0, styleObject;
       // eslint-disable-next-line
       for (var p1 in obj) {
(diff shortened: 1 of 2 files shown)


src/mixins/text_style.mixin.js
```diff
@@ -63,31 +63,42 @@
...
+          if (stylePropertyHasBeenSet) {
+            if (!style) {
+              style = styleObject[property];
+            }
+            else if (styleObject[property] !== style) {
+              allStyleObjectPropertiesMatch = false;
+            }
+
+            if (styleObject[property] === this[property]) {
+              delete obj[p1][p2][property];
```
asturur: this should be a reference to styleObject, should be fine to use it to delete too.

src/mixins/text_style.mixin.js
```diff
@@ -63,31 +63,42 @@
       if (!this.styles || !property || property === '') {
         return false;
       }
-      var obj = this.styles, stylesCount = 0, letterCount, foundStyle = false, style,
-          canBeSwapped = true, graphemeCount = 0;
+      var obj = this.styles, stylesCount = 0, letterCount, style,
+          allStyleObjectPropertiesMatch = true, graphemeCount = 0, styleObject;
```
asturur: what do you think about:
style  => stylePropertyValue

stefanhayden: clearly I love verbose names!

src/mixins/text_style.mixin.js
stefanhayden: true!

Issue #4755, opened 2018-02-21
setSelectionStyle stopped working in 2.0.x ?


blobinabottle:
## Version
2.0.3

## Test Case
In 2.0.3 http://jsfiddle.net/blobinabottle/yab4fk2z/3/
Vs 1.7.20 http://jsfiddle.net/blobinabottle/yab4fk2z/2

## Steps to reproduce
Select some texts and the end of paragraph and hit bold. On 1.7.20 it's ok. Not in 2.0.3

## Expected Behavior
Text should go bold. Selection start & end are correct. 

## Actual Behavior
Text do nothing.

Note: in 2.0 rc3 it's working. Don't see what changes could have an impact on this...

asturur:
i wonder what unit tests are doing

blobinabottle:
And if you set it by code on a defined range, it can go super crazy: 
http://jsfiddle.net/blobinabottle/yab4fk2z/10/

But seems ok if the style is only set on the first line....

asturur:
that crazy part is the one in the other ticket.
That thing is unfixable.
Everytime you fix a bit you break it somewhere else

blobinabottle:
But here it's setSelectionStyles that's messing things up. I just called textbox.setSelectionStyles({fontWeight:'bold'},100, 115); 
15 characters, and it set the whole line in bold. It doesn't know what's bold or not...

asturur:
![image](https://user-images.githubusercontent.com/1194048/36509448-146f7a0a-1760-11e8-9754-aa68e9bd0864.png)

on kitchensink works, looking at the code i see it has failures points.

asturur:
This is the straight forward kitchensink code

```javascript
function setActiveStyle(styleName, value, object) {
  object = object || canvas.getActiveObject();
  if (!object) return;

  if (object.setSelectionStyles && object.isEditing) {
    var style = { };
    style[styleName] = value;
    object.setSelectionStyles(style);
    object.setCoords();
  }
  else {
    object.set(styleName, value);
  }

  object.setCoords();
  canvas.renderAll();
};
```

asturur:
Ok i see the error but i cannot fix it now since it needs some thinkering.
This should fix also the error when extending.

If you want to try by yourself:

file: text_style.mixin.js, function _extendStyles, line ```var loc = this.get2DCursorLocation(index);```  becomes ```var loc = this.get2DCursorLocation(index, true);```

and function getStyleAtPosition same change.

[...]

PR #4762, opened 2018-02-24
isEmptyStyles for textbox

asturur:
IsEmptyStyles receives unwrapped lineindex, since is used during rendering.
That means that textbox class requires its own version in order to work properly.

Needs an added test.

related to #4755


asturur:
Added textbox file for tests, since we had nothing.

PR #4763, opened 2018-02-24
Fixed incorrect Object.isOnScreen functionality.

scriptspry:
I have not found any usages of Object.isOnScreen being called with calculate = false. I thought about removing it but, was not sure. Hence, left it as it is, added a couple more tests, removed setCoords calls that were kind of bypassing the functionality being tested.

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -176,12 +176,12 @@
         }
       }
       // no points on screen, check intersection with absolute coordinates
-      if (this.intersectsWithRect(pointTL, pointBR, true)) {
+      if (this.intersectsWithRect(pointTL, pointBR, true, calculate)) {
         return true;
       }
       // worst case scenario the object is so big that contains the screen
       var centerPoint = { x: (pointTL.x + pointBR.x) / 2, y: (pointTL.y + pointBR.y) / 2 };
-      if (this.containsPoint(centerPoint, null, true)) {
+      if (this.containsPoint(centerPoint, null, true, calculate)) {
         return true;
       }
       return false;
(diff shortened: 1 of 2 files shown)


asturur:
We do not want to use calculate true during rendering, that will add unnecessary calculations.

scriptspry:
The travis build... it is running same tests as `npm test` ? I am seeing all tests pass in my system. Is there another test system for me to check before pushing?

As for the calculations during rendering, you're absolutely right, but these coords calculations, they are everywhere, can't we do something about these, performance wise?

I mean, something on lines of when canvas/object's properties that affect the coords are changed, calculate the coords. Probably by having setter's to object properties might help. But this is all another different issue I think.

Going back to thread, I really thought that this isOnScreen not calculating coords is a big problem. It is returning wrong results because of this reason. I think it should calculate and performance issue should be fixed in a different manner, immediately.

asturur:
So Fabricjs offers an api and some built in interactions.

The built in interactions, are used by user of a product, and are probably outside dev complete control, and in that case fabricjs takes care of updating the coordinates.

For all other situations, to move an object, you have to write some code, in that code you will call setCoords.
You can also override the _set method and for one of the many prop that influences coords, you can call setCoords automatically.

But should not be  a default behaviours. If you can make setCoords faster, even by just removing some add and multiply, please do it.

All the usage of trigonometric functions like sin cos atan and atan2, have been minimized recently because often they give back weird float results.

Also isOnScreen needs just the 4 main corners, while setCoords i think updates everything.

asturur:
I added the extra line so i can merge it, i wrote that comment but probably forgot to submit review

PR #4768, opened 2018-02-26
Use clientWidth and clientHeight of upperCanvas to size hiddenTextarea

nateevans:
Use clientWidth and clientHeight of upperCanvas to size hiddenTextarea to work with dynamic sized canvas.

See https://github.com/kangax/fabric.js/issues/4767

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -525,8 +525,8 @@
             y: boundaries.top + boundaries.topOffset + charHeight
           },
           upperCanvas = this.canvas.upperCanvasEl,
-          maxWidth = upperCanvas.width - charHeight,
-          maxHeight = upperCanvas.height - charHeight;
+          maxWidth = upperCanvas.clientWidth - charHeight,
+          maxHeight = upperCanvas.clientHeight - charHeight;
 
       p = fabric.util.transformPoint(p, m);
       p = fabric.util.transformPoint(p, this.canvas.viewportTransform);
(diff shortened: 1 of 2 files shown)


asturur:
@nateevans could you try if is possible to add a small test, if jsdom works good with it and then we can merge this?

nateevans:
ok @asturur I'll see what I can do for a test

nateevans:
@asturur Ok, I've been messing with this for awhile, and honestly I'm stumped. Not sure how to test this in a reasonable way outside of something like an integration test.

If you take a look at this updated Fiddle, you can see I was able to find and log the undesired width and height in the DOM when the iText is in edit mode. https://jsfiddle.net/8k4qogkc/31/ (This is without the fix)

But I am unsure of how to check for this within QUnit. I don't want to just jam some DOM elements in with styling... Though it seems like I would need the wrappers and styles from that Fiddle above to reproduce the conditions that cause the bug.

Thoughts?

nateevans:
Oh.... are you referring to usage of Fabric in Node? Doesn't look like we'll have access to clientHeight and clientWidth in jsdom?

Maybe something like this will work as a fallback in **_calcTextareaPosition**?

`
upperCanvasWidth = upperCanvas.clientWidth || upperCanvas.width;
upperCanvasHeight = upperCanvas.clientHeight || upperCanvas.height;
maxWidth = upperCanvasWidth - charHeight,
maxHeight = upperCanvasHeight - charHeight;
`

asturur:
share the test you drafted, i ll help to finish it. no worries

nateevans:
I think I got it covered now @asturur... let me know what you think on my latest commit.

I just exited the test if `isLikelyNode` since I am pretty sure that this does not apply to the Node environment.

nateevans:
@asturur ping, anything I can do to move this along? 😄

asturur:
I have been having many busy days, i got like 50 unopened emails about fabricJS. Trying to catch up.
I actually tried to write a test some days ago and then i fall asleep.

asturur:
i think i can merge it and then swap with a simpler test.

Issue #4507, opened 2017-11-27
Could you please provide TypeScript definition 


leehavin:
There are a lot of typescript projects now, hoping to provide a typescript definition

asturur:
someone should do it. i can't.
I do not know typescript i have no reason to learn it as of now.
If you know typescript please do it i will add the file where needed.
Someone should also then mantain it.

Is providing a typescript definition mandatory to make it work on typescript?

asturur:
https://stackoverflow.com/questions/27417107/how-use-an-external-non-typescript-library-from-typescript-without-d-ts

Is there anything that generates type definition from jsdoc?

CaptainCodeman:
I'm using https://www.npmjs.com/package/@types/fabric which seems to work OK

asturur:
is based on fabricjs 1.5

May lead to errors with new versions.

CaptainCodeman:
Thanks, yes I'm so far just using core features that haven't changed so it's working for me. It would be nice to have "official" versions but I'm guessing if someone is maintaining them they will be updated soon enough and not everyone is going to be using Typescript to make it an urgent priority.

asturur:
well the point is that javascript is still a losely typed langauge.
Is not an issue for me at all, i could produce a file that says that everything is `Any`

I think that who wants to use typescript should create one and mantain it as long as he want.

I do not feel like is a feature at all.
Maybe because i do not use typescript.

CaptainCodeman:
Typescript is a superset of Javascript with types added, while you can definitely use "any" for things, doing that doesn't really give you the true benefits (like type-safety and intellisense).

It's really easiest if things are coded in typescript but is still valuable for a lib with a large API to make it easier to consume. As I said, I can understand it not making sense for you to add or support it though.

asturur:
i do not have time really.
Also i think that in some points can be also problematic.
For example i like to use transformPoint passing a fabric.Point or a simple Object.
I wonder if that is still possible with a type definition.

CaptainCodeman:
Yes, you can define a signature as accepting multiple types, e.g.

`function doSomething(whatever : fabric.Point | fabric.Object)`

asturur:
put help wanted label, whoever wants to do it will do it.

@CaptainCodeman @leehavin  since you are typescript users, you are good candidates.

rebendajirijr:
Still no update at DefinitelyTyped/DefinitelyTyped#19911.

[...]

PR #4793, opened 2018-03-07
initial commit

JPAhnen:
As requested in #4507.
This shouldn't be merged yet as it isn't completed yet.


asturur:
are the jsdoc necessary for the typescript?
i mean if the type script supporting ide, read the inline documentation from the ts file instead of source file.
In this case keeping update 2 copies of docs may be bothersome.

JPAhnen:
I can take them out once the definition is done, but while I'm writing it, it's easier for me leaving it in.

PeterBenko:
FWIW docs don't seem to land in the distributed fabric package (from npm). But naturally, documenting in two places shouldn't be necessary.

asturur:
Docs are on the website, with a small banner.
The banner money are used to print stickers  and ship them.

i did one shipping and no sticker was ever delivered.

asturur:
Is this thing testable in a way that we can be sure it matches fabricJS, what tools are used?

asturur:
would it be possible to move this to a separate repository? like fabricjs/typescript?
Would it be still viable for people to use it from different repo?

I would imagine they would need to install both the typescript definition and the library.

How does typescript normally works? Do file need to be in the same folder as source files?

JPAhnen:
Usually, there are two packages where the definitions should be put: Either the main package itself (i.e. the fabric package), or the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo (where already is a fabric type definition by someone else - but I'm not sure for which fabric version)

asturur:
googled 3 jsdoc to d.ts file converter.
I ll try them soon.

In this way keeping jsdoc updated will be more important and we will get definition for free each release.

Issue #4808, opened 2018-03-13
Font names with trailing numbers don't work


dennisrjohn:
## Steps to reproduce
We have a font named "Teenage Girl 2"
When the font is drawn to the context using drawText, the font declaration that is generated looks like:

"30px Teenage Girl 2"

The browser is seeing the 2 at the end as a separate style, and the context doesn't find the font.

## Expected Behavior
The font draws properly

## Actual Behavior
The font isn't used and the text is drawn using the browser default font

I've fixed this by always forcing the fontFamily to be surrounded by quotes in _getFontDeclaration of text.class.js:
```
 var fontFamily = style.fontFamily === undefined ||
          style.fontFamily.includes('\'') ||
          style.fontFamily.includes('"')
        ? style.fontFamily : '"' + style.fontFamily + '"';
      return [
        // node-canvas needs "weight style", while browsers need "style weight"
        (fabric.isLikelyNode ? style.fontWeight : style.fontStyle),
        (fabric.isLikelyNode ? style.fontStyle : style.fontWeight),
        forMeasuring ? this.CACHE_FONT_SIZE + 'px' : style.fontSize + 'px',
        fontFamily
      ].join(' ');
```
Interestingly, the fontFamily for Textbox was already surrounded by quotes, but the fontFamily for any styles in the Textbox weren't, thus, the check to see if there is already a quote around the fontName.

Will attach PR.

dennisrjohn:
https://github.com/kangax/fabric.js/pull/4809

asturur:
I always skipped this because i have noticed that there are problems with safari or IE11 when using quotes, depending on the quote used ( double or single ).
What are your experiences?

dennisrjohn:
I just tested on IE11 and single quotes work fine.

asturur:
@dennisrjohn Can you check my comment in the PR?

dennisrjohn:
Yes. I’ll take a look at it tomorrow. Sorry, I’ve been working on something
else this week.

dennisrjohn:
Done.

PR #4811, opened 2018-03-13
Fix for #4808 

dennisrjohn:
This fixes issue #4808


asturur:
We need the change to the text.class.js file only, and if possible an added test that shows that fontnames with quotes do not get a double quoting.

dennisrjohn:
Closing and creating a new PR with your suggestions

PR #4831, opened 2018-03-16
New PR fo 4808

### LINKED ISSUE #4808: Font names with trailing numbers don't work

dennisrjohn:
Fixes #4808

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1203,12 +1203,16 @@
      */
     _getFontDeclaration: function(styleObject, forMeasuring) {
       var style = styleObject || this;
+      var fontFamily = style.fontFamily === undefined ||
+      style.fontFamily.includes('\'') ||
+      style.fontFamily.includes('"')
+        ? style.fontFamily : '"' + style.fontFamily + '"';
       return [
         // node-canvas needs "weight style", while browsers need "style weight"
         (fabric.isLikelyNode ? style.fontWeight : style.fontStyle),
         (fabric.isLikelyNode ? style.fontStyle : style.fontWeight),
         forMeasuring ? this.CACHE_FONT_SIZE + 'px' : style.fontSize + 'px',
-        (fabric.isLikelyNode ? ('"' + style.fontFamily + '"') : style.fontFamily)
+        fontFamily
       ].join(' ');
     },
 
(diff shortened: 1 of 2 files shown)


src/shapes/text.class.js
```diff
@@ -1203,12 +1203,16 @@
      */
     _getFontDeclaration: function(styleObject, forMeasuring) {
       var style = styleObject || this;
+      var fontFamily = style.fontFamily === undefined ||
+      style.fontFamily.includes('\'') ||
+      style.fontFamily.includes('"')
```
asturur: sorry to be late with this comment.
includes is something forever supported by IE, just recently included in firefox and chrome as per new specs.
I think there is a string method that is supported by everyone and should be es5 compatible.
FabricJS is es5 and .includes is not.

dennisrjohn:
@asturur Right, we use an ES6 polyfill in webpack, so I didn't notice.

I changed the code to use indexOf instead of includes.

asturur:
Thank you!

PR #4840, opened 2018-03-19
Added single quoting to font names in toSVG

### LINKED ISSUE #4839: toSVG not quoting font names
## Version
2.2.2

Related to #4808, toSVG isn't single quoting font names.

Fixed and added test in attached PR.

dennisrjohn:
Fixes #4839

### DIFF
--- a/src/mixins/object.svg_export.js
+++ b/src/mixins/object.svg_export.js
@@ -9,8 +9,8 @@
     }
     else {
       var color = new fabric.Color(value),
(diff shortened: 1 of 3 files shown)


asturur:
i need to see svg specs, i bet is idifferent, but i need to check between single and double what they want.

dennisrjohn:
According to the specs, is should be quoted, but it doesn't specify single or double quotes:
https://www.w3.org/TR/2008/REC-CSS2-20080411/fonts.html#propdef-font-family

PR #4846, opened 2018-03-22
removed references to context when disposing

### LINKED ISSUE #4844: nodejs, memory leak with toDataURL
## Version
fabric: 2.2.2
node: v6.10.3

## Test Case
https://gist.github.com/peernohell/db5513b45e015a5ed7f2b4b860211413

## Steps to reproduce
copy the gist.
npm install fabric
node index.js --expose-gc

## Expected Behavior
providing as first object an empty object or an object with a small image should have the same memory footprint.

## Actual Behavior
If you execute the code as is you will see in the terminal that the RSS is growing (to ~1.2Go)
If you uncomment the lien `obj = {}` then the RSS grow to ~500Mo then go down.

here is what I have in my terminal,
With obj uncomment
```
❯ node --expose-gc index.js
START  103
IMG0  506
IMG1  506
IMG2  506
END  506
RSS  104
RSS  104
RSS  77
RSS  76
RSS  76
RSS  76
RSS  76
```

with obj with the image
```
❯ node --expose-gc index.js
START  105
END  125
IMG0  526
RSS  526
IMG1  888
IMG2  1285
RSS  1284
RSS  1284
RSS  1284
RSS  1283
```

I know that the RSS is not a good metric to detect leak, but on our production server, node take so much memory that it hit all the free memory and crash.

asturur:
close #4844

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1555,7 +1555,10 @@
       this.removeListeners();
       wrapper.removeChild(this.upperCanvasEl);
       wrapper.removeChild(this.lowerCanvasEl);
-      delete this.upperCanvasEl;
+      this.upperCanvasEl = null;
+      this.cacheCanvasEl = null;
+      this.contextCache = null;
+      this.contextTop = null;
       if (wrapper.parentNode) {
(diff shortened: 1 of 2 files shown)


asturur:
can be tested, needs a test.

PR #4852, opened 2018-03-26
Canvas SVG export

### LINKED ISSUE #4801: Inconsistency in Canvas Zoom and SVG Background behavior
It seems to me that the goal of toSVG() is to replicate the current view of the canvas, including its zoom level, and because of that, setZoom() impacts SVG viewBox in the output. 

The problem is, if setBackgroundcolor() is used, the background <rect> element stays the same size, regardless of the zoom level, so for cases that zoom is less than 1, the SVG output looks different that canvas view (background is small):
http://jsfiddle.net/Da7SP/1127/

Canvas view:
image

SVG View:
image

A related issue is, if user sets a custom viewBox for SVG output, the background color won't be extended all the way to the edge of the viewBox.

If confirmed as a bug, I think the easiest fix is to change width and height of background <rect> to 100%, so both of the above use cases will be covered. 

`<rect x="0" y="0" width="100%" height="100%" fill="red"></rect>`

asturur:
fix #4801 
Export wider background when canvas is zoomed


semiadam:
@asturur  thanks so much for working on this.

PR #4856, opened 2018-03-26
isPartiallyOnScreen method

keyurpatel:
Close https://github.com/kangax/fabric.js/issues/4823,

Issue:  to know if the object is partially off screen, completely off screen.

Solution: create a simple helper methods which return boolean for object partially off screen `isPartiallyOnScreen`, completely offscreen `isOffScreen`.

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -189,7 +189,6 @@
 
     /**
      * Checks if object is partially contained within the canvas with current viewportTransform
-     * the check is done stopping at first point that appears off screen
      * @param {Boolean} [calculate] use coordinates of current position instead of .oCoords
      * @return {Boolean} true if object is partially contained within canvas
      */
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -199,7 +199,7 @@
       }
       return false;
     },
-    
+
     /**
      * Checks if object is partially contained within the canvas with current viewportTransform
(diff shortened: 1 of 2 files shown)


src/mixins/object_geometry.mixin.js
```diff
@@ -187,6 +187,35 @@
       return false;
     },
 
+    /**
+     * Checks if object is partially contained within the canvas with current viewportTransform
+     * the check is done stopping at first point that appears off screen
+     * @param {Boolean} [calculate] use coordinates of current position instead of .oCoords
```
asturur: use coordinates of current position, calculated just in time, instead of precalculated .aCoords

asturur: apply to the other function too.

test/unit/object_geometry.js
```diff
@@ -802,4 +802,38 @@
...
+    assert.equal(cObj.isPartiallyOnScreen(cObj.aCoords), true, 'object is partially on screen because it include the canvas');
+    cObj.left = 100;
+    cObj.top = 100;
+    cObj.setCoords();
+    assert.equal(cObj.isPartiallyOnScreen(cObj.aCoords), false, 'object is not partially on screen because it include the canvas');
+    cObj.left = -1000;
+    cObj.top = -1000;
+    cObj.setCoords();
+    assert.equal(cObj.isPartiallyOnScreen(cObj.aCoords), false, 'object is not partially on screen because it is outside the canvas');
+  });
```
asturur: we have more test to cover and edge cases i believe. Intersect can return true for coincidents segments too.

PR #4866, opened 2018-03-29
Fix getPointer on touch devices

nmorel:
When dragging an object on touch device on top or left edge of a screen, the getPointer function returns an incorrect value (NaN).  
To reproduce it, you can use the following demo : http://fabricjs.com/intersection and move an object to the left side. 

When clientX and pageX are 0, the internal _getPointer method returns event[clientProp] which is undefined instead of 0.

### DIFF
--- a/src/util/dom_event.js
+++ b/src/util/dom_event.js
@@ -212,11 +212,17 @@
 
   function _getPointer(event, pageProp, clientProp) {
     var touchProp = event.type === 'touchend' ? 'changedTouches' : 'touches';
+    var pointer;
 
-    return (event[touchProp] && event[touchProp][0]
-      ? (event[touchProp][0][pageProp] - (event[touchProp][0][pageProp] - event[touchProp][0][clientProp]))
-        || event[clientProp]
-      : event[clientProp]);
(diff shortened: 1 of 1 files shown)


src/util/dom_event.js
```diff
@@ -212,11 +212,17 @@
...
 
-    return (event[touchProp] && event[touchProp][0]
-      ? (event[touchProp][0][pageProp] - (event[touchProp][0][pageProp] - event[touchProp][0][clientProp]))
-        || event[clientProp]
-      : event[clientProp]);
+    if (event[touchProp] && event[touchProp][0]) {
+      pointer = event[touchProp][0][pageProp] - (event[touchProp][0][pageProp] - event[touchProp][0][clientProp]);
+    }
+
+    if (!pointer && pointer !== 0) {
```
asturur: isn't this better as typeof pointer === 'undefined' ?  that is the real meaning of it, if yuo did manage to assign anything to pointer in the first if, right?

nmorel: yea it should work the same. I'll change it.

PR #4877, opened 2018-03-31
Reset width and height for an image when swapping src

### LINKED ISSUE #4733: Image setSrc, setElement and width/height value
## Version
2.0.1

## Test Case

Tried to load the same image with exact same code and got different outputs.

v1.7.22 - [Code](https://jsfiddle.net/10t7b7kh/) - [Output](https://user-images.githubusercontent.com/19214202/36344091-37b49af4-143b-11e8-946d-24c8631facf7.png)

v2.0.1 - [Code](https://jsfiddle.net/huby9uuy/) - [Output](https://user-images.githubusercontent.com/19214202/36344093-3fd8a89c-143b-11e8-849a-9e53470c030a.png)

## Expected Behavior

Expected the full image to be rendered.

## Actual Behavior

Only got 1/4th? of the image rendered.

---

First off, I do have a "working" solution for even v2.0.1 [here](https://jsfiddle.net/a27dozgb/). But, it is a hack. Also, I have already gone through the v1 to v2 breaking changes documents, and read and understood the parts about fabric.Image and fabric.Image.fromObject.

Now, I have already spent a couple of hours of time on this, and came to a conclusion that library is misbehaving. Here's why:

1. width = 128 & scaleX = 1 !== width = 256 & scaleX = 0.5 !!!

If the width and scaleX semantics for other shapes: rect, circle etc are rendering same output for both cases. Why does Image behave differently? Is it not an inconsistency with interpretation of what width and scaleX values are?

2. We can almost never have scaleX as 1 for fabric.Image instances, except when image's width, height match fabric.Image object's width height.

This feels like a painful and limiting enforcement.

asturur:
close #4733 

needs tests

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -369,6 +369,7 @@
     setSrc: function(src, callback, options) {
       fabric.util.loadImage(src, function(img) {
         this.setElement(img, options);
+        this._setWidthHeight();
         callback(this);
       }, this, options && options.crossOrigin);
       return this;
@@ -555,13 +556,13 @@
      * @param {Object} [options] Object with width/height properties
      */
     _setWidthHeight: function(options) {
-      this.width = 'width' in options
+      this.width = options && ('width' in options)
         ? options.width
         : (this.getElement()
           ? this.getElement().width || 0
           : 0);
 
-      this.height = 'height' in options
+      this.height = options && ('height' in options)
         ? options.height
         : (this.getElement()
           ? this.getElement().height || 0
--- a/test/unit/image.js
+++ b/test/unit/image.js
@@ -134,6 +134,22 @@
     });
   });
 
+  QUnit.test('setSrc', function(assert) {
+    var done = assert.async();
+    createImageObject(function(image) {
+      image.widh = 100;
+      image.height = 100;
+      assert.ok(typeof image.setSrc === 'function');
+      assert.equal(image.width, 100);
+      assert.equal(image.width, 100);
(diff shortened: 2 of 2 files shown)


src/shapes/image.class.js
```diff
@@ -369,6 +369,7 @@
     setSrc: function(src, callback, options) {
       fabric.util.loadImage(src, function(img) {
         this.setElement(img, options);
+        this._setWidthHeight();
```
scriptspry: Umm... did you forget to pass options to `_setWidthHeight` or is this intentional? Just wondering.

sandro: @scriptspry Great question. I just tried using `setSrc` with `width` and `height` specified in options, and the provided options were not set. I ended up needing to use `loadImage` directly, which shouldn't be necessary. I vote for options to be passed in to `this._setWidthHeight()` or better, don't call `_setWidthHeight` at all and let `setElement` take care of it.

asturur: There was some reason this was done.
I'll try to recover it.
It could be that setSrc will intentionally reset the image from scratch, and if you need to set it with different widht/height/scropX/cropY you do that in the callback.
What would be wrong with this approach?

sandro: @asturur Hi, nothing wrong with the approach of changing the width and height in the callback, it's just when I looked at the fabric source code I thought for sure not passing options along was a bug. I almost created a PR to fix it, then I noticed this comment and wanted to follow up.

PR #4897, opened 2018-04-10
Fixing issues with selectionDashArray bleeding into free drawing (#4894)

### LINKED ISSUE #4894: selectionDashArray causing issues when free drawing
## Version
2.2.3

## Test Case
http://jsfiddle.net/Da7SP/1459/

## Steps to reproduce
1. initialize canvas with selectionDashArray
2. click and drag to show that selectionDashArray exists
3. set canvas isDrawingMode to true
4. free draw on the canvas

## Expected Behavior
While drawing, the line should appear solid

## Actual Behavior
While drawing, the line appears as dashed, but on mouseUp returns to a solid line

arlen-yu:
close #4894

Proposed fix: call `setLineDash` on an empty dash array once the selection box is done drawing.

### DIFF
--- a/src/brushes/base_brush.class.js
+++ b/src/brushes/base_brush.class.js
@@ -78,8 +78,8 @@ fabric.BaseBrush = fabric.util.createClass(/** @lends fabric.BaseBrush.prototype
     ctx.lineCap = this.strokeLineCap;
     ctx.miterLimit = this.strokeMiterLimit;
     ctx.lineJoin = this.strokeLineJoin;
-    if (this.strokeDashArray && fabric.StaticCanvas.supports('setLineDash')) {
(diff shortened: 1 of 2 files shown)


asturur:
While this fix the issue probably, i think the best way to fix it is to ensure the brush is calling setDashArray by itself on the prepareDrawing ( or similarly named ) method.

asturur:
Well... i see that _prepareForDrawing is calling _reset, that is calling _setBrushStyle that is resetting the brush strokeDashArry in the file base_brush.class.js

Sadly there is an extra if that is setting the strokeDashArray just if is different than null that is the default value.

Is at like 74 the function, can you fix it there?

asturur:
Thanks!

PR #4926, opened 2018-04-26
Added Canvas.onBeforeMove

scriptspry:
Hi,

When implementing undo functionality for an object, we can know the state before scale and rotate using the existing Canvas.onBeforeScaleRotate callback.

But there's no such thing for movement. I needed to know the state of the object before it's going to move.

Hence, this pull request. Added `Canvas.onBeforeMove` method that gets before the component is going to be moved.

I have a small problem though: it also triggers when mouse down happens on the object. I couldn't find any way to distinguish these both and thought about renaming onBeforeMove to `onLikelyToMove` but didn't feel it is clean solution. Do you have any thoughts about this?

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -601,6 +601,9 @@
       if (target._findTargetCorner(this.getPointer(e, true))) {
         this.onBeforeScaleRotate(target);
       }
+      else {
+        this.onBeforeMove(target);
+      }
 
     },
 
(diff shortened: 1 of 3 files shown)


asturur:
I think this thing should be normalized and we could use input of more people.
What about starting a short conversation in the api-discussion issue tracker?

asturur:
i think with the new events this should be setted.
What do you think?

scriptspry:
Yes. I've checked the new events just a couple of days ago. They seem very powerful and much cleaner than this. Closing. 👍

PR #4949, opened 2018-05-04
isEqual fails hard.

keyurpatel:
Issue: - https://github.com/fabricjs/fabric.js/issues/4950

**small description.**
stateful mixin method `isEqual` throw an exception when it tries to compare an array value with null.


keyurpatel:
@asturur Thanks for adding a test. Sure will take a look to fix the object case.

keyurpatel:
Everything passed locally !!!
screen shot 2018-05-07 at 1 10 03 pm

asturur:
So why it was going locally no idea. The test was actually catching a bug, calling null is of type Object and Object.keys(null) fails.

PR #4954, opened 2018-05-07
Fix: use the same options when adding or removing touch event listeners

tehmaestro:
Backport https://github.com/fabricjs/fabric.js/pull/4804/commits/a710b12f0d9a2d25b1e69c1bf55b905eb92aa02a


tehmaestro:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

I do not know for sure, but I could test.

asturur:
thanks for finding the link, yes is mandatory. So i have to fix the 2.x master before release.

tehmaestro:
Yes, I was about the warn you guys that this needs to be fixed in 2.x too. Thanks :)

PR #4955, opened 2018-05-07
isTargetTransparent() now samples the cache directly.

AndrewJDR:
This is a work in progress. It seems to work for the limited things I've tried it with -- mostly hit-testing paths. I know that it doesn't work on apple platforms where retina scaling is enabled, so I'd like some advice on how to get that working properly, and any other feedback/notes.

For example, I check for cache validity with the following:
```
if (target.shouldCache && target._cacheCanvas && !target.isCacheDirty())
```
Is there a better way to be doing this, or other things to consider checking about the state of the cache?

And finally, does the math here correct?

Thank you!

---

Previously, isTargetTransparent would do a paint from the cache to the contextCache which has more overhead and is slower. With this commit, fabricjs will test the cache canvas directly using x/y coordinates calculated with respect to the target object center. This is more performant since it avoids an extra render
on every call to isTargetTransparent.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -492,7 +492,7 @@
      * @return {Boolean}
      */
     isTargetTransparent: function (target, x, y) {
-      if (target.shouldCache() && target._cacheCanvas && !target.isCacheDirty()) {
+      if (target.shouldCache() && target._cacheCanvas) {
         var normalizedPointer = this._normalizePointer(target, {x: x, y: y}),
             targetRelativeX = target.cacheTranslationX + (normalizedPointer.x * target.zoomX),
             targetRelativeY = target.cacheTranslationY + (normalizedPointer.y * target.zoomY);
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -492,6 +492,17 @@
      * @return {Boolean}
      */
     isTargetTransparent: function (target, x, y) {
+      if (target.shouldCache() && target._cacheCanvas && !target.isCacheDirty()) {
+        var normalizedPointer = this._normalizePointer(target, {x: x, y: y}),
+            targetRelativeX = target.cacheTranslationX + (normalizedPointer.x * target.zoomX),
+            targetRelativeY = target.cacheTranslationY + (normalizedPointer.y * target.zoomY);
+
+        var isTransparent = fabric.util.isTransparent(
+          target._cacheContext, targetRelativeX, targetRelativeY, this.targetFindTolerance);
+
+        return isTransparent;
+      }
+
       var ctx = this.contextCache,
           originalColor = target.selectionBackgroundColor, v = this.viewportTransform;
 


src/canvas.class.js
```diff
@@ -492,6 +492,20 @@
      * @return {Boolean}
      */
     isTargetTransparent: function (target, x, y) {
+      if (target.shouldCache && target._cacheCanvas && !target.isCacheDirty()) {
```
asturur: the logic that tells us if the target is on its own cache canvas is a little bit more complex ( i have to verify )

src/canvas.class.js
```diff
@@ -492,6 +492,20 @@
...
+        var center = target.getCenterPoint(),
+            curRelTargetX = x - center.x,
+            curRelTargetY = y - center.y;
+
+        curRelTargetX = (target._cacheCanvas.width / 2) + (curRelTargetX / target.zoomX);
+        curRelTargetY = (target._cacheCanvas.height / 2) + (curRelTargetY / target.zoomY);
```
asturur: for blurriness reasons more than width/2 there is another property where we can see the center of the obect on the canvas, i think is called cacheTranslateX/Y or something like that, you find it in UpdateCacheCanvas function.

Also the translation respect the center should take care of the object rotation using the matrices directly.

AndrewJDR: Okay, I tried using cacheTranslateX/Y in various ways, but I was getting weird results. I'll give it another try though.

src/canvas.class.js
AndrewJDR: Ah ok. I've now noticed a canvas/path that this code doesn't properly handle, so I'm wondering if it's due to not checking its cache canvas properly or something.

src/canvas.class.js
```diff
@@ -492,6 +492,17 @@
      * @return {Boolean}
      */
     isTargetTransparent: function (target, x, y) {
+      if (target.shouldCache() && target._cacheCanvas && !target.isCacheDirty()) {
+        var normalizedPointer = this._normalizePointer(target, {x:x, y:y}),
```
AndrewJDR: Changed to use _normalizePointer() here, which uses the xform matrix so it handles rotation etc.

src/canvas.class.js
```diff
@@ -492,6 +492,17 @@
      * @return {Boolean}
      */
     isTargetTransparent: function (target, x, y) {
+      if (target.shouldCache() && target._cacheCanvas && !target.isCacheDirty()) {
```
AndrewJDR: shouldCache() returns the result of `this.objectCaching && (!this.group || this.needsItsOwnCache() || !this.group.isOnACache())`. Does this seem sufficient?

AndrewJDR: shouldCache() returns the result of `this.objectCaching && (!this.group || this.needsItsOwnCache() || !this.group.isOnACache())`. Does this seem sufficient?

src/canvas.class.js
```diff
@@ -492,6 +492,17 @@
...
+        var normalizedPointer = this._normalizePointer(target, {x: x, y: y}),
```
AndrewJDR: Changed to use _normalizePointer() here, which uses the xform matrix so it handles rotation etc.

AndrewJDR:
Updated, accounted for rotations and it also seems to work with retina scaling enabled, now.

asturur:
I m a little bit busy and i have to wrap up 2 things and release 2.2.4, but i l check this next.

asturur:
it looks good anyway

asturur:
i think is good, i m gonna merge it.
Writing a test is not straight forward, let's keep it as it is.

Issue #4946, opened 2018-05-04
Error - Access of includeDefaultValues when object has no group property


iongion:
For a certain reason, some of my objects/selection have no `group` property but the code hits this path that triggers an exception

## Version
latest

## Test Case
This code:
```
        toObject: function(propertiesToInclude) {
            var objsToObject = this.getObjects().map(function(obj) {
                var originalDefaults = obj.includeDefaultValues;
                obj.includeDefaultValues = obj.group.includeDefaultValues;
                var _obj = obj.toObject(propertiesToInclude);
                obj.includeDefaultValues = originalDefaults;
                return _obj;
            });
            return extend(this.callSuper("toObject", propertiesToInclude), {
                objects: objsToObject
            });
        },
```

## Steps to reproduce
Trigger code hit of this code when `obj` has no `group` property

## Expected Behavior
```
obj.includeDefaultValues = obj.group ? obj.group.includeDefaultValues : false;
```

## Actual Behavior
It is hard to provide a fiddle with this.

asturur:
in some way you are creating a broken selection, but i agree this pattern should be avoided and safeguarded.

Issue #4598, opened 2018-01-01
loadFromDatalessJSON: load multiple svg file as object from saved design error


SanjayChaudhary02:
@kangax While i am trying to load dataless json from saved design , which contain multipath svg files as object it returns error on:fabric.require.js:441: 
`objects.forEach(function(o, index) {`

while i am saving datalessjson by adding `obj.sourcePath = obj.origingSVG(oringinal path of svg)`, 
Multipath svg file contain objects key in objects data so after first iteration value of objects replace by url and foreach stop. 

- issue only comes while i am loading multipath svg
Load svg method:
- Initially while we add svg from url to canvas
` fabric.loadSVGFromURL(src, function(objects, options) {
                                
             var img = fabric.util.groupSVGElements(objects, options);
             img.set('sourcePath', src);
                                current.options.currentCanvas.add(img).renderAll();
                                  img.set({
                                      left: 0,
                                      top: 0,
                                      objID: newID,
                                      objType : targetArea,
                                      uploadBaseType: 'svg',
                                      origingSVG: src,
                                      originX:'center',
                                      originY:'center',
                                      selectable: false,
                                      scaleX : current.options.canvasWidth / img.width, 
                                      scaleY : current.options.canvasHeight / img.height, 
                                      applyFill: 0,
                                  });
                                current.options.currentObject.center();
                                current.options.currentObject.setCoords();
                                current.options.currentCanvas.renderAll();
                                current.options.currentCanvas.setActiveObject(current.options.currentObject);
                              },null,true);`

Output  :: 
Datalessjson:
[...]

PR #4994, opened 2018-05-23
Fix issue with saving/restoring datalessJSON while retaining the edit…

brentcappello:
In reference to #4598 and #4946

This was really a two part issue. 

Part 1. was the objects.forEach bug. (objects.forEach would bug out when it received a url string)
Fix: in toDatalessObject: I removed the if/else assigning the sourcePath to the objects prop. Instead I figured it made more sense to keep all the objects default values minus the paths. This way we can retain the relatively small properties while still achieving the goal of dumping the large paths for the small sourcePath url.

Part 2. was rebuilding the svg/group. (Svg group options should rebuild to how they were exported) 
Fix: I leveraged the implementation used in Paths and the previous PathGroup(pre 2.0) and modified the fabric.Group.fromObject to check for a sourcePath. If the sourcePath was found load the SVG from the url and set each object element to the object “elements” original value. I used ._setObject for clarity but .set works just as well. I then grouped the elements using groupSVGElements and set the options to the original object. 

Hope that makes sense!

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -248,18 +248,16 @@
      */
     toDatalessObject: function(propertiesToInclude) {
       var objsToObject, sourcePath = this.sourcePath;
-      if (sourcePath) {
-        objsToObject = sourcePath;
-      }
-      else {
-        objsToObject = this.getObjects().map(function(obj) {
-          var originalDefaults = obj.includeDefaultValues;
-          obj.includeDefaultValues = obj.group.includeDefaultValues;
-          var _obj = obj.toDatalessObject(propertiesToInclude);
-          obj.includeDefaultValues = originalDefaults;
-          return _obj;
-        });
-      }
+      objsToObject = this.getObjects().map(function(obj) {
+        var originalDefaults = obj.includeDefaultValues;
+        obj.includeDefaultValues = obj.group.includeDefaultValues;
+        var _obj = obj.toDatalessObject(propertiesToInclude);
+        obj.includeDefaultValues = originalDefaults;
+        if (sourcePath) {
+          delete _obj.path;
+        }
+        return _obj;
+      });
       return extend(this.callSuper('toDatalessObject', propertiesToInclude), {
         objects: objsToObject
       });
@@ -558,12 +556,26 @@
    * @param {Object} object Object to create a group from
    * @param {Function} [callback] Callback to invoke when an group instance is created
    */
-  fabric.Group.fromObject = function(object, callback) {
-    fabric.util.enlivenObjects(object.objects, function(enlivenedObjects) {
-      var options = fabric.util.object.clone(object, true);
-      delete options.objects;
-      callback && callback(new fabric.Group(enlivenedObjects, options, true));
-    });
-  };
+fabric.Group.fromObject = function(object, callback) {
+    var options = fabric.util.object.clone(object, true);
 
+    if (typeof object.sourcePath === "string") {
+      var sourcePath = object.sourcePath;
+      fabric.loadSVGFromURL(sourcePath, function(elements) {
+        elements.forEach(function (obj, i) {
+          obj._setObject(object.objects[i]);
+        });
+        var group = fabric.util.groupSVGElements(elements);
+        group.setOptions(options);
+        callback && callback(group);
+      });
+    } 
+    else {
+      fabric.util.enlivenObjects(object.objects, function(enlivenedObjects) {
+        delete options.objects;
+        callback && callback(new fabric.Group(enlivenedObjects, options, true));
+      });
+    }
+  };
+  
 })(typeof exports !== 'undefined' ? exports : this);
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -556,10 +556,10 @@
    * @param {Object} object Object to create a group from
    * @param {Function} [callback] Callback to invoke when an group instance is created
    */
-fabric.Group.fromObject = function(object, callback) {
+  fabric.Group.fromObject = function(object, callback) {
     var options = fabric.util.object.clone(object, true);
 
-    if (typeof object.sourcePath === "string") {
+    if (typeof object.sourcePath === 'string') {
       var sourcePath = object.sourcePath;
       fabric.loadSVGFromURL(sourcePath, function(elements) {
         elements.forEach(function (obj, i) {
@@ -569,7 +569,7 @@ fabric.Group.fromObject = function(object, callback) {
         group.setOptions(options);
         callback && callback(group);
       });
-    } 
+    }
     else {
       fabric.util.enlivenObjects(object.objects, function(enlivenedObjects) {
         delete options.objects;
@@ -577,5 +577,5 @@ fabric.Group.fromObject = function(object, callback) {
       });
     }
   };
-  
+
 })(typeof exports !== 'undefined' ? exports : this);


asturur:
Hwey i missed this,
tests are failing can you check why?
I still did not read it fully, and i have to remember how this dataless should work.
I think that maybe is too complicated and we should have a more imperative way to do this.

Saving an object with a sourcePath is a bit wrong since you may have modified values in the group that you will not find back restoring. Wondering if it makes sense at all to keep this feature.

brentcappello:
Sorry about that. I fixed the linter formatting errors. The failing unit test is the assertion that if a sourcePath is present on a Group toDatalessObject that the object.objects should be the sourcePath URL. This is precisely what I changed with the implementation edits I made. To your comments, Yes I agree that saving with source path did not work how I think people would have expected it. Myself included. I modified it so the svg path property options (fills, strokes, etc) for each path would be saved if you made edits and then restored when you loaded it back. This way only the paths points (the bulk of the SVG) are using the sourcePath URL. For most implementations I think this makes sense. I do know that it drastically lowers our heap size, storage and processing times. It just didn't make sense not being able to restore the edits. But, now we can. I do think it's a useful addition to the core. The only requirement that might be different then before is sourcePath must be included in the custom props definition. toDatalessJSON(['sourcePath']); If I missed anything or if additional edits are required let me know! Thanks.

brentcappello:
@asturur Would you like me to write a test that asserts the sourcePath is still valid? I can simply adjust the the 1 failing test to check for the sourcePath inside the object.objects instead of the current test which expects object.objects to equal the sourcePath.

asturur:
i want definitely fix this issue, i have been buys finishing the svg import fix and did not have time to look into this.
Please do fix the current failing test, and let's move this on in some way.
I have to be honest i did not stop reflecting on side effects of changes.
I m wondering is someone is using toDataless in production environment and we are going to blow up some apps.

brentcappello:
Yes, I completely agree. I will write some additional tests to verify!

brentcappello:
@asturur Sorry or the delay, I have been working on multiple variations of this problem. I was hoping to uncover a more "elegant" solution. But really I ended up running into various road blocks of one form or another. Here is what I attempted. I am open to suggestions. Otherwise, I am going to clean up the existing PR and write a few tests around options numner 1.  

## These are the four proposed recommendations:
1. if/else SourcePath in Group.fromObject (still the most performant)
2. Use the current enlivenObject utility and id deep group paths (An interesting idea but slow)
3. Created redundant groupings (not recommended unless we can fix the overhead)
4. Possibly Create a new Class specific to handeling multiple paths that share the same sourceUrl (Similar to how PathGroup was, but as a MUCH smaller utility possibly)

### 1. Use the Group class and fromObject method to wrap an if/else on discovery of an object level sourcePath string. 
 **_group.class.js - line 560_**

```
fabric.Group.fromObject = function(object, callback) {
  var options = fabric.util.object.clone(object, true);

  if (typeof object.sourcePath === 'string') {
    var sourcePath = object.sourcePath;
    fabric.loadSVGFromURL(sourcePath, function(elements) {
      elements.forEach(function (obj, i) {
        obj._setObject(object.objects[i]);
      });
      var group = fabric.util.groupSVGElements(elements);
      group.setOptions(options);
      callback && callback(group);
    });
  }
  else {
    fabric.util.enlivenObjects(object.objects, function(enlivenedObjects) {
      delete options.objects;
      callback && callback(new fabric.Group(enlivenedObjects, options, true));
    });
  }
};
```

Delete the path prop if sourcePath.
**_group.class.js - line 240_**
```
toDatalessObject: function(propertiesToInclude) {
  var objsToObject, sourcePath = this.sourcePath

  objsToObject = this.getObjects().map(function(obj, index) {
    var originalDefaults = obj.includeDefaultValues;
    obj.includeDefaultValues = obj.group.includeDefaultValues;
    var _obj = obj.toDatalessObject(propertiesToInclude);
    obj.includeDefaultValues = originalDefaults;
    if (sourcePath) {
      delete _obj.path;
    }
    return _obj;
  });

  return extend(this.callSuper('toDatalessObject', propertiesToInclude), {
    objects: objsToObject,
    sourcePath: sourcePath
  });

}
```

### 2. Use the current enlivenObject Utility to parse an piece together "grouped paths"
**_group.class.js - line 560_**
```
fabric.Group.fromObject = function(object, callback) {
  fabric.util.enlivenObjects(object.objects, function(enlivenedObjects) {
    var options = fabric.util.object.clone(object, true);
    delete options.objects;
    callback && callback(new fabric.Group(enlivenedObjects, options, true));
  });
};
```

I needed to be able to find the correct path to apply the object options to. I added a pathPosition property and sourcePath to each path when the object was created and then use the Group.fromGroup enlivenObject utility to route the path to the Path class fromObject where it finds the correct path and a parses it. (This was inneficent and slow due to the necessary parser loops).
**_path.class.js - line 884_**
```
fabric.Path.fromObject = function(object, callback) {
  if (typeof object.sourcePath === 'string') {
    var path, pathUrl = object.sourcePath
    var pos = object.pathPosition || 0
    fabric.loadSVGFromURL(pathUrl, function(elements) {
        path = elements[pos];
        path.setOptions(object);
        callback && callback(path);
    });
  }
  else {
    fabric.Object._fromObject('Path', object, callback, 'path');
  }
};
```

Notice the addition of the pathPositon Prop.
**_group.class.js - line 240_**
```
toDatalessObject: function(propertiesToInclude) {
  var objsToObject, sourcePath = this.sourcePath

  objsToObject = this.getObjects().map(function(obj, index) {
    var originalDefaults = obj.includeDefaultValues;
    obj.includeDefaultValues = obj.group.includeDefaultValues;
    var _obj = obj.toDatalessObject(propertiesToInclude);
    obj.includeDefaultValues = originalDefaults;
    if (sourcePath) {
      _obj.sourcePath = sourcePath;
      _obj.pathPosition = index;
      delete _obj.path;
    }
    return _obj;
  });

  return extend(this.callSuper('toDatalessObject', propertiesToInclude), {
    objects: objsToObject,
    sourcePath: sourcePath
  });

},
```

### 3. Use the current enlivenObject Utility to parse an piece together "grouped paths"
**_path.class.js - line 884_**

```
fabric.Path.fromObject = function(object, callback) {
  if (typeof object.sourcePath === 'string') {
    var pathUrl = object.sourcePath;

    fabric.loadSVGFromURL(pathUrl, function (elements) {
      var path;
      if (elements.length > 1) {
        path = elements[object.pathId];
      } else {
        path = elements[0]
      }
      path.setOptions(object);
      callback && callback(path);
    });
  }
  else {
    fabric.Object._fromObject('Path', object, callback, 'path');
  }
};
```

We can add the property identifier in the Group fromObject (like below) or in the toDatalessObject method like mentioned above in section 2.
**_group.class.js - line 560_**
```
fabric.Group.fromObject = function(object, callback) {
  object.objects.forEach(function (obj, i) {
    obj.pathId = i;
  });
  fabric.util.enlivenObjects(object.objects, function(enlivenedObjects) {
    var options = fabric.util.object.clone(object, true);
    delete options.objects;
    callback && callback(new fabric.Group(enlivenedObjects, options, true));
  });
};
```
We can take advantage of the Path Class toDatalessObject method (line 458) to
delete the path for us. However we get a slight performance hit because of it.

```
toDatalessObject: function(propertiesToInclude) {
  var objsToObject, groupSourcePath = this.sourcePath;
  objsToObject = this.getObjects().map(function(obj) {
    var originalDefaults = obj.includeDefaultValues;
    obj.includeDefaultValues = obj.group.includeDefaultValues;
    if (!obj.sourcePath) {
      obj.sourcePath = groupSourcePath;
    }
    var _obj = obj.toDatalessObject(propertiesToInclude);
    obj.includeDefaultValues = originalDefaults;
    return _obj;
  });

  return extend(this.callSuper('toDatalessObject', propertiesToInclude), {
    objects: objsToObject
  });
},
```

### 4. Or we can write a new Class to handle multiple paths objects that share the same sourcePath while needing to retain the options for each path. Just an option.

Thanks!

PR #5017, opened 2018-06-01
reHSLa and reRGBa ignore case

durga598:
ignore case for rgba and hsla, as if pass in uppercase it was returning default color

### DIFF
--- a/test/unit/color.js
+++ b/test/unit/color.js
@@ -88,8 +88,8 @@
     var oColor = new fabric.Color('ffffffff');
     assert.ok(typeof oColor.toHexa === 'function');
     assert.equal(oColor.toHexa(), 'FFFFFFFF');
-    oColor.setSource([255,255,255,0.8]);
-    assert.equal(oColor.toHexa(), 'FFFFFFCC');
+    oColor.setSource([255,255,255,0.1]);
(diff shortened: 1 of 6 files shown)


asturur:
I think this is OK.

asturur:
Do you mind adding a small test to the suite to test it?

like at file test/unit/color.js
dduplicate the test at line 176 making it with uppercase and do the same for an hsla test.

durga598:
I added.

test/unit/color.js
```diff
@@ -88,8 +88,8 @@
     var oColor = new fabric.Color('ffffffff');
     assert.ok(typeof oColor.toHexa === 'function');
     assert.equal(oColor.toHexa(), 'FFFFFFFF');
-    oColor.setSource([255,255,255,0.8]);
-    assert.equal(oColor.toHexa(), 'FFFFFFCC');
+    oColor.setSource([255,255,255,0.1]);
+    assert.equal(oColor.toHexa(), 'FFFFFF1A');
```
asturur: can i ask you the reason of this change here? i remember this one was here because of weird decimals.

durga598: It was in my forked version.

durga598: changed that to original

PR #5020, opened 2018-06-03
remove more events

### LINKED ISSUE #5013: Fabricjs doesn't release mouse event handlers on dispose
## Version
2.3.0

## Test Case
http://jsfiddle.net/Da7SP/1660/

## Goal
Refresh the canvas object frequently, refreshing means to dispose() the current fabric canvas and then recreate it immediately.

## Steps to reproduce
- Open your chrome console (or any console)
- Click your mouse on the canvas(in it) and do no release it (mouse:down)
- Wait 1 second and now release and move your mouse around
- Watch as the console is getting filled with infinite errors any time you move the mouse anywhere on the webpage.

## Expected Behavior
Dispose() should clean everything related to fabric, allowing you to recreate a new fabric if you wish so, and should **always clean all event handlers and all data**

## Actual Behavior
Dispose() won't really clean fabric if a mouse event is already in progress (mouse down), causing a bug that will fail the browser until a full webpage refresh

## Final words
It seems like fabric is registering some GLOBAL(document) events on mouse:down and do not release them on dispose causing errors if you happen to call dispose() while the user clicked on something in the canvas

When I run `getEventListeners(document)` [on my webpage not on the fiddle, in the fiddle you would need to get the document of the html VM window not the fiddle window] I get:
`{click: Array(2)}`
Each time I recreate this bug and list the `getEventListeners(document)`  I get:
`{click: Array(2), touchend: Array(1), touchmove: Array(1), mouseup: Array(1), mousemove: Array(1)}`
And if I cause it again:
`{click: Array(2), touchend: Array(2), touchmove: Array(2), mouseup: Array(2), mousemove: Array(2)}`

So it seems like `touchend,touchmove,mouseup,mousemove` do not get cleaned if an event was mid-air while disposing.

asturur:
close #5013 

When disposing, remove also eventual event listener bound to document.window

Also made similar code common

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -122,6 +122,11 @@
       removeListener(this.upperCanvasEl, 'dragenter', this._onDragEnter);
       removeListener(this.upperCanvasEl, 'dragleave', this._onDragLeave);
       removeListener(this.upperCanvasEl, 'drop', this._onDrop);
+      // if you dispose on a mouseDown, before mouse up, you need to clean document to...
+      removeListener(fabric.document, 'mouseup', this._onMouseUp);
+      removeListener(fabric.document, 'touchend', this._onMouseUp, addEventOptions);
+      removeListener(fabric.document, 'mousemove', this._onMouseMove, addEventOptions);
+      removeListener(fabric.document, 'touchmove', this._onMouseMove, addEventOptions);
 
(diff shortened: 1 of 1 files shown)


asturur:
looks like is fixiing

asturur:
Also fix a bug of double click being registered with dblclick and tried to remove as doubleclick.

Issue #992, opened 2013-11-18
Free drawing mode does not support multitouch


kccheung:
When I test the free drawing functionality by navigating http://fabricjs.com/freedrawing/
on my mobile safari of iPad @ iOS6. It does not support multitouch drawing. Once I apply my second touch during first touch drawing, it immediately cancel the first touch drawing.
##

wikett:
Is there any update about freeDraw multitouch (3, 4 or 5 pens at the same time drawing?)

If not, could you tell me what it will be the best approach?

Thanks in advance

asturur:
Please describe better what you need.
Let's suppose for a moment that multitouch get supported.

How are you going to discern one finger/pen from the another?

wikett:
Thanks @asturur 
The idea is to do something similar like this: http://www.paulirish.com/demo/multi

(Tested on Chromebook)

PR #5045, opened 2018-06-14
Solve the problem that path will disppear when second finger touch sc…

boomyao:
#992


asturur:
we need to properly support multi touch events. we cannot patch just piece of code around of specific features.

asturur:
this has been fixed with recent changes

PR #5072, opened 2018-06-26
Added support for per object min and max height and width

dennisrjohn:
Sometimes our customers accidentally make objects too small, this can have a couple of averse effects. 

If the object is a group, containing an image, and the image is scaling with the group, this can cause an error because the object is trying to draw on a canvas that has a 0 dimension. 

If not, and they resize the object to 0 height, width, or both, it's very difficult to grab and resize the object.

This PR allows this for any fabric object:
let groupProps = {
                left: 10,
                top: 10,
                width: 100,
                height: 100,
                minWidth: 20,
                minHeight: 20,
                maxWidth: 300,
                maxHeight: 300
            };

group = new fabric.Group(groupItems, groupProps);

I based the code off of similar functionality in textbox_behavior.mixin.js

### DIFF
--- a/src/mixins/object_behavior.mixin.js
+++ b/src/mixins/object_behavior.mixin.js
@@ -1,0 +1,37 @@
+(function() {
+
+  /**
+   * Override _setObjectScale and apply max and min dimensions
+   */
+  var setObjectScaleOverridden = fabric.Canvas.prototype._setObjectScale;
+
+  fabric.Canvas.prototype._setObjectScale = function (localMouse, transform,
+    lockScalingX, lockScalingY, by, lockScalingFlip, _dim) {
+
+    var t = transform.target;
+    if (by === 'x') {
+      var tw = t._getTransformedDimensions().x;
+      var w = t.width * (localMouse.x / tw);
+      if (w >= t.getMinWidth() && w <= t.getMaxWidth()) {
+        t.set('width', w);
+        return true;
+      }
+    }
+    else if (by === 'y') {
+      var th = t._getTransformedDimensions().y;
+      var h = t.height * (localMouse.y / th);
+      if (h >= t.getMinHeight() && h <= t.getMaxHeight()) {
+        t.set('height', h);
+        return true;
+      }
+    }
+    else if (by === 'equally') {
+      var tw = t._getTransformedDimensions().x;
+      var w = t.width * (localMouse.x / tw);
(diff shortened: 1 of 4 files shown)


dennisrjohn:
I'm not sure why the tests are failing, that's not in any of the code I changed.

src/shapes/object.class.js
```diff
@@ -448,6 +448,35 @@
...
+     * @default
+     */
+    minHeight:                null,
+
+    /**
+     * When not null, will not allow the height of the object to be larger than the value
+     * @type Number
+     * @default
+     */
+    maxHeight:                null,
```
asturur: if we ask for numbers as type, we do not want to initialize as null. you can initialize as Number.INFINITY as you did in the functions. and let the getters out.

src/mixins/object_behavior.mixin.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+    else if (by === 'equally') {
+      var tw = t._getTransformedDimensions().x;
+      var w = t.width * (localMouse.x / tw);
+      if (w >= t.getMinWidth() && w <= t.getMaxWidth() && h >= t.getMinHeight() && h <= t.getMaxHeight()) {
+        return setObjectScaleOverridden.call(fabric.Canvas.prototype, localMouse, transform,
+          lockScalingX, lockScalingY, by, lockScalingFlip, _dim);
+      }
+    }
+  };
```
asturur: The problem of this kind of behaviour is that it does not take in account
Images, Path, Polygon and Polylines.

Is also mixing widht/height changes with Scaling changes, something i m not completely understand.

dennisrjohn:
I followed your advice above and removed the need for getters and setters.

As far as mixing width and height changes with scaling changes, I was expanding the pattern established in textbox_behavior.mixin.js, which sets a minimum width for textboxes.

asturur:
That basically works JUST for textboxes, There is a way to do this, and can be also done, let me think a bit about it.

dennisrjohn:
Ok, thanks, that would be very much appreciated.

I tested this with groups, rects and circles, and it worked like a charm. But didn't try paths and polygons.

asturur:
Maybe i misread the code.

2018-06-27 17:45 GMT+02:00 dennisrjohn <notifications@github.com>:

PR #5075, opened 2018-06-27
 toSVG() excludeFromExport for backgroundImage and overlayImage

durga598:
added `excludeFromExport` check for backgroundImage and overlayImage in toSVG(), all other objects `excludeFromExport = true` working fine.


asturur:
The PR is ok. Do not worry about the commit list, just check that the tab `Files Changed` is clean.

asturur:
Going to merge this, i just need to slim down the test

PR #5085, opened 2018-07-04
Fix Circle toSVG method adding an incorrect double quote

nmorel:
I also wonder if the condition `if (angle === 0)` should be changed.  
I encountered the double quote even though my circle was "full".  
The calculated angle was `1.4210854715202004e-14`.  
The condition could be something like `if (angle < 0.001 && angle > -0.001)` ?

### DIFF
--- a/src/shapes/circle.class.js
+++ b/src/shapes/circle.class.js
@@ -114,7 +114,7 @@
           '" transform="', this.getSvgTransform(),
           ' ', this.getSvgTransformMatrix(), '"',
           this.addPaintOrder(),
-          '"/>\n'
+          '/>\n'
         );
       }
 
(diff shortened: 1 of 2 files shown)


asturur:
Can you open a different issue for the angle? we should try to make the angle be calculated as 0 and figure out where the approximation error come from

asturur:
Also since we do not have tests for this case, otherwise it would have fail, would you add one?

asturur:
Perfect!

asturur:
oh there is some linting issue.
Can you try to run `npm run lint_tests` and see the output?

nmorel:
Looking at Travis, it seems the issue is in master branch, on your last commit

asturur:
fixed lint on master, rebuilding this and if is green, merging.

PR #5093, opened 2018-07-09
simpler test for excludeFromExport for bgImage an overlayImage

asturur:
@durga598 this is a way to test the same thing you did but simpler.

Your test was testing how many times a reviver was called as an indication if background and overlay where there. That test could have failed for a series of reasons not related to exclude from export.

In this way the test is focused on the svg output.

### DIFF
--- a/test/unit/canvas_static.js
+++ b/test/unit/canvas_static.js
@@ -867,44 +867,23 @@
   });
 
   QUnit.test('toSVG with exclude from export background', function(assert) {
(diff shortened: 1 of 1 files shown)


durga598:
From where I will get the SVG data? As you used expectedSVG?

asturur:
i just output the svg itself, verified it was correct with the eyes, and decided it would become the expected one

Issue #4555, opened 2017-12-12
Fabricjs 2.0 Text decoration


DangerD256:
2.0

## Test Case
https://jsfiddle.net/bLn363j1/1/

Can't set underline/overline/line-through...
Worked well in 1.x

DangerD256:
line-through: true/false - doesn't work

samael205:
Please read the FabricJS Docs

http://fabricjs.com/docs/fabric.IText.html

You must use canvas.getActiveObject().set("linethrough", true);

darron1217:
Doc says its type is `String` but actually it accepts `boolean`
Doc should be fixed 

https://github.com/fabricjs/fabric.js/blob/efa490fff020b287561a31d8d388850b5762152c/src/shapes/text.class.js#L93-L114

pidugusundeep:
`linethrough:True` works just fine for me

asturur:
Just in case someone can still hit this issue, i updated the official doc with a note.

http://fabricjs.com/fabric-text

PR #5101, opened 2018-07-13
Added callback in setPatternFill

durga598:
As Pattern constructor taking callback, we can use it.
And `fabric.util.loadImage` used if we use url in source,  so removed in demo.

text.class doc changed #4555

In setCoords() doc, there was a link when to call, added @link to act as a link, jsdoc


asturur:
Thank you @durga598  this is good.

PR #5121, opened 2018-07-25
Fix multiple applications of blend image filter with transparent images

brandondong:
Repro case: https://jsfiddle.net/ojt42731/94/


asturur:
the change looks good.
Would you mind adding a small visual test for it? We have a filter test for resizing and adding a test for blend image should be straight forward

brandondong:
No idea why the tests are failing. Passing locally for me.

asturur:
i ll check them

asturur:
i think i ll merge as it is and fix locally since is easier. We need a more colorfull image if we want to make a test, if is very faint is harder to see visually what is going wrong.

PR #5124, opened 2018-07-27
Reset superscript subscript

kalpeshdeo:
As per current implementation there is no way to unsubscript or unsuperscript which is equally important as subscript and superscript. So added support for the same.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -163,7 +163,7 @@
       size:       0.60,
       baseline:   0
     },
-  
+
     /**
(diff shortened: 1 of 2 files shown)


asturur:
Hi @kalpeshdeo, thanks for the PR.
What is wrong with approach of removing dy and fontSize style information from the text object at desired range?

Do we really need a new method?

kalpeshdeo:
Hi @asturur,
There is nothing wrong with that. But, I always prefer to have separate methods so that, existing code will untouched as well as its easy to maintain in future. That's why.

PR #5130, opened 2018-07-29
Fix visual tests

asturur:
Rewrite the blend image filter test reusing code and using images easier to inspect with the eye


asturur:
Ok this was hard. So much confusion while i just made an error in .travis.yml

PR #5138, opened 2018-08-03
Added requestRenderAllBound

durga598:
As `requestRenderAllBound` is undefined in `_initStatic` method of `fabric.Canvas` class `initialize` method. Here is [jsfiddle](https://jsfiddle.net/durga598/utaodqwm/).

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -61,6 +61,7 @@
     initialize: function(el, options) {
       options || (options = { });
       this.renderAndResetBound = this.renderAndReset.bind(this);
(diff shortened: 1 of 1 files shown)


asturur:
Hi Durga, i m fine with this change.
I did not see the utility in first place, but at the end it can be useful too if you have multiple callback finishing very near in time.

Just to be clear to whoever read this PR, the bound method are just a typing save utility to avoid putting in callbacks

```this.loadSomething(canvas.renderAll.bind(canvas))```

that becomes:

```this.loadSomething(canvas.renderAllBound)```

PR #5153, opened 2018-08-08
Fix interaction between grouping and enter edit for text

### LINKED ISSUE #5152: Active selection + textbox is bugging
## Version
2.3.0

## Test Case
KitchenSink

## Steps to reproduce
Edit a text, next click and drag to make an active selection but release the mouse on top of text that you were editing. It will put that textbox back in editing mode. And it's mixing editing mode with selection mode... 
Check the video here: https://cl.ly/0U2D1g2u3m0s

## Expected Behavior
Keep all textbox in selection mode

## Actual Behavior
Textbox are switched in editing mode with no reason

asturur:
fix #5152


asturur:
@Aubron testing this fix is becoming a problem

so weird

Issue #5155, opened 2018-08-08
Href parsing in SVG use directive


DylanPaulusSEL:
(https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href)
xlink:href is being deprecated for href

## Version
2.3.0

## Test Case
http://jsfiddle.net/uybeqc5r/

## Steps to reproduce

Use `loadSVGFromUrl`/`loadSVGFromString` with an SVG `<use />` tag that contains `href` instead of `xlink:href`. (`<use href="#myId" />`)

## Expected Behavior

Should parse the SVG into the fabric canvas.

## Actual Behavior

Throws a null reference error on `xlink = el.getAttribute('xlink:href').substr(1)` when using `href`.
![image](https://user-images.githubusercontent.com/30325527/43851028-7453376a-9aee-11e8-8069-033365140c6a.png)

--------

Willing to make a PR for this if everything seems fine. Thinking of adding checks for either `xlink:href` and `href`. (eg. `xlink = (el.getAttribute('xlink:href') || el.getAttribute('href')).substr(1)`

asturur:
it make sense to support href on top of xlink:href to do not immediately being completely incompatible with svg 2.

But we still focus on svg 1.1 ( with 1.2 addons where possible ) compatibility, so is not deprecated for us.

Most of svg2 features are really out of reach for us.

asturur:
merged

PR #5156, opened 2018-08-08
Feature/add href parsing

DylanPaulusSEL:
Resolves issues in parsing SVG2.0 `<use />` tags as referenced #5155.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -464,7 +464,7 @@
 
     while (nodelist.length && i < nodelist.length) {
       var el = nodelist[i],
-          xlink = el.getAttribute('xlink:href').substr(1),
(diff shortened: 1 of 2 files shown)


asturur:
This looks good!
Was there any Xlink:href test too?
If there is not, can you add it please?

DylanPaulusSEL:
@asturur I believe [this test](https://github.com/fabricjs/fabric.js/blob/41fceef824634dadcdce29d5ecb8b05981daff63/test/unit/parser.js#L333) implicitly tests `xlink:href`.

asturur:
it does, but implicitly. It can be removed because ie9 is no more supported and that getById polifyll is no more needed.

Would you mind jsut duplicate your test as it is, using xlink:href and changing the title to it?

that would make me happy that we locked down the situation of the href parsing

PR #5176, opened 2018-08-18
fixed bug positioning group

### LINKED ISSUE #5169: Free Drawing Mode: Circle and Spray move after completing line
## Version
2.3.5 (I assume the demos use the most recent release)

## Test Case
http://fabricjs.com/freedrawing

## Steps to reproduce
Use the Circle or Spray modes. Draw something.

## Expected Behavior
The line should be drawn using the given mode, and the line should be drawn where you drew it.

## Actual Behavior
The line moves up and to the left after you release the mouse.
I've tested in Safari and Chrome, both show the same issue.

asturur:
close #5169

### DIFF
--- a/src/brushes/circle_brush.class.js
+++ b/src/brushes/circle_brush.class.js
@@ -99,7 +99,7 @@ fabric.CircleBrush = fabric.util.createClass(fabric.BaseBrush, /** @lends fabric
 
       circles.push(circle);
     }
-    var group = new fabric.Group(circles, { originX: 'center', originY: 'center' });
+    var group = new fabric.Group(circles);
     group.canvas = this.canvas;
 
     this.canvas.add(group);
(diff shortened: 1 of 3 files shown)


src/brushes/circle_brush.class.js
```diff
@@ -99,7 +99,7 @@ fabric.CircleBrush = fabric.util.createClass(fabric.BaseBrush, /** @lends fabric
 
       circles.push(circle);
     }
-    var group = new fabric.Group(circles, { originX: 'center', originY: 'center' });
+    var group = new fabric.Group(circles);
```
asturur: no need to force this origin

src/brushes/spray_brush.class.js
```diff
@@ -109,7 +109,7 @@ fabric.SprayBrush = fabric.util.createClass( fabric.BaseBrush, /** @lends fabric
       rects = this._getOptimizedRects(rects);
     }
 
-    var group = new fabric.Group(rects, { originX: 'center', originY: 'center' });
+    var group = new fabric.Group(rects);
```
asturur: no need to force this origin

PR #5259, opened 2018-09-22
avoid run selection logic on mouseup for clicks

### LINKED ISSUE #5203: Cursor events not detected on overlapping objects with selectable:false
## Version
2.3.5

## Test Case
http://jsfiddle.net/Da7SP/2732/

## Steps to reproduce
Case 1: Hover over the red rectangle overlapped by the blue
Case 2: Click the red rectangle overlapped by the blue
Case 3: Mouse down and drag the red rectangle overlapped by the blue, then mouse up

## Expected Behavior
Case 1: Mouse cursor should have a 'pointer' cursor
Case 2: Red rectangle should be selected
Case 3. Red rectangle should be dragged

## Actual Behavior
Case 1: It takes default cursor. Something like css pointer-events: none is needed for objects with selectable: false
Case 2: Red rectangle is selected (Pass)
Case 3: It creates a selection box and then the red rectangle is selected. If any other object is nearby the whole group is selected.

I'm not sure if this is a bug or a feature request. Is it possible to detect the object underneath while hovering, when objectPreserveStacking is true? How should this be handled? Is there a CSS property like pointer-events: none?

asturur:
Run the grouping logic just in case of selection drags.
The mouseup click should not generate selections

close #5203

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -359,7 +359,7 @@
         shouldRender = transform.actionPerformed;
       }
 
-      if (target || !isClick) {
+      if (!isClick) {
         this._maybeGroupObjects(e);
         shouldRender || (shouldRender = this._shouldRender(target));
       }


thotar:
A side effect of this change is that if there is a selectable object under a not selectable object the below object was selected on the click but now it is not.

asturur:
can you open an issue with a jsfiddle that show the problem? so we can talk about it.

Issue #5252, opened 2018-09-20
loadSVGFromURL error


AlexSergey:
## Version
2.3.6

## Test Case
When we try to set incorrect URL we get empty xml variable. It call callback with null, but we need prevent function, something like return false, etc. Because after callback(null) this code doesn't dying and try to call parseSVGDocument

http://joxi.ru/gmvpaj4TL85vPA

## Steps to reproduce
1. Go to http://fabricjs.com/kitchensink
2. Set "Execute"
3. Paste the code:
fabric.loadSVGFromURL('https://www.cleverbrush.com/f/mu27s/w5jspy20v+.svg', objects => {
console.log(objects);
});

## Expected Behavior
We need to get only null in console

## Actual Behavior
We get also error - Cannot read property 'documentElement' of null

PR #5263, opened 2018-09-23
Fix loadSVGFromUrl bug #5252

AlexSergey:
I blocked calling parse function after error happend


asturur:
Can you explain a little better what the problem is in this case?

AlexSergey:
This case - https://github.com/fabricjs/fabric.js/issues/5252

asturur:
oh right!

PR #5270, opened 2018-09-25
Can not scale pattern image with patternTransform

nqcuong72:
See sample code as following
          canvas.setZoom(3);
          fabric.Image.fromURL(url, function(img) {
            var pattern = new fabric.Pattern({
              source: img.getElement(),
              // 1 is keep 100% image size it does not affected by canvas scale
              patternTransform: [1, 0, 0, 1, 0, 0], 
              repeat: 'repeat'
            });
            obj.setPatternFill(pattern);
            canvas.renderAll();

Output pattern image is all zoom 3 times, not keep as 1 as patternTransform parameter

Solution: Update from version 2.3.6 at line 13564
From: ctx.transform(t[0], t[1], t[2], t[3], t[4], t[5]);
To: ctx.setTransform(t[0], t[1], t[2], t[3], t[4], t[5]);

Please update in the next version


nqcuong72:
I have updated the request.
Problem: Rendered pattern image inside svg element is zoomed to 300%. It is not keep as 100% as input data to patternTransform parameter.

asturur:
But this is a pull request, why not an issue?

is the solution included here?

nqcuong72:
Yes, I have solution
Solution: Update from version 2.3.6 at line 13564
From: ctx.transform(t[0], t[1], t[2], t[3], t[4], t[5]);
To: ctx.setTransform(t[0], t[1], t[2], t[3], t[4], t[5]);

PR #5272, opened 2018-09-25
Changelog Syntax - Converting version numbers to headings

curtisj44:
This is a simple suggestion/fix:

GitHub automatically provides anchor links around each heading and the ability to link to a specific version summary is useful, i.e:

> That one Safari issue we've been fighting has been fixed in v2.2.4: https://github.com/curtisj44/fabric.js/blob/master/CHANGELOG.md#224.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,21 @@
-**Version 2.4.0**
+# Changelog
+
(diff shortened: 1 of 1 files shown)


asturur:
looks good to me, thanks

PR #5282, opened 2018-09-30
fixed background / overlay image not recalculating zoom value

### LINKED ISSUE #5280: SVG backgroundImage does not re-render on canvas zoom
## Version
2.4.0

All browser

## Steps to reproduce
Set an svg as the canvas background image, then use canvas.zoomToPoint to zoom in.

## Expected Behavior
SVG re-render to zoom factor and look sharp. was working with version 2.2.
http://jsfiddle.net/at60ojdc/

## Actual Behavior
SVG does not re-render so it get zoom but look super pixelated.
http://jsfiddle.net/t34La5yk/

asturur:
Add a canvas reference so background or overlay image know the zoom value

Close #5280

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -14,6 +14,8 @@
       toFixed = fabric.util.toFixed,
       transformPoint = fabric.util.transformPoint,
       invertTransform = fabric.util.invertTransform,
+      getNodeCanvas = fabric.util.getNodeCanvas,
+      createCanvasElement = fabric.util.createCanvasElement,
 
       CANVAS_INIT_ERROR = new Error('Could not initialize `canvas` element');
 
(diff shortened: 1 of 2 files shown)


src/static_canvas.class.js
```diff
@@ -459,13 +467,18 @@
     __setBgOverlayImage: function(property, image, callback, options) {
       if (typeof image === 'string') {
         fabric.util.loadImage(image, function(img) {
-          img && (this[property] = new fabric.Image(img, options));
+          if (img) {
+            var instance = new fabric.Image(img, options);
+            this[property] = instance;
+            instance.canvas = this;
+          }
           callback && callback(img);
```
durga598: Can we return fabric image object i.e. instance, instead of img?

asturur: That is another method, is called `fabric.Image.fromUrl(...)`

durga598: Yes but on callback of `setBackgroundImage/ setOverlayImage` it returns image element not fabric.Image instance.

asturur: Does it even make sense that those callback return something?
You are right they should at least return the same thing.

I would remove the method completely to be honest. Devs should create theyr background object and set it normally.

PR #5284, opened 2018-09-30
Rework SVG code to have clipPath friendly transformations

### LINKED ISSUE #5264: clipPath toSVG bug
## Version
2.4.0

## Test Case
I load svg with clip path. In canvas it looks good but after I call toSVG it is disapear.
http://jsfiddle.net/he53rzy6/3/

asturur:
Wrap every object in a group with transform, so that clipPath can be embedded and auto transformed.

Improve code reuse for toSVG export

close #5264


asturur:
I may have broken some SVG exports, i ll add more tests before publishin 2.4.2

PR #5285, opened 2018-10-01
Textbox character spacing issue fix for multiple languages

sakthikanth:
If we apply charSpacing for a text box / iText the characters in english are split into single characters correctly.But for other languages the single characters are split with their consonant letters while using the charspacing in the textbox.

The current method which split tamil language as follows
`let graphemes = ["த","ம","ி","ழ","்"];`
Fix :
`let graphemes = ["த","மி","ழ்"];`

For the demo of character spacing for multiple languages.
Please visit : http://52.14.184.2/lang/demo_lang.html

Thank You./
			.


asturur:
I understand the idea behind this fix.

Grapheme splitting should be canvasless, and grapheme split is also overridable to support many languages ( with specific libraries ).

We should not rely on canvas to split something without the knowledge of how it should be splitted and rely on a graphic output.

This approach would not work for emoji for example if you miss the font for it.

Or would not work for hindi if you miss the font for it.

PR #5287, opened 2018-10-01
perPixelTargetFind not working in nested group.

codertx:
perPixelTargetFind property is not working with object in nested group. When I replace isTargetTransparent methods params with pointer relative to canvas, everything works fine.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1158,6 +1158,7 @@
 
       var ignoreZoom = true,
           pointer = this.getPointer(e, ignoreZoom),
+          globalPointer = pointer,
           activeObject = this._activeObject,
           aObjects = this.getActiveObjects(),
           activeTarget, activeTargetSubs;
@@ -1167,7 +1168,7 @@
       // if active group just exits.
       this.targets = [];
 
-      if (aObjects.length > 1 && !skipGroup && activeObject === this._searchPossibleTargets([activeObject], pointer)) {
+      if (aObjects.length > 1 && !skipGroup && activeObject === this._searchPossibleTargets([activeObject], pointer, globalPointer)) {
         return activeObject;
       }
       // if we hit the corner of an activeObject, let's return that.
@@ -1175,7 +1176,7 @@
         return activeObject;
       }
       if (aObjects.length === 1 &&
-        activeObject === this._searchPossibleTargets([activeObject], pointer)) {
+        activeObject === this._searchPossibleTargets([activeObject], pointer, globalPointer)) {
         if (!this.preserveObjectStacking) {
           return activeObject;
         }
@@ -1185,7 +1186,7 @@
           this.targets = [];
         }
       }
-      var target = this._searchPossibleTargets(this._objects, pointer);
+      var target = this._searchPossibleTargets(this._objects, pointer, globalPointer);
       if (e[this.altSelectionKey] && target && activeTarget && target !== activeTarget) {
         target = activeTarget;
         this.targets = activeTargetSubs;
@@ -1196,13 +1197,14 @@
     /**
      * @private
      */
-    _checkTarget: function(pointer, obj) {
+    _checkTarget: function(pointer, obj, globalPointer) {
+      globalPointer = globalPointer || pointer;
       if (obj &&
           obj.visible &&
           obj.evented &&
           this.containsPoint(null, obj, pointer)){
         if ((this.perPixelTargetFind || obj.perPixelTargetFind) && !obj.isEditing) {
-          var isTransparent = this.isTargetTransparent(obj, pointer.x, pointer.y);
+          var isTransparent = this.isTargetTransparent(obj, globalPointer.x, globalPointer.y);
           if (!isTransparent) {
             return true;
           }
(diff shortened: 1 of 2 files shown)


asturur:
So we would search target with the normalized pointer for bounding box, but the per pixel target would work with the non normalized one.

That would make sense. I wonder if a nicer fix is possible. And tests, we need tests to ensure it works.

codertx:
Would you please indicate which part of code can be improved. Thanks.

I'll add some test cases soon.

src/canvas.class.js
```diff
@@ -1196,13 +1199,14 @@
     /**
      * @private
      */
-    _checkTarget: function(pointer, obj) {
+    _checkTarget: function(pointer, obj, globalPointer) {
+      globalPointer = globalPointer || pointer;
```
asturur: is this check necessary? it looks like that we always pass in globalPointer.

src/canvas.class.js
```diff
@@ -1216,18 +1220,18 @@
     /**
      * @private
      */
-    _searchPossibleTargets: function(objects, pointer) {
```
asturur: even if those 2 methods are private, we could still document them
```
Function used to search inside objects an object that contains pointer in bounding box or that contains pointerOnCanvas when painted
@param [fabric.Object] objects array to look into
...
```

can you complete the JSDOC information for _searchPossibleTargets  and checkTarget ?

asturur:
If address the comments looks good.
Let me know how it goes for testing.

src/canvas.class.js
codertx: I wonder if someone might call this private method directly.
Should I remove this check or just document this method?

src/canvas.class.js
codertx: sure

src/canvas.class.js
asturur: Document the method is enough, is a private method.
If someone call it directly is gonna have problems anyway since those methods get changed with no information to devs.
We can put this check in searchPossibleTargets that while being private too, is at least one level up of this.

what do you think?

codertx: That make sense to me.

test/unit/canvas.js
```diff
@@ -790,6 +790,36 @@
...
+    var triangle = makeTriangle({ left: 0, top: 0 }),
+        target,
+        group1 = new fabric.Group([triangle], {
+          subTargetCheck: true
+        }),
+        group2 = new fabric.Group([group1], {
+          subTargetCheck: true
+        });
+
+    canvas.add(group2);
```
asturur: While this test is ok, the generated group is too simple.
can we use something like this:

```javascript
var circle = new fabric.Circle({ radius: 30, top: 30, left: 30, fill: 'yellow'});

var circle2 = new fabric.Ellipse({ rx: 10, ry: 20, top: 40, left: 0, fill: 'blue'});

var circle3 = new fabric.Circle({ scaleX: 2, scaleY: 2, radius: 10, top: 120, left: -20, fill: 'red'});

var rect = new fabric.Rect({ width: 100, height: 80, top: 50, left: 60, fill: 'purple' });

var group = new fabric.Group([circle, circle2], { scaleX: 3, angle: 45, opacity: 0.5 });

var group2 = new fabric.Group([group, rect, circle3], { opacity: 0.5, top: 20, left: 20, scaleX: 0.8, scaleY: 1.2 });

canvas.add(group2);
```
that looks like this:

image

and test edge cases? we should be able to succesfully target all the shapes.

codertx:
It's seems that object in nested group should normalize the pointer just once or we will get wrong result when calling containsPoint.

Please test cases I just added :).

test/unit/canvas.js
```diff
@@ -790,6 +790,67 @@
...
+    assert.equal(canvas.targets.length, 2, 'Subtargets length should be 2');
+    assert.equal(canvas.targets[0], rect, 'The deepest target should be rect');
+    target = canvas.findTarget({
+      clientX: 9, clientY: 145
+    });
+    assert.equal(target, group3, 'Should return the group3 now');
+    assert.equal(canvas.targets.length, 2, 'Subtargets length should be 2');
+    assert.equal(canvas.targets[0], circle2, 'The deepest target should be circle2');
+    canvas.perPixelTargetFind = false;
+    canvas.remove(triangle);
```
asturur: this is group3, not triangle

asturur:
yes i think t his is because normalizePointer will normalize everything up to the canvas level, so everytime you normalize starting from absolute.

Those changes are less invasive, i need some time to download and test the branch, but i think we are good. We still avoided angle and skew in the test case, i do not understand why

test/unit/canvas.js
codertx: Sorry about that. I will fix it soon.

codertx: Sorry about that. I will fix it soon.

codertx:
Please let me know if there's any problem.

asturur:
Hey, thank you.
This is a nice pr, a nice fix and thanks for the flexibility in rewriting the test.

PR #5289, opened 2018-10-02
Fix typo in resize_filter.class.js.

charlie-s:
The `sliceHack` filter had a typo in the documentation; was referred to as `sliceHacl`.


asturur:
thanks

melchiar:
while you're at it maybe fix the spelling of resizyType too :)

PR #5312, opened 2018-10-16
when enter shift key then enter mousedown other selected objects unse…

yanzhou-share:
1、select some object
2、enter shift key
3、click canvas and not on any object
4、selected objects is unselect


yanzhou-share:
Or override _shouldClearSelection  add e.shiftkey

asturur:
Generic apps abbehaviour for multi select is that it will destroy the selection. Why should we do differently?

yanzhou-share:
ok. good idea！

Issue #5335, opened 2018-10-25
surface chrome 70 can't touch draw line


zhishaofei3:
## Version
1.7.22 and lastest version

## Test Case
http://fabricjs.com/freedrawing

## Information about environment
Nodejs or browser?
Which browsers?
Microsoft ** Surface** & **Chrome 70.0.3538.77**

## Steps to reproduce
freedrawing mode use mouse draw line is normal
but **touch draw line, no response**

## Expected Behavior
add a line

## Actual Behavior
no line

zhishaofei3:
```javascript
var pointerX = function(event) {
    // looks like in IE (<9) clientX at certain point (apparently when mouseup fires on VML element)
    // is represented as COM object, with all the consequences, like "unknown" type and error on [[Get]]
    // need to investigate later
      if (event.touches && event.touches.length) {
        return event.touches[0].clientX
      } else if(typeof event.clientX != 'undefined') {
        return event.clientX;
      } else {
        return 0
      }
        // return (typeof event.clientX !== unknown ? event.clientX : 0);
    },

      pointerY = function(event) {
        if (event.touches && event.touches.length) {
          return event.touches[0].clientY
        } else if(typeof event.clientY != 'undefined') {
          return event.clientY;
        } else {
          return 0
        }
        // return (typeof event.clientY !== unknown ? event.clientY : 0);
      };
```

zhishaofei3:
in suface win10 chrome70
"ontouchstart" in document.documentElement // false
fabric.isTouchSupported // false
in mac chrome70 
"ontouchstart" in document.documentElement // true
fabric.isTouchSupported // true

PR #5337, opened 2018-10-25
pointerX pointerY compatible touches

zhishaofei3:
pointerX pointerY compatible touches
fixed surface chrome 70 can't touch draw line

https://github.com/fabricjs/fabric.js/issues/5335


IanYates:
Hi @zhishaofei3 - this fix works well.  I had this same issue in Chrome v70 reported today.   I've also verified correct behaviour still remains in Edge.

Any reason why you closed the PR?  Was another fix for this merged in instead?

asturur:
I have no clear situation of how events outside mouse should be handled in the whole.
I think we can ditch touches in place of pointer, but i m not sure how much effort is it.

PR #5338, opened 2018-10-25
check surface device chrome 70 touch supports

zhishaofei3:
check surface device chrome 70 touch supports

https://github.com/fabricjs/fabric.js/issues/5335
https://codepen.io/katwhocodes/post/using-javascript-to-detect-touch-events


asturur:
https://github.com/fabricjs/fabric.js/pull/4302  <--- what about this?
We switched from one to the other 1 year ago.

asturur:
also
https://github.com/maaaaark/bcSwipe/issues/4

asturur:
https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/

tatsuyafw:
@zhishaofei3 

Any updates for this PR?
I've also faced with same problem when using a touch monitor.
（Ref: [898045 \- Free drawing using a touch monitor shows nothing with fabric\.js\. \- chromium \- Monorail](https://bugs.chromium.org/p/chromium/issues/detail?id=898045)）

JSteunou:
I @asturur been a while :) We are also facing the same problem. Any chance to see this fix published?

JSteunou:
I think previous fix #4302 was to aggressive. Chrome 70 comes with better touch support, so we need more fine grain now. This PR should be rewrite against source file, but the idea is good.

asturur:
is just unclear to me if this is fixing because:

`"ontouchstart" in fabric.document.documentElement`

OR because the new condition is:

`navigator.maxTouchPoints > 0`

The original authour never answered.

JSteunou:
I will search about this monday. Ping me if I forget 😉

JSteunou:
:+1: 

If someday this bug appears again, this can be a good resource https://patrickhlauke.github.io/touch/touchscreen-detection/

Issue #5141, opened 2018-08-05
group object with subTargetCheck area be smallerwhen canvas zoomed


liajoy:
## Version
2.3.0

## Test Case
https://jsfiddle.net/8ea1ux97/3/

## Steps to reproduce
1. set canvas zoom.
2. add a group with objects and set subTargetCheck to true.
3. listen to mousemove event, the area can fire mousemove be smaller.
## Expected Behavior
The whole red rect should fire mousemove event.
## Actual Behavior
Only the center part of rect responds event.

rikkarv:
I suggest that you should set canvas dimensions, like:

canvas id ="c" width="300" height="300"
or
 new fabric.Canvas('c', {width: 300, height: 300}).

 Plus, as in documentation [http://fabricjs.com/docs/fabric.Group.html#addWithUpdate](url) addWithUpdate(object) has a parameter, the object that you are adding to the group.
You should set a variable rectangle and in line 14 you call: g.addWithUpdate(rectangle), instead of g.addWithUpdate(). 
Finally, get rid of g.item(0).setCoords().

[https://jsfiddle.net/8ea1ux97/6/](url)

liajoy:
@rikkarv 
I know the key point of event area zoom is setCoords method.But someTimes haven't call setCoords make object get wrong position?
https://github.com/fabricjs/fabric.js/wiki/When-to-call-setCoords

rikkarv:
I really don't know and i have no personal experience with such setCoords issues.

liajoy:
@asturur  I have found the reason. Maybe create a pr.

asturur:
no is also probably how you create the group with add and addwithupdate.
You should not call setCoords when object is grouped

liajoy:
But if not call setCoords, how to make coords right when set position of object?
https://jsfiddle.net/8ea1ux97/20/

asturur:
you call setCoords on the group not of the items(0).
I m not sure it work in that case.

Also the best way to add a group is:

`new fabric.Group([object1, object2, ...], options);`

asturur:
@liajoy please check #5347

liajoy:
got it. The best practice is not control the object on a group.

PR #5347, opened 2018-10-27
Add more test for subTargetCheck and zoom + pan

asturur:
Just add some tests to ensure that when there is a viewport transform, the subTargetCheck works correctly.


asturur:
followup of #5141

PR #5375, opened 2018-11-07
more compliant gradient parsing

### LINKED ISSUE #5370: SVG not being properly rendered
## Version
Kitchen Sink Version

Chrome Browser

## Steps to reproduce
Go to fabric's [Kitchen Sink demo page](http://fabricjs.com/kitchensink)
Paste in the following SVG code: https://pastebin.com/B6sRU0Bf

## Expected Behavior
An SVG that resembles this:
https://i.imgur.com/WNbjSQh.png

## Actual Behavior
An SVG that looks like this:
https://i.imgur.com/78lX8O8.jpg

asturur:
close #5370 

There were 2 errors in our gradients xlink:href parsing:

1) should have been recursive
2) should copy properties just if missing ( was never copying properties )
3) should copy children only if missing ( was always adding children )

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -229,14 +229,7 @@
     }
 
     // identity matrix
-    var iMatrix = [
-          1, // a
-          0, // b
-          0, // c
-          1, // d
-          0, // e
-          0  // f
-        ],
+    var iMatrix = fabric.iMatrix,
 
         // == begin transform regexp
         number = fabric.reNum,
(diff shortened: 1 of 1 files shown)


src/parser.js
```diff
@@ -440,25 +433,6 @@
...
-      return el;
-    }
-    var node, i, len, nodelist = doc.getElementsByTagName('*');
-    for (i = 0, len = nodelist.length; i < len; i++) {
-      node = nodelist[i];
-      if (id === node.getAttribute('id')) {
-        return node;
-      }
-    }
-  }
```
asturur: it was meant for old version of ie. let's see if it works fine without in node and ie10+ in case i will add it back

PR #5379, opened 2018-11-09
Fix the draggability of small objects

### LINKED ISSUE #5369: Cannot drag small objects
## Version
2.4.0

## Test Case
http://fabricjs.com/many-objects

Safari 12.0

## Steps to reproduce
Select one of the dots in the right canvas and drag it around.

## Expected Behavior
The selected dot should be dragged.

## Actual Behavior
The selected dot is resized. The problem is that apparently for very small objects resizing has a higher priority over dragging.

asturur:
Add a force rule ( we had it before somehow ) that objects cannot be scaled if not selected.
Realistically no one guess the corner position and start to drag it to scale.
Is often a drag operation that ends up with a scale one.
If you want to use a corner, first select the object.

close #5369

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -644,8 +644,8 @@
     /**
      * @private
      */
-    _getActionFromCorner: function(target, corner, e) {
-      if (!corner) {
+    _getActionFromCorner: function(alreadySelected, corner, e) {
+      if (!corner || !alreadySelected) {
         return 'drag';
       }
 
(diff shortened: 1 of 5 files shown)


src/canvas.class.js
```diff
@@ -644,8 +644,8 @@
     /**
      * @private
      */
-    _getActionFromCorner: function(target, corner, e) {
-      if (!corner) {
+    _getActionFromCorner: function(alreadySelected, corner, e) {
+      if (!corner || !alreadySelected) {
```
asturur: target was there, unused.
Liinntttt where are you?

asturur:
Not proud of this code, looks like a patch. The selection and action flow is complicated enough that i do not get away with a clean and straight forward code.

asturur:
needs an extensive test where on an object, a simulated click checks each of the 9 corners and the center of the object and inspect the resulting action type both for an unselected object and for a selected one.

PR #5392, opened 2018-11-14
save/restore the ctx around dashed line selections.

AndrewJDR:
Previously, the line dash setting could be left enabled, which could cause other user-side stroke calls on the top canvas to be dotted when they weren't intended to be.


asturur:
https://github.com/fabricjs/fabric.js/blob/57ef7bb2c11e7bad40b60eeb1d674569678eb1b6/src/canvas.class.js#L400

What about putting the save/restore couple at start and end of that function i linked so that we cover also all the strokeStyle and fillStyle that happen in that function?

Also please let's reproduce your use case of bug ( how were you drawing on the top canvas? ) and let's make a visual test from it.

asturur:
@AndrewJDR any chance to follow up?

asturur:
As as i said the PR looks good if we can move the save/restore couple in the function i suggested.

AndrewJDR:
Should be all set :+1:

PR #5428, opened 2018-12-13
Fix for IE11 IndexSizeError for negative source crops

CaptEmulation:
This codepen will not render on IE11 https://codepen.io/captemulation/pen/PXPORb (copy to your own codepen and replace `pen` with `debug` to view on IE11)

In IE11, if a negative number is passed in for sx or sy when using the 9 argument version of `drawImage` an `IndexSizeError` exception will be thrown.  This can occur with rounding errors making a number just past 0, e.g. -2.4e-12.

Negative numbers do not really make sense here.

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -507,8 +507,8 @@
     _renderFill: function(ctx) {
       var elementToDraw = this._element,
           w = this.width, h = this.height,
-          sW = Math.min(elementToDraw.naturalWidth || elementToDraw.width, w) * this._filterScalingX,
-          sH = Math.min(elementToDraw.naturalHeight || elementToDraw.height, h) * this._filterScalingY,
+          sW = Math.min(elementToDraw.naturalWidth || elementToDraw.width, w * this._filterScalingX),
+          sH = Math.min(elementToDraw.naturalHeight || elementToDraw.height, h * this._filterScalingY),
           x = -w / 2, y = -h / 2,
           sX = Math.max(0, this.cropX * this._filterScalingX),
           sY = Math.max(0, this.cropY * this._filterScalingY);


CaptEmulation:
> Can you fix that too?

Taking a looksee

CaptEmulation:
Add the additional boundary check and added a test.  Could not follow test style because I could not run anything in image test suite (timeout).  IDE also picked up on some reformatting.

dist/fabric.js
```diff
@@ -19931,14 +19931,15 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
...
-          x = -w / 2, y = -h / 2, elementToDraw = this._element;
-      elementToDraw && ctx.drawImage(elementToDraw,
-        this.cropX * this._filterScalingX,
-        this.cropY * this._filterScalingY,
-        sW,
-        sH,
-        x, y, w, h);
+      var elementToDraw = this._element,
+          w = this.width, h = this.height,
+          sW = Math.min(elementToDraw.naturalWidth || elementToDraw.width, w * this._filterScalingX),
```
asturur:
```suggestion
          sW = Math.min(elementToDraw.naturalWidth || elementToDraw.width, w) * this._filterScalingX,
```
we want to limit the source area, this._filterScaling apply anyway

dist/fabric.js
```diff
@@ -19931,14 +19931,15 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
...
+          sH = Math.min(elementToDraw.naturalHeight || elementToDraw.height, h * this._filterScalingY),
```
asturur:
```suggestion
          sH = Math.min(elementToDraw.naturalHeight || elementToDraw.height, h) * this._filterScalingY,
```

asturur:
Thanks for the test too.
Can you checkout away the changes to dist folder?

i usually do `git checkout master dist` and i rest them.

asturur:
image
ok we broke something.
i probably suggested something wrong?

asturur:
oh well if there is a filter scaling, the elementToDraw is already resized to the right dimension of width * filterScalingX. So as you did in the first version, the code was right, with filterScaling inside the parenthesis.

PR #5438, opened 2018-12-23
fixing path exports to SVG

### LINKED ISSUE #5436: Canvas toSVG ,the svg will break
## Version
2.4.5 bad
2.4.3 and 2.4.4 show well
## Test Case

brower
Which browsers?
chrome 70.0.3538.110
## Steps to reproduce
http://fabricjs.com/docs/fabric.Canvas.html#toSVG
canvas toSVG image
## Expected Behavior

image
image

asturur:
Close #5436 

I did not look at all manual tests last release, and i broke path svg export. This is a fix.
I have still to verify a path + clipPath path svg import/export


asturur:
ok this is not right yet. This breaks some kind of clipPaths.
OMG so hard.

asturur:
Added a test that fails for the broken part right now.

PR #5461, opened 2019-01-04
Fire "selection:changed" on click only if the selection is actually changed

finom:
The library shouldn't fire "selection:changed" and don't need to update textarea if user is clicked on the same cursor position because the selection coordinates remain the same.

Test case http://jsfiddle.net/Da7SP/3383/

### DIFF
--- a/src/mixins/itext_click_behavior.mixin.js
+++ b/src/mixins/itext_click_behavior.mixin.js
@@ -179,7 +179,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       this.selectionStart = newSelection;
       this.selectionEnd = newSelection;
     }
-    if (this.isEditing) {
+    if (this.isEditing && (start !== newSelection || end !== newSelection)) {
       this._fireSelectionChanged();
       this._updateTextarea();
     }


asturur:
looks good, but please add a tests, so that this behaviour gets removed we have an alarm.

The test should verify that event is fired when a click happen, and that is not fired when a click appens in the same position.
Eventually this should happen with keyboard strokes too ( left-right arrow ) when we are at the start or end of text and so nothing changes?

asturur:
Let me know if you need help with setting up tests, looks like we do not have any around for this area.

finom:
> Eventually this should happen with keyboard strokes too ( left-right arrow ) when we are at the start or end of text and so nothing changes?

Makes sense.

> Let me know if you need help with setting up tests, looks like we do not have any around for this area.

I will need to figure out how to simulate a click and calc coordinates where to simulate it. So far I have no any specific questions because I haven't looked on ways to do that. If you have some informational article or something like that it may help me in my research.

asturur:
i ll merge a test that fires a single click, you can branch different use cases from that

asturur:
link comment here
https://github.com/fabricjs/fabric.js/pull/5462#issuecomment-451704127

asturur:
run `npm run build` to be sure you are running on latest  build version
(tests run on dist/fabric.js)

Be sure node-canvas is installed, and run the tests with `npm run test` they will run on node only and they will be faster. The travis script will then run on all platforms.

install libraries for node-canvas ---> brew install pkg-config cairo libpng jpeg giflib

i can run tests reliably on various environments.
I m actually on highsierra, node 10.14.1, npm 6.4.1

finom:
@asturur it's much better now, thank you. But I still get tons of "timed out" tests. I've set `QUnit.config.testTimeout = 150000;` (10x times more than default value) but it didn't help.

asturur:
because for some reason when it was in devDeps if it was failing everything was catching fire listening to devs. Never happened to me.
The least i could do was moving it in optionaDependency.

Is also a production dependency if you do everything under node.

asturur:
@finom, i m planning a release today, 2.5.1 Should i wait for you or i can go straight to publish?
No pressure of any kind! Do as you feel like.

finom:
Ping. Just to make this issue to be open yet.

PR #5462, opened 2019-01-06
added a test for click and set selection on IText

asturur:
as support for #5461

### DIFF
--- a/test/unit/itext_click_behaviour.js
+++ b/test/unit/itext_click_behaviour.js
@@ -4,6 +4,7 @@
 
   QUnit.module('iText click interaction', {
     afterEach: function() {
+      canvas.clear();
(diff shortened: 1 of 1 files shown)


asturur:
@finom i ll merge this test as soon as the build pass.
Please merge back master in your branch and add more tests on this kind to validate your change.

you should test that:

- Clicking twice in the same point won't fire the event again
- Clicking in a point and then in another point more right will actually fire the event again
- Clicking with shift pressed ( add shiftKey: true to event data ) will fire the event and set selection start and selection end with different numbers

Try to set as i did maximum 2 clicks per test, or the timeout code become hard to read.

you can also use `canvas._onDoubleClick` to start the code that happens as a double click ( it selects a full word or full text )

PR #5473, opened 2019-01-10
strokeUniform property

stefanhayden:
new Object property strokeUniform allows the stroke width to always stay the same width as object scales.

This code has the stroke staying uniform but I'm still doing something wrong with the bounding box as it's still growing and shrinking as if the stroke would be scaling.

example5

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -592,8 +592,16 @@
       if (skewX === 0 && skewY === 0) {
         return { x: dimensions.x * this.scaleX, y: dimensions.y * this.scaleY };
       }
-      var dimX = dimensions.x / 2, dimY = dimensions.y / 2,
-          points = [
+      var dimX, dimY;
+      if (this.strokeUniform) {
+        dimX = this.width / 2;
+        dimY = this.width / 2;
+      }
+      else {
+        dimX = dimensions.x / 2;
+        dimY = dimensions.y / 2;
+      }
+      var points = [
             {
               x: -dimX,
               y: -dimY
@@ -616,7 +624,10 @@
         points[i] = fabric.util.transformPoint(points[i], transformMatrix);
       }
       bbox = fabric.util.makeBoundingBoxFromPoints(points);
-      return { x: bbox.width, y: bbox.height };
+      return this.strokeUniform ?
+        { x: bbox.width + this.strokeWidth, y: bbox.height + this.strokeWidth }
+        :
+        { x: bbox.width, y: bbox.height };
     },
 
     /*
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -595,7 +595,7 @@
       var dimX, dimY;
       if (this.strokeUniform) {
         dimX = this.width / 2;
-        dimY = this.width / 2;
+        dimY = this.height / 2;
       }
       else {
         dimX = dimensions.x / 2;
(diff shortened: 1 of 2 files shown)


src/shapes/object.class.js
```diff
@@ -560,6 +560,16 @@
      */
     noScaleCache:              true,
 
+    /**
+     * When `false`, the stoke width will scale with the object.
+     * When `true`, the stroke will always match the exact pixel size entered for stroke width.
+     * default to false
+     * since 2.5.1
```
asturur:
```suggestion
     * @since 2.6.0
     * @type Boolean
     * @default false
```
this is a new feature, need a minor bump.

asturur: ops i got it wrong

asturur:
```suggestion
     * @since 2.5.1
```
since is something that jsdoc recognize

asturur: oh jesus

asturur:
This will be for sure a multi pr feature. I would start with this. I m adding a couple of tests and merge it
I ll follow up with control box and then svg import/export

asturur:
is strokeUniform a good name?

asturur:
and text, we need to workout a way for text. Because in our prototype we did not need it for other reasons

src/mixins/object_geometry.mixin.js
```diff
@@ -592,8 +592,16 @@
       if (skewX === 0 && skewY === 0) {
         return { x: dimensions.x * this.scaleX, y: dimensions.y * this.scaleY };
       }
```
asturur: this was supposed to be a shortcut to save on calculations... keeping it now kills the feature. Or maybe we need to reorganize a bit the code. Most of the peple work without the skew values, not point in going for 4 point * matrix + bbox calculations. But i can agree is not probably a bottleneck.

asturur: what about this? would keep the shortcut somehow.

```javascript
_getTransformedDimensions: function(skewX, skewY) {
      if (typeof skewX === 'undefined') {
        skewX = this.skewX;
      }
      if (typeof skewY === 'undefined') {
        skewY = this.skewY;
      }
      var dimensions = this._getNonTransformedDimensions(), dimX, dimY,
            noSkew = skewX === 0 && skewY === 0;

      if (this.strokeUniform) {
        dimX = this.width;
        dimY = this.height;
      }
      else {
        dimX = dimensions.x;
        dimY = dimensions.y;
      }
      if (noSkew) {
       return _finalizeDiemensions(dimX, dimY);
      }
      else {
        dimX /= 2;
        dimY /= 2;
      }
      var points = [
            {
              x: -dimX,
              y: -dimY
            },
            {
              x: dimX,
              y: -dimY
            },
            {
              x: -dimX,
              y: dimY
            },
            {
              x: dimX,
              y: dimY
            }],
          i, transformMatrix = this._calcDimensionsTransformMatrix(skewX, skewY, false),
          bbox;
      for (i = 0; i < points.length; i++) {
        points[i] = fabric.util.transformPoint(points[i], transformMatrix);
      }
      bbox = fabric.util.makeBoundingBoxFromPoints(points);
      return this._finalizeDiemensions(bbox.width, bbox.height);
    },

    _finalizeDiemensions(width, height) {
      return this.strokeUniform ?
        { x: width + this.strokeWidth, y: height + this.strokeWidth }
        :
        { x: width, y: height };
    }
```

stefanhayden: yup that makes sense.

asturur:
need to finish this but i broke my branches, merging this and fixing.

Issue #5476, opened 2019-01-11
Support Relative Image URL src


nateevans:
**Affected Version**
Fabric 2.5.0

**Is your feature request related to a problem? Please describe.**
We save canvas JSON in a portable database. So when we save Images to Fabric canvas, we need relative Image src url. The browser and Fabric seem to work correctly with relative url like `/api/path/to/my/image`, but the Fabric api does not handle this as expected.

If we call `setSrc()` with a relative url like `/api/path/to/my/image`, update canvas etc, then call `getSrc()` the result will have the current browser schema and hostname prepended to the src string. Like: `http://127.0.0.1/api/path/to/my/image`.

This seems counter-intuitive behavior and maybe not intended to happen by Fabric.js team.

**Describe the solution you'd like**
I've narrowed down the cause to calling `return element.src` in `fabric.Image.prototype.getSrc`. It seems that the behavior of `HTMLImageElement` causes it to prepend hostname when accessing the property in this way. 

I've found that calling `element.getAttribute('src')` is capable of returning the un-altered relative url and doesn't seem to interfere with current behavior.

```
getSrc: function(filtered) {
  var element = filtered ? this._element : this._originalElement;
  if (element) {
    if (element.toDataURL) {
      return element.toDataURL();
    }
    return element.getAttribute('src');
  }
  else {
    return this.src || '';
  }
}
```

**Additional context**
I can submit a PR for this, but it is not clear to me how testing this should be handled within your testing framework.

melchiar:
A simple override to `fabric.Image.prototype.toObject` can accomplish this. Native support might not be a bad idea though. 

See https://stackoverflow.com/a/51903965/7693185

nateevans:
Yep, I already overrode this in my app, but would prefer not to maintain it for the foreseeable future. Thanks though

asturur:
I would add a step 0 before.
Verify that this is a spec, finding any reference of it somewhere (mdn, w3c )
Verify that browser do this all the same
And verify that jsdom + nodecanvas2 do that too

nateevans:
Correct @asturur it is `getAttribute()`, my bad. I will update the original comment.

nateevans:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement is what is being used and seems to be essentially the same across all browsers for this use. 

[...]

PR #5484, opened 2019-01-14
Support Relative URL for Image src

nateevans:
see https://github.com/fabricjs/fabric.js/issues/5476


src/shapes/image.class.js
```diff
@@ -349,7 +357,13 @@
         if (element.toDataURL) {
           return element.toDataURL();
         }
-        return element.src;
+
+        if (this.srcFromAttribute && typeof element.getAttribute === 'function') {
```
asturur: We support ie10 ( sort of ) ie11 and then modern browsers.
Can you verify if getAttribute is really unsupported from some of those?

asturur:
@nateevans a couple of comments:
Please branch of your master from where you are, restore your master to the top of fabric.js master, and before committing the code remember to do:

`git checkout master dist` so that you do not commit changes to the dist folder.

Can you try to add a test in the relative file in the test/unit folder?

dist/fabric.js
```diff
@@ -19479,6 +19479,14 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
      */
     strokeWidth: 0,
 
+    /**
+     * When calling {@link fabric.Image.getSrc}, return value from element src with `element.getAttribute('src')` (if available).
+     * This allows for relative urls as image src.
+     * @type Boolean
+     * @default
```
asturur: can you add:
@since 2.7.0 ?

src/shapes/image.class.js
asturur:
```suggestion
        if (this.srcFromAttribute) {
```

asturur:
@nateevans a couple of comments:
You should not commit to the dist folder.
In order to roll it back either

branch off your master from where you are, restore your master to the top of fabric.js master, and before committing the code remember to do:
`git checkout master dist` so that you do not commit changes to the dist folder.

or find the commit where dist was not changed and do
`git checkout ..commithash... dist`

asturur:
i wonder how we can properly test node. as it is now the test under node does not make much sense.

nateevans:
@asturur 
> You should not commit to the dist folder.

Ah ok. How do you run tests normally? seems like I had to build those in order to run `testem` since it runs off of `dist/`

> i wonder how we can properly test node. as it is now the test under node does not make much sense.

I don't think the use case for this actually applies to node. With JSDOM in Node, calling `element.src` doesn't seem to add any schema/hostname. Which makes sense, I don't know what it would even add there... it just seems to affect usage in browser.

nateevans:
I'll close out this PR and make one from the new branch

PR #5487, opened 2019-01-15
Support Relative URL for Image src

nateevans:
Update from https://github.com/fabricjs/fabric.js/pull/5484

See https://github.com/fabricjs/fabric.js/issues/5476


nateevans:
@asturur 
> We support ie10 ( sort of ) ie11 and then modern browsers.
> Can you verify if getAttribute is really unsupported from some of those?

I noticed `getAttribute()` is actually used all over the place in Fabric. Seems essential to the library. So it must be ok to use for this?

asturur:
looks good to me.

asturur:
i did not talk about adding this property in the toObject json export because this sounds more like a project setting that you set always in your app code, more than something you have to store and restore each time.

Does it sound right?

nateevans:
Ah, I didn't think of the `toObject` use case. We do store and restore in our app since we have "saved projects". But, we already use and rely on `propertiesToInclude` which should still cover this I think.

asturur:
But in your projects do you have images that need relative URL and images that do not?
Considering that where you do not need relative links, you pass to src all the ling and so in that case getAttribute is giving you back the full link.

asturur:
This will be probably turned default to true in fabric 4. Seems logic to be true.
Fabric 3.0 will be just a node major change requiring node 8 and no api changes

PR #5553, opened 2019-03-05
Various spelling/grammar corrections

melchiar:
First PR so let me know if I'm missing anything. 

Fixed spelling in `_finalizeDimensions` method along with various corrections within documentation. Still not too sure of the meaning of line 297 "the box is intented as aligned to axis of canvas." but corrected to "intended"

Similarly, the meaning of line 510 is unclear: "return transform Matrix for object and not go upward with parents"

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -189,7 +189,7 @@
      * @param {Fabric.Point} pointTL Top Left point
      * @param {Fabric.Point} pointBR Top Right point
      * @param {Boolean} calculate use coordinates of current position instead of .oCoords
-     * @return {Boolean} true if the objects containe the point
+     * @return {Boolean} true if the objects contains the point
      */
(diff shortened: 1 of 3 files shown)


asturur:
Line 510 means that the transform matrix calculated is the one of the object, and not the stacked effect of containers in which the object may be.

Thanks for the fixes, if you manage to rephrase line 510 even better.

melchiar:
Hmm, so I'm still getting the hang of versioning with github and I'm not sure why my commit from #5558 is referenced above (I guess I should get in the habit of separating unrelated things into different branches)

asturur:
it tells me 'unknown repository' otherwise i could give an help.

a branche per feature is like mandatory. Or everything stacks.

melchiar:
Yeah, I had previously committed from master but have since deleted and reforked the repository in order to branch things properly moving forward. Would it be better to just delete this PR and do a new one?

asturur:
yes easier. thanks for your efforts

Issue #5549, opened 2019-03-03
toObjectMethod being called by different argument types


anthify:
## Version
**EDIT**: ~2.4.0~ 2.6.0

## Details
Firstly, sorry to go off the issue format. I'm working on a biggish project which extends quite a few different classes so it might be a problem of my own making.

The `_toObjectMethod` is called recursively if `clipPath` resolves to true, but instead of `methodName` being passed which (afaik) should be a `string`, and in this instance should have the value`"toObject"`. But as you can see below it is called with whatever is assigned to `clipPath`, which in my case is a fabric.js `klass`, and this breaks the following call to `this._toObjects`>`this._toObject` when trying to locate a `methodName` on the klass, but the `methodName` is also a klass assigned to `clipPath`. 

```javascript
    /**
     * @private
     */
    _toObjectMethod: function (methodName, propertiesToInclude) {

      var clipPath = this.clipPath, data = {
        version: fabric.version,
        objects: this._toObjects(methodName, propertiesToInclude),
      };
      if (clipPath) {
        /** the line below passes clipPath into should be methodName **/
        data.clipPath = this._toObjectMethod(clipPath, methodName, propertiesToInclude);
      }
      extend(data, this.__serializeBgOverlay(methodName, propertiesToInclude));

      fabric.util.populateWithProperties(this, data, propertiesToInclude);

      return data;
    },
```

```javascript
    /**
     * @private
     */
    _toObjects: function(methodName, propertiesToInclude) {
      return this._objects.filter(function(object) {
        return !object.excludeFromExport;
      }).map(function(instance) {
        return this._toObject(instance, methodName, propertiesToInclude);
      }, this);
    },

    /**
     * @private
     */
    _toObject: function(instance, methodName, propertiesToInclude) {
      var originalValue;

      if (!this.includeDefaultValues) {
        originalValue = instance.includeDefaultValues;
        instance.includeDefaultValues = false;
      }
     
     /** it breaks here as methodName is in fact the clipPath klass **/
      var object = instance[methodName](propertiesToInclude);
      if (!this.includeDefaultValues) {
        instance.includeDefaultValues = originalValue;
      }
      return object;
    },
```

For my use I've done following, which seems to work 🤞 

```javascript
fabric.Canvas.prototype._toObjectMethod = function (methodName, propertiesToInclude) {
  var clipPath = this.clipPath, data = {
    version: fabric.version,
    objects: this._toObjects(methodName, propertiesToInclude),
  };
[...]

PR #5556, opened 2019-03-05
Fix for when toObjects is called on a canvas with a clipPath

anthify:
## Details

Regarding issue: https://github.com/fabricjs/fabric.js/issues/5549

### Bugs
* The private method `_toObjectMethod` is calling itself recursively if a clipPath is present on the root canvas.
* In addition to that, it is called with the first argument being the clipPath `klass` object itself, instead of the expected argument  `String` `methodName`, which breaks the subsequent to `toObject` method. The reason it breaks is because it's trying to find a method with `methodName`, but because it's an `object` it fails and throws an error.

### Fix
* remove the _seemingly redundant_ recursive call of `_toObjectMethod` and instead apply the clipPath to the scoped variable `data` if `clipPath` exists, and returns that instead. This results in the expected behaviour when applying a clipPath with no errors.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1177,7 +1177,7 @@
         objects: this._toObjects(methodName, propertiesToInclude),
       };
       if (clipPath) {
-        data.clipPath = this._toObjectMethod(clipPath, methodName, propertiesToInclude);
+        data.clipPath = this.clipPath;
       }
       extend(data, this.__serializeBgOverlay(methodName, propertiesToInclude));
 
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1177,7 +1177,7 @@
         objects: this._toObjects(methodName, propertiesToInclude),
       };
       if (clipPath) {
-        data.clipPath = this.clipPath;
+        data.clipPath = this._toObject(this.clipPath, methodName, propertiesToInclude);
       }
       extend(data, this.__serializeBgOverlay(methodName, propertiesToInclude));
 
--- a/test/unit/canvas.js
+++ b/test/unit/canvas.js
@@ -1297,7 +1297,42 @@
     var expectedObject = {
       'version': fabric.version,
       objects: canvasWithClipPath.getObjects(),
-      'clipPath': clipPath.toObject()
+      clipPath: {
+        type: 'rect',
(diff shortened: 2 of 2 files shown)


src/static_canvas.class.js
```diff
@@ -1177,7 +1177,7 @@
         objects: this._toObjects(methodName, propertiesToInclude),
       };
       if (clipPath) {
-        data.clipPath = this._toObjectMethod(clipPath, methodName, propertiesToInclude);
+        data.clipPath = this.clipPath;
```
asturur: wait, this should call _toOject anyway no?
In this way we are exporting the instance itself.
We want it's object representation.

test/unit/canvas.js
```diff
@@ -1290,6 +1290,25 @@
...
+      'clipPath': clipPath.toObject()
+    };
+
+    assert.ok(typeof canvasWithClipPath.toObject === 'function');
+    assert.deepEqual(expectedObject, canvasWithClipPath.toObject());
+
+    var rect = makeRect();
+    canvasWithClipPath.add(rect);
+
+    assert.equal(canvasWithClipPath.toObject().objects[0].type, rect.type);
```
asturur: to be a test that catch regressions, this last assert is not important.
you have to dump the output of a canvas `toObject()` and copy paste it in the test.
That is fixed data and we check that the code respect the fixed data ( apart for the version that is not fixed ).

So without objects, you get an empty array, and that is fine.

```
    var expectedObject = {
      'version': fabric.version,
      objects: [],
      'clipPath': { type: 'rect', width.... }
    };
```

anthify:
@asturur changes to the test have been made now

asturur:
thanks!

melchiar:
Hey guys, I just ran a quick test of this fix. While this does seem to resolve the issue with serializing a canvas with a clipPath, there appears to be a new bug when running `loadFromJSON()` on that output (see fiddle below).

Is this worth opening up a new issue?

image

http://jsfiddle.net/melchiar/68op1vLd/

asturur:
i think there is a pr open for the load fix

melchiar:
ah thanks, I missed that - https://github.com/fabricjs/fabric.js/pull/5641

PR #5558, opened 2019-03-06
Add nonScaling shadow property

melchiar:
A perfect complement to the new `strokeUniform` object property in 2.6.0 is to add a `shadowUniform` property to the shadow class to allow a shadow's `offsetX`, `offsetY`, and `blur` to be unaffected by an object's scale. 

shadowuniform


src/shapes/object.class.js
```diff
@@ -1395,9 +1395,9 @@
       }
       ctx.shadowColor = shadow.color;
       ctx.shadowBlur = shadow.blur * fabric.browserShadowBlurConstant *
-        (multX + multY) * (scaling.scaleX + scaling.scaleY) / 4;
-      ctx.shadowOffsetX = shadow.offsetX * multX * scaling.scaleX;
-      ctx.shadowOffsetY = shadow.offsetY * multY * scaling.scaleY;
+        (multX + multY) * (shadow.shadowUniform ? 1 : scaling.scaleX + scaling.scaleY) / 4;
```
asturur: scalingX + scalingY when object is not scaled, has a value of 2. i think in this case this number here should be 2.

asturur:
this would be fine if:
- we take care of object SVG export too
- we avoid calculate this.getObjectScaling if shadowUniform is true
- we give it a different name. Stroke uniform is because stroke actually get deformed by scaling, the shadow get scaled but is not `non uniform`.

how could we name this?
Shadow.scaleWithObject?
Shadow.withoutScaling?
Shadow.scaleEffect?
Shadow.noScaling?

Naming is hard.

asturur:
oh and tests. a visual test and unit tests.

melchiar:
> how could we name this?
> Shadow.scaleWithObject?
> Shadow.withoutScaling?
> Shadow.scaleEffect?
> Shadow.noScaling?
> 
> Naming is hard.

Naming is hard... I like `Shadow.noScaling` since it's precise and simple plus similar to some existing property names (eg. `centeredScaling`, `lockScalingX`, `lockUniScaling`, etc.)

src/shapes/object.class.js
melchiar: right, good catch!

melchiar:
So I ended up going with `nonScaling` if that works for you, since it's more descriptive rather than prohibitive. It also fits with the SVG attribute `non-scaling-stroke` which follows a similar principle.

I'm thinking now though that adding SVG support isn't going to be easy since the filters used to replicate shadows are subject to a matrix transform based on the parent object's scaling. Any ideas?

asturur:
is all mixed with the fixes PR :(

melchiar:
Should be fixed now. Apparently I hadn't properly reverted the files in my working directory back to master when I reforked the repo.

asturur:
well i can add svg support.
But let's add a unit test for export and a visual test in the generic_rendering.js file under test/visual/generic_rendering.

Adding a visual test is easy enough.

0) update your master to official master.
(to do so in your local git folder add my remote and give it a name you want)
`git remote add nameyouwant git@github.com:fabricjs/fabric.js.git`

then in your master:

`git pull nameyouwant master`

at this point your master is updated. if you want to update your fork copy in github too, because you prefer later to use the web interface:

`git push` from your master folder, this will send the content to your fork.

now go back in your shadow branch and do:

`git merge master` 

at this point your branch should be updated with latest master + your changes.

1) update to latest master since runnning it is easeir ( nodecanvas-2)
2) npm install
3) npm run build // to have latest fabric running

the visual test has this structure

```
  function functionwithaname(canvas, callback) {
    // canvas is a fabricCanvas ready to use
    // some drawing code with a final renderAll() you have to add.
    callback(canvas.lowerCanvasEl);
  }

  tests.push({
    test: 'test name',
    code: yourFunctionReference,
    golden: 'pngnameasyourfunction.png',
    percentage: 0.09,
    width: 150, // width of the canvas you want to use ( do not exagerate)
    height: 60, // height of the canvas you want to use ( do not exagerate)
  });
```

run the visual test with:

`npm run test:visual:single test/visual/generic_rendering.js`

at this point the missing golden is created.

Verify that looks as it should look with eyes.

then `git add .` to add the image.

the restore fabric to master

`git checkout master dist`

then commit as usual.

asturur:
image

you have some extra trailing space somewhere.
I think we are near!

thanks for trying hard.

asturur:
you did not run test locally, did you?
I'll fix them

melchiar:
only the generic rendering visual test. i can do more with some more instruction. Still figuring stuff out here :)

asturur:
if you simply run npm run test ( after a npm build ) you should see a listof tests, and some will fail.

melchiar:
k, I'm still getting 12 tests failing, but with no obvious connection to the shadow changes. 

> not ok 1187 fabric.util > fabric.util.loadImage with crossOrigin
> not ok 495 fabric.Image > getSrc
> not ok 277 fabric.StaticCanvas > toJSON overlayImage

...etc. 

Am I missing something?

asturur:
i think we are all set!

PR #5567, opened 2019-03-09
BREAKING changing objectCaching

asturur:
Change the meaning of Object.objectCaching.

ObjectCaching can now deactivate cache unless the cache is needed for clipPath or other funcionalities ( future ).


asturur:
needs a unit test to verify that works as intended.
one for Object, one for Image at least.

melchiar:
Do you think conditional caching should apply in cases where caching is preferred to be disabled, but may be required for objects with both stroke and shadow? I recently had to implement this for text since caching can cause text to appear blurry at small font sizes, but it's required to make stroke and shadows work properly together.

asturur:
Before adding cache and requestAnimationFrame, fabric was way slower.

Now with cache everything is faster. Cache blurryness has been mostly solved last year but some things cannot be solved because are part of the caching.

Said so, i agree sometimes is better not having cache, especially for final renderings where you want best possible picture, while during animations or interaction you do not really care.

You can conditionally avoid caching per object type, if you want, assigning objectCaching false only to text for example.

I also created a different rendering method that does not cache, allow for clipping, and gives best resolution.

But for now i digged it.

I think is fine to disable caching for some objects if you want maximum clarity at expense of frame time.

i was thinking of a rendering engine that can finally update just the change objects rather than full canvas, that would give the best of both world

melchiar:
For most object types caching renders the everything flawlessly. It's only with text that I've noticed a decrease in clarity in certain cases which is why I generally disable caching when possible.
image

asturur:
So for how the code is write, cached text should look perfectly fine when displayed at postion of 1px at time. If it has no stroke width for example,
Or there is a but in calculation.
If you try to set strokeWidth to 0 or to 2, with or without a stroke color, to text and leave cache on, you should have a fine result. This because with strokeWidth of 1 or 3 or 5, the text is aliased and wrote with half pixel of difference and everything is drawn across pixels.
So all the calculation i do to place the text on in the correct position are thrown off.

I cannot compensate them, because compensating would mean moving the text of 0.5px left or right and that is not right.

melchiar:
Makes sense. The issue is also practically non-existent on higher dpi displays which most devices seem to have these days so the number of people that would actually notice this is likely minimal. 

A big thanks for your amazing work on this project. 👍 I only started using fabricjs a year and a half ago and have been amazed at the frequency and quality of improvements you've put into it in that short amount of time.

Issue #5583, opened 2019-03-19
strokeUniform issue when image paintFirst: "stroke"


melchiar:
## Version
2.7.0

## Test Case
https://jsfiddle.net/melchiar/41vy30ka/

## Information about environment
tested with Chrome, Edge

## Steps to reproduce
Add an image with `strokeUniform: true` and `paintFirst: "stroke"` and the stroke experiences clipping issues at certain sizes.

## Actual Behavior
![strokeUniform bug](https://user-images.githubusercontent.com/9057412/54582271-5eda5100-49cd-11e9-8b9f-2851e2d5a86c.gif)

asturur:
Understood, this is the issue of the PR that i commented before reading this.
I think the fix should be in cache dimension calculations not in the `nonTransformedDimensions`.

```diff
    _getCacheCanvasDimensions: function() {
      var objectScale = this.getTotalObjectScaling(),
+ // probably we should use transformedDimensions here 
+ // and then multiply just by the value given by zoom and retina?
          dim = this._getNonTransformedDimensions(),
          zoomX = objectScale.scaleX,
          zoomY = objectScale.scaleY,
          width = dim.x * zoomX,
          height = dim.y * zoomY;
      return {
        // for sure this ALIASING_LIMIT is slightly crating problem
        // in situation in wich the cache canvas gets an upper limit
        width: width + ALIASING_LIMIT,
        height: height + ALIASING_LIMIT,
        zoomX: zoomX,
        zoomY: zoomY,
        x: dim.x,
        y: dim.y
      };
    },
```

asturur:
Ideally we need to add some numeric unit testing for the cacheCanvasSize ( we should already have some of them ), add some for the strokeUniform case.
Then change the code and verify that only the one with strokeUniform change results ( and insert correct results )

Whatever else we do we risk to break stuff silently.

PR #5584, opened 2019-03-20
fixed strokeUniform cache dimensions

melchiar:
The `_getNonTransformedDimensions` function used for calculating cache canvas dimensions wasn't correctly accounting for the strokeUniform property when `paintFirst: "stroke"` #5583


asturur:
I'm not sure this is right.
StrokeUniform influences transformed objects, whatever is the state of the object, its non transformed size is what it is.

What bug did you encounter that put you in need of this fix?
Maybe the fix should be somewhere else?

asturur:
Ok so we have 3 failing tests.
I m sure this is harder than it seems. I ll try to write a list of tests before you continue this, so that we know each point of logic of that function ( i also forgot about skewing is involved here )

asturur:
replaced by #5626

Issue #4981, opened 2018-05-19
Pasted text does not get the applied style


semiadam:
By design, my app does not allow rich text. Meaning that, all text styles are applied to the entire text element by applying .set() on the itext object. However, if user copy/paste some text within the itext, the applied style will not be applied to the pasted text, so I ended up having a textbox with different text styles in it (see snapshot and fiddle). 

## Version
2.3.0

## Test Case
http://jsfiddle.net/Da7SP/1647/

## Steps to reproduce
Edit text box. Select a few letters. Hit CTRL+C and CTRL+V to duplicate those letter. Click on button to change the 

## Expected Behavior
The entire  content of textbox must get the applied style.

## Actual Behavior
Pasted letter do not get the applied style.
![image](https://user-images.githubusercontent.com/5418638/40273918-cb445f7c-5b7e-11e8-8310-110b3b4cbee9.png)

semireg:
This worked for me:

`
    document.addEventListener('keydown', this.keyHandler.keydown, false);
`

then...

`
  keydown = (e: any) => {
    fabric.copiedTextStyle = null;
  }
`

melchiar:
I just encountered the same issue with my app since I'm trying to keep the styling uniform within each textbox as well. I ended up just overriding `insertNewStyleBlock` function like so to prevent the creation of style blocks.

`fabric.Textbox.prototype.insertNewStyleBlock = function() {}`

@asturur, would you be okay if I did a PR to add a config option for modifying this behavior as you suggested?

asturur:
i thought we did something like that.

asturur:
![image](https://user-images.githubusercontent.com/1194048/54864506-942dc880-4d58-11e9-8f7f-c7d20e059ce2.png)

i think we can get away with a boolean disableStyleCopying property on the Text Class so that that object has an empty object or whatever it has to have to do not carry over any style?

sponno:
Adding this line fixes the issue
`fabric.disableStyleCopyPaste = true`

PR #5590, opened 2019-03-24
added disableStyleCopyPaste property

melchiar:
Added a new property `disableStyleCopying` to the text class, as discussed in #4981. Enabling this property essentially turns of the default _Keep Source Formatting_ behavior of the existing copy paste functionality. Useful for applications where a single text style is desired for each text object.

@asturur, do you think we need any unit tests for this?

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -239,7 +239,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     }
 
     fabric.copiedText = this.getSelectedText();
-    if(!this.disableStyleCopying) {
+    if (!this.disableStyleCopying) {
       fabric.copiedTextStyle = this.getSelectionStyles(this.selectionStart, this.selectionEnd, true);
     }
     this._copyDone = true;
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -239,7 +239,9 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     }
 
     fabric.copiedText = this.getSelectedText();
-    fabric.copiedTextStyle = this.getSelectionStyles(this.selectionStart, this.selectionEnd, true);
+    if(!this.disableStyleCopying) {
+      fabric.copiedTextStyle = this.getSelectionStyles(this.selectionStart, this.selectionEnd, true);
+    }
     this._copyDone = true;
   },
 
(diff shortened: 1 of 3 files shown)


src/mixins/itext_key_behavior.mixin.js
```diff
@@ -239,7 +239,9 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     }
 
     fabric.copiedText = this.getSelectedText();
-    fabric.copiedTextStyle = this.getSelectionStyles(this.selectionStart, this.selectionEnd, true);
+    if (!this.disableStyleCopying) {
+      fabric.copiedTextStyle = this.getSelectionStyles(this.selectionStart, this.selectionEnd, true);
+    }
```
asturur: we need to assign something to this object anyway.
Imagine someone copy with style.
Then enable the property.
Then copy again. The old copied style is still there.
I think an else statement is ok or just assign it to null ( whatever is the default )  before the if.

asturur:
in test/unit/itext_key_behaviour.js line 266 we have a test for copy.
We can agument it ( or create a copy of it ) to test this new property.

src/mixins/itext_key_behavior.mixin.js
melchiar: I guess I had been thinking of this more like an application level setting, in which case the object would remain empty regardless. If you see this as a property that some users might want to set differently for different text objects then I wonder if this approach is going to be problematic since it won't prevent a box that's meant to contain only one style from still receiving style blocks from text with this option turned off. Would it perhaps be better to add a `disableStylePasting` property instead? Or maybe both?

asturur: the property is on `fabric` so is at webpage level, not even application.
The point is that then devs do what they want to do, that is a property that you can enable or disable and people will do in the way we did not think of.

So better it behaves consistently

asturur: Well we can call it `disableStyleCopyPaste` and can stop copy and pasting all together.

asturur: Wait ok, i see you added it on `this` i would expect it to be fabric.disableStyleCopyPasting. Per object does not make sense in my opinion.

melchiar: Right, yes attaching the property to fabric does make more sense (I can't imagine why anyone would need it to be per object)

asturur:
@melchiar i m merging this and adding a unit test for it.
I have been busy and i do not want drag you around for long.
Generally we want always to add one or more tests for the following reasons:

1) future contribution may break this feature
2) a test usually fix what is the intent of the code and in case of breakage explains you what you were expecting and how you should fix it.
3) peoples/companies can trust libraries with tests more.

PR #5606, opened 2019-04-01
Consistent return in stopObserving() and fire() methods.

severyanov:
Consistent return in stopObserving() and fire().
Methods are declared to be chainable and always return this reference, but in some cases return undefined.


asturur:
thanks!

asturur:
A small test would probably expand line converage a bit and ensure no one breaks it again.

asturur:
Any chance to add a small test?

asturur:
i'll add the test and merge

PR #5629, opened 2019-04-09
Hooks to allow overriding text align at a line level

SLKnutson:
For my use case, I need to be able to change text alignment on a per line basis.
This PR creates a method called per line that consumers can override to return different text alignments for different lines. It satisfies my requirement, and should satisfy most similar requirements, without drastic changes to any fabric data structures. The default implementation does not change any fabric behavior (as evidenced by none of the tests breaking).
Thanks! Let me know if there is anything else you need!

### DIFF
--- a/src/mixins/itext.svg_export.js
+++ b/src/mixins/itext.svg_export.js
@@ -112,7 +112,7 @@
     _setSVGTextLineText: function(textSpans, lineIndex, textLeftOffset, textTopOffset) {
       // set proper line offset
       var lineHeight = this.getHeightOfLine(lineIndex),
-          isJustify = this.textAlign.indexOf('justify') !== -1,
+          isJustify = this._getLineTextAlign(lineIndex).indexOf('justify') !== -1,
           actualStyle,
           nextStyle,
           charsToRender = '',
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -406,7 +406,6 @@
 
       var selectionStart = this.inCompositionMode ? this.hiddenTextarea.selectionStart : this.selectionStart,
           selectionEnd = this.inCompositionMode ? this.hiddenTextarea.selectionEnd : this.selectionEnd,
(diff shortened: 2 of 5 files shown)


asturur:
I'm neutral to this change, i think is fine.
I'll ask some people what they think of it.

asturur:
Hi @SLKnutson i talked about this with some devs, and i came to a conclusion.
Either we make it a full feature or we do not do it.

I prefer we create an array of textAlings ( something like `this.linesTextAlign` or name to be defined ) and use the getter to retrieve that at specific line or return the default one `textAlign` if the one per line is not set.

We can discuss it further, but i prefer to have a full feature rather than an hook for who knows how to use it

SLKnutson:
I assume the text aligns would correspond to the lines the user passes in, not the wrapped lines? I think using similar logic to how the correct style is found, it should be simple enough to figure out which text align to use.
There is a lot of complexity that comes with creating a property for it. You can probably think of more things because you are more familiar, but off the top of my head we would have to:
1. Possibly combine text aligns when text is removed.
2. Possibly add a new text align when text is added (assuming we would copy the text align of the paragraph in which the additional text started?)
Do you know of anything else that would need to change?

asturur:
well how were you going to make it work?
You defined an hook, but what were you doing in your app?

There are no parapgraph in fabricJS, and in word editing a per-line textaling does not exist and i do not want to invent it.

So the closest thing we have to paragraph are hard break lines.

So 10 wrapped lines are 1 line with 1 text alignment possible.

For the thing of combining lines when text is added /removed we could also not do it.

SLKnutson:
To avoid changing too much core fabric.js, in my app I shoved alignment onto each style of each character. Then in the hook, I return whatever the alignment is of the first character of that wrapped line.
Then the trick is to make sure that whenever the text changes, apply the alignment of the first character of each line to all of the subsequent characters of that line (where line is hard line break separated line).

PR #5632, opened 2019-04-10
Fixing onSelect being called when objects were not intersected

### LINKED ISSUE #5596: onSelect called when the object is not selected.
## Version
2.6.0

## Test Case
http://jsfiddle.net/rpfilipe/rjx23sek/4/

Chrome, FIrefox

## Steps to reproduce
Perform a selection by dragging the mouse that doesn't intersect the object.

## Expected Behavior
onSelect() should not be called.

## Actual Behavior
onSelect() is caled

## Where's the problem
The onSelect() is called in the __collectObjects() method.

On the [docs](http://fabricjs.com/docs/fabric.Object.html#onSelect) is stated that "This callback function is called every time _discardActiveObject or _setActiveObject try to to select this object".

Full Call Stack
__collectObjects() : fabric.js:12024
_groupSelectedObjects(): fabric.js:12015
_maybeGroupObjects(): fabric.js:12074
__onMouseUp(): fabric.js:11290
_onMouseUp(); fabric.js:11187

rpFilipe:
PR fixing the bug stated on the issue #5596

The onSelect method is only called after we determine that the multiselect intersects the object. If the onSelect returns 'false' we discard them from the group.

close #5596

### DIFF
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -139,8 +139,7 @@
             (allowIntersect && currentObject.containsPoint(selectionX1Y1)) ||
             (allowIntersect && currentObject.containsPoint(selectionX2Y2))
         ) {
-          
-            group.push(currentObject);
+          group.push(currentObject);
           // only add one object if it's a click
           if (isClick) {
             break;
@@ -148,7 +147,7 @@
         }
       }
 
-      if(group.length > 1) {
+      if (group.length > 1) {
         group = group.filter(function(object) {
           return !object.onSelect({ e: e });
         });
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -149,7 +149,9 @@
       }
 
       if(group.length > 1) {
-        group = group.filter((object) => !object.onSelect({ e: e }));
+        group = group.filter(function(object) {
+          return !object.onSelect({ e: e });
+        });
       }
 
       return group;
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -139,17 +139,17 @@
             (allowIntersect && currentObject.containsPoint(selectionX1Y1)) ||
             (allowIntersect && currentObject.containsPoint(selectionX2Y2))
         ) {
-          if(currentObject.onSelect({ e: e })) {
+          
             group.push(currentObject);
-          } else {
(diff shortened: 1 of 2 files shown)


src/mixins/canvas_grouping.mixin.js
```diff
@@ -140,14 +140,19 @@
...
           // only add one object if it's a click
           if (isClick) {
             break;
           }
         }
       }
 
+      if (group.length > 1) {
+        group = group.filter(function(object) {
+          return !object.onSelect({ e: e });
```
asturur: what about wrapping `group.push(currentObject)` in the if with this condition?

```
if (!object.onSelect({ e: e })) {
  group.push(currentObject);
}
```

so we avoid the other if and the filter operation, just because is 5 lines extra not because i do not like it.

asturur:
Do not forget we need to add a unit test for this change.

If you can't run them locally or you do not know how to do, please ask. I can help

src/mixins/canvas_grouping.mixin.js
rpFilipe: That was my initial implementation. However, if the selection only intersected one object the _onSelect()_ method would be called 2 times. one in _this method_ and another on  __setActiveObject()_. That requires changing the __setActiveObject()_ to have a flag to not call the _onSelect()_ if we are 'coming' from the __onCollectObjects()_. I was reluctant to change as it could impact many operations. However, if you like that aproach better let me know and I'll do a refactoring

rpFilipe:
I'll see the implemented tests and try to create a test for this change. I'll probaly need yout help :) I'll let you know.

src/mixins/canvas_grouping.mixin.js
asturur: no, at this point this solution is better.

We still need tests. Can you write them?

rpFilipe: yes, i wasn't able this weekend, but I'll work on it this week. I'll keep you updated.

rpFilipe: @asturur Already implemented the tests. If you need anything more just let me know

asturur:
i think is ok!
I ll test it better as soon as i can, but i think we can merge it.
The code changes looks fine and the test cover enough of it.

asturur:
This will be in the bundle of fabricJS 3.0
I wanted to solve the clipPath blurryness with that release, but since is taking too long i think i ll move on with a release anyway.

PR #5634, opened 2019-04-12
Exposing methods for typescript react related to canvas

teddyseyed:
The DefinitelyTyped version of FabricJS for react doesn't have the ability to modify anything related to freedrawing and the brush, so I exposed those methods to do so.

Exposed methods for:

1. Setting Brush color
2. Setting Brush width
3. Setting Brush shadow
4. Setting Brush shadow blur
5. Setting Brush shadow x offset
6. Setting Brush shadow y offset
7. Setting the Brush to be a pencil brush
8. Setting the Brush to be a circle brush
9. Setting the Brush to be a spray brush


asturur:
Why do we want to create methods for things that are direct access to properties ? Is there any reason to do so?

vleandersson:
Couldn't the same result be achieved by updating the @types/fabric typings? If so, shouldn't we aim to update those instead?

PR #5637, opened 2019-04-14
Avoid finding target if we are transforming

### LINKED ISSUE #5504: Lose `__corner` property when you resize the object outside the canvas area
## Version
2.4.0

## Test Case
http://jsfiddle.net/nLaxumbj/

Nodejs or browser?
Which browsers?
All major browsers
## Steps to reproduce
1.Open the console.
2. Select the rectangle and resize it from 'mr' corner until the rectangle pass the canvas area.
3. When you come with the cursor back into the canvas area the `__corner` property is lost.
I think the problem comes from `_onMouseEnter` callback. This function calls `findTarget` and from here 
it goes to `_findTargetCorner` where no __corner is found.
I guess the fix is to check the `_currentTransform` property and if is different from null we just trigger the `mouse:over` event
## Expected Behavior
The __corner property must remain the same
## Actual Behavior
No `__corner` property when you resize the rectangle outside the canvas

asturur:
close #5504

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -163,7 +163,13 @@
      * @param {Event} e Event object fired on mouseenter
      */
     _onMouseEnter: function(e) {
-      if (!this.findTarget(e)) {
+      // This find target and consequent 'mouse:over' is used to
+      // clear old instances on hovered target.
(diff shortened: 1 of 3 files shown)


asturur:
1 boolean changed, 43 lines of tests and 15 of docs...

PR #5642, opened 2019-04-18
fix load SVG with the broken image

AlexSergey:
If we try to load SVG with the broken image (ERR_CONNECTION_TIMED_OUT, not 404) it will be error

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -714,7 +714,7 @@
    */
   fabric.Image.fromURL = function(url, callback, imgOptions) {
     fabric.util.loadImage(url, function(img) {
-      callback && callback(new fabric.Image(img, imgOptions));
+      callback && callback(!img ? null : new fabric.Image(img, imgOptions));
     }, null, imgOptions && imgOptions.crossOrigin);
   };
 


asturur:
Images with null element should be possible imho.
What is the output now, can you describe better?

AlexSergey:
If we set null image to fabric.Image it will be error -
http://joxi.ru/D2Pzl65CqJqqlr
I think we need to avoid null images

asturur:
i think we should have a null image definied as an image that can live in fabricJS but with 0 dimensions and that does not throw errors.

Do you have a failing SVG that gives us a failed image?

AlexSergey:
Hmmm... I can't reproduce it from the original fabric.js version. Maybe it happens only our modified fabric.js

asturur:
i remember we fixed it.... but since i said the same thing and i was wrong, i did not say that this time.

asturur:
what is the situation with this bug? can you confirm if the bug exist or not?

AlexSergey:
I investigated this task and I think this happened only my project. We can close it and this won't be merge to fabric.js.

PR #5654, opened 2019-04-27
Swap NODE11 with NODE12

asturur:
add latest version of node latest.


asturur:
the tests fail in travis, but run fine locally on mac os

Issue #5677, opened 2019-05-08
Bundled "fabric.js" doesn't  work on IE10


junghwan-park:
## Version
2.7.0

## Test Case
```
Test case not needed.
```

## Information about environment
Web Browser / IE10

## Steps to reproduce

1. Execute IE10 browser in Windows OS.
2. Load web page that loads bundled [fabric.js](https://cdnjs.cloudflare.com/ajax/libs/fabric.js/2.7.0/fabric.js).
3. Check debugger window console tab.

## Expected Behavior

There should be no errors. Script parsed successfully, and ready to use fabric.

## Actual Behavior

Browser hasn't `HTMLDocument` constructor but bundled file header contains `HTMLDocument`. I can't use `fabric` in IE10 at all.

asturur:
ie10 should be supported, i'll check the issue.
The problem is that i do not have a  ie10 computer anymore. I need to setup a vm.

junghwan-park:
@asturur
As far as I know, `HTMLDocument` constructor not exists in IE9, IE10. But, It has non HTML5 specification constructor `Document`. But modern browsers has `Document` too.

I think [HEADER.js#L13](https://github.com/fabricjs/fabric.js/blob/765f11537b5002ffe5b556f47ccccbfb8428d022/HEADER.js#L13)'s condition statement should be check `document` is instance of `Document` constructor when `HTMLDocument` constructor is `undefined`.

Issue #5530, opened 2019-02-20
Support Fabric in XML documents.


ajvincent:
Fabric.js assumes that it runs inside a HTML document.  Which is usually not a bad assumption, but it does mean if I try to load Fabric in a XML document, it just breaks all over the place.  In particular, when referring to properties of element.style, it throws exceptions because element is an instance of [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element), not [HTMLElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement).  The latter has the style property; the former does not.

Fortunately the fix is simple:  in HEADER.js, I use `fabric.document = document.implementation.createHTMLDocument("");` to make this work (in the case where document isn't a HTMLDocument).  I will provide a pull request momentarily.

I can provide a testcase upon request, written in SVG and using the `<foreignObject>` element.  (I know, I know, why use FabricJS and Canvas if I have SVG?  Well, I have my reasons - specifically, being able to capture a video stream in the browser from the canvas.)  In general, though, I think the pull request I will generate will work with any XML document.

ajvincent:
[editor.zip](https://github.com/fabricjs/fabric.js/files/2882671/editor.zip)
This has the SVG file and a FabricController.js, and assumes the existence of libraries/fabric/fabric.js .

asturur:
i ll try the editor.zip, understand the use case and probably merge.

cnsdato:
This breaks IE10 and IE9: both do not support HTMLDocument.
Now I have to go back to 2.6.0

asturur:
ie9 is no more supported since 2.0
Ie10 there is a PR to restore functionality, if you would be able to test it, i can merge it.

asturur:
https://github.com/fabricjs/fabric.js/pull/5678

PR #5678, opened 2019-05-09
fix: Check HTMLDocument constructor exists first for IE10

junghwan-park:
## AS-IS
- IE10 is supported browser. But `HTMLDocument` constructor is not exists in IE10.
- `fabric.js` bundle file using `HTMLDocument` constructor for check `document` is instance of it.

## TO-BE
- Use `Document` constructor instead of `HTMLDocument` constructor if `HTMLDocument` is not exists in browser global variable.

## Resolution
- Check `HTMLDocument` exists first.
  - If not, Use `Document` for check `document` is instance of it.

ref: #5677

### DIFF
--- a/HEADER.js
+++ b/HEADER.js
@@ -10,9 +10,7 @@ else if (typeof define === 'function' && define.amd) {
 }
 /* _AMD_END_ */
 if (typeof document !== 'undefined' && typeof window !== 'undefined') {
-  var DocumentConstructor = typeof HTMLDocument !== 'undefined' ? HTMLDocument : Document;
-
-  if (document instanceof DocumentConstructor) {
+  if (document instanceof (typeof HTMLDocument !== 'undefined' ? HTMLDocument : Document)) {
     fabric.document = document;
   }
   else {
--- a/HEADER.js
+++ b/HEADER.js
@@ -10,7 +10,9 @@ else if (typeof define === 'function' && define.amd) {
 }
 /* _AMD_END_ */
 if (typeof document !== 'undefined' && typeof window !== 'undefined') {
-  if (document instanceof HTMLDocument) {
+  var DocumentConstructor = typeof HTMLDocument !== 'undefined' ? HTMLDocument : Document;
+
+  if (document instanceof DocumentConstructor) {
     fabric.document = document;
   }
   else {


asturur:
@ajvincent can you verify this change does not kill XML document support please? i do not have time for it.

asturur:
@junghwan-park thanks ot be proactive with the fix

ajvincent:
Apologies for not responding yet:  I have been rather busy myself.  I have this ticket pinned in my browser to remind me to test it, but on the surface it looks fine.

cnsdato:
@asturur I have build @junghwan-park branch (https://github.com/junghwan-park/fabric.js/tree/fix-ie10-support) and it looks fine, so I will wait until this is merged.

ajvincent:
The proposed change does pass the editor.zip test I posted in #5530, but on second look at the patch included, I have a minor concern.

Specifically, the code adds a top-level var DocumentConstructor.  Think about namespace pollution, or just being nice to neighboring code in the same scope.

If this block were enclosed in a self-executing function which then defined the necessary properties, I'd be fine with this.  That said, be careful how much you enclose in such a function:  `DOMParser = require('xmldom').DOMParser; ` is also in the HEADER.js file in the node.js path, and turns up much later.

Consider this a peanut-gallery code review: if @asturur wants to accept the patch as-is, or require a change similar to what I suggest, that is his decision.

HEADER.js
```diff
@@ -10,7 +10,9 @@ else if (typeof define === 'function' && define.amd) {
 }
 /* _AMD_END_ */
 if (typeof document !== 'undefined' && typeof window !== 'undefined') {
-  if (document instanceof HTMLDocument) {
+  var DocumentConstructor = typeof HTMLDocument !== 'undefined' ? HTMLDocument : Document;
+
+  if (document instanceof DocumentConstructor) {
```
asturur:
```suggestion
  if (document instanceof (typeof HTMLDocument !== 'undefined' ? HTMLDocument : Document)) {
```

asturur:
Your review is what i was looking for actually. I did not notice/remember the header wasn't wrapped in an anti-leak function wrapping, so that `VAR` cannot be there at all.
Thanks for spotting this.

HEADER.js
asturur: Another solution is to wrap all the file in a self executed function.

asturur:
@junghwan-park sorry for the delay, can you look at the comment/discussion?

thanks for opening this!

HEADER.js
junghwan-park: Okay, I'll use that ternary operation. Thanks.

junghwan-park:
@asturur I've changed `var` definition into ternary operation.

asturur:
Guys thanks for collaboration

junghwan-park:
@asturur @ajvincent  Thanks.
I'll looking forward to this feature releasing. :)

asturur:
i ll merge another pr and release 3.0.1

PR #5699, opened 2019-05-20
fix(canvas): prevent issues with multitouch events

dirkluijk:
This PR fixes issues with multi-touch events and free drawing.

When you are using free drawing, and you are drawing a path, it "vanishes" when you (accidentally) touch the screen with a second pointer (e.g. other finger).

This PR makes Fabric JS ignore additional `touchstart` and `touchmove` events from other fingers.

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -244,11 +244,11 @@
      */
     _onMouseDown: function (e) {
       // ignore events from other pointers
-      if (this._pointerId && this._pointerId !== this._getPointerId(e)) {
+      if (this._pointerId !== undefined && this._pointerId !== this._getPointerId(e)) {
         return;
       }
 
-      this._cachePointerId(e);
+      this._rememberPointerId(e);
       this.__onMouseDown(e);
       this._resetTransformEventData();
       addListener(fabric.document, 'touchend', this._onMouseUp, addEventOptions);
@@ -275,7 +275,7 @@
      */
     _onMouseUp: function (e) {
       // ignore events from other pointers
-      if (this._pointerId && this._pointerId !== this._getPointerId(e)) {
+      if (this._pointerId !== this._getPointerId(e)) {
         return;
       }
 
@@ -310,7 +310,7 @@
      */
     _onMouseMove: function (e) {
       // ignore events from other pointers
-      if (this._pointerId && this._pointerId !== this._getPointerId(e)) {
+      if (this._pointerId !== this._getPointerId(e)) {
         return;
       }
 
@@ -661,7 +661,15 @@
      * @param {Event} event Event object
      */
     _getPointerId: function(event) {
-      return this.enablePointerEvents ? event.pointerId : 1;
+      if (this.enablePointerEvents) {
+        return event.pointerId;
+      }
+
+      if (event.changedTouches) {
+        return event.changedTouches[0].identifier;
+      }
+
+      return 0;
     },
 
     /**
@@ -678,7 +686,7 @@
      * @private
      */
     _resetPointerId: function() {
-      this._pointerId = null;
+      this._pointerId = undefined;
     },
 
     /**
@@ -698,7 +706,7 @@
      * @private
      * @param {Event} event Event object fired on event
      */
-    _cachePointerId: function(event) {
+    _rememberPointerId: function(event) {
       this._pointerId = this._getPointerId(event);
     },
 
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -243,6 +243,12 @@
      * @param {Event} e Event object fired on mousedown
      */
     _onMouseDown: function (e) {
+      // ignore events from other pointers
+      if (this._pointerId && this._pointerId !== this._getPointerId(e)) {
+        return;
+      }
+
+      this._cachePointerId(e);
       this.__onMouseDown(e);
       this._resetTransformEventData();
       addListener(fabric.document, 'touchend', this._onMouseUp, addEventOptions);
@@ -268,8 +274,14 @@
      * @param {Event} e Event object fired on mouseup
      */
     _onMouseUp: function (e) {
+      // ignore events from other pointers
+      if (this._pointerId && this._pointerId !== this._getPointerId(e)) {
+        return;
+      }
+
       this.__onMouseUp(e);
       this._resetTransformEventData();
+      this._resetPointerId();
       var canvasElement = this.upperCanvasEl,
           eventTypePrefix = this.enablePointerEvents ? 'pointer' : 'mouse';
 
@@ -297,6 +309,11 @@
      * @param {Event} e Event object fired on mousemove
      */
     _onMouseMove: function (e) {
+      // ignore events from other pointers
+      if (this._pointerId && this._pointerId !== this._getPointerId(e)) {
+        return;
+      }
+
       !this.allowTouchScrolling && e.preventDefault && e.preventDefault();
       this.__onMouseMove(e);
     },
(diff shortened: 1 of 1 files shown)


dirkluijk:
I could also add some unit tests, if that is required. In which spec file should I add such a test?

asturur:
Unit tests are required and hard to write for events.
an acceptance test framework is missing here.

I was wondering if this is the correct solution.
This solution basically stops you from implementing any multi touch drawing.
What is breaking the drawing? how can we make so that we do not return from the event but indeed we do not make the brush trigger the problem?

Can we dig a little bit deeper in what is going on?

asturur:
And first of all, thanks for opening a pr to fix a problem.

dirkluijk:
Sorry for being a bit unclear about the issues I encountered. ;)

```
I was wondering if this is the correct solution.
This solution basically stops you from implementing any multi touch drawing.
What is breaking the drawing? how can we make so that we do not return from the event but indeed we do not make the brush trigger the problem?
```
Okay, first of all: I would expect Fabric.js so support multi-touch in a proper way. I guess that would mean that you can draw multiple lines with multiple fingers. In `PointerEvent` API you could make use of [`PointerEvent.pointerId`](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerId). However, Fabric.js is based on `TouchEvent` API which makes it a bit harder. You would need to do something with the `TouchEvent.touches` array and [`Touch​.identifier`](https://developer.mozilla.org/en-US/docs/Web/API/Touch/identifier) of the touch objects.

However, that would be a lot of work. I feel a bit unable to file a PR for that at this moment.

Currently, when you use multiple fingers (which can easily happen accidentally!), things really gets messed up.

1) A large portion of the (already drawn) path just vanishes, because a new "touchstart" has been fired by your second pointer device (e.g. other finger). A lot of our users has been complaining about drawn figures disappearing during the drawing operation.
2) When the second pointer device (e.g. other finger) stops touching the screen, the drawing operation is being finished, although the initial finger is still drawing. This is annoying as well.

Both issues can be simply solved by ignoring additional touch events (which is done by this PR).

A better improvement would be (imho) to move Fabric.js to PointerEvent API. It would be a huge improvement for the Fabric.js codebase because we don't need to distinguish between Mouse Event API and Touch Event API. There is a very good polyfill available for Apple devices. The PointerEvent API would also make proper multi-touch support for Fabric much easier. 😄 

I hope this clarifies my PR. ;)

asturur:
so we got pointer-events last week.
But we do not have any gesture related code yet.
I still belive that we should work on a different fix, instead of returning, move some of this code in the base-brush class that will let it continue to work rather than starting a new brush or closing one?

Using this code, with a note `hey this code is a patch because at this point in time the feature X was an annoyance but we did not have time to build the correct solution` is not good for a library. Because eventually will be there 3 years and then context of what was not good and what we would like to have get lost.

I would be happy to help to move this fix in the base brush with an option, default as false, `enableMultiTouch` that will obviously not work, but will make it possible to continue to draw with one finger while other fingers touch the screen.

dirkluijk:
```
so we got pointer-events last week.
```
Oops, didn't know that. Is there a PR / branch?

In that case, I would suggest remembering the `PointerEvent.pointerId` of the current touch operation, and ignore all events that do not match that identifier (or handle them correctly).

What about that?

asturur:
For me is ok, as long as it work in conjuction with mouse events.
( was more than a week i see now)
https://github.com/fabricjs/fabric.js/pull/5589 <--- this is the PR.

As long has it does not break with mouseEvents, is good.

dirkluijk:
OK, I'll upgrade to the pointer events support and investigate the issue (and possible solution) again.

dirkluijk:
@asturur I just pushed a new fix, based on pointer events. It is backwards compatible, when pointer events are not enabled, every mouse/touch event gets the same id (`1`).

dirkluijk:
@asturur I just pushed a new fix, based on pointer events. It is backwards compatible, when pointer events are not enabled, every mouse/touch event gets the same id (1).

dirkluijk:
@asturur Can you review this fix? 😄 Please let me know if you miss something.

asturur:
Sorry @dirkluijk did not have time to go over. I would like to understand if we can move the fix inside the brush rather than in the event system. I ll reserve some time today to look at it.

src/mixins/canvas_events.mixin.js
```diff
@@ -639,6 +656,14 @@
       (shouldRender || shouldGroup) && this.requestRenderAll();
     },
 
+    /**
+     * @private
+     * @param {Event} event Event object
+     */
+    _getPointerId: function(event) {
+      return this.enablePointerEvents ? event.pointerId : 1;
+    },
```
asturur: what is the equivalent version on touch  events for determine wich touch in multi touch are we using? any idea?
i would prefer something like:
```js
return event.pointerId || 1;
```

unless pointerId can be 0.

src/mixins/canvas_events.mixin.js
```diff
@@ -662,6 +694,14 @@
       this._target = this._currentTransform ? this._currentTransform.target : this.findTarget(e) || null;
     },
 
+    /**
+     * @private
+     * @param {Event} event Event object fired on event
+     */
+    _cachePointerId: function(event) {
+      this._pointerId = this._getPointerId(event);
+    },
```
asturur: let's avoid the caching logic, extracing a value from an event is not worth the extra methods.

asturur:
@dirkluijk i have some questions and i think i prefer a different solution.
I would like to leave in the canvas class only the getPointerId method.

In all the related events of freeDraw where we pass down the object `{ e, pointer }` pass `{ e, pointer, pointerId }` in this way we move the fix in the pencilBrush.

In the pencilBrush class on mousedown we memorize the pointerId passed, and we execute code in mousemove and mouseup inside the brush only if the pointerId of the following events matches.

In this way we can fix the brush and leave the canvas events brush-problems-unaware

What do you think?

dirkluijk:
@asturur seems like a good idea. I'll try to implement a new fix today.

asturur:
We can release a 3.2.1 when you are done. No rush

dirkluijk:
@asturur I tried implementing it in the brush system.

I had 2 issues when attempting to implement the fix in the brush system:
*  There is currently no mechanism in the `BaseBrush` class to add the pointer tracing, so that would require changes in all the current brushes (not backwards compatible?)
* The canvas event system does a `removeEventListener()` on `touchend`/`mouseup`/`pointerup` in the `_onMouseUp()` function. This means that preventing secondary pointer behaviour from inside the Brush system does not work, because the "pointer up" of the first pointer will not fire if a second pointer has been used in the mean time.

My suggestion:
Please reconsider accepting the fix in the canvas event system. I believe this is the best location for it and won't break anything else. 😄 

I just added a new commit which makes the fix also work for touch events. So it now works for both touch and pointer events.

src/mixins/canvas_events.mixin.js
dirkluijk: Fixed!

For touch events, `event.changedTouches[0].identifier` is the equivalent of `event.pointerId`.  For mouse events, no secondary pointers are relevant, so I use `0` as fallback value.

src/mixins/canvas_events.mixin.js
dirkluijk: It is needed to remember the pointerId. I renamed the method to make it more clear.

dirkluijk:
Please see https://stackblitz.com/edit/fabric-fix for a simple example of the fix.

asturur:
ok so fabric wants to track movements and mouse up out of the canvas when you draw or drag. That is dirty but understandable.

It does it registering document events when you mouse down.

Then on mouse up it wants to remove those from the document.

i would like to get keep canvas as much as brush neutral possible.
Can we just count how many pointers are active and avoid doing usual stuff?

let me throw out a prototype now.

asturur:
I started a counter proposal in #5785, please have a look

asturur:
replaced by #5785

Please @dirkluijk let me know how does it work for you, and let's keep improving it.

PR #5714, opened 2019-05-23
could we add validation on fabric.Image.fromObject

pimpin:
This is an idea to be discussed and hopefully developped ahead to become a mergeable PR.

Our use case is server side (NodeJs) image building from stored json data.
For some reason, fabricJs silently fail to build expected image.
I found some of those potentials reasons.

The purpose of this PR is to

1. check validity of object before building image
2. handle unvalid cases to give feedback to dev

Is this ok for you @asturur ? Shall I go ehead with this PR ?

For point 1 : could you help listing most recurring error/mis used people get you know from issues.
For point 2: I guess we should throw errors and cath them else where... whats your opinion ?

### DIFF
--- a/test/unit/image.js
+++ b/test/unit/image.js
@@ -437,6 +437,13 @@
     });
   });
 
+  QUnit.skip('fromOject validates object or throws error (before doing anything else)'), function(assert) {
+    // check src attribute presence
+    // validated src attribute is a string
+    // validate src attribute is a valid url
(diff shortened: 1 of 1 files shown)


asturur:
well, what we can do is we can either provide an additional option in loadFromJSON ( from Object ) to accept a validation function.

I would not take care of the possible validation functions that people may want to implement, but i m ok to provide an hook to validate an object.

But in case you won't find an object valid, what would you do? Stop loading anything? Stop loading that single object?

asturur:
Let's open an issue here: https://github.com/fabricjs/api-discussion/issues to see if someone else join the topic, and the we move forward with an implementation.

PR #5731, opened 2019-05-31
Fix parsing number values using scientific notation w/ a capital E.

rodovich:
Fixes parsing numbers like `1.2E3` → 1200 in `fabric.reNum`.

Per https://www.w3.org/TR/css-syntax-3/#number-token-diagram, either lowercase `e` or uppercase `E` can be used in scientific notation numbers. (That page is called "CSS Syntax" but the SVG spec at https://www.w3.org/TR/SVG/types.html#InterfaceSVGNumber delegates to that section of the CSS spec.)

I updated the existing tests for parsing `matrix` and `points`. `points` already worked fine because it uses `parseFloat` and not `fabric.reNum`, but the `matrix` test failed before fixing the regular expression.

### DIFF
--- a/HEADER.js
+++ b/HEADER.js
@@ -72,7 +72,7 @@ fabric.SHARED_ATTRIBUTES = [
  * Pixel per Inch as a default value set to 96. Can be changed for more realistic conversion.
  */
 fabric.DPI = 96;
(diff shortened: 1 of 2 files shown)


asturur:
i m ok with this PR.
I wonder if svg 1.1 or 1.2 should be considered stopped at the CSS specs that were available those years.

asturur:
thanks!

PR #5746, opened 2019-06-08
IN PROGRESS - enhance retina upper canvas, move controls on the upper canvas

asturur:


melchiar:
Since controls will be moving to the upper canvas, would you be open to supporting an option that would allow the upper canvas to have larger dimensions than the lower canvas? (ie. to allow  resizing/manipulating objects that are outside of the main canvas's bounds)

Although you can achieve this type of setup by setting the canvas larger than needed and then clipping it, it's a bit of a messy solution.

asturur:
@melchiar your comment/request seems related to the often asked bleed out of canvas objects.
In case i m wrong ignore me.

If not, i think i should make an example. I think that with an overlay and a background, that effect is completely achievable.

Issue #5739, opened 2019-06-04
toSVG() losing path opacity


mediumhust:
## Version
3.1.0

## Test Case
https://jsfiddle.net/tinyhustlee/v8dyb2wj/3/

## Information about environment
Which browsers?

## Steps to reproduce
Load SVG from URL, set object opacity 0.3. Then use the function toSVG() to export canvas to svg ( click the button To SVG in jsfiddle ).
## Expected Behavior
Exported SVG has the same opacity with canvas
## Actual Behavior
Exported SVG losing opacity

asturur:
yes is a bug!

PR #5755, opened 2019-06-15
Fix Group.toSVG export missing opacity and visibility

asturur:
close issue #5739


asturur:
Bug is fixed, but requires writing a new test and also fixing tests that changed because of the changed code.
I hope to manage it in the weekend

asturur:
Needs a test case for group with multiple layers of opacity and shado and shadow on 1 item only

PR #5779, opened 2019-06-26
Rm reserved keyword char

### LINKED ISSUE #5778: Compression with Yui Compressor broken
## Version
at least from 3.0.0 and until 3.2.0

similar to #2934 

Compression broken by usage of reserverd "char" variable name  in `getHeightOfChar` here : https://github.com/fabricjs/fabric.js/blob/v3.0.0/dist/fabric.js#L25206

You can test it here : https://fr.piliapp.com/minify/yui-compressor/

pimpin:
Should fix #5778

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -25203,8 +25203,8 @@ fabric.Image.filters.BaseFilter.fromObject = function(object, callback) {
      * @param {Number} char the character number
      * @return {Number} fontSize of the character
      */
-    getHeightOfChar: function(line, char) {
-      return this.getValueOfPropertyAt(line, char, 'fontSize');
+    getHeightOfChar: function(line, _char) {
+      return this.getValueOfPropertyAt(line, _char, 'fontSize');
     },
 
     /**
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -683,8 +683,8 @@
      * @param {Number} char the character number
      * @return {Number} fontSize of the character
      */
-    getHeightOfChar: function(line, char) {
-      return this.getValueOfPropertyAt(line, char, 'fontSize');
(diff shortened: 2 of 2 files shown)


pimpin:
Notice I did not actualy build the dist file but manually fix it.
The minified file is not fixed. But both ot them should be fixed on next commited built.
And we have to backport fix into master branch.

asturur:
Hi @pimpin please open a PR only on the master branch, without committing to the dist file at all.
I will release then a new version with the fix applied

dist/fabric.js
```diff
@@ -25203,8 +25203,8 @@ fabric.Image.filters.BaseFilter.fromObject = function(object, callback) {
      * @param {Number} char the character number
      * @return {Number} fontSize of the character
      */
-    getHeightOfChar: function(line, char) {
-      return this.getValueOfPropertyAt(line, char, 'fontSize');
+    getHeightOfChar: function(line, _char) {
+      return this.getValueOfPropertyAt(line, _char, 'fontSize');
```
asturur: should also fix the JSDOC 3 lines above

pimpin: ok. LEts do that. Will you backport it and build all of 3.0.1 3.1.1 and 3.2.1 ?

asturur:
i will put in 3.2.1 and if you really need i can do a one off backport and publish, but if ypir app works with 3.0 will work with 3.2.1 too

pimpin:
Actually we are not sure it's working well with 3.0.0. We are upgrading from 1.5.0 and we did not complete all our checkup yet.
Currently I've patched the source file but we might bump to 3.2.1.

asturur:
1.5 wow..... my first big release as contributor. Fabric changed a lot.

PR #5785, opened 2019-06-29
Differently support multi mouse events

asturur:
This PR is a proof of concepts for handling multi touch / multi pointer events
It aims at taking track of all events that gets generated on the canvas and identify the owner of the transformation, while letting everything else flow to eventually brushes and events firing.

- Removes old ie < 10 code to handle events ( and so not supported )
- Register specific events for touch based
- Put a guard against double re-registering of mouse down


src/mixins/canvas_events.mixin.js
```diff
@@ -16,7 +16,7 @@
       addEventOptions = { passive: false };
 
   function checkClick(e, value) {
-    return 'which' in e ? e.which === value : e.button === value - 1;
+    return e.button === value - 1;
```
asturur: this was something pre ie10. We do not need it anymore

src/mixins/canvas_events.mixin.js
```diff
@@ -44,14 +60,23 @@
...
       this.addOrRemove(addListener, 'add');
     },
 
+    /**
+     * return an event prefix pointer or mouse.
+     * @private
+     */
+    _getEventPrefix: function () {
+      return this.enablePointerEvents ? 'pointer' : 'mouse';
+    },
```
asturur: this is just to allow the minifier to work better and reuse code.

src/mixins/canvas_events.mixin.js
```diff
@@ -238,21 +263,80 @@
...
+     * @param {Event} event Event object
+     */
+    getPointerId: function(event) {
+      if (this.enablePointerEvents) {
+        return event.pointerId;
+      }
+
+      var changedTouches = event.changedTouches;
+      if (changedTouches) {
+        return changedTouches[0] && changedTouches[0].identifier;
```
asturur: need to verify what changeTouches contains and in what order.

asturur:
@dirkluijk @ arcatdmz
I'm trying to get a grasp of the pointer event situation.

Today i digged up my pointer event device and formatted/reinstalled it ( forgot password ) so that i can move on with this.

We need to understand how we can:
Uniform pointer/touch/mouse events in a way that up/down make sense in the same way for every device ( a mouse generate down/up with buttons and PER button , while pointer do not do it, but they generate a move event with different button state )

Define if we want hardcode in fabric that object transformations are done with left mouse button and with no buttons for touch/pointer events ( ok fingers do not have button ).

How touch events and pointer events fire things with multiple finger/pencils that move or goes up ath the same time.

if i have 2 fingers down on the screen and i pull up one, do i get a mouse up events? I did not try yet.

src/mixins/canvas_events.mixin.js
```diff
@@ -268,10 +352,11 @@
      * @param {Event} e Event object fired on mouseup
      */
     _onMouseUp: function (e) {
+      this._removePointer(e);
       this.__onMouseUp(e);
       this._resetTransformEventData();
       var canvasElement = this.upperCanvasEl,
-          eventTypePrefix = this.enablePointerEvents ? 'pointer' : 'mouse';
+          eventTypePrefix = this._getEventPrefix();
```
asturur: we want to do the next 15 lines of code, only when we are finished with pointers, or maybe JUST for mouse events, since pointer events can carry this information anyway in the touches / changedPointers ( or whatever they have ).

src/mixins/canvas_events.mixin.js
```diff
@@ -368,6 +453,10 @@
         return;
       }
 
+      if (!this._isMainEvent(e)) {
+        return;
+      }
+
```
asturur: this repeated for down/up/move is needed to avoid doing stuff in fabricjs object transformations when we are outside the single pointer movement, main touch changes.

asturur:
For now is broken, but i m here with my multitouch stuff to test it

src/mixins/canvas_events.mixin.js
```diff
@@ -60,8 +85,7 @@
       functor(canvasElement, 'wheel', this._onMouseWheel);
       functor(canvasElement, 'contextmenu', this._onContextMenu);
       functor(canvasElement, 'dblclick', this._onDoubleClick);
-      functor(canvasElement, 'touchstart', this._onMouseDown, addEventOptions);
-      functor(canvasElement, 'touchmove', this._onMouseMove, addEventOptions);
+      functor(canvasElement, 'touchstart', this._onTouchStart, addEventOptions);
```
asturur: do not register touchMove by default. is not used if not after a touchStart. Use specifct onTouchStart instead of _onMouseDown.

src/mixins/canvas_events.mixin.js
```diff
@@ -238,29 +264,113 @@
...
+      // Wait 400ms before rebinding mousedown to prevent double triggers
+      // from touch devices
+      var _this = this;
+      if (this._willAddMouseDown) {
+        clearTimeout(this._willAddMouseDown);
       }
+      this._willAddMouseDown = setTimeout(function() {
+        addListener(_this.upperCanvasEl, eventTypePrefix + 'down', _this._onMouseDown);
+        _this._willAddMouseDown = 0;
+      }, 400);
```
asturur: add some clearTimeout logic to avoid registering multiple mousedown when another touchend would happen before 400 ms

src/util/dom_event.js
```diff
@@ -191,46 +191,21 @@
...
+  function getTouchInfo(event) {
+    var touchProp = event.changedTouches;
+    if (touchProp && touchProp[0]) {
+      return touchProp[0];
     }
-
-    return pointer;
-  }
-
-  if (fabric.isTouchSupported) {
```
asturur: remove this.  `isTouchSupport` is randomly broken depending on browsers and security and privacy concerns. If the event has a `changedTouches` then is a touch event.

asturur:
Ok this starts to work

dirkluijk:
Nice work @asturur ! I will test this PR today and come back at you with feedback.

dirkluijk:
I tested this on a touch screen. The fix does not seem to work.

**Steps to reproduce**
1. Start drawing with one finger.
2. Start drawing with second finger (don't release first finger)
3. Release either first or second finger.

**Expected result**
I expect either the second finger to be ignored at all, or a second path.

**Actual result**
* First part of the path (until when the second finger touches) is removed
* Remaining part of the path are lines between both fingers

asturur:
This is not going to fix brushes.
This allows you to fix the brush in the brush file, without bringing specific pencil logic on the canvas.

You can add to the brush events informations like pointerId, activePointers, and mainPointer andavoid it do stop drawing or drawing too much when there are too many fingers, or when the finger is not the main one.

dirkluijk:
All right, I'll check if this PR is OK for a brush fix. I'll get back at you.

dirkluijk:
Some comments:
* maybe rename `activePointers` to `activePointerIds`?
* if feels a bit weird to have both `activePointerIds` and `mainPointerId`. For example, the `_isMainEvent()` method checks them both.

I tried to fix the brush issue in this branch, but I encountered very strange behaviour:

* with touch events, I still cannot prevent issues with second pointers. The second path is not drawn, but the `touchend` of the second pointer is still messing things up.
* with pointer events, things get even worse (different behaviour). Everything seems to break.

In `pencil_brush.class.js` I tried several things, including:
```
onMouseDown: function(pointer, data) {
    if (this.canvas.activePointerIds.length > 1) {
      return;
    }

  /* ... */
}

onMouseMove: function(pointer, data) {
    if (this.canvas.getPointerId(data.e) !== this.canvas.mainPointerId) {
      return;
    }

  /* ... */
}

onMouseUp: function(data) {
    if (this.canvas.getPointerId(data.e) !== this.canvas.mainPointerId) {
      return;
    }

  /* ... */
}
```
However, I cannot get it working. I feel a bit more comfortable with my original PR.

Can you take a look?

asturur:
Yes but your original PR makes an hard choice of locking events completely when are from more than one finger. That feels bad to me.
I'll put up the touch device again later today and continue.

asturur:
> if feels a bit weird to have both activePointers and mainPointerId. For example, the _isMainEvent() method checks them both.

Can you explain me this a bit more? activePointers give us a glimps of how many touches are there, mainPointerId is just a memo to know what event started the current ongoing event flow.

dirkluijk:
> Yes but your original PR makes an hard choice of locking events completely when are from more than one finger. That feels bad to me.
Yeah, I agree with that.
> I'll put up the touch device again later today and continue.
Awesome! 🎉 
> Can you explain me this a bit more? activePointers give us a glimps of how many touches are there, mainPointerId is just a memo to know what event started the current ongoing event flow.
Yeah, my comment was about `_isMainEvent()`:
```
_isMainEvent: function(evt) {
  return this.activePointers.length <= 1 || this.getPointerId(evt) === this.mainPointerId;
},
```
I don't understand this function really. What does it really do? Shouldn't it only check `this.activePointers.length <= 1`? Because is there a difference when I release the "main" pointer before the "second" pointer or vice versa?

asturur:
well i wrote it so it would work most of the time.
I think we can call an experimental phase, we do not know if moving away the main finger should interrupt fabricJS transformation or not.

dirkluijk:
Any chance to take a look at my issue? 😎

asturur:
To be totally honest i didn't. I have been taken by some videogame ( wow, new patch, new farming ).

asturur:
here i  am, wired with the touch device, resuming work.

asturur:
I was about to give up today.
Then i decided for this compromise: when pointer events are enabled, touch events are disabled.
This allow for simpler code all around.
Ie10 works nice too.

I did not get to the brush problem yet.

asturur:
This is not perfect by any mean, but the brush now works correctly. There is some strange interaction with firefox but ie10, ie11, edge, chrome work good.

asturur:
Ok i m going to merge it. The pencil works good now with many fingers on the screen.
At works is as broken as before!

dirkluijk:
Thanks @asturur! I'll test it tomorrow morning right away.

dirkluijk:
Works perfect!!

bricejar:
I don't think it fixed this issue https://github.com/fabricjs/fabric.js/issues/5801 .

asturur:
I have microsoft surface, a stylus , firefox and i could freedraw both with finger and pencil.

bricejar:
@asturur  it works, sorry about that.

src/brushes/pencil_brush.class.js
```diff
@@ -49,7 +52,10 @@
      * Inovoked on mouse move
      * @param {Object} pointer
      */
-    onMouseMove: function(pointer) {
+    onMouseMove: function(pointer, options) {
```
durga598: Need to add extra options parameter in doc

PR #5788, opened 2019-07-02
Word boundary search update

durga598:
Update to check in grapheme text rather to normal text

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -226,7 +226,7 @@
           index++;
         }
       }
-      while (/\S/.test(this._text[index]) && index < this.text.length) {
+      while (/\S/.test(this._text[index]) && index < this._text.length) {
         offset++;
         index++;
       }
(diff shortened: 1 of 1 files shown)


asturur:
can you help me figure out what is broken and how we are fixing it? what kind of text was failing the word search?

durga598:
Check this [jsfiddle](https://jsfiddle.net/durga598/b0n52e9c/show)

asturur:
Ok seems good to me.
I do not understand this language ( hindi? ) and i do not now what is the expected user effect.

Take this grapheme cluster:
image

It has 2 cursor positions, one at the start and one at the middle, if you  press space, it moves as a single block.
I think this is a fabricJs error, or the grapheme splitter you use is not working as expected?

durga598:
Yes it is Hindi, and most of Indian language works like that way. In google doc/notepad I tried, it works like you have mentioned. It removes one piece at a time, as the complex letter comprised of three letters.
letters
And while moving cursor it takes as single letter
text
But using fabricjs, it takes another step in between, is it because of grapheme splitter?
fabric text

asturur:
probably because of the grapheme splitter used that is consodering it 2 pieces ( you can inspect the _text array i suppose

asturur:
going to merge this, should not break anything

Issue #5810, opened 2019-07-19
Fabric fails to correctly import SVG gradients


ajvincent:
## Version
2.4.0

## Test Case
http://jsfiddle.net/ajvincent/b3xL0v2g/4/

## Information about environment
Mozilla Firefox 68.0, Google Chrome 75.0.3770.142

## Steps to reproduce
None.

## Expected Behavior

The red, green and blue sides of the triangle should each have a gradient leading to yellow in the center.

## Actual Behavior

The red, green, and blue sides of the triangle are solid colors.

## Other notes
- Possible duplicate of #3665 .
- The SVG source code is included in-line, which should make it relatively easy to compare to the generated canvas.
- There are gradients that partially work in this image (red-to-green, green-to-blue, blue-to-red), but it's worth noting that the opacity isn't working.

ajvincent:
http://victorblog.com/html5-canvas-gradient-creator/
This site has a pretty good example of what we should expect.  The code is MIT-licensed, but considering that FabricJS does its work with matrices, it won't port over for us anyway.

ajvincent:
I'm afraid I really stink at linear algebra.  Not nearly enough practice in recent years.

I think the place to fix this is in src/gradient.class.js, fromElement, just after gradient.gradientTransform is established.  The problem is that the coords object has no information regarding the _height_ of the bounding box.  This coords object is fully converted away from percentages in _convertPercentUnitsToValues.  

Without the height, determining the correct rotation of the bounding box is impossible.

asturur:
a suggestion:
http://jsfiddle.net/nkej2d7y/
When you have to show a difference in gradients or svg import/export, please show both of them in the fiddle.

ajvincent:
I think my earlier assertion that fromElement is the right place to fix this is incorrect now, as it doesn't cover the case in #3665... which is essentially the same problem without the SVG.

asturur:
ok i get the geometric problem.
In `_convertPercentUnitsToValues` we need to parse the gradientTransform and decompose it with fabric.util.qrDecomose.

Once we do that, we take away the angle property of the  decomposition and we redo the matrix with the rest.

We use the angle to rotate x1, x2, y1, y2. Indeed a gradient rotated of 90 degree is swapping width with height basically. ( x1 - x2 with y1 - y2 ).

May break everything else, so needs to be done with an eye over:
http://fabricjs.com/test/svg_import/

asturur:
[...]

PR #5812, opened 2019-07-20
Implement fill-opacity on Path objects with gradients.

ajvincent:
This fixes the smaller issue in #5810 .  That said, I'm not sure I put the test in the right place.  It took me longer than I would've liked just to craft a test that works.

Please review with a bit of caution, and if the test is in the wrong place, please provide a better test for the right file.  (I'm still quite an amateur at the fabric.js codebase.)

### DIFF
--- a/src/elements_parser.js
+++ b/src/elements_parser.js
@@ -72,7 +72,18 @@ fabric.ElementsParser = function(elements, callback, options, reviver, parsingOp
   proto.resolveGradient = function(obj, property) {
     var gradientDef = this.extractPropertyDefinition(obj, property, 'gradientDefs');
     if (gradientDef) {
-      obj.set(property, fabric.Gradient.fromElement(gradientDef, obj));
+      var gradient = fabric.Gradient.fromElement(gradientDef, obj);
+
+      if ((property + "Opacity") in obj) {
+        var multiplier = obj[property + "Opacity"];
+        if (multiplier != 1) {
+          gradient = new fabric.Gradient(gradient);
+          for (var i = 0; i < gradient.colorStops.length; i++)
+            gradient.colorStops[i].opacity *= multiplier;
+        }
+      }
+
+      obj.set(property, gradient);
     }
   };
 
--- a/src/elements_parser.js
+++ b/src/elements_parser.js
@@ -72,19 +72,8 @@ fabric.ElementsParser = function(elements, callback, options, reviver, parsingOp
   proto.resolveGradient = function(obj, el, property) {
     var gradientDef = this.extractPropertyDefinition(obj, property, 'gradientDefs');
     if (gradientDef) {
-      var gradient = fabric.Gradient.fromElement(gradientDef, obj);
-
       var opacityAttr = el.getAttribute(property + '-opacity');
-      var multiplier = parseFloat(opacityAttr) / (/%$/.test(opacityAttr) ? 100 : 1);
-      multiplier = multiplier < 0 ? 0 : multiplier > 1 ? 1 : multiplier;
-
-      if (!isNaN(multiplier) && (multiplier !== 1)) {
-        gradient = new fabric.Gradient(gradient);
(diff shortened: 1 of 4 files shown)


src/elements_parser.js
```diff
@@ -72,7 +72,19 @@ fabric.ElementsParser = function(elements, callback, options, reviver, parsingOp
   proto.resolveGradient = function(obj, property) {
     var gradientDef = this.extractPropertyDefinition(obj, property, 'gradientDefs');
     if (gradientDef) {
-      obj.set(property, fabric.Gradient.fromElement(gradientDef, obj));
+      var gradient = fabric.Gradient.fromElement(gradientDef, obj);
```
asturur: Ok this is good, but i would like to have it differently.
resolveGradient is called from createCallback, where `el` the original element is available.

I would like to have resolveGradient taking the `el` as additional argument, extract el.getAttribute('fill-opacity') in this function, and have this multiplication logic in the `fromElement` itself.

All in all would be less code.
We could extract fillOpacity inside Gradient.fromElement since we pass the obj in the function, but the point is that obj is already a fabric instance and fillOpacity, while collected during svg parsing, is not an object property and should be removed one day.

I think the solution is good, thank you for finding it.

test/unit/parser.js
```diff
@@ -451,6 +451,28 @@
...
+    '</svg>';
+
+    fabric.loadSVGFromString(string, function(objects) {
+      assert.equal(objects[0].fill.colorStops[0].opacity, 0.5);
+      assert.equal(objects[0].fill.colorStops[0].color, 'rgb(255,0,0)');
+      assert.equal(objects[0].fill.colorStops[1].opacity, 0.5);
+      assert.equal(objects[0].fill.colorStops[1].color, 'rgb(0,255,0)');
+      done();
+    });
+  });
```
asturur: test is good.

test/unit/parser.js
```diff
@@ -451,6 +451,28 @@
...
 
+  QUnit.test('parseSVGFromString path fill-opacity with gradient', function(assert) {
+    var done = assert.async();
+    var string = '<?xml version="1.0" encoding="UTF-8"?>' +
+    '<svg version="1.2" baseProfile="tiny" xml:id="svg-root" width="300" height="400" ' +
+      'viewBox="0 0 300 400" xmlns="http://www.w3.org/2000/svg" ' +
+      'xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xe="http://www.w3.org/2001/xml-events">' +
+        '<linearGradient id="red-to-red">' +
+          '<stop offset="0%" stop-color="#00ff00"/>' +
+          '<stop offset="100%" stop-color="#ff0000"/>' +
```
asturur: if you add here stop-opacity, we can test that the two values gets multiplied.

asturur:
As soon this is corrected and merged i can look in the geometry problem.

src/elements_parser.js
ajvincent: I think this didn't turn out to be less code after all...

asturur: can t you bring the opacity value inside gradient.fromElement? Wasn't that doable?

ajvincent: I don't think so, since the fill-opacity is on the path element, not the gradient element.

ajvincent: Oh.  Duh.

ajvincent:
I am uncomfortable with this latest patch, because I have changed the API to fabric.Gradient.fromElement:  specifically, the second argument has become the third, and the opacity attribute has become the second attribute.  Feel free to reject (again) if you would rather the new opacity attribute be appended to the end of the arguments to preserve API.

asturur:
I agree that would be better to preserve the argument order, even if fromElement is supported just for internal use for the whole SVG parser.
But I really cannot predict if someone is using it stand alone, and if they managed succesfully would be nice to do not break the API for them.

asturur:
i will add you svg example as a visual test ( that will probably fail for now )

PR #5813, opened 2019-07-20
Trigger onChange and onComplete with the right values

### LINKED ISSUE #5700: Animate Calculation Issue in 3.0.0
## Version
3.0.0
Not present in previous 2.x versions.

## Test Case
http://jsfiddle.net/OrthATG/x5nkqbgf/

Tested in Chrome 74 and Safari; macOS 10.12

## Steps to reproduce
In the fiddle, press Toggle Properties to animate some properties. Opacity is toggled from 1 to 0. Left is changed from 0 to 8.

## Expected Behavior
On animation complete we should see `rect.opacity = 0` and `rect.left = 8`

## Actual Behavior
On animation complete we _actually_ see `rect.opacity = 1.1102230246251565e-16` and `rect.left = 7.999999999999999`

It seems animating values between 0 and 8 causes the issue (possibly due to floating point calculation errors).

asturur:
Close #5700 

Trigger onChange and onComplete with values as accurate as possible.

### DIFF
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -4,6 +4,10 @@
     return false;
   }
 
+  function defaultEasing(t, b, c, d) {
+    return -c * Math.cos(t / d * (Math.PI / 2)) + c + b;
+  }
+
   /**
    * Changes value from one to another within certain period of time, invoking callbacks as value is being changed.
    * @memberOf fabric.util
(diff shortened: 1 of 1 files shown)


idelsink:
Would this change not fall under the 'incompatible API changes' section of semver? And this should cause the version to not be 3.4.0 but 4.x.x?

asturur:
the difference is minimal.
You get the arguments on onComplete, exactly as you would get on `onComplete`.
You get the onChange arguments capped at max value if the time is bigger than finish time.

Did you encounter a breaking change?

idelsink:
I did encounter a breaking change but that was mostly my 'incompetence' :-) for depending on the `undefined` value of the onComplete function argument. I wanted to know if another animation was running in the onComplete method so that I can hide something when it was not moving. I reworked it by using the `onChange` percentage and the `abort` method. This seems to work as expected now.

asturur:
the fabricjs animation system is not that great. sometimes i wonder if i should remove and just get some examples with other animation libraries.

idelsink:
You mean like incorporate some external library or take inspiration from another library? I do think that as a whole animation support is a real strong feature of FabricJS

asturur:
i was meaning that any animation library can animate a fabricjs object.
like tween.js or others ( that i do not know of ).
I m not sure it make sense to have our animation code.

Issue #5822, opened 2019-07-31
feat: Emit 'fabric' module to CommonJS export method in HEADER.js


junghwan-park:
## **Is your feature request related to a problem? Please describe.**

First of all, I appreciate your awesome project. And thank you for your great effort to  maintaning `fabric.js`.

 I've difficulty in running cross browsing test using Karma. That difficulty due to `fabric` exported by *named export* only. 
 
`HEADER.js`
```js
//...
if (typeof exports !== 'undefined') {
  exports.fabric = fabric;
}
//..
```

My project is using fabricjs as external library. Webpack doesn't include **fabric.js** code when bundling, so users of my project has to do manually load their webapp by themselves own. That's why all of importing `fabric` codes are written `import fabric from 'fabric'`.

But, Karma environment is using `fabric.js` as commonJS module. and bundle The tests are failed with *default import* syntax because of current `fabric.js`'s _named export_ method. 

- In tests:
```js
import fabric from 'fabric'; 
// ...
const MyRect = fabric.util.createClass( // Error occurs - Because 'fabric' is undefined. :( 
```

As I said earlier, In the web browser environment, `fabric` defined as `global.fabric`.

- In web browser:
```js
import fabric from 'fabric'; 
// ...
const MyRect = fabric.util.createClass( // It's working.
```

I think this issue came from different depth of module definition.

## **Describe the solution you'd like**

Why don't you emit `fabric` to `module.exports` as same as `exports`?
```js
if (typeof exports !== 'undefined') {
  if (typeof module !== 'undefined') {
    (module.exports = fabric).fabric = fabric;
  }
  exports.fabric = fabric;
}
```

To do this, exporting module has same depth I guess.

## **Describe alternatives you've considered**
Webpack [externals](https://webpack.js.org/configuration/externals/#object) configuration not helped. Because, `fabric` module is exported as _named export_ already. :( 

`fabric.js` export code: 
```js
var fabric = fabric || { version: '3.3.0' };
if (typeof exports !== 'undefined') {
  exports.fabric = fabric;
}
```

Webpack bundled code:
```js
//...
        /***/ }),
	/* 8 */
	/***/ (function(module, exports) {

		module.exports = __WEBPACK_EXTERNAL_MODULE_8__; // 

	/***/ }),
//...
```

And second hurdle was webpack generating `_interopRequireDefault()` function.

```js
function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
```

[...]

PR #5823, opened 2019-07-31
feat: Add CommonJS export method support to HEADER.js

junghwan-park:
refer: fabricjs/fabric.js#5822

### DIFF
--- a/HEADER.js
+++ b/HEADER.js
@@ -2,6 +2,9 @@
 
 var fabric = fabric || { version: '3.3.0' };
 if (typeof exports !== 'undefined') {
+  if (typeof module !== 'undefined') {
+    (module.exports = fabric).fabric = fabric;
+  }
   exports.fabric = fabric;
 }
 /* _AMD_START_ */


asturur:
I'm not opposed to this, but i need to understand some things firsts.

Why the default and name import is necessary? i would prefer to have one only.
In your issue you refer to `import fabric from 'fabric'` why can't you change that to `import { fabric } from 'fabric';`

junghwan-park:
@asturur 
As I issued earlier in #5822, I've encounter the module importing problem in difference of module exporting depth.
I issued #5823 yesterday, but I resolved karma test issue via karma and webpack configuration today. I've mis-configured karma & webpack config file yesterday.  :(

Anyway, Here is my thought. CommonJS module definition depth is one step deeper from others.

**three way to get `fabric` module**
- In Web Browser : using `fabric` direct
```js
fabric.util.createClass();
```
- In AMD module : using `fabric` direct
```js
define(['fabric'] , function (fabric) {
    return function () {
        fabric.util.createClass();
    };
});
```
- In CommonJS module : `( { fabric: { /* genuine */ } } )`
```js
require('fabric').fabric.util.createClass();
```

It might have bring some confusing situation to someone as I experienced. 

If you think this is not needed right now, just close PR.

Thanks :)

iongion:
Could this be solving the problems caused by attempting to use fabric as external in webpack  ?.

Now, if I try to make it an external and load it from CDNJS , the imported object will be empty with babel import statements like:

```javascript
import { fabric } from 'fabric;
```

asturur:
what do you mean by 'an external'?

iongion:
@asturur sorry, I mean this:

https://webpack.js.org/configuration/externals/

In my case I use the CDN of CDNJS

https://cdnjs.cloudflare.com/ajax/libs/fabric.js/3.4.0/fabric.js

And the index.html looks like this

image

Allowing people not to bundle the entire fabric library in a large vendors file.

asturur:
i m not sure what  is wrong in having a vendor file as large as you need.
i would like to solve as many as possible import issues anyway.

the one for which the pr has been open is still a bit unclear to me

hga77:
I think I am having the same issue... Look here
https://github.com/fabricjs/fabric.js/issues/6552

PR #5836, opened 2019-08-10
Fix gradient parsing

### LINKED ISSUE #3665: Shape gradients with percentage based coordination don't work correctly.
## Version
1.7.3

## Test Case
https://jsfiddle.net/human_a/9xfa5u77/

## Steps to reproduce
In my app, I am using the following function to turn angle (degrees based) to X and Y coordinations, but even using the examples from [this page](https://github.com/kangax/fabric.js/wiki/Working-with-gradients) would not work if I hard code them.
```javascript
const anglePI = (-parseInt(angle, 10)) * (Math.PI / 180)
const angleCoords = {
    'x1': Math.round(50 + Math.sin(anglePI) * 50) + '%',
    'y1': Math.round(50 + Math.cos(anglePI) * 50) + '%',
    'x2': Math.round(50 + Math.sin(anglePI + Math.PI) * 50) + '%',
    'y2': Math.round(50 + Math.cos(anglePI + Math.PI) * 50) + '%',
}
```

This function works perfectly fine and I am using the same technique to create SVG gradients for canvas background, while it works for the canvas background flawlessly, it does not work for shapes unless I use shape width and height which will limit my app for using different angles (I have a knob using which I can change the angle on mouse drag)

## Expected Behavior
Percentage based coordinations should work the same as width and height based coords, but it does not produce correct gradients for shapes.

### LINKED ISSUE #5810: Fabric fails to correctly import SVG gradients

asturur:
close #5810 
close #3665

This PR rewrite fabricJS gradient parsing and fixes those use cases:
- better support for gradient transform
- better support for ObjectBoundingBox gradient units
- better support for percentage coords on UserSpaceOnUse ( use the svg viewbox as reference )

It also enable the following feature:
- now gradients have a property `gradientUnits` that can be pixels or percentage and allows you to  express the gradient properties in percentage relative to the object utilizing it.


ajvincent:
I will offer a code review momentarily, but I strongly suggest you get another code review from someone who understands the internals of Fabric (because I don't).

src/gradient.class.js
```diff
@@ -95,18 +95,59 @@
      */
     offsetY: 0,
 
+    /**
+     * A transform matrix to apply to the gradient before paiting.
+     * It shares the same name with the svg property but behaves differently.
+     * @type Array[Number]
+     * @default null
+     */
+    gradientTransform: null,
```
ajvincent: 1. painting
1. behaves differently how?  Describe the layout and the members of this matrix, please, enough to describe them for people coming along in the future.

```javascript
/**
 * | x1 x2 0 0 
 * | ... |
 * ...
 * x1: a description of x1
 * ...
 */
```

Yes, I know, JavaDoc is probably insufficient for this (I've never seen a 2-D array defined in JavaDoc in detail), but some documentation of the members is better than none.

Even better would be to make this an instance of an explicit unit-testable matrix class.  I realize that would be a significant refactor, but I'd be very surprised if you didn't have a need for it by now elsewhere in the code.  I do see there's a fabric.ColorMatrix class, but that appears to be something else.

src/gradient.class.js
```diff
@@ -95,18 +95,59 @@
...
+
+    /**
+     * coordinates units for coords.
+     * If `pixels`, the number of cords are in the same unit of width/ height.
```
ajvincent: coords

src/gradient.class.js
```diff
@@ -95,18 +95,59 @@
...
+     * @param {Object} [options.offsetX] SVG import compatibility
+     * @param {Object} [options.offsetY] SVG import compatibility
+     * @param {Object} options.coords contains the coords of the gradient
+     * @param {Array[Object]} options.colorStops contains the colorstops.
+     * @param {Number} [options.coords.x1]
+     * @param {Number} [options.coords.y1]
+     * @param {Number} [options.coords.x2]
+     * @param {Number} [options.coords.y2]
+     * @param {Number} [options.coords.r1]
+     * @param {Number} [options.coords.r2]
```
ajvincent: options.coords should probably be next to options.coords.x1.  Move colorstops up one line?

src/util/misc.js
```diff
@@ -165,9 +165,15 @@
     /**
      * Returns coordinates of points's bounding rectangle (left, top, width, height)
      * @param {Array} points 4 points array
+     * @param {Array} [transform] 6 number trasnform matrix
```
ajvincent: transform (typo on the second use).  Also mark it optional?

Is this array 1x6, 6x1, 2x3, 3x2? :-)

src/util/misc.js
```diff
@@ -658,7 +664,7 @@
     },
 
     /**
-     * Decomposes standard 2x2 matrix into transform componentes
+     * Decomposes standard 2x3 matrix into transform componentes
```
ajvincent: components

src/gradient.class.js
```diff
@@ -307,7 +344,7 @@
      * @see http://www.w3.org/TR/SVG/pservers.html#LinearGradientElement
      * @see http://www.w3.org/TR/SVG/pservers.html#RadialGradientElement
      */
-    fromElement: function(el, instance, opacityAttr) {
+    fromElement: function(el, instance, opacityAttr, svgOptions) {
```
ajvincent: You might as well add a javadoc line for svgOptions.

src/util/misc.js
```diff
@@ -681,6 +687,61 @@
...
+      if (options.skewX) {
+        scaleMatrix = fabric.util.multiplyTransformMatrices(
+          scaleMatrix,
+          [1, 0, Math.tan(fabric.util.degreesToRadians(options.skewX)), 1],
+          true);
+      }
+      if (options.skewY) {
+        scaleMatrix = fabric.util.multiplyTransformMatrices(
+          scaleMatrix,
+          [1, Math.tan(fabric.util.degreesToRadians(options.skewY)), 0, 1],
```
ajvincent: Just a little consistency question here.  Your code style uses fabric.util.cos, fabric.util.sin, but then you abruptly switch to Math.tan.  Maybe you should just define fabric.util.tan as well.

src/util/misc.js
```diff
@@ -681,6 +687,61 @@
...
+     * @param  {Number} [options.angle]
+     * @param  {Number} [options.scaleX]
+     * @param  {Number} [options.scaleY]
+     * @param  {Number} [options.skewX]
+     * @param  {Number} [options.skewX]
+     * @param  {Number} [options.translateX]
+     * @param  {Number} [options.translateY]
+     * @return {Array[Number]} transform matrix
+     */
+    componeMatrix: function(options) {
```
ajvincent: componentMatrix?

ajvincent:
While you're at it, see if this fixes #3665 as well.

src/util/misc.js
asturur: i think is 'compose'.
I used it in a precedent implementation of the parsing, i do not need to keep it.
i ll remove it.

src/util/misc.js
asturur: Math.tan is good enough to be used as it is.
cos and sin never return 0 or 1 when they are at multiples of 90 degrees. That makes number go crazy, like 0.0000009999 instead of 0. That is why we have util.cos and util.sin, to return 0 and 1 when needed.

src/gradient.class.js
asturur: sure, i m not done yet, svg export is now broken and needs to be adequate to the new logic.

src/util/misc.js
asturur: All the matrices are 2 rows, 3 columns and are the standard geometric affine trasformations you use in 2d graphic. the ones you can pass to ctx.transform

PR #5858, opened 2019-08-29
Update _fireSelectionEvents method

RifatBinReza:
The object:selected and object:deselected events doesn't return the latest selected object and last selected object. Attaching the objects in 'opt' while returning the option will let the invoke function to get the returned object.
Before, calling 'selected' or 'deselected' returned the object in event.selected[i] and event.deselected[i]. But it was deleted soon after returning.
This patch will ensure that these events will return the last or latest selected object.

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -10814,6 +10814,7 @@ fabric.PatternBrush = fabric.util.createClass(fabric.PencilBrush, /** @lends fab
       oldObjects.forEach(function(oldObject) {
         if (objects.indexOf(oldObject) === -1) {
           somethingChanged = true;
+          opt.object = oldObject;
           oldObject.fire('deselected', opt);
           removed.push(oldObject);
         }
@@ -10821,6 +10822,7 @@ fabric.PatternBrush = fabric.util.createClass(fabric.PencilBrush, /** @lends fab
       objects.forEach(function(object) {
         if (oldObjects.indexOf(object) === -1) {
           somethingChanged = true;
+          opt.object = object;
           object.fire('selected', opt);
           added.push(object);
         }


asturur:
I m not opposed to this extra value in the option object.

There  is something to consider tho, the object gets mutated and will be reused for many events and if in those events there is some asyncronous processing, the value will change during the handling. That is confusing at least.

In the case of those objects, the handlers are bound to the object instance, so in the handler the object is the 'this'. what does stops you from using the this?

RifatBinReza:
The values were being changed in the 'removed' and 'added' array. There was no way to use the deselected or selected objects from the event. The 'before:selection:cleared' has the target object though. I think the 'removed' and 'added' array attachments in the opt.deselected and opt.selected is is not working as expected.

asturur:
let's talk of what was being changed and what does it mean not working as expected.

RifatBinReza:
Consider this:
object.on('deselected', e =>{
  console.log(e) // we can find e.deselected[i] here from browser console which contains the deselected object
  console.log(e.deselected[0]) //cannot access 0 of undefined. e.deselected is undefined here.
}

Same goes for 'object:selected'.
This was the original problem which can be solved attaching the deselected or selected object to 'opt' when firing from _fireSelectionEvents.

asturur:
i think the problem comes from using a fat arrow there.
The fat arrow codeblock cannot be bound and so you do not get the right 'this' that would be the deselected object.

RifatBinReza:
yeah it works with a normal function block. But it wasn't mentioned in the documentation so it was confusing.

asturur:
that will be added to docs.

The change as proposed is not ideal, i would prefer to eventually create a new options object because of mutation traps dev can fall into

PR #5866, opened 2019-09-05
ISSUE 5814: Paths in path created not cleared properly when removing all objects.

### LINKED ISSUE #5814: Paths in path created not cleared properly when removing all objects.
## Version
Upgrading from 3.2.0 to ^3.3.0

## Test Case
https://codesandbox.io/s/vigilant-ptolemy-n2ztg

browser

Which browsers?
All

## Steps to reproduce
Open sandbox and alternate between v3.2.0 and 3.3.0.
Draw a couple of paths

## Expected Behavior
No black paths remain. All paths are red.

## Actual Behavior
In 3.3.0 the last black path remains, but is not listed in the canvas objects

niels-van-den-broeck:
Change renderAll in finalizePath for pencil brush to requestRenderAll to be in line with other brushes.
resolves 
Fixes #5814

### DIFF
--- a/src/brushes/pencil_brush.class.js
+++ b/src/brushes/pencil_brush.class.js
@@ -280,7 +280,7 @@
       var path = this.createPath(pathData);
       this.canvas.clearContext(this.canvas.contextTop);
       this.canvas.add(path);
-      this.canvas.renderAll();
+      this.canvas.requestRenderAll();
       path.setCoords();
       this._resetShadow();
 


asturur:
Sorry i still did not have time to look at this. Is really a one liner but i need to check the recent pencil brush PR and try to remember why i did not use it in first place.

asturur:
i think this change is fine, and i have absolutely no way to remember why it was not done in first place.

PR #5912, opened 2019-10-04
Implemented `cornerRegionSize` option for objects' controls region size.

ickata:
On touch devices it is really hard to resize/rotate objects with the corner controls. They need to be bigger, bug bigger corner controls look ugly.

This pull request introduces new option in `fabric.Object` called `cornerRegionSize` (default `0`). It adds (in px) extra invisible region around corner controls, so the clickable/touchable area is big enough while the controls can be kept small.

### DIFF
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -43,21 +43,44 @@
           continue;
         }
 
-        lines = this._getImageLines(this.oCoords[i].corner);
+        lines = fabric.util.object.clone(this._getImageLines(this.oCoords[i].corner), true);
+
+        var t = 'topline',
+            b = 'bottomline',
+            l = 'leftline',
+            r = 'rightline',
+            size = this.cornerRegionSize;
+
+        lines[t].d.x += size;
+        lines[t].o.x -= size;
+        lines[t].d.y += size;
+        lines[t].o.y += size;
+        lines[b].d.x += size;
+        lines[b].o.x -= size;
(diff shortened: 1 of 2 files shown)


asturur:
this is great and as soon as i can read it i ll either ask for chamges or merge it.

melchiar:
👍 Great idea!

src/mixins/object_interactivity.mixin.js
```diff
@@ -43,21 +43,44 @@
...
+        lines[b].d.y -= size;
+        lines[b].o.y -= size;
+        lines[l].d.x -= size;
+        lines[l].o.x -= size;
+        lines[l].d.y -= size;
+        lines[l].o.y += size;
+        lines[r].d.x += size;
+        lines[r].o.x += size;
+        lines[r].d.y += size;
+        lines[r].o.y -= size;
```
asturur: ok so i see some red flags here.
We are cloning the data in a performance loop and adding the new size here.
We can do that when we calculate those points first time.

asturur:
Before proceeding with any implementation we should answer those questions:

- how this feature is gonna work for mouse users.?
- should it work for touch events only?
- when the area is large and overlap with other controls how we distinguish one versus another?

ickata:
It looks like this has been implemented with #6098 so I'm closing this.

asturur:
had to implement out of necessity for work. Code will be simplified before release of 4.0

current master code is already 4.0 beta

PR #5933, opened 2019-10-24
added font size to the font cache key

artichox:
Text boxes were not recalculating their width properly after changing the font size. This caused problems such as the text being cut short as it fell out of the text box. This was because the font cache was not updating when the font size was changed.


artichox:
apologies, a file from the distribution folder seems to have made it in, working on removing that now.

artichox:
Closing so I can create another with a cleaner commit tree

PR #5935, opened 2019-10-25
add fontSize to fontCache to fix width calculation bug

artichox:
Text box doesn't update its width correctly when changing the font size. This is because the fontCache doesn't take into account the font size. Adding the font size to the fontCache fixes the bug. Includes tests.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -605,7 +605,9 @@
         fabric.charWidthsCache[fontFamily] = { };
       }
       var cache = fabric.charWidthsCache[fontFamily],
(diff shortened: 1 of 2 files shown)


asturur:
Hi @artichox, in theory the cache is made fontSize-less on purpose.
Fonts should scale linearly with the fontSize and so we measure the width at a fixed font size
(400px) and then we scale down.

What kind of non correctness are you experiencing? can you make an example?

asturur:
Hi @artichox, any chance to look at my quesiton?

asturur:
If you want to bring back the topic i will reopen the pull request.

Issue #5939, opened 2019-10-28
DashArray wrong with strokeUniform 


awehring:
## Version
3.4.0
(also in version 2.7.0)

## Test Case
https://jsfiddle.net/awehring/osv8Lr2k/23/

## Information about environment
Firefox 70.0
Edge 44.18362.387.0

## Steps to reproduce
Define a DashArray for a Stroke of a Rect, Circle, Triangle or other object with `strokeUniform = false`.
Define a DashArray for a Stroke of an identical object with `strokeUniform = true`.
Compare the two objects.

## Expected Behavior
The strokes for both objects should look the same.

## Actual Behavior
The strokes for the second object (yellow) are about 5 times as long as for the first (green).

![image](https://user-images.githubusercontent.com/13434187/67715053-62a16d80-f9c9-11e9-9e8b-7ea2fd8e70af.png)

awehring:
I found a (partial) workaround: Divide the DashArray parameters by the `strokeWidth` when `strokeUniform == true`, like

`obj.set({strokeDashArray: [15 / obj.strokeWidth, 7.5 / obj.strokeWidth]});`

It works well for Circle, Triangle, Ellipse. Not so well for Rect.

This hint might also help with bugfixing.

asturur:
eventually this can be fixed dividing the array elements by the object full scale.

korzo89:
I will try to fix it as a part of PR #5950, since it touches a common area of problems.

asturur:
Deleted my comment, the bug here is that i have no memory of why the strokeWidth is playing a role in the stroke dash array.

I hope there are some comments in the code or in the PR that introduced the feature

asturur:
https://github.com/fabricjs/fabric.js/pull/5473/files#diff-840d0c62989de96492c73f1de1f574d0R1340

This is the code.
I'm not sure is a bug or not,  i'm really undecided.

asturur:
i think the idea of that code was to avoid having strokeDashes of 1 with big stroke width that would look awful, but eventually that is not a problem of the strokeUniform, nor a problem of fabricJS.

I'll meet the guy that wrote that code later, i ll ask him if he remember

asturur:
@korzo89 opene a separate PR, this is an easier fix, while i have no idea what we should do with the other one. Start with removing the code i linked, and let's see what is the end result.

awehring:
[...]

PR #5950, opened 2019-11-04
Fix uniform stroke rendering for objects in groups

### LINKED ISSUE #5611: StrokeUniform attributes don't work， when a Rect in Group
## Version
2.7.0

## Test Case
https://codepen.io/anon/pen/LvVVaP

 browser
chrome

## Steps to reproduce
add a Rect to group with  set StrokeUniform  ”true“， when scale the group， stroke also scale

korzo89:
Fixes the following object rendering issues when using the `strokeUniform: true` property:
- grouped object strokes are stretched after scaling the group despite setting the `strokeUniform` property (issue #5611),
- selection borders are too large for scaled child objects in a selected group (see attached image).
border_bug
close #5611

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -565,7 +565,7 @@
      * @return {Object} .y height dimension
      */
     _getNonTransformedDimensions: function() {
-      var strokeWidth = this.strokeUniform ? 0 : this.strokeWidth,
+      var strokeWidth = this.strokeWidth,
           w = this.width + strokeWidth,
           h = this.height + strokeWidth;
       return { x: w, y: h };
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -565,7 +565,7 @@
      * @return {Object} .y height dimension
      */
     _getNonTransformedDimensions: function() {
-      var strokeWidth = this.strokeWidth,
+      var strokeWidth = this.strokeUniform ? 0 : this.strokeWidth,
           w = this.width + strokeWidth,
           h = this.height + strokeWidth;
       return { x: w, y: h };
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1548,7 +1548,13 @@
 
       ctx.save();
       if (this.strokeUniform) {
-        ctx.scale(1 / this.scaleX, 1 / this.scaleY);
+        var scaleX = this.scaleX;
+        var scaleY = this.scaleY;
+        if (this.group) {
+          scaleX *= this.group.scaleX;
+          scaleY *= this.group.scaleY;
+        }
+        ctx.scale(1 / scaleX, 1 / scaleY);
       }
       this._setLineDash(ctx, this.strokeDashArray, this._renderDashedStroke);
       if (this.stroke.toLive && this.stroke.gradientUnits === 'percentage') {
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -194,7 +194,7 @@
      */
     drawBordersInGroup: function(ctx, options, styleOverride) {
       styleOverride = styleOverride || {};
-      var p = this._getNonTransformedDimensions(),
+      var p = { x: options.width, y: options.height },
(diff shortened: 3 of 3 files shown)


awehring:
@korzo89 
Thank you for fixing this !

There is a bug in the same corner: #5939 
Maybe it is easy to fix too. Would appreciate it very much!

asturur:
Are we talking of active selections or static groups?

korzo89:
The stroke stretching issue is related mostly to user-created `fabric.Group` objects, since individual object strokes are rendered correctly. Since, the active selection group also uses the `fabric.Group` class internally, the border size problem occurs for both cases.

src/mixins/object_geometry.mixin.js
```diff
@@ -565,7 +565,7 @@
      * @return {Object} .y height dimension
      */
     _getNonTransformedDimensions: function() {
-      var strokeWidth = this.strokeWidth,
+      var strokeWidth = this.strokeUniform ? 0 : this.strokeWidth,
```
asturur: this part does not convince me.
This is to get object dimensions, and the objects dimensions do not change with stroke uniform, before transformation.
So this change, while can be useful fir this case, is not correct for others.

asturur:
I need some time to wrap my head around what is going on here.

src/mixins/object_geometry.mixin.js
korzo89: Yeah, I also had some reservations about this change. 

The original implementation unfortunately doesn't take stroke scaling into account (since those are "non-transformed dimensions"). I found out that transformed dimension finalization in `_finalizeDimensions` includes the stroke width, but it is after the initial stroke width has been included in the `_getNonTransformedDimensions` calculation. After applying the scaling, the initial `strokeWidth` gets scaled as well in the final dimensions, which is wrong for `strokeUniform: true`.

I'll try to think of a different solution, which doesn't modify the `_getNonTransformedDimensions` function significantly.

asturur: you can use naked width and height instead of calling _getNonTransformedDimensions, not a big deal `{ x: this.wdith, y: this.height }`

korzo89: You're right, it will be simpler and safer that way :) I changed the code in a new commit.

src/shapes/object.class.js
```diff
@@ -1548,7 +1548,13 @@
...
       ctx.save();
       if (this.strokeUniform) {
-        ctx.scale(1 / this.scaleX, 1 / this.scaleY);
+        var scaleX = this.scaleX;
+        var scaleY = this.scaleY;
+        if (this.group) {
+          scaleX *= this.group.scaleX;
+          scaleY *= this.group.scaleY;
+        }
+        ctx.scale(1 / scaleX, 1 / scaleY);
```
asturur: let's try to fix this.
This code will work only for object inside one group.
We need ALL the groups up to the canvas, and those can be more than 1.

Another example is that if you have an object rotated of 90degrees, the scaleX of the group will influence the scaleY of the object.

So things are never so easy.

If you want to bring this forward you can divide this code in 2 parts.

```js
if (this.strokeUniform && !this.group) {
// the old code, the most common case and more performant
} else if (this.strokeUniform) {
// there is a group then. Use calcTransformMatrix from the object, that will give you the full transform
// then decompose the matrix in scalex/scaleY operations.
var options = fabric.util.qrDecompose(this.calcTransformMatrix());
ctx.scale(1 / options.scaleX, 1 / options.scaleY);
}
```

See if this work, it would be a more generic case and would work with any rotation.

asturur:
@korzo89 do you have time to move this forward?

korzo89:
Sorry for such a long delay, I didn't have enough time to take a deeper look into the issue. I'll try to get back to that in the next following days.

asturur:
no worries, i asked because i would like to include the fix and if you do not have time i can follow up

asturur:
i need to take over this at least for the  object.class.js part, since i have to fix a production bug somewhere else

asturur:
@korzo89 i m merging this as it is, and changing the things i found out with a following PR. Is the only way i can give you a minimum of attribution for the work done.

I just need it now, and i can't wait for you to finish it.

PR #5953, opened 2019-11-06
Fix dashed stroke rendering with strokeUniform property

korzo89:
Fixes issue #5939.

I removed the code lines pointed by @asturur in the following comment:
https://github.com/fabricjs/fabric.js/issues/5939#issuecomment-549788287

It seems that the code was added to prevent "ugly" rendering of short dash array distances, however IMHO this should not be a feature of fabric.js - the user should be solely responsible for properly defining dash arrays. The rationale behind this is that unscaled objects should look the same, regardless of the `strokeUniform` property (see images below).

I used the following code for testing:
```js
canvas.add(new fabric.Rect({
    left: 10,
    top: 10,
    width: 300,
    height: 75,
    fill: '#00ff00',
    stroke: '#000000',
    strokeWidth: 5,
    strokeDashArray: [1, 1],
    strokeUniform: false
}));

canvas.add(new fabric.Rect({
    left: 10,
    top: 120,
    width: 300,
    height: 75,
    fill: '#ffff00',
    stroke: '#000000',
    strokeWidth: 5,
    strokeDashArray: [1, 1],
    strokeUniform: true
}));
```
For unmodified fabric.js code the test produces the following result:
image
As you can see, the strokes look dramatically different.

After modifying the code, the result looks as follows:
image
The strokes are now identical.

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1381,9 +1381,6 @@
       else {
         alternative && alternative(ctx);
       }
-      if (this.strokeUniform) {
-        ctx.setLineDash(ctx.getLineDash().map(function(value) { return value * ctx.lineWidth; }));
-      }
     },
 
     /**


asturur:
Hey @stefanhayden do you remember why we decided that in the dash stroke, the stroke width should matter?
I'm sure it counts in our app, but why we added this when we ported back here the strokeUniform?

Reading it today seems unrelated.

stefanhayden:
I though it was to fix how it looks as it scales. the whole point of `strokeUniform` is visual consistency of the stroke as you scale it. I'll have to look how it scales with `strokeDashArray` with the proposed code but if we loose visual consistency during scaling then we might need a different solution. Clearing the above static example of the current state looks wrong.

awehring:
@stefanhayden 
You may look at the original issue #5939. I show there how I handle it in my application. Scaling the DashArray linear with the strokeWidth is not pretty.
I'm not shure if this can be handled in a general approach in fabric.js or is better left to the application.

asturur:
Yes the point here was consistency.

this is what i would expect from a stroke uniform ( yellow ) and a normal ( green )
gsplit2

While this is what we have now
gsplit2

What we are trying to fix is the bigger dash, in static condition, with the same values of non uniform.

I think a dev would expect that 2 stroke dash with same values would look the same at scale 1

stefanhayden:
yeah I agree that @asturur's first image is toally what should happen and if it doesn't then it seem like a bug.

asturur:
@korzo89 what are you missing now is a visual test added to close the PR.
Please look in the tests/visuals folder of one to add.
If you need guidance, ask.

asturur:
I will add a visual test.

PR #5957, opened 2019-11-11
Use SVG Namespace for SVG Elements

rockerBOO:
- Replaces `createElement` with `createElementNS` for SVG elements. 
- Replaces `setAttribute` with `setAttributeNS` for SVG attributes.

Doing some code quality checks and found this suggestion. 

https://webhint.io/docs/user-guide/hints/hint-create-element-svg/
http://zhangwenli.com/blog/2017/07/26/createelementns/

Side note, `dist/fabric.js` and `.gitattributes` are not intentional to this request.


asturur:
Hi @rockerBOO, thanks for the PR.
To remove the non intentional files,
run `git checkout master dist` and then commit again. This will remove the changes to the dist folder.

the file `.gitattributes` you can just do `git rm .gitattributes`

Can you explain what exactly will change with this different way to set the attributes?

Also i think it would be nice to have the SVG namespace defined on the fabric object like

```
fabric.svgNameSpace = 'http://www.w3.org/2000/svg'
```

So that we do not have to repeat it all over the place.

rockerBOO:
This change is for consistency. The `setAttributeNS` further increases consistency with applying the correct SVG element attribute from the namespace, instead of a unknown HTML element attribute . The actual behavior may conflict with HTML or other namespaces. 

We apply the namespace in all other examples (through raw SVG files) but the behavior of these tests may change as browser adaptions implement fallback behavior differently. As all the tests pass without changing the tests, we can assume most browsers fallback appropriately currently.

I can make the modifications shortly.

asturur:
@rockerBOO let me know if you can make the changes, or give me the write auth on the branch i'll move it forward

rockerBOO:
I believe I have given you write access.

Updating to this would be the last step.
```
fabric.svgNS = 'http://www.w3.org/2000/svg'
```

asturur:
i'll do the change on top of your, since is easier to me. Less fiddling with git.

PR #5966, opened 2019-11-14
fix: sometimes active selector did not render when setActive

monologueH:
the active selector did not render when I triggered the setActiveObject API by click a button outside canvas.


asturur:
setActiveObject does not want to render.
Is developer task to do so, and we do not want fabricJS to have too many automatic renders.

PR #5983, opened 2019-11-18
Clippath transform parsing

### LINKED ISSUE #5512: SVG Image cropped with fabric.Image.fromURL
## Version
2.4.6

## Test Case
https://jsfiddle.net/kbcn4mpd/

Tested on Chrome 71.

## Steps to reproduce
See fiddle.
I try to add an image to the canvas, using `fabric.Image.fromURL()` or `new fabric.Image()`. This image is a SVG file but I don't want to be able to manipulate objects inside this SVG, I just need to manipulate this SVG as a whole image.

## Expected Behavior
The canvas should show the entire image (french flag).

## Actual Behavior
The image appears, but is cropped: we can only see the upper left corner. I you remove the canvas' `height` attribute, then it appears correctly but then I am not able to correctly size the canvas.

### LINKED ISSUE #5882: SVG rendering bug
## Version
3.4.0

## Test Case
http://jsfiddle.net/3equ92m0/1/

Windows 10 64-bit 1903 (Build 18362.295). Chrome 76.0.3809.132.

## Steps to reproduce
Load SVG file. You can download it [here](https://drive.google.com/file/d/1vvuyQTFRPGyaQ2DLTfDoArVRs3mTS-0G/view?usp=sharing).

## Expected Behavior
The entire SVG should be rendered:
image

## Actual Behavior
The SVG doesn't render correctly. It does render correctly when opening it on Chrome or Adobe Illustrator.
image

asturur:
close #5512
close #5882 

The idea behind this change is to move the clipPath as a sibling near the element that is referencing it. This will allow for nested clipPath parsing with correct transform. At least some of them.


src/shapes/path.class.js
```diff
@@ -500,7 +500,7 @@
           coords = [],
           currentPath,
           parsed,
-          re = /([-+]?((\d+\.\d+)|((\d+)|(\.\d+)))(?:e[-+]?\d+)?)/ig,
+          re = fabric.rePathCommand,
```
asturur: this saved a bunch of memory GC during a high number of path parsing.

PR #5985, opened 2019-11-19
Feature: Add vertical alignment for Text

yassipad:
This enables the setting of an extra "verticalAlign" property on text objects ("top", "middle" or "bottom"). Cursor position, text selection and SVG exports also work.

Please check the jsfiddle for a demo: https://jsfiddle.net/1fcmse6o/161/


asturur:
While i have nothing to say about the execution that is good, i wonder what is the purpose of this functionality.

Can you explain me some practical example in which this feature gets used?
Why should you get bigger text boxes compared to the text content?

yassipad:
I am currently building some kind of a slide editor where for alignment purposes, centering text inside another object or dividing the space equally, being able to have the same size across textboxes is very useful :)

asturur:
But this is how you got to the solution. What is the actual feature you are looking for? can you formulate it?

Is it alignment/centering?
If is it, we need to see how to get there without developing a text vertical alignment tool, that outside your use case is not useful.

yassipad:
I don't really understand what you're asking for? As the title states, the feature is being able to set the vertical alignment of a textbox, so why would we want not to develop "a text vertical alignment tool"?

asturur:
FabricJS does fit the box of the text automatically to text content. There is never space to align text vertically.

So the feature is not implemented.

In your application you created an use case and you wrote some code to have bigger bounding box for text and then you need to align the text inside those box.

This seems, as you also briefly said, an alignment and distribution use case more than vertical text align.

So what i m asking is, why you think we should vertically align text? isn't something else that solve your final goal that is already doable in fabricJS?

yassipad:
Ok, I see what you mean, thanks for the clarification :)
I'm not a fabric.js so I can't say I'm a 100% sure but it seems to me that what I'm trying to achieve is not currently doable. Basically, the idea is that in "creation mode", one may edit a canvas as if you were editing a slide on PowerPoint/Google Slides for instance. So the first use case is when you design a slide, you may need to set a fixed size for your textboxes irrespective of their content for templating purposes. I've also implemented a feature that allows me to add text "inside" another object directly (e.g. double click on a circle to add content to it). In such a scenario, the bounding box of the textbox needs to be the same as the object it is inside of and follow all its rotating/scaling/moving actions. Having a simple "verticalAlign" property to set allows me to do all that without having to recalculate the coords of the textbox (to align it vertically) everytime the parent object is modified.

I don't know if that makes sense and if you think there's a better approach to what i'm trying to achieve, please share your thoughts :)

barnes7td:
@asturur 

_Can you explain me some practical example in which this feature gets used?
Why should you get bigger text boxes compared to the text content?_

Many applications, including the one my team is building for a customer want the height of the textbox to be a manual adjustment by the user, like the width is. This allows users to layout the maximum area of text a page can contain. A practical example would be laying out columns for a page. 

 Alongside this features is usually inferred that the text can be larger than the textbox and be masked out or have the ability to flow the remaining text into another textbox. 

It is a common feature in native design applications.

asturur:
I think we should go over the ability to have a free height textbox as soon as the new control interface is released. And as soon as we have that reason on this PR.

lk77:
Hello,

any news on this ?

yassipad:
hi there!

in case you're still interested, i had made a module with some extensions I use on my projects (including vertical alignment for text), and i just published it.

```node
yarn add @yassidev/fabric-extensions
```

```js
import fabric from 'fabric'
import { textVerticalAlign } from '@yassidev/fabric-extensions'

textVerticalAlign(fabric)
```

i'm using a lot of other small extensions included in the package like keyboard shortcuts, copy/paste from clipboard, drop images on canvas, etc. so feel free to add other extensions as well ;)

asturur:
Great!
Custom constrols api is also something people could look  for too.

PR #5997, opened 2019-11-22
ISSUE-5996 fixing pattern svg export

### LINKED ISSUE #5996: toSVG() ignores objects with fill type pattern with patternTransform?
## Version
3.5.1

## Test Case
http://jsfiddle.net/BrandApp_io/3zk9njxb/48/

Browser: Chrome Version 78.0.3904.70 (Official Build) (64-bit) - on Ubuntu 19

## Steps to reproduce
I'm trying to create a "image crop functionality" (a bit like this thread #4606) in my application, and using the fill type: "pattern" on an object (rectangle is used in the JsFiddle), and then further adjust the pattern image with the offsetX and offsetY and scaling with the patternTransform parameters.

## Expected Behavior
toSVG() should handle the patternTransform

## Actual Behavior
toSVG() seems to ignore patternTransform?

image

asturur:
close #5996 

Need a proper fix and tests, both UTs and visual.

This is an half fix.
In the case mentioned in the issue, this fix SEEMS to work. But if you introduce an offset to the pattern and a repeat schema you will notice that the wrapping part of the pattern is different.
The hack used to scale the pattern with the GROUP instead of pattern trasnform has some kind of side effect hard to understand for now.

If using patternTransform directly some kind of offset is needed and is unknow to me how to calculate it.

### DIFF
--- a/src/pattern.class.js
+++ b/src/pattern.class.js
@@ -130,7 +130,7 @@
           patternHeight = patternSource.height / object.height,
           patternOffsetX = this.offsetX / object.width,
           patternOffsetY = this.offsetY / object.height,
-          patternImgSrc = '';
+          patternImgSrc = '', patternTransform = '';
       if (this.repeat === 'repeat-x' || this.repeat === 'no-repeat') {
         patternHeight = 1;
         if (patternOffsetY) {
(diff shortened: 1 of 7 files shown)


asturur:
needs a visual test.

PR #6013, opened 2019-12-03
ISSUE-4115 - triggers in/out events for sub targets

### LINKED ISSUE #4115: no mouseover/mouseout event on objects in groups (mousedown/mouseup exist, though)
I have a fabric.Group to which I added a couple of fabric.Rect via group.AddWithUpdate(). (All are displayed correctly in my canvas.)

this event doesn't fire (same with 'mouseout'):

    rect.on('mouseover', function () {
        console.log('mouseover');
    });

yet this event DOES fire (same with 'mouseup'):

    rect.on('mousedown', function () {
        console.log('mousedown');
    });

As per http://fabricjs.com/fabric-intro-part-2#events and https://github.com/kangax/fabric.js/wiki/Working-with-events, it should work though:
"For convenience, Fabric takes event system even further, and allows you to attach listeners directly to canvas objects. [...] We're attaching event listeners directly to rectangle and circle instances. Instead of "object:selected", we're using "selected" event." 

my group and rect code:

    var buttonsGroup = new fabric.Group([], {
        left: 0,
        top: 55,
        selectable: false,
        hoverCursor: 'auto',
        subTargetCheck: true,
        myID: 'group_menubuttons'
    });

    var menuButton = new fabric.Rect({
        left: buttonPosX,
        top: buttonPosY,
        width: buttonSize.width,
        height: buttonSize.height,
        fill: '#FFF',
        hoverCursor: 'pointer',
        hasControls: false,
        myClass: 'menuButton',
        myID: 'menuButton_' + type
    });

buttonsGroup.addWithUpdate(menuButton);

AndreaBogazzi confirmed on Stackoverflow, that "Sadly mouseUp, mouseDown events are for now the only one supported inside groups."

Example for use case: buttons in a button container. the container is a group, to which the buttons are added to. the buttons shall change their appearance on mouseover.

Adding these events would be greatly appreciated.

jakedowns:
+ and calls _setCursorFromEvent for sub targets


jakedowns:
Shout-out to #4115 Here's a little preview of this PR in action to get you excited! :D

Screen Recording 2019-12-03 at 10 02 AM

jakedowns:
@asturur I am noticing something weird with this implementation, when I scroll (via `canvas.viewportTransform[5]`) the hover Over seems to work ok, but the hover Out isn't

I'll try to put together a reduced test case scenario tomorrow. 

https://share.getcloudapp.com/o0uQylBp

asturur:
i need some time to read it carefully.

asturur:
i need some time to read it carefully.
You wrote lot of code, 0 comments. :)
takes time to review.

Try to explain what is the approach you took, there are difficulties in handling hover and out, on a stack of objects.

I'm happy to help you trough this and merge it, as long as you are ok accepting that is a mid long process of understanding and refining.

jakedowns:
Oh yes, I totally understand. Thanks for being receptive & responsive to this PR.

My initial approach followed your original suggestion to the letter, I simply added a targetName argument, and used the key of the .targets[] array for naming, and tracking ._hoveredTargetN

however, this didn't seem sufficient, as, I'd roll across from one subtarget to the next, and sometimes, depending on the nesting, ._hoveredTarget0 would match (signaling an "oldTarget") but, really, there was a _new_ ._hoveredTarget0 trying to be set. So it would sort of do an outFire & overFire, leading to it *sometimes* working, sometimes not, as it would get false-positives for "targetChanged"

My second attempt was to switch from a list of targets, keyed by their target array index to a dictionary keyed by a `guid`.

1. i added a pseudo-private `.__guid` getter to the base `Object` class (similar to the .__uid property already present)
1. `canvas.hoveredTarget|N` becomes => `canvas._hoveredTargets = {}`
    - i also added `canvas._hoveredTargetsOrdered = [string_guid1, string_guid2]` in case a developer needed to reference the exact stacking order of the hovered objects, but maybe that was overkill. 
1. I updated `fireSyntheticInOutEvents` to accept an array of Objects as it's first `targets` argument
1. Since `_fireEnterLeaveEvents` also uses `fireSyntheticInOutEvents`, i modified it to match, using *targets<array>* instead of *target* and adding a corresponding `canvas._draggedoverTargets<dictionary>, canvas._draggedoverTargetsOrdered<array>`
1. I did my best to null out / delete lingering Object references in these two new places to avoid memory leaks as you suggested
1. I updated all the related tests I could find, but I have not yet added any *new* tests for checking that triggering in/out events for sub targets works as expected😳

asturur:
So i have some requests:
I would leave the `_hoveredTargetsOrdered` out for this PR.
Let's implement the basic functionality first  and test it.

I would also like to avoid to introduce the getter for the _guid.
I do not like the concept that _guid is related to graphical interfaces ( maybe i m wrong? ). I would like to understand if we can relate to the group structure to find the objects rather than adding an identifier for this.

asturur:
I would image somehow the group structure is enough.
The targets we find and we put in the targets array should change only when some target change.
At that point every item in the target array from the change on is no more hovered, and all the one before the changed one are still hovered.

I would like to see this boiled down to the minimum amount of changes we need to support the basic feature.

jakedowns:
Totally understand. Ok, I'll revert back to my first pass at implementing this and squash it into this PR

src/canvas.class.js
```diff
@@ -1477,8 +1477,21 @@
...
       }
-      if (this._hoveredTarget === obj) {
-        this._hoveredTarget = null;
+      // TODO: need a way to cleanly check if obj is one of _hoveredTargetN and de-ref it
+      // if (obj === this._hoveredTarget){
+      //   this._hoveredTarget = null;
+      // }
+      // is there a more sane way than looping through ALL properties of *this* ?
+      // i wanted to put them into a ._hoveredTargets[Array]
+      // but, that complicates the logic of fireSyntheticInOutEvents
```
asturur: strict equality has to work, or we did something wrong somewhere.

asturur:
Today i ll sit down and help with this.

asturur:
i m actually on a local setup validating an idea. I will share asap

src/canvas.class.js
jakedowns: yeah, i understand the strict equality check, i was just wondering about when i have to search Canvas for all of the properties prefixed with "_hoveredTarget" to search through `_hoveredTarget0`...`_hoveredTargetN` to find a match

my gut instinct was to put ._hoveredTargets in an array that would be easier to loop thorugh, rather than all of `Object.keys(this)`

but then it would require a significant change to how `fireSyntheticInOutEvents` so, i kept it simple for now, even though it's probably not the most efficient way to check for target matches

jakedowns:
for what it’s worth. the simpler solution appears to be working on my
branch. the only thing i’m having an issue with is once i pan/scroll using
viewport transform as described in the official docs example, the hover no
longer seems to work. i’m working on putting together a reduced test case
on codepen

asturur:
```
{"version":"3.5.1","objects":[{"type":"activeSelection","version":"3.5.1","left":-152,"top":656.25,"width":356.5,"height":356.5,"scaleX":0.45,"scaleY":0.45,"objects":[]},{"type":"group","version":"3.5.1","left":11,"top":6,"width":511.5,"height":511.5,"objects":[{"type":"rect","version":"3.5.1","left":-255.75,"top":-255.75,"width":50,"height":50,"fill":"#6ce798","scaleX":10.03,"scaleY":10.03,"opacity":0.8},{"type":"group","version":"3.5.1","left":-179.75,"top":22,"width":356.5,"height":356.5,"scaleX":0.54,"scaleY":0.54,"objects":[{"type":"rect","version":"3.5.1","left":-178.25,"top":-178.25,"width":50,"height":50,"fill":"#4862cc","scaleX":6.99,"scaleY":6.99,"opacity":0.8},{"type":"group","version":"3.5.1","left":-163.25,"top":-161.25,"width":177.5,"height":177.5,"objects":[{"type":"rect","version":"3.5.1","left":-88.75,"top":-88.75,"width":50,"height":50,"fill":"#5fe909","scaleX":3.48,"scaleY":3.48,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-59.75,"top":-68.75,"width":50,"height":50,"fill":"#f3529c","opacity":0.8},{"type":"triangle","version":"3.5.1","left":36.03,"top":-38.12,"width":50,"height":50,"fill":"#c1124e","angle":39.07,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-65.75,"top":17.25,"width":50,"height":50,"fill":"#9c5120","opacity":0.8}]},{"type":"group","version":"3.5.1","left":-34.25,"top":-31.25,"width":177.5,"height":177.5,"scaleX":1.08,"scaleY":1.08,"objects":[{"type":"rect","version":"3.5.1","left":-88.75,"top":-88.75,"width":50,"height":50,"fill":"#5fe909","scaleX":3.48,"scaleY":3.48,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-59.75,"top":-68.75,"width":50,"height":50,"fill":"#f3529c","opacity":0.8},{"type":"triangle","version":"3.5.1","left":36.03,"top":-38.12,"width":50,"height":50,"fill":"#c1124e","angle":39.07,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-65.75,"top":17.25,"width":50,"height":50,"fill":"#9c5120","opacity":0.8}]}]},{"type":"group","version":"3.5.1","left":-202.75,"top":-228.5,"width":356.5,"height":356.5,"scaleX":0.61,"scaleY":0.61,"objects":[{"type":"rect","version":"3.5.1","left":-178.25,"top":-178.25,"width":50,"height":50,"fill":"#4862cc","scaleX":6.99,"scaleY":6.99,"opacity":0.8},{"type":"group","version":"3.5.1","left":-163.25,"top":-161.25,"width":177.5,"height":177.5,"objects":[{"type":"rect","version":"3.5.1","left":-88.75,"top":-88.75,"width":50,"height":50,"fill":"#5fe909","scaleX":3.48,"scaleY":3.48,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-59.75,"top":-68.75,"width":50,"height":50,"fill":"#f3529c","opacity":0.8},{"type":"triangle","version":"3.5.1","left":36.03,"top":-38.12,"width":50,"height":50,"fill":"#c1124e","angle":39.07,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-65.75,"top":17.25,"width":50,"height":50,"fill":"#9c5120","opacity":0.8}]},{"type":"group","version":"3.5.1","left":-34.25,"top":-31.25,"width":177.5,"height":177.5,"scaleX":1.08,"scaleY":1.08,"objects":[{"type":"rect","version":"3.5.1","left":-88.75,"top":-88.75,"width":50,"height":50,"fill":"#5fe909","scaleX":3.48,"scaleY":3.48,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-59.75,"top":-68.75,"width":50,"height":50,"fill":"#f3529c","opacity":0.8},{"type":"triangle","version":"3.5.1","left":36.03,"top":-38.12,"width":50,"height":50,"fill":"#c1124e","angle":39.07,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-65.75,"top":17.25,"width":50,"height":50,"fill":"#9c5120","opacity":0.8}]}]},{"type":"group","version":"3.5.1","left":138.3,"top":-90.22,"width":356.5,"height":356.5,"scaleX":0.42,"scaleY":0.42,"angle":62.73,"objects":[{"type":"rect","version":"3.5.1","left":-178.25,"top":-178.25,"width":50,"height":50,"fill":"#4862cc","scaleX":6.99,"scaleY":6.99,"opacity":0.8},{"type":"group","version":"3.5.1","left":-163.25,"top":-161.25,"width":177.5,"height":177.5,"objects":[{"type":"rect","version":"3.5.1","left":-88.75,"top":-88.75,"width":50,"height":50,"fill":"#5fe909","scaleX":3.48,"scaleY":3.48,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-59.75,"top":-68.75,"width":50,"height":50,"fill":"#f3529c","opacity":0.8},{"type":"triangle","version":"3.5.1","left":36.03,"top":-38.12,"width":50,"height":50,"fill":"#c1124e","angle":39.07,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-65.75,"top":17.25,"width":50,"height":50,"fill":"#9c5120","opacity":0.8}]},{"type":"group","version":"3.5.1","left":-34.25,"top":-31.25,"width":177.5,"height":177.5,"scaleX":1.08,"scaleY":1.08,"objects":[{"type":"rect","version":"3.5.1","left":-88.75,"top":-88.75,"width":50,"height":50,"fill":"#5fe909","scaleX":3.48,"scaleY":3.48,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-59.75,"top":-68.75,"width":50,"height":50,"fill":"#f3529c","opacity":0.8},{"type":"triangle","version":"3.5.1","left":36.03,"top":-38.12,"width":50,"height":50,"fill":"#c1124e","angle":39.07,"opacity":0.8},{"type":"rect","version":"3.5.1","left":-65.75,"top":17.25,"width":50,"height":50,"fill":"#9c5120","opacity":0.8}]}]}]}]}
```

Storing here a json loadable nested group that is usefull for tests

asturur:
@jakedowns  i made #6032 a POC i did not even have time to run. I'm going to bed is midnight.
Please have a look at the idea, and think if it can work. I'm sure we can do something on those lines.

jakedowns:
Thank you for your assistance with this. It's always great to have a second set of eyes looking at a problem to come up with the most elegant & straight-forward solution. Also great to have someone to bounce ideas back & forth with.

I've implemented the changes from your POC and your comments from your POC PR.
Thank you for allowing me to take ownership/credit for this PR as well! 🙇

I've tested on my local using the following setup and it appears to be working as expected:
> https://gist.github.com/jakedowns/0fbca45fd8796e109e94635c2f221980

next, I believe, is just a matter of making sure the unit tests account for these changes. Any suggestions as to what you'd like to see in terms of test coverage?

src/canvas.class.js
```diff
@@ -339,7 +339,7 @@
 
     /**
      * hold the list of nested targets hovered
-     * @type fabric.Object
+     * @type Array[fabric.Object]
```
jakedowns: my linter was complaining that i couldn't set this._hoveredTargets to an array

src/canvas.class.js
```diff
@@ -1491,21 +1491,9 @@
...
-      for (var i = 0; i < keys.length; i++){
-        var key = keys[i];
-        if (key.indexOf('_hoveredTarget') > -1){
-          if (obj === this[key]){
-            this[key] = null;
-          }
-        }
+      if (obj === this._hoveredTarget){
+        this._hoveredTarget = null;
+        this._hoveredTargets = [];
```
jakedowns: seems like the correct thing to do here is clear out _hoveredTargets as well, yes? should we be firing mouseout on these here?

src/mixins/canvas_grouping.mixin.js
```diff
@@ -53,20 +53,7 @@
...
-          if (key.indexOf('_hoveredTarget') > -1){
-            this[key] = null;
-          }
-        }
-        // ISSUE-4115: loop through this.targets and assign them as hovered as well?
-        // why don't we fire mouse:over here?
-        for (var i = 0; i < this.targets.length; i++){
-          this['_hoveredTarget' + i] = this.targets[i];
-        }
+        this._hoveredTargets = this.targets;
```
jakedowns: does this seem right to you @asturur ?

src/mixins/canvas_events.mixin.js
```diff
@@ -812,14 +814,6 @@
...
         this._setCursorFromEvent(e, target);
         this._fireOverOutEvents(target, e);
-        // handle triggering on SubTargets
-        this.targets.map(function(subTarget,k){
-          _this._fireOverOutEvents(subTarget, e, '_hoveredTarget' + k);
-        });
-        // hoverCursor should come from top-most subtarget
-        this.targets.slice(0).reverse().map(function(subTarget){
-          _this._setCursorFromEvent(e, subTarget);
-        });
```
jakedowns: what do you think about this call to ._setCursorFromEvent call? Can I leave it in, or should it be a separate PR? I feel like it's relevant to add it in here.

src/mixins/canvas_events.mixin.js
```diff
@@ -902,6 +897,8 @@
         inOpt = { e: e, target: target, previousTarget: oldTarget };
         outOpt = { e: e, target: oldTarget, nextTarget: target };
       }
+      inFires = target && targetChanged;
+      outFires = oldTarget && targetChanged;
```
jakedowns: accidentally removed this in a earlier merge

src/mixins/canvas_grouping.mixin.js
```diff
@@ -53,7 +53,7 @@
       if (activeSelection.contains(target)) {
         activeSelection.removeWithUpdate(target);
         this._hoveredTarget = target;
-        this._hoveredTargets = this.targets;
+        this._hoveredTargets = this.targets.concat();
```
jakedowns: added a concat like you used (TIL that concat is faster than slice(), good to know)

src/mixins/canvas_grouping.mixin.js
```diff
@@ -74,6 +74,8 @@
       var currentActives = this.getActiveObjects(), group = this._createGroup(target);
       this._hoveredTarget = group;
       // ISSUE 4115: should we consider subTargets here?
```
jakedowns: @asturur not really sure what to do here

src/canvas.class.js
asturur: JSDOCS has to be as much good as possible, hoping one day we can pull a ts definition directly from them.

src/canvas.class.js
asturur: well that is not important for now. Seems right to delete them since _hoveredTargets are just part of the main target, if the main target goes, also _hoveredTargets goes.

src/mixins/canvas_events.mixin.js
```diff
@@ -854,22 +891,21 @@
...
         inOpt = { e: e, target: target, previousTarget: oldTarget };
         outOpt = { e: e, target: oldTarget, nextTarget: target };
-        this[config.targetName] = target;
       }
       inFires = target && targetChanged;
       outFires = oldTarget && targetChanged;
       if (outFires) {
         canvasEvtOut && this.fire(canvasEvtOut, outOpt);
-        oldTarget.fire(config.evtOut, outOpt);
+        oldTarget && oldTarget.fire(config.evtOut, outOpt);
```
asturur: this extra `oldTarget &&` is not necessary, since outFires already include `oldTarget` to be non null.

src/mixins/canvas_events.mixin.js
```diff
@@ -854,22 +891,21 @@
...
       }
       if (inFires) {
         canvasEvtIn && this.fire(canvasEvtIn, inOpt);
-        target.fire(config.evtIn, inOpt);
+        target && target.fire(config.evtIn, inOpt);
```
asturur: same as above.

src/mixins/canvas_grouping.mixin.js
asturur: we can skip for now. There are already enough moving pieces.

src/shapes/object.class.js
```diff
@@ -2116,4 +2116,23 @@
...
+    //   var r = Math.random() * 16 | 0, v = c == 'x' ? r : (r & 0x3 | 0x8);
+    //   return v.toString(16);
+    // });
+  }
+  fabric.Object.prototype.__defineGetter__('__guid', function(){
+    if (!this.__myGUID) {
+      this.__myGUID = generateGUID();
+    }
+    return this.__myGUID;
+  });
```
asturur: do we still need this?

src/mixins/canvas_events.mixin.js
```diff
@@ -819,12 +830,38 @@
...
       this.fireSyntheticInOutEvents(target, e, {
-        targetName: '_hoveredTarget',
+        oldTarget: this._hoveredTarget,
         canvasEvtOut: 'mouse:out',
         evtOut: 'mouseout',
         canvasEvtIn: 'mouse:over',
         evtIn: 'mouseover',
       });
+      var _this = this;
+      this.targets.forEach(function(_target, index) {
```
asturur: i was thinking to save space here:
```js
var _hoveredTarget = this._hoveredTarget, _hoveredTargets = this._hoveredTargets, targets = this.targets,
      diff = _hoveredTargets.length - targets.length;
[target].concat(targets, new Array(diff)).forEach(function(_target, index) {
        _this.fireSyntheticInOutEvents(_target, e, {
          oldTarget: _hoveredTargets[index - 1],  // because we know we shifted one place with the new array trick
          canvasEvtOut: 'mouse:out',
          evtOut: 'mouseout',
          canvasEvtIn: 'mouse:over',
          evtIn: 'mouseover',
        });
});
this._hoveredTarget = target;
this._hoveredTargets = this.targets.concat();
```

asturur: oh we have to handle if diff is bigger than 0. But at least is more terse.
I often pull the variables away from `this` with their own name. I do that just because fabrics is big also when minimized and `this.hoveredTarget` does not get minimized. So if i use it more than once in a function, pulling it in a variable allow me to minimize it and save after the first occurrence.

src/mixins/canvas_events.mixin.js
```diff
@@ -835,7 +872,7 @@
      */
     _fireEnterLeaveEvents: function(target, e) {
       this.fireSyntheticInOutEvents(target, e, {
-        targetName: '_draggedoverTarget',
+        oldTarget: '_draggedoverTarget',
```
asturur: this is still half way between new and old.

asturur:
I think we are nearly there, i would happy with even just a new test case using that json string and mocking up some event data. i can try to write one.

Thanks for being patient and collaborative.

src/shapes/object.class.js
jakedowns: nope, forgot to remove. will do

src/mixins/canvas_events.mixin.js
jakedowns: updated in this commit: https://github.com/fabricjs/fabric.js/pull/6013/commits/0cd41961d47a39153bb322bbcf725fc839ed77ac

src/shapes/object.class.js
jakedowns: removed here: https://github.com/fabricjs/fabric.js/pull/6013/commits/deb64f039a8f43aa526b2f6aa088a179f76cc454

src/mixins/canvas_events.mixin.js
jakedowns: updated here: https://github.com/fabricjs/fabric.js/pull/6013/commits/426e68cb407bced5b8205be211339fcdb7de3631

jakedowns: tweaked one aspect of this in my commit
`oldTarget: index === 0 ? _hoveredTarget : _hoveredTargets[index - 1],`

jakedowns:
@asturur Thanks again for all your help and attention with this PR. I stubbed in a new unit test: 

https://github.com/fabricjs/fabric.js/pull/6013/commits/74831428001d603668350fd4d5b54e7795ef0db7

Not sure of your preferred method of simulating the mouseMove events though. couldn't find an existing example of that.

I also adjusted `_setCursorFromEvent` to account for subTargets. I hope that's ok, my project relies on it, so i'm hoping it can be included when this finally releases. https://github.com/fabricjs/fabric.js/pull/6013/commits/1909d1f49ee1c147475e6d71d68299126a33c5b0

src/mixins/canvas_events.mixin.js
```diff
@@ -231,7 +242,7 @@
     _onDragOver: function(e) {
       e.preventDefault();
       var target = this._simpleEventHandler('dragover', e);
-      this._fireEnterLeaveEvents(target, e);
+      this._fireEnterLeaveEvents([target].concat(this.targets), e);
```
asturur: this seems off to me. The signature of the function still wants just a target.

test/unit/canvas_events.js
```diff
@@ -481,6 +483,39 @@
...
   });
 
+  QUnit.test('Fabric mouseover, mouseout events fire for subTargets when subTargetCheck is enabled', function(assert){
+    var counterOver = 0, counterOut = 0, canvas = new fabric.Canvas();
+    function setSubTargetCheckRecursive(obj) {
+      if (obj._objects) {
+        obj._objects.forEach(setSubTargetCheckRecursive);
+      }
+      // TODO: maybe this property could've been baked into the JSON?
+      // not sure if loadFromJSON accounts for it tho
```
asturur: so the JSON file always ever cared for the graphic restore. No funcionalities. This is not written in stone, but for now is out of scope. This same problem has to be revised again when i get in the new control interface.

src/mixins/canvas_events.mixin.js
jakedowns: not sure how this snuck back in, reverted.

jakedowns: good catch

asturur:
I would proceed with a merge and then eventually fix if we broke anything. I can add more tests later.

jakedowns:
let me know if there's anything else you need from me before proceeding with the merge :D

asturur:
No, i want just to add more testing, but i realize we have not the infrastructure. Is easier for me to look at master than working on your branch, the build pass i think is fine.

src/mixins/canvas_events.mixin.js
```diff
@@ -818,13 +827,20 @@
...
-      this.fireSyntheticInOutEvents(target, e, {
-        targetName: '_hoveredTarget',
-        canvasEvtOut: 'mouse:out',
-        evtOut: 'mouseout',
-        canvasEvtIn: 'mouse:over',
-        evtIn: 'mouseover',
+      var _this = this, _hoveredTarget = this._hoveredTarget,
+          _hoveredTargets = this._hoveredTargets, targets = this.targets,
+          diff = _hoveredTargets.length - targets.length;
+      [target].concat(targets, new Array(diff > 0 ? diff : 0).fill(null)).forEach(function(_target, index) {
```
asturur: @jakedowns sorry i did not notice this.
We still support internet explorer and fill is not an option there.

test/unit/canvas_events.js
```diff
@@ -505,17 +503,33 @@
...
+      canvas._fireOverOutEvents(null, moveEvent);
+      assert.equal(counterOut, 4, 'mouseout fabric event fired 4 times for primary hoveredTarget & subTargets');
       assert.equal(canvas._hoveredTarget, null, '_hoveredTarget has been set to null');
       assert.equal(canvas._hoveredTargets.length, 0, '_hoveredTargets array is empty');
     });
   });
 
+  // TODO: QUnit.test('mousemove: subTargetCheck: setCursorFromEvent considers subTargets')
+  // TODO: QUnit.test('mousemove: subTargetCheck: setCursorFromEvent considers subTargets in reverse order, so the top-most subTarget's .hoverCursor takes precedence')
+
```
jakedowns: dunno if you saw this, but I did update this test so that it at least checks that the expected events fire for the subtargets

asturur:
Yes i have seen the tests.
Indeed i said more testing :)

Il giorno mer 18 dic 2019 alle ore 17:34 Jake Downs <
notifications@github.com> ha scritto:

src/mixins/canvas_events.mixin.js
jakedowns: replaced with a for loop: https://github.com/fabricjs/fabric.js/pull/6013/commits/dbaf781eb48feeaf32a96845daa6657f18d457c1

src/mixins/canvas_events.mixin.js
```diff
@@ -834,11 +855,22 @@
...
-        evtOut: 'dragleave',
-        evtIn: 'dragenter',
+      var _this = this, _draggedoverTarget = this._draggedoverTarget,
+          _hoveredTargets = this._hoveredTargets, targets = this.targets,
+          diff = _hoveredTargets.length - targets.length,
+          diffArrayLength = diff > 0 ? diff : 0,
+          diffArray = [];
+      for (var i = 0; i < diffArrayLength; i++){
+        diffArray.push(null);
+      }
```
asturur: oh i see, the array with 3 undefined will not iterate. Well ok this is fine, if will find away to have it more terse, it can be a follow up.

image

the length is 7 but forEach will not care

jakedowns: yeah forEach ignores the empties :/ interested to see what you come up with that's shorter than that. I looked at an Array.prototype.fill polyfill, but it was way more lines than a 3 for loop.

asturur:
to be honest we have a fabric.util.array.fill, but since i'm planning to remove those generic utils from the library ( non standard and out of scope ) i did not want to use it.
Probably there isn't a readable shorter version. Whatever i can think of makes the code unreadable for a couple of bytes of savings. not worth.

PR #6023, opened 2019-12-10
took parser from svgo package

shahafan:
Took path parser from https://github.com/svg/svgo/blob/master/plugins/_path.js

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -497,46 +497,83 @@
      */
     _parsePath: function() {
       var result = [],
-          coords = [],
-          currentPath,
-          parsed,
-          re = fabric.rePathCommand,
-          match,
-          coordsStr;
+          rNumber = String.raw`[-+]?(?:\d*\.\d+|\d+\.?)(?:[eE][-+]?\d+)?\s*`,
+          rCommaWsp = String.raw`(?:\s,?\s*|,\s*)`,
+          rNumberCommaWsp = `(${rNumber})${rCommaWsp}`,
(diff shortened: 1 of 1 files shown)


asturur:
Can i ask what is the goal here?

We are picking up other people's code without proper attribution.
Is not es5 compatible and uglifyJS will probably fail.
We are swapping 32 lines of code with 68.

What made you change the code?
Is there something our current path parser cannot parse?
Is this code sensibly faster?

shahafan:
The current parser isn't compatible with the svg standards, it's just looking for spaces / commas.
On elliptical arc you don't have to include spaces and the parser won't draw the right paths.

Any svg that will get optimization with svgo will fail.

We've started to use this parser now and this has solves us tons of breaks.

I can rewrite their code to be es5 compatible and also try to make it shortener..

If you want an example you can try this one:

https://static.wixstatic.com/shapes/db42c0_af158af3367e4dc9bee9fe009507285d.svg

Let me know if you want me to do here anything else.

asturur:
Those are valuable informations. Thanks for clarifying.
I would like if is possible for you to open a PR with one change at time.
SVGO will compress the path value, using specs, and something will fail.
We should take one of those failing paths and update our code to handle it, without switching to SVGO parser entirely.
That SVG path you provided is good.
We also need tests.

shahafan:
I'll do my best to find the time.

gloriousjob:
@shahafan My PR got pulled into 4.2.0. Would that resolve this PR?

asturur:
yes this can be close

PR #6055, opened 2019-12-28
Reduce event emission on canvas

asturur:
This PR eliminates over firing of events on the canvas when subtargets are firing mouseover/mouseout

To mirror what we do on mousedown/mouseup, events for subtargets are fired only on objects and not for both canvas and object.

I prefer to start with this more restrictive way of firing and widening it later or adding more info to the canvas events.

@jakedowns does this make the feature less usefull for you?


asturur:
i will merge and make a release today, because i have  time to do so in the weekend.
I would like to reiterate if this is not optimal with a patch release later.

Issue #6072, opened 2020-01-05
Convolution filter index edge conditions are off by one


rfinck:
## Version
3.6.0

## Test Case
Unit tests as well as a fix added here: https://github.com/rfinck/fabric.js/blob/fix-convolution-indexing/test/unit/image_filters.js#L703 
I can make a pull request but there are some workarounds I had to do to get the Node tests to run that may or may not be desirable in the project.

## Information about environment
Nodejs or browser? Both 
Which browsers? Any

## Steps to reproduce
See unit test linked above, which applies the following convolution filter to the following data:
```
var filter = new fabric.Image.filters.Convolute({
  matrix: [
    0, 1, 0,
    1, 2, 1,
    0, 1, 0
  ]
});
var pixelData = new Uint8Array([
  0, 0, 0, 255, 0, 0, 0, 255, 0, 0, 0, 255,
  0, 0, 0, 255, 1, 2, 3, 255, 0, 0, 0, 255,
  0, 0, 0, 255, 0, 0, 0, 255, 0, 0, 0, 255,
]);
```
## Expected Behavior
```
var expected = new Uint8Array([
  0, 0, 0, 255, 1, 2, 3, 255, 0, 0, 0, 255,
  1, 2, 3, 255, 2, 4, 6, 255, 1, 2, 3, 255,
  0, 0, 0, 255, 1, 2, 3, 255, 0, 0, 0, 255,
]);
```
## Actual Behavior
```
var actual = new Uint8Array([
  0, 0, 0, 255, 1, 2, 3, 255, 0, 0, 0, 255,
  1, 2, 3, 255, 2, 4, 6, 255, 0, 0, 0, 255,
  0, 0, 0, 255, 0, 0, 0, 255, 0, 0, 0, 255,
]);
```

asturur:
Can you give more info?
the pixel data is 9 pixels, we need to know how they are organized. i suppose 3 by 3 and not 9 by 1 or 1 by 9.

Can you link to the actual fix?
I also wonder the difference between the webgl version and the js version.
Probably webgl is doing something automatic at the edges ( padding or mirroring ).

rfinck:
@asturur In the example above, the pixels are 3x3.

There were actually two different issues
[...]

PR #6088, opened 2020-01-11
ISSUE-6072 Fix image index overrun and add test for js convolution filter

rfinck:
Previously, the convolution filter tried to extend beyond the boundaries of the image, which would attempt to add `undefined` to the last row and column of pixels and thus throw off the results.

This is the js half of #6072 ; the gl part is split out since it includes bigger changes to the testing setup.


asturur:
this is good as it is.
I wonder if i should add a test suite of filters running on images, and compare results.

PR #6089, opened 2020-01-11
ISSUE-6072 Add test for gl convolution and update shader so that it passes

### LINKED ISSUE #6072: Convolution filter index edge conditions are off by one

rfinck:
This is the gl half of #6072 ; the js part is in #6088 since it's a more straightforward change to the tests. It's not clear how the changes here will perform in ci so partly creating this PR to check.

Here I added a test for the gl convolution filter, which required some changes to the shaders to pass in Node. The issue the previous shaders had may not have been a problem when run in the browser.

close #6072


rfinck:
It looks like tests are failing on Node.js 12 and up due to a known issue: https://github.com/stackgl/headless-gl#supported-platforms-and-nodejs-versions

Locally, tests pass for my with Node.js 8 and Node.js 10, as well as with Chrome.
I can reproduce the failures with Firefox and Node.js > 12.

src/filters/convolute_filter.class.js
```diff
@@ -311,7 +311,7 @@
      */
     getUniformLocations: function(gl, program) {
       return {
-        uMatrix: gl.getUniformLocation(program, 'uMatrix'),
+        uMatrix: gl.getUniformLocation(program, 'uMatrix[0]'),
```
asturur: can you explain me this? because i have no idea what does it do.
I would expect this to be just a reference to how is named the variable in the shader code.

asturur:
i have to look a bit at how gl libray work and if it can be integrated with node-canvas or jsdom

src/filters/convolute_filter.class.js
rfinck: Yes it is a reference to how it's named, my understanding is that either referencing it as `'uMatrix'` or as its first element `'uMatrix[0]'` should work but only the second one worked when I was gl in tests; both seem to work when using it normally.

rfinck:
By running Xvfb in travis, unit tests now pass for Node < 13. The remaining issues are Firefox and Node 13 where the gl install errors.

Job | Node.js | ENV | State
-- | -- | -- | --
6643.3 | 12 | LAUNCHER=Chrome | passed
6643.4 | 12 | LAUNCHER=Firefox | failed
6643.5 | 13 | LAUNCHER=Node | errored
6643.6 | 12 | LAUNCHER=Node | passed
6643.7 | 10 | LAUNCHER=Node | passed
6643.8 | 8 |   | passed

asturur:
When running in firefox we should not need the GL module at all, its requiring should be conditional.
That is why i wanted to know if we can just connect it to JSDOM and let JSDOM emulate a normal canvas with webgl support.

PR #6109, opened 2020-01-21
feat(controls): visibility call fix

asturur:
The getVisibility call accepts as argument the instance of the object we are looking at and its name, we are not actually passing them in as of now.

Should name really be needed? the API should push people to write clean simple handlers and not handlers that do many things depending on the control they are.


asturur:
removed the name argument in the getVisibility call

PR #6111, opened 2020-01-21
feat(): remove deprecated code from older versions

asturur:
- removed Object.setShadow, and BaseBrush.setShadow. change `rect.setShadow(options)` to `rect.set('shadow', new fabric.Shadow(options))`
- removed Object.transformMatrix. If you still have transform matrix in your designs look at how fabric removes the transform matrix from svg and preserve visual correctness.
- removed `object:selected` event. use `selection:created`. In the callback you will still find  `target` in the options, but also you will find `selected` with all the objects selected during that single event.
- removed Gradient.forObject. It wasn't that helpful. The new gradient options introduced in 3.4.0 make up for the missing method.
- removed Object and canvas `clipTo`. Use Object.clipPath;
- removed Canvas.loadFromDatalessJSON, it was just an alias for `loadFromJSON`
- removed `observe`, `stopObserving`, `trigger` from observable. Keep using `on`, `off`, `fire`.
- removed the Object.set ability to take a function as a value. Was rather strange to use.
- removed Object.setGradient. Change `rect.setGradient(options)` with `rect.set('fill', new fabric.Gradient(otherOptions))`. The options format is slightly different, but keeping 2 formats does not really make sense.
- removed Object.setPatternFill. Change `rect.setPatternFill(options)` to `rect.set('fill', new fabric.Pattern(options))`;
- removed Object.setColor. Change `rect.setColor(color)` to `rect.set('fill', color)`
- removed fabric.util.customTransformMatrix. Use the replacement fabric.util.composeMatrix


juzhiyuan:
Hi @asturur , I think this wiki page[1] is out of date, such as `object:selected` event was changed to `selection:created`, how about removing it or added some tip[2] on it?

[1] https://github.com/fabricjs/fabric.js/wiki/Working-with-events
[2] http://fabricjs.com/v2-breaking-changes-2

asturur:
That is a breaking change for v4.
I think i should remove the wiki pages, since i already have problem mantaining the fabricjs.com doc website.

juzhiyuan:
Agreed!

Issue #6047, opened 2019-12-24
How to create an instance of Fabric image from Buffer object


Shohin93:
Is there a way to build a Fabric instance from Buffer object, for example, from the following buffer: <Buffer 89 50 4e 47 0d 0a 1a 0a 00 00 00 48 89 95 97 ... >. I'm passing a buffer (fetched from S3 bucket) from another service/application to my current application where I do image manipulations. There's this API `fromObject(object, callback)` (http://fabricjs.com/docs/fabric.Image.html), but passing the above buffer does not seem to be doing job. Anybody knows how to do it?

asturur:
If you are retriving the buffer from a service, is probably coming in a format that is not Buffer.
What format is it?
I would imagine is a url or a data url.

Shohin93:
@asturur it's something like this:
```
{ AcceptRanges: 'bytes',
  LastModified: 2019-12-26T16:10:00.000Z,
  ContentLength: 220518,
  ETag: '"c6288cc91a3993c207a96d169a4921ab"',
  ContentType: 'application/octet-stream',
  Metadata:
   { extension: 'png',
     mime: 'image/png',
     width: '1019',
     height: '776',
     name: 'Screen_Shot_2019_12_18_at_11_01_04_AM' },
  Body: <Buffer 89 50 4e 47 0d 0a 1a 0a 00 00 00 0d 49 48 44 52 00 00 03 fb 00 00 03 08 08 06 00 00 00 97 7e f5 8d 00 00 0a b6 69 43 43 50 49 43 43 20 50 72 6f 66 69 ... > }
```

So, `body` is `Buffer`. Is there any api in in Fabric that works with buffers? `fromObject` doesn't really do the job

asturur:
oh ok you are getting the file itself.
if the api returning you the file is GET and not authenticated, you can pass that URL to the fromURL image method and you should be fine.

Otherwise, assuming you are running under node, the only thing you could try to do is to instantiate a new HTMLImageElement using `fabric.util.createImage`, then you could then try to access the node-canvas implementation of that image using  `fabric.util.getNodeCanvas` and then check the node-canvas docs on how to load an image from Buffer.

(i think is image.src = Buffer but i may be wrong )

Shohin93:
is it something like this:
```
const createdImage = fabric.util.createImage();
const nodeCanvas = fabric.util.getNodeCanvas(createdImage);
```
?
`nodeCanvas` returns `undefined`

Also, tried:
[...]

PR #6141, opened 2020-02-08
If it's a file, then create a node-canvas image

Shohin93:
This PR is followed after https://github.com/fabricjs/fabric.js/issues/6047 and brings back the capability to create an image from file (buffer) as well.

### DIFF
--- a/src/util/misc.js
+++ b/src/util/misc.js
@@ -368,7 +368,7 @@
       // IE10 / IE11-Fix: SVG contents from data: URI
       // will only be available if the IMG is present
       // in the DOM (and visible)
-      if (url.substring(0,14) === 'data:image/svg') {
+      if (!fabric.isLikelyNode && url.substring(0,14) === 'data:image/svg') {
         img.onload = null;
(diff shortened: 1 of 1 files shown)


asturur:
I do not want to create images that are outside the JSDOM HTMLImage elements.
Why this would help compared to the solution proposed in the relevant issue?

Shohin93:
@asturur That solution https://github.com/fabricjs/fabric.js/issues/6047 looks "hacky" to me tbh :smiley: I was wondering why or what was the reason this https://github.com/fabricjs/fabric.js/pull/6141/files#diff-7e5f78926304ab958df2a0f5216c19b2R627-R629 was removed since `v2.x`? Feels like the API for building an image from a buffer file is gone

asturur:
Was removed because mantainability is important here, since i'm alone. JSDOM role it to provide a DOM like implementation in node and node-canvas is compatible with JSDOM. We can still load an image from buffer if JSDOM implementation would collaborate. I do not feel like have 2 different branch of code handling images.

asturur:
another possibility would be inspect if object url are available somehow under node

PR #6152, opened 2020-02-13
Fix centering guidelines for canvas with zoom factor

yongfook:
Before:
Center position is incorrect for canvas with a zoom factor applied

After:
Width is now calculated based on canvas width and zoom factor, so center position is now correct

### DIFF
--- a/lib/centering_guidelines.js
+++ b/lib/centering_guidelines.js
@@ -6,8 +6,8 @@
  */
 function initCenteringGuidelines(canvas) {
 
-  var canvasWidth = canvas.getWidth(),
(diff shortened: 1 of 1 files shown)


asturur:
Hi @yongfook , thank you for your contribution.
Would it be possible to show with a jsfiddle/codepen what is wrong with the actual implementation?

Thanks.

yongfook:
Apologies - I was incorrect. My usage of setZoom() was a bit backwards so it needed edits to initCenteringGuidelines()

With proper usage of setZoom(), everything works as-is :)

PR #6154, opened 2020-02-14
Update return type on toCanvasElement to use HTMLCanvasElement

rockerBOO:
I am not sure if I am misrepresenting the return documentation but the return type is usually HTMLCanvasElement. 

This updates the type documentation and some info about what is being returned.


asturur:
thank you!

PR #6163, opened 2020-02-20
Adding a safe check for _setImageSmoothing function to make sure that variable is not a null

BishoyBishai:


asturur:
How did you trigger this bug? fabric in node?

BishoyBishai:
> How did you trigger this bug? fabric in node?

We use fabric in our project and this bug has been trigged by Sentry.

asturur:
what is Sentry?

BishoyBishai:
> what is Sentry?

@asturur [Sentry](https://sentry.io/about/)

PR #6166, opened 2020-02-21
hoaily add enableKerned for text rendering like font-kerning

hoaily:
I just add enableKerned attribute to allow using kernedWidth or only width of character.

First add this code to define new attribute
/**
 * Attribute measure character enable or disable kerned
 * @Author: Hoai Ly(lyhoai2006tc@gmail.com)
 * @type boolean
 */
fabric.enableKerned = true;

Then, modify the returning of _measureChar method:
- From: 
return { width: width * fontMultiplier, kernedWidth: kernedWidth * fontMultiplier };

- To:
if (fabric.enableKerned) {
          return { width: width * fontMultiplier, kernedWidth: kernedWidth * fontMultiplier };
      } else {
          return { width: width * fontMultiplier, kernedWidth: width * fontMultiplier };
      }


asturur:
Hi @hoaily , would explain the reason of not using the kernedWidth? is this for some specific font, language or feature?

PR #6195, opened 2020-03-03
Check font faces markup for objects within groups

yassipad:
Currently, the "createSVGFontFacesMarkup" is only looking for fontFamilies among objects of the canvas. This PR allows it to check for fontFamilies within (nested) groups.


test/unit/itext.js
```diff
@@ -676,6 +676,46 @@
...
+      assert.equal(typeof iText2.toSVG, 'function');
+      var parser = new DOMParser();
+      var svgString = canvas.toSVG(),
+          doc = parser.parseFromString(svgString, 'image/svg+xml'),
+          style = doc.getElementsByTagName('style')[0].firstChild.data;
+      assert.equal(
+        style,
+        '\n\t\t@font-face {\n\t\t\tfont-family: \'Plaster\';\n\t\t\tsrc: url(\'path-to-plaster-font-file\');\n\t\t}\n\t\t@font-face {\n\t\t\tfont-family: \'Engagement\';\n\t\t\tsrc: url(\'path-to-engagement-font-file\');\n\t\t}\n'
+      );
+    });
```
asturur: i would ask you if you could make up 2 additional font names. We have Plaster and Engagement, used in style and object props. If we had 4 names instead of 2, we could be sure that everyone is captured.

asturur:
Looks good to me to merge, if you could update the test i would appreciate.

asturur:
great thanks!

PR #6198, opened 2020-03-03
Register + store events

yassipad:
This is yet another feature i've been adding for my own project and thought this may be useful for others, so here it is:

- This PR allows for registering "global" events, which can then be reused simply by name across shapes:
```js
fabric.util.registerEvent('alert', function(event, object) {
  alert(`You have triggered the alert event on object "${object.name}" at position { x: ${event.e.x}, y: ${event.e.y} }`)
})
```

- This PR allows for storing events in the JSON representation of an object with an extra "trigger" option for custom interaction on a given event:
```js
object.addEvent('alert', 'mousedown')
fabric.Object._fromObject('Object', object.toJSON(), function(newObject) {
  newObject.events => [{
    name: 'alert',
    trigger: 'mousedown'
  }]
})
```
Here is a working example: https://jsfiddle.net/brfojpq0/1/

Tests are included for registering events and storing events.

PS: sorry for all the PRs, i'm just sharing my own extensions of your amazing package

### DIFF
--- a/build.js
+++ b/build.js
@@ -135,6 +135,7 @@ var filesToInclude = [
   'src/util/dom_style.js',
   'src/util/dom_misc.js',
   'src/util/dom_request.js',
+  'src/util/register_events.js',
 
   'src/log.js',
 
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -627,6 +627,12 @@
     absolutePositioned: false,
 
     /**
+     * List of stored events.
+     * @type array
+     */
+    events: [],
+
+    /**
      * Constructor
      * @param {Object} [options] Options object
      */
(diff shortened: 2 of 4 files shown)


asturur:
I m not sure i want do this or the animation one.
FabricJS stores in JSON whatever is necessary to restore visual representation of the canvas ( controls excluded )
This new things would bring more features of which we are full already and we strive to mantains quality/compatibility.

Also, does this store functions in JSON? i just don't get it from the code.

yassipad:
Hi, 
Yes I see your concern and totally get it. I’m personally using events/animations storing to create interactive canvases that can be recreated from JSON directly. And no it doesn’t store the events, it only stores the name of the event and its trigger, the list of available events is extended using the registerEvents method. This way I can simply add an array of events in my json and call the corresponding method on the client. I find it quite useful but it doesn’t have to be in the core package :)

PR #6199, opened 2020-03-03
Trigger events on objects within group

yassipad:
Currently, events such as "added" or "removed" are not triggered on the objects contained in a group, only on said group. This PR adds a "triggerObjectsEvents" property on the fabric.Group class in which an array of event names can be passed to trigger the desired events on its objects as well.

```js
const object = new fabric.Object({})
object.on('added', () => console.log('object was just added'))
const group = new fabric.Group([object], {
    triggerObjectsEvents: ['added', 'removed', 'customEvent']
})
```

Simple tests are included.


src/shapes/group.class.js
```diff
@@ -195,6 +219,9 @@
     _onObjectAdded: function(object) {
       this.dirty = true;
       object.group = this;
+      if (!object.canvas && this.triggerObjectsEvents.indexOf('added') > -1) {
+        object.fire('added');
+      }
```
asturur: this is a different fix i guess, and i bet is missing the remove counterpart?

src/shapes/group.class.js
```diff
@@ -100,6 +108,22 @@
...
+     */
+    _triggerObjectsEvents: function () {
+      var _this = this;
+      _this.triggerObjectsEvents.forEach(function (eventName) {
+        _this.on(eventName, function (event) {
+          _this.forEachObject(function (object) {
+            object.fire(eventName, event);
+          });
+        });
+      });
```
asturur: does this have to be in the library? eventually in my app i can add an events for each group created that does exactly this.

I can wrap my contstructor in a new one that adds this extra functionality:
for each event i m interested in, i can add to each group an event handler that will fire for each object the same event.

```js
MyGroup.on('eventX', function(opts) {
  this.forEachObject(function(object) { object.fire('eventX', opts) });
})
```

src/shapes/group.class.js
asturur: would make sense that each object that gets added either to canvas or to group fire added. For canvas happens, for groups not.

yassipad: No actually this is the same fix, it’s just that I was looking for a common « add » method that would also be called on group initialization if an array of objects is passed as first argument but realized it wasn’t the case. So, basically, if objects are passed in initialization, the added event will get fired as soon as the group is added, but if you then manually add another object and this object wasn’t already on the canvas, it will also trigger the added event as expected

yassipad: Oh and no I didn’t add the removed counterpart because if I understand correctly, if an object gets removed from the group it doesn’t necessarily mean that it gets removed from the canvas? Is that correct?

src/shapes/group.class.js
yassipad: Yes of course, i just thought it’d be nicer and more consistent to only have to set the event names to listen to directly on the object. Though I realize in terms of consistency that one would expect calling set(‘triggerObjectsEvents’, [´smth’]) to modify the listeners accordingly.

Issue #6065, opened 2020-01-02
Cannot read property 'bringToFront' of undefined


paulconlin:
There seems to be cases where bringToFront can be called when this.canvas is undefined. It's tricky to reproduce, but it seems like  this.canvas just needs an existence check before bringToFront is called. Let me know if you think this should be fixed on my end.

Thanks!

![image](https://user-images.githubusercontent.com/10995964/71682451-5e8d2580-2d55-11ea-978d-c8803ba9b37c.png)

asturur:
I wonder how. I m ok with a safeguard, but i also would like to figure out how you got there.
There may be a real bug somewhere else

paulconlin:
I have it in a mouse:down listener, so I wonder if i'm removing the canvas and the listener is still firing after that 
```
    fabricCanvas.on('mouse:down', options => {
      if (fabricCanvas.isDrawingMode) this.setControlsVisibility(false);

      // Bring to front on click if not drawing
      if (!fabricCanvas.isDrawingMode && options.target && options.target.__corner === 0) options.target.bringToFront();
    });
```

asturur:
Well it does not look like you are removing the canvas. Is strange that the canvas is not there.

paulconlin:
I am doing some very non-typical things with it though, where the canvas is used during certain steps of our recording flow, and then is removed during playback. You can see what I'm doing here - https://kinetoscope.flipgrid.com/ 

Fabric powers all of the drawing and sticker management on the recording step.

asturur:
Yes so probably in the process of moving an object from a canvas to another something is off.

(anyway the bring to front on click is an embedded feature called `preserveObjectStacking` i m not sure if is something that is usefull for you that avoid the custom code).

asturur:
i'm ok with adding a `else if (this.canvas) {`
I'm also ok to remove the feature later completely, since the idea of having the same functionality on both canvas and objects, seems to be bad on the long run.

As a dev calling bringToFront of an object and having it reuse either the canvas function to bring it up, or the same function but called with the group as context ( so that is brought top in the group stack ) seems just a magical confusion.

As today, as opposed as me of 4 years ago, i would just kill that method completely.

paulconlin:
Good to know about `preserveObjectStacking` and thanks as always for your prompt responses!

asturur:
if you have time for adding the guarding if, please open a pr.

paulconlin:
PR opened!

PR #6207, opened 2020-03-06
Adding existence check for this.canvas on object stacking mixins

paulconlin:
Looks to resolve Cannot read property 'bringToFront' of undefined #6065

I was having instances where the canvas instance was null and bringToFront was being called. I added this existence check to all of the object stacking mixins and matches the approach to check for `this.group` in the if block above.


asturur:
you forgot all the if after the else keyword. Indeed is not building.

asturur:
you can also have `this.canvas && this.canvas.bringTo...` if you want.
Is the same for me, the difference in bytes saved is minimal

paulconlin:
🤦‍♂️sorry about that, updating now

paulconlin:
Opted to stay with the else block, but can update to `this.canvas &&` if you'd like

asturur:
is fine as it is!

PR #6211, opened 2020-03-09
 fix aligning guideline position issue with scaled canvas

jhondge:
fix(guidelines): fix guideline position issue with scaled canvas

### DIFF
--- a/lib/aligning_guidelines.js
+++ b/lib/aligning_guidelines.js
@@ -33,9 +33,13 @@ function initAligningGuidelines(canvas) {
     ctx.save();
     ctx.lineWidth = aligningLineWidth;
     ctx.strokeStyle = aligningLineColor;
+    var v = viewportTransform;
+    ctx.transform(v[0], v[1], v[2], v[3], v[4], v[5]);
     ctx.beginPath();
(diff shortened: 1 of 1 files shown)


lib/aligning_guidelines.js
```diff
@@ -33,9 +33,13 @@ function initAligningGuidelines(canvas) {
     ctx.save();
     ctx.lineWidth = aligningLineWidth;
     ctx.strokeStyle = aligningLineColor;
+    var v = viewportTransform;
+    ctx.transform(v[0], v[1], v[2], v[3], v[4], v[5]);
```
asturur: this will make the guidelines stroke size grow with zoom.
What is wrong in the math  down?
probably viewportTransform[4] and [5] shouldn't be multipled by zoom?

jhondge: yes, when you set the scale not equal 1, the guidelines will be draw on incorrect position. I think if the stroke size will grow with zoom, we can dynamic change the line width?

asturur: I think we should fix the actual math to respect he matrix transformation.

GreatAuk:
same issue, need help...

PR #6222, opened 2020-03-18
fix ISSUE-6220

Swedish-li:
parse font property (start with font size) error

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -134,7 +134,7 @@
         value = 'stroke';
       }
     }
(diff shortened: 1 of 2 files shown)


asturur:
Thanks for the PR.
Do you have any link to the specs for the css font attribute on svg?
It should be the same as standard html, but better check.

Swedish-li:
this is the same as css specs.

https://www.w3.org/TR/SVG11/text.html#FontFamilyProperty

asturur:
well seems fine to me then! Thanks.

PR #6238, opened 2020-03-29
Added more error flag passing throughout functions

### LINKED ISSUE #6227: Better error handling for Image.fromURL
**Is your feature request related to a problem? Please describe.**
The problem is that some functions swallow errors.

For example, the function `Image.fromURL`
Doc: http://fabricjs.com/docs/fabric.Image.html#.fromURL
Source: http://fabricjs.com/docs/fabric.js.html#line20735

The line 20736 in the code clearly shows that the callback passed to `fabric.util.loadImage()` ignores the error. I can open a PR to fix that if needed.

Also, that would be nice to respect the JS standards and having an `Error` as first argument and the `Image` as second argument, instead of `Image` as first argument and a `boolean` as second argument. Moreover this boolean seems to not be documented.

And that would be even more nice to have all functions taking callback to have the callback as the last argument, but anyway. Those deeper changes (order of arguments) would introduce massive breaking changes, so of course that's not doable.

**Describe the solution you'd like**
The solution would be to change http://fabricjs.com/docs/fabric.js.html#line20736 to add the `isError` argument, and change the code as follow:
```js
fabric.Image.fromURL = function(url, callback, imgOptions) {
    fabric.util.loadImage(url, function(img, isError) {
        if (!callback) {
            return;
        }
        if (isError) {
            callback(null, true);
        } else {
            callback(new fabric.Image(img, imgOptions), false);
        }
    }, null, imgOptions && imgOptions.crossOrigin);
}
```
or maybe still passing the image, and so doing the following:
```js
fabric.Image.fromURL = function(url, callback, imgOptions) {
    fabric.util.loadImage(url, function(img, isError) {
        callback && callback(new fabric.Image(img, imgOptions), isError);
    }, null, imgOptions && imgOptions.crossOrigin);
}
```
but I don't really see the point in instancing an empty image.

The implementation of `fabric.util.loadImage()` implies that solution 1 would be a better candiadate.

**Describe alternatives you've considered**
An alternative would be to pass an `Error` object as the second argument instead of a boolean, but that would break people who did `if (secondParam === true) { ... }`. Also, the `` element does not provide any useful information when raising the `onerror` callback, so maybe that's not really necessary.

**Additional context**
Because this feature is all about networking (loading image from a remote), errors are very likely to happen, therefore proper handling is critical.

Thank you for considering.

TanukiSharp:
Context: Some function are swallowing error flags and do not pass them to subsequent functions.
Related ticket: https://github.com/fabricjs/fabric.js/issues/6227

Closes #6227

### DIFF
--- a/src/pattern.class.js
+++ b/src/pattern.class.js
@@ -75,9 +75,9 @@
         // img src string
         var _this = this;
         this.source = fabric.util.createImage();
-        fabric.util.loadImage(options.source, function(img) {
+        fabric.util.loadImage(options.source, function(img, isError) {
           _this.source = img;
-          callback && callback(_this);
+          callback && callback(_this, isError);
         }, null, this.crossOrigin);
       }
     },
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -380,10 +380,10 @@
      * @chainable
      */
     setSrc: function(src, callback, options) {
-      fabric.util.loadImage(src, function(img) {
+      fabric.util.loadImage(src, function(img, isError) {
         this.setElement(img, options);
         this._setWidthHeight();
-        callback && callback(this);
+        callback && callback(this, isError);
       }, this, options && options.crossOrigin);
       return this;
     },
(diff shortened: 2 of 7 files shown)


asturur:
Seems good to me.

PR #6256, opened 2020-04-09
fix(text): incorrect cursor/selection position when text in group

jhondge:
There has a incorrect position issue when render the cursor/selection in group.

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -271,7 +271,19 @@
       this._clearTextArea(ctx);
       skipRestore || ctx.restore();
     },
-
+    transform: function(ctx) {
+      var m;
+      // the contextTop layer will not be transform in group render function. so we need
+      // manual apply the group transform for render cursor or selection which used the contextTop
+      if ((this.group && !this.group._transformDone) ||
+          (this.group && this.canvas && ctx === this.canvas.contextTop)) {
+        m = this.calcTransformMatrix();
+      }
+      else {
+        m = this.calcOwnMatrix();
+      }
+      ctx.transform(m[0], m[1], m[2], m[3], m[4], m[5]);
+    },
     /**
      * Renders cursor or selection (depending on what exists)
      * it does on the contextTop. If contextTop is not available, do nothing.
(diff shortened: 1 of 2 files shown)


src/shapes/itext.class.js
```diff
@@ -271,7 +271,19 @@
...
+      // manual apply the group transform for render cursor or selection which used the contextTop
+      if ((this.group && !this.group._transformDone) ||
+          (this.group && this.canvas && ctx === this.canvas.contextTop)) {
+        m = this.calcTransformMatrix();
+      }
+      else {
+        m = this.calcOwnMatrix();
+      }
+      ctx.transform(m[0], m[1], m[2], m[3], m[4], m[5]);
+    },
```
asturur, line 286: i wonder if we can generalize this in the main Object class.
Does it make sense to duplicate it?
This is the standard function
```
    /**
     * Transforms context when rendering an object
     * @param {CanvasRenderingContext2D} ctx Context
     */
    transform: function(ctx) {
      var m;
      if (this.group && !this.group._transformDone) {
        m = this.calcTransformMatrix();
      }
      else {
        m = this.calcOwnMatrix();
      }
      ctx.transform(m[0], m[1], m[2], m[3], m[4], m[5]);
    },
```

it can be changed in the Object class in this way:

```js
    /**
     * Transforms context when rendering an object
     * @param {CanvasRenderingContext2D} ctx Context
     */
    transform: function(ctx) {
      var needFullTransform = (this.group && !this.group._transformDone) ||
         (this.group && this.canvas && ctx === this.canvas.contextTop);
      var m = this.calcTransformMatrix(!needFullTransform);
      ctx.transform(m[0], m[1], m[2], m[3], m[4], m[5]);
    },
```

We would need at least a visual test for it of some sort anyway.

asturur:
@jhondge any chance to fix the comment? Do you agree with it?

jhondge:
> @jhondge any chance to fix the comment? Do you agree with it?

yeah, of course.

asturur:
@jhondge any chance to follow up on the pr?

asturur:
thank you!

PR #6259, opened 2020-04-10
docs: Fix simple typo, usefull -> useful

timgates42:
There is a small typo in dist/fabric.js, src/canvas.class.js, src/mixins/text_style.mixin.js.

Should read `useful` rather than `usefull`.


asturur:
Thanks!

PR #6296, opened 2020-04-25
feat(fabric.Pattern): remove the ability to restore patterns from functions.

### LINKED ISSUE #1621: Fabric js is not csp unsafe-eval compliant

asturur:
close #1621

BREAKING: patterns that uses functions are no more supported

Pattern will transition to fabric v4, but will be sort of deprectated.
A new Pattern will be created, that will allow to define patterns with fabric objects.

### DIFF
--- a/src/pattern.class.js
+++ b/src/pattern.class.js
@@ -66,11 +66,6 @@
         callback && callback(this);
         return;
       }
-      // function string
-      if (typeof fabric.util.getFunctionBody(options.source) !== 'undefined') {
-        this.source = new Function(fabric.util.getFunctionBody(options.source));
-        callback && callback(this);
-      }
       else {
         // img src string
         var _this = this;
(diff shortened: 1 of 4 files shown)


sz5000:
This changes breaks fabric.PatternBrush. 

brush = new fabric.PatternBrush(canvas);

... use the brush

results in an browser error

this is because this code is removed:

// function string
      if (typeof fabric.util.getFunctionBody(options.source) !== 'undefined') {
        this.source = new Function(fabric.util.getFunctionBody(options.source));
        callback && callback(this);
      }

this code is required if you use the getPatternSrc function of fabric.PatternBrush

Issue #6213, opened 2020-03-10
Textbox can't be edited after adding to another canvas


shanicerae:
## Version
3.6.0

## Test Case
http://jsfiddle.net/2m0gvbc1/18/

## Information about environment
Browser (Chrome)

## Steps to reproduce
Edit a textbox in a canvas.
Remove said textbox from canvas, create new canvas, and then add said textbox to new canvas.
-> User can only type on initial click. This is because the `hiddenTextarea` on the Textbox does not get focused upon second click. This seems to be a side-effect of `onClick` not getting bound to the second canvas because `_clickHandlerInitialized` on the Textbox is false.

## Expected Behavior

User should be able to type in a textbox that has been removed from a canvas and added to another canvas. A textbox's `hiddenTextarea` should get focused upon every click and a textbox should have `_clickHandlerInitialized` set to false when removed from the canvas that the click handler was added to.

## Actual Behavior

The `hiddenTextarea` on a Textbox doesn't always focus on click causing users to not be able to type in a textbox that has been added to canvases other than the first one that `initHiddenTextarea` was called in.

asturur:
hi @shanicerae thanks for pointing out the issue in so much detail.
I ll give it a look asap.

asturur:
Sorry for not getting back sooner.

ShaMan123:
ancient and seems fixed

PR #6382, opened 2020-06-11
Set _clickHandlerInitalized to false when adding iText to canvas

shanicerae:
A fix for issue #6213 .

Sets `_clickHandlerInitialized` on the iText instance to false when it gets added to the canvas. This handles the case when an iText instance gets added to more than one Canvas instance - in this case, the listener needs to be initialized again.

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -28,7 +28,7 @@
       this.on('added', function() {
         var canvas = _this.canvas;
         if (canvas) {
-          _this._restoreClickHandlerProps();
+          _this._clickHandlerInitialized = false;
 
           if (!canvas._hasITextHandlers) {
             canvas._hasITextHandlers = true;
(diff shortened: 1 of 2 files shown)


src/mixins/itext_behavior.mixin.js
```diff
@@ -28,6 +28,8 @@
       this.on('added', function() {
         var canvas = _this.canvas;
         if (canvas) {
+          _this._restoreClickHandlerProps();
```
asturur, line 31:
```suggestion
          this._clickHandlerInitialized = false;
```

I would love to have just this and a small test that can verify it works. Do we need to check if the canvas is the exact same canvas?

asturur: I also wonder if today we do multiple registarations.

shanicerae: @asturur Sorry for taking so long to get back on this. I've pushed the suggested change with an update to the test. To answer about checking if the canvas is exactly the same, I don't think this is necessary as I believe the listener would need to be added regardless of if the canvas was the same as previous

PR #6390, opened 2020-06-17
Animation logic improvement: testing concept

asturur:
- Add to canvas a `renderAtTime` method. that will work exactly as `renderAll` but invoking for each object the `renderAtTime` method instead the `render` one

- The object `renderAtTime` method will first calculate the new object properties and then render

- is duty of thedeveloper to call `this.saveState({ propertySet: 'animatableProperties' });` to copy the original values and to restore them calling `this.restoreState({ propertySet: 'animatableProperties' });`

- out of scope but possible: animation canvas viewport ( zoom, panning )

Possible animation format:

```js
fabric.Rect.animation = [
{ property: fill, value: 'red', offset: 33, duration: 4000, easing: 'cubic' },
{ property: scaleX, value: 4, offset: 33, duration: 4000, easing: 'linear' },
{ property: scaleX, value: 1, offset: 4033, duration: 4000, easing: 'linear' },
]
```

### DIFF
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -4,6 +4,10 @@
     return false;
   }
 
+  function identityEase(currentTime, startValue) {
+    return startValue;
+  }
+
   function defaultEasing(t, b, c, d) {
     return -c * Math.cos(t / d * (Math.PI / 2)) + c + b;
   }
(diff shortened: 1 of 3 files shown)


src/util/animate.js
```diff
@@ -90,7 +94,69 @@
...
+      if (current.offset > time) {
+        // this aniamtion has to be started yet, ignore
+        return accumulator;
+      }
+      var property = current.property;
+      // at this point however it ends, i need a key for a valid animation
+      if (!accumulator[property]) {
+        accumulator[property] = {
+          previouslyEndedAt: 0,
+          previousState: object[property]
```
asturur, line 122: this is fine moving forward in time, but has problems moving backward.

lyzs90:
how would i go about specifying the t=0 state of an object's prop i.e. the first keyframe?

asturur:
That would be the canvas state. So the canvas would be the t=0, and all the animations will play as difference from t=0.

There are things to understand that are clearly not strictly part of the library, but are more or less:
- how do i enable a developer/user to move objects in the middle of a non keyframe with the stanadard interactions?
- how do i make possible identify the current animation i m rendering for a given time ?

lyzs90:
> That would be the canvas state. So the canvas would be the t=0, and all the animations will play as difference from t=0.
this sounds fair

> how do i enable a developer/user to move objects in the middle of a non keyframe with the stanadard interactions?
i see i'm looking at how this is handled in [After Effects](https://youtu.be/6tR3fpv4Aco?t=742). they require you to select which keyframe to operate on, so you are never modifying an object in between keyframes

PR #6414, opened 2020-06-27
fix(fabric.Image): ISSUE-6397 modify crossOrigin behaviour for setSerc

### LINKED ISSUE #6397: issue with setSrc for crossorigin images
Hello,
i have some trouble with changing the src of an image.

As you can see in the fiddle, i want to change the src of an image via the `setSrc` function.
After calling `canvas.renderAll` in the `setSrc`-callback nothing happens.
The image will be loaded twice (you can see it in the network panel of the chrome devtools). So the callback will be triggered after the first load, but the image will be rendered correctly after the image is loaded the second time. Also all image options like width and height will be reseted.

This problem occurs only if i set the crossOrigin option (to anonymous).

http://jsfiddle.net/v6e0hd5o/1/

Tested with fabricjs v. 4.0.0-beta.12 on chrome 83.

asturur:
Image.crossOrigin does not make sense.
crossOrigin is a property of the image element, and is needed to load and reload from object the image.
The developer should not be able to change it at will.

changes:
- add getCrossOriginMethod to read only the actual value
Breaking:
- removing setCrossOrigin method
- avoid setSrc to change element crossOrigin after load.

close #6397

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -30,14 +30,6 @@
     type: 'image',
 
     /**
-     * crossOrigin value (one of "", "anonymous", "use-credentials")
-     * @see https://developer.mozilla.org/en-US/docs/HTML/CORS_settings_attributes
-     * @type String
-     * @default
-     */
-    crossOrigin: '',
-
-    /**
      * Width of a stroke.
      * For image quality a stroke multiple of 2 gives better results.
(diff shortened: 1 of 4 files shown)


asturur:
this introduce a breaking change, so i cannot backport it. The Pr is easily cherry-pickable

PR #6454, opened 2020-07-11
fix(fabric.IText): copy style in non full mode when typing text

### LINKED ISSUE #6426: `IText` doesn't get updated on canvas after style changes
## Version
4.0.0-beta.12

## Test Case
In fabric 3.x.x, if I I change a style property on an IText (say, `myIText.set('fill', hslColor);`), and request a render, the element gets updated on the canvas. On 4.0 it does not. 

On 4.0, If I explicitly call `myIText.cleanStyle('fill')` before changing the style, the changes are applied.

Browsers, tested on Chrome, Edge & Firefox

## Steps to reproduce
1. Create an IText and requestRender
1. After the object is in the canvas, change some visual property, such as font family or color.
1. Request a new render

## Expected Behavior
New style property should be visible on existing text

## Actual Behavior
No visible changes on existing text

### LINKED ISSUE #6436: Font update  - does not work after new Text input
## Version
3.6.3

## Test Case
https://youtu.be/Ka-lTABJXjg

Chrome Browser  83.0.4103.116

## Steps to reproduce
1 select Textbox with a text, 
2. make sure cursor inside Textbox
3. ctrl+a to mark all Text
4. overwrite the old Text with some new words (by marking the old text)
5. After new input, you are not able to change font-Family anymore or other Font settings like font size. Shadows still works.

## Expected Behavior
update font settings

## Actual Behavior
Does not update Font Settings anymore

asturur:
Copying style without using the full properties available, will avoid creating extra props that make the text completely styled and not reacting anymore to the generic properties of the text object.

close #6436 
close #6426

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -189,6 +189,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
         // we want to copy the style before the cursor OR the style at the cursor if selection
         // is bigger than 0.
         copiedStyle = this.getSelectionStyles(selectionStart, selectionStart + 1, false);
+        console.log(copiedStyle)
         // now duplicate the style one for each inserted text.
         copiedStyle = insertedText.map(function() {
           // this return an array of references, but that is fine since we are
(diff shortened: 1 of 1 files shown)


rlidev:
Thanks ;)

PR #6491, opened 2020-07-31
Try to update gesture module to v4

asturur:
Update gesture handlers to use the scaling code, so that event emitter and lock control is shared.
Added to the control handler the extra code needed to handle a. gesture scaling.


juzhiyuan:
Not sure why failed on Node 14

PR #6492, opened 2020-08-01
feat(Brushes): add beforePathCreated event

### LINKED ISSUE #6486: Can the path object generated by the PencilBrush during mouseup be returned to the outside to customize some attributes?
I am using fabricjs to build a canvas for multi-person collaboration.
When one of the users is free drawing, the point obtained during mousemove is transmitted to other clients for drawing at the same time.

I hope that the path object generated by PencilBrush or other brushes in `_finalizeAndAddPath` will be exposed to the outside world to customize some attributes such as setting id through `.set`.

It is too late to set the id in the `path:created` event, because the `id` of this element needs to be kept the same for all clients. Because I can’t use `canvas.contains` on other clients to determine the element to unify the `id`.

So I expect whether the newly created path object can be exposed to the outside world during `onMouseUp`？ https://github.com/fabricjs/fabric.js/blob/6c28f5a0b6d0fd9d30d338770fbacce8f419c69f/src/brushes/pencil_brush.class.js#L90

Looking forward to your reply, thanks very much.

linrz:
Related to https://github.com/fabricjs/fabric.js/issues/6486, before the path added into canvas fire `before:path:created`. 

close #6486


test/unit/brushes.js
```diff
@@ -76,7 +79,7 @@
           brush.onMouseMove(pointer2, { e: {} });
           brush.onMouseMove(pointer3, { e: {} });
           brush.onMouseUp({ e: {} });
-          assert.equal(fired, true, 'event is fired');
```
asturur, line 79: please make a specific variable that check both firing, not the sum of both.

asturur:
PR looks good if you address the test

asturur:
If you merge the latest master in your branch, the error on node14 fill get fixed.

asturur:
perfect!
I'm releasing 4.0 now because i need to move forward, then i ll merge this right after, will be in 4.0.1 that we can publish quickly

PR #6504, opened 2020-08-07
feat(fabric.Path): Change the way path is parsed and drawn

asturur:
We want to simplify path code in order to make space for text on path ( curved text ) 🎉 

First step of this work is to remove from the path:
all the non absolute commands, all the Arcs, and V and H.

Less command types will also mean less drawing and bounding box code.


src/util/path.js
```diff
@@ -1,17 +1,30 @@
 (function() {
-
   var _join = Array.prototype.join;
 
+  function segmentToBezier(th2, th3, cosTh, sinTh, rx, ry, cx1, cy1, mT, fromX, fromY) {
+    var costh2 = fabric.util.cos(th2),
```
juzhiyuan, line 5: Hi, just for sure, this PR means the fabric.js will support Bezier Curve soon?

asturur: What do you mean? fabric always drawn beizer curves but does not give you the ability to interact and modify them.

asturur:
OK UT pass, but i first need a visual regression test PR to be merged on the current master. and run this gain

src/shapes/path.class.js
```diff
@@ -354,36 +205,6 @@
...
-              current[2],
-              current[3],
-              current[4],
-              current[5],
-              current[6] + l,
-              current[7] + t
-            ]);
-            x = current[6];
-            y = current[7];
-            break;
```
asturur, line 385: @melchiar i wrote code in the file util/path.js to handle 'a' and 'A' at parse time, and also all the lower case commands.
I would like to get rid in this function also of case 'T' and case 'S'. and keep 'C' and 'Q'.
An 'S' would become a paricular different 'C' and a 'T' a particular different 'Q'.

The drawing code in this function explains how. and the code should be added in util/path.js function makePathSimpler. The transtion from s to S and to t to T is already handled.

src/util/path.js
juzhiyuan: Oh yes, I mean if this PR could give us ability to modify shapes with Becker Curves.

asturur: This PR aims at removing everything that is not M L C Q from complex path. so that is easier to write code to handle paths manually.
So yes, it will help definitely with you text/path thing.

PR #6507, opened 2020-08-09
feat(fabric.Path) Simplify S and T command in C and Q.

asturur:
More path semplification code.

Add C and Q counterpart of S and T.
Removed respective bounding box and render code.


test/visual/svg_import.js
```diff
@@ -38,6 +41,10 @@
 
   QUnit.module('Simple svg import test');
 
+  function identity (a) {
+    return a;
```
juzhiyuan, line 45: Hi, what's this function's ability?

asturur:
i use it as a quick hack to filter out tests that have no code with
array.filter

i had some problem with a test and needed a solution but is temporary i
have to solve the test before merging

asturur:
Ok this PR was a pain. Lots of hidden hard to spot small mistakes during development.

PR #6516, opened 2020-08-12
Draft of measurePath

melchiar:
Began the draft of fabric.util.measurePath function for calculating the length of a pre-simplified path (using `makePathSimpler`)

TODO: still needs methods for C and Q


src/util/path.js
```diff
@@ -395,6 +395,41 @@
...
+              y1 = current[2];
+            break;
+            case 'C':
+              //todo
+            break;
+            case 'Q':
+              //todo
+            break;
+        }
+    }
```
asturur, line 430: you probably have lint issue with those 4 spaces indent.

melchiar: hmm... i'll take a look

asturur: build is failing at lint step https://github.com/fabricjs/fabric.js/pull/6516/checks?check_run_id=977074861

melchiar: should be good now

melchiar:
@asturur has something stalled the tests?

asturur:
did not check, let me see.

asturur:
github must have had some probelm with the checks, the build passed.

PR #6525, opened 2020-08-16
feat(fabric.util): Add a function to work with path measurements

asturur:
This function ( currently untested ) should provide us a way to measure paths.

### DIFF
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -396,39 +396,184 @@
     return destinationPath;
   };
 
+  /**
+   * Calc lenght from point x1,y1 to x2,y2
+   * @param {Number} x1 starting point x
+   * @param {Number} y1 starting point y
+   * @param {Number} x2 starting point x
+   * @param {Number} y2 starting point y
+   * @return {Number} length of segment
+   */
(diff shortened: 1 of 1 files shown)


src/util/path.js
```diff
@@ -396,39 +396,192 @@
...
+    return 3 * t * t * (1 - t);
+  }
+  function CB3(t) {
+    return 3 * t * (1 - t) * (1 - t);
+  }
+  function CB4(t) {
+    return (1 - t) * (1 - t) * (1 - t);
+  }
+
+  function getPointOnCubicBezierIterator(P1x, P1y, P2x, P2y, P3x, P3y, P4x, P4y) {
```
asturur, line 426: need to make those lowercase.

src/util/path.js
```diff
@@ -396,39 +396,192 @@
     return destinationPath;
   };
 
+  /**
+   * Calc lenght from point x1,y1 to x2,y2
```
juzhiyuan, line 400: Length please

asturur: i spelled it good almost everywhere, after various checks and corrections 😂

PR #6545, opened 2020-08-24
Fire resizing event for textbox width

### LINKED ISSUE #6544: Resizing textbox width doesn't fire the scaling event
## Version
4.1.0

Browser

## Explanation of issue
Resizing a textbox with ml and mr isn't firing the scaling event as it did pre-4.0

melchiar:
Made the textbox `changeWidth()` function fire the scaling event. Function returns false if no scaling has occured. Fixes #6544

### DIFF
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -31,8 +31,8 @@
   function fireEvent(eventName, options) {
     var target = options.transform.target,
         canvas = target.canvas,
-        canasOptions = Object.assign({}, options, { target: target });
-    canvas && canvas.fire('object:' + eventName, canasOptions);
+        canvasOptions = Object.assign({}, options, { target: target });
+    canvas && canvas.fire('object:' + eventName, canvasOptions);
     target.fire(eventName, options);
   }
 
@@ -140,7 +140,7 @@
   }
 
   /**
-   * Inspect event , control and fabricObject to return the correct action name
+   * Inspect event, control and fabricObject to return the correct action name
    * @param {Event} eventData the javascript event that is causing the scale
    * @param {fabric.Control} control the control that is interested in the action
    * @param {fabric.Object} fabricObject the fabric object that is interested in the action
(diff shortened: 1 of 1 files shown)


src/controls.actions.js
```diff
@@ -680,9 +680,14 @@
...
         multiplier = isTransformCentered(transform) ? 2 : 1,
+        oldWidth = target.width, hasScaled,
         newWidth = Math.abs(localPoint.x * multiplier / target.scaleX) - strokePadding;
     target.set('width', Math.max(newWidth, 0));
-    return true;
+    hasScaled = oldWidth !== newWidth;
+    if (hasScaled) {
+      fireEvent('scaling', commonEventInfo(eventData, transform, x, y));
+    }
+    return hasScaled;
```
asturur, line 690: i would swear it was firing `resizing` did i dream of it?

asturur: Must have been some conversation with some developer, but i would prefer not to fire scaling since scale is actually not change, can we call the event something else and you subscribe to both? we also may have extra events somewhere after an action is performed. does `resizing` sounds good?

asturur:
We need anyway a little test and the JSDOC where the other events are declared

src/controls.actions.js
melchiar: Sure, either way works for me. If we do change it to resizing we should list it in the breaking changes doc.

asturur:
can we have a little test? We have a test that tests the resize width, extending to verify the event is fired should be easy.

melchiar:
sure, I assume it'd be in the unit\controls_handlers.js test? Do we have docs for that somewhere?

asturur:
i think you can mirror this test here:
https://github.com/fabricjs/fabric.js/blob/master/test/unit/controls_handlers.js#L126
and in the upper part of the file you find the changeWidth test from which you can infer how to call the function.

asturur (approved):
i ll add the test, i actually need to merge this.

melchiar:
Yeah sorry I hadn't gotten to that

PR #6557, opened 2020-08-30
upgrade dev dependencies based on npm audit suggestions

TiraO:
I ran `npm audit fix` and then discarded the package-lock.json. I ran tests / build and all appeared to pass.


asturur:
seems good to me, being all dev deps, if the pipeline pass we are good.

asturur (approved):

PR #6560, opened 2020-08-31
fix(fabric.staticCanvas): return clipTo for canvas

andrewzolotukhin:
Hi @asturur , 

Now I'm trying to upgrade to the V4 in my editor (CleverBrush.com). 
In new version you have removed clipTo() for canvas. I was using it to clip the canvas when user zooms/pans inside the editor.
I see in the upgrade notes that you consider using clipPath instead of it, but it looks to me like a different (useful as well) feature.
If I set clipPath to the canvas it will be eventually saved to the scene JSON which I don't really wan't to happen, that's why this clipTo solution was perfect for me. 
Probably there is another correct solution that I'm not aware of.

I tried to reimplement the clipTo functionality by subscribing to before:render on canvas, but it doesn't work because I can't do a ctx.restore() (see line 912-914 at static_canvas.class.js).

Could we return clipTo for canvas only? In this PR I just reverted clipTo only for canvas.

Thank You,
Andrew


adamgamble:
👍 This also would help out some projects that i'm working on. Would love to see this merged.

dist/fabric.js
```diff
@@ -555,7 +555,12 @@ fabric.CommonMethods = {
       this._setObject(key);
     }
     else {
-      this._set(key, value);
+      if (typeof value === 'function' && key !== 'clipTo') {
+        this._set(key, value(this.get(key)));
+      }
+      else {
+        this._set(key, value);
+      }
```
asturur, line 563: why are we restoring this?

asturur:
ClipTo was deprecated since 2.4, one of the main reason was that it wasn't easy to nest clip area and that was unsafe to save and restore.
There are alternatives tho.
Did you try to do a save + clipTo in `before:render` and a restore in `after:render`?

What are you doing with clipTo that you can't do with a global composite operation later?

andrewzolotukhin:
Thanks for your help, @asturur , I'll close this PR, your solution with before:render and restore on after:render seems works for me.
But I think it would be great if we have some kind of internal context clipping in Fabric. Then we'd just skip render for objects that are off of the current clipping region. But I understand that it could be a challenging task, especially if clipping region has arbitrary form.

andrewzolotukhin:
@asturur , I just need to clip a canvas a bit when user zooms/pans. We have a 100% width <canvas> and instead of changing its size we just clip it depending of current zoom/pan.

asturur:
so you want a bleeding edge to define an inner canvas area, that is like your drawing board, but using full page canvas to host anyway controls and selection?

andrewzolotukhin:
Basically yes, it's cosmetic things, you can try here: https://www.cleverbrush.com/editor/v3pm0dopykntu7g94n
When you zoom out I clip background area. Also you can select objects which are "outside" the canvas, because selection is still working there. Also we could add a border with shadow around clipped area.
The sad thing that I clip controls. So if object is partly visible then controls that are ouside the clipping edge are not visible. 
That's  why I'm cosidering it to be a useful feature for Fabric.

asturur:
so i would suggest you to try a different approach.
Clipping will be active from the first shape drawn, on the canvas. so basically for every drawing operation you are going to calculate clip region and draw ( not that that is a perf hit, but just to add context ).

If you make a very small override to the renderOverlayMethod, where you call your function before or after the official code, you can paint a layer on top of your final canvas that acts as bleeding mask.
You can choose to have an hard clip or soft clip using alpha, you can leave controls in or outside of it, since controls can be drawn before or after the overlay.

you can use one of the effects here:
https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Compositing
(do not remember exactly which one) to draw on the objects outside the canvas only, without overdrawing the pre existing application background. 
You can use ctx.save/restore to draw this layer according to the canvas zoom ( you basically need to create a square donut with a square hole in it)

PR #6562, opened 2020-08-31
Ensure mouse corners match custom control corner size

### LINKED ISSUE #6556: Cursor range on custom control does not extend to cornerSize of control
## Version
4.1.0

## Test Case
http://fabricjs.com/custom-control-render

OS: Windows 10, 2004
Browser: Version 84.0.4147.135 (Official Build) (64-bit)

## Steps to reproduce
Move mouse over delete icon in either codepen. Move to top or bottom of X and note that the red area does not have the cursor icon but cursor icon is fine when within the X.

## Expected Behavior
Mouse should work over whole icon of control.

## Actual Behavior
Mouse only works within a certain range of the icon, likely tied to the object's corner size

## Potential fix?
Note that I was able to make this work better by replacing this code. I think it came originally from https://github.com/fabricjs/fabric.js/blob/master/src/mixins/object_interactivity.mixin.js
I also wonder if the touch corner should be handled but not sure on your thoughts. I also wasn't sure if this call was frowned upon (const controlObject = fabric.Object.prototype.controls[control]):
```js
    /**
     * Sets the coordinates of the draggable boxes in the corners of
     * the image used to scale/rotate it.
     * note: if we would switch to ROUND corner area, all of this would disappear.
     * everything would resolve to a single point and a pythagorean theorem for the distance
     * @private
     */
    _setCornerCoords: function() {
	var coords = this.oCoords,
		newTheta = fabric.util.degreesToRadians(45 - this.angle),
		cosTheta = fabric.util.cos(newTheta),
		sinTheta = fabric.util.sin(newTheta),
		/* Math.sqrt(2 * Math.pow(this.cornerSize, 2)) / 2, */
		/* 0.707106 stands for sqrt(2)/2 */
		cornerHypotenuse = this.cornerSize * 0.707106,
		touchHypotenuse = this.touchCornerSize * 0.707106,
		cosHalfOffset = cornerHypotenuse * cosTheta,
		sinHalfOffset = cornerHypotenuse * sinTheta,
		touchCosHalfOffset = touchHypotenuse * cosTheta,
		touchSinHalfOffset = touchHypotenuse * sinTheta,
		x,
		y;

	for (var control in coords) {
		x = coords[control].x;
		y = coords[control].y;
		// handle custom control corner sizes
		const controlObject = fabric.Object.prototype.controls[control];
		const cornerHyp = controlObject.cornerSize
			? controlObject.cornerSize * 0.707106
			: cornerHypotenuse;
		const cornerCosHalfOffset = controlObject.cornerSize ? cornerHyp * cosTheta : cosHalfOffset;
		const cornerSinHalfOffset = controlObject.cornerSize ? cornerHyp * sinTheta : sinHalfOffset;
		coords[control].corner = {
			tl: {
				x: x - cornerSinHalfOffset,
				y: y - cornerCosHalfOffset,
			},
			tr: {
				x: x + cornerCosHalfOffset,
				y: y - cornerSinHalfOffset,
			},
			bl: {
				x: x - cornerCosHalfOffset,
				y: y + cornerSinHalfOffset,
			},
			br: {
				x: x + cornerSinHalfOffset,
				y: y + cornerCosHalfOffset,
			},
		};
		coords[control].touchCorner = {
			tl: {
				x: x - touchSinHalfOffset,
				y: y - touchCosHalfOffset,
			},
			tr: {
				x: x + touchCosHalfOffset,
				y: y - touchSinHalfOffset,
			},
			bl: {
				x: x - touchCosHalfOffset,
				y: y + touchSinHalfOffset,
			},
			br: {
				x: x + touchSinHalfOffset,
				y: y + touchCosHalfOffset,
			},
		};
	}
};
```

gloriousjob:
Pertains to issue 6556. I've added code to ensure that the mouse will highlight over all parts of the custom corner size instead of the original corner size.
I also added a test based on just changing the size of one corner.

I'm not sure if more should be done for the touch code, as I'm unfamiliar with it.

This has morphed to also provide support for rectangular custom controls (a feature desired for 4.0 that didn't make it).
Closes #6556


asturur:
we are going to talk about this, i m just busy right now.

src/mixins/object_interactivity.mixin.js
```diff
@@ -88,24 +88,32 @@
           x, y;
 
       for (var control in coords) {
+        // handle custom control corner sizes
+        var controlObject = fabric.Object.prototype.controls[control];
+        var cornerHyp = controlObject.cornerSize
```
asturur, line 93: what about extending the controls to return their own corner shape? is a bit more complicated but would solve the problem of both different sized controls and corners that are rectangular.

controlObject.cornerSizeX and cornerSizeY and would also change the math down to something i can't calculate now on the top of my head. But definitely doable

gloriousjob:
@asturur What do you think about this? I wasn't sure how if you wanted cornerSize vs cornerSizeX, cornerSizeY. I also added touchCornerSizeX,Y. More still needs to be done but wanted your thoughts on which direction to take if you want to change some things.

asturur:
Hey sorry for not coming back.
I need this exact same feature, and i m a bit short of time.
I want to be sure this PR fix your use case and my use case and i m trying to get the chance to do this at work.

gloriousjob:
Just let me know if you have concerns. The chief thing I'm not sure about 
is handling rectangle vs. square since I didn't really need the rectangle 
for my square icon. But having one more prop doesn't really bother me.

asturur:
The idea looks good completely.
I m not entirely sure if we want to keep that function in the corner or in the canvas, it does not matter much.
Did you had a chance to try if the calculations are correct?

gloriousjob:
@asturur What do you mean by "try if the calculations are correct"? I adjusted the sizes and noticed that the mouse was properly changing the cursor when within the rectangle. I'm not sure if you mean just verified that things were working? I used basic trig to make the calculations.
I also did a recent update when I found that doing the custom polygon controls wasn't working with this update but looks good now.

src/control.class.js
```diff
@@ -217,6 +249,55 @@
...
+          y: centerY - sinHalfOffset,
+        },
+        bl: {
+          x: centerX - cosHalfOffset,
+          y: centerY + sinHalfOffset,
+        },
+        br: {
+          x: centerX + cosHalfOffset,
+          y: centerY + sinHalfOffset,
+        },
```
asturur, line 297: this bunch of numbers here i remember they work with square and are a simplified version. Once the object is rotated i believe with a strongly rectangular controls, this does not work.
This same thing overly complicated the padding logic elsewhere.

the x and the y coordiante of each point when rotated are both affected by cos and sin of the 2 offsets.

asturur: There should be some debug code to uncomment to see those points on the canvas.

src/control.class.js
```diff
@@ -217,6 +249,59 @@
...
+        controlTriangleAngle = fabric.util.radiansToDegrees(Math.atan2(ySize, xSize));
+        cornerHypotenuse = Math.sqrt(xSize * xSize + ySize * ySize) / 2;
+      }
+      else {
+        // use default object corner sizes
+        controlTriangleAngle = 45;
+        /* 0.7071067812 stands for sqrt(2)/2 */
+        cornerHypotenuse = objectCornerSize * 0.7071067812;
+      }
+      var newTheta = controlTriangleAngle - fabric.util.degreesToRadians(objectAngle),
```
asturur, line 277: Here we are subtracting degrees with radians

gloriousjob: Thanks, I fixed this in my test code but didn't transfer it late last night when I committed. I need to not let lack of sleep cause blunders.

src/control.class.js
gloriousjob: Yes, I think I understand the original code now. It's using the complementary angle to triangulate the other corner (which both angles are 45 when in a square, which is why stuff works but it's almost hard-coding to run the original way). Things were working with right angles when I'd rotate so thanks for challenging me to check other rotations.

I'm not sure about debug code though. I did find things below a getImageLines call that I was able to uncomment and see some squares show up so that was hepful.

gloriousjob: @asturur Is there anything else you'd like to see here? I tried adding padding to a shape and the rectangular corner worked, even when rotating.

asturur:
Testing this now and trying to merge it.
I think that we can dramatically reduce the complexity somehow but will be a follow up.

asturur:
image

There is some issue with rendering changing only some controls.
Trying to verify what it is, seems a mutability problem somehow.

asturur:
Changes i applied:

- removed the name cornerSizeX/Y for sizeX/Y and similar for the touch version. Those are controls, corner is an old name and since those are positioned freely, it does not make sense anymore to reference to those as corners.
- fixed the mutability bug for styleOverride
- made the circle drawing control one level less nested

gloriousjob:
I like the rename... I was concerned about it but didn't want to change things without your approval so glad that's good!
I'm not sure I get the styleOverride thing but if things are working, that's what's important (probably because of the fabricObject addition when trying to get the value of the x and y sizes).

Issue #6555, opened 2020-08-29
Object.lockUniScaling alternative in V4


andrewzolotukhin:
I see that in V4 we have a lockUniScaling property removed for Object. Probably anybody knows what is propper alternative in V4 to get the same result?

asturur:
http://fabricjs.com/upgrade-guide

Look at the v4 breaking changes.

andrewzolotukhin:
Hi @asturur , I saw this link, there we have: " If uniformScaling is true, objects get scaled proportionally, if uniscaleKey is pressed, the behaviour get swapped." But what if I want only one object to get scaled uniformly but others no?

asturur:
It is a canvas only property right now.
It could have been done on the object, so that you could have act on the protype to get all objects.
But it is not.

You can still act on controls on this function here:

```js
  /**
   * Inspect event and fabricObject properties to understand if the scaling action
   * @param {Event} eventData from the user action
   * @param {fabric.Object} fabricObject the fabric object about to scale
   * @return {Boolean} true if scale is proportional
   */
  function scaleIsProportional(eventData, fabricObject) {
    var canvas = fabricObject.canvas, uniScaleKey = canvas.uniScaleKey,
        uniformIsToggled = eventData[uniScaleKey];
    return (canvas.uniformScaling && !uniformIsToggled) ||
    (!canvas.uniformScaling && uniformIsToggled);
  }
```
and let it consider an object property you like more.

I'm not opposed to reintroduce it for objects too, as it was before in 3.x was built broken.

andrewzolotukhin:
Great, thanks, I'll check it.

andrewzolotukhin:
As I understand now I should user controls API for that?

asturur:
You are right this has been a misjudgment when deciding to keep it on the canvas or on the objects.
It has to be on both.

I have to run on the v3 code to remember what exactly lockUniScaling was doing and how it cooperated with `uniScaleTransform` on the canvas.

andrewzolotukhin:
Would you like me to create a PR?

adamgamble:
PR here: https://github.com/fabricjs/fabric.js/pull/6575

asturur:
Added a message on the PR with an alternative, let me know guys what do you think.

andrewzolotukhin:
Thanks, sent you a response there.

SLKnutson:
Hi.
[...]

PR #6575, opened 2020-09-07
feature(controls): get back of lockUniscaling for objects

andrewzolotukhin:
Hi @asturur , I've ressurected lockUniScaling. Could you please review this draft?

### DIFF
--- a/src/control.class.js
+++ b/src/control.class.js
@@ -187,6 +187,16 @@
     },
 
     /**
+     *  
+     * @private
+     * @param {String} controlKey key where the control is memorized on the
+     * @return {Boolean}
+     */
+    _isHiddenWhileUniscalingLocked: function(controlKey) {
+      return controlKey === 'mr' || controlKey === 'mt' || controlKey === 'mb' || controlKey === 'ml';
+    },
+
+    /**
      * Returns controls visibility
      * @param {fabric.Object} object on which the control is displayed
      * @param {String} controlKey key where the control is memorized on the
@@ -194,8 +204,16 @@
      */
     getVisibility: function(fabricObject, controlKey) {
       var objectVisibility = fabricObject._controlsVisibility;
+      var hideWhileUniscalingLocked = fabricObject.lockUniScaling && this._isHiddenWhileUniscalingLocked(controlKey);
+
       if (objectVisibility && typeof objectVisibility[controlKey] !== 'undefined') {
+        if(hideWhileUniscalingLocked) {
+          return false;
+        }
         return objectVisibility[controlKey];
+      }
+      if(this.visible) {
+        return !hideWhileUniscalingLocked;
       }
       return this.visible;
     },
--- a/src/control.class.js
+++ b/src/control.class.js
@@ -207,12 +207,12 @@
       var hideWhileUniscalingLocked = fabricObject.lockUniScaling && this._isHiddenWhileUniscalingLocked(controlKey);
 
       if (objectVisibility && typeof objectVisibility[controlKey] !== 'undefined') {
-        if(hideWhileUniscalingLocked) {
+        if (hideWhileUniscalingLocked) {
           return false;
         }
         return objectVisibility[controlKey];
       }
-      if(this.visible) {
+      if (this.visible) {
         return !hideWhileUniscalingLocked;
       }
       return this.visible;
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1820,7 +1820,7 @@
    * @example <caption>JSON without additional properties</caption>
    * var json = canvas.toJSON();
    * @example <caption>JSON with additional properties included</caption>
-   * var json = canvas.toJSON(['lockMovementX', 'lockMovementY', 'lockRotation', 'lockScalingX', 'lockScalingY']);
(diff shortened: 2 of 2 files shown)


asturur:
i forgot what lockUniscaling was supposed to do, that is why i need time to understand what we want to do.

andrewzolotukhin:
Sure, when you remember just please take a look at this PR it looks like it's the single thing we should add

adamgamble:
This relates to https://github.com/fabricjs/fabric.js/issues/6555

asturur:
I'm sorry @andrewzolotukhin  took me a while to get back here to fabricjs.
I went to check up what uniscaling was doing in 3.x, and basically, it was hiding  the medium side controls.
Was also blocking scaling in other ways but the main point was disable medium side controls and the important function was saved in the design JSON.

I do not want to bring back this functionality as it was i want to search a way to enable it with the new control API.

One example is different control set per object class.
You can have images and text for example to have unilockScaling in this way

```js
const lockedControlSet = {
   ...fabric.Object.prototype.controls;
}
// remove the 4 medium side conrtols
delete lockedControlSet.mr;
delete lockedControlSet.ml;
delete lockedControlSet.mt;
delete lockedControlSet.mb;

// apply this control set to images and text

fabric.Image.prototype.controls = lockedControlSet;
fabric.Text.prototype.controls = lockedControlSet;
```

This is an example.
Are you looking for something completely different?

andrewzolotukhin:
Hi @asturur !

Thank you for your reply. Does that getting serialized to JSON? I understand that I could just path different controls set to some object types. But the problem is that user could disable uniscale locking for particular object from UI and he should still have it disabled once he saved/reloaded project from JSON. 

As I understand, controls state is not being saved to JSON.

asturur:
If the User has to disable per object of the same class, and we want to support it, your solution is the only one that works.

Give me some thinkering time, i really wanted to remove some extra feature that would interact with the others somehow.

In case we need to give it a different name, something like `hideSomethingControls` since it is what it is doing, and we need to add back the property in the object with the default false, where it was, with a description.

```js
    /**
     * When `true`, object non-uniform scaling is locked
     * @type Boolean
     * @default
     */
    lockUniScaling:           false,
```

Just we need better name and better description

asturur:
An alternative is extending the control api to custom getVisibility too.
The issue i have with this approach is that then we have a priority fight between 4 controls hide by a boolean and a specific control enabled manually. Who should win? those are the kind of questions i would like to not have to do.

andrewzolotukhin:
Hi @asturur , would you like me to update a PR with a better naming?

asturur:
Hey sorry for taking so much time here. I got your email, i do not want to bring back lockUniscaling in the official codebase. The controlApi is built to solve those kinds of problems here.

you can write a simple function in this way:

```js

var originalGetVisibility = fabric.Control.prototype.getVisibility;
var getVisibilityForLockUniScaling = function(fabricObject, controlKey) {
  if (this.lockUniScaling) {
    return false;
  }
  return originalGetVisibility(fabricObject, controlKey);
}

// add this function on your controls of choice:
fabric.Object.prototype.controls.tl.getVisibility = getVisibilityForLockUniScaling;
fabric.Object.prototype.controls.tr.getVisibility = getVisibilityForLockUniScaling;
fabric.Object.prototype.controls.bl.getVisibility = getVisibilityForLockUniScaling;
fabric.Object.prototype.controls.br.getVisibility = getVisibilityForLockUniScaling;
```

This is completely in the api idea and you are literally just adding the IF condition you need in the controls you need.

andrewzolotukhin:
Hi @asturur , thanks, your suggestion is working for me. I've just added this field to JSON and used the code you provided. Closing this PR.

r043v:
a bit of trouble as i would easily assume than i could filter object type using this method, ie _fabric.IText.prototype.controls.ml.getVisibility = returnFalseFn;_ but yet do it will in fact set it globally like in your exemple with fabric.Object

so for now i use a system like this :

```
const visibilityRules = {
	'i-text':{
		ml:false, mr:false, mt:false, mb:true, mtr:false
	},
	circle:{
		mtr:false
	}
};

fabric.Control.prototype.originalGetVisibility = fabric.Control.prototype.getVisibility;
fabric.Control.prototype.getVisibility = function( obj, border ){
	if( ( v = visibilityRules[ obj.type ] ) !== undefined )
		if( ( v = v[ border ] ) !== undefined )
			return v ;
	return fabric.Control.prototype.originalGetVisibility.call( this, obj, border );
}
```

r043v:
right, it work fine !

now i could have custom function or enable my own rebind only where it need :)

thanks a lot!

```
const visibilityRules = {
	'i-text':{
		ml:false, mr:false, mt:false, mb:true, mtr:false
	},
	circle:{
		mtr:false
	}
};

const customControl = Object.assign( {}, fabric.Control );
Object.setPrototypeOf( customControl, fabric.Control );

customControl.prototype.originalGetVisibility = customControl.prototype.getVisibility ;
customControl.prototype.getVisibility = function( obj, border ){
	if( ( v = visibilityRules[ obj.type ] ) !== undefined )
		if( ( v = v[ border ] ) !== undefined )
			return v ;

	return customControl.prototype.originalGetVisibility.call( this, obj, border );
}

fabric.IText.prototype.Control = customControl ;
fabric.Circle.prototype.Control = customControl ;

```

Issue #6117, opened 2020-01-24
Mouse up returning target of mouse down after performing drag operation


kartikanand:
## Version
3.6.1

## Test Case
https://jsfiddle.net/eupbgtmd/24/

## Information about environment
Browser, tested in Chrome

## Steps to reproduce
1. Create a line from the lower right anchor (pink circle) to top left anchor (pink circle)

## Expected Behavior
A line should get drawn between the two anchors and the new anchor should be returned on mouse up event.

## Actual Behavior
The line doesn't get drawn because the old anchor is returned instead of the new anchor on mouse up event

The issue seems related to #2508 but in this case in the context of groups. The issue is reproducible only when `selectable: true` on the group. Making it `false` returns the correct target on mouse up event.
 This can be seen by changing `selectable` to `false` on line no. 41 in the fiddle.

Also this seems to be a regression from `1.7.22` since the correct target is returned on using the old version.

asturur:
I'm sorry for me the fiddle is too much to parse.
Can you focus on where do you think the bug is, without using your particular application use case?

asturur:
Modify the fiddle and i will reopen the issue for investigation

kartikanand:
Hello,

I've modified the fiddle, reduced it as much as I can while retaining the code for connecting lines.

In the fiddle, if `selectable` is turned off on lines 19 and 27 then the mouse:up event returns the correct target, otherwise, the old target of mouse:down is returned.

Fiddle:
https://jsfiddle.net/b1hm2avd/41/

asturur:
if the problem is the target on mouseup, eventually having 2 objects only on canvas, both with lockmovementX and Y should do the trick without the need to write drawing line code?

i imagine if there is a bug can be triggered clicking down on an object and then dragging the mouse over another?

kartikanand:
Yep, my bad!

Created a small test case:
https://jsfiddle.net/2bnkygjv/4/

The issue can be seen by clicking down on an object and dragging the mouse over another as you just described. Also the issue doesn't depend on groups.

The problem is solved when `selectable: false`

kartikanand:
Hi @asturur, the issue is still there in the latest version of Fabric. Can we reopen it ?

asturur:
yes we should.

asturur:
Labeled as a bug won't be closed again by the stale bot

asturur:
I added the label, but i did not kept the issue open.

PR #6591, opened 2020-09-13
Fix mouse up target while performing drag operation

kartikanand:
Hi @asturur 

I'm trying to fix the bug describer here in #6117. Based on my limited understanding of the code, I feel the bug is triggered because a transform event of "Drag" is fired. The problem here is that the node's x and y positions are locked. If they were unlocked then surely the mouse up target will be the same as mouse down. But here since the positions are locked, the mouse up target should be wherever the mouse currently is. That I found was coming out as incorrect because of the drag transform being fired.

Can you review this draft and provide your suggestions ?

Thanks,
Kartik

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -524,6 +524,10 @@
             absolutePointer: this._absolutePointer,
             transform: this._currentTransform
           };
+      if (eventType === 'up') {
+        options.currentTarget = this.findTarget(e);
+        options.currentSubTargets = this.targets;
+      }
       this.fire('mouse:' + eventType, options);
       target && target.fire('mouse' + eventType, options);
       for (var i = 0; i < targets.length; i++) {
--- a/test/unit/canvas_events.js
+++ b/test/unit/canvas_events.js
@@ -301,6 +301,21 @@
     assert.equal(isClick, false, 'moving the pointer, the click is false');
   });
 
+  QUnit.test('mouse:up should return currentTarget', function(assert) {
+    var e = { clientX: 30, clientY: 30, which: 1 };
+    var e2 = { clientX: 31, clientY: 31, which: 1 };
+    var rect = new fabric.Rect({ left: 0, top: 0, width: 50, height: 50 });
+    canvas.add(rect);
+    var opt;
+    canvas.on('mouse:up', function(_opt) {
+      opt = _opt;
+    });
+    canvas.__onMouseDown(e);
+    canvas.__onMouseMove(e2);
+    canvas.__onMouseUp(e2);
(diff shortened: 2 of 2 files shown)


asturur:
My suggestion is that:
- i think the mouseup returning the same target on drag was an optimization to not research the target.
- we should remove the optiomization

Let me find it.

asturur:
```
    /**
     * Cache common information needed during event processing
     * @private
     * @param {Event} e Event object fired on event
     */
    _cacheTransformEventData: function(e) {
      // reset in order to avoid stale caching
      this._pointer = this.getPointer(e, true);
      this._absolutePointer = this.restorePointerVpt(this._pointer);
      this._target = this._currentTransform ? this._currentTransform.target : this.findTarget(e) || null;
    },
```

If there is a currentTransform, always return the target of the current transform.

asturur:
We cannot actually skip executing _setupCurrentTransform or the event system could break

asturur (changes requested):
PLease try to apply the fix in the comment

kartikanand:
Hi @asturur, Yes you're correct we cannot simply remove the transform. I did some tests with my current fix and it does'nt work for all cases. Let me try the one you suggested and I'll update this pull request with both the change and the test.

Thanks,
Kartik

src/mixins/canvas_events.mixin.js
```diff
@@ -524,6 +524,10 @@
             absolutePointer: this._absolutePointer,
             transform: this._currentTransform
           };
+      if (eventType === 'up') {
+        options.currentTarget = this.findTarget(e);
+        options.currentSubTargets = this.targets;
```
asturur, line 529: oh good point.

test/unit/canvas_events.js
```diff
@@ -301,6 +301,21 @@
...
+    var rect = new fabric.Rect({ left: 0, top: 0, width: 50, height: 50 });
+    canvas.add(rect);
+    var opt;
+    canvas.on('mouse:up', function(_opt) {
+      opt = _opt;
+    });
+    canvas.__onMouseDown(e);
+    canvas.__onMouseMove(e2);
+    canvas.__onMouseUp(e2);
+    assert.equal(opt.currentTarget, rect, 'options match model - currentTarget');
```
asturur, line 316: we would need to mimick your use case just to make a test that make sense.

Add 2 rects, one near the other, the first locked, mouse down on the first, do a mouse move on the second and mouseup on the second. verify that target and currentTarget match rect1 and rect2.

asturur (approved):

asturur:
Seems good to me, as soon as is green we can merge. Thanks!

kartikanand:
Thank you @asturur for your guidance!

PR #6617, opened 2020-09-26
feat(fabric.controlsUtils) Move drag to actions to control handlers

### LINKED ISSUE #6565: IE11 - object doesn't support property or method 'assign'
## Version
   4.1.0 

browser?
Which browsers?
IE11

## Steps to reproduce

- add an image to Canvas.
- try to resize or drag (move) that image on canvas using mouse pointer.
- fbaric js 4.1.0 crashes in IE11 (works fine in Chrome)

## Expected Behavior

fabric JS should not crash in IE11.

## Actual Behavior

fabric js 4.1.0 crashes with error " object doesn't support property or method 'assign'". it works fine in Chrome.

i just shifted from fabric js 3.6.3 to 4.1.0. i simply updated the fabric js file and noticed this issue in IE11.

is there any way to fix it ?

Thanks

### LINKED ISSUE #6607: Move action in custom controls api (4.1.0)
I just updated my project to 4.1.0 and recreated my custom controls using the built in custom controls api (which is great).

There is only one last thing, that is missing: A simple move action for my move icon.
Here you can see my move icon (old working version): https://lettering.org/lettering-generator/

I can't finde a move action here: https://github.com/fabricjs/fabric.js/blob/master/src/mixins/default_controls.js

The old custom controls plugin just needed the "action: move". Is there something similar in fabric.js?
I searched for move actions everywhere, but I did not find anything.

`	fabric.Object.prototype.controls.customMove = new fabric.Control({
		x: -0.5,
		y: -0.5,
		//actionHandler: move?,
		cursorStyle: 'pointer',
		render: renderIcon(icon_move),
		cornerSize: icon_size
	});`

Thanks for your help!

asturur:
No practical changes.
Move the ability to move an object outside the Canvas class, into controls utils.
Also fix a ie11 bug

close #6565
close #6607

### DIFF
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -31,8 +31,9 @@
   function fireEvent(eventName, options) {
     var target = options.transform.target,
         canvas = target.canvas,
-        canasOptions = Object.assign({}, options, { target: target });
-    canvas && canvas.fire('object:' + eventName, canasOptions);
+        canvasOptions = fabric.util.clone(options);
+    canvasOptions.target = target;
+    canvas && canvas.fire('object:' + eventName, canvasOptions);
     target.fire(eventName, options);
   }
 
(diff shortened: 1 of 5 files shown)


src/controls.actions.js
```diff
@@ -31,7 +31,8 @@
   function fireEvent(eventName, options) {
     var target = options.transform.target,
         canvas = target.canvas,
-        canvasOptions = Object.assign({}, options, { target: target });
+        canvasOptions = fabric.util.clone(options);
```
asturur, line 34: fix #6565

PR #6640, opened 2020-10-09
Check if style exists and is not null

yassipad:
Style at lineIndex and charIndex can be null, which throws an error.

### DIFF
--- a/src/mixins/text_style.mixin.js
+++ b/src/mixins/text_style.mixin.js
@@ -43,7 +43,7 @@
       for (var p1 in obj) {
         // eslint-disable-next-line
         for (var p2 in obj[p1]) {
-          if (typeof obj[p1][p2][property] !== 'undefined') {
+          if (obj[p1][p2] && typeof obj[p1][p2][property] !== 'undefined') {
             return true;
           }
         }


asturur:
How do you get a null there?

yassipad:
I believe it happens when i copy/paste an object and then modifies the text but i'm not entirely sure...

asturur:
Can you try to replicate it?

yassipad:
Ok actually, I know why: the style remains correct in the JSON but if one tries to store that value on a (PHP) server, the styles object becomes a numeric array (since the keys are numeric) with null values wherever the key was not set.

asturur:
The style object shouldn't become an array, is still
wrapped in a couple of curly braces. How is that possible?

yassipad:
Ok nevermind, it was my mistake, here's the explanation in case you want to know what happened:

I'm using Laravel as a backend PHP server application and the request data input is turned into an associative array automatically using `json_decode($data, true)`, which has the drawback of turning any empty object `{}` into an array `[]`. You can checkout the behaviour on this [link](http://phpfiddle.org/lite?code=%3C?php\necho%20%22assoc_array%20=%20false%20=%3E%20%22%20.%20json_encode(json_decode(%27{%20%22styles%22:%20{}%20}%27))%20.%20%22%20//%20%22%20.%20%22assoc_array%20=%20true%20=%3E%20%22%20.%20json_encode(json_decode(%27{%20%22styles%22:%20{}%20}%27,%20true));\n?%3E\n). I solved my issue by setting a custom casting that will force the "styles" property to be considered an object whether or not it's empty.

```php
json_encode(json_decode('{ "styles": {} }')) => { "styles": {} }

json_encode(json_decode('{ "styles": {} }', true)) =>  { "styles": [] }
```

Issue #6649, opened 2020-10-14
v4: 'scaling' event triggered before object position is adjusted


SLKnutson:
## Version
4.2.0

## Expected Behavior
The 'scaling' event should be fired after the object position is adjusted to pin to a specific side/corner. That is how fabric behaved pre-v4, and in my opinion, what most people would expect.

## Actual Behavior
The 'scaling' event is triggered after the scale is changed but before the position is adjusted to pin to a specific side/corner.

Thanks!

asturur:
i think you are rigth. this is changed and messy right now.

I m not sure how to fix it at this point.

Probably also half people prefer the old position, half the new. But i agree should be fixed.

SLKnutson:
I'm also running into this issue with rotation. I think skewing also has this issue. I'll take a stab at fixing it.

Eltonfjunior:
Good afternoon.
I am also having the same problem, do you already have any return on this demand?

asturur:
We have a PR, that is fine, but we need to avoid breaking the interface and i did not have a time to sit down and see what is the solution.

https://github.com/fabricjs/fabric.js/pull/6650

ShaMan123:
closed by #6650

PR #6650, opened 2020-10-15
fix v4: 'scaling' event triggered before object position is adjusted

SLKnutson:
Hopefully this is satisfactory. It worked in my project.
https://github.com/fabricjs/fabric.js/issues/6649
Thanks!


src/controls.actions.js
```diff
@@ -188,15 +188,19 @@
   /**
    * Wrap an action handler with saving/restoring object position on the transform.
    * this is the code that permits to objects to keep their position while transforming.
+   * @param {string} eventName
    * @param {Function} actionHandler the function to wrap
    * @return {Function} a function with an action handler signature
    */
-  function wrapWithFixedAnchor(actionHandler) {
+  function wrapWithFixedAnchor(eventName, actionHandler) {
```
asturur, line 195: we can't change wrapWithFixedAnchor argument signature, or we would break the compatibility of fabric v4.

asturur: we need an event wrapper, or something like that. I ll get some fixes done and then i ll try to help.

asturur:
This still a valid issue

asturur:
I do not have the pipeline to check it, or the time to fix the test errors, but we definitely need this as a reminder to fix it.

SLKnutson:
@asturur I updated the PR to not change the argument signature

asturur:
hi @SLKnutson i know a lot of time passed, but now the pipeline is working again.
If you can update your PR with a fresh master branch, it would run again and we can continue to fix it.

SLKnutson:
@asturur 
It seems like the failure isn't related to the code?

asturur:
hi @SLKnutson i ll review asap

asturur:
seems great tho

ShaMan123:
+1 exactly what I need. Bleeded a day on this.

melchiar:
Thanks for this! Turns out this was the cause of a bug in my application that started a few versions ago.

Issue #6651, opened 2020-10-16
SVG ouput for element with clipPath and shadow is incorrect


phil-r:
## Version
4.2.0

## Test Case
https://codesandbox.io/s/hopeful-lake-vvm9n?file=/src/index.js

<img width="409" alt="image" src="https://user-images.githubusercontent.com/577316/96287641-87c82b00-0fe2-11eb-90c1-34eff05febf7.png">

## Information about environment
Browsers, latest chrome and firefox

## Steps to reproduce

Create an element that has both shadow and clipPath

## Expected Behavior

SVG output should look the same as canvas

## Actual Behavior

SVG output looks different from the canvas

phil-r:
issue is still there

asturur:
labeled as a bug won't be labeled as stale anymore.

PR #6653, opened 2020-10-16
Fix element with clipPath and shadow svg serialization

phil-r:
This fixes the issue for the example provided in #6651, but sadly it breaks other tests and I'm not sure how I should approach this.

Before: 

https://codesandbox.io/s/hopeful-lake-vvm9n?file=/src/index.js

image

After: 

https://codesandbox.io/s/pensive-visvesvaraya-bs738?file=/index.html

image

I would be happy if you could provide some guidance into how broken tests can be fixed, or if my approach is completely wrong, thanks.


asturur:
i understand the issue, the shadow-filter should be applied on the clipped object an not vice versa.
I m just unsure how to deal with it, it was a bit complicated to have clipPath working correctly in svg export

PR #6654, opened 2020-10-16
Fixed SvgExport in path with gradient and added tests

claas-c:
There seems to be a problem in fabric with the svg export for a percentage based gradient on a path element. Here is a comparison between the rendering on canvas and in the exported svg:
https://codesandbox.io/s/clever-mountain-u0fxv?file=/src/index.js

When removing the subtraction of pathOffset from offset in the ```toSvg``` function of the ```gradient``` with the unit ```percentage```, the exported svg looks like the rendering on canvas, as can be seen in this codesandbox:
https://codesandbox.io/s/nice-vaughan-l3sj4?file=/src/index.js

I added two visual tests to this PR, the first one tests if the gradient is correctly rendered on canvas and succeeds also for the current version. The second one tests the exported svg image and only succeeds with my changes. I am not sure if these tests are in the correct files or are performed at the right position. I could not find any guidelines for creating visual tests.


asturur:
Pr is good! thanks.

asturur (approved):

PR #6655, opened 2020-10-17
[feature] RadioInput class

CharlesRA:
This PR contains the progress of the RadioInput element. This component contains 3 callbacks, which can be called when the input is checked or not
Can this be useful to the project?


CharlesRA:
[Example](https://www.awesomescreenshot.com/video/1388310?key=645b257e4d5261b4cb195c3434e5a7fe)

Maybe the fill style needs to be improved.

Btw the default check status or the checkFillColor can be customize during the object initialization

asturur:
I wish we do not introduce UI elements in fabricJS, seems wrong.
But if the class is good, it can be totally in a repo in this organization if you are willing to maintain the class or answer eventual questions from people.

src/shapes/radioInput.class.js
```diff
@@ -0,0 +1,162 @@
...
+        );
+      },
+
+      /**
+       * @private
+       * @param {CanvasRenderingContext2D} ctx Context to render on
+       */
+      _render: function (ctx) {
+        this.callSuper("_render", ctx);
+      },
```
asturur, line 148: you do not need to call this. It will automatically call the circle one.

src/shapes/radioInput.class.js
```diff
@@ -0,0 +1,162 @@
...
+   * Returns fabric.RadioInput instance from an object representation
+   * @static
+   * @memberOf fabric.RadioInput
+   * @param {Object} object Object to create an instance from
+   * @param {Function} [callback] Callback to invoke when an fabric.Path instance is created
+   */
+  fabric.RadioInput.fromObject = function (object, callback) {
+    callback && callback(new fabric.RadioInput(object));
+  };
+  fabric.RadioInput.async = true;
```
asturur, line 161: async is deprecated, all classes are async.

src/shapes/radioInput.class.js
```diff
@@ -0,0 +1,162 @@
...
+      toObject: function (propertiesToInclude) {
+        return fabric.util.object.extend(
+          this.callSuper(
+            "toObject",
+            [
+              "checked",
+              "onCheckFillColor",
+              "onChange",
+              "onCheckCallback",
+              "onUnCheckCallback",
```
asturur, line 136: we do not support anymore serialization of functions. Is not secure and hard to grasp for developers.

PR #6686, opened 2020-11-11
fix(fabric.Canvas): rotation handle should take origin into account

### LINKED ISSUE #6679: centeredRotation is not taken into account for rotation handle
## Version
4.2.0

## Test Case
http://jsfiddle.net/qkpf3abu/1/

## Steps to reproduce
Run the fiddle
Select the Rect and rotate it

## Expected Behavior
Since centeredRotation is set to false, i expect it to rotate around the top/left corner.

## Actual Behavior
It always rotates around the center

virror:
Remove the fixed center origin from the code.
Default origin should depend from `_shouldCenterTransform` function in the `_setupCurrentTransform`;

close #6679

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -608,10 +608,6 @@
       else if (corner === 'bl' || corner === 'mb' || corner === 'br') {
         origin.y = 'top';
       }
-      else if (corner === 'mtr') {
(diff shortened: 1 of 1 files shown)


asturur:
@virror a small description would have been on spot.

PR #6719, opened 2020-12-07
limitedToCanvasSize property to brush

CharlesRA:
In fabricjs it's quite difficult to control the different tools and manage their limits. Moreover, the [stackoverflow question](https://stackoverflow.com/questions/44046162/fabricjs-prevent-freedrawing-outside-of-canvas-boundaries) remained unanswered.

So I added the limitedToCanvasSize property to prevent free drawing outside canvas boudaries.


asturur:
I'm not opposed to this, but i would like it to a method of base brush class, and used in all the brushes. It can't be a one thing for the pencil brush.

src/brushes/base_brush.class.js
```diff
@@ -3,122 +3,146 @@
...
+    },
+
+    /**
+     * Removes brush shadow styles
+     * @param {Object} pointer
+     * @private
+     */
+
+    _isOutSideCanvas: function(pointer) {
+      if (pointer.x < 0 || pointer.x > this.canvas.getWidth() || pointer.y < 0 || pointer.y > this.canvas.getHeight()) {
```
asturur, line 140: I would like to return this expression rather than having an if that return true or false.

src/brushes/circle_brush.class.js
```diff
@@ -76,6 +76,9 @@ fabric.CircleBrush = fabric.util.createClass(fabric.BaseBrush, /** @lends fabric
       this._render();
     }
     else {
+      if (this.limitedToCanvasSize === true && this._isOutSideCanvas(pointer)) {
+        return;
+      }
```
asturur, line 81: in the pencil brush we are returning immediately if the condition match.
Why are we allowing the full render here?

src/brushes/base_brush.class.js
```diff
@@ -3,122 +3,146 @@
...
-  /**
-   * Sets brush shadow styles
-   * @private
-   */
-  _setShadow: function() {
-    if (!this.shadow) {
-      return;
-    }
+fabric.BaseBrush = fabric.util.createClass(
+  /** @lends fabric.BaseBrush.prototype */ {
```
asturur, line 7: sending this comment one line down makes all the diff bigger than necessary.
This is kind of a problem for an open project since is less clear what has changed and makes the PR harder to review.

asturur:
Hi @CharlesRA i left a couple of minor comments and questions.
When you make the PR, please do not commit the full build library.
To restore it back you can user `git checkout master dist` will bring back the dist folder to the master version.
We build it each release.

CharlesRA:
Hi !
Sorry for the mistakes, I'm new to open source contribution and programming 😊.
I made the changes you asked me to do, are there other things that are wrong ?

asturur:
Testing this today and merging

CharlesRA:
Hello, do I need to do something else for this PR ?

asturur:
I do not remember outstanding comments, i have just been busy.

asturur:
Thank you for the PR, and sorry for taking so much time to handle it.

asturur:
I'll add a test for it, please have a look at how is done, so next time you can write one too

SatwinderKaur:
@asturur I hwas having issues with LimitedToCanvasSize when we have zoom on Canvas. I could not draw on the whole canvas, only on a part of it. I think because we ignore zoom , when getting pointer. :
`_onMouseMoveInDrawingMode(e: any) {
      if (this._isCurrentlyDrawing) {
        const pointer = this.getPointer(e); //  this.getPointer(e,true) seemed to work in my case`

Can we please include zoomed canvas support in the code and pass correct pointer in ` _isOutSideCanvas: function(pointer) `

KonMost:
Hi all,

can someone help me on how to modify the circle_brush.class in order to get an ellipse/slanted line instead of a circle as brush tip to draw the freedrawing lines? i think this would be the place to add a calligraphy kind of brush. i tried changing my min.js but it did not take the slanted line or oval as brush tip. please advise

PR #6730, opened 2020-12-15
Redundant conditional expression

jasonsturges:
This appears to be a redundant conditional expression, but maybe explicitly stating `true` / `false` helps to provide clarity.


asturur:
yes definitely redundant.

asturur (approved):

PR #6732, opened 2020-12-16
Documentation errata

jasonsturges:
Some documentation updates, if interested - minor errata in documentation and tests.


asturur (approved):
Thank you very much, i wonder why i type
so bad.

PR #6733, opened 2020-12-16
Util `fromArcToBeziers()` API update

jasonsturges:
Proposing API change:
- Currently: `fromArcToBeizers`
- Proposing: `fromArcToBeziers`

### DIFF
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -702,6 +702,12 @@
   fabric.util.makePathSimpler = makePathSimpler;
   fabric.util.getPathSegmentsInfo = getPathSegmentsInfo;
   fabric.util.fromArcToBeziers = fromArcToBeziers;
+    /**
+   * Typo of `fromArcToBeziers` kept for not breaking the api once corrected.
+   * Will be removed in fabric 5.0
+   * @deprecated
+   */
+  fabric.util. fromArcToBeizers = fromArcToBeziers;
   fabric.util.getBoundsOfCurve = getBoundsOfCurve;
   fabric.util.getPointOnPath = getPointOnPath;
   // kept because we do not want to make breaking changes.
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -115,7 +115,7 @@
    * @param {Number} y1
    * @param {Number} x2 secondo control point
    * @param {Number} y2
-   * @param {Number} x3 end of beizer
+   * @param {Number} x3 end of bezier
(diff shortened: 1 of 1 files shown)


asturur:
i have to check how old is that util, in case we can make a correct alias, and deprecate the old one, but not remove it right away.

jasonsturges:
@asturur Good idea, sorry - hadn't thought of that.

Looks like it was added August 20, 2020
SHA: 1c2164bab523da2405bbafb395d31c2b0584fc33
Pull Request: https://github.com/fabricjs/fabric.js/pull/6504

Would you like me to add a deprecation?  

Sorry for the hassle - certainly no major issue here, if you'd like to close and consider this some other time.

asturur:
you can correct the function name and export, then add an alias with the wrong old name and a @deprecated notic in the jsdoc, with a line explaining that was a typo

src/util/path.js
```diff
@@ -692,16 +692,16 @@
...
+    beziers.forEach(function(bezier) {
+      ctx.bezierCurveTo.apply(ctx, bezier.slice(1));
     });
   };
 
   fabric.util.parsePath = parsePath;
   fabric.util.makePathSimpler = makePathSimpler;
   fabric.util.getPathSegmentsInfo = getPathSegmentsInfo;
-  fabric.util.fromArcToBeizers = fromArcToBeizers;
+  fabric.util.fromArcToBeziers = fromArcToBeziers;
```
asturur, line 704:
```suggestion
  fabric.util.fromArcToBeziers = fromArcToBeziers;
    /**
   * Typo of `fromArcToBeziers` kept for not breaking the api once corrected.
   * Will be removed in fabric 5.0
   * @deprecated
   */
  fabric.util. fromArcToBeizers = fromArcToBeziers;
```

asturur: I wonder if this suggestion from one line to many can just work.

jasonsturges:
@asturur Faster than me - just beat me to it.

I had added:

image

Which I think is working as intended:

image

But I won't push, as it looks like we did the same thing.

jasonsturges:
Removed some whitespace and had to revert the change I made (sorry, I had pushed at the same time).

I think this is good now if you want to review the final pull request.

asturur:
I added the suggestion to save you typing!
Thanks for the PR merging now

asturur (approved):

PR #6747, opened 2020-12-22
Methodos for backgorundImage controls

matheuscamarques:
backgroundImage

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -407,7 +407,14 @@
       }
       return objsToRender;
     },
-
+    getBackgroundImage(){
+      return this..backgroundImage.src;
+    },
+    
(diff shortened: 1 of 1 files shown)


asturur:
Hi @matheuscamarques
Could you please try to explain what is the added benefits of a method like this?

Also as an added note, backgroundImage property supports any kind of object, so getting the src property from it is not useful.

matheuscamarques:
I am making a photoshop style editor and with setters and modifiers in that part that I miss. I had to do a local implementation. Note: I missed an error in the commit '..'.

matheuscamarques:
https://stackoverflow.com/questions/14171378/fabricjs-delete-backgroundimage

PR #6756, opened 2020-12-25
Feature/promises

zhangsai521314:


asturur:
Let's have a discussion in the proper area if you want to work on this, so we agree before on method signatures and other small details

PR #6768, opened 2020-12-30
Added 'path' to Text toObject properties

### LINKED ISSUE #6762: canvas.loadFromJSON doesn't work when there is text on path on canvas
Hi, I am implementing a redo/undo feature as well as saving of json data, the text on path json data seems to be unusable when exported

## Version
4.3.0

## Test Case
https://jsfiddle.net/deceive9/zef528vq/13/

Google Chrome

## Steps to reproduce
Use "Add text on paths" demo on fabricjs site, draw a line
Use canvas.toJSON(['path']) function to export JSON data
Load that same json on the canvas

## Expected Behavior
canvas will be loaded

## Actual Behavior
Error:  Cannot read property 'x' of undefined

melchiar:
Added 'path' to the array of properties serialized with text objects.

close #6762

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1576,7 +1576,7 @@
    * @param {Function} [callback] Callback to invoke when an fabric.Text instance is created
    */
   fabric.Text.fromObject = function(object, callback) {
-	  var path = object.path;
+    var path = object.path;
     delete object.path;
     return fabric.Object._fromObject('Text', object, function(textInstance) {
       if (path) {
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1576,7 +1576,7 @@
    * @param {Function} [callback] Callback to invoke when an fabric.Text instance is created
    */
   fabric.Text.fromObject = function(object, callback) {
-    var path = object.path;
+    var path = clone(object.path);
     delete object.path;
     return fabric.Object._fromObject('Text', object, function(textInstance) {
       if (path) {
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1416,6 +1416,7 @@
       ].concat(propertiesToInclude);
       var obj = this.callSuper('toObject', additionalProperties);
       obj.styles = clone(this.styles, true);
+      obj.path = this.path && this.path.toObject();
       return obj;
     },
 
@@ -1575,7 +1576,19 @@
    * @param {Function} [callback] Callback to invoke when an fabric.Text instance is created
    */
   fabric.Text.fromObject = function(object, callback) {
-    return fabric.Object._fromObject('Text', object, callback, 'text');
+	  var path = object.path;
+    delete object.path;
+    return fabric.Object._fromObject('Text', object, function(textInstance) {
+      if (path) {
+        fabric.Object._fromObject('Path', path, function(pathInstance) {
+          textInstance.set('path', pathInstance);
+          callback(textInstance);
+        }, 'path');
+      }
+      else {
+        callback(textInstance);
+      }
+    }, 'text');
   };
 
   fabric.Text.genericFonts = ['sans-serif', 'serif', 'cursive', 'fantasy', 'monospace'];
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1412,6 +1412,7 @@
         'textAlign',
         'textBackgroundColor',
         'charSpacing',
+        'path'
       ].concat(propertiesToInclude);
       var obj = this.callSuper('toObject', additionalProperties);
       obj.styles = clone(this.styles, true);


asturur:
Ok this is not as simple :)
But thanks for starting.

Let me drop a couple of hints

asturur:
After this line we
https://github.com/fabricjs/fabric.js/blob/20bb6ca93f9c20cb8b9fa346a5ebc769d946ef09/src/shapes/text.class.js#L1418
We need something like

```obj.path = this.path && this.path.toObject()```

Since we need to serialize the path too, we can't just reference the value.

**Comment by asturur** [comment]
In restoring the object we need some extra care:
https://github.com/fabricjs/fabric.js/blob/20bb6ca93f9c20cb8b9fa346a5ebc769d946ef09/src/shapes/text.class.js#L1577
Now i ll mock up the code, but i think you can have a look at what we do with images where we restore filters before restoring the image itself.

i **THINK** ( not sure ) we can restore the path and then text or viceversa, depending what make us easier to read/write the code.

```
var path = object.path;
delete object.path;
fabric.Object._fromObject('Text', object, function(textInstance) {
  if (path) {
    fabric.Object._fromObject('Path', path, function(pathInstance) {
      textInstance.set('path', pathInstance);
      callback(textInstance);
    }, 'path');
  }
  else {
    callback(textInstance);
  }
}, 'text');
```

**Comment by melchiar** [comment]
ahhh, I see the problem. I had done a quick test of serializing and then loading the canvas which seemed to load the path data just fine (https://jsfiddle.net/melchiar/rb85gsn9/). I now see that since my serialized canvas wasn't being stringified it was just referencing the old path object when loading rather than reinitializing it.

**Comment by melchiar** [comment]
Thanks for the hints. Just thinking this through a bit more and have a couple of thoughts. If path properties like pathOffset are going to recalculate at deserialization anyways, should those props be excluded from serialization? Also, I believe we're going to run into a problem with the object's cache canvas dimensions being too small in some situations causing clipping. 

image

Should we perhaps be adding the text height to the path's dimensions similar to how stroke and padding are added? Or maybe just to the cache canvas dimensions?

**Comment by asturur** [comment]
I do not think pathOffset is serialized, it shouldn't be.

The bounding box is a separate issue as others we are going to find for sure, text on path is in beta and i m sure isn't done yet.

Let's address one at time, and keep this PR on serialization.

We can open an issue on the bounding box already.

**Comment by melchiar** [comment]
K, I've basically used your exact approach and it seems to work well. Unless you think we should change the order of how the text and path are loaded?

File: src/shapes/text.class.js [hunk 1]
```diff
@@ -1574,7 +1576,19 @@
    * @param {Function} [callback] Callback to invoke when an fabric.Text instance is created
    */
   fabric.Text.fromObject = function(object, callback) {
-    return fabric.Object._fromObject('Text', object, callback, 'text');
+    var path = object.path;
+    delete object.path;
```
asturur, line 1580: i made a wrong suggestion here.
In this case we are mutating the input of the function, meaning that the second time you restore the same object, we deleted the path property from it.
We need to clone this object before deleting the path. ( fabric.util.object.clone is what we use generally )

deocampoej:
Thank you so much, i applied these changes on my personal copy of fabricjs

asturur:
I wish we could get Travis going again

melchiar:
probably something wrong with the flux capacitor 😉

asturur:
i have to switch to github actions, i do not feel like bothering travis for free usage, i asked twice, if they do not have time to answer or evaluate i cannot blame them.  the issue is that the system has been abused by miners and now is pay to use. It seems to me  that the minimum is 69$ per month and i do not want to spend that even if the banner on the website could theoretically cover it for the most part.

asturur:
@melchiar if you could add any commit to this PR here, it would get the new pipeline running

melchiar:
k, let me know if any changes are needed.

melchiar:
@asturur any idea what's going on with the unit tests?

asturur:
you broke them, for sure.
Adding a property to export, breaks a bunch of tests.
And without having jest snapshot testing, this is a pain to fix.

I think i should start of letting go browser based unit tests and move forward to pure functional unit tests in node + jest and leaving the browser for the visual tests

But that is a huge refactor.

melchiar:
just looking at this test: https://github.com/fabricjs/fabric.js/pull/6768/checks?check_run_id=1763906236

It seems that the outputted string's properties are just in a different order to the expected output, so not actually related to this PR?

asturur:
@melchiar you have push permission on the repo i belive.
You should be able to open branches directly on fabricjs repo rather than on your own fork.
Doing so i can finish things or you can do as well, in this case i opened a new branch to replace this PR to which i cannot commit. Your commits are still part of the job no credit is lost.

PR #6773, opened 2021-01-02
Update static_canvas.class.js

sponno:
Added one line to the allow smooth scrolling help docs.  This property can only be set when the canvas is instiantied, there is no effect you try and change this property after the canvas has been created.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -119,7 +119,8 @@
     controlsAboveOverlay: false,
 
     /**
-     * Indicates whether the browser can be scrolled when using a touchscreen and dragging on the canvas
+     * Indicates whether the browser can be scrolled when using a touchscreen and dragging on the canvas.
+     * This must a property passed to the canvas when instantiated. Setting this property after, has no effect.
      * @type Boolean
      * @default
      */


src/static_canvas.class.js
```diff
@@ -119,7 +119,8 @@
     controlsAboveOverlay: false,
 
     /**
-     * Indicates whether the browser can be scrolled when using a touchscreen and dragging on the canvas
+     * Indicates whether the browser can be scrolled when using a touchscreen and dragging on the canvas.
+     * This must a property passed to the canvas when instantiated. Setting this property after, has no effect.
```
asturur, line 123: Can we change this in:
`In order to work, this property has to be passed in the canvas options at creation time. Changing it later won't have any effect`

asturur:
@sponno thank you for the doc fix

src/static_canvas.class.js
sponno: sounds great.

sponno:
@asturur Love your work, I've been working with FabricJS for about a year now in a project that's about to launch.
Just sponsored this repo 🎉, thanks for the hard work

asturur:
Hey thanks for the nice words!
I ll merge the pr as soon as i can get the CI to roll again. There is a problem with the account and is blocked

PR #6777, opened 2021-01-03
fix(fabric.Canvas): Update backgroundImage and overlayImage coordinates on zoom change

### LINKED ISSUE #6646: Rotated background image disappearing upon panning
Hi awesome FabricJS team :wave:  and first of all, thank you for all the awesome work on this library :bow: 
This is also my first bug report so please correct me if I made a mistake somewhere in the process :bow: 

## Version
4.2.0

## Test Case
(Unfortunately I was not able to reproduce behavior below in JSfiddle, but managed it in CodePen. I hope that is also OK)
https://codepen.io/driek/pen/yLJLVwW
I have placed two buttons, "rotate 48 degrees" and "rotate 54 degrees".
When pressing either button, the image will be rotated and panned accordingly to create illusion that image stays centered.
For reference, I also display a semi-transparent rectangle to indicate the original position of the background image.
When pressing the former button, the image is still visible (and so is said rectangle).
However, when pressing the latter button, the background is not visible anymore (see also behavior description below).
**NOTE:** since pressing a button will invoke `canvas.relativePan`, the buttons cannot be pressed in sequence but the page needs to be refreshed to reset to initial state before another button can be pressed.

Nodejs or browser?
Browser

Which browsers?
Verified Codepen test case with:
- Chrome (85.0.4183.121)
- Firefox (81.0)

## Steps to reproduce
- Initialize a canvas which is in "landscape" with a background image that has a wide width but shallow height. The canvas has a zoom ratio to fit the image exactly.
- Rotate the background image using `setBackgroundImage` and update `angle` option to the degree that the original position lays outside the viewport.
  (for convenience I set `originX` and `originY` to "center" and for values of `top` and `left` I used center point of minimum required dimensions to fit the rotated image in my example)
- use `relativePan` to create the illusion that the background image stays centered

## Expected Behavior
The background image to always be visible

## Actual Behavior
The image disappears. 
My best guess is that the position/dimension of the background image before rotation is being used to check whether or not to render the image as part of optimization. If you un-comment the line in my example so `setCoords`is invoked on the background image however, the image will be visible as expected. 

Currently I think this behavior is applicable to panning which means it also happens with `zoomToPoint`. (<= not part of the test case, but came across this while developing locally)
If you are aware of other places where the same kind of thing could happen, pointing out those places is highly appreciated. (for risk assessment to higher ups and QA on my side)
However, if my approach is wrong to begin with, please point that as well. Thank you in advance for your time and effort :bow:

asturur:
As someone noticed, changing the viewport does not automatically updates the coordinates of the background image.
This because responsible function was skipping it entirely.

Needs a test case

close #6646

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "nyc": "^15.1.0",
     "onchange": "^3.x.x",
     "pixelmatch": "^4.0.2",
-    "qunit": "2.9.2",
+    "qunit": "^2.13.0",
     "testem": "^3.2.0",
     "uglify-js": "3.3.x"
   },
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -676,7 +676,10 @@
      * @chainable true
      */
     setViewportTransform: function (vpt) {
-      var activeObject = this._activeObject, object, i, len;
+      var activeObject = this._activeObject,
+          backgroundObject = this.backgroundImage,
+          overlayObject = this.overlayImage,
+          object, i, len;
       this.viewportTransform = vpt;
       for (i = 0, len = this._objects.length; i < len; i++) {
         object = this._objects[i];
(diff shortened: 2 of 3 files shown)


asturur:
Well i need to merge without a CI fo now.

PR #6778, opened 2021-01-03
fix(fabric.Image): fix wrong scaling value for the y axis in renderFill

### LINKED ISSUE #6729: Resize filters work wrong, changing only scaleY (down) or scaleX (up) causes wrong ratio and undesired padding
## Version
4.2.0

## Test Case
http://fabricjs.com/lanczos-webgl

Any browser

## Steps to reproduce
You can see this behavior on http://fabricjs.com/lanczos-webgl when you try to scale image up by dragging only horizontal handles or scale down by dragging vertical handles

## Expected Behavior
Image fills the whole selection box

## Actual Behavior
Some strange padding appears under the image its selection box

Also it is not clear why there are scaleX and scaleY instead of just scale, it is not clear what is the use case for having 2 and docs don't say anything about that.

asturur:
Probably a copy-paste error, we were using scaleX in place of scaleY in a place.

close #6729


asturur:
Needs a small test.

PR #6791, opened 2021-01-06
implemented targetHasOneFlip using xor operator

KastriotSalihu:
This function was implemented using  a longer form of expressing an XOR operation. So used the javascript  '^' operator for  XOR.

### DIFF
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -240,7 +240,7 @@
    * @return {Boolean} true if one flip, but not two.
    */
   function targetHasOneFlip(target) {
-    return target.flipX ^ target.flipY;
+    return target.flipX != target.flipY;
   }
 
   /**
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -240,7 +240,7 @@
    * @return {Boolean} true if one flip, but not two.
    */
   function targetHasOneFlip(target) {
-    return (target.flipX && !target.flipY) || (!target.flipX && target.flipY);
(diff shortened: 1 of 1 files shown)


asturur:
I looked at this already then i got lost.
I noticed that the XOR operator requires numbers, how is it doing with booleans?

KastriotSalihu:
The XOR operator truth table with booleans would be:
1 1 (0)
1 0 (1)
0 1 (1)
0 0 (0)
Which is the same as what is required from the function.
Perhaps an easier to read approach, would be:
`return target.flipX != target.flipY`
Which again gives the same result.

asturur:
I was referring to the fact the XOR is a bitwise operator and is supposed to be used between numbers and not booleans.
Since  booleanA !== booleanB gives the same result i prefer that a bit more.

KastriotSalihu:
Well since boolean values can be represented as integers (1) and (0), so they would be implicitly converted to integers and be  represented as 32 bit numbers, i.e. 
00000000000000000000000000000001 (1)
00000000000000000000000000000000 (0)
basically
booleanA ^ booleanB 
would be equivalent to
Number(booleanA) ^ Number(boolenB)
where a true-> 1, and false->0
so the XOR operator is being implicitly used between numbers

asturur:
Sure, if you could just use the booleans i would happily merge it.

asturur:
I mean i understand that it will work with implict conversion, i just prefer using the booleans

PR #6823, opened 2021-01-20
implement targetHasOneFlip using shorthand

KastriotSalihu:
implemented targetHasOneFlip using boolean operator !== which returns the same results but is shorter and easier to understand


asturur:
thanks!

asturur (approved):

PR #6828, opened 2021-01-25
fix: add caveats for ES6 syntax

prog-rajkamal:
fabric does not support es6 modules. and requires bundlers to transpile import statements to commonjs imports at compile time. 

However, as browsers have started supporting ES6 modules. The readme gives the impression that fabric is supported via browser's es module which leads to confusion and errors[1]

[1] https://stackoverflow.com/questions/57715326/fabric-js-as-es6-import-in-chrome-doesnt-work

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Or you can use this instead if your build pipeline supports ES6 imports:
 import { fabric } from "fabric";
 ```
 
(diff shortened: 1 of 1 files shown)


asturur:
looks good to me. There is a with spelled as wtf

asturur:
sorry as wth

prog-rajkamal:
@asturur fixed the typo

PR #6844, opened 2021-01-31
fix(Fabric.text): Add path to text export

### LINKED ISSUE #6762: canvas.loadFromJSON doesn't work when there is text on path on canvas

asturur:
replace #6768 
close #6768 
close #6762

- Add path to text export
- add tests


src/shapes/text.class.js
```diff
@@ -1570,11 +1580,23 @@
...
    * @static
    * @memberOf fabric.Text
-   * @param {Object} object Object to create an instance from
+   * @param {Object} object plain js Object to create an instance from
    * @param {Function} [callback] Callback to invoke when an fabric.Text instance is created
    */
   fabric.Text.fromObject = function(object, callback) {
-    return fabric.Object._fromObject('Text', object, callback, 'text');
+    var objectCopy = clone(object), path = object.path;
+    delete objectCopy.path;
```
asturur, line 1588: @melchiar when i made that clone comment, i meant this:
we clone the input object, so that we can safely delete the path from our copy without mutating the input.

asturur:
This coverage in the PR feels so rewarding. Was a mess to setup.

src/shapes/text.class.js
melchiar: got it, i wasn't sure.

PR #6881, opened 2021-02-20
fix for #6841

### LINKED ISSUE #6841: Exception thrown on copy and paste in Textbox
Steps to reproduce this:

- create a new TextBox
- copy and paste some text eg
```
constructor(public name: String){
}
```
this is fine.

Now copy and paste the above code with some newlines in it:
```

constructor(public name: String){
}


```

Exception thrown because of the newlines:

```
fabric.js:28418 Uncaught TypeError: Cannot set property '0' of undefined
    at klass.insertNewStyleBlock (fabric.js:28418)
    at klass.onInput (fabric.js:28974)
insertNewStyleBlock @ fabric.js:28418
onInput @ fabric.js:28974
```

This is repoducible even on the FabricJS demo site - make sure to refresh the page between the above tests

```
http://fabricjs.com/loadfonts

```
Tested in Edge Version 89.0.774.4 (Official build) dev (64-bit)
Tested in Chrome Version 88.0.4324.96 (Official Build) (64-bit)

Same issue in both, not tested in FF

jzburns:
addressing issue #6841, adding:

```
        else if (copiedStyle) {
          // this test is required in order to close #6841
          // when a pasted buffer begins with a newline then
          // this.styles[cursorLoc.lineIndex + i] and copiedStyle[0]
          // may be undefined for some reason
          if (this.styles[cursorLoc.lineIndex + i] && copiedStyle[0]) {
            this.styles[cursorLoc.lineIndex + i][0] = copiedStyle[0];
          }
        }
```
into ``itext_behavior_mixin.js``

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -866,7 +866,13 @@
           this.insertCharStyleObject(cursorLoc.lineIndex + i, 0, addedLines[i], copiedStyle);
         }
         else if (copiedStyle) {
-          this.styles[cursorLoc.lineIndex + i][0] = copiedStyle[0];
+          // this test is required in order to close #6841
+          // when a pasted buffer begins with a newline then
+          // this.styles[cursorLoc.lineIndex + i] and copiedStyle[0]
+          // may be undefined for some reason
+          if (this.styles[cursorLoc.lineIndex + i] && copiedStyle[0]) {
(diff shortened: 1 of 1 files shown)


asturur:
I have some difficulties remember how this work.
There is a insertStyle for new line and for normal text.
I understand this IF is blocking the error, but is it pasting the style correctly then?

PR #6924, opened 2021-03-10
fix(fabric.Object): Cache CropX and CropY Object Properties

### LINKED ISSUE #6907: Modify the cropX cropY parameter that contains the clipPath object, it not be updated after canvas.renderAll()
ClipPath is applied to my picture, and the cropX cropY parameter needs to be dynamically modified, but when I modify the parameter, it not be updated after canvas.renderAll().

## Version
4.3.1

## Test Case
https://codepen.io/xh4010/pen/ZEBjrqe

browser

rileygowan:
In response to https://github.com/fabricjs/fabric.js/issues/6907, add `cropX` and `cropY` to `fabric.Image`'s list of `cacheProperties`.
> /**
     * cacheProperties:
     * List of properties to consider when checking if cache needs refresh
     * Those properties are checked by statefullCache ON ( or lazy mode if we want ) or from single
     * calls to Object.set(key, value). If the key is in this list, the object is marked as dirty
     * and refreshed at the next render
     */

Demo made using https://codepen.io/xh4010/pen/ZEBjrqe, but without explicitly setting `dirty: true`:
pug

close #6907

### DIFF
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -93,6 +93,15 @@
     stateProperties: fabric.Object.prototype.stateProperties.concat('cropX', 'cropY'),
 
     /**
+     * List of properties to consider when checking if cache needs refresh
+     * Those properties are checked by statefullCache ON ( or lazy mode if we want ) or from single
+     * calls to Object.set(key, value). If the key is in this list, the object is marked as dirty
+     * and refreshed at the next render
+     * @type Array
+     */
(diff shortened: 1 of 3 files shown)


asturur:
Thanks!

asturur:
oops @rileygowan! we added it in fabric.Object, while this is specific of fabric.Image !

asturur (approved):

asturur:
Thank you!

PR #6935, opened 2021-03-15
Bug fix for Object.toCanvasElement()

GreenFootballs:
Replaces the bounding rect calculation when an object has a shadow.

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1698,25 +1698,25 @@
     toCanvasElement: function(options) {
       options || (options = { });
 
-      var that = this,
+      var _this = this,
           utils = fabric.util,
           origParams = utils.saveObjectTransform(this),
           originalGroup = this.group,
           originalShadow = this.shadow,
           multiplier = (options.multiplier || 1) * (options.enableRetinaScaling ? fabric.devicePixelRatio : 1),
-          objs = [that],
+          objs = [_this],
           el = fabric.util.createCanvasElement(),
           rects = [];
 
-      delete that.group;
+      delete _this.group;
       if (options.withoutTransform) {
-        utils.resetObjectTransform(that);
+        utils.resetObjectTransform(_this);
       }
       if (options.withoutShadow) {
-        that.shadow = null;
+        _this.shadow = null;
       }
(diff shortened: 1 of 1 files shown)


GreenFootballs:
The node visual unit tests appear to be failing because of the same bug this commit tries to fix - it's getting a different size for a text object with a shadow.

src/shapes/object.class.js
```diff
@@ -1711,30 +1711,49 @@
...
           boundingRect = this.getBoundingRect(true, true),
-          shadow = this.shadow, scaling,
-          shadowOffset = { x: 0, y: 0 }, shadowBlur,
-          width, height;
+          shadow = this.shadow;
 
       if (shadow) {
-        shadowBlur = shadow.blur;
-        if (shadow.nonScaling) {
-          scaling = { scaleX: 1, scaleY: 1 };
```
asturur, line 1723: In the new code we are missing both the nonScaling attribute effect from the shadow and the getObjectScaling usage that will count for the groups

asturur:
UTs fail because the dataURL code changed behaviour, so that may be normal. And we need to update the tests.
I'm just not convinced this code is entirely better than the old one.

GreenFootballs:
OK, got it. The nonScaling attribute is a simple fix, but accounting for groups and active selections will take a bit more work because it needs to find an overall bounding box that includes the shadows of each element. I assume I should refrain from using ES5/ES6 methods like Array.reduce()?

asturur:
I'll get to this PR soon

GreenFootballs:
Here's another try. This one accounts for groups and active selections that contain objects with shadows by looping over the objects, getting the extended bounding box for each one and storing it in an array, then using the array to calculate the total box size.

The only thing I wasn't sure about is whether it's OK to use a forEach loop, or should I convert that to an old-school for loop?

asturur:
no this is a slow call anyway. This is not in the render hot loop. ForEach is the preferred way.

GreenFootballs:
Here's a Codepen example of this code - you can toggle the original toCanvasElement() in and out to see the difference.

https://codepen.io/green_footballs/pen/VwmNeWQ?editors=1011

asturur:
Closing this PR because we found out were the issue is.

Issue #6936, opened 2021-03-17
canvas.clear() does not remove "canvas" property of the canvas's objects


Basa0:
The "clear" function of a Canvas does not apply the same cleanup to its objects that the "remove" function does. I would expect that the objects no longer have their "canvas" property after clear, because they are no longer on that canvas.

Until I discovered that _canvas.clear_ does not remove this property, I expected that it was safe to verify if _object.canvas_ is defined to tell if that object is (still) on a canvas (rather than something like _canvas.getObjects().includes(object)_).

I would be open to attempting a fix and creating a PR, if you agree that this is a bug (unexpected behavior).

## Version
4.3.1

## Test Case
http://jsfiddle.net/mL3z25dp/

## Information about environment
browser (latest Chrome)

## Steps to reproduce
1. Add at least 1 object to a canvas
2. Call canvas.clear()
3. Verify the canvas property of one of the objects previously added

## Expected Behavior
1. object.canvas is not defined (same as if I were to call canvas.remove on each of its objects)

## Actual Behavior
1. object.canvas is defined - reference to the canvas that no longer contains that object

asturur:
I would assume you are right, to be honest i have to think if there is a reason why this done in this way.

Basa0:
Fixed in #6937

PR #6937, opened 2021-03-17
#6936: call remove on every object on canvas.clear

Basa0:
Suggested fix for #6936: calls canvas.remove on every canvas object, instead of only clearing the canvas._objects array (this would result in firing additional removal events and removing the canvas property of the objects previously added to the canvas).

I understand we should be careful if the initial implementation may have skipped this removal on purpose. In terms of performance, I don't think removal itself is particularly costly in any way. Calling canvas.remove will fire 'object:removed' on canvas and 'removed' on each object for every object, the listeners may be costly or unwanted in some way.

If this is a cause of concern, either because it may be a performance issue or because it may be considered a breaking change: we could maybe implement an option parameter to canvas.remove, maybe called "suppressObjectRemoveEvents", that would default to true - so that this new behavior would be opt-in (at least temporarily).

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -813,7 +813,9 @@
      * @chainable
      */
     clear: function () {
-      this._objects.length = 0;
+      this.forEachObject((function (obj) {
+        this.remove(obj);
+      }).bind(this));
       this.backgroundImage = null;
       this.overlayImage = null;
       this.backgroundColor = '';
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -813,9 +813,7 @@
      * @chainable
      */
     clear: function () {
-      this.forEachObject((function (obj) {
-        this.remove(obj);
-      }).bind(this));
+      this.remove.apply(this, this.getObjects());
       this.backgroundImage = null;
       this.overlayImage = null;
       this.backgroundColor = '';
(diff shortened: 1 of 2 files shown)


asturur:
thanks for opening the pr and sorry for the delay.
I think the fix is right, i m just not sure if we should fire the `remove` event here or just manually remove the canvas reference from the objects.
I'm usually worried from uncontrolled events firing.
There is also the risk that calling `remove` will try to fire a render for each object ( that eventually is then cancelled by the next one called, but still extra work ).

Basa0:
Thanks @asturur, no worries on any delays - look how late I am 😅 

I made a change to call remove only once with all the object (instead of once per object), to avoid calling requestRenderAll after each of those.

As to the overall issue, I'm unsure how to approach it. I believe it is the correct behavior: whatever handler a dev would add to removal events should theoretically be valid and even welcomed - objects are actually being removed. But because this was not the behavior until now, it may lead to unwanted results in some implementations. I'm thinking particularly from a performance perspective, that handlers of such events may be expensive in some way.

I would suggest that if there isn't a way to make an educated decision on whether this is safe to introduce, consider this as a breaking change for a future major release (should there ever be another one).

asturur:
i'm keen to risk on this change.
It seems more logical to me to have clear do a bunch of stuff rather than developers knows to hunt references of canvases and activeObjects around.

asturur (approved):

PR #6943, opened 2021-03-21
Fix bug - onMouseUp

adam-26:
If the mouse is held down and dragged off the "corner", then `__onMouseUp` **can not find the corner** because the pointer is no longer positioned over the corner. As a result, the `mouseUpHandler` is not invoked when the mouse is released.

This fixes that bug, by using the `__corner` assigned during `__onMouseDown`.

### DIFF
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -456,10 +456,7 @@
           shouldRender = true;
         }
         else {
-          var corner = target._findTargetCorner(
-            this.getPointer(e, true),
-            fabric.util.isTouchEvent(e)
-          );
+          var corner = target.__corner;
           var control = target.controls[corner],
               mouseUpHandler = control && control.getMouseUpHandler(e, target, control);
           if (mouseUpHandler) {


asturur:
there was a specific reason why we look for corner again during mouse up. We cannot assume that the target did not change between mouse down and mouse move/mouse up.
one example is that you may have locked objects that won't drag, and you want to do something dragging from a corner of object A to a corner of object B ( like connecting an arrow ).

Can you try to replicate the bug you are talking about in a fiddle so we can understand it and talk about how to fix it?

adam-26:
Ok, I can use 'onMouseUpHandler' and '_onMouseDownHandler' to achieve what I need. Is there a reason that the 'IText' '_onMouseDownHandler' is prefixed with and underscore and the up handler is not?

asturur:
Can you point me to this handler? i can't find it.

adam-26:
Lol! I had the same problem. [You can see it here](https://github.com/fabricjs/fabric.js/blob/2eabc92a3221dd628576b1bb029a5dc1156bdc06/src/mixins/itext_click_behavior.mixin.js#L99)

asturur:
That is an internal fabricJS functionality and you should not fiddle with it, you can of course, but just makes more complicated in case of updates.

Is still unclear what are you doing and what is your bug, for me, the english explanation isn't clear.

adam-26:
I need to invoke a function before the mouse starts to scale an object, and I need to run a different function after the mouse button is released and scaling finishes.

I initially tried to do this using a custom control, and that is how I found the "bug".

Issue #6951, opened 2021-03-22
Free Drawing Brush with decimate can remove the end of the line


BinaryMoon:
## Version
Latest

## Test Case
http://jsfiddle.net/BinaryMoon/z1kbepwq/

## Information about environment
This happens in all browsers.

## Steps to reproduce
Setup free drawing with a high decimate value, and then do some drawing. The line will draw fine until you release the pointer, and then the line will shorten. I assume this is because of the decimate feature.

## Expected Behavior
The line ends where I release the mouse.

## Actual Behavior
This lines gets shorter.

My suggestion is to stop decimate from deleting the last point on the line.

asturur:
yes is probably because of that.

asturur:
I mark this as beginner friendly so that anyone can jump in.

This function is inside the file pencil_brush.class.js

```diff
    /**
     * Decimate points array with the decimate value
     */
    decimatePoints: function(points, distance) {
      if (points.length <= 2) {
        return points;
      }
      var zoom = this.canvas.getZoom(), adjustedDistance = Math.pow(distance / zoom, 2),
          i, l = points.length - 1, lastPoint = points[0], newPoints = [lastPoint],
          cDistance;
+     // acting on the liimit of this loop can probably solve the issue,
+     // keeping in mind that a brush stroke an have also just 2 commands and points.lenght-2 can get to 0.
      for (i = 1; i < l; i++) {
        cDistance = Math.pow(lastPoint.x - points[i].x, 2) + Math.pow(lastPoint.y - points[i].y, 2);
        if (cDistance >= adjustedDistance) {
          lastPoint = points[i];
          newPoints.push(lastPoint);
        }
      }
      if (newPoints.length === 1) {
        newPoints.push(new fabric.Point(newPoints[0].x, newPoints[0].y));
      }
      return newPoints;
    },
```

BinaryMoon:
Hi - I was going to attempt to fix this myself but I'm having trouble building the code locally.

If I run `npm run build:watch` then I get the error:

`cp: ../fabricjs.com/lib/fabric.js: No such file or directory`

And if I run `node build.js` I get the error:

`Command failed: uglifyjs  --compress --mangle --output fabric.min.js fabric.js`

I have downloaded all the things to my computer and run npm install. Is there something I have missed?

BinaryMoon:
I found it worked if I ran `npm run build:fast` so that's what I did. :)

[...]

PR #6966, opened 2021-03-30
Fix decimate deleting end of a freedrawing line

BinaryMoon:
Stop decimate from deleting the last point on the line, whilst ensuring the line is always > 1 point in length (so that it actually displays)

Fixes this ticket: https://github.com/fabricjs/fabric.js/issues/6951

### DIFF
--- a/src/brushes/pencil_brush.class.js
+++ b/src/brushes/pencil_brush.class.js
@@ -253,9 +253,11 @@
           newPoints.push(lastPoint);
         }
       }
-      if (newPoints.length === 1) {
-        newPoints.push(new fabric.Point(newPoints[0].x, newPoints[0].y));
-      }
+      /**
+       * Add the last point from the original line to the end of the array.
+       * This ensures decimate doesn't delete the last point on the line, and ensures the line is > 1 point.
+       */
+      newPoints.push(points[l]);
       return newPoints;
     },
 
--- a/src/brushes/pencil_brush.class.js
+++ b/src/brushes/pencil_brush.class.js
@@ -246,7 +246,7 @@
       var zoom = this.canvas.getZoom(), adjustedDistance = Math.pow(distance / zoom, 2),
           i, l = points.length - 1, lastPoint = points[0], newPoints = [lastPoint],
           cDistance;
-      for (i = 1; i < l; i++) {
+      for (i = 1; i < l - 1; i++) {
         cDistance = Math.pow(lastPoint.x - points[i].x, 2) + Math.pow(lastPoint.y - points[i].y, 2);
         if (cDistance >= adjustedDistance) {
           lastPoint = points[i];
--- a/test/unit/brushes.js
+++ b/test/unit/brushes.js
@@ -21,6 +21,16 @@
       assert.deepEqual(brush._points, [], 'points is an empty array');
     });
 
+	  QUnit.test('decimate points', function(assert) {
+	    var brush = new fabric.PencilBrush(canvas);
+	    var points = [{ x: 1, y: 0 }, { x: 2, y: 0 }, { x: 3, y: 0 }, { x: 4, y: 0 }, { x: 5, y: 0 }];
+	    var distance = 6;
+	    var newPoints = brush.decimatePoints(points, distance);
+	    assert.equal(newPoints[0], points[0], 'first point is always present');
+	    assert.equal(newPoints[1], points[points.length-1], 'last point is always present');
+	    assert.equal(newPoints.length, 2, 'All points removed except first and last');
+	  });
+    
     [true, false].forEach(function(val) {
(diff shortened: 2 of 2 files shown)


asturur:
this could work, but can you check what happens for single click points with drawing brush? We should still get a circle.

src/brushes/pencil_brush.class.js
```diff
@@ -253,9 +253,11 @@
           newPoints.push(lastPoint);
         }
       }
-      if (newPoints.length === 1) {
-        newPoints.push(new fabric.Point(newPoints[0].x, newPoints[0].y));
-      }
+      /**
```
asturur, line 256: or couldn't we just check in the loop 

```js
if (cDistance >= adjustedDistance || i === l - 1) {
```
?

asturur: Because with your solution we risk to duplicate the last point in case is already there? or am i reading something wrong?

BinaryMoon:
> this could work, but can you check what happens for single click points with drawing brush? We should still get a circle.

Yes - a single click still works. Previously (the code I removed) added the last point again. Since there was only one point `point[0]` is the last point. Essentially all this code does is remove the check for line length and always add the last point to the end. In my tests it worked well on a variety of drawing speeds, just clicking, and different decimate amounts.

src/brushes/pencil_brush.class.js
BinaryMoon: That is also an option. I liked the simplicity of my solution, which actually removed code, but I can see that it may sometimes duplicate the last point. I shall have another think and try some other options.

src/brushes/pencil_brush.class.js
```diff
@@ -246,16 +246,18 @@
       var zoom = this.canvas.getZoom(), adjustedDistance = Math.pow(distance / zoom, 2),
           i, l = points.length - 1, lastPoint = points[0], newPoints = [lastPoint],
           cDistance;
-      for (i = 1; i < l; i++) {
+      for (i = 1; i < l - 1; i++) {
```
BinaryMoon, line 249: I went a slightly different way to your suggestion and changed the loop to ignore the last point, and then always add it at the end.

With this method the line will always be at least 2 points long, because newPoint has the first point added when it's initialized, and the last point added after the decimation. The decimation then only happens to the points in the middle of the line.

This removes the need to check the line length, and reduces the amount of checks made when decimating the line.

Admittedly very small improvements but I like it :)

I have tested this with single clicks and long lines.

If you're not happy with this I'd be happy to use your suggestion. I don't mind what the solution is as long as it works. :)

asturur: this looks good to me.
Way better than my try.
Now the last step. Can we write a test that will fail if this commit is removed?

Let me see if i can find a semi ready test

BinaryMoon: I was able to run the test. That's great. Thanks!

I've never done any JavaScript testing before.

asturur:
@BinaryMoon can you add a test for this change? so we can merge the pr

BinaryMoon:
Hi there - Sorry for the very delayed reply. Work and real life got in the way.

I'd really like to work on this and add a test but I have no idea what I'm doing. I've never written a unit test and I don't know how to generate the data.

Sorry :(

asturur:
No worries @BinaryMoon this is a free open project and people dedicate the time they can.
I'll guide you to write a test, is very important for open software to have tests, and i want your contrbution to come with one.
When is ready, is ready.

BinaryMoon:
Hi - thanks for your understanding! :)

I had some time to go through this today and it was easier than I thought it would be. I've added the test you mentioned and added a couple of additional tests.

Hopefully that works ok.

BinaryMoon:
Also - do you have any coding standards to follow? The formatting standards I use for work are very different to what I can see here so it makes it harder to edit things without reformatting the entire file. I've been working out the changes locally then applying the changes on Github directly - but that's not the niceset workflow :)

asturur:
Hi @BinaryMoon sorry for not answering, i moved fabricJS emails out of my inbox and i m still getting adjusted to it.
In the repo you find a lint command and a lint command for tests.
If you run `npm run lint` or `npm run lint_tests` you get the check for the only code standards that are catched as regressions.
Other than that we still use ES5, so that is the other coding standard you have to follow.
We still keep ie11 working with no transpiler ( i think this is the last version that does it )

asturur:
In theory your IDE should pick up the eslint file from the project and adjust itslef to respect the rules.
What ide do you use?

asturur (approved):

BinaryMoon:
Thanks for merging this! :)

I have been using Atom to edit this. When I saved the file(s) it reformatted the same thing. I don't know if that's because of an extension or something else but I suspect that's not what you want.

I'll see if I can work out how to get the eslint file to be recognised.

Issue #6964, opened 2021-03-30
Free drawing over non-transparent .PNG file parts


PhilCliv:
The _globalCompositeOperation_ in DrawingMode seems not to be working the way I expect. Or my approach is wrong. 
I'm trying to do **free drawing** only **on non-transparent parts of a PNG image**. 
I've been doing deep searches on how to make it work but with no success. Here is what I have so far:
    
    var canvas = new fabric.Canvas('canvas');
    canvas.backgroundColor = '#ffffff';

    fabric.Image.fromURL('https://i.pinimg.com/originals/49/af/b1/49afb1d21ae594cb7ac3534a15383711.png', function (myImg) {
        var img1 = myImg.set({ left: 0, top: 0, width: 503, height: 343 });
        canvas.add(img1);
    });

    canvas.isDrawingMode = 1;
    canvas.getContext().globalCompositeOperation = 'source-atop';
    canvas.freeDrawingBrush.color = "purple";
    canvas.freeDrawingBrush.width = 10;
    canvas.renderAll();

Here is an example of the result I would like to have, but it's working in native canvas (without Fabric.js) https://stackoverflow.com/a/49228110/6660372
Any suggestions for a solution using Fabric.js? Thank you!

asturur:
No globalComposite operation won't work because the drawing happens on a separate layer and gets merged later.
You can still do it using the image as a clipPath object.
But i would need to write an example because i m not sure is easy to get there or supported out of the box.
So trying alone could be a huge use of time with uncertain results.
We have at least another issue asking the same thing.

PhilCliv:
Thank you for the answer. I tried clipping with the image but it doesn't seem to work like this. Here is the jsfiddle link https://jsfiddle.net/PhilCliv/dpf840qn/14/
Is it something like this you were talking about?

```
var canvas;

$(function () {
    canvas = window._canvas = new fabric.Canvas('canvas');
    canvas.backgroundColor = '#efefef';
    canvas.isDrawingMode= 1;
    canvas.freeDrawingBrush.color = "purple";
    canvas.freeDrawingBrush.width = 10;
     var imgInstance = fabric.Image.fromURL('https://i.pinimg.com/originals/49/af/b1/49afb1d21ae594cb7ac3534a15383711.png', function (img) {
        canvas.add(img);
        canvas.renderAll();
    });
 
    canvas.renderAll();

    canvas.on("path:created", (e) => {
      var path = e.path;
      path.set({selectable:false})
[...]

Issue #6356, opened 2020-05-27
The eraser is not for all path objects


Mamikon95:
Hello, I have such a problem, there is an eraser, I want it to not work for all objects. For example, there is a red path and a green path. When using the eraser, I want the green color to be removed, and so that it does not hurt the red path, in this example, for example - http://prntscr.com/soyu6l. In the combat project, there are more colors and everything is dynamic.

Test Case - https://jsfiddle.net/hybumkcp/8/

asturur:
Hi @Mamikon95 there is no official support for erasers and i won t be able to provide an example for you

ShaMan123:
The eraser has moved https://github.com/ShaMan123/erase2d

Issue #6261, opened 2020-04-11
Eraser


Lyqxyz:
Add eraser function to free painting

asturur:
We have open issues for suggestions already.
Erases can't be made easily, you have to make your own combination of 2 drawing paths, one that draws and the other that cancel with the globalCompositeOperation.
duplicate of #5742

Issue #5742, opened 2019-06-05
freehand eraser with real time erasing


TomVanHaver:
**Is your feature request related to a problem? Please describe.**
We would like to implement a classic eraser functionality that can erase "real time".

**Describe the solution you'd like**
When freehand drawing in erase mode (= globalComposteOperation set to "destination-out" on the path) over an existing drawing, the erasing should be applied "real time", so onMouseMove, not onMouseUp.

**Describe alternatives you've considered**
- set "globalCompositeOperation" to "destination-out" on the brush
- apply the erase path intermediate (onMouseMove)
- apply the erasing on the "main" canvas, and not on the topcanvas (**most plausible idea ?**)

**Additional context**
image a: onMouseOver while erasing.
![image a](https://user-images.githubusercontent.com/7240053/58961862-4bf44300-87aa-11e9-96f6-40e0cd56e856.png)

image b: erased after onMouseUp.
![image b](https://user-images.githubusercontent.com/7240053/58961865-4bf44300-87aa-11e9-8ce7-9dd1ae3accf8.png)

alexwenbj:
@TomVanHaver Hi,Tom,I am looking for a eraser for the free drawing scence,would you please share your eraser code for us?Thank you.

PR #6994, opened 2021-04-07
Eraser 

ShaMan123:
## This PR adds an ERASER brush. 
**You can test it in [codesandbox](https://codesandbox.io/s/focused-christian-7s5gx?file=/src/EraserBrush.js).**

### Inspiration 
https://github.com/fabricjs/fabric.js/issues/1225#issuecomment-499620550

## Object Logic
1. When path is created it is attached to all objects that intersect the eraser path
2.  Add the new path to existing object eraser paths
3. use `clipPath` attr and `EraserPath` to achieve erasing effect

## Brush Rendering Logic
`EraserBrush` extends `PencilBrush`.
Use top context to achieve an erasing effect while drawing - draw canvas on top context and leave only non erasable objects on main context. This way when the user erases the top context is drawn out and the main context with non erasables is revealed. 🚀💪👍👍
1. Draw all non-erasable objects on **bottom ctx** with the exception of overlays `renderBottomLayer`
2. Draw all objects and erasable drawables on top ctx `renderTopLayer`
3. Draw eraser `fabric.PencilBrush#_render` at `renderTopLayer`
4. Draw non-erasable overlay `renderOverlay`

The canvas is prepared for every step/layer according to appropriate logic, hiding unnecessary objects and restoring them after painting.

Layer 1 is drawn on main/bottom ctx, the rest on top ctx.
Layers 2, 3 are affected by the brush.

### Usage
added `erasable` option to `Object` controlling whether it can be erased by the brush.
EraserBrush is the same as PencilBrush in terms of use.

### Caveats
~~1. Currently using `absolutePositioned` for clipPath - this causes a problem when selecting and moving an erased object. The eraser paths don't move with the object. I'm pretty sure there's some kind of transform function available to render the clipPath relatively. Would be glad to know which.~~ Fixed 6d9c7ff
~~2. `onMouseMove` is a challange - I have changed the drawing context (from `canvas.contextTop` to `canvas.getContext()`) so that the brush behaves as a clip path for the entire canvas. Not desirable though, in case of objects which are not erasable. I've seen that animated objects are drawn on a different canvas as well so that needs to be addressed.~~ DONE 127de4b
3. ~~I haven't addressed export/import of eraser data as part of the clip path~~ ~~Need to handle `toSVG`~~ export to svg now supported 1257489 47e6873, **REMAINING** `fromSVG`
4. ~~Using animation when erasing needs handling~~ DONE

### Performance
Superb!

### Motivation is obvious
#1225 #5742 #6356 #6261 #5595 #5273 #5188 #6964


ShaMan123:
Regarding 302a23b
I suggest renaming `absolutePositioned: boolean` to `clipPathPosition: 'relative' | 'absolute' | 'fixed'`  so data export/import can be addressed properly

andreafalzetti:
I'd love to have the eraser functionality in FabricJS, I can help with testing this. @ShaMan123 how can I reach out to you for a chat?

ShaMan123:
@andreafalzetti here?
Take a look at the code sandbox.
I've updated the PR description.

asturur:
Sorry for not being able to look at this yet.
For easyness of review, please do not commit modified dist folder.
```
git checkout master dist
```
will retore dist folder to latest master and clean up the diff view here.

ShaMan123:
> Sorry for not being able to look at this yet.
> For easyness of review, please do not commit modified dist folder.
> 
> ```
> git checkout master dist
> ```
> 
> will retore dist folder to latest master and clean up the diff view here.

Done
```
git checkout upstream/master dist
```

src/brushes/eraser_brush.class.js
```diff
@@ -0,0 +1,272 @@
...
+      }));
+      this._parent = parent;
+      this._paths = paths || [];
+    },
+
+    _render: function (ctx) {
+      this.callSuper('_render', ctx);
+      this._paths.forEach(({ path, transformMatrix }) => {
+        ctx.save();
+        ctx.transform(transformMatrix[0], transformMatrix[1], transformMatrix[2], transformMatrix[3], transformMatrix[4], transformMatrix[5]);
```
ShaMan123, line 236: I need help here, struggling calculating the transform matrix for the eraser path so I simply use `ctx.transform`.
This isn't good practice - it should somehow be calculated and added to the path's transform matrix

ShaMan123: Then this class will behave more like a group.
see other comment.

src/brushes/eraser_brush.class.js
```diff
@@ -0,0 +1,272 @@
...
+          obj.calcTransformMatrix()
+        );
+
+        var clipObject;
+        if (!obj.eraser) {
+          clipObject = new fabric.StrokeClipPath(obj);
+        } else {
+          clipObject = obj.clipPath;
+        }
+        clipObject.addPath(path, transformMatrix);
```
ShaMan123, line 146:
```suggestion
        fabric.util.addTransformToObject(path, transformMatrix);
        clipObject.addPath(path);
```

> I need help here, struggling calculating the transform matrix for the eraser path so I simply use ctx.transform.
This isn't good practice - it should somehow be calculated and added to the path's transform matrix

ShaMan123 (commented):
I need help here, struggling calculating the transform matrix for the eraser path so I simply use `ctx.transform`.
This isn't good practice - it should somehow be calculated and added to the path's transform matrix

andreafalzetti:
> @andreafalzetti here?
> Take a look at the code sandbox.
> I've updated the PR description.

Are you available for freelance work around FabricJS?

ShaMan123:
> > @andreafalzetti here?
> > Take a look at the code sandbox.
> > I've updated the PR description.
> 
> Are you available for freelance work around FabricJS?

PM

andreafalzetti:
> > > @andreafalzetti here?
> > > Take a look at the code sandbox.
> > > I've updated the PR description.
> > 
> > 
> > Are you available for freelance work around FabricJS?
> 
> PM

PM? Where?

src/brushes/eraser_brush.class.js
```diff
@@ -0,0 +1,361 @@
...
+      this._paths = paths || [];
+      this._paths.forEach(function (p) {
+        p.path.set({ globalCompositeOperation: "destination-out" });
+      })
+    },
+
+    setParent: function (parent) {
+      this.set({
+        width: parent.width,
+        height: parent.height,
```
ShaMan123, line 295: When `parent` is `fabric.Path` it behaves a bit differently. `width` and `height` aren't exactly correct so the clip path removes the edges. How can I handle this correctly?

ShaMan123:
This PR is mature and stable.

ShaMan123:
These final commits enable erasing canvas' `overlayColor` `overlayImage`.

ShaMan123:
I see that all brushes draw on top of canvas overlay. 
This doesn't seem to be the expected behavior.
I can migrate the logic for it from `EraserBrush` to `BaseBrush`
What do you think?

ShaMan123:
Finished on my part.
Waiting for comments regarding my review.
Thanks!

asturur:
hi @ShaMan123 , so there are so many things to talk about this PR.
First and foremost this is great work and thanks for putting up this.
Then we need to decide how to package this. I do not think this should be part of the default fabricJS distruibution.
This is an application feature, a made tool for a picture editor, and shouldn't be part of the graphic library.

Many features create many interaction between them and as you noticed with animations, is not easy to keep everything work together. Developers that need an eraser should definetely pick up this but as a plugin.

There are a couple of things in the fabricJS future and one of them is handling the upper canvas with a stack of operations and define an interface for brushes, and this PR may make too much harder to create those while maintain full compatibility.

Those are the reasons for me to thinking this should be a plugin. 

There is a lot of code in the PR, and the PR description does not explain it clearly.
Can you talk about the eraserPath and what is role?
Are the options of having objects erasable or not, necessary? Can't we just explain how to obtain it but not making it an option? same apply for background image/overlay.

ShaMan123:
Thanks @asturur. This library is awesome and pretty simple to start with (though docs are challenging). 
I had to build an eraser for a project of mine so I went for it and saw no reason not to publish back. 

### Regarding packaging:
I think that the `EraserBrush` must be available along side `PencilBrush`. I'd even say that `PencilBrush` without an eraser is very missing UX. This was my motivation to code it and in fact it was the biggest disadvantage when researching `fabric` and vs. `konva` in particular. I terms of size it's pretty slim.
I don't exactly understand the considerations though. I simply think it belongs with the rest of the brushes.

### Regarding animations:
This is a technical matter that I haven't addressed yet. I am hopeful of the possibilities of making it work with the eraser. I couldn't track the animation rendering logic from the code so I couldn't wrap my head around it to provide a solution as of yet. Would love an explanation of how an animated object is rendered and by whom. You can uncomment the `animate()` line in the example to see how it behaves, but again I didn't do anything to make it work so it makes sense that it doesn't.

I believe that the interactions between the features is what makes this library so powerful. So, I strongly advocate for the continuation of this approach.

If you would like to explain more about the upper canvas that could be helpful (for other contributors as well). I have written comprehensive comments embedded in the code regarding the rendering logic of the upper canvas. I don't mind discussing it here. It will probably help us both, me for understanding how to make animations compatible and you for understanding this PR.

### `EraserPath` 
`EraserPath` is a workaround for clipping paths that are **strokes and not fills**. 
1. Clipping is done with the **fill** of the clip path, so to enable **clipping paths by their stroke** and achieving an eraser effect, `EraserPath` fills a rect where the object that it needs to clip is drawn.
2. Then it draws the paths that the `EraserBrush` has drawn onto the rect in `globalCompositionMode=destination-out`. This removes the path from the drawn rect. The result of this is an erased rect.
3. `EraserPath` is attached to it's owning object as a clip path. So basically it achieves what the `inverted` prop achieves, only for stroked objects.
4. There is a part of the logic in `EraserPath` regarding the transform matrix of each path that I struggled with. I left comments in the review above. I tried to transform the paths before attaching them to `EraserPath` so it could behave more like a group. This is crucial for svg import/export and it is the best approach that is compatible with the structure of the entire library. I would be glad for help in this matter. Transform matrices are not my forte to say the least. Because the `EraserPath` is drawn as a clip path it is really hard (for me) to understand what are the transforms that are applied to it or how to calculate them.
5. Once (4) is resolved I think it will be a good idea to expose a simple method to handle the erased paths of an object for adding, removing etc. This is minor work.

### `erasable` option:
Why are you against it? Could you explain?.
I think it's fantastic and so simple. 
Did you play with the sandbox? 
With so little code in the [example](https://codesandbox.io/s/async-water-johjz?file=/src/App.tsx) I achieved such a flexible and smooth drawing experience. For me as a developer it says a lot about the maturity and state of a library.
The way I see it `erasable` option fits exactly into the DNA of `fabric` as it provides fine grained **configurability and simplicity**. And it makes erasing ever more powerful. For me as a developer, when I research a library to see if it fits my needs, configurability and simplicity are of the highest importance. This is something that fabric is good at.
By the way, the lack of erasing was something that made me think hard if to pick `fabric` rather than `konva`. 
I don't see a reason why not to support this option. It defaults to `true` and can be handled globally by `fabric.Object.prototype.erasable = true/false;`. There's no reason for it to be misleading either.
I think that something like `absolutePositioned` or `inverted` are much more esoteric and confusing (I myself misplaced these props and assigned them to the clip path owner instead of on to the clip path itself).
I put in more work on the background/overlay just for compatibility. In my opinion, it doesn't make sense being able to erase a triangle and not being able to erase the overlay color in terms of UX. I myself don't use overlays in my project but it must be supported for the validity of the feature. Any other option makes the `EraserBrush` awkward and damages it's scope and will require developers a long excruciating workaround (let alone time) to get it to function properly and statefully.

What do you think?

asturur:
Ok let's clarify better.
I'm not against the erasable option, i just want carefully consider what goes in as default
FabricJS has to cut down on features. And has to because is unmaneageable right now.
Every library that survives in time has a reasonably small core and a good interface to define additional features.
One first step i did was definitng the control api and let people add icon controls on their own way.
We need to solve filters yet and brushes too.
Eventually we move all the brushes out, apart the pencil.

Said so, the eraser functionality would be still in this repo, or later in a repo in this organization dedicated to plugins ( whenever we have it ) with clear instructions on how to enable them.

EraserPath:

So right, you need to clip by stroke, instead of clip by fill that is what clipPath does. Why not  merging and using masks? The mask PR would solve you this right away, having a black pencil brush and a white background will act as the eraser brush maybe without need to use the eraserPath.

I still need to read every single line of code, i may have more questions or some questions will be answered automatically when i read everything.

I'll also look at the transform issue and see if i understand what is your concern

I'll get back on the erasable option too as soon as i understand why is necessary rather than being true for everything and not optional

asturur:
```
I think that something like absolutePositioned or inverted are much more esoteric and confusing (I myself misplaced these props and assigned them to the clip path owner instead of to the clip path itself).
```

Yes those 2 options aren't straight forward, but when clipPath was added, there was need to have it eventually not moving but influencing a single object and not the entire canvas.

The inverted can probably get removed because mask will do a better job at it, but before removing it we need to have a major version bump

ShaMan123:
@asturur Sorry for changing things. I had an idea and it turned out to be great so I refactored the code to implement better logic.

ShaMan123:
My original approach was to clone the canvas and draw the bottom ctx with the clone but that was completely unnecessary and bad for performance.
The new approach is much simpler.


## Rendering Logic
1. Draw all non-erasable objects on **bottom ctx** with the exception of overlays `renderBottomLayer`
2. Draw all erasable objects on top ctx `renderTopLayer`
3. Draw eraser `fabric.PencilBrush#_render` at `renderTopLayer`
4. Draw non-erasable overlay `renderOverlay`

The canvas is prepared for every step/layer according to appropriate logic, hiding unnecessary objects and restoring them after painting.

ShaMan123:
BTW CD git actions are spamming my email like crazy 😅

andreafalzetti:
> Ok let's clarify better.
> I'm not against the erasable option, i just want carefully consider what goes in as default
> FabricJS has to cut down on features. And has to because is unmaneageable right now.
> Every library that survives in time has a reasonably small core and a good interface to define additional features.
> One first step i did was definitng the control api and let people add icon controls on their own way.
> We need to solve filters yet and brushes too.
> Eventually we move all the brushes out, apart the pencil.
> 
> Said so, the eraser functionality would be still in this repo, or later in a repo in this organization dedicated to plugins ( whenever we have it ) with clear instructions on how to enable them.
> 
> EraserPath:
> 
> So right, you need to clip by stroke, instead of clip by fill that is what clipPath does. Why not merging and using masks? The mask PR would solve you this right away, having a black pencil brush and a white background will act as the eraser brush maybe without need to use the eraserPath.
> 
> I still need to read every single line of code, i may have more questions or some questions will be answered automatically when i read everything.
> 
> I'll also look at the transform issue and see if i understand what is your concern
> 
> I'll get back on the erasable option too as soon as i understand why is necessary rather than being true for everything and not optional

+1 on the idea of a core only package. What about having plugin-based repos in this org, which publish independent NPM packages, and a method in FabricJS core to load them? 

I guess it's already possible to extend FabricJS but if there's an official way of doing it, such as a list of official and community plugins for FabricJS, it might get more traction and appetite for developing and publishing extra features on top of FabricJS

asturur:
A great start would have this merged, with no default build, an option to define the eraser module to add everything is need to build it.
a page demo on fabricJS demos on how to use it ( basically replicating the codesandbox in a codepen since the website support those.

I'm not strict on how this actually work since is an option and since i won't have time to dig in the exact functionality of this but also i think the effort should be considered on this work and not get it lost.

ShaMan123:
Great.
Where does the source code of the docs/demos reside?

asturur:
we can address the demo in a follow up PR, is in another repo. fabricjs/fabricjs.com

src/canvas.class.js
```diff
@@ -418,6 +425,11 @@
         this.clearContext(this.contextTop);
         this.contextTopDirty = false;
       }
+      // while erasing the brush is in charge of rendering the canvas so we return
+      if (this.isErasing()) {
+        this.freeDrawingBrush._render();
+        return;
+      }
```
asturur, line 432: i do not think we should have this. renderAll has the duty to render, and should not know about the actual brush. Devs may want to call it at some point and in case, is renerTopLayer that should refresh the brush.

src/canvas.class.js
```diff
@@ -408,6 +408,13 @@
       return objsToRender;
     },
 
+    isErasing: function () {
+      return this.isDrawingMode &&
+        this.freeDrawingBrush &&
+        this.freeDrawingBrush.type === 'eraser' &&
+        this.freeDrawingBrush._isErasing;
+    },
```
asturur, line 416: can you extend this in the brush file ? So we do not add functionalities to the canvas that aren't used by default.

src/brushes/eraser_brush.class.js
```diff
@@ -0,0 +1,592 @@
+(function () {
```
asturur, line 1: Let's rename this file as eraser_mixin, and let's move it in the mixin folder, since it contains a brush and changes to canvas and object.

src/brushes/eraser_brush.class.js
```diff
@@ -0,0 +1,592 @@
...
+  var _proto = fabric.util.object.clone(fabric.StaticCanvas.prototype);
```
asturur, line 2: why do we need to clone the prototype? can't we just reference it?

src/shapes/object.class.js
```diff
@@ -592,7 +592,7 @@
       'top left width height scaleX scaleY flipX flipY originX originY transformMatrix ' +
       'stroke strokeWidth strokeDashArray strokeLineCap strokeDashOffset strokeLineJoin strokeMiterLimit ' +
       'angle opacity fill globalCompositeOperation shadow visible backgroundColor ' +
-      'skewX skewY fillRule paintFirst clipPath strokeUniform'
+      'skewX skewY fillRule paintFirst clipPath erasable strokeUniform'
```
asturur, line 595: erasable isn't a state property. A state property is meant to cause a re-render or fire a modified when changed in stateful mode. I do not think erasable is useful in that case.

src/shapes/object.class.js
```diff
@@ -645,6 +645,11 @@
      */
     absolutePositioned: false,
 
+    /**
+     * Indicates whether this object can be erased by @class fabric.EraserBrush
+     */
+    erasable: true,
```
asturur, line 651: Let's add this in the mixin file, where getEraser is, and let's add proper JSDOC ( with @type )

src/shapes/object.class.js
```diff
@@ -866,6 +871,7 @@
             globalCompositeOperation: this.globalCompositeOperation,
             skewX:                    toFixed(this.skewX, NUM_FRACTION_DIGITS),
             skewY:                    toFixed(this.skewY, NUM_FRACTION_DIGITS),
+            erasable:                 this.erasable
```
asturur, line 874: by default to JSON exports only properties that influence the rendering of the objects.
Everything that is related to behaviour of objects isn't exported.
You can add this in the mixin wrapping toObject

```js
var toObject = fabric.Object.prototype.toObject;
fabric.Object.prototype.toObject = function(additionalProperties) {
  return toObject.call(this, ['erasable'].concat(additionalProperties));
}
```

src/canvas.class.js
ShaMan123: I agree. However, eraser rendering logic hijacks the canvas to render in layers to achieve an erased effect.
I can move it to the mixin logic. What do you think?

src/brushes/eraser_brush.class.js
ShaMan123: Because it creates a circular ref after it's extended

src/brushes/eraser_brush.class.js
ShaMan123: DONE
c80f5645

src/shapes/object.class.js
ShaMan123: DONE

src/shapes/object.class.js
ShaMan123: DONE

src/shapes/object.class.js
ShaMan123: DONE

src/canvas.class.js
ShaMan123: Moved logic to mixin

src/canvas.class.js
ShaMan123: DONE

ShaMan123 (commented):
@asturur There's still the matrix issue
Please take a look.

src/mixins/eraser_brush.mixin.js
```diff
@@ -0,0 +1,653 @@
...
+          clipObject = new fabric.EraserPath();
+          clipObject.setParent(obj);
+        } else {
+          clipObject = obj.clipPath;
+        }
+
+        var transformMatrix = fabric.util.invertTransform(
+          obj.calcTransformMatrix()
+        );
+        //fabric.util.applyTransformToObject(path, transformMatrix);
```
ShaMan123, line 444: @asturur  This is where I need your help.
How can the matrix be applied to the path so there won't be a need to handle it in EraserPath.

src/mixins/eraser_brush.mixin.js
```diff
@@ -0,0 +1,653 @@
...
+        var m = o.transformMatrix;
+        m && ctx.transform(m[0], m[1], m[2], m[3], m[4], m[5]);
+        o.path.render(ctx);
+        ctx.restore();
+      });
+    },
+
+    addPath: function (path, transformMatrix) {
+      path.set({ globalCompositeOperation: 'destination-out' });
+      this._objects.push({ path: path, transformMatrix: transformMatrix });
```
ShaMan123, line 593: if the transform matrix is applied to path before it is added then this line is redundant

src/mixins/eraser_brush.mixin.js
```diff
@@ -0,0 +1,653 @@
...
+      this.dirty = true;
+    },
+
+    toObject: function (propertiesToInclude) {
```
ShaMan123, line 597: if the transform matrix is applied to path this needs refactoring

src/mixins/eraser_brush.mixin.js
```diff
@@ -0,0 +1,653 @@
...
+        clipPath: parent.clipPath
+      });
+    },
+
+    _render: function (ctx) {
+      this.callSuper('_render', ctx);
+      this._objects.forEach(function (o) {
+        ctx.save();
+        var m = o.transformMatrix;
+        m && ctx.transform(m[0], m[1], m[2], m[3], m[4], m[5]);
```
ShaMan123, line 585: if the transform matrix is applied to path this can be removed

src/mixins/eraser_brush.mixin.js
ShaMan123:
```js
fabric.util.applyTransformToObject(path, transformMatrix);
```
This didn't work

ShaMan123 (commented):
Could you help with applying the matrix to path?

ShaMan123:
0e227a5 adds support for **nested selective erasing**.
Supports erasing a group with both erasable and non-erasable objects (these can be groups as well).

ShaMan123:
I have managed to solve the transform matrix issue thanks to this awesome tutorial of yours [using-transformations](http://fabricjs.com/using-transformations).
It introduced a minor bug with overlay color. Will be solved soon.

src/mixins/eraser_brush.mixin.js
ShaMan123: 7fd927d

src/mixins/eraser_brush.mixin.js
ShaMan123: 7fd927d

src/mixins/eraser_brush.mixin.js
ShaMan123: 7fd927d

src/mixins/eraser_brush.mixin.js
ShaMan123: 7fd927d

ShaMan123:
f6126bf This introduces a fine option for handling the bounding box of a group. It allow to control the group's bounding box by setting the the objects that the group's bounding box should be bound to.
It is crucial for deprecating `EraserPath` in favor of `fabric.Group` because it allows to add multiple erasing paths to a group (which will act as the object's clip path) without changing it's size.
I believe this functionality is desired in other cases as well

asturur:
been out a bit. I ll update myself on this pr asap

ShaMan123:
export to svg now supported
1257489
47e6873

I need help to handle `fromSVG`

jianglinjie-nd:
Hi, @ShaMan123 . I found a bug in the demo when erased drawing path.  https://codesandbox.io/s/focused-christian-7s5gx?file=/src/EraserBrush.js

https://user-images.githubusercontent.com/43067064/116519950-ad74ac00-a904-11eb-8380-0ba04f69ccc8.mov

jianglinjie-nd:
@ShaMan123 And I have another question. Is it possible deleting the object when has been erased entirely?

ShaMan123:
@jianglinjie-nd I've noticed the bug as well #7042 #7043
Regarding your question you can listen to the event `erasing:end` 6291cf5 and test if the path covers the object using intersection logic found at [`src/mixins/object_geometry.mixin.js`](../tree/master/src/mixins/object_geometry.mixin.js) and transparency testing [src/canvas.class.js#isTargetTransparent](../tree/master/src/canvas.class.js#L474) logic
```js
canvas.on("erasing:end", ({ targets, drawables }) =>
// test intersection and transparency
          console.log(
            "objects:",
            targets.map((t) => t.type),
            "drawables: canvas bg/overlay image/color:",
            Object.keys(drawables)
          )
        );

```

asturur:
With the changes to group bounding box we moved a step away from the mergeability imho.
What was wrong with eraser path that required to make this change? I see the comment that introduces it, but i can't  figure out the story about it.
Why the standard bounding box for group wasn't good enough?

ShaMan123:
> [f6126bf](https://github.com/fabricjs/fabric.js/commit/f6126bf4268e08a689b14e4fb730923bef78a771) introduces a fine option for handling the bounding box of a group. It enables controlling the group's bounding box by setting the the objects that the group's bounding box should be bound to.
> It is crucial for deprecating `EraserPath` in favor of `fabric.Group` because it allows to add multiple erasing paths to a group (which will act as the object's clip path) without changing it's size.
> I believe this functionality is desired in other cases as well

I wanted `EraserPath` to conform to fabric's object behavior/logic. It was too non standard in logic:
- It was a rect that behaved like a group (but didn't have the proper logic to resolve group behavior - it was causing bugs in edge cases)
- The paths had `transformMatrix` property because I wasn't able to resolve the transform properly). 
The issue with the group that I had to fix was that it calculates it's bounds by the bounds of it's objects. This behavior it not wanted for `EraserPath` because when a path is added to the `EraserPath` it should not be accounted for calculating the group's bounds.

I made the change to the group non breaking and I saw issues requesting this functionality **BUT** I can instead re-introduce `EraserPath` as `Eraser`, a subclass of `Group`. *No problem.*
What do you think?

build.js
```diff
@@ -232,6 +232,9 @@ var filesToInclude = [
 
   ifSpecifiedInclude('textbox', 'src/shapes/textbox.class.js'),
   ifSpecifiedInclude('interaction', 'src/mixins/default_controls.js'),
+
+  //  extends fabric.StaticCanvas, fabric.Canvas, fabric.Object, depends on fabric.PencilBrush and fabric.Rect
+  ifSpecifiedInclude('freedrawing', 'src/mixins/eraser_brush.mixin.js'),
```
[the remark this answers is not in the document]
ShaMan123: 00a191c1

ShaMan123:
Strangely enough the eraser brush wasn't using the group logic I've added. Came across it by chance so I've simply removed it and everything seems to work as expected

ShaMan123:
So it seems ready to merge.
The only caveat is `loadFromSVG`

build.js
```diff
@@ -232,6 +232,9 @@ var filesToInclude = [
 
   ifSpecifiedInclude('textbox', 'src/shapes/textbox.class.js'),
   ifSpecifiedInclude('interaction', 'src/mixins/default_controls.js'),
+
+  //  extends fabric.StaticCanvas, fabric.Canvas, fabric.Object, depends on fabric.PencilBrush and fabric.Rect
+  ifSpecifiedInclude('erasing', 'src/mixins/eraser_brush.mixin.js'),
```
asturur, line 237: 👍

asturur:
I my opinion this is way better. For the svg quirks, it can't be perfect first time. it never is.
We just miss removing the changes to fabric.js, fabric.min.js and package.json.
You should be able to do so using git checkout specifically from the original master branch or just copy over the actual one.

ShaMan123:
reverted dist + package-lock.json to  0d7db7b

package.json
```diff
@@ -40,7 +40,7 @@
   },
   "license": "MIT",
   "scripts": {
-    "build": "node build.js modules=ALL requirejs exclude=gestures,accessors",
+    "build": "node build.js modules=ALL requirejs exclude=gestures,accessors,erasing",
     "build:fast": "node build.js modules=ALL requirejs fast exclude=gestures,accessors",
```
ShaMan123, line 44:
```diff
-"build:fast": "node build.js modules=ALL requirejs fast exclude=gestures,accessors",
+"build:fast": "node build.js modules=ALL requirejs fast exclude=gestures,accessors,erasing",
??
```

asturur:
Merging now. Sorry if it took a while but i have been very busy this period and did not work on fabricJS much.

ShaMan123:
I see the tests fail because of `erasable` option that isn't in the test files...

package.json
asturur: yes

asturur:
Let me see if with build:fast fixed it pass.

asturur:
Ok there are just some minor lint issues. can you run `npm run lint` locally and fix them?

ShaMan123:
lint passed 0f81013

asturur (approved):

asturur:
Would be cool if we can work on a demo for the main website

johannb75:
Hi ! Do you know when this feature will be published to the NPM package ?

asturur:
Next release is in automatically. Soon since we also merged RTL text

ShaMan123:
> Would be cool if we can work on a demo for the main website

@asturur Could you point me to the exact place where I can add a demo in the website repo?
I don't understand what goes where

asturur:
https://github.com/fabricjs/fabricjs.com/blob/gh-pages/posts/demos/_posts/2020-2-15-custom-control-render.md

This is an example of the demo.
You create a new file in the same folder, and start writing.

To load the website locally:
you clone the repo in the same parent folder of fabricJS, so that fabric.js and fabricjs.com are siblings.
Then in fabricjs.com folder you run `jekyll serve`.
If you are on windows, i have no idea how to exactly have it running. It works well on mac and linux.

goldingdamien:
Is there any user documentation for this?
I installed version 4.5.0 but there is no fabric.EraserBrush.
How do I reference this feature from the fabric npm package?

ShaMan123:
https://github.com/fabricjs/fabricjs.com/pull/306

It is excluded from the default build so you need to create a custom build

goldingdamien:
@ShaMan123 

I would like to use this feature while using the npm module for simplicity.
Can you confirm, doing the following is correct.
Thank you.

Add and execute below:
https://github.com/ShaMan123/fabric.js/blob/6d9c7fffa4c5c6ece49a97380fa0965aeb7bb7e1/src/brushes/eraser_brush.class.js

```
// Enable EraserBrush:
canvas.freeDrawingBrush = new fabric.EraserBrush(canvas);
canvas.freeDrawingBrush.width = 10;
canvas.isDrawingMode = true;

// Optionally use "erasable" property to enable/disable ability for object to be erased:
const t = new fabric.Triangle({
  top: 300,
  left: 210,
  width: 100,
  height: 100,
  fill: "blue",
  erasable: false
});

```

ShaMan123:
> @ShaMan123
> 
> I would like to use this feature while using the npm module for simplicity.

As would I.
Looks good.

matsu-motsu:
@ShaMan123 
> fabricjs/fabricjs.com#306

Thank you!
However, It seems to be incompatible with the SprayBrush.
After spraying, I used an eraser and my computer's memory boiled.
https://codepen.io/matsu-motsu/pen/xxqWjdZ

ShaMan123:
> @ShaMan123
> 
> > [fabricjs/fabricjs.com#306](https://github.com/fabricjs/fabricjs.com/pull/306)
> 
> Thank you!
> However, It seems to be incompatible with the SprayBrush.
> After spraying, I used an eraser and my computer's memory boiled.
> https://codepen.io/matsu-motsu/pen/xxqWjdZ

@matsu-motsu  Great, look at the fix #7100

asturur:
You can use it from the npm module anyway, you can add a post-install script in your installation that will rebuild fabric adding the module.

goldingdamien:
Thanks latest fix got rid of erased area returning after mouseup issue.

goldingdamien:
@ShaMan123 There is an issue in my environment where I am using IText and when using this eraser function, the last IText object disappears while mouse is down but returns on mouseup.

ShaMan123:
@goldingdamien I can't reproduce https://codesandbox.io/s/exciting-architecture-bcgej?file=/src/App.tsx

goldingdamien:
@ShaMan123 Ok, thanks for checking, if I can replicate the problem in the sandbox I'll let you know.

tanzyy96:
I'm currently using this feature and it seems great! One issue though is that even after completely deleting all a drawing, the drawing remains selectable as an invisible group. Is there any way to clear the "erased" drawings? I need to export the JSON and I'm worried it becomes infinitely large.

ShaMan123:
@tanzyy96 https://github.com/fabricjs/fabric.js/pull/6994#issuecomment-829075360_

brunobrante:
How can I remove erased objects in canvas, only when I have fully erased the object?

ShaMan123:
https://github.com/fabricjs/fabric.js/discussions/8586

PR #7007, opened 2021-04-10
Fix uniform scaling bug 

vvn050:
With current implementation , uniform scaling works only in case shift key was pressed.


asturur:
The current implementation is supposed to work in this way:
- you select uniformScaling as true, and you will get free form scaling if you press shift
- you select uniformScaling as false, you get always freeform, but when you press shift you will get proportional scaling.
This is how we want it to work because it gives you 4 combinations and should cover every case.

What are you experiencing that you think is wrong?

vvn050:
From documentation : http://fabricjs.com/docs/fabric.Canvas.html
uniformScaling :Boolean
When true, objects can be transformed by one side (unproportionally) when dragged on the corners that normally would not do that.
According to it , it should behave exactly the opposite.

asturur:
ok that is a mistake by my side not updating the description.
in the v4 breaking changes we clarified:
http://fabricjs.com/v4-breaking-changes
image

And i remember waiting a breaking version to change it, because the old behaviour/code did not allow for the 4 combinations:
always on
always off
ON with toggle key
OFF with toggle key

vvn050:
Cool, thanks for the clarification and fast response.

PR #7030, opened 2021-04-21
fix(svg parser): Ensure that applyViewboxTransform returns an object and not undefined/null

rockerBOO:
In fabric.parseSVGDocument, the options object was being set as `undefined`/`null` and causing `options.crossOrigin` to be invalid. 

```
    TypeError: Cannot set property 'crossOrigin' of undefined

      19 | ): Promise<EcogardenFabricObject> =>
      20 |  new Promise((resolve) => {
    > 21 |      fabric.loadSVGFromString(
         |             ^
```

```
      at Object.<anonymous>.fabric.parseSVGDocument (node_modules/fabric/dist/fabric.js:4614:25)
      at Object.loadSVGFromString (node_modules/fabric/dist/fabric.js:4992:14)
      at src/js/lib/fabric.ts:21:10
```

```
    var svgUid =  fabric.Object.__uid++, i, len,
        options = applyViewboxTransform(doc),
        descendants = fabric.util.toArray(doc.getElementsByTagName('*'));
    options.crossOrigin = parsingOptions && parsingOptions.crossOrigin;
```

This pull request resolves `applyViewboxTransform` to always return an object.

Using fabricjs 4.4.0 and this was not erroring previously in 3.6.


asturur:
@rockerBOO  can you remove the build dist file with the master one? so i can merge.
(i usually do `git checkout master dist` before committing)

rockerBOO:
Reverted dist/fabric.js to master. Thank you!

Issue #7042, opened 2021-04-26
Path incorrect bbox values


ShaMan123:
## Version
Latest

## Test Case
https://codesandbox.io/s/fervent-sunset-5d2x8?file=/src/App.tsx

## Information about environment
Chrome

## Steps to reproduce /  Actual Behavior
```js
var path;
path.strokeWidth = 20 // manipulate this
var coords = path.calcCoords();
var badSize = {
  width: path.width,
  height: path.height
};
var size = {
  width: coords.br.x - coords.tl.x,
  height: coords.br.y - coords.tl.y
};
console.log(badSize.width === size.width, badSize.height === size.height);

// output false, false
```

## Expected Behavior
`width` and `height` should reflect the bounding box regardless of `strokeWidth` property

ShaMan123:
#6727

asturur:
Is unclear what are you trying to do in the codepen seems like you want to add a clipPath to the drawing, but you can't find out the correct size of the path.

Try to look for the patth bounding box using
http://fabricjs.com/docs/fabric.Object.html#getBoundingRect

ShaMan123:
Sorry for not being clear enough.
If filed the bug because `width` and `height` don't reflect the path's actuals bounding box.
**For the EXAMPLE:**
Once a path is created and added to the canvas a rect clip path of the size of the path will be added to the path.
The desired result would be that the clip path will not clip any part of the path because it's size is the same as the path. That means the bounding box was calculated correctly.
Click the different buttons to see how results vary between calculations showing that `width | height` properties are incorrect for the path.
I have patched `fabric.Path` to fix this. You can try it by clicking `Patch Path class`.

ShaMan123:
> Is unclear what are you trying to do in the codepen seems like you want to add a clipPath to the drawing, but you can't find out the correct size of the path.
> 
> Try to look for the patth bounding box using
> http://fabricjs.com/docs/fabric.Object.html#getBoundingRect

This works. But again the width and the height properties of the path are different from the result of `getBoundingRect`

ShaMan123:
> > Is unclear what are you trying to do in the codepen seems like you want to add a clipPath to the drawing, but you can't find out the correct size of the path.
> > Try to look for the patth bounding box using
> > http://fabricjs.com/docs/fabric.Object.html#getBoundingRect
> 
> This works, thanks. However, the width and the height properties of the path are different from the result of `getBoundingRect`


I've updated the example

[...]

PR #7043, opened 2021-04-26
fix Path._calcDimensions (account for strokeWidth)

ShaMan123:
Motivation #7042 

It seems that the dimensions do not account for `strokeWidth` though it is expected.

Test: https://codesandbox.io/s/fervent-sunset-5d2x8?file=/src/patchPath.ts:100-440

The test reveals another bug that seems to originate from `PencilBrush`, try changing the brush `strokeWidth`

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -322,11 +322,12 @@
           deltaX = maxX - minX,
           deltaY = maxY - minY;
 
+      //  account for strokeWidth
       return {
-        left: minX,
-        top: minY,
-        width: deltaX,
-        height: deltaY
+        left: minX - this.strokeWidth / 2,
+        top: minY - this.strokeWidth / 2,
+        width: deltaX + this.strokeWidth,
+        height: deltaY + this.strokeWidth
       };
     }
   });


asturur:
@ShaMan123 i think this is wrong. The path width / height DOES NOT take in account stroke width. Is just the width/height of the area described buy the line ( dimensionless ) that describe the path.
This is the same for all the shapes. The circles is twice the radius and not radius + strokewdith

What does it make you think we need to change this?
We have specific methods to understand how big is an object, in the 2 kind of different coordinates.

ShaMan123:
I simply think it is unclear.
I would expect `width` and `height` to be equal to the bbox.

ShaMan123:
What is the behavior is html?

asturur:
The behaviour in HTML is probably different but is not relevant.
FabricJS has its own code in order to be able to position and draw things, and this is the contract.
Width and height define the shape without a stroke.
Everything else break otherwise.

Adding strokeWidth to the width/height is not enough when there are very acute angle for example.

To get the current size of the shape we have:
Object._getNonTransformedDimensions and Object._getTransformedDimensions that try to take in account at best other properties like strokeUniform.

ShaMan123:
OK I understand, thanks.
I do think this should be documented in [fabric gotchas](http://fabricjs.com/fabric-gotchas)
closing

ShaMan123:
Continuing this on #7042

asturur:
Feel free to open a PR for fabric gotchas here:
https://github.com/fabricjs/fabricjs.com/blob/gh-pages/posts/_posts/2019-08-26-fabric-gotchas.html

PR #7044, opened 2021-04-27
Fix svg parsing in case it uses empty use tag or use with image href

andrejslogins:
Fix cases where parsing string to svg was crashing.

Added the two test cases where this occurs.
1st case (empty <use>) occurs because xLinkAttribute is null (line 476)
2nd case (image href) occurs because element.parentNode is null

Didn't bump dist files as i'm not aware if this should be done in scope of PRs or not

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -471,12 +471,23 @@
     var nodelist = _getMultipleNodes(doc, ['use', 'svg:use']), i = 0;
     while (nodelist.length && i < nodelist.length) {
       var el = nodelist[i],
-          xlink = (el.getAttribute('xlink:href') || el.getAttribute('href')).substr(1),
+          xlinkAttribute = el.getAttribute('xlink:href') || el.getAttribute('href');
+
+      if (xlinkAttribute === null) {
+        return;
+      }
+
+      var xlink = xlinkAttribute.substr(1),
           x = el.getAttribute('x') || 0,
           y = el.getAttribute('y') || 0,
           el2 = elementById(doc, xlink).cloneNode(true),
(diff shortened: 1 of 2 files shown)


src/parser.js
```diff
@@ -471,12 +471,23 @@
     var nodelist = _getMultipleNodes(doc, ['use', 'svg:use']), i = 0;
     while (nodelist.length && i < nodelist.length) {
       var el = nodelist[i],
-          xlink = (el.getAttribute('xlink:href') || el.getAttribute('href')).substr(1),
+          xlinkAttribute = el.getAttribute('xlink:href') || el.getAttribute('href');
+
+      if (xlinkAttribute === null) {
+        return;
```
asturur, line 477: what is this function returning on the others return early case? are we safe returning undefined?

andrejslogins: no explicit returns are there in the following code, so it's safe

andrejslogins:
For checks:
- Pr quality check / coverage (pull_request) :`Error: Resource not accessible by integration` , looks like this is unrelated to the PR itself

src/parser.js
andrejslogins: Only usage here:

andrejslogins:
@asturur All good from your side?

asturur:
Sorry i have been away

asturur:
So to clarify my question, the first time i looked at the PR i must have opened to much diff lines or to few, and i was seeing some other returns, but i see i was watching something else.

Pr looks good, i m merging it. Next time try to avoid unnecessary format changes, however they may look better, they create non necessary diff changes to look at.

Unless is a specific linting PR.

andrejslogins:
sure, will keep in mind. thanks

PR #7082, opened 2021-05-27
fix(fabric.Path): setting `path` property after initialization

ShaMan123:
### Motivation
Setting `path` property after `fabric.Path` initialization doesn't work out of the box and needs too much handling and internal knowledge of this repository's logic to make it successful.
**This is bad for DEV UX.**

### Fix
The fix re-uses the same logic to reset the path correctly.

```js
const path = new fabric.Path(...);
// do stuff
path.set('path', .......);
```

### Test it
https://codesandbox.io/s/reverent-ives-464sc?file=/src/App.tsx

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -374,7 +374,7 @@
   fabric.Path.convertPointsToPath = function (points) {
     var start = points[0].x;
     var minMax = points.reduce(
-      (prev, curr) => {
+      function (prev, curr) {
         return {
           min: Math.min(curr.x, prev.min),
           max: Math.max(curr.x, prev.max)
@@ -384,7 +384,7 @@
     );
     var width = minMax.max - minMax.min;
     return fabric.PencilBrush.prototype.convertPointsToSVGPath
-      .call({ width }, points)
+      .call({ width: width }, points)
       .join(' ');
   };
 
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -50,10 +50,15 @@
     initialize: function (path, options) {
       options = options || {};
       this.callSuper('initialize', options);
-      this.setPath(path || [], options);
+      this._setPath(path || [], options);
     },
 
-    setPath: function (path, options) {
+    /**
+    * @private
+    * @param {Array|String} path Path data (sequence of coordinates and corresponding "command" tokens)
+    * @param {Object} [options] Options object
+    */
+    _setPath: function (path, options) {
       var fromArray = _toString.call(path) === '[object Array]';
 
       this.path = fromArray
@@ -69,11 +74,14 @@
       fabric.Polyline.prototype._setPositionDimensions.call(this, options || {});
     },
 
+    /**
+     * @private
+     */
     _set: function (key, value) {
       if (key === 'path') {
         var left = this.left,
             top = this.top;
-        this.setPath(value);
+        this._setPath(value);
         this.left = left;
         this.top = top;
         this.setCoords();
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -53,7 +53,7 @@
       this.setPath(path || [], options);
     },
 
-    setPath: function (path, options = {}) {
+    setPath: function (path, options) {
       var fromArray = _toString.call(path) === '[object Array]';
 
       this.path = fromArray
@@ -66,7 +66,7 @@
       if (!this.path) {
         return;
       }
-      fabric.Polyline.prototype._setPositionDimensions.call(this, options);
+      fabric.Polyline.prototype._setPositionDimensions.call(this, options || {})
     },
 
     _set: function (key, value) {
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -66,18 +66,19 @@
       if (!this.path) {
         return;
       }
-      fabric.Polyline.prototype._setPositionDimensions.call(this, options || {})
+      fabric.Polyline.prototype._setPositionDimensions.call(this, options || {});
     },
 
     _set: function (key, value) {
       if (key === 'path') {
         var left = this.left,
-          top = this.top;
+            top = this.top;
         this.setPath(value);
         this.left = left;
         this.top = top;
         this.setCoords();
-      } else {
+      }
+      else {
         this.callSuper('_set', key, value);
       }
     },
@@ -368,7 +369,7 @@
   /**
    * @static
    * @memberOf fabric.Path
-   * @param {fabric.Point[]} points 
+   * @param {fabric.Point[]} points
    * @returns {string} svg path
    */
   fabric.Path.convertPointsToPath = function (points) {
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -79,7 +79,11 @@
      */
     _set: function (key, value) {
       if (key === 'path') {
+        var left = this.left,
+            top = this.top;
         this._setPath(value);
+        this.left = left;
+        this.top = top;
         this.setCoords();
       }
       else {
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -6,6 +6,7 @@
       min = fabric.util.array.min,
       max = fabric.util.array.max,
       extend = fabric.util.object.extend,
+      clone = fabric.util.object.clone,
       _toString = Object.prototype.toString,
       toFixed = fabric.util.toFixed;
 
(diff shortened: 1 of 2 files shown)


ShaMan123:
Another thought for **DEV UX** bc5499a5

ShaMan123:
Build + test pass

ShaMan123:
I think `fabric.PencilBrush.prototype.convertPointsToSVGPath` should reside under `Path` or `util`.
What do you think?

src/shapes/path.class.js
```diff
@@ -66,7 +71,24 @@
...
+     * @private
+     */
+    _set: function (key, value) {
+      if (key === 'path') {
+        var left = this.left,
+            top = this.top;
+        this._setPath(value);
+        this.left = left;
+        this.top = top;
+        this.setCoords();
```
asturur, line 87: this is a bit of a stretch for a setter and is also a breaking change because is radically changing how the app behaves for who uses set today
This is not wrong but should be an utility on its own.

Something like `path.updatePath` that can do exactly

```
        var left = this.left,
            top = this.top;
        this._setPath(value);
        this.left = left;
        this.top = top;
        this.setCoords();
```

ShaMan123: You are right. It was breaking.
Now it isn't a2373e2.
So to preserve the position one can easily do the following:
```js
        var left = path.left, top = path.top;
        path.set('path', value);
        path.left = left;
        path.top = top;
```

asturur:
Sorry for not being responsive, i don't have much time lately.

ShaMan123:
not stale

asturur:
So ok there are good things there, mostly all of them.
There is some consistency that we are breaking and i would like not to.
Also some plan i was thinking of was to completely trash away the polygon/polyline code and leave them to Path entirely. So PathFromPoint is good, but it can't rely on the brush to happen and also can't be the default Quadratic curve version.
The whole drawing code is optional and can be excluded and this would create a dependency between the 2 modules.

With those changes the path is going to call setCoords at initialization while no other class does that.
I would prefer to avoid it.
That would call setCoords automatically also when someone use set.

For me to merge this pr we need to apply the following changes
I'll describe them briefly, and then focus on the first one.

- make `convertPointsToSVGPath` an utility in the file util/path.js and have the brush utilize it.
- have a Pr that adds path from points using that new utility
- have a pr that will add setPath that does what you did without setCoords
- have a pr that can add an higher level method that can take care of setCoords and position

If we do small PRs focused on a single thing, i don't need much time to understand what is changing, think of side effects, and review and explain what i would like to change and why.

So if you agree you can open a new PR with the first 2 points that should be relatively simple and really useful.
Now `convertPointsToSVGPath` returns a non useful array of strings.
If you look at how we handle paths, a SVG path looks like this:

```js
const path = 'M 3 4 Q 3 5 2 1 4 0  Q 9 12 2 1 4 0'
```
The path returned from `convertPointsToSVGPath` is:
```js
const path = ['M', ' ', 3, ' ', 4, 'Q', ' ', 3, ....] // and so on
```
Is basically number and stings and spaces in the array.

The path format for the Path class is:

```js
const path = [
['M', 3, 4],
['Q', 3, 5, 2, 1, 4, 0],
['Q', 9, 12, 2, 1, 4, 0],
]
```

Fabric.Path constructor expect either the SVG string or the last one.
The middle one we can forget about it, is non useful, is bad to write bad to read.

Now `convertPointsToSVGPath` wants to handle a special case in which the path is dimensionless and so it uses a small fraction of the brush width to change the path points (this.width/1000). This correction factor can be an argument together with points, and instead of pushing spaces and number converted in strings

```diff
    convertPointsToSVGPath: function(points, correction) {
      var path = [], i,
          p1 = new fabric.Point(points[0].x, points[0].y),
          p2 = new fabric.Point(points[1].x, points[1].y),
          len = points.length, multSignX = 1, multSignY = 0, manyPoints = len > 2;

      if (manyPoints) {
        multSignX = points[2].x < p2.x ? -1 : points[2].x === p2.x ? 0 : 1;
        multSignY = points[2].y < p2.y ? -1 : points[2].y === p2.y ? 0 : 1;
      }
        // and below changes similar to this, no spaces but nested arrays
-       path.push('M ', p1.x - multSignX * correction, ' ', p1.y - multSignY * correction, ' ');
+       path.push(['M', p1.x - multSignX * correction, p1.y - multSignY * correction]);
      for (i = 1; i < len; i++) {
        if (!p1.eq(p2)) {
          var midPoint = p1.midPointFrom(p2);
          // p1 is our bezier control point
          // midpoint is our endpoint
          // start point is p(i-1) value.
          path.push('Q ', p1.x, ' ', p1.y, ' ', midPoint.x, ' ', midPoint.y, ' ');
        }
        p1 = points[i];
        if ((i + 1) < points.length) {
          p2 = points[i + 1];
        }
      }
      if (manyPoints) {
        multSignX = p1.x > points[i - 2].x ? 1 : p1.x === points[i - 2].x ? 0 : -1;
        multSignY = p1.y > points[i - 2].y ? 1 : p1.y === points[i - 2].y ? 0 : -1;
      }
      path.push('L ', p1.x + multSignX * width, ' ', p1.y + multSignY * width);
      return path;
    },

/* final output:
const path = [
['M', 3, 4],
['Q', 3, 5, 2, 1, 4, 0],
['Q', 9, 12, 2, 1, 4, 0],
]
*/
```

So the output of this function would be compatible with fabricPath with no other changes, while the brush would use something like

```
finalPath = stringPath = path.map((command) => command.join(' ')).join(' ');
```

or eventually it wouldn't at all since  what he wants to do is create a fabric.Path and the output format is good enough.

Then there is another small detail.
A brush creates ( because we built like this ) a smooth path using Q commands.
But a series of point could totally also just generate a polygon, there isn't a solution for this yet, but something to keep in mind, maybe instead of calling the `Path.PathFromPoints` we can call it `Path.PathFromBrush` or find a similar names, and leaving fromPoints for the Polygon.


I thank you for trying to improve the library, and in order for your efforts to be efficient we need to find a common ground where collaboration is useful.

So in general:
- small prs with one small change at time ( and tests )
- speak openly about what you find is missing and what you would like to fix before diving deep down in a pr
- or open a discussion like this: https://github.com/fabricjs/fabric.js/discussions/7128 in the IDEAS section and ping both me and @melchiar

ShaMan123:
Agreed! I appreciate you being open minded and open from collaboration and for caring about my time, effort and work.
First of all I'd like to know how you develop the library. I couldn't understand how to set up a dev environment. Up to now I worked on a project of mine but I want an order in my environment.
What you mention seems perfectly fine. It is sometimes hard to PR small patches because the context of the major fix is important for progress. However I do understand the necessity. I want to understand your view on this, is #7100 a big/small/normal PR? It is a good ref because many of the fixes are necessary for the context of the major fix but I did push a few commits out of scope. I'll kind this in mind.
- If I remember correctly the call to `setCoords` was done because touch area was wrong. What if we call `setCoords` from the `added` event or `_onObjectAdded`?
- regarding `XXXFromPoints` I have 2 ideas:
  1. rename to `CurveFromPoints`. `PathFromBrush` it's not clear what it does.
  2. expose a function: 
  ```ts
   type PathBuilder = 'polygon' | 'polyline' | 'curve' | 'smooth-path' ..... etc.
   Path.FromPoints(type: PathBuilder, points: Point[]): Path
   ```

asturur:
#7100 is ok as PR, some pr can't be broken in smaller pieces. The eraser is kind on its own case, since i don't know it enough and i know that being confined in a module i don't have to worry too much of side effects.

setCoords gets called when we add objects on the canvas, when a user initiated transform end, or when the developer call it manually. I think since we introduced the control api it also gets called during a trasnform in order to render the controls in the right place, but in general we don't want it to be called without knowing, as we don't things to be called  too often.

Regarding the path building, let's do what we need now, so starting to get in the direction of having the path adjustable at runtime, and building the extra parts when needed.

ShaMan123:
What about starting the dev environment? How do I do that? Is there a gist for contributing?

melchiar:
This should help setup your dev environment

https://github.com/fabricjs/fabric.js/pull/5558#issuecomment-471491694

ShaMan123:
> This should help setup your dev environment
> 
> [#5558 (comment)](https://github.com/fabricjs/fabric.js/pull/5558#issuecomment-471491694)

Thanks, How about running fabric on web, actually interacting with it?

ShaMan123:
Closed via #7140 #7141

Issue #7086, opened 2021-05-27
Group selector uses canvas coordinates which can correspond to different fabric coordinate when panning/zooming while selecting


SLKnutson:
## Version
4.5.0, don't think this has changed recently

## Test Case
http://fabricjs.com/fabric-intro-part-5

## Information about environment
Any browser

## Steps to reproduce
1. Go to the first fabric canvas on http://fabricjs.com/fabric-intro-part-5
1. Use a mouse with a scroll wheel
1. Hold down the left mouse button
1. Move the mouse slightly to create a blue selection area
1. While keeping the mouse button held down, zoom in or out using the mouse wheel

## Expected Behavior
The grouping should stay anchored to the original fabric coordinate that it was created in

## Actual Behavior
The grouping is anchored to the canvas offset that it was originally created in

## Reasoning
Not super obvious with the zooming in/out, but this is really evident when implementing autopanning when the cursor moves to the edge of the canvas. Theoretically someone could want it to behave the way that it currently does, but I can't imagine why.

## Thanks!

PR #7088, opened 2021-05-28
Record group selection in transformed coordinates to support panning/zooming while group selecting

SLKnutson:
Fixes this issue: https://github.com/fabricjs/fabric.js/issues/7086

There is still a slight issue because the group select controls don't get immediately rerendered when transforming, but I didn't want to shove too much into a single PR. I thought about making a feature toggle on fabric.Canvas, but I couldn't think of any use cases for the way it behaved before. I didn't see any existing methods of testing the contents of _drawSelection(). I'm glad to make changes as needed.

I changed _drawSelection to accommodate varying selectionLineWidth, since the previous implementation was essentially hardcoded to 1, and would draw outside the box for values > 1, now it should always draw within the box.

Thank you!

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -714,21 +714,8 @@
       maxX -= strokeOffset;
       maxY -= strokeOffset;
       // selection border
-      if (this.selectionDashArray.length > 1 && !supportLineDash) {
-        ctx.beginPath();
-
-        fabric.util.drawDashedLine(ctx, minX, minY, maxX, minY, this.selectionDashArray);
-        fabric.util.drawDashedLine(ctx, minX, maxY, maxX, maxY, this.selectionDashArray);
-        fabric.util.drawDashedLine(ctx, minX, minY, minX, maxY, this.selectionDashArray);
-        fabric.util.drawDashedLine(ctx, maxX, minY, maxX, maxY, this.selectionDashArray);
-
-        ctx.closePath();
-        ctx.stroke();
-      }
-      else {
-        fabric.Object.prototype._setLineDash.call(this, ctx, this.selectionDashArray);
-        ctx.strokeRect(minX, minY, maxX - minX, maxY - minY);
-      }
+      fabric.Object.prototype._setLineDash.call(this, ctx, this.selectionDashArray);
+      ctx.strokeRect(minX, minY, maxX - minX, maxY - minY);
     },
 
     /**
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -2,10 +2,8 @@
 
   var getPointer = fabric.util.getPointer,
       degreesToRadians = fabric.util.degreesToRadians,
-      abs = Math.abs,
       supportLineDash = fabric.StaticCanvas.supports('setLineDash'),
-      isTouchEvent = fabric.util.isTouchEvent,
-      STROKE_OFFSET = 0.5;
+      isTouchEvent = fabric.util.isTouchEvent;
 
   /**
    * Canvas class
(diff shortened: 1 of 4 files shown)


src/mixins/canvas_events.mixin.js
```diff
@@ -712,8 +712,8 @@
       if (this.selection && (!target ||
         (!target.selectable && !target.isEditing && target !== this._activeObject))) {
         this._groupSelector = {
-          ex: pointer.x,
-          ey: pointer.y,
+          ex: this._absolutePointer.x,
+          ey: this._absolutePointer.y,
```
asturur, line 716: i dream a day in which the local pointer goes away entirely and fabric uses only the absolute values, and we remove most of the conversions.

src/canvas.class.js
```diff
@@ -712,30 +709,25 @@
...
         ctx.beginPath();
 
-        fabric.util.drawDashedLine(ctx, px, py, px + aleft, py, this.selectionDashArray);
-        fabric.util.drawDashedLine(ctx, px, py + atop - 1, px + aleft, py + atop - 1, this.selectionDashArray);
-        fabric.util.drawDashedLine(ctx, px, py, px, py + atop, this.selectionDashArray);
-        fabric.util.drawDashedLine(ctx, px + aleft - 1, py, px + aleft - 1, py + atop, this.selectionDashArray);
+        fabric.util.drawDashedLine(ctx, minX, minY, maxX, minY, this.selectionDashArray);
+        fabric.util.drawDashedLine(ctx, minX, maxY, maxX, maxY, this.selectionDashArray);
+        fabric.util.drawDashedLine(ctx, minX, minY, minX, maxY, this.selectionDashArray);
+        fabric.util.drawDashedLine(ctx, maxX, minY, maxX, maxY, this.selectionDashArray);
```
asturur, line 723: would you check if IE11 supports dashed stroke? if it does, we can finally get rid of  the block (!supportedLineDash)

asturur:
I think this PR is fine.
Not sure we can write an automated test for this, but ideally the only task of _drawSelection is to draw that rect. If while not zooming it works as before we are ok.
Can you capture in some way the issue you are referring to?

src/canvas.class.js
SLKnutson: Here is the support link. Looks like IE 11 supports but I'm not sure if the support is good enough on other browsers like chrome? https://caniuse.com/mdn-api_canvasrenderingcontext2d_setlinedash

asturur: yes setLineDash is an old feature that was missing on ie10 only.
Since we do not support ie10 anymore, this is safe to go.

src/canvas.class.js
```diff
@@ -2,10 +2,8 @@
 
   var getPointer = fabric.util.getPointer,
       degreesToRadians = fabric.util.degreesToRadians,
-      abs = Math.abs,
       supportLineDash = fabric.StaticCanvas.supports('setLineDash'),
-      isTouchEvent = fabric.util.isTouchEvent,
-      STROKE_OFFSET = 0.5;
```
asturur, line 8: i think STROKE_OFFSET is still needed. it is used because mouse events are always round numbers, while we want to draw single pixels line at 0.5.

asturur:
@SLKnutson any chance we can move this forward.
If you are busy is ok, just let me know if you know you won't have time, i ll do it ( if i'm enabled to push on your branch )

src/canvas.class.js
SLKnutson: If I remember correctly, I think this actually to account for canvas drawing strokes down the center of the line. The hardcoded stroke offset of 0.5 works well for the default selectionLineWidth of 1, but doesn't look right for larger values. I think what I did accomplishes the same thing but allows for selectionLineWidth != 1

SLKnutson:
@asturur You can make changes as needed. I will eventually have time for this, but I'm not sure how far out in the future that would be. Thanks!

PR #7100, opened 2021-06-04
feat(eraser_brush): erasable=`deep`

ShaMan123:
### Changes
This PR introduces fine grained control over a group's `erasable` property.
Added erasable property value: `deep`, `erasable` now accepts `boolean | 'deep'`.

The `deep` option will erase nested objects in a group if they are erasable, 
`true` will erase the entire group (once the group changes the eraser is propagated to its children for proper functionality), 
`false` will leave all objects including the group untouched.

### Fixes
1. Erasing huge groups like the result of `SprayBrush` or `CircleBrush`.
2. Eraser behavior in a group after the group has changed (object added/removed). Because the group's dimensions change after add/remove the eraser must be handled or else it will be transformed incorrectly according to the group's previous bounding box. To handle this, before the change is applied to the group the eraser is propagated to the relevant objects and is removed from the group. This happens when the group's `erasable` property is set to `true`.
3. Object resizing is now supported out of the box, including canvas resizing.

### Test It
https://codesandbox.io/s/exciting-architecture-bcgej?file=/src/App.tsx

_Originally posted by @matsu-motsu in https://github.com/fabricjs/fabric.js/issues/6994#issuecomment-854459262_

#### Behavior before this PR
Once erasing started `EraserBrush` traversed through all objects, nested or not to render the layers for erasing.
Once erasing had completed (`mouseup`) all objects were iterated through, including objects that were nested in groups. If an object was erasable a path would have been added to it's eraser. 
`EraserBrush` didn't add erasing paths to groups to enable a group to have both erasable and non erasable objects.

In the case of `SprayBrush`, what happened was:
1. Rendering the layers for erasing was extremely heavy because it iterated through all the dots, unnecessarily.
2. When erasing had completed an erasing path was created for each dot of the spray, resulting in a huge and extremely heavy operation. 

This behavior is exactly what the `deep` option does and is **undesirable** as default behavior as shown above. 
So now, with this PR, a single path will be added **only** to the group's eraser as long as `erasable=true` and that's it.

---
This tests object and canvas resizing with eraser:
https://codesandbox.io/s/romantic-williams-fwbz9?file=/src/App.tsx
Erase the blue overlay and then change the canvas size by dragging the window borders. Notice that the eraser is now sized automatically; the red rect changes size while it's eraser stays in place.

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/7100#issuecomment-864380592_

---
> I have come across an edge case.
> I am interested in your feedback.
> Say there's a group with a clip path. When erasing the group the eraser path will be clipped properly.
> Once the eraser paths propagate to child objects the group clip path will not. This will make nested erasing with nested clipping hard to handle from the event `path:created` |`before:path:created`.
> How should we address this?
> It's easy enough to check if the parent group of an object has a clip path and if it does to apply it to the object. But what object nested objects in nested groups with clip paths?

**SOLVED** 82a3d7d -> e2c6774
https://codesandbox.io/s/angry-browser-49sc3?file=/src/EraserBrush.js

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/7100#issuecomment-857810155_


ShaMan123:
Ready to merge

ShaMan123:
DONE

src/mixins/eraser_brush.mixin.js
```diff
@@ -124,7 +124,12 @@
   fabric.util.object.extend(fabric.Object.prototype, {
     /**
      * Indicates whether this object can be erased by {@link fabric.EraserBrush}
-     * @type boolean
+     * The `deep` option introduces fine grained control over a group's `erasable` property.
+     * When set to `deep` the eraser will erase nested objects if they are erasable, leaving the group and the other objects untouched.
+     * When set to `true` the eraser will erase the entire group. Once the group changes the eraser is propagated to its children for proper functionality.
+     * When set to `false` the eraser will leave all objects including the group untouched.
+     * @tutorial {@link http://fabricjs.com/erasing#erasable_property}
+     * @type boolean | 'deep'
```
asturur, line 132: boolean or string isn't my favorite.
With people also asking for typescript we want to keep types as much as standard as possible.
How harder it is to have this with its own parameter?

Also why the property can't stack normally?
How does it work if you have nested groups?
Seems that this deep is more a `childrenOnly` or a `deferToChildren`?

ShaMan123: `childrenOnly` sounds fitting.
Passing another option sounds bad to me. Makes something that should be simple error prone.
What about `enum` or `erasable | children-only | none`?

ShaMan123: >Also why the property can't stack normally?
>How does it work if you have nested groups?

What do mean by stack normally?
It traverses through all nested groups if they are `erasable=deep`, see [code](https://github.com/fabricjs/fabric.js/pull/7100/files#diff-d9e0aff690fbd7c86841642ab3ba18ce2cb07c9b4e87ca83c5e334f25572b9b0R659)

ShaMan123: BTW Typescript doesn't complain in such a case (`boolean` | `string`) . `enum` it standard in TS.

ShaMan123: What's your opinion?

ShaMan123:
I have come across an edge case.
I am interested in your feedback.
Say there's a group with a clip path. When erasing the group the eraser path will be clipped properly. 
Once the eraser paths propagate to child objects the group clip path will not. This will make nested erasing with clipping hard to handle.
How should we address this?
It's easy enough to check if the parent group of an object has a clip path and if it does to apply it to the object. But what object nested objects in nested groups with clip paths?

ShaMan123:
> I have come across an edge case.
> I am interested in your feedback.
> Say there's a group with a clip path. When erasing the group the eraser path will be clipped properly.
> Once the eraser paths propagate to child objects the group clip path will not. This will make nested erasing with nested clipping hard to handle from the event `path:created` |`before:path:created`.
> How should we address this?
> It's easy enough to check if the parent group of an object has a clip path and if it does to apply it to the object. But what object nested objects in nested groups with clip paths?

**SOLVED** 82a3d7d
https://codesandbox.io/s/angry-browser-49sc3?file=/src/EraserBrush.js

ShaMan123:
@asturur ?

ShaMan123:
This tests 7c4e732
Erase the blue overlay and then change the canvas size by dragging the window borders. Notice that the eraser is now sized automatically and the red rect changes size while it's eraser stays in place.
https://codesandbox.io/s/romantic-williams-fwbz9?file=/src/App.tsx

ShaMan123:
Update CHANGELOG
I want to move forward with this PR

src/mixins/eraser_brush.mixin.js
ShaMan123: @asturur ?

ShaMan123:
@asturur This PR fixes a major bug and adds a major feature. It is sad seeing it going stale.
Could you reply on the open matters so we can agree on how to finalize it?

asturur (approved):

asturur:
This is fine, i have no time arguing over the deep/erasable naming.

ShaMan123:
Thanks!
I didn't mean to argue. Just want this to be a good feature that both you and I are happy about.

asturur:
Argue is the wrong word, I apologize. I just meant is not a detail to which i can dedicate time nor stop you with.

ShaMan123:
You are right
Thanks again

בתאריך יום ג׳, 27 ביולי 2021, 19:10, מאת Andrea Bogazzi ‏<
***@***.***>:

PR #7120, opened 2021-06-14
ignored justify when last line in textbox

khajaamin:


SLKnutson:
@khajaamin 
Not sure what you are going for here, but you might be looking for the existing text align option of 'justify-left'

asturur:
i tried to ask him what he was trying to achieve or help with, but i never got an answer

PR #7121, opened 2021-06-14
Justify ignore lastline

khajaamin:


asturur:
@khajaamin  is unclear what are you trying to do and why.
Care to explain?

PR #7129, opened 2021-06-17
ci(): fix visual test running in browsers

asturur:
browser tests were not running in the ci, missing a double dash to concat the options with the npm run command


asturur:
Font testing requires some kind of help in general, but this is better than nothing.
Anyway for testing Arial > Helvetica.
Are more similar across OS/Env.

PR #7140, opened 2021-06-21
Smooth Path utils

ShaMan123:
This PR replaces most of #7082
It focuses on smooth path utils.

Changes:
1. Extracted `convertPointsToSVGPath` to path util and renamed to `getSmoothPathFromPoints`
2. Added default correction value to `getSmoothPathFromPoints`
2. Adapted pencil brush and eraser brush
3. Adjusted tests

### DIFF
--- a/src/brushes/pencil_brush.class.js
+++ b/src/brushes/pencil_brush.class.js
@@ -179,43 +179,26 @@
     /**
      * Converts points to SVG path
      * @param {Array} points Array of points
-     * @return {String} SVG path
+     * @return {(string|number)[][]} SVG path commands
      */
-    convertPointsToSVGPath: function(points) {
-      var path = [], i, width = this.width / 1000,
-          p1 = new fabric.Point(points[0].x, points[0].y),
-          p2 = new fabric.Point(points[1].x, points[1].y),
-          len = points.length, multSignX = 1, multSignY = 0, manyPoints = len > 2;
+    convertPointsToSVGPath: function (points) {
+      var correction = this.width / 1000;
+      return fabric.util.getSmoothPathFromPoints(points, correction);
+    },
 
-      if (manyPoints) {
-        multSignX = points[2].x < p2.x ? -1 : points[2].x === p2.x ? 0 : 1;
-        multSignY = points[2].y < p2.y ? -1 : points[2].y === p2.y ? 0 : 1;
-      }
-      path.push('M ', p1.x - multSignX * width, ' ', p1.y - multSignY * width, ' ');
-      for (i = 1; i < len; i++) {
-        if (!p1.eq(p2)) {
-          var midPoint = p1.midPointFrom(p2);
-          // p1 is our bezier control point
-          // midpoint is our endpoint
-          // start point is p(i-1) value.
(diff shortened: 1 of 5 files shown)


src/brushes/pencil_brush.class.js
```diff
@@ -272,8 +255,8 @@
       if (this.decimate) {
         this._points = this.decimatePoints(this._points, this.decimate);
       }
-      var pathData = this.convertPointsToSVGPath(this._points).join('');
-      if (pathData === 'M 0 0 Q 0 0 0 0 L 0 0') {
+      var pathData = this.convertPointsToSVGPath(this._points);
+      if (this._isEmptySVGPath(pathData)) {
```
ShaMan123, line 259: Maybe there's a better way to determine if the path is empty.
What about checking points?

ShaMan123:
Should I add the path builder to this PR?
It is now very simple and minor because of the refactoring + tests.
```ts
fabric.Path.pointsToSmoothPath = function (points, options) {
    return new fabric.Path(fabric.util.getSmoothPathFromPoints(points), options);
  };
```

asturur:
here i am late again!

src/brushes/pencil_brush.class.js
asturur: Probably there is, but for now we don't care. we know this is a bit bad because we need to join it all of it to check if is empty. If is a 20 meter long path with 10.000 points is time wasted. But for now we don't care. for this PR is fine as it is.

src/util/path.js
```diff
@@ -729,6 +741,62 @@
...
+          return {
+            min: Math.min(curr.x, prev.min),
+            max: Math.max(curr.x, prev.max)
+          };
+        },
+        { min: start, max: start }
+      );
+      var width = minMax.max - minMax.min;
+      correction = width / 1000;
+    }
```
asturur, line 771: if correction is undefined, we set it to 0. we don't want to correct every path. is just a edge case of the brush.
please remove this if, let's just add a `correction = correction || 0;` and this PR is good to go

asturur (changes requested):
If we remove that extra if cycle we are good to go with this pr!

src/util/path.js
ShaMan123: 500abd6, af9045e
e740a16 extracts the logic to a separate util. What do you think?

src/util/path.js
```diff
@@ -729,6 +741,70 @@
...
+          min: Math.min(curr.x, prev.min),
+          max: Math.max(curr.x, prev.max)
+        };
+      },
+      { min: start, max: start }
+    );
+    var width = minMax.max - minMax.min;
+    var correction = width / 1000;
+    return getSmoothPathFromPoints(points, correction);
+  }
```
asturur, line 806: we don't need this automatic correction, at all.
this width can't be inferred from the point, is just the brush chosen width, the brush stroke, and has nothing to do with the final size of the path.

The correction is applied for when a final user does a single click and to avoid get a dimensionless path we move a tiny tiny bit the first  and last point.

This is specific of the brush and does not need to be automatic.

ShaMan123: REVERTED adba2157

asturur (approved):

PR #7141, opened 2021-06-21
fix(fabric.Path): setting `path` during runtime

ShaMan123:
This PR with #7140 finalizes the replacement of #7082 

Apparently calling `setCoords` was redundant.
Everything looks good

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -70,9 +70,6 @@
           fabric.util.parsePath(path)
         );
 
-      if (!this.path) {
-        return;
-      }
       fabric.Polyline.prototype._setPositionDimensions.call(this, options || {});
     },
 
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -6,6 +6,7 @@
       min = fabric.util.array.min,
       max = fabric.util.array.max,
       extend = fabric.util.object.extend,
+      clone = fabric.util.object.clone,
       _toString = Object.prototype.toString,
       toFixed = fabric.util.toFixed;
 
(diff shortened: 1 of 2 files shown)


src/shapes/path.class.js
```diff
@@ -66,7 +73,19 @@
       if (!this.path) {
         return;
       }
-      fabric.Polyline.prototype._setPositionDimensions.call(this, options);
+      fabric.Polyline.prototype._setPositionDimensions.call(this, options || {});
```
ShaMan123, line 76: Do you want to extract `_setPositionDimensions` from Polyline?

ShaMan123:
@asturur merge?

ShaMan123:
@asturur

ShaMan123:
stale bot 😠

ShaMan123:
@asturur ?

asturur:
Ok i finally found the time to come back here.
I had many thoughts on this change and there are some thing we shouldn't really do.

A fabric.Path is defined by its own path data, and updating it is not so much common.
We shouldn't execute that logic with the set command.
I would prefer we get an imperative this.updatePath or this.changePath method.

If you want we can merge this PR removing the changes to _set/set, and is good to go.
Or we can work on adding changePath in this PR too ( honestly i would merge and reiterate ).
updatePath, changePath, should take of course path as an argument, and another argument to specify a point of the path that won't change position.

Changing a path data with something entirely different does not make sense and we can't really fix that.
But changing a path with another where only an handful of points have been moved is a more common use case, and the hard part is always having the 2 render in sequence look good

ShaMan123:
Done

asturur:
i will add the changePath method with the optional anchor point

asturur:
i forgot to merge this?

asturur:
now i have as a task to write the function that call _setPath but also keep the position steady.

PR #7147, opened 2021-06-21
fix(EraserBrush): check intersection taking in account canvas vpt

ShaMan123:
The check used by Eraser Brush to determine if to add a path to an object needs to account for canvas viewport transform.
This fixes the issue.


asturur:
Add a line to changelog please ( maybe add here or on the other approved PR 2 lines for both pr or you will get unnecessary merge conflicts )

asturur (approved):

PR #7156, opened 2021-06-25
feat(fabric.Text): improve text on a path

asturur:
- PR to monitor progress looking at diffs


src/shapes/text.class.js
```diff
@@ -741,23 +757,49 @@
...
+        }
+        if (this.direction == "ltr") {
+          positionInPath += this.startOffset;
+        }
+        else if (this.direction == "rtl") {
+          positionInPath -= this.startOffset;
+        }
       }
-      for (i = 0; i < line.length; i++) {
+      for (i = reverse ? line.length - 1 : 0; reverse ? i >= 0 : i < line.length; reverse ? i-- : i++) {
```
asturur, line 792: why you want to loop through graphemes in reverse?

src/shapes/text.class.js
```diff
@@ -204,6 +204,22 @@
...
+     */
+    startOffset:               0,
+
+    /**
+     * Which side of the path the text should be drawn on.
+     * Only used when text has a path
+     * @type String
+     * @default
+     */
+    side:               "left",
```
asturur, line 221: left/right can be up/down either. But why we do want this now?
textAlign and startoffset are big enough for a single PR.

src/shapes/text.class.js
melchiar: It's reversed when the `side` property is 'right` or `direction` is `rtl`

src/shapes/text.class.js
melchiar: Side left/right is the name of the svg equivalent property so I figured it'd be good to be consistent. 
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/side

The side logic is tied pretty closely to rtl support so I can open a separate pr for those if you think this too much for one pr.

asturur: what happen if we just let the current logic run over a path during rtl mode?
is it completely garbage?

melchiar: ok, I see the problem. I was misunderstanding how right-to-left text works. I'll make some changes.

src/shapes/text.class.js
melchiar: This seemed like the simpler approach than figuring out how to reverse the path itself. Would you approach this differently?

src/shapes/text.class.js
asturur: My suggestion is to make textAlign on the path work. Cover it with visual tests as we can,
We can then looks like how it behaves with RTL text, and then if really needs some extra tweaks we can do that.

asturur:
@melchiar i hijacked the PR. I want to merge textAlign to move forward, and talk about rtl and side.
Side was sort of working, but rtl was complicating everything else. (inlcuding side).
Let's get this in, and let's takle rtl + side on a separate pr.

melchiar:
Looks good!

PR #7175, opened 2021-07-01
feat(BaseBrush): `clipPath`

### LINKED ISSUE #5431: clipPath for canvas ignores free drawing
## Version
2.4.5

## Test Case
http://jsfiddle.net/synamqvu/

## Steps to reproduce
Try to draw anything, free drawing mode enabled.

## Expected Behavior
A circle is used as clipPath for canvas, it should clip contextTop as well.
## Actual Behavior
It is clipping only after mouse up. clipPath is not applying to contextTop while using free drawing.

### LINKED ISSUE #6955: Adding Freehand with ClipPath to Object Group
Hello,

I am working with Fabricjs and i'd like to draw inside of a circle and have the path constrained to the circle. This includes the selection area of the group.

UPDATE:

Seem to have solved the first point

http://jsfiddle.net/42snph1d/14/

**However, the selection are is too big. Anyway to reduce it so it is the side of the circle and not the path?**

**Is it possible to truncate the PATH to only include points inside the circle?**

--------------------------------------------------------------------------------------------------------------

http://jsfiddle.net/42snph1d/12/

I am seeing weird artefacts when moving the object around. 

1) The freehand drawing can be seen in the canvas outside of the circle when moving
2) The group selection is large, includes the entire freehand drawing. I'd like the selection area to only contain the circle.

Any obvious fixes here?

circle

Any advice would be appreciated.

ShaMan123:
### Overview
This PR adds a `clipPath` property to all brushes, aligning them with `fabric.Object`.
It **BREAKS** `SprayBrush`, it can be avoided with different naming if necessary.

### Motivation
closes #7134 
closes #6955 
closes #5431

### Demo
~~https://codesandbox.io/s/upbeat-microservice-mkbzs?file=/src/App.tsx~~
[Edit fabric_app (forked)](https://codesandbox.io/s/fabric-app-forked-9hcztm?fontsize=14&hidenavigation=1&module=%2Fsrc%2FBrushClipPath.tsx&theme=dark)

### Logic
- Added a `clipPath` property to `BaseBrush` and exposed clip path methods
- Exposed a common `render` + `_render` method on `BaseBrush` which conforms to Object#render and is in charge of:
   1. Preparing the context (transformation etc.)
   2. calling `_render` which is implemented by each brush
   3. render clipPath
   4. cleanup
- Made all subclasses use `_render` ~~instead of `_render` - **BREAKING** `SprayBrush` to make it compatible with all brushes (renamed it's previous `render` method to `renderChunck`), renamed `EraserBrush` `render` method to `renderAll`~~
- Added clip path support to segment rendering
- Added logic to apply the clip path to the resulting object created by the brush

### To Do
- [x] Eraser clip path - once #7470 is merged can probably be supported by clipping the mask created by eraser brush

### Watch
Pay no attention to eraser bug. It's fixed by the group PRs.
ezgif com-gif-maker (3)

**With Eraser patched up**
ezgif com-gif-maker (4)

image

**`absolutePositioned`** + viewportTransform
ezgif com-gif-maker (6)


asturur:
i do not think we should do that.
The brush should just respect the canvas clipPath if any, the issue is that rendering the topLayer, the contextTop is currently unmanaged, and brush, group selection, text selection, all fight for it.

We should work toward having a clear order of operation there and adding the clipPath there if necessary.

ShaMan123:
Did you try the demo?
What about a use case that you need the draw over parts of the canvas. You can't achieve this right now. Say you want to draw only over some objects.
Ohh you mean that if I use active selection it will be clipped..... I get you. What if we render a cache canvas to workaround this issue?

ShaMan123:
Thinking again it can't happen. Because once `mouseup` the brush clears the top context and create a Path object so it won't collide with group/text selection.

ShaMan123:
Isn't canvas clip path applied to all of it? That is not the use case described or the motivation of this PR

ShaMan123:
I think this is a cool feature.
I'll put it at hold for now.

asturur:
@ShaMan123 this is a cool feature, yes.
But i want to redo the contextTop or topLayer entirely.
This code will create a feature that will get in the way of reorganizaing.
In the top layer we need to be able to handle:
- controls
- group selection
- brush creation ( not always because for example the eraser brush needs to be in the main canvas )
- eventually small custom stuff for the developers

brush themselves need to be changed in something that is more similar to custom controls, one class and callbacks for your own rendering.

ShaMan123:
@asturur what about this?
Want this in? I think it's great. Simply need to address the eraser and then the feature is supported by all brush interactions

ShaMan123 (commented):
Updated from master.
Suitable for v6.
Eraser supported (needs group PRs to merge into master to function properly)
~~Something is off with Eraser. I don't know if it's eraser or the clip path.
Need to check deeper.~~

ShaMan123 (commented):
Ready to merge!

ShaMan123 (commented):
Fixed `absolutePositioned`
Visual tests fail but after examining myself I don't see any visual difference.

ShaMan123 (commented):
Updated from master, resolved conflicts and tested

PR #7181, opened 2021-07-04
fix(fabric.StaticCanvas): restore canvas size when disposing

ShaMan123:
When `enableRetinaScaling=true` the canvas element is resized.
When disposing canvas it is important to restore it's size in case the app will be reusing the canvas element.
With react, for example, `hot reload/fast refresh` disposes the fabric canvas instance but reuses the canvas element. 
This means that `_initRetinaScaling` is run again on the same canvas element, causing unwanted size changes.

### Sandbox!!
https://codesandbox.io/s/react-sandbox-forked-kh2vb?file=/src/App.tsx
I love the sandbox script !

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -523,7 +523,7 @@
       }
 
       fabric.util.addClass(this.lowerCanvasEl, 'lower-canvas');
-
+      this._originalCanvasStyle = this.lowerCanvasEl.style;
       if (this.interactive) {
         this._applyCanvasStyle(this.lowerCanvasEl);
       }
@@ -1770,13 +1770,13 @@
       this.overlayImage = null;
       this._iTextInstances = null;
       this.contextContainer = null;
+      // restore canvas style
+      this.lowerCanvasEl.classList.remove('lower-canvas');
+      this.lowerCanvasEl.style = this._originalCanvasStyle;
+      delete this._originalCanvasStyle;
       // restore canvas size to original size in case retina scaling was applied
       this.lowerCanvasEl.setAttribute('width', this.width);
       this.lowerCanvasEl.setAttribute('height', this.height);
-      if (this.upperCanvasEl) {
-        this.upperCanvasEl.setAttribute('width', this.width);
-        this.upperCanvasEl.setAttribute('height', this.height);
-      }
       fabric.util.cleanUpJsdomNode(this.lowerCanvasEl);
       this.lowerCanvasEl = undefined;
       return this;
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1770,6 +1770,13 @@
       this.overlayImage = null;
       this._iTextInstances = null;
       this.contextContainer = null;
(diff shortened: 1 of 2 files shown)


src/static_canvas.class.js
```diff
@@ -1770,6 +1770,13 @@
       this.overlayImage = null;
       this._iTextInstances = null;
       this.contextContainer = null;
+      // restore canvas size to original size in case retina scaling was applied
+      this.lowerCanvasEl.setAttribute('width', this.width);
+      this.lowerCanvasEl.setAttribute('height', this.height);
+      if (this.upperCanvasEl) {
+        this.upperCanvasEl.setAttribute('width', this.width);
+        this.upperCanvasEl.setAttribute('height', this.height);
+      }
```
asturur, line 1779: the upperCanvasElement goes away with Canvas, that is the class that instantiate it.
While lowerCanvasEl gets back into the dom on place of the original one, so resizing it as you did is a good idea.

asturur:
So this is good as it is once we removed the upperCanvasElement.
I want to add this consideration:
At initialization time fabricJS wrap the canvas with a wrapper and change the canvas style.
If we want to be nice to the dom we should save the original style attribute in a string and apply it back at dispose time.
There may be other attributes set on the canvas before we initialize and keep track of all of them may be hard.
A more generic solution and less probably less correct would also be to just remove the applied style on the canvas at dispose time.
As of now if you dispose in useEffect in your sandbox you will notice the canvas is absolute positioned in a corner of the screen

src/static_canvas.class.js
ShaMan123: I didn't understand what you meant about upperCanvasElement. 
You mean that it gets removed so there's no point in these code lines?

ShaMan123:
If I understand correctly you prefer saving the style. Am I right?

ShaMan123:
This really helps in creating a good `useCanvas` hook for the template.

asturur:
Well we can save the style in another PR, this is still fixing the issue that a disposed canvas will twice larger from when it started since the container with retina scaling style gets removed

src/static_canvas.class.js
asturur: Yes if you have a StaticCanvas you have no `upperCanvasEl`. if you have a Canvas, you call Canvas.dispose() that will first remove the upperCanvasEl, and then call his parent function, this one.
so when this function runs there should be never a upperCanvasEl

ShaMan123:
I want to finish it here. I'll push a commit

src/static_canvas.class.js
ShaMan123: 4c03569

ShaMan123:
Size and style are now restored when disposing.
I am not sure regarding `classList`, is it supported in node environment? I guess the test should fail if not?

asturur:
JSDOM support most of the html dom methods and properties, but this is good enough, is still better than it was before.

asturur (approved):

asturur (approved):

PR #7186, opened 2021-07-04
WIP fix(fabric.Object): render clipPaths sharper

### LINKED ISSUE #5494: Resizing objects with a clipPath gives blurry edges
## Version
2.6.0

## Test Case
https://jsfiddle.net/mattiasr/h7t6yp8d/12/

Nodejs or browser? Browser
Which browsers? Chromium, Firefox

## Steps to reproduce
Resize the object in the fiddle quite a bit larger.

## Expected Behavior
The edge of the circle gets redrawn to be sharp

## Actual Behavior
The edge of the circle stays blurry

asturur:
This is an issue we have since 2.4.0, clipPaths don't scale up well with object.
The issue, that is also a lesson, is that with the first round of clipPath release we added too many features ( nested clippaths, absolute clippaths, group of clipPaths ) and we thought we would fix the sharpness later.
Fixing the sharpness and keeping things working is hard, this has been in my to do list for long, i m not sure how to completely solve it and this is just a start.

I need to write a bunch of tests for regression since the one we have are not that much extensive

close #5494

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1083,6 +1083,10 @@
       ctx.restore();
     },
 
+    /**
+     * this function render the cache if necessary.
+     * if not, won't do anything
+     */
     renderCache: function(options) {
       options = options || {};
       if (!this._cacheCanvas) {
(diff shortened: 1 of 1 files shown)


asturur:
replaced by #9774

PR #7189, opened 2021-07-06
Added vibrance filter (for increasing saturation of muted colors)

melchiar:
Added a Vibrance filter for increasing the saturation of an image's more muted colors. Formula based on [Caman.js](https://github.com/meltingice/CamanJS/blob/master/src/lib/filters.coffee)

image

https://jsfiddle.net/melchiar/oq4eydrj/


src/filters/vibrance_filter.class.js
```diff
@@ -0,0 +1,114 @@
...
+        'float max = max(color.r, max(color.g, color.b));\n' +
+        'float avg = (color.r + color.g + color.b) / 3.0;\n' +
+        'float amt = (abs(max - avg) * 2.0) * uVibrance;\n' +
+        'color.r += max != color.r ? (max - color.r) * amt : 0.00;\n' +
+        'color.g += max != color.g ? (max - color.g) * amt : 0.00;\n' +
+        'color.b += max != color.b ? (max - color.b) * amt : 0.00;\n' +
+        'gl_FragColor = color;\n' +
+      '}',
+
+    vibrance: 0,
```
asturur, line 47: does vibrance works between 0 to 1 or -1 to 1? This information is useful

src/filters/vibrance_filter.class.js
```diff
@@ -0,0 +1,114 @@
...
+
+    mainParameter: 'vibrance',
+
+    /**
+     * Constructor
+     * @memberOf fabric.Image.filters.Vibrance.prototype
+     * @param {Object} [options] Options object
+     * @param {Number} [options.vibrance=0] Vibrance value for the image (between -1 and 1)
```
asturur, line 55: ok is written here, but we can put this also over the vibrance property

asturur:
Should we add a test?

asturur:
i can setup a test suite for all the filters maybe since you don't have examples.
We could use that image as reference

src/filters/vibrance_filter.class.js
melchiar: done

melchiar:
yeah, I guess we don't really have any visual tests for filters yet, do we

src/filters/saturate_filter.class.js
```diff
@@ -15,7 +15,7 @@
    * @see {@link http://fabricjs.com/image-filters|ImageFilters demo}
    * @example
    * var filter = new fabric.Image.filters.Saturation({
-   *   saturation: 100
+   *   saturation: 1
```
melchiar, line 18: changed this to 1 (100 makes no sense when the recommended range is -1 to 1

src/filters/saturate_filter.class.js
```diff
@@ -43,6 +43,14 @@
...
       '}',
 
+    /**
+     * Saturation value, from -1 to 1.
+     * Increases/decreases the color saturation.
+     * A value of 0 has no effect.
+     * 
+     * @param {Number} saturation
+     * @default
+     */
```
melchiar, line 53: added a description to the variable saturation filter too since it was missing

asturur (approved):

asturur:
I had this idea on my head for a bit:

this code here:

```js
    /**
     * Return WebGL uniform locations for this filter's shader.
     *
     * @param {WebGLRenderingContext} gl The GL canvas context used to compile this filter's shader.
     * @param {WebGLShaderProgram} program This filter's compiled shader program.
     */
      return {
        uVibrance: gl.getUniformLocation(program, 'uVibrance'),
      };
    },

    /**
     * Send data from this filter to its shader program's uniforms.
     *
     * @param {WebGLRenderingContext} gl The GL canvas context used to compile this filter's shader.
     * @param {Object} uniformLocations A map of string uniform names to WebGLUniformLocation objects
     */
    sendUniformData: function(gl, uniformLocations) {
      gl.uniform1f(uniformLocations.uVibrance, -this.vibrance);
    },
```

is sort of repeated for each filter, but at the end is conceptually:

```
uniformsValeus = {
  uVibrance: function (filter) { return -filter.vibrance };
}
```

We still need a smart way to identify the type of the filter, in this case `1f`

All the rest is wrapping code that could be in the base filter class.
i'm sure this would make the filters more easy to read and maybe save some byte here and there.

If not all, at least `getUniformLocation can be streamlined to a data declaration.

PR #7201, opened 2021-07-07
fix(shapes): fabric.Object._fromObject never should return

rockerBOO:
`fabric.Object._fromObject` doesn't return anything, and is processed through the callback.

This removes the return and fixes the documentation. This is mostly to upstream this fix to Typescript types as it incorrectly thinks Polygon, Circle, Ellipse return an object.

Thank you!

### DIFF
--- a/src/shapes/polygon.class.js
+++ b/src/shapes/polygon.class.js
@@ -74,7 +74,7 @@
    * @param {Function} [callback] Callback to invoke when an fabric.Path instance is created
    */
   fabric.Polygon.fromObject = function(object, callback) {
-    return fabric.Object._fromObject('Polygon', object, callback, 'points');
+    fabric.Object._fromObject('Polygon', object, callback, 'points');
   };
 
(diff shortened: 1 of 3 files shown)


src/shapes/polygon.class.js
```diff
@@ -74,7 +74,7 @@
    * @param {Function} [callback] Callback to invoke when an fabric.Path instance is created
    */
   fabric.Polygon.fromObject = function(object, callback) {
-    return fabric.Object._fromObject('Polygon', object, callback, 'points');
+    fabric.Object._fromObject('Polygon', object, callback, 'points');
```
asturur, line 77: can you add return void to this one too?

asturur:
Hi @rockerBOO thanks this is good. Can you address the comment?

src/shapes/polygon.class.js
rockerBOO: Added

asturur (approved):

asturur:
Thank you for the contribution

Issue #7171, opened 2021-06-30
How to add Erasing with Eraser Brush from Fabric.js to Next.JS(React.JS) project


iamibrahimkholil:
I need to add an eraser to my fabric js canvas from fabric js Eraser Brush in the following link. [http://fabricjs.com/erasing](url). Now I the documentation it is mentioned that it's not included in the default build of fabric js. So I googled a lot about how to add that to my NextJS project. Can anyone please help with this? Now to add to this problem I tried to download a custom build  JS file from fabric js website both Dev Build and Production Build and paste it in node_modules/fabric/dist/ folder but then I'm getting Reference Error: Window is not defined. I get it, Its Next.JS project and code do run on server side so window will be undefined. So how can I solve this problem. I need to use Erasing with Eraser Brush from fabric js.

## Version
4.5.1

## Information about environment
Next.JS

ShaMan123:
try `node build.js modules=ALL requirejs exclude=gestures,accessors` from `fabric` directory

asturur:
@iamibrahimkholil does your project has an npm install step?
If yes, you can define a postinstall script in your package.json.

In your post install part add a script that change directory to fabric js ( node_modules/fabric ) and that run `npm run build.js modules=ALL requirejs exclude=gestures,accessors`. In this way the build you need will be prepared every time you install a new version or when you deploy your app.

iamibrahimkholil:
[https://codesandbox.io/s/focused-christian-7s5gx?file=/src/EraserBrush.js](url) I used this link and imported Eraser Lib as an ES6 module. Is it a good solution to follow?

ShaMan123:
I suggest you file a bug report for the `fill.render` issue (I thought I pushed a commit that fixed it but apparently not) after you read this carefully http://fabricjs.com/erasing, there's a section about bg/overlay

Sounav-unstudio:
@iamibrahimkholil  how did you exactly manage to install the custom build of fabric in your dev environment as well as in production ? I am also using Next,js and am facing issues. Whenever I try to run the node build.js modules=ALL requirejs exclude=gestures,accessors, it gives the following error
`Built distribution to dist/fabric.js (requirejs-compatible)
Minification failed using uglifyjs with uglifyjs  --compress --mangle --output fabric.min.js fabric.js
Minifier error output:
Error: Command failed: uglifyjs  --compress --mangle --output fabric.min.js fabric.js
'uglifyjs' is not recognized as an internal or external command,
[...]

PR #7220, opened 2021-07-16
Fix TypeError: fill.render is not a function

chr314:
check if fill.render is function

fixing https://github.com/fabricjs/fabric.js/issues/7171#issuecomment-876153693

### DIFF
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -112,7 +112,9 @@
         if (needsVpt) {
           ctx.transform(v[0], v[1], v[2], v[3], v[4], v[5]);
         }
-        fill && fill.render(ctx);
+        if (fill && typeof fill.render === 'function') {
+          fill.render(ctx);
+        }
         object && object.render(ctx);
         ctx.restore();
       }


asturur:
@ShaMan123 does this seems a legit fix for you?

ShaMan123:
Yes and no.
Yes it will fix the error but I am not sure why it is happening. 
The logic assumes that the background/overlay color has been replaced by a `Rect` so it could be erased by the eraser. 
That is where the true bug originates from.
I think I can reproduce now that I understand.

When I built the eraser mixin this specific logic was handled by the eraser but later on I moved it to Canvas.

If I am not mistaken this is where the bug comes from:
https://github.com/fabricjs/fabric.js/blob/e6d16d3a03b972cf63babb59b351916d007fa563/src/mixins/eraser_brush.mixin.js#L77
https://github.com/fabricjs/fabric.js/blob/e6d16d3a03b972cf63babb59b351916d007fa563/src/mixins/eraser_brush.mixin.js#L95

This logic was correct when the eraser brush was in charge of replacing bg/overlay color to rects if they were erasable.
Now if fails because it has no replacement mechanism.
I moved it because it seemed too expensive to run every mouse move and I stick to it.

So if my assumptions are correct the true fix would be to change the else blocks and add rects instead of filters/string values.
In addition, this proposed fix will only remove the error but it will not render the actual bg/overlay.

**Therefore I think this is not a fix after all.**

There is another bug regarding this block of code that I pushed into #7100:
The bg/overlay color should not be affected by canvas transform.

ShaMan123:
@chr314 Could you help me reproduce?
Did you load the canvavs from JSON/SVG before this error happened?

ShaMan123:
I can't reproduce.
@chr314 are you setting bg/overlay color directly (assigning) or using the `setBackgroundColor` method as **STATED** in the instructions of the demo??

look at these 2 repos:
https://codesandbox.io/s/d16uq
https://codesandbox.io/s/dl79j

Read their comments to understand what they check

ShaMan123:
> Yes and no.
> Yes it will fix the error but I am not sure why it is happening.
> The logic assumes that the background/overlay color has been replaced by a `Rect` so it could be erased by the eraser.
> That is where the true bug originates from.
> I think I can reproduce now that I understand.
> 
> When I built the eraser mixin this specific logic was handled by the eraser but later on I moved it to Canvas.
> This logic was correct when the eraser brush was in charge of replacing bg/overlay color to rects if they were erasable because it checked it every render.
> Now it fails because it has no replacement mechanism.
> I moved it because it seemed too expensive to run every mouse move and I stick to it.
> 
> If I am not mistaken this is where the bug comes from:
> https://github.com/fabricjs/fabric.js/blob/e6d16d3a03b972cf63babb59b351916d007fa563/src/mixins/eraser_brush.mixin.js#L77
> 
> 
> https://github.com/fabricjs/fabric.js/blob/e6d16d3a03b972cf63babb59b351916d007fa563/src/mixins/eraser_brush.mixin.js#L95
> 
> So if my assumptions are correct the true fix would be to change the else blocks and add rects instead of filters/string values.
> In addition, this proposed fix will only remove the error but it will not render the actual bg/overlay.
> 
> **Therefore I think this is not a fix after all.**
> 
> There is another bug regarding this block of code that I pushed into #7100 (which is waiting for your feedback):
> The bg/overlay color should not be affected by canvas transform.

After looking closely at serialization I can say confindently that the eraser mixin works as expected with no problems since serialization calls `setBackgroundColor` or `setOverlayColor` that the eraser mixin extends.

https://github.com/fabricjs/fabric.js/blob/2eabc92a3221dd628576b1bb029a5dc1156bdc06/src/mixins/canvas_serialization.mixin.js#L142

ShaMan123:
What I would do is add a warning here instead explaining how to set `bg/overlay` color when using eraser and referencing the docs

chr314:
@ShaMan123  thanks
yeah I think I was using the bg color the wrong way

PR #7233, opened 2021-07-22
fix(fabric.util): iterator step could`t be 0.01

meetwhere-dev:
ex: 0.1 + 0.2 = 0.0000...00034

it will loss last one step when cal Path`s totalLength.


melchiar:
Sigh... Floating point errors. 😞 

Nice find.

asturur:
@meetwhere-dev  this is ok, but can you remove the built dist file?
`git checkout master dist` is what i usually do.

meetwhere-dev:
> @meetwhere-dev this is ok, but can you remove the built dist file?
> `git checkout master dist` is what i usually do.

oh Sorry about that.
I will take it

asturur:
Sorry you have to fix tests too, because the tests are based on the old weird values

asturur:
Do you know how to run tests?

PR #7236, opened 2021-07-23
Fix grammar

suman:
Fix grammar


melchiar:
Thanks for the PR! 

Just to be sure though I researched the rules regarding use of a-vs-an for acronyms and as with other words, it goes by the phonetic sound, not by whether the first letter is a vowel. 

So actually, because SVG begins with a vowel Ess sound, "an SVG" is technically correct. 

https://grammar.yourdictionary.com/vs/a-vs-an-basic-rules-and-exceptions.html

Issue #7037, opened 2021-04-23
Objects added via group don't fire 'added'


VanillinT:
## Version
4.4.0

## Test Case
https://jsfiddle.net/67y5jco8/1/

## Information about environment
Chrome 89.0.4389.128

## Steps to reproduce
Call addWithUpdate on a group with any fabric object as a parameter

## Expected Behavior
Objects fire 'added' when added to a group

## Actual Behavior
They don't fire 'added' when added to a group

asturur:
We had a PR for this somewhere to fix it.

ShaMan123:
+1

ShaMan123:
If I am not mistaken this happens because group doesn't call `Collection#add` and instead pushes the object to `_objects`:
https://github.com/fabricjs/fabric.js/blob/master/src/shapes/group.class.js#L167

ShaMan123:
#6619

Issue #6619, opened 2020-09-26
Group.addWithUpdate should leverage Group.add.


asturur:
We should reuse the code in some specific functions rather than duplicating it.
AddWithUpdate is supposed to Add and do extra work for the update.

Group.add should be reused so that eventual changes will go automatically in addWithUpdate.

Issue #6130, opened 2020-02-02
ActiveSelection does not fire Events on addWithUpdate(), removeWithUpdate()


awehring:
## Version
3.6.1

## Test Case
https://jsfiddle.net/awehring/ygxmbt6d/73/

## Information about environment
Browser. 
Tested with current Edge, FireFox

## Bug Description
An ActiveSelection does not fire Events on `addWithUpdate()`, `removeWithUpdate()`.
I would expect that `selection:updated` events are fired.

## Steps to reproduce
Run the fiddle and inspect the browser console. 

Click 3 times the  "Add to Selection" button until all 3 objects are selected.
Click 3 times the "Reduce Selection" button until no object is selected any more.

The console reports the fired events.

## Expected Behavior
`addWithUpdate()`, `removeWithUpdate()` should fire `selection:updated` events.

## Actual Behavior
These two Methods don't fire `selection:updated` events.

asturur:
they won't fire it. you made that button and you are in control of it, you do not need an event. On the button handler, run the code you need.

Does it make sense?

awehring:
Yes, thats possible.
But somewhat inconsistent:  `canvas.setActiveObject()` fires `selection:created` or `selection:updated` events (depending if the selection already existed). 
`canvas.discardActiveObject()` fires `before:selection:cleared` and `selection:cleared`.

asturur:
well but your button is going to add or remove from selection, and before doing so you know if you are going to modify or create a selection, depending if you already have one. I do not think addWithUpdate or removeWithUpdate should fire events.

i think that the selection is also firing object added and object removed being it part of the group class?

awehring:
The selection / group does fire events?

What's the Name of these events? Never heared of them.

asturur:
No it doesn't.
I was confused with the collection.mixin. Just canvas has events... that is silly too since users can't really add objects or remove them with embedded actions.

Please if you fill strongly about it, try to prove why you think is necessary. I can be wrong about it.

awehring:
I appreciate that you are interested in an open discussion.

My use case is a special kind of drawing application, where the user is able to select fabric objects with a button at the user interface. Afterwards (s)he can change some properties like `color`, `stroke`, `fontStyle`, etc.
[...]

PR #7241, opened 2021-07-25
fix(Group): fire added/removed events

ShaMan123:
### Motivation
#7037 #6619 #6130
I looked for the PR you mentioned @asturur, found something that looked off #6199

### Test it LIVE:
https://codesandbox.io/s/dev-sandbox-forked-ifg6i?file=/src/App.tsx

### Logic
I followed `StaticCanvas` impl of these events.

### Considerations:
1. Is this breaking? This fix fires an `added` `removed` event on object. While until now this was fired on objects that were added to/removed from canvas. This might break stuff for devs... I added a target property (the group) to the event to differentiate.
2. What about initialization? Should events fire on objects added within the initialization block? **Currently** they don't.

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -164,7 +164,7 @@
           // if this group is inside another group, we need to pre transform the object
           fabric.util.removeTransformFromObject(object, this.group.calcTransformMatrix());
         }
-        this._objects.push(object);
+        this.add(object);
         object.group = this;
         object._set('canvas', this.canvas);
       }
@@ -205,12 +205,16 @@
       this.dirty = true;
       object.group = this;
       object._set('canvas', this.canvas);
+      this.fire('object:added', { target: object });
+      object.fire('added', { target: this });
     },
 
     /**
      * @private
      */
-    _onObjectRemoved: function(object) {
+    _onObjectRemoved: function (object) {
+      this.fire('object:removed', { target: object });
+      object.fire('removed', { target: this });
       this.dirty = true;
       delete object.group;
     },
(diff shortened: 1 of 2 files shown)


ShaMan123:
#6619

src/shapes/group.class.js
```diff
@@ -164,7 +164,7 @@
           // if this group is inside another group, we need to pre transform the object
           fabric.util.removeTransformFromObject(object, this.group.calcTransformMatrix());
         }
-        this._objects.push(object);
+        this.add(object);
         object.group = this;
         object._set('canvas', this.canvas);
```
ShaMan123, line 169: these 2 assignments can be removed (redundant, assigned in `_onObjectAdded`)

ShaMan123: @ShaMan123
remove redundant assignment
3dc8c33

ShaMan123:
46036ed make this PR good because it fixes the firing of the events to when object and group are in the correct state (after group logic has finished resizing, transforming etc.)

src/shapes/group.class.js
```diff
@@ -177,6 +177,7 @@
       else {
         this.setCoords();
       }
+      object && this._fireGroupEvent('added', object);
```
ShaMan123, line 180: Should this fire only if object is not nested?

ShaMan123:
Leaving this open, but I think this should an will be addressed by the group rewrite.
I might merge this into v6 with some patches.

ShaMan123:
stale 
Old dead group

PR #7242, opened 2021-07-25
Update documentation for clone and extend methods

Serene47:
Add more information in the documentation for `clone` and `extend` methods.


asturur:
You can also be more strong.
`This will break a fabricObject if used over it`, `is used internally in some cases as a shield for mutation during serialization and restore`

asturur (approved):

PR #7250, opened 2021-07-28
fix: fixed aligning_guidelines issues in zoom mode

beidan:
the old aligning_guidelines in zoom mode will cause a bug，this pull request can fixed it


asturur:
hi @beidan thank you for the PR.
Is it possible for you to demo the fix in some way?
with a codepen or a fiddle.

PR #7258, opened 2021-07-29
feat(EraserBrush): better `erasing:end` event

ShaMan123:
1. Fire `erasing:end` event on object.
This is important since #7100 because erasable targets may be nested deep in the object tree and there is no way of knowing if a nested object was erased.
2. Add `subTargets` property to canvas `erasing:end` event.

### Context
Canvas fires `erasing:end` event with a `targets` property. Since #7100 `targets` is not adequate because erasable targets may be nested deep in the object tree.
I am thinking of a way to add a `subTargets` property to canvas `erasing:end` event that will contain all erased nested objects. To achieve this I need some advice.
Can I assume that cloning a path is a sync and not an async process?

### DIFF
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -867,6 +867,9 @@
           obj.fire('erasing:end', {
             path: path
           });
+          if (obj.group && Array.isArray(_this.__subTargets)) {
+            _this.__subTargets.push(obj);
+          }
         });
       },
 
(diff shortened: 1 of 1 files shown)


ShaMan123:
867c809 adds `subTargets` by assuming that `path.clone` is a sync process.

asturur:
path clone takes a callback, but is async only if the content of the path is.
(ex: an image in pattern as a fill/stroke)

asturur (approved):

PR #7259, opened 2021-07-29
added pathSide property to text on path

melchiar:
Here's my newest implementation of the `side` property for text on a path. Note that I've used the name `side` to be consistent with the related [SVG property](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/side), although SVG `side` is currently supported only in Firefox.

Feel free to test:
https://jsfiddle.net/melchiar/oLq91r2t/

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -793,7 +793,7 @@
             break;
           //todo - add support for justify
         }
-        positionInPath += this.pathStartOffset * reverse ? -1 : 1;
+        positionInPath += this.pathStartOffset * (reverse ? -1 : 1);
       }
       if (path) {
         for (i = reverse ? line.length - 1 : 0;
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -210,7 +210,7 @@
      * @type Number
      * @default
      */
-    startOffset:               0,
+    pathStartOffset:               0,
 
     /**
      * Which side of the path the text should be drawn on.
@@ -218,7 +218,7 @@
      * @type {String} 'left|right'
      * @default
      */
-    side:               'left',
+     pathSide:               'left',
 
     /**
      * @private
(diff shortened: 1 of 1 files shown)


src/shapes/text.class.js
```diff
@@ -212,6 +212,14 @@
      */
     startOffset:               0,
 
+    /**
+     * Which side of the path the text should be drawn on.
+     * Only used when text has a path
+     * @type {String} 'left|right'
+     * @default
+     */
+    side:               'left',
```
asturur, line 221: for the previous discussion, should we call this `pathSide`?

src/shapes/text.class.js
```diff
@@ -818,6 +832,14 @@
...
       graphemeInfo.renderTop = info.y - startingPoint.y;
       graphemeInfo.angle = info.angle;
+      switch (this.side) {
+        case 'left':
+          graphemeInfo.angle = info.angle;
+          break;
+        case 'right':
+          graphemeInfo.angle = info.angle + Math.PI;
+          break;
+      }
```
asturur, line 842: why in other part of the code we use a ternary and here a switch case? Are there are other values of side we should be aware of? if is just left/rigght we can have 
```
graphemeInfo.angle = info.angle + this.side ===  'right' ? Math.PI : 0;
```

src/shapes/text.class.js
```diff
@@ -818,6 +832,14 @@
       graphemeInfo.renderLeft = info.x - startingPoint.x;
       graphemeInfo.renderTop = info.y - startingPoint.y;
       graphemeInfo.angle = info.angle;
```
asturur, line 834: this seems a leftover line?

melchiar: oops

src/shapes/text.class.js
melchiar: you're right, that's better

src/shapes/text.class.js
melchiar: sounds good. I think let's change `startOffset` to `pathStartOffset` as well for the same reason

asturur: yes i agree is fresh enough that no one noticed it yet.

asturur:
i'm merging.
Can you follow up with a visual test?

PR #7268, opened 2021-08-02
ci(): publish @next tag on PR merge

ShaMan123:
#7256

### DIFF
--- a/.github/workflows/npmpublish-next.yml
+++ b/.github/workflows/npmpublish-next.yml
@@ -1,0 +1,20 @@
+name: Node.js Package @next
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  publish-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/npmpublish-next.yml
```diff
@@ -0,0 +1,20 @@
...
+    branches: [ master ]
+
+jobs:
+  publish-npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
```
asturur, line 14: let's bump this to 16 since is the current LTS.

asturur: (we will do it for the other scripts too later)

ShaMan123:
Did it work?

ShaMan123:
This try didn't work right?

asturur:
no i don't think is working at all.
Not sure if is permission or else, i did not check anymore

ShaMan123:
up

PR #7269, opened 2021-08-03
fix(EraserBrush): safeguard eraser from `prototype.fill` changes

ShaMan123:
### Motivation
I came across an edge case.
I was changing `Object.prototype.fill` to `null`.
It made erasing break.

The eraser logic uses a `Rect` and draw paths over it so clipping can be done. 
If fill is `null` the eraser erases everything.

This PR safeguards the eraser from such a case.

### DIFF
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -835,6 +835,7 @@
         if (!obj.getEraser()) {
           var size = obj._getNonTransformedDimensions();
           var rect = new fabric.Rect({
+            fill: 'rgb(0,0,0)',
             width: size.x,
             height: size.y,
             clipPath: obj.clipPath,


ShaMan123:
I am thinking of another safeguard for eraser motivated by #7220

asturur:
The eraser is an high level feature that can have its own requirements, is fine to have this.

ShaMan123:
I was thinking this because:
1. It is not straight forward.
2. It is breaking and requires migrating all bg/overlay color setting logic before adding the eraser brush, which is annoying to say the least (I hate a lot of setup for features that I perceive as a one line plugin).
2. It creates noise for devs: errors and frustration (the only place to read about it is in the demo)
2. It creates noise for maintainers/repo: false/repetitive issues and frustration (*"Did you read the how to??? 😠😠"*)
2. It feels bad from the dev point of view, makes fabric feel unstable, bad PR (public relations)

If I were a dev using fabric and opting in to erasing with existing code I would find myself with a broken app. 
I don't think it's major to make it backward compatible.

Anyways I need to prioritize so I take your advice and move it to the far end.

ShaMan123:
Regardless this can be merged

Issue #5187, opened 2018-08-22
eraser and recover brush for fabricjs( canvas)


komagic:
I created some methods to implement eraser and recover brush in fabricjs（by globalCompositeOperation）,but i still dont know  how to use original canvas methods
 in fabricjs?

asturur:
Hi @komagic, i do not understand your question.

Can you rephrase it, please?

komagic:
how to create a path with alpha feather edges( fabricjs)?
![2018-08-22 3 30 59](https://user-images.githubusercontent.com/13415767/44449386-92642100-a620-11e8-92ae-b21274c7041b.png)

asturur:
There is no such a tool.
If you can define a path for your shape you can overlay a series of strokes with different opacity to imitate the effect.

komagic:
how to copy the content of the canvas to another canvas correctly?   use todataURL pls. (do not use toJson)

komagic:
success! Then I want to drop the picture back to the old fabric canvas, the picture should be put correctly?

asturur:
i can t really follow step by step your project. You should make a good open stack overflow snippet and ask help on stackoverflow. Here we can t provide this kind of help

Euphoria9702:
我最近也在想用fabric.js您当时实现的内容，demo可以放到git上给我参考学习一下吗，哈哈

Euphoria9702:
@ @komagic

ShaMan123:
@asturur can shadow be used here?

ShaMan123:
> haha，i am try to create a simple editor by fabricjs， i created an eraser and a recovery brush, as show below:
> ![2018-08-22 10 09 23](https://user-images.githubusercontent.com/13415767/44448736-adce2c80-a61e-11e8-81ba-4b0fbc2df2ce.png)
> 
> however, i do not know how to create a path with feather edges by fabricjs，can you show me some samples?

This is a cool feature

ShaMan123:
Tried it out, looks interesting:
https://codesandbox.io/s/exciting-architecture-bcgej?file=/src/App.tsx

hit the `erase` button

Euphoria9702:
how to created  recovery brush @komagic

ShaMan123:
Try finddling with `inverted erasing`
https://codesandbox.io/s/blissful-sutherland-yk05j?file=/src/EraserBrush.js:25469-25473
It simply changes the `globalCompositeOperation `
```js
[...]

PR #7271, opened 2021-08-04
feat(EraserBrush): inverted brush

ShaMan123:
### Motivation 
#5187 

### POC
https://codesandbox.io/s/blissful-sutherland-yk05j?file=/src/EraserBrush.js:25469-25473
Try finddling with `inverted erasing`
It simply changes the `globalCompositeOperation `
```js
(path | ctx).globalCompositeOperation = this.inverted
          ? "source-over"
          : "destination-out";

```
With existing erasing logic this approach simply adds a path that draws over erasing paths in the object's clip path.

---

### The Challenge
The problem is the rendering of the brush in such a case.
It will draw color and I would expect it won't and only recover erased paths.
This might be possible but will require some work.

#### Context
https://github.com/fabricjs/fabric.js/blob/master/src/mixins/eraser_brush.mixin.js#L420
>When erasing occurs, the path (generated by mouse move) clips the top ctx and reveals the bottom ctx.
 >  * This achieves the desired effect of seeming to erase only erasable objects.
 >  * After erasing is done the created path is added to all intersected objects' `clipPath` property.

#### Drawing Logic
https://github.com/fabricjs/fabric.js/blob/master/src/mixins/eraser_brush.mixin.js#L646

### Thoughts
To achieve an inverted/recovery effect the canvas should be prepared differently:
1. Render everything on bottom ctx Render all objects on bottom ctx 
2. Render top ctx in a way that when erasing it will reveal a rendered canvas with objects rendered **without** their eraser:
    - using a filter/[pattern](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createPattern) of canvas?
    - **render canvas on top ctx, clip entire ctx, set brush `globalCompositionOpertion =  "source-over"` to reveal top canvas**
3. On mouse up add the path as any eraser path (with existing logic), set path `globalCompositionOpertion =  "source-over"`

### Workaround
The effect of inverted erasing on a specific object can be achieved by setting the brush `globalCompositionOpertion =  "source-over"` and applying a color/filter with conjunction with #7175 

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/5187#issuecomment-892493315_

### DIFF
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -696,6 +696,7 @@
       _saveAndTransform: function (ctx) {
         this.callSuper('_saveAndTransform', ctx);
         ctx.globalCompositeOperation = this.inverted ? 'source-over' : 'destination-out';
+        this.inverted && (ctx.strokeStyle = 'rgba(255,255,255,0)');
       },
 
       /**
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -436,6 +436,11 @@
       type: 'eraser',
 
       /**
+       * When set to `true` the brush will create a visual effect of undoing erasing
+       */
+      inverted: false,
+
+      /**
        * Indicates that the ctx is ready and rendering can begin.
        * Used to prevent a race condition caused by {@link fabric.EraserBrush#onMouseMove} firing before {@link fabric.EraserBrush#onMouseDown} has completed
        *
@@ -690,7 +695,7 @@
        */
       _saveAndTransform: function (ctx) {
         this.callSuper('_saveAndTransform', ctx);
-        ctx.globalCompositeOperation = 'destination-out';
+        ctx.globalCompositeOperation = this.inverted ? 'source-over' : 'destination-out';
       },
 
       /**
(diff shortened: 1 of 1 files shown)


ShaMan123:
The best solution seems to be creating another brush (`AntiEraserBrush`) extending PencilBrush that will draw the canvas as a pattern (all erasers removed)
```js
 _saveAndTransform: function (ctx) {
          this.callSuper('_saveAndTransform', ctx);
          var canvas = fabric.util.createCanvasElement();
          canvas.width = this.canvas.width;
          canvas.height = this.canvas.height;
          var patternCtx = canvas.getContext('2d');
          var objects = this.canvas.getObjects().filter(function (obj) {
            return obj.erasable;
          });
          // prepare object tree by hiding/removing erasers for erasable objects
          this.prepareCollectionTraversal(this.canvas);
          this.canvas.renderCanvas(patternCtx, objects);
          this.restoreCollectionTraversal(this.canvas);
          ctx.strokeStyle = ctx.createPattern(canvas, 'no-repeat');
        
      },
```

ShaMan123:
This is low on my list

ShaMan123:
#7470 solves this

PR #7273, opened 2021-08-04
ci(): Update npmpublish-next to trigger on merged PR.

asturur:
@ShaMan123 That yml file isn't working, probably because a master push is not a merge PR, and in general we want to avoid direct push to master.

Let's try if this works


ShaMan123:
Any progress?

asturur:
no i need to incremenet the. version number, probably adding the commit to it.
We have all the code in place somehow, let me do it

PR #5936, opened 2019-10-27
Add ability to cancel animation

asturur:
Can make sense to have the animate utility to return a function that will let you cancel the animation imperatively.

the actual abort callback does not provide information to the callback itself and so is not that much useful.

close #5829

PR #7275, opened 2021-08-04
feat(animate): imperative abort

ShaMan123:
### Motivation
I use imperative abort a lot so I'd like the util to support it out of the box.
Current abort makes my code over grown and inconvenient.

### DIFF
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -22,6 +22,10 @@
    * @param {Function} [options.abort] Additional function with logic. If returns true, onComplete is called.
    */
   function animate(options) {
+    var aborted = false;
+    var abort = function () {
+      aborted = true;
+    }
 
     requestAnimFrame(function(timestamp) {
       options || (options = { });
@@ -47,7 +51,7 @@
             timePerc = currentTime / duration,
             current = easing(currentTime, startValue, byValue, duration),
             valuePerc = Math.abs((current - startValue) / byValue);
-        if (abort()) {
+        if (this.aborted || abort()) {
           onComplete(endValue, 1, 1);
           return;
         }
@@ -62,6 +66,8 @@
         }
       })(start);
     });
+
+    return abort;
   }
 
   var _requestAnimFrame = fabric.window.requestAnimationFrame       ||
--- a/src/util/animate_color.js
+++ b/src/util/animate_color.js
@@ -32,7 +32,7 @@
         originalOnChange = options.onChange;
     options = options || {};
 
-    fabric.util.animate(fabric.util.object.extend(options, {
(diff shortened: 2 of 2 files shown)


ShaMan123 (commented):
2892af4 makes all animation methods return an abort function.
This **BREAKS** chainability, hence 2 tests don't pass.
Made this a separate commit so it could be reverted with ease.

src/util/animate.js
```diff
@@ -47,7 +52,7 @@
             timePerc = currentTime / duration,
             current = easing(currentTime, startValue, byValue, duration),
             valuePerc = Math.abs((current - startValue) / byValue);
-        if (abort()) {
+        if (aborted || abort()) {
```
asturur, line 55: let's divide it as it was here:
https://github.com/fabricjs/fabric.js/pull/5936/files#diff-4c788168c81ce977ee91d228587e70e438fdd375b832462bcc998fa1e111ff4aR55

So that then the old abort() can be deprecated easily.

asturur:
https://github.com/fabricjs/fabric.js/pull/5936/files

This is pretty much similar to this PR i started 2 years ago and then i never finished.

Can we bring this part in too?
https://github.com/fabricjs/fabric.js/pull/5936/files#diff-6d25d09956ab859c1fffa84497d6edbc7c217ce9185766cd594aee58a9f56a2bR66

src/mixins/animation.mixin.js
```diff
@@ -73,8 +73,6 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         onComplete();
       }
     });
-
-    return this;
```
asturur, line 77: all the changes to `src/mixins/animation.mixin.js` can't be done because are breaking changes.
We can add them later when we have a major release

ShaMan123:
So do you want me to merge #5936 into this one or do you want to merge #5936 into master and drop this PR?

asturur:
No just add that small differences of that PR here.
Are just few tweaks

src/util/animate.js
ShaMan123: Don't you want to support both imperative and declarative abort?

ShaMan123:
> https://github.com/fabricjs/fabric.js/pull/5936/files
> 
> This is pretty much similar to this PR i started 2 years ago and then i never finished.
> 
> Can we bring this part in too?
> https://github.com/fabricjs/fabric.js/pull/5936/files#diff-6d25d09956ab859c1fffa84497d6edbc7c217ce9185766cd594aee58a9f56a2bR66

What part do you mean? The util additions and tests?

PR #7284, opened 2021-08-08
ci(): Fix the version for publishing many times

asturur:


asturur:
@ShaMan123 let's see if this fixes.

PR #7287, opened 2021-08-08
ci(): See if this publishing is accepted

asturur:


asturur:
@ShaMan123 this is not easy without tagging every single release.
npm needs a patched package.json to get the next release, commiting just with the git hash could be possible, but would bring to versions that aren't consecutives.
so you should install fabric@commit-hash rather than fabri@next because the tag will not have consecutive versions

ShaMan123:
I understand.
What if the build script changes the package and then reverts? 
I guess we can store the tag somewhere or [query npm](https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md) from within the build script to get the next semver tag.

ShaMan123:
```js
axios.get('https://registry.npmjs.org/fabric').then(res => {
    console.log(res.data['dist-tags'])
})
```
response:

```js
{ latest: '4.5.1', beta: '3.6.5' }
```
```

**Comment by ShaMan123** [comment]
1. I don't see a problem with how this works, the script prereleases and when you publish a patch/minor/major it increments accordingly.
2. I see the command throws an error if the tag exists

```bash
C:\Users\DELL\Desktop\DEV\fabric.js>npm version prerelease --preid=alpha
v4.5.2-alpha.3
npm ERR! code 128
npm ERR! Command failed: git -c core.longpaths=true tag v4.5.2-alpha.3 -m 4.5.2-alpha.3
npm ERR! fatal: tag 'v4.5.2-alpha.3' already exists
npm ERR!

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\DELL\AppData\Roaming\npm-cache\_logs\2021-08-09T05_43_37_083Z-debug.log

C:\Users\DELL\Desktop\DEV\fabric.js>npm version prerelease --preid=alpha
v4.5.2-alpha.4
```

asturur:
yes probably we can do like this, we pull and we increment it manually. But the script won't pre-release because last try it did not let me commit. And there isn't a `patch version but don't commit` option.

asturur:
i don't mind also tagging everything officially in the repo with the action.

ShaMan123:
I think that tagging everything is the best practice.
Very orderly...

ShaMan123:
Any updates?

Issue #7278, opened 2021-08-05
Background color doesn't scale when zoom out on fabric build with eraser


alperen:
## Version
4.1.0 (Erasing enabled)

## Test Case
[https://codesandbox.io/s/pedantic-leaf-7kvkh](https://codesandbox.io/s/pedantic-leaf-7kvkh)

## Information about the environment
All major browsers with the latest versions

## Steps to reproduce
- Click to "Draw" button
- Draw something into the canvas
- Click to "Erase" button
- Erase some part of that you drawed (bug isn't based on your drawed)
- Click to "Zoom Out" button
- Click to "Export" button

Exported image (it opens in a new tab) has no background color over the image, on the standard builds of the fabric, the background color is spreading over the exported image. I spread the background color by setting `backgroundVpt: false` over the canvas but it doesn't affect exporting.

You can see the difference between the standard build and with the eraser build by changing `index.html:25` to 

```html
<script
      id="fabric-build"
      src="src/fabric-builds/standart-build.js"
    ></script>
```

## Expected Behavior
Background color spread over the export image, for example:

![image](https://user-images.githubusercontent.com/3244438/128335160-9a251424-ac82-4944-8af5-cdc7435758d2.png)

## Actual Behavior
Background color doesn't spread over the exported image, it doesn't scale when the zoom out on the exported image.

![image](https://user-images.githubusercontent.com/3244438/128335253-499dab16-c6bc-443f-bff0-807516150f99.png)

asturur:
The eraser brush uses a rect to emulate a color, that is an hack.
@ShaMan123  why someone would want the background color to be erasable?
Why the developer shouldn't use an object as background if erasing is needed, and decide on his own how large this background is?

alperen:
Here is additional information, once you'll comment out line 32 to 34, it will erase the background color, makes it transparent

ShaMan123:
@asturur It is definitely a hack, explained here http://fabricjs.com/erasing.
So if I understand you, you vote for removing the ability to erase bg/overlay color?
This can be fixed for sure but there are more pit falls such as #7220.

ShaMan123:
I think someone would want to erase the background color in case they want to reveal something underneath. But it is in the scope of adding an object.

alperen:
I forked the sandbox with the new build which includes #7289, it seems the pr fixed by problem well. You can check out it from: https://codesandbox.io/s/aged-resonance-doied

[...]

PR #7289, opened 2021-08-09
fix(EraserBrush): dump canvas bg/overlay color support

ShaMan123:
### Motivation
fixes #7220 #7278  by removing erasability support for bg/overlay color


ShaMan123:
Once/If this is merged I'll PR the docs

asturur:
This can be probably be supported with some thoughts.
Eventually we can find a way and describe it in a clear example.
The developer can still use a rect, just he has to be aware that the rect has to be as big as the largest level of viewport he wants to use.

PR #7290, opened 2021-08-09
fix(EraserBrush): `intersectsWithObject` edge cases

### LINKED ISSUE #7277: Fabric.js Eraser not working properly if canvas is zoom in or zoom out
## Version
4.5.1

## Test Case
https://codesandbox.io/s/sparkling-dew-2e038

Although I am using it in NextJS application. I made the test case in Vanilla JS
Which browsers?
Chrome, Firefox etc.

## Steps to reproduce
First, draw something on the board then zoom in a lot then try to Erase. 
## Expected Behavior
It Should erase the drawing.

https://user-images.githubusercontent.com/79884206/128333869-cf06abde-3bb5-4286-82f2-83f34df51d32.mp4

## Actual Behavior
it won't work. It mostly happening while I am trying to eraser around the corner of the drawing. It is also happening if I zoom out the canvas a lot.

ShaMan123:
closes #7277

### DIFF
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -929,6 +929,8 @@
         }
 
         var path = this.createPath(pathData);
+        //  needed for `intersectsWithObject`
+        path.setCoords();
         canvas.fire('before:path:created', { path: path });
 
         // finalize erasing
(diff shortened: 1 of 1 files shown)


src/mixins/eraser_brush.mixin.js
```diff
@@ -929,6 +929,8 @@
         }
 
         var path = this.createPath(pathData);
+        //  needed for `intersectsWithObject`
+        path.setCoords();
```
asturur, line 933: i m worried this won't work if there is some zoom level going on
But this for sure is fixing something, so for now is fine.

PR #7298, opened 2021-08-13
fix(Fabric.text): account for fontSize in textpath dimensions

melchiar:
Text on a path is currently using the dimensions of the assigned path which can cause the cache size to be too small.

image

This fix adds the calculated text height to the object's dimensions to ensure that nothing is clipped in the cache canvas. 
image

Honestly though I'm not crazy about this solution since depending on the path's shape and direction this means the bounding box will sometimes be unnecessarily large. Let me know if you can think of a better idea.

image

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -795,8 +795,6 @@
           //todo - add support for justify
         }
         positionInPath += this.pathStartOffset * (reverse ? -1 : 1);
-      }
-      if (path) {
         for (i = reverse ? line.length - 1 : 0;
           reverse ? i >= 0 : i < line.length;
           reverse ? i-- : i++) {
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -400,13 +400,14 @@
       }
       this._splitText();
       this._clearCache();
+      var textHeight = this.calcTextHeight();
       if (this.path) {
-        this.width = this.path.width;
-        this.height = this.path.height;
+        this.width = this.path.width + textHeight;
+        this.height = this.path.height + textHeight;
       }
       else {
         this.width = this.calcTextWidth() || this.cursorWidth || this.MIN_TEXT_WIDTH;
-        this.height = this.calcTextHeight();
+        this.height = textHeight;
       }
       if (this.textAlign.indexOf('justify') !== -1) {
         // once text is measured we need to make space fatter to make justified text.


melchiar:
Obviously this change would requires fixing all of the visual tests since the new dimensions affect positioning. I'd like to try to find another solution to this issue if we can though. I've experimented with simply factoring the font size into the cache size calculation and leaving the object's bounding dimensions untouched but doing that seems to cause the cache's resolution suffer.

asturur:
a better, but complicated idea, would be to get the object size looking at the Path + the character boxes we create. For each char we have the width and the position + angle. That means we can collect all the 4 corners of each character and throw inside a bucket of points that will give us a bounding box.

src/shapes/text.class.js
```diff
@@ -400,13 +400,14 @@
       }
       this._splitText();
       this._clearCache();
+      var textHeight = this.calcTextHeight();
       if (this.path) {
-        this.width = this.path.width;
-        this.height = this.path.height;
+        this.width = this.path.width + textHeight;
+        this.height = this.path.height + textHeight;
```
asturur, line 406: This works because you make a single line of text.
It grows if is more than one line. I would use just the function that calculate the lineHeight ( the one called from calcTextHeight)

melchiar:
> a better, but complicated idea, would be to get the object size looking at the Path + the character boxes we create. For each char we have the width and the position + angle. That means we can collect all the 4 corners of each character and throw inside a bucket of points that will give us a bounding box.

That might be a bit over complicated. I'm imagining a scenario like this where text is curved around the top and bottom of two circles. Aligning the two circles to each other is a lot easier if the bounding boxes are uniform and you can use the same center position.

image

src/shapes/text.class.js
melchiar: right. I can fix that

melchiar:
I've decided to take a different approach and simply adjust the cache dimensions based on the font size. I think this is a better solution than increasing the text object's dimensions since from a UX perspective the path dimensions are more important in my opinion (this is also the way Illustrator does it).

Anyone who wants to include the text in the bounding box can do so with padding.

melchiar:
No clipping!
text-path-caching

emielmolenaar:
Thanks for this fix! Nice work.

Clemweb:
There is a way to export this result in svg ? (textPath...)

asturur:
In the latest version of 5.x svg export is working and so in 6.x

Clemweb:
Hi @asturur , I'm not sur we speak about the same thing : I have Fabric 6.x (lastest version) and when I export my canvas to SVG curved texts are not curved but straight whithout any textPath specification... 
Is there something to activate it ?

Clemweb:
@asturur ok you can do it ?

asturur:
What i can do is polish and merge the PR i have open and that you should be able to see in the PR list.
Then if that is what you are looking for i do not know

Clemweb:
I see this PR : https://github.com/fabricjs/fabric.js/pull/10284 
Can you make a release with it ?

asturur:
I miss tests but yes, i need to write tests

PR #7300, opened 2021-08-14
draft of the function

asturur:
@melchiar this utility function would transform a path to a new path that unscaled/unrotate is equivalent to the one scaled/rotated.

I did not try it yet, i ll try to make a test tomorrow with text on a path or other paths effects.

### DIFF
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -780,6 +780,36 @@
     path.push(['L', p1.x + multSignX * correction, p1.y + multSignY * correction]);
     return path;
   }
+  /**
+   * Transform a path by transforming each segment.
+   * it has to be a simplified path or it won't work.
+   * WARNING: this depends from pathOffset for correct operation
+   * @param {Array} path fabricJS parsed path commands
+   * @param {Array} transform matrix that represent the transformation
+   * @param {Object} [pathOffset] the fabric.Path pathOffset
+   * @param {Number} pathOffset.x
+   * @param {Number} pathOffset.y
+   * @returns {Array} the transformed path
+   */
+  function transformPath(path, transform, pathOffset) {
+    if (pathOffset) {
+      transform = fabric.util.multiplyTransformMatrices(
+        transform,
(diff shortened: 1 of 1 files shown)


asturur:
The idea is that you do your transform operation as usual ( scale/ rotate / skew ).
instead of applying the transform to the text, you apply it to the path in the text.
Then you use this function to put the trasnformation in the path segments,
you swap the path segments, you calculate the bounding box again.
And you assign the path to the text. 
Eventually having a new path is better so that the text understand that a change happened.

For path on canvas, this code is going to change the path properties, but keeping a correct top/left is a different issue entirely.

melchiar:
Cool, thanks. I will test this evening. Also, what exactly is the advantage to recalculating the path coordinates rather than simply having text respond to a path's scale values?

asturur:
The difference is that the only way to do that is with uniform scaling, you can't really do anything else.
So you would scale the path, scale the path length values returned from the calculation, and you would be done.
But if you need to do anything else like:
- scaling from one side only
- skewing
- rotate
this is the only way to do it.

asturur:
search-gif

this is what it does, i m sure caching is doing something wrong and that hiccup can be fixed

asturur:
i did not write the custom controls but you can do this either with custom controls or with events.
The complexity i think is the same.

melchiar:
This is great! I'll need to play with it some more though before I've got it auto transforming in the scale event. Any chance you'd be able to throw your example into a fiddle?

melchiar:
K, I've got the resizing event working although there is the still caching problem to sort out. When all the pieces have come together I'll put together a demo for the website. Are you okay with merging this transformPath function or was this just meant as as an example?

textpath-resizing

asturur:
Sorry not following back on pulling out my example. The baby is taking all my time.
Yes this function is good. i have to update the changelog and the JSDOC slightly.
Ping me again if i don't do it today

melchiar:
> Sorry not following back on pulling out my example. The baby is taking all my time.

No worries, babies tend to do that 😂

PR #7317, opened 2021-08-20
feat(Observable): once event handler

ShaMan123:
This is very useful and saves repetitive `on` `off` logic for calling an event once and it aligns `Observable` with node's `EventEmitter`

### DIFF
--- a/src/mixins/observable.mixin.js
+++ b/src/mixins/observable.mixin.js
@@ -46,6 +46,27 @@
     return this;
   }
 
+  function _once(eventName, handler) {
+    var _handler = function () {
+      handler();
+      this.off(eventName, _handler);
+    }.bind(this);
+    this.on(eventName, _handler);
+  }
+
+  function once(eventName, handler) {
+    // one object with key/value pairs was passed
+    if (arguments.length === 1) {
+      for (var prop in eventName) {
+        _once.call(this, prop, eventName[prop]);
+      }
+    }
+    else {
+      _once.call(this, eventName, handler);
+    }
+    return this;
+  }
+
   /**
    * Stops event observing for a particular event handler. Calling this method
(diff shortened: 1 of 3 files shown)


src/mixins/observable.mixin.js
```diff
@@ -46,6 +46,27 @@
     return this;
   }
 
+  function _once(eventName, handler) {
+    var _handler = function () {
+      handler();
+      this.off(eventName, _handler);
```
asturur, line 52: I wonder how this work with hoisting.
We are referencing the function before we finish create it.
Is not a question about your code but more on javascript.

asturur:
This is good.

asturur (approved):

src/mixins/observable.mixin.js
```diff
@@ -46,6 +46,27 @@
     return this;
   }
 
+  function _once(eventName, handler) {
+    var _handler = function () {
+      handler();
```
asturur, line 51: @ShaMan123 in this case handler is not bound and is not getting the arguments.
I think this should  somethng like this:

```js
var _handler = function () {
   handler.apply(this, arguments)
   this.off(...)
}.bind(this)
```

handler needs to be bound when executed and needs to receive the arguments.
Can you verify and update the tests?

src/mixins/observable.mixin.js
ShaMan123: I guess I can call the function directly and not from the mixin

src/mixins/observable.mixin.js
ShaMan123: @ShaMan123
bind handler
c63e746

asturur:
i ll fix the semicolon and merge

asturur (approved):

PR #7318, opened 2021-08-21
fix(text.class.js) added pathStartOffset and pathSide to props lists in text.class.js

melchiar:
The new pathStartOffset and pathSide properties were missing from the toObject property list, the dimensionAffectingProps list, and the additionalProps list.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -12,7 +12,7 @@
 
   var additionalProps =
     ('fontFamily fontWeight fontSize text underline overline linethrough' +
-    ' textAlign fontStyle lineHeight textBackgroundColor charSpacing styles path').split(' ');
+    ' textAlign fontStyle lineHeight textBackgroundColor charSpacing styles path pathStartOffset pathSide').split(' ');
 
   /**
    * Text class
@@ -39,7 +39,9 @@
       'charSpacing',
       'textAlign',
       'styles',
-      'path'
+      'path',
+      'pathStartOffset',
+      'pathSide'
     ],
 
     /**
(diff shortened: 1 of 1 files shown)


src/shapes/text.class.js
```diff
@@ -1515,6 +1517,8 @@
         'charSpacing',
         'path',
         'direction',
+        'pathStartOffset',
+        'pathSide'
       ].concat(propertiesToInclude);
```
asturur, line 1522: i wonder if we can replace this custom array here with additionalProps higher up in the file.
Can you check if they match?

asturur:
Fixing tests is a pain here

src/shapes/text.class.js
melchiar: This file has 3 very similar property arrays (`additionalProperties`, `additionalProps`, and `_dimensionAffectingProps`) . All 3 share the following props:

path, fontSize, charSpacing, fontStyle, fontWeight, pathStartOffset, pathSide, lineHeight, fontFamily, textAlign, text

Here are the props that are not shared by every array:

**additionalProps:** 
textBackgroundColor, linethrough, overline, underline, styles

**additionalProperties (toObject method):** 
textBackgroundColor, linethrough, overline, underline, direction

**_dimensionAffectingProps:** 
styles

Perhaps we could simplify all three arrays by having a master array for the shared values and then doing a concat for the unique values?

asturur: `direction has to be added to additionalProps`, for sure. That invalidate caching between the things

asturur: `styles` would be extra in the internal toObject array, but is overridden a line after.

Please add direction to additionalProps, and delete additionalProperties, use additionalProp in its place.

PR #7358, opened 2021-09-07
feat(text.class.js) added pathAlign property for text on path

melchiar:
This adds an additional property `pathAlign` to the text class for setting the perpendicular alignment of each character on a path. It's basically the equivalent to the text path alignment settings in InDesign/Illustrator. Not sure how useful the `ascender` and `descender` options will be, but I expect `center` will see a lot of use.

image

image

image

image

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -200,12 +200,25 @@
     shadow:               null,
 
     /**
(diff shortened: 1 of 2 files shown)


melchiar:
ugh...this branch is messed up. I'm going to close this and try again.

PR #7362, opened 2021-09-08
feat(text.class.js) added pathAlign property for text on path

melchiar:
(note - this replaces #7358 which had branch conflicts)

This adds an additional property pathAlign to the text class for setting the perpendicular alignment of each character on a path. It's basically the equivalent to the text path alignment settings in InDesign/Illustrator. Not sure how useful the ascender and descender options will be, but I expect center will see a lot of use.

I've taken the approach of offsetting each grapheme's position based on the path's tangent rather than just modifying the canvas's `textbaseline` since the `textbaseline` prop gives inconsistent results across different browsers.

image

image

image

image

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -244,6 +244,7 @@
      * How text is aligned to the path. This property determines
      * the perpendicular position of each character relative to the path.
      * (one of "baseline", "center", "ascender", "descender")
+     * This feature is in BETA, and its behavior may change
      * @type String
      * @default
      */
@@ -563,7 +564,20 @@
      * @param {String} [charStyle.fontStyle] Font style (italic|normal)
      */
     _setTextStyles: function(ctx, charStyle, forMeasuring) {
-      ctx.textBaseline = 'alphabetic';
+      ctx.textBaseline = 'alphabetical';
+      if (this.path) {
+        switch (this.pathAlign) {
+          case 'center':
+            ctx.textBaseline = 'middle';
+            break;
+          case 'ascender':
+            ctx.textBaseline = 'top';
+            break;
+          case 'descender':
+            ctx.textBaseline = 'bottom';
+            break;
+        }
+      }
       ctx.font = this._getFontDeclaration(charStyle, forMeasuring);
     },
 
@@ -854,31 +868,13 @@
      */
     _setGraphemeOnPath: function(positionInPath, graphemeInfo, startingPoint) {
       var centerPosition = positionInPath + graphemeInfo.kernedWidth / 2,
-          path = this.path, offsetDist = 0;
+          path = this.path;
 
       // we are at currentPositionOnPath. we want to know what point on the path is.
       var info = fabric.util.getPointOnPath(path.path, centerPosition, path.segmentsInfo);
       graphemeInfo.renderLeft = info.x - startingPoint.x;
       graphemeInfo.renderTop = info.y - startingPoint.y;
       graphemeInfo.angle = info.angle + (this.pathSide ===  'right' ? Math.PI : 0);
-
-      switch (this.pathAlign) {
-        case 'center':
-          offsetDist = graphemeInfo.height / 4;
-          break;
-        case 'ascender':
-          offsetDist = graphemeInfo.height / 1.5;
-          break;
-        case 'descender':
-          offsetDist = graphemeInfo.height / -3.75;
-          break;
-      }
-
-      if (offsetDist) {
-        // offsets the grapheme in the direction perpenticular to the path's tangent
-        graphemeInfo.renderLeft -= fabric.util.sin(graphemeInfo.angle) * offsetDist;
-        graphemeInfo.renderTop += fabric.util.cos(graphemeInfo.angle) * offsetDist;
-      }
     },
 
     /**
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -247,7 +247,7 @@
      * @type String
      * @default
      */
-     pathAlign:               'baseline',
+    pathAlign:               'baseline',
 
     /**
      * @private
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -862,18 +862,16 @@
       graphemeInfo.renderTop = info.y - startingPoint.y;
       graphemeInfo.angle = info.angle + (this.pathSide ===  'right' ? Math.PI : 0);
 
-      if (this.type === 'text' && this.path) {
-        switch (this.pathAlign) {
-          case 'center':
-            offsetDist = graphemeInfo.height / 4;
-            break;
-          case 'ascender':
-            offsetDist = graphemeInfo.height / 1.5;
-            break;
-          case 'descender':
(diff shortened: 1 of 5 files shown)


src/shapes/text.class.js
```diff
@@ -239,6 +240,15 @@
...
     pathSide:               'left',
 
+    /**
+     * How text is aligned to the path. This property determines
+     * the perpendicular position of each character relative to the path.
+     * (one of "baseline", "center", "ascender", "descender")
+     * @type String
+     * @default
+     */
+    pathAlign:               'baseline',
```
asturur, line 250: So this is actually the baseline, not the pathAlign, sort of pathBaseline.
We do not have much to calculate ascender and descender.
Maybe advanced text metrics.

src/shapes/text.class.js
```diff
@@ -844,13 +854,34 @@
...
-          path = this.path;
+          path = this.path, offsetDist = 0;
 
       // we are at currentPositionOnPath. we want to know what point on the path is.
       var info = fabric.util.getPointOnPath(path.path, centerPosition, path.segmentsInfo);
       graphemeInfo.renderLeft = info.x - startingPoint.x;
       graphemeInfo.renderTop = info.y - startingPoint.y;
       graphemeInfo.angle = info.angle + (this.pathSide ===  'right' ? Math.PI : 0);
+
+      if (this.type === 'text' && this.path) {
```
asturur, line 865: why should we check if we have a path and if this is the function `setGraphemeOnPath`? I would also skip the check `this.type === 'text'` since we do not support itext-textbox for now because of editing but anyway this function is going to run anyway.

src/shapes/text.class.js
```diff
@@ -844,13 +854,34 @@
...
+        switch (this.pathAlign) {
+          case 'center':
+            offsetDist = graphemeInfo.height / 4;
+            break;
+          case 'ascender':
+            offsetDist = graphemeInfo.height / 1.5;
+            break;
+          case 'descender':
+            offsetDist = graphemeInfo.height / -3.75;
+            break;
```
asturur, line 875: all those magic numbers should be explained.

src/util/path.js
```diff
@@ -107,6 +107,20 @@
...
+   * @param {Number} distance the distance/length of the vector
+   * @return {Object.x} x coordinate of point
+   * @return {Object.y} y coordinate of point
+   */
+  function calcVectorPoint(angle, distance) {
+    var coords = {};
+    coords.x = fabric.util.cos(angle) * distance;
+    coords.y = fabric.util.sin(angle) * distance;
+    return coords;
+  }
```
asturur, line 122: i do not feel good about this function.
The name is not really descriptive, and for now is really used just in your specific use case, i can't see a clear reusable case for it.
i would prefer doing this directly where are you using vec.x and vec.y.

```
raphemeInfo.renderLeft -= fabric.util.cos(angle) * offsetDist
```

src/shapes/text.class.js
```diff
@@ -844,13 +854,34 @@
...
+        }
+      }
+
+      if (offsetDist) {
+        var angle = graphemeInfo.angle - Math.PI / 2;
```
asturur, line 880: Subtracting math.pi/2 does not explain why we are doing that, and can be probably be done swapping sin/cos between them ( and maybe changing the cos sign too )

asturur:
I'm happy to do this, but i m not sure if the baseline approach that you discarded is really worse.
some points:
- Magic numbers for fonts won't work across different font sizes
- descender/ascender can't be supported with magic numbers
- chinese/japanes will look weird because ascender / baseline/ descender are much different

I would ask you to check the following:
- how does it looks like with styles that change the font size across the path
- defaulting to using lineHeight for ascender/descender, and see how it looks like.

src/shapes/text.class.js
melchiar: The terms I used are actually taken straight from from Illustrator/InDesign.

image

In typography, the baseline is defined as "the line upon which most letters "sit" and below which descenders extend" (taken from [Wikipedia](https://en.wikipedia.org/wiki/Baseline_(typography)))

image


I'm not sure why canvas uses `textBaseline` as the general property name, but I personally feel like calling it the Baseline setting is a bit confusing. Thoughts?

src/shapes/text.class.js
melchiar: right!

src/util/path.js
melchiar: sure, I can change that

src/shapes/text.class.js
melchiar: you're right, that is much cleaner

src/util/path.js
melchiar: done

melchiar:
The magic numbers are essentially estimates to try to match the character positions you get when changing the `textbaseline` prop. Unfortunately the `alphabetic` option is the only one that gives consistent results across browsers. Here's a comparison of Chrome vs Firefox.

image

InDesign/Illustrator don't even try to hit the ascender/descender positions, instead just aligning with the center, top and bottom according to the font size.

image

melchiar:
Here are the current results for text with varied styles. Perhaps using lineHeight for the offset would be better.

image

melchiar:
and here are the results from using lineheight instead of grapheme height for the offset distance. I think this is the better approach since this way the baseline isn't shifting between each character

image

src/shapes/text.class.js
asturur: My only issue here is that each font family has those different. Especially the one that aren't latin fonts.
So baseline is the one we know and should work, then ewe can have descener and ascender using lineHeight, i m not sure we can get center working.

asturur: i m open to test this as it is, but i wonder if changing the real baseline at render time ( for path case only ) would yeld better results and valid across different fonts families.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/textBaseline

changing this at render time, would be exactly 1 line and no magic numbers. But i m not sure of which results would give.
The reason why we could use it in this particular case compared to normal rendering is because in this cases the bounding box is broken most of the time.

asturur:
if you can fix the tests, i m ok with this.
Decide by yourself if going the extra mile and testing ctx.textBaseline effects

src/shapes/text.class.js
melchiar: The textBaseline property does seem to give better results across fonts so if you're not bothered by the slight differences in browser rendering I'm fine with going that route instead.

melchiar: One problem though is that if we go with the textBaseline prop there's no way to use the lineheight for the offset distance since every character is aligned to its own vertical center. Having the baseline shift between characters of different styles seems odd to me, and doesn't match how InDesign/Illustrator treats text of different styles on a path.

image

I'm conflicted on this. 😐

asturur:
Sorry @melchiar i went rogue.
No one of the approaches will be perfect for various reasons.
Let's keep it as it is, let's mark it a beta/unstable so that if we find a better approach we can change it.
Let's fix the tests

asturur (approved):

PR #7419, opened 2021-10-05
feat(Observable): prepend

ShaMan123:
Added the ability to prepend event handlers in order to control an event's firing sequence.
```js
const handler = () => {};
canvas.on('foo', handler, true)
canvas.once('foo', handler, true)
```

### DIFF
--- a/src/mixins/observable.mixin.js
+++ b/src/mixins/observable.mixin.js
@@ -24,10 +24,11 @@
    * @alias on
    * @param {String|Object} eventName Event name (eg. 'after:render') or object with key/value pairs (eg. {'after:render': handler, 'selection:cleared': handler})
    * @param {Function} handler Function that receives a notification when an event of the specified type occurs
+   * @param {boolean} [prepend] whether to prepend the handler so it will be fired before the rest of the handlers
    * @return {Self} thisArg
    * @chainable
    */
-  function on(eventName, handler) {
+  function on(eventName, handler, prepend) {
     if (!this.__eventListeners) {
       this.__eventListeners = { };
     }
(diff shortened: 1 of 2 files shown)


ShaMan123:
Fr some disturbing reason I face a `ReferenceError: fabric is not defined` when I try to run the test locally.

asturur:
While this is very few code, i wonder, do other event systems do so?

ShaMan123:
Node.js has a `prependListner | prependOnceListener`:
https://nodejs.org/api/events.html#events_emitter_prependlistener_eventname_listener

ShaMan123:
This is important in cases when you want an object that is at the top of the stack (and therefore an object that will get handlers fired last) to be the first to handle an event and block other handlers from handling it by some kind of flagging (preventDefault).

ShaMan123:
Actually this doesn't entirely solve this issue.

> This is important in cases when you want an object that is at the top of the stack (and therefore an object that will get handlers fired last) to be the first to handle an event and block other handlers from handling it by some kind of flagging (preventDefault).

Take for example [_simpleEventHandler](https://github.com/fabricjs/fabric.js/blob/master/src/mixins/canvas_events.mixin.js#L493). Let's say I want to handle a drop event on multiple levels, canvas level being default behavior. When a drop event is fired I want targets to be able to handle the drop event and prevent the default behavior from executing. This can't be achieved unless I override `canvas._onDrop`.

A solution to this issue might be to make `Observable` scoped to canvas, so when objects use observable methods the handlers register on the canvas with a reference to the object.
What do you think?
I believe such use cases should be easier to handle.

PR #7434, opened 2021-10-15
fix: when style doesn't have anything to match, parser failed

k88manish:
I faced this issue recently when I used and SVG which had only google font imported in that case it failed to parse the style content and was not able to proceed any further.

### DIFF
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -1,7 +1,6 @@
 import * as fabric from 'fabric';
 import './styles.css';
 import { testCase } from './testcases/simpleTextbox';
-// import { testCase } from './testcases/polygons';
 
 const el = document.getElementById('canvas');
 const canvas = (window.canvas = new fabric.Canvas(el));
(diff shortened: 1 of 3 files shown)


asturur:
And does that SVG works in the browser today? what spec is that?

k88manish:
> And does that SVG works in the browser today? what spec is that?

This is trimmed down version of large SVG which got me this exception. But if you need the complete SVG file let me know

asturur:
my question is:
i can't find the import spec for svg in `www.w3.org`.
What kind of SVG is it? is it standard? Does it render in browser?

k88manish:
> 

Yes this SVG works in browser. 
I will try to find the spec but to be honest I am not very familiar with SVG specs

asturur:
Handling the @import statement inside svg should be the correct solution, but that may be hard. this solution could be backported on the current codebase

asturur:
I wanted to make the PR from your branch but is not possible

asturur:
@k88manish thank you for this fix and pr.
It has been merged as #9602

PR #7442, opened 2021-10-19
feat(): `drop:before` event

ShaMan123:


ShaMan123:
Another use case would be handling drop on objects in a different order than object stack order

asturur:
Sorry i m late on this.
Ok i m not super into this, but is not bad.
I will change the comment and merge it.

asturur:
that object you set the drop:before event on, could be the object that is in the `target` of the normal drop event.
I understand that sometimes we don't want to have big even thandlers that take care of everything in a single function.

Issue #7142, opened 2021-06-21
`absolutePositioned` with objects nested in groups


ShaMan123:
## Version
latest

## Test Case
http://jsfiddle.net/2185reh6/19/

## Information about environment
browser chrome

## Steps to reproduce
Add an `absolutePositioned` clip path to an object inside a group

## Expected Behavior
Clip path should be positioned relative to canvas, http://fabricjs.com/docs/fabric.Object.html#absolutePositioned

## Actual Behavior
Because group manages it's own plane the clip path is relative to the group

ShaMan123:
This is a part of the major problem that group creates on it's objects.
I think the solution should handle changing `calcTransformMatrix` so it can calculate the matrix  relative to the canvas as well with a flag.
Related #7130 #6776

ShaMan123:
I think the problem is that `calcTransformMatrix` is called from within a the clipPath. This call is unaware of the context it is running in, meaning it doesn't know it is a clip path transformed to a relative plane, therefore, it can't account for the transform matrix of the containing object.

i-prikot:
Hi. I think it works as intended.
http://jsfiddle.net/i_prikot/fgkyns8z/

ShaMan123:
http://fabricjs.com/docs/fabric.Object.html#absolutePositioned

asturur:
this is a though one.
Both use cases are fine, and discerning between the 2 with more options is not that much convenient.
The absolute positioned idea was to be able to have a clipPath on the canvas that would influence only an object.

It is working as intented from what i can see in the fiddle, but the point is that it does not invalidate the group cache on movement.

Not sure if this in an edge case that deserve fixing or not.
https://jsfiddle.net/qzo7r3y9/
you can also just set objectCaching: false on the group when you create it.

ShaMan123:
I found a work around
```js
            const __drawClipPath = fabric.Object.prototype._drawClipPath;
            fabric.util.object.extend(fabric.Object.prototype, {
                _drawClipPath: function (ctx: CanvasRenderingContext2D) {
                    __drawClipPath.call(this, ctx);
                    //  render stuff here
                }
            })
```

Now I am able to render many clip paths with varying `absolutePositioned` value.

ShaMan123:
fixed by the rewrite 
https://codesandbox.io/s/0d1pbh
Once caching is resolved this is gone for good

Issue #7130, opened 2021-06-18
Object geometry methods for grouped objects


ShaMan123:
## Version
latest

## Test Case
http://jsfiddle.net/mkdve1x9/15/

## Information about environment
Browser.

## Steps to reproduce
1. Create a `Group`/`ActiveSelection`
2. Try using a method from [`object_geomerty`](https://github.com/fabricjs/fabric.js/blob/2eabc92a3221dd628576b1bb029a5dc1156bdc06/src/mixins/object_geometry.mixin.js#L191) on an object that belongs to the group

## Expected Behavior
Should work for nested objects without additional code

## Actual Behavior
It returns wrong values because the object's transform matrix is used and is relative to the group

ShaMan123:
I suggest enhancing `calcTransformMatrix`.
https://github.com/fabricjs/fabric.js/blob/2eabc92a3221dd628576b1bb029a5dc1156bdc06/src/mixins/object_geometry.mixin.js#L584
```js
matrix = multiplyMatrices(this.group.calcTransformMatrix(true), matrix); 
```
This will open the door for many fixes.

asturur:
Yes.
The more i look at it and the more i think all the whole aCoord,oCoord is, excuse me the term, fucked up.
Fix it is a mess, it requires a full rewrite of so many internals and a ton of regression test/unit tests.

Ideally we want to get rid of oCoords entirely and all that absolute: true/false on those kind of geometry functions.
The only way to fix this is to add inside a 
```
if (this.group) {
// take group transform matrix
// multiply the points for group transform matrix
}
// proceed with normal checks
```

ShaMan123:
Man I absolutely agree.
You rock.

ShaMan123:
#7599

Issue #6776, opened 2021-01-03
Create ActiveSelection with objects in group


jameszhong2008:
I want to create a multiple selection from objects in group.
After create selection with ActiveSelection, bound box of the ActiveSelection is wrong. 
How to select multiple object within group or nested group?

check it here.
https://jsfiddle.net/zhongjian2008/49wdsg2q/

asturur:
http://fabricjs.com/manage-selection

Please first check if this answers your questions, and if it doesn't be specific on what.

septemberjustin:
@asturur I think James wants to select specific objects in a group, which means if there are 3 objects in a group. 
When he select 1 or 2 objects, or even 3 objects in this group. He got wrong bound box.

asturur:
Oh ok, that does not work, and is not in the planning for now.
It should't be hard to fix the bounding box to be honest, not hard at all.
And that would still help people that want to show as selected objects not directly by clicks but from some other kind of UI in an application.

septemberjustin:
Thanks Andrea, you are very responsible. 👍 
Yes, users need to select objects on the object tree of the UI, especially there are multiple layers and groups.

septemberjustin:
We really need this to work. (ʃƪˆ▿ˆ)

asturur:
This should be fine and useful for me too. I just wonder about controls. Controls are mostly useless inside a group. I wonder who is responsible for hiding/showing them. The  developer or an option in the library.

jameszhong2008:
@asturur @septemberjustin Thanks for your time and response, so you mean it should be implemented by developer rather than this library? 
ActiveSelection is convenient when we resize/move multiple objects together. So I want to know is there any chance that you can implement it within the library?(add multiple objects in a group or different group to ActiveSelection)

asturur:
Jameszhong, i do not get what are you looking for. I think the library should correctly draw the box for a select object inside a group.

what are you looking for?

septemberjustin:
@asturur James means the library should correctly draw the box and also show Controls for selected objects inside a group. 

Because when user drawing, they will group some objects, and they still need to adjust them inside a group. 
They don't want to ungroup them, then group them again. So, Controls are important inside a group. 

And users will also select multiple objects inside one group or between different groups. :) 

[...]

Issue #2073, opened 2015-03-31
Group Position Doesn't Retain


darkfrog26:
I am currently creating a group with an empty array (no children) to start with and I want it to be centered at 100x100.  Unfortunately when I set those values and then addWithUpdate the left and top seem to be getting reset to 0x0.
##

darkfrog26:
If, after adding the items, I then set left and top everything looks correct.

darkfrog26:
There also seems to be a glitch that causes neither the group nor items in the group to be selectable.

asturur:
there is an old discussion about this. mainly the group wrap himself around object positions. Only after the group is created you can force is position.

after setting the group top and left you have to call setCoords to make it selectable.

darkfrog26:
this seems really clunky to me and painful to use....are there any plans to change this?

asturur:
we have mayor bugs and features to solve and implement.

but we are always open to suggestions.
.setCoords is called manually because it is calculation intensive, so we do not want to call it automatically on every top or left modifications.

yeah grop is a little bit unconfortable to work with.

darkfrog26:
That's a shame. Most of the functionality of FabricJS I have been very impressed with. Group is the first major frustration point I've hit.

asturur:
I was thinking about this.
I cannot really see how to accomplish this feature.

I position a group at 50,50.
I add a circle that has top left of 60,60.
i add a rect that has top left  of 10,10

i want to keep the group at 50,50, where should rect and circle be?

darkfrog26:
@asturur sorry for the very late reply:

The absolute position within the canvas of circle and rect should be 110x110 and 60x60 respectively.  Perhaps I'm not fully understanding your question, but I don't see how adding anything to the group would change anything.

asturur:
so you mean translating the given top and left to add from the top left of the group. That can be done by code, but when you are grouping from graphic interface it looks very strange.

dmccuskey:
i just posted a new bug report which may be similar in nature: https://github.com/kangax/fabric.js/issues/3837

the current behavior seems broken when using group.add(). however, things seem to work correctly if objects are added in the group constructor. (i haven't tested extensively)

*basically, when an object is added to a new container, its position should be relative to that container's originX/Y (oX/oY).*

say we have a Rect with oX/oY Top Left with pos 10, 20 in canvas.

[...]

PR #7449, opened 2021-10-23
feat(Layer): Group Rewrite

ShaMan123:
### Motivation
#7316  #6776 #7136 #7299 #7130 #7142 #2073
Group is obsolete. We need something much better for fabric.
I think this is a successful predecessor that can replace Group entirely.

### Gist
I managed to come up with this solution thanks to this https://github.com/fabricjs/fabricjs.com/pull/315:
> Creating custom objects built from other objects is an extremely powerful option, but it takes you closer to fabric's edge.<br/>
You may experience problems. Let's use an example to explain.<br />
When an object (a) renders, it transforms the rendering context to it's center point.
In order to render another object (b) as a child of (a) we need to take that into account.
We can handle (b)'s transform matrix relative to (a)'s center point. This will render (b) correctly.<br />
However, it will cause many problems.<br />
Let's take, for example, one of the simplest and most powerful rendering optimizations to explain.<br />
Before an object renders it checks if it's visible on screen. If it isn't it skips the rendering process.
Checking if it's visible means the object checks it's transform matrix relative to the canvas.
But we've handled it's transform matrix relative to (a)....<br />
This will cause (b) to skip renders in some cases and not show at all.<br />
This gotcha affects almost everything fabric has to offer.<br />
The solution is simple.<br />
Handle the transform matrix relative to canvas, <strong>ALWAYS</strong>. When rendering (b), transform the rendering context from (a)'s center point back to canvas.<br />
For more information and detailed explanation, check the dedicated pages: <a href="/using-transformations">Using transformations</a>, <a href="/fabric-intro-part-3#subclassing">Subclassing</a>.

**This is why Layer objects are transformed relative to canvas**

### Impl

- In order to propagate transformation from Layer to it's objects I've extended `calcOwnMatrix`. The method adds the transform diff from the previous call to objects.
- In order to layout the Layer we need to disable transform propagation.

### To Do
- [x] object caching (copy from group?)
- [ ] performance - in case of a very large amount of objects (SprayBrush) + stress tests
- [ ] additional layout strategies + overflow (exposed public method `getLayoutStrategyResult`, need to support rotation)
- [ ] test svg import/export
- [x] make interactive!! selectable nested objects - Ohh yes it's coming.
- [ ] tests - cover all group cases, etc.


ShaMan123:
The latest commits fixed object stacking when selecting. Of course this needs to be tested with n>2 levels of nesting.
The major thing I can't get working is `getObjectsBoundingBox` when instance is rotated.
After that's solved I think this is mature to merge.

ShaMan123:
active selection tests need adjusting

ShaMan123:
Anyone interested in trying this out:
Modify [`onSelect` method of object](https://github.com/fabricjs/fabric.js/pull/7449/files#diff-2ae13b6c50b012abf8c7d30712d3c99c21005f724557d769ebead822ccec8dfd) and of layer to achieve different UX

ShaMan123:
closed by #7473

PR #7462, opened 2021-10-28
fix(fabric.StaticCanvas): error of changing read-only style field

gamtiq:
Calling `StaticCanvas.prototype.dispose` in IE causes "Assignment to read-only properties is not allowed in strict mode" error.
The fix eliminates this error.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1776,13 +1776,7 @@
       this.contextContainer = null;
       // restore canvas style
       this.lowerCanvasEl.classList.remove('lower-canvas');
-      // some browsers (IE) don't allow to redefine style
-      try {
-        this.lowerCanvasEl.style = this._originalCanvasStyle;
-      }
-      catch (err) {
-        fabric.util.setStyle(this.lowerCanvasEl, this._originalCanvasStyle);
-      }
+      fabric.util.setStyle(this.lowerCanvasEl, this._originalCanvasStyle);
       delete this._originalCanvasStyle;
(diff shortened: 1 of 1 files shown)


src/static_canvas.class.js
```diff
@@ -1776,7 +1776,13 @@
       this.contextContainer = null;
       // restore canvas style
       this.lowerCanvasEl.classList.remove('lower-canvas');
-      this.lowerCanvasEl.style = this._originalCanvasStyle;
+      // some browsers (IE) don't allow to redefine style
+      try {
+        this.lowerCanvasEl.style = this._originalCanvasStyle;
+      }
+      catch (err) {
+        fabric.util.setStyle(this.lowerCanvasEl, this._originalCanvasStyle);
```
asturur, line 1784: if this would work 100% of the time, why not using it as default?
dispose is not a fast operation, even if setStyle is for some reason more complex, is ok.

gamtiq: Because `setStyle` is more slow.
I suppose it should be used as a workaround for IE and similar browsers with restrictions.

asturur: this is a slow method, is fine to be always slow. please use setStyle 100% of the time

gamtiq: Updated.

asturur (approved):

asturur:
Just a clarification that while we are not removing this fix here, probably fabric 5.0 will remove specific workarounds or limitation for ie11. Seems about time.

Issue #7445, opened 2021-10-20
EraserBrush doubles the opacity of semi-transparent non-erasable shapes during free drawing


porterde:
## Version
4.6.0 with eraser module

## Test Case
[requires custom build of fabric with eraser brush]

## Information about environment
Browser

## Steps to reproduce
1. Create a custom build of fabric including EraserBrush
2. Add a non-erasable Rect with fill black and opacity 0.2
3. Set EraserBrush as the free drawing brush and enable isDrawingMode
4. When you press down the mouse button, the rectangle gets darker
5. You can "erase" the darkness, to reveal the lighter rectangle underneath
5. When you release the mouse button, the entire rectangle returns to normal

## Expected Behavior
Non-erasable shapes which are semi-transparent do not change during free drawing erasing
Semi-transparent shadows of non-erasable shapes should not change darkness during free drawing erasing

## Actual Behavior
Non-erasable shapes which are semi-transparent get darker during free drawing erasing
This also affects the semi-transparent shadows of non-erasable shapes

The logic in `renderBottomLayer` and `renderTopLayer` in https://github.com/fabricjs/fabric.js/blob/master/src/mixins/eraser_brush.mixin.js means non-erasable objects are rendered on both the top and bottom canvases, so the opacity of the semi-transparent areas is doubled.

For the special case where the non-erasable object has no erasable objects beneath it, I imagine this issue could be fixed by filtering out such objects during `renderTopLayer`. Is there a solution for the general case?

ShaMan123:
@porterde look at #7470

ShaMan123:
eraser v2 is now stable

ShaMan123:
The eraser has moved https://github.com/ShaMan123/erase2d

Issue #7454, opened 2021-10-25
Group With ClipPath and Text with Shadow Renders Incorrectly


edcurtis:
## Test Case
https://codepen.io/edcurtis/pen/dyzvdJj

## Information about environment
4.6.0 in Chrome

## Steps to reproduce
Create a group that is clipped via ClipPath and contains Text that has a Shadow with an offset. The rendering messes up and hides the entire canvas background.  It is ok if the shadow offset is 0.  You can see in my example/test case that the canvas should be red, but is showing up white.  Change the offsetX and offsetY to 0 and it will render correctly as red.

ShaMan123:
I think this is an edge case caused by conflicts between group, caching and shadow.
Tested with #7473 (still WIP), yields good results.
https://codesandbox.io/s/dq826
So I think it proves my assumption on this issue.

ShaMan123:
After discussing this issue with @asturur I now understand it's complexity.
Shadow cannot be cached because once the object is rotated the shadow will remain static and cast at a wrong angle.
This means that setting a clip path on a group and a shadow on a child object will cause this bug.
I have in mind a solution. It's similar to one I have implemented for the eraser.
Basically it propagates the clip path to all children.
Then we can mitigate caching by separating objects into sub group.
This sounds to me like a robust solution.

PR #7470, opened 2021-11-02
feat(EraserBrush): V2

ShaMan123:
### Motivation
Solves eraser problem (opaic objects) by applying a mask over the brush.
As a side effect solved other problems such as deep clipping and erasing.
And enabled the awesome feature of undoing erasing.
While boosting performance so it seems.

### Previous Drawing Logic
Render objects in 3 steps:
1. All non-erasables on bottom on main/bottom context
2. All erasables on top context
3. While erasing use the path to clip step 2.
4. Non-erasable overlay

This was great except when you had a non erasable opaic object. It would draw on step 1 and 2, doubling itself.
Trying to solve it in context would not work, somewhere it would still be doubled or missing.

### Current Drawing Logic
While erasing:
1. clip main context with path
2. render path with a mask on top context. Mask is produced by rendering all non-erasable objects.

This approach supports `inverted` erasing, AKA recovering, out of the box. We simply produce a mask rendered with all erasable objects with their eraser (clip path) inverted.
Use this feature as follows:
```
brush.inverted = true;
```
### Outcome
1. Better performance
2. #7445 is solved
3. animations while erasing - if object is erasable it produces a funny output (dev should handle it making the animated objects non-erasable: once #7528 is merged the brush can handle this case as well by default)
4. Supports/solves #5187 #7271 

### Try it out
https://codesandbox.io/s/2hhyp
https://codesandbox.io/s/soz8h (tests erasing with inverted and nested clip paths - erase and select objects to see)
The left top circle is opaic and not erasable.

### Migration from V1
The Eraser object is now a subclass of Group. This means that loading from JSON will break between versions. 
Use this [code](https://gist.github.com/ShaMan123/6c5c4ca2cc720a2700848a2deb6addcd) to transform your json payload to the new version.

### Perf

pff

### Next
I've pushed the use of clipping beyond the limit of `svg`. It seems fabric has advanced further than `svg` in clipping abilities so `toSVG` should be amended to using masks, not clip paths in order to support all functionalities.


src/brushes/pencil_brush.class.js
```diff
@@ -168,12 +168,13 @@
     /**
      * Draw a smooth path on the topCanvas using quadraticCurveTo
      * @private
+     * @param {CanvasRenderingContext2D} [ctx]
```
asturur, line 171: those changes here on top are fine for me.

ShaMan123:
I want a way to check performance compared to prev version.
Need ideas

asturur:
You can use the inspector of google chrome.
It will give you scripting time and FPS on the canvas.

src/mixins/eraser_brush.mixin.js
```diff
@@ -357,192 +326,95 @@
...
-      renderOverlay: function () {
-        this.prepareCanvasForLayer('overlay');
-        var canvas = this.canvas;
-        var ctx = canvas.contextTop;
-        canvas._renderOverlay(ctx);
-        this.restoreCanvasFromLayer('overlay');
+      _setBrushStyles: function (ctx) {
+        ctx = ctx || this.canvas.contextTop;
+        this.callSuper('_setBrushStyles', ctx);
+        ctx.strokeStyle = null;
```
ShaMan123, line 417: should be
```diff
- ctx.strokeStyle = null;
+ ctx.strokeStyle = 'transparent';
```

but for some strange reason group goes crazy after erasing.
related #7454 ?

src/mixins/eraser_brush.mixin.js
```diff
@@ -357,192 +326,95 @@
...
        */
-      restoreCanvasObjectsFromLayer: function (layer) {
-        if (layer !== 'bottom') { return; }
-        this.restoreCollectionTraversal(this.canvas);
+      preparePattern: function (ctx) {
+        if (!this._patternCanvas) {
+          this._patternCanvas = fabric.util.createCanvasElement();
+        }
+        var canvas = this._patternCanvas;
+        canvas.width = this.canvas.width;
```
ShaMan123, line 373: @asturur I think this canvas needs retina scaling.
Tried it but couldn't get the correct size.
Can you help?

GavinLi2016:
ShaMan123, i found after erased an image. this image can not use clipPath to clip any polygon again.
can you take a see if it is my mistake. thanks
https://codepen.io/gavinli2016/project/editor/AnrdOk

ShaMan123:
@GavinLi2016 I don't understand what you are trying to achieve, you should try to explain your goal.
There is a working sandbox with a link in the first comment.

ShaMan123:
If I'm taking eraser to v2 then I want to separate it from clip path.
It was bad design to attach to clip path but at the time it was too much to get it working standalone.
So I will open a dedicated PR.
I will be breaking for sure.

ShaMan123:
I merged #7585 into this branch because I wanted to finalize some common stuff.
I necessary we can revert the merge so it an be easier to understand.
Did some profiling with chrome inspector.
V2 is winning across the charts, so it seems.
I have no idea how to export the charts...
pff

src/mixins/eraser_brush.mixin.js
ShaMan123: If I understand correctly I should apply retina scaling to the pattern canvas and then scale top context by `1/retinaScaling`.
However this is an edge case fabric doesn't handle well. Events get transformed as well by this transformation, so it sounds bad to do this.
Ideas?

ShaMan123: look at the text being erased.
text is not erasable so what you see is the pattern drawing over it.
because retina scaling is accounted for it is blurred
https://user-images.githubusercontent.com/34343793/148683724-a60fff60-d40d-4978-8576-89fcac511498.mp4

src/mixins/eraser_brush.mixin.js
ShaMan123: I am not sure the above is correct.
There is an undesirable visual effect happening when erasing a non-erasable (pattern draws over it) I want to solve.
It appears to draw a very thin border around the path drawn by the brush.

ShaMan123: I think it's a side effect caused by the absence of retina scaling

GavinLi2016:
> @GavinLi2016 I don't understand what you are trying to achieve, you should try to explain your goal. There is a working sandbox with a link in the first comment.

sorry to reply late. I just want to erase one image with eraserbrush and shape paths both. but it didn't work as the codepen link https://codepen.io/gavinli2016/project/editor/AnrdOk
I found when using eraserbrush, the clipPath only accepted dot line paths, any other shape paths(for example a rectangle) added to the clipPath not effective

ShaMan123:
Your explanation is still unclear so I can't understand.
But by guessing I think you want to clip the image with a rect and have it erased as well.
If that's the case this PR solves it because the eraser property has been moved away from the clip path property.
Simply set the clip path to the rect and it should work

GavinLi2016:
> Your explanation is still unclear so I can't understand. But by guessing I think you want to clip the image with a rect and have it erased as well. If that's the case this PR solves it because the eraser property has been moved away from the clip path property. Simply set the clip path to the rect and it should work. **Be sure to update the dist file in your repro**

yes, I want to inverted clip the image and also can erase it as well. I try to use the updated dist file again. thanks

ShaMan123:
All right! Happy to say I've managed to resolve the last bit that was bothering me (retina scaling).
**So  this PR is ready to be merged.**
It is excellent if you don't mind me saying.
I think we should add test, but because it isn't part of the build they won't pass. Anyways that's another PR.

ShaMan123:
All right! Happy to say I've managed to resolve the last bit that was bothering me (retina scaling).
**So  this PR is ready to be merged.**
It is excellent if you don't mind me saying.
I think we should add a test, but because it isn't part of the build it won't pass. Anyways that's another PR.

ShaMan123:
All right! Happy to say I've managed to resolve the last bit that was bothering me (retina scaling and a small glitch with the inverted brush).
**So  this PR is ready to be merged.**
It is excellent if you don't mind me saying (only thing I think of fixing is a very thin outline of the eraser revealing context behind what is erased, but is **extremely** minor).
I think we should add a test, but because it isn't part of the build it won't pass. Anyways that's another PR.

src/shapes/object.class.js
```diff
@@ -1225,13 +1225,13 @@
...
-      this._drawClipPath(ctx);
+      this._drawClipPath(ctx, this.clipPath);
       this.fill = originalFill;
       this.stroke = originalStroke;
     },
 
-    _drawClipPath: function(ctx) {
-      var path = this.clipPath;
+    _drawClipPath: function (ctx, path) {
+      path = path || this.clipPath;
```
asturur, line 1234: since we are going for 5.0 and the _drawClipPath calls aren't many, please make path a required argument and let's skip the  `path = path || this.clipPath;`

asturur: same for `drawClipPathOnCache`

src/util/misc.js
```diff
@@ -580,6 +580,25 @@
...
 
+    /**
+     * Creates corresponding fabric instances residing in an object, e.g. `clipPath`
+     * @see {@link fabric.Object.ENLIVEN_PROPS}
+     * @param {Object} object
+     * @param {Object} [context] assign enlived props to this object (pass null to skip this)
+     * @param {(objects:fabric.Object[]) => void} callback
+     */
+    enlivenObjectEnlivables: function (object, context, callback) {
+      var enlivenProps = fabric.Object.ENLIVEN_PROPS.filter(function (key) { return !!object[key]; });
```
asturur, line 591: i can't find this ENLIVEN_PROPS

asturur:
So this should go in first?
Can you show me where the ENLIVEN_PROPS are? i can't find them by searching.

src/brushes/base_brush.class.js
```diff
@@ -68,9 +68,10 @@ fabric.BaseBrush = fabric.util.createClass(/** @lends fabric.BaseBrush.prototype
   /**
    * Sets brush styles
    * @private
+   * @param {CanvasRenderingContext2D} [ctx]
    */
-  _setBrushStyles: function() {
-    var ctx = this.canvas.contextTop;
+  _setBrushStyles: function (ctx) {
+    ctx = ctx || this.canvas.contextTop;
```
asturur, line 74: we are doing for 5.0 and setBrush style is internal, it is fine to check where this is called all over fabric an just update it to be used with the ctx parameter.

src/shapes/object.class.js
```diff
@@ -1958,6 +1958,15 @@
...
   fabric.Object.NUM_FRACTION_DIGITS = 2;
 
+  /**
+   * Defines which properties should be enlivened from the object passed to {@link fabric.Object._fromObject}
+   * @static
+   * @memberOf fabric.Object
+   * @constant
+   * @type string[]
+   */
+  fabric.Object.ENLIVEN_PROPS = ['clipPath'];
```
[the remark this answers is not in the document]
asturur: well i used the search function extensively and nothing it wasn't coming out

src/shapes/object.class.js
ShaMan123: 6a73507a

src/brushes/base_brush.class.js
ShaMan123: 3ad6470d

src/brushes/pattern_brush.class.js
```diff
@@ -29,17 +29,19 @@ fabric.PatternBrush = fabric.util.createClass(fabric.PencilBrush, /** @lends fab
 
   /**
    * Creates "pattern" instance property
+   * @param {CanvasRenderingContext2D} ctx
    */
-  getPattern: function() {
```
ShaMan123, line 33: BREAKING

ShaMan123: but is the right approach

src/brushes/pencil_brush.class.js
```diff
@@ -168,12 +168,13 @@
...
      */
-    _render: function() {
-      var ctx  = this.canvas.contextTop, i, len,
+    _render: function(ctx) {
+      var i, len,
           p1 = this._points[0],
           p2 = this._points[1];
-
+      ctx = ctx || this.canvas.contextTop;
```
ShaMan123, line 177: What about `_render`?
```diff
- _render() {

+ _render(ctx) {
```

ShaMan123:
Looking Good

ShaMan123:
ready to merge

asturur:
ok let's see what happens

asturur (approved):

ShaMan123:
What about changelog?

ShaMan123:
CHANGELOG DONE

asturur (approved):

asturur:
Can you fix the dist folder? i tried but i can't fixing it by check out

ShaMan123:
image

asturur:
try from commandline with the classing `git checkout master dist` ?

ShaMan123:
DONE!

asturur (approved):

asturur:
do we need to merge the second one now?

ShaMan123:
yes.
it's supposed have no diff because I merged it into this one.

ShaMan123:
Well no. I'll shut it down

komagic:
bad case， i will show how to solve it

ShaMan123:
@komagic if you've encountered a bug do open a bug report. There's no way to understand what you are experiencing.
And unfortunately setting `globalCompositionOperation` does not suffice.

komagic:
https://www.cutout.pro/remove-background/upload

i just  discuss the main solution , dont want do anything.

ShaMan123:
> @komagic if you've encountered a bug do open a bug report. There's no way to understand what you are experiencing. And unfortunately setting `globalCompositionOperation` does not suffice.

my comment still stands.

ShaMan123:
The eraser has moved https://github.com/ShaMan123/erase2d

PR #7473, opened 2021-11-04
feat(ICollection/Layer): Group Rewrite

asturur:
### Motivation
#7316  #6776 #7136 #7299 #7130 #7142
Group is obsolete. We need something much better for fabric.
I think this is a successful predecessor that can replace Group entirely.
This PR aims first and foremost to enable/support nested selection under an Object.

### Gist
I managed to come up with this solution thanks to this https://github.com/fabricjs/fabricjs.com/pull/315:
> Creating custom objects built from other objects is an extremely powerful option, but it takes you closer to fabric's edge.<br/>
You may experience problems. Let's use an example to explain.<br />
When an object (a) renders, it transforms the rendering context to it's center point.
In order to render another object (b) as a child of (a) we need to take that into account.
We can handle (b)'s transform matrix relative to (a)'s center point. This will render (b) correctly.<br />
However, it will cause many problems.<br />
Let's take, for example, one of the simplest and most powerful rendering optimizations to explain.<br />
Before an object renders it checks if it's visible on screen. If it isn't it skips the rendering process.
Checking if it's visible means the object checks it's transform matrix relative to the canvas.
But we've handled it's transform matrix relative to (a)....<br />
This will cause (b) to skip renders in some cases and not show at all.<br />
This gotcha affects almost everything fabric has to offer.<br />
The solution is simple.<br />
Handle the transform matrix relative to canvas, <strong>ALWAYS</strong>. When rendering (b), transform the rendering context from (a)'s center point back to canvas.<br />
For more information and detailed explanation, check the dedicated pages: <a href="/using-transformations">Using transformations</a>, <a href="/fabric-intro-part-3#subclassing">Subclassing</a>.

**This is why Layer objects are transformed relative to canvas**

### Impl

- Expose a method that adds the transform diff from the previous call to objects.
- In order to layout we need to disable transform propagation.

### To Do
- [x] object caching (copy from group?)
- [x] performance - in case of a very large amount of objects (SprayBrush)
- [ ] `objectCaching=false`: need a better hook for `_applyMatrixDiff`
- [x] additional layout strategies (exposed public method `getLayoutStrategyResult`)
- [ ] `getLayoutStrategyResult` with clip path?
- [ ] support rotation
- [ ] overflow - clipping during selection??
- [x] test svg import/export
- [ ] svg import: handle object doubled opacity, handle `layout` prop, transform, subTargetCheck
- [x] deep interactivity.
- [x] tests - cover all group cases, stress tests, etc.
- [ ] active selection cases

### Additional
- [ ] migrate `SprayBrush` from group?


ShaMan123:
Changed `_applyMatrixDiff` hook to `render` and `setCoords`. This introduced a problem with object caching.
Hooking to `render` is OK, hooking to `setCoords` only is not good because it renders a frame+ late

ShaMan123:
Last commits optimized layer in such a way I think it is now more performant than group.
Instead of rendering all the time after applying the matrix diff it does the following:
1. keep objects up to date from `setCoords`
2. renders once an object is selected/deselected

ShaMan123:
I have come across another issue demonstrated by the following screencast.
IMO `_checkTarget` should:
1. not return true for objects that aren't selectable (not a big deal)
2. not return true if pointer is outside the parent's clipped area
ezgif com-gif-maker (3)

ShaMan123:
Today's commits:
1. Rename to `ICollection`
2. Create `Layer` subclass (it is the conventional layer as we discussed, spreading over the entire canvas)

I am pleased with this PR. It's good.
@asturur  I need your help with these:
1. bbox when ICollection is rotated
2. finding the performance hit: https://codesandbox.io/s/l327k, follow the instruction in the comments

src/shapes/icollection.class.js
```diff
@@ -0,0 +1,569 @@
...
+        options = clone(object, true);
+    delete options.objects;
+    if (typeof objects === 'string') {
+      // it has to be a url or something went wrong.
+      fabric.loadSVGFromURL(objects, function (elements) {
+        var group = fabric.util.groupSVGElements(elements, object, objects);
+        group.set(options);
+        group._restoreObjectsState();
+        callback && callback(group._objects, options);
+      });
```
ShaMan123, line 543: importing from SVG need to be solved.

src/shapes/icollection.class.js
```diff
@@ -0,0 +1,569 @@
...
+            width = (bounds.max.x - bounds.min.x) / (this.scaleX || 1),
+            height = (bounds.max.y - bounds.min.y) / (this.scaleY || 1),
+            rad = fabric.util.degreesToRadians(this.angle || 0),
+            cos = Math.abs(Math.cos(rad)),
+            sin = Math.abs(Math.sin(rad));
+        return {
+          left: center.x,
+          top: center.y,
+          width: width * cos + height * sin,
+          height: width * sin + height * cos,
```
ShaMan123, line 404: I don't know how to calculate the bbox correctly if instance is rotated.

src/shapes/icollection.class.js
```diff
@@ -0,0 +1,569 @@
...
+       *
+       * @param {object} opt
+       * @param {fabric.Object[]} opt.subTargets
+       * @returns true to abort selection, a `subTarget` to select that or false to defer to default behavior and allow selection to take place
+       */
+      onSelect: function (opt) {
+        return this.callSuper('onSelect', opt) || (opt.subTargets && opt.subTargets.length > 0 && opt.subTargets[0]);
+      },
+
+      isCacheDirty: function (skipCanvas) {
```
ShaMan123, line 265: Is this correct?
I used `Group` logic.

ShaMan123 (changes requested):
@asturur I've commented some things I need help with

ShaMan123:
test 8a9142e: https://codesandbox.io/s/v37mr

src/parser.js
```diff
@@ -29,6 +29,7 @@
         display:              'visible',
         visibility:           'visible',
         transform:            'transformMatrix',
+        'fabric-layout':      'layout',
```
asturur, line 32: what is this?

asturur:
i m starting to review and taking notes. Can you leave the code steady for a bit?

ShaMan123:
Yes.

בתאריך יום ה׳, 2 בדצמ׳ 2021, 13:52, מאת Andrea Bogazzi ‏<
***@***.***>:

ShaMan123:
I will work with PRs and merge into this one so you won't get overwhelmed with commits or changes while trying to review this.
Next up #7529

src/canvas.class.js
```diff
@@ -1138,18 +1147,71 @@
      * @param {Event} [e] Event (passed along when firing "object:selected")
      * @return {Boolean} true if the selection happened
      */
-    _setActiveObject: function(object, e) {
-      if (this._activeObject === object) {
+    _setActiveObject: function (object, e) {
+      var isCollection = Array.isArray(object._objects), activeObject = this._activeObject;
```
ebigtm, line 1151: For fixing the drag of multiple object isCollection should be fixed  and line 1151 should be

`var isCollection = (Array.isArray(object._objects) &&  object.type != 'activeSelection'), activeObject = this._activeObject;`

ebigtm (changes requested):

src/canvas.class.js
ShaMan123: Thanks! I'll look into it.
Just wanted to make sure you've tested this with #7529 merged into your local branch.

ShaMan123: I didn't test it but I have a few thoughts.
If I understand correctly if we accept your proposal selecting objects nested in active selection will become impossible.
Can you confirm?
If so this case should be handled by the dev overriding `onSelect` of active selection.

asturur:
https://codesandbox.io/s/yiliv

https://codesandbox.io/s/fabric-app-forked-cvewn

src/canvas.class.js
ebigtm: for me its working as it should selecting in icollection when its there but doesnt cancel selection for active selection, tested on #7529 merged branch.

ShaMan123:
> after the fix i presented yesterday a new issue popped, after dragging the objects it casts ray on the object the mouse up was on and selects it, the fix i found is a bit rough: `var result = object.type == 'iCollection' ? object.onSelect({ e: e, object: activeObject, subTargets: this.targets.filter(function (object) { return object.isSelectable(); }), activeSubTargets: object._objects.length > 0 ? true : subTargets && subTargets.filter(function (object) { return object.isSelectable(); }) }) : object.onSelect({ e: e, object: activeObject, });`
> 
> maybe it can be done differntly but generally i added a condition regarding the type, maybe we can use isCollection var instead?
> 
> https://github.com/fabricjs/fabric.js/blob/5571a3a597abae989ec872efb64583afa3e978f6/src/canvas.class.js#L1192-L1197

ShaMan123:
closing this in favor of #7669 
There's still selection logic that might be useful

PR #7482, opened 2021-11-10
minor empty check fix in getCSSRules

ijaz-saeed:
files with following style tags are not processed, needs this empty or undefined check. 

 **<style type="text/css">@import url ...**

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1001,7 +1001,7 @@
         if (styleContents.trim() === '') {
           continue;
         }
-        rules = styleContents.match(/[^{]*\{[\s\S]*?\}/g);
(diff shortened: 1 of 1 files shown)


asturur:
Can you tell me more about those style tags?
Can you reference me some kind of spec, i never seen those.

asturur:
PR looks fine.

ijaz-saeed:
> Can you tell me more about those style tags? Can you reference me some kind of spec, i never seen those.

just noticed , the similar issue is reported by someone else , https://github.com/fabricjs/fabric.js/pull/7434

asturur:
closing this since we changed the parser and there are no regex anymore.

PR #7497, opened 2021-11-18
Fix(fabric.Text): Avoid reiterating measurements when width is 0 and measure also empty lines for consistency.

avra-m3:
Very small PR, I noticed we were hitting measure text more than expected and found this to be the cause.

Happy for this to be declined If I've misunderstood these checks :)

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1347,15 +1347,8 @@
         return this.__lineWidths[lineIndex];
       }
 
-      var width, line = this._textLines[lineIndex], lineInfo;
-
-      if (!line || line.join('') === '') {
-        width = 0;
-      }
-      else {
-        lineInfo = this.measureLine(lineIndex);
-        width = lineInfo.width;
-      }
+      var lineInfo = this.measureLine(lineIndex);
+      var width = lineInfo.width;
       this.__lineWidths[lineIndex] = width;
       return width;
     },
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1343,13 +1343,13 @@
      * @return {Number} Line width
      */
     getLineWidth: function(lineIndex) {
-      if (this.__lineWidths[lineIndex]) {
+      if (this.__lineWidths[lineIndex] !== undefined) {
         return this.__lineWidths[lineIndex];
       }
 
       var width, line = this._textLines[lineIndex], lineInfo;
 
-      if (line === '') {
+      if (!line || line.join('') === '') {
         width = 0;
       }
       else {


src/shapes/text.class.js
```diff
@@ -1343,13 +1343,13 @@
      * @return {Number} Line width
      */
     getLineWidth: function(lineIndex) {
-      if (this.__lineWidths[lineIndex]) {
+      if (this.__lineWidths[lineIndex] === undefined) {
```
Hammer-Inc, line 1346: When the line is width 0, this cache check incorrectly fell through

src/shapes/text.class.js
```diff
@@ -1343,13 +1343,13 @@
...
         return this.__lineWidths[lineIndex];
       }
 
       var width, line = this._textLines[lineIndex], lineInfo;
 
-      if (line === '') {
+      if (!line || line.join('') === '') {
```
Hammer-Inc, line 1352: this.textlines is actualy string[][] so line is string[], line === '' is never true

src/shapes/text.class.js
```diff
@@ -1343,13 +1343,13 @@
      * @return {Number} Line width
      */
     getLineWidth: function(lineIndex) {
-      if (this.__lineWidths[lineIndex]) {
+      if (this.__lineWidths[lineIndex] !== undefined) {
```
Hammer-Inc, line 1346: Avoid causing a cache miss when the cached value is zero

Hammer-Inc:
Looking into this further, fabric is actually dependent on the line === '' check failing as it doesn't create the charBounds value otherwise

src/shapes/text.class.js
asturur: this is fine

src/shapes/text.class.js
```diff
@@ -1343,19 +1343,12 @@
...
-      var width, line = this._textLines[lineIndex], lineInfo;
-
-      if (line === '') {
-        width = 0;
-      }
-      else {
-        lineInfo = this.measureLine(lineIndex);
-        width = lineInfo.width;
-      }
+      var lineInfo = this.measureLine(lineIndex);
```
asturur, line 1350: does measureLine have the same check inside?

Hammer-Inc: Not as far as I have seen.

Hammer-Inc:
> Run ChristiaanScheermeijer/jest-reporter-action@v0.4.0
> Error: Resource not accessible by integration
How can I fix this test failure?

src/shapes/text.class.js
Hammer-Inc: When I stepped through line by line, I got as far as this._measureChar with an empty string.

asturur: So why we want to skip the shortcut and go far down with the empty line?

asturur: I m happy about the fix, i want to understand why we also remove the shortcut. Is not of immediate understanding to me.

Hammer-Inc: Originally I fixed the shortcut, it caused an error when a line consisted solely of a blank space because `_charBounds[x]` is undefined.

My understanding is that `this._measureLine` sets `_charBounds` for each line so we shouldn't skip it otherwise other parts of the code will fail :(.

We could restore the check but currently it's the equivalent of `if(false)` because `['']` will never equal `''`, 

Fixing it to act as a shortcut is probably possible, I'd need to investigate further and may have consequences when it comes to how different platforms measure blank newlines?.

asturur:
that failure is just the coverage, i have no idea how to make it work on forks.

src/shapes/text.class.js
asturur: no is ok as it is.
The fact that skipping measurements was creating error elsewhere is good to me.

asturur (approved):

Issue #5275, opened 2018-09-28
fix(geometry): object bounding box, clip path taken into account


blobinabottle:
Hello, clipPath seems working pretty well (in canvas, I didn't test toSVG yet). But I think we should improve the way controls are displayed. For the moment it's like this:
![image](https://user-images.githubusercontent.com/8090162/46200839-f90ce680-c312-11e8-8b52-c52eba203ee5.png)

So boundaries are based on all objects inside the clipPath. 

But I think it makes more sense to only consider the clipPath for the boundaries:
![image](https://user-images.githubusercontent.com/8090162/46200761-c7941b00-c312-11e8-91a8-4efcde7e76f6.png)

But maybe I'm missing something.

rikkarv:
Perhaps, when using "absolutePositioned", the present implementation makes more sense. 

For example the clipping area is positioned in top left of the canvas, however the clipped object is still in the center of canvas and i want to move it towards the clipping area. How could i drag him if the controls are in the clipping area?

@blobinabottle good point to bring this to discussion

rikkarv:
Please look at issue #5266 and http://jsfiddle.net/utryjax4/85/ where we have a situation in which is needed controls to the full clipped object and controls to other object that marks the clipping area

asturur:
there should be some logic that would be able to get the intersection between the 2 boxes.

The width and height is not the one or the other but is where they overlap.

While this could be still be done drawing the controls, i have no idea what happens when you want to scale something with this logic

blobinabottle:
maybe using sectable true/false on clipping area and/or content?

rikkarv:
Look at this [crazy implementation](https://jsfiddle.net/rikkarv/kanejp47/4/) inspired by @releasecreative issue #5266 and @TsuchiyaMasahiro idea of setting a interactive clipping fabric object.

It is possible to interact in many ideas with the clipping area (scaling, rotating, moving, stretching) on the second canvas.

It seems a little off-topic here but i think it's connected to @asturur reference to "_i have no idea what happens when you want to scale something with this logic_". Couldn't it be a workaround?

ShaMan123:
This has been solved by #7858 
Make sure the clipped object is a group and set `layout: 'clip-path'`
#7868

ierover:
@ShaMan123 It doesn't work in fabric v5.3
you can try here:
https://jsfiddle.net/1yjvLhbe/

ShaMan123:
[...]

PR #7507, opened 2021-11-20
fix(geometry): bbox of clipped object

ShaMan123:
Motivation #5275 #7473 

What should happen when an object is clipped?
Should it be bound to the clipped area?
If the clip path isn't centered what then?
Should originX/Y be relative to the clip path center?

Test
https://codesandbox.io/s/wpu1g

### DIFF
--- a/dist/fabric.js
+++ b/dist/fabric.js
@@ -1,4 +1,4 @@
-/* build: `node build.js modules=ALL exclude=gestures,accessors,erasing requirejs minifier=uglifyjs` */
+/* build: `node build.js modules=ALL requirejs minifier=uglifyjs` */
 /*! Fabric.js Copyright 2008-2015, Printio (Juriy Zaytsev, Maxim Chernyak) */
 
 var fabric = fabric || { version: '4.6.0' };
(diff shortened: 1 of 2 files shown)


asturur:
the issue here is that:
- clipPath can be larger than the object
- clipPath can be scaled independently of the object and break calculations
- clipPath can be positioned in a super weird way so that its dimensions combined with left/top of the object would be mess up.

I wonder if a possible solution is to separate the code that renders controls/bounding box from the code that is used for rendering and calculations of other kinds.
I think this _getTransformedDimension is shared between the 2.

ShaMan123:
It seems that [`drawBorders`](https://github.com/fabricjs/fabric.js/blob/8ecbdb10f797ce07fb4dccca348fe63ff1558b62/src/mixins/object_interactivity.mixin.js#L124) relies on [`_calculateCurrentDimensions`](https://github.com/fabricjs/fabric.js/blob/8ecbdb10f797ce07fb4dccca348fe63ff1558b62/src/mixins/object_geometry.mixin.js#L707)
`drawControls` relies on `oCoords` that in turn relies on `_calculateCurrentDimensions`.
So it seems `_calculateCurrentDimensions` is used by all of them.

ShaMan123:
In think this PR is a great example that demonstrates the need of a coherent object API.
This way with a simple override of getDimensions or what ever I can handle special cases myself.

ShaMan123:
This can die out because of group's layout engine

Issue #7327, opened 2021-08-24
Path Distance Measurement Inconsistency


melchiar:
## Version
4.5.1

## Test Case
https://jsfiddle.net/melchiar/br06s23p/

## Information about environment
Browser

## Description

It looks like the path distance calculation for the text on path code can be a bit inconsistent at times when measuring curves. This results in uneven text kerning in some cases. 

Note how the distance between some dots can jump a bit near the tops and sides of curves when adjusting the pathStartOffset. Maybe something to do with the path simplification?

![path-distance-wiggle](https://user-images.githubusercontent.com/9057412/130657249-599a6d2a-5e92-4d50-bd71-ff7f45fcfce6.gif)

melchiar:
It looks like the jumps occur at the connection point between two curves. Since we're simplifying arcs and quadratic arcs into bezier curves, an arc ellipse like this:

`M 50 0 a 50 100 0 1 1 1 0 Z`

Gets converted into an ellipse composed of 4 bezier curves, with joints at the middle of each edge.

`M 50 0 C 22.48472060094895 -0.5503331053242485 0.3630400233350102 -45.46238353072359 0.5006250195322721 -100.49500612526722 C 0.6382100157295341 -155.52762871981085 22.9833447337881 -199.99499987499377 50.5 -199.99499987499377 C 78.0166552662119 -199.99499987499377 100.36178998427046 -155.52762871981088 100.49937498046773 -100.49500612526725 C 100.636959976665 -45.46238353072362 78.51527939905108 -0.5503331053242809 51.00000000000004 0`

![image](https://user-images.githubusercontent.com/9057412/130662606-f385f89c-7a4b-46df-8891-ed7d2f0edcbf.png)

It's at these joints that the distance calculation loses accuracy.

melchiar:
I'm still not exactly sure what's going on here, but I've been able to eliminate this issue by changing the default value of `nextStep` in the `findPercentageForDistance` function from 0.01 to 0.1. I'm thinking this is another floating point error but I haven't figured out exactly why yet.

asturur:
i can look into this now

TheAlemazing:
I changed the condition from 1 to 1.1 and it's worked

```
while (tmpLen < distance && perc < 1.1 && nextStep > 0.0001) {
...
}
```

PR #7511, opened 2021-11-22
Path Distance Measurement Inconsistency

TheAlemazing:
close: https://github.com/fabricjs/fabric.js/issues/7327

### DIFF
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -507,7 +507,9 @@
         p, nextLen, nextStep = 0.01, angleFinder = segInfo.angleFinder, lastPerc;
     // nextStep > 0.0001 covers 0.00015625 that 1/64th of 1/100
     // the path
-    while (tmpLen < distance && perc <= 1 && nextStep > 0.0001) {
+    
+    // floating point made perc <= 1 ignore some calculation, so perc < 1.1 do the trick
+    while (tmpLen < distance && perc < 1.1 && nextStep > 0.0001) {
       p = iterator(perc);
       lastPerc = perc;
       nextLen = calcLineLength(tempP.x, tempP.y, p.x, p.y);
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -507,8 +507,7 @@
         p, nextLen, nextStep = 0.01, angleFinder = segInfo.angleFinder, lastPerc;
     // nextStep > 0.0001 covers 0.00015625 that 1/64th of 1/100
     // the path
-    // floating point made perc <= 1 ignore some calculation, so perc < 1.1 do the trick
-    while (tmpLen < distance && perc < 1.1 && nextStep > 0.0001) {
+    while (tmpLen < distance && perc < (perc + nextStep) && nextStep > 0.0001) {
       p = iterator(perc);
       lastPerc = perc;
       nextLen = calcLineLength(tempP.x, tempP.y, p.x, p.y);
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -507,7 +507,7 @@
         p, nextLen, nextStep = 0.01, angleFinder = segInfo.angleFinder, lastPerc;
     // nextStep > 0.0001 covers 0.00015625 that 1/64th of 1/100
     // the path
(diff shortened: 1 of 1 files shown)


src/util/path.js
```diff
@@ -507,7 +507,9 @@
         p, nextLen, nextStep = 0.01, angleFinder = segInfo.angleFinder, lastPerc;
     // nextStep > 0.0001 covers 0.00015625 that 1/64th of 1/100
     // the path
-    while (tmpLen < distance && perc <= 1 && nextStep > 0.0001) {
+    
+    // floating point made perc <= 1 ignore some calculation, so perc < 1.1 do the trick
+    while (tmpLen < distance && perc < 1.1 && nextStep > 0.0001) {
```
asturur, line 512: why did they even bother adding floats to JS.

asturur: You left some trailing spaces, lint is failing.
This is good to go probably.
Can we look into reducing 1.1 to minimum possible? 1.01? 1.1 would be 10% of tolerance seems high.
I would also like to see a test that fails with 1 and pass with 1.0x

TheAlemazing:
I will add the test soon.

src/util/path.js
TheAlemazing: I think the correct should be
```
while (tmplen < distance && perc < (perc + nextstep) && nextstep > 0.0001)
```

src/util/path.js
```diff
@@ -507,7 +507,7 @@
         p, nextLen, nextStep = 0.01, angleFinder = segInfo.angleFinder, lastPerc;
     // nextStep > 0.0001 covers 0.00015625 that 1/64th of 1/100
     // the path
-    while (tmpLen < distance && perc <= 1 && nextStep > 0.0001) {
+    while (tmpLen < distance && perc < (perc + nextStep) && nextStep > 0.0001) {
```
asturur, line 510: mmm this smells like infinite loop

asturur: can be this just < ( 1 + nextStep ), would this work?

TheAlemazing: yeah. I think 1 + nextStep is the correct one.

TheAlemazing: I tried, 1 + nextStep does not work. I think we can remove the percentage check as `tmpLen < distance` is enough.

TheAlemazing: I think we can remove the percentage check. Only `tmpLen < distance && nextStep > 0.0001` are enough

TheAlemazing:
Some animate ezgif com-gif-maker

src/util/path.js
```diff
@@ -507,15 +507,15 @@
...
+    while (tmpLen < distance && nextStep > 0.0001) {
       p = iterator(perc);
       lastPerc = perc;
       nextLen = calcLineLength(tempP.x, tempP.y, p.x, p.y);
       // compare tmpLen each cycle with distance, decide next perc to test.
       if ((nextLen + tmpLen) > distance) {
         // we discard this step and we make smaller steps.
-        nextStep /= 2;
         perc -= nextStep;
+        nextStep /= 2;
```
asturur, line 518: This could have been the bug.

asturur:
@melchiar since you are the highest text on a path stakeholder here, do you want to have a look at this code?
I think is fine believe it will work fine.

src/util/path.js
TheAlemazing: 1. Nope, it's not. I changed the line order to match the comment.
2. I think a bug relate to perc calculation, at the end of the path there are calculation steps that made perc greater than 1 (eg 1.00000000007) so the position calculation ignored. ( as we can see the text is not moving )

TheAlemazing: 3. We do not need to care about perc >= 1 because Distance and the tmpLen cover that.

melchiar:
> @melchiar since you are the highest text on a path stakeholder here, do you want to have a look at this code? I think is fine believe it will work fine.

Sorry, been crazy busy and haven't had time to test yet. I will try to take a look on Monday

melchiar:
Looks good to me!

asturur (approved):

asturur:
You can formally approve next time!

Issue #7471, opened 2021-11-03
`canvas.viewportTransform` (scale bug)


suncs2001:
## Version
3.6.0

## Information about environment
Nodejs or browser?
browser
Which browsers?
chrome
## Steps to reproduce
1、
canvas = document.getElementById('canvas')
let context = this.canvas.getContext('2d')
context.transform(1, 0, 0, -1, 0, canvasHeight)
2、 zoom in or zoom out the canvas 
3、 mouse move
## Expected Behavior
when mouse move，in mouse move listener , the pos obtained by getPointer  is canvas pos
## Actual Behavior
pos.x is actual canvas position and it is correct,  but pos.y is incorrect

ShaMan123:
Your issue doesn't provide enough context.
However, it seems you are directly manipulating the context. Which can cause many things if you don't own it 100%.
For example which context are you transforming? the top or the bottom/main? which is another way of asking which canvas did you obtain.
You should use `canvas.viewportTransform` or `setViewportTransform`

suncs2001:
hi ShanMan, thanks for your replay. 
yes ,i should not use context transform but instead of  using fabricjs viewportTransform, and after that, the getPointer() function return the right value.
but i found that zoomToPoint() function did not support transform change, because vpt[1] and vpt[3] all use param 'zoom, that means if i call zoomToPoint() after viewportTransform([1,0,0,-1,0,height]), its transform back to [1,0,0, 1,0,0] again.
so i had to implement zoomToPoint() myself, i copy the source code of zoomToPoint(),  just modify "vpt[3] = value"  to "vpt[3] = -1 * value",   but there is a strange thing,  sometimes the object  added to canvas can not display.

i test it in codepan
https://codepen.io/suncs2001/pen/BadbyBP
i found that in fabricjs 1.6.2, it worked well, but in fabricjs 4.5.0, the circle can not display

ShaMan123:
It's still unclear.
You should provide steps to reproduce.
I have no idea what your pen should show.
You should look at the demos to see how zoomToPoint works

suncs2001:
i edit my pen again, and  it is easier to understand . from my pen , i do two things.  
1) make viewportTransfrom,  flip y axis.  
`canvas.setViewportTransform([1,0,0,-1,0, 400]);`	
2) draw a circle in (50,50)    
`  var circle = new fabric.Circle({
        top :  50, 
        left : 50, 
        radius : 50, 
        fill : 'red', 
[...]

PR #7515, opened 2021-11-23
fix(Canvas): flipped viewport transform coords

ShaMan123:
### Motivation
#7471 

### gist
When flipped (negative scale) `Object.isOnScreen` breaks because it tried to check if a point is within `vptCoords` which always returned `false` because when flipped `tl` was greater than `br`.

### Important
I've adjusted tests. 
I need you to review that they are good enough, viewport transform is confusing.
The test enhancement exposed another bug with `isPartiallyOnScreen`.
It's test used stale canvas state, that is why it passed. Now it fails.
The reason it fails is because the rect has a common line with canvas vpt so the intersection test running in `isPartiallyOnScreen` returns true. But it is wrong. `isPartiallyOnScreen` should return false in this case, which is unhandled.


asturur:
Didn't we fix this already?

asturur:
@ShaMan123 if you have time to move this PR in a branch i can look into test failures and finish it.

ShaMan123:
> @ShaMan123 if you have time to move this PR in a branch i can look into test failures and finish it.

@asturur ?

asturur (dismissed):

ShaMan123:
Should we rebase? Update from master? I'll leave it to you.
I haven't visited this PR in a long time

ShaMan123:
Cause I'm afraid of unwanted overwrite

asturur (approved):

asturur:
one test needed to be updated, since it had a flipped viewport, and so the new function was supposed to change the test result.
The other i don't know but was effectively an edge case, so i just made less edgy.

PR #7520, opened 2021-11-26
fix(fabric.parser): attempt to resolve some issues with regexp

asturur:
It looks like the regexp we use to parse CSS may be susceptible of high CPU usage with crafted payloads.
This would make servers that use fabric to parse SVGs at risk of crash or DOS.
This PR removes the regexes and uses a simpler approach to parse style rules.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1012,7 +1012,7 @@
 
           var match = rule.split('{'),
               ruleObj = { }, declaration = match[1].trim(),
-              propertyValuePairs = declaration.split(';');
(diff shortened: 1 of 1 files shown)


asturur:
@JamieSlome  do you want to verify if this fix your concens?

JamieSlome:
@asturur - thanks for the bump 🤛 

I am just tagging the researcher @ready-research - can you take a look at the fix here?

ready-research:
@asturur - Yeah, proposed changes LGTM. Sorry for the late reply.

asturur:
why i can't merge? lol

asturur:
@ShaMan123 can you try approve this PR?

PR #7523, opened 2021-11-29
feat: Customize line width for controls

rockerBOO:
Allows for custom controls to define the line width that goes around the control.

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineWidth


src/controls.render.js
```diff
@@ -44,8 +45,8 @@
     else {
       size = xSize;
     }
-    // this is still wrong
-    ctx.lineWidth = 1;
+    // this is still wrong (what is still wrong? original [ctx.lineWidth = 1])
+    ctx.lineWidth = lineWidth;
```
asturur, line 49: i think this property exists already, or at least existed before.

PR #7524, opened 2021-11-29
feat(Control) Add control fill/stroke options

rockerBOO:
This allows custom controls to have specific fill/stroke outside specific
object options (fabric.Object.cornerStrokeStyle).

```javascript
const object = new fabric.Rect({ width: 5, height: 10 })
object.set('controls', {
 mtr: new fabric.Control({ fill: 'red', stroke: 'blue' })
})
```

### DIFF
--- a/src/controls.render.js
+++ b/src/controls.render.js
@@ -24,7 +24,9 @@
         transparentCorners = typeof styleOverride.transparentCorners !== 'undefined' ?
           styleOverride.transparentCorners : fabricObject.transparentCorners,
         methodName = transparentCorners ? 'stroke' : 'fill',
-        stroke = !transparentCorners && (styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor || this.stroke),
+        stroke = !transparentCorners && (
+          styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor || this.stroke
+        ),
         myLeft = left,
         myTop = top, size;
     ctx.save();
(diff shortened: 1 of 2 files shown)


src/controls.render.js
```diff
@@ -24,12 +24,14 @@
         transparentCorners = typeof styleOverride.transparentCorners !== 'undefined' ?
           styleOverride.transparentCorners : fabricObject.transparentCorners,
         methodName = transparentCorners ? 'stroke' : 'fill',
-        stroke = !transparentCorners && (styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor),
+        stroke = !transparentCorners && (
+          styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor || this.stroke
+        ),
```
asturur, line 29: this gets hairy.
How do we decide priority?
ideally the  most specific part always win.
so do i have to paint the stroke?

well if !transparentCoreners
if this stroke is not undefined and is not defined as transparent or not defined as empty
if styleOverride.cornerStrokeColor is not undefined and is not defined as transparent or not defined as empty
if ....

:(

rockerBOO: I don't like the complexity but not sure of how to cascade options like this without it?

```
transparentCorners, stroke = false,
!transparentCorners
  > styleOverride
  > fabricObject
  > stroke

If styleOverride is set, it will override all other options
If fabricObject is set (object specific options for controls should take priority)
If stroke is set on the control (control settings are the default option and will fall through in priority until the last option.)
```
I can separate it into 2 statements for the requested color and if the corners are transparent.

asturur: what take precedence is a bit a matter of opinion, the important is that we document it.
Controls are generally shared between objects, this is how the library is setup by default, so it makes sense that object settings take precedence.
I m not sure what is the default value of those properties, because if is not undefined it becomes complicated.
Maybe we should consider the empty space a lazy default value, while if you want to take the precedence with a transparent one, instead of using false or empty space you use the word `transparent`

Issue #1719, opened 2014-10-01
animation delay


heaversm:
Hi - is there a delay parameter for animations? I'm running a javascript loop to add a series of circles to the canvas, and I want them to pop in, but staggered, at a random delay.
##

alexiskattan:
+1

yezongyang:
+1

Tsourdox:
+1

charmjeh:
+1

dellvolk:
+1

asturur:
Are we sure it does not exists today?

PR #7528, opened 2021-12-01
feat(animation): animations registry

ShaMan123:
### Motivation
1. There's a need to cleanup animations in applications.
Currently the dev needs to implement something by themselves.
For example in react while developing when react does a hot reload or whatever and fabric is animating an error raises because the canvas is trying to access the context which react just disposed.
2. Another use case I can think of is with erasing. It makes sense to iterate over all animations before erasing starts to make the objects non erasable. (I wanted this to be default behavior but lacked the mechanism to do so)
In general, knowing what is animating is crucial information for good UX/UI. 

I think this should be handled by fabric.

### gist
Exposed `fabric.runningAnimations` (is this a good name? is fabric the right level to hold this ref?
It is an array containing the options object passed to `animate` and the `cancel` function. This means that it can contain additional context if such is passed to animate allowing the dev freedom. 
Exposed the following methods: `cancelAll`, finders
For example we could pass the object/canvas as context, this will make it very simple to cancel relevant animations.

IMO It makes sense that an object holds a ref to all it's running animations and that it's `dispose` method should cancel them. 
What do you think?

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1200,6 +1200,7 @@
      */
     dispose: function () {
       var wrapper = this.wrapperEl;
+      fabric.runningAnimations.cancelAll();
       this.removeListeners();
       wrapper.removeChild(this.upperCanvasEl);
       wrapper.removeChild(this.lowerCanvasEl);
--- a/src/mixins/animation.mixin.js
+++ b/src/mixins/animation.mixin.js
@@ -25,6 +25,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         _this = this;
 
     fabric.util.animate({
+      target: this,
       startValue: object.left,
       endValue: this.getCenter().left,
       duration: this.FX_DURATION,
@@ -60,6 +61,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         _this = this;
 
     fabric.util.animate({
+      target: this,
       startValue: object.top,
       endValue: this.getCenter().top,
       duration: this.FX_DURATION,
@@ -95,6 +97,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         _this = this;
 
     fabric.util.animate({
+      target: this,
       startValue: object.opacity,
       endValue: 0,
       duration: this.FX_DURATION,
@@ -196,6 +199,7 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
     }
 
     var _options = {
+      target: this,
       startValue: options.from,
       endValue: to,
       byValue: options.by,
--- a/src/mixins/animation.mixin.js
+++ b/src/mixins/animation.mixin.js
@@ -13,8 +13,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
    * @param {Object} [callbacks] Callbacks object with optional "onComplete" and/or "onChange" properties
    * @param {Function} [callbacks.onComplete] Invoked on completion
    * @param {Function} [callbacks.onChange] Invoked on every step of animation
-   * @return {fabric.Canvas} thisArg
-   * @chainable
+   * @return {fabric.AnimationContext} context
    */
   fxCenterObjectH: function (object, callbacks) {
     callbacks = callbacks || { };
@@ -24,7 +23,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         onChange = callbacks.onChange || empty,
         _this = this;
 
-    fabric.util.animate({
+    return fabric.util.animate({
       target: this,
       startValue: object.left,
       endValue: this.getCenter().left,
@@ -39,8 +38,6 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         onComplete();
       }
     });
-
-    return this;
   },
 
   /**
@@ -49,8 +46,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
    * @param {Object} [callbacks] Callbacks object with optional "onComplete" and/or "onChange" properties
    * @param {Function} [callbacks.onComplete] Invoked on completion
    * @param {Function} [callbacks.onChange] Invoked on every step of animation
-   * @return {fabric.Canvas} thisArg
-   * @chainable
+   * @return {fabric.AnimationContext} context
    */
   fxCenterObjectV: function (object, callbacks) {
     callbacks = callbacks || { };
@@ -60,7 +56,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         onChange = callbacks.onChange || empty,
         _this = this;
 
-    fabric.util.animate({
+    return fabric.util.animate({
       target: this,
       startValue: object.top,
       endValue: this.getCenter().top,
@@ -75,8 +71,6 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
         onComplete();
       }
     });
-
-    return this;
   },
 
   /**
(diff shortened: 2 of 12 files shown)


src/util/animate.js
```diff
@@ -52,14 +115,13 @@
           return;
         }
         if (abort(current, valuePerc, timePerc)) {
-          // remove this in 4.0
-          // does to even make sense to abort and run onComplete?
-          onComplete(endValue, 1, 1);
```
asturur, line 57: my comment still stands

src/util/animate.js
```diff
@@ -11,22 +71,27 @@
...
-   * @param {Function} [options.onChange] Callback; invoked on every value change
-   * @param {Function} [options.onComplete] Callback; invoked when value change is completed
-   * @param {Number} [options.startValue=0] Starting value
-   * @param {Number} [options.endValue=100] Ending value
-   * @param {Number} [options.byValue=100] Value to modify the property by
-   * @param {Function} [options.easing] Easing function
-   * @param {Number} [options.duration=500] Duration of change (in ms)
-   * @param {Function} [options.abort] Additional function with logic. If returns true, onComplete is called.
-   * @returns {Function} abort function
+   * @param {AnimationOptions} [options] Animation options
```
asturur, line 74: this helps with tpyescript, i m not sure if jsdocs still sees it.

src/util/animate.js
```diff
@@ -11,22 +71,27 @@
...
+   * @returns {CancelFunction} cencel function
```
asturur, line 75: cencel = cancel

src/util/animate.js
```diff
@@ -1,4 +1,64 @@
...
+
+    /**
+     *
+     * @param {CancelFunction} cancelFunc the function returned by animate
+     * @returns {number}
+     */
+    findAnimationIndex: function (cancelFunc) {
+      return this.findIndex(function (animation) {
+        return animation.cancel === cancelFunc;
+      });
```
asturur, line 48: we still support ie11.
Since we have a loop to find findAnimation and findIndex i think is not supported in ie11, this should be:

```
return this.indexOf(findAnimation(cancelFunc));
```

src/util/animate.js
```diff
@@ -1,4 +1,64 @@
...
+   * @type {AnimationContext[]}
+   */
+  var RUNNING_ANIMATIONS = [];
+  fabric.util.object.extend(RUNNING_ANIMATIONS, {
+
+    /**
+     * cancel all running animations
+     */
+    cancelAll: function () {
+      var animations = this.splice(0, this.length);
```
asturur, line 34: this.splice(0) is enough, i think also this.concat(); is good enough

src/util/animate.js
asturur: since we did not remove this in 4.0 we can't remove it now. now it has to wait for 5.0

src/util/animate.js
ShaMan123: I see in VSCode... not sure as well.
What's the decision?

src/util/animate.js
ShaMan123: af7cdc3f9cff33cd74e83b77ac479273269ad59d

src/util/animate.js
ShaMan123: 456a36052dda7a16009849017a8122f504e81b00

src/util/animate.js
ShaMan123: 456a36052dda7a16009849017a8122f504e81b00

src/util/animate.js
ShaMan123: 456a36052dda7a16009849017a8122f504e81b00

src/util/animate.js
ShaMan123: 456a36052dda7a16009849017a8122f504e81b00

ShaMan123:
ready

ShaMan123:
> is this a good name? is fabric the right level to hold this ref?
@asturur

ShaMan123:
I have added current state to animation context so now it enables even more freedom of use

src/util/animate.js
asturur: test if docs build is enough. i can do that don't worry about it now

ShaMan123: OK.
So, this is stable and can be merged

ShaMan123:
I remembered you mentioned there's a need to cleanup `runningAnimations` from object/canvas dispose.
So I'll commit that.

ShaMan123:
1. Once an animation ends it is removed from the registry
2. added a `target` property to animation context
2. all object animations get registered with `target` set
3. object `dispose` cancels the target's animations
4. canvas `dispose` cancels all animations

No memory leaks should exist

src/util/animate.js
```diff
@@ -48,18 +147,28 @@
...
+        //  update context
+        context.currentValue = current;
+        context.completionRate = valuePerc;
+        context.durationRate = timePerc;
         if (cancel) {
           return;
         }
         if (abort(current, valuePerc, timePerc)) {
-          // remove this in 4.0
+          // remove this in 5.0
```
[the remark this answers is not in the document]
ShaMan123: This is good to remove because then the option to pause an animation comes to play.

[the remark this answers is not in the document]
ShaMan123: 0695460

ShaMan123:
I think it's done so ready to review

ShaMan123:
There's something important!
When using `object.animate` I can't access the value returned by fabric.util.animate because the function is chainable.
Should we break? I think so. Because it is awkward trying to access the animation context without it.

https://github.com/fabricjs/fabric.js/blob/7a406e85c128fcea79f2eecc65ca41829beac417/src/mixins/animation.mixin.js#L154

There are a few more places such as this

src/canvas.class.js
```diff
@@ -1200,6 +1200,7 @@
      */
     dispose: function () {
       var wrapper = this.wrapperEl;
+      fabric.runningAnimations.cancelAll();
```
asturur, line 1203: this needs to be executed per canvas.
EIther you pass in the canvas as argument and cancelAll becomes a cancelAll by canvas, or in an application with 2 canvases this would stop everything.

src/mixins/animation.mixin.js
```diff
@@ -13,8 +13,7 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
    * @param {Object} [callbacks] Callbacks object with optional "onComplete" and/or "onChange" properties
    * @param {Function} [callbacks.onComplete] Invoked on completion
    * @param {Function} [callbacks.onChange] Invoked on every step of animation
-   * @return {fabric.Canvas} thisArg
-   * @chainable
```
asturur, line 17: maybe i m wrong, but is chainability still cool to do?
I understand it saves typing and as a result bits in the final bundle.
But i find having half methods chainable and half not is weird. Maybe is something we should drop in the rewrite.

ShaMan123: I agree. Chaining is not worth the weirdness.
I say we drop it.

ShaMan123: And it is less breaking (for future impl)

src/static_canvas.class.js
```diff
@@ -1117,14 +1104,11 @@
...
      * @param {fabric.Object} object Object to center
      * @param {fabric.Point} center Center point
-     * @return {fabric.Canvas} thisArg
-     * @chainable
      */
     _centerObject: function(object, center) {
       object.setPositionByOrigin(center, 'center', 'center');
       object.setCoords();
       this.renderOnAddRemove && this.requestRenderAll();
-      return this;
```
asturur, line 1127: i would like to limit breaking changes in 5. Those are not part of the animation changes. What is the reason to remove chainability now?

src/canvas.class.js
ShaMan123: Good point.
When an objects animates does it have to be attached to a canvas? I guess not. So I'm not sure how to do what you suggest.
I can add a canvas property to the animation context object but what if there's no canvas set to the object??
Or maybe we should iterate over all running animations and check at that moment who is attached to the canvas arg and cancel that?

asturur:
i m going to merge this and open a following PR to restore some extra changes that are for now not needed and complicate the release note of 5.0

ShaMan123:
BTW I should I push a small commit for the delay option (#1719 ) or leave it for another PR?

ShaMan123:
> i m going to merge this and open a following PR to restore some extra changes that are for now not needed and complicate the release note of 5.0

I you think it's better to release v5 without this that's alright. We'll make it part of v6.

src/static_canvas.class.js
ShaMan123: It's as you said. I thought it's too weird to have the animated version not chainable while the regular one is.

asturur:
i m fine adding this now, i just don't want to remove the chaining operators that do not return animations.
I have seen the commit for the delay, i wonder what are we doing with the start option then? isn't the start a delay itself?

ShaMan123:
I don't see a start option. Or anything resembling a delay option

Issue #7502, opened 2021-11-18
Canvas blur after rescale page


vitonsky:
Canvas is not rescale content after resize page.
When we load web page with 50% page zoom and then set zoom to 100% canvas looks  very bad like low resolution pucture.

I did try to manually rescale fabricjs inner objects while resize page like that:
```ts
export const fixCanvasSize = (canvas: fabric.Canvas) => {
    const lowerCanvasEl = (canvas as any).lowerCanvasEl;

    const width = (canvas as any).width;
    const height = (canvas as any).height;

    lowerCanvasEl.width = width;
    lowerCanvasEl.height = height;

    canvas.renderAll();
};
```

And this fix problem, but controls to resize objects are render wrong and i can't fix it.
If you have solution, tell about it.

Seems it impossible to fix on library user side

## Version
3.6.0

## Test Case
It reproduce on any page with fabricjs, try it on the http://fabricjs.com/

## Information about environment
All browsers

## Steps to reproduce
- Scale page with canvas to 50%
- Reload page
- Scale page to 100%

## Expected Behavior
- Fine canvas picture

## Actual Behavior
- Blur on the canvas, because it was render as 50% size and then resized without rescale

rockerBOO:
Try using setDimensions

```
   canvas.setDimensions({
      width: window.innerWidth,
      height: window.innerHeight
    });
```

vitonsky:
@rockerBOO i've try it, but it not work for me.

I did try use only `setDimensions`, also tried set size for `lowerCanvasEl` and then call `setDimensions` and vice versa.
I also tried all options combinations. It did not fix this bug.

After call `setDimensions` canvas resize without fix of scale factor.

rockerBOO:
```
  canvas.setWidth(window.innerWidth);
  canvas.setHeight(window.innerHeight);

  canvas.calcOffset();
  canvas.requestRenderAll();
```

I have also used this option. Don't try to adjust the lowerCanvasEl or upperCanvasEl. They are generally abstracted away where you shouldn't need to access them. You can infer things from them but manually adjusting is probably going to cause some conflict.

vitonsky:
@rockerBOO unfortunately `calcOffset` is not fix canvas too. Is it work for you?
Btw, i not resize canvas object, i only rescale web page (but i set size explicitly for test it as you say)

I adjust `lowerCanvasEl` because this method is fix a image, but after this fix an object controls render with incorrect position.
[...]

PR #7537, opened 2021-12-05
prevent render canvas with quality less than 100%

vitonsky:
Fix issue https://github.com/fabricjs/fabric.js/issues/7502

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -240,7 +240,7 @@
      * @return {Number} retinaScaling if applied, otherwise 1;
      */
     getRetinaScaling: function() {
-      return this._isRetinaScaling() ? fabric.devicePixelRatio : 1;
+      return Math.max(1, fabric.devicePixelRatio);
     },
 
     /**
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -232,7 +232,7 @@
      * @private
      */
     _isRetinaScaling: function() {
-      return (fabric.devicePixelRatio !== 1 && this.enableRetinaScaling);
+      return (fabric.devicePixelRatio > 1 && this.enableRetinaScaling);
     },
 
     /**


src/static_canvas.class.js
```diff
@@ -232,7 +232,7 @@
      * @private
      */
     _isRetinaScaling: function() {
-      return (fabric.devicePixelRatio !== 1 && this.enableRetinaScaling);
+      return (fabric.devicePixelRatio > 1 && this.enableRetinaScaling);
```
asturur, line 235: you also need to change getRetinaScaling to return a number that is at minimum 1 i think.

vitonsky: Done

Dok11: Maybe make sense to switch places of these parts for performance? Checking a boolean variable should be faster than comparing two values. Like this: `this.enableRetinaScaling && fabric.devicePixelRatio > 1`.

vitonsky: @Dok11 i don't sure. JS engine probably is don't know whats variable contain, so will not check variable as boolean directly. But symbol `>` may suggest that it's just math operation which is instant

Dok11: @vitonsky I even tested it, and it seems like it doesn't care :D
https://jsbench.me/zkkxq3rjv0/1

image

ShaMan123:
@vitonsky could you push a commit that calculates retina scaling for within [`_onResize`](https://github.com/fabricjs/fabric.js/blob/7efc732d8a050d743a66c89d8b1c4e5f942191f4/src/mixins/canvas_events.mixin.js#L372) as suggested in https://github.com/fabricjs/fabric.js/issues/7502#issuecomment-986287369

src/static_canvas.class.js
asturur: i m sure performance is not an issue here, but in the JS bench seems like Math.random and toFixed are actually the 2 things that will matter for performance. no?

asturur (approved):

asturur:
we broke so many tests. not sure why it wasn't visible in the pr

src/static_canvas.class.js
```diff
@@ -232,15 +232,15 @@
...
     },
 
     /**
      * @private
      * @return {Number} retinaScaling if applied, otherwise 1;
      */
     getRetinaScaling: function() {
-      return this._isRetinaScaling() ? fabric.devicePixelRatio : 1;
+      return Math.max(1, fabric.devicePixelRatio);
```
asturur, line 243: ok this was the error.
We lessen the importance of the boolean here.
If retina scaling is set as false, whatever is the devicePixelRatio, we return 1.
My bad for not noticing it.

vitonsky: Probably we can fix it like this?
```js
this._isRetinaScaling() ? Math.max(1, fabric.devicePixelRatio) : 1
```

vitonsky:
@ShaMan123 
> @vitonsky could you push a commit that calculates retina scaling from within [`_onResize`](https://github.com/fabricjs/fabric.js/blob/7efc732d8a050d743a66c89d8b1c4e5f942191f4/src/mixins/canvas_events.mixin.js#L372) as suggested in [#7502 (comment)](https://github.com/fabricjs/fabric.js/issues/7502#issuecomment-986287369)

I'm not sure that we can do it on the library level. If you wish, make own pull request with this change if possible

PR #7585, opened 2022-01-08
feat(Eraser): v2 #2

ShaMan123:
### Summary
This PR along with #7470 consist of Eraser's v2.
This PR **deprecates** eraser's bad use of `clipPath`.
It was bad practice to do it as I did (at the time I was overwhelmed by fabric's internals), so know it's fixed and for the best!
This fix removes all awkwardness that was caused by riding on clip path (handling clip path and eraser was brittle).
This makes eraser much more robust, more legitimate and less a hack. **MATURE**. a proper part of fabric.
Previously introducing eraser to an existing project would break clip path in some cases and require a lot of touch up.
Now it doesn't.
Simply plug and play.

### Changes
1. clip path has returned to it's normal behavior.
2. Object has a dedicated prop `eraser` (instead of riding on clip path).
3. Eraser is now a subclass of group. This means that `fromJSON` will break between versions. Correct it by changing the `type` prop from `group` to `eraser` and remove the first (rect) object.
4. dropped `getEraser` `setClipPath` `getClipPath`
5. fixed eraser clipped by `absolutePositioned` clip path 
4. fixed `toClipPathSVG` -  support nested clip paths (*group still remains problematic #7586 )

### Next Up
Tests


src/shapes/object.class.js
```diff
@@ -1958,6 +1958,15 @@
...
   fabric.Object.NUM_FRACTION_DIGITS = 2;
 
+  /**
+   * Defines which properties should be enlivened from the object passed to {@link fabric.Object._fromObject}
+   * @static
+   * @memberOf fabric.Object
+   * @constant
+   * @type string[] 
+   */
+  fabric.Object.ENLIVEN_PROPS = ['clipPath'];
```
asturur, line 1968: this is missing in the first PR.

src/mixins/object.svg_export.js
```diff
@@ -180,15 +180,26 @@
...
+      var commonPieces = [
+        this.getSvgTransform(true, additionalTransform),
+        this.getSvgCommons(),
+      ].join(''),
+        // insert commons in the markup, style and svgCommons
+        index = objectMarkup.indexOf('COMMON_PARTS');
       objectMarkup[index] = commonPieces;
-      return reviver ? reviver(objectMarkup.join('')) : objectMarkup.join('');
+      markup.push(objectMarkup.join(''));
+      return reviver ? reviver(markup.join('')) : markup.join('');
```
asturur, line 202: ok this is a tentative to make the clipPath svg export recursive, right?
This is unrelated to brush, is a general improvement/fix and deserve its own pr with test ( testing svg export is very easy luckly )
Sorry for the extra work but i prefer we split it.

src/shapes/object.class.js
asturur: Could FILL, STROKE be a part of the enlivable props and handle gradient and patterns?
maybe a tought for a later pr.

As today we have this discrepancy where you can send in the object options a non enlived gradient and pattern and the object will iniitialize them. While this does not work for clipPath.
I wonder if we should make the same for everything

ShaMan123: Sounds easy to adjust

src/mixins/object.svg_export.js
ShaMan123: Alright

src/shapes/object.class.js
ShaMan123: And a good idea

ShaMan123: > this is missing in the first PR.

I see it on both PRs

ShaMan123:
@asturur do you mind I shut this PR down and work only on #7470 ?
It's easier for me beause some changes are here and some there and then I need to resolve conflict and so on.

ShaMan123:
merged into #7470

asturur:
ok i was about to ask

Issue #7598, opened 2022-01-10
Editing text inside groups


kirill-konshin:
## Version
4.6.0

## Test Case

http://jsfiddle.net/pce9h6fv/6/

## Steps to reproduce

1. Open fiddle
2. Click "edit" (calls `text.enterEditing()`)
3. Type something

I am aware that IText editing is not supported when it's in groups https://github.com/fabricjs/fabric.js/issues/3787 and there are custom solutions like here: https://stackoverflow.com/questions/24449481/fabric-js-grouped-itext-not-editable, for example to create a temporary text node https://stackoverflow.com/a/67689690/5125659. 

But this is not very optimal.

I am wondering, if there's a way to route events from a group to particular text node, given that I know, there will be only one such node per group, or it will be more work that fake text node way?

I also attempted to set custom SVG background to IText node, in order to avoid groups at all, but this is not supported either: https://github.com/fabricjs/fabric.js/discussions/7297.

## Expected Behavior

Text is properly shown

## Actual Behavior

Text input is missing, yet the cursor is moving.

Additionally if group is rotated/moved during editing it leaves a cursor trail:

![image](https://user-images.githubusercontent.com/1530410/148826234-1739ba68-bbd6-4177-9e96-51e34e1ddbd6.png)

ShaMan123:
Your barging through an open door.
Take a look at #7473 #7599 
They should fix your issues (I saw you raised another regarding group).
The plan is to rewrite group to support interacting with nested objects.

kirill-konshin:
@ShaMan123 yep, I've had another ticket regarding groups, I'm glad you saw it too :)

What is the projected timeline for the collections? Are we talking months or weeks? It would be great to cleanup all our own hacks and workarounds...

ShaMan123:
I saw weeks, hopefully quicker.
I would appreciate it if you would test it.
The more testers the faster it evolves.

kirill-konshin:
@ShaMan123 sure, if you can provide some instructions how to migrate old code to new proposal, and the NPM build I can test it in my app.

ShaMan123:
You can try 
```bash
npm i --save fabricjs/fabric.js/#layer-proposal
```

kirill-konshin:
Surprisingly, I had a slightly better results with objectCaching: false on group/collection.

[...]

PR #7599, opened 2022-01-10
fix(Geometry): account for group transform

ShaMan123:
Shockingly I managed to fix Group to some degree, enabling some form of nested selection.
I was trying to solve something for ICollection and while reading #7130 again I came up with a fix.
I am quite embarrassed, after all the work I put into ICollection.
I still think ICollection is much better and solves many issues. 
So I am moving forward as planned.
What do you think @asturur?

Test it
Move around the group (circle+rect).
On mouse up the circle gets selected.
Try moving and transforming it.
Try transforming the group.

http://jsfiddle.net/q0hk1ve7/2/

This example has selection enabled (merged #7473 ): http://jsfiddle.net/h2omjyw1/4/

TO DO Next (?)
- [x] rotation control handler
- [x] control handlers when parent group is transform 
- [ ] render border box and controls of nested object when parent is transformed (scale/skew/rotate)
- [x] text editing

### DIFF
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -25,7 +25,7 @@
    */
   function findCornerQuadrant(fabricObject, control) {
     //  angle is relative to canvas plane
-    var angle = fabricObject.group ? fabric.util.qrDecompose(fabricObject.calcTransformMatrix()).angle : fabricObject.angle;
+    var angle = fabricObject.getTotalAngle();
     var cornerAngle = angle + radiansToDegrees(Math.atan2(control.y, control.x)) + 360;
     return Math.round((cornerAngle % 360) / 45);
   }
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -231,7 +231,7 @@
         control = target.controls[transform.corner],
         zoom = target.canvas.getZoom(),
         padding = target.padding / zoom,
-        localPoint = target.toLocalPoint(new fabric.Point(x, y), originX, originY);
+        localPoint = target.normalizePoint(new fabric.Point(x, y), originX, originY);
     if (localPoint.x >= padding) {
       localPoint.x -= padding;
     }
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -24,7 +24,9 @@
    * @return {Number} 0 - 7 a quadrant number
    */
   function findCornerQuadrant(fabricObject, control) {
-    var cornerAngle = fabricObject.angle + radiansToDegrees(Math.atan2(control.y, control.x)) + 360;
+    //  angle is relative to canvas plane
+    var angle = fabricObject.group ? fabric.util.qrDecompose(fabricObject.calcTransformMatrix()).angle : fabricObject.angle;
+    var cornerAngle = angle + radiansToDegrees(Math.atan2(control.y, control.x)) + 360;
     return Math.round((cornerAngle % 360) / 45);
   }
 
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -942,7 +942,8 @@
           transform = this._currentTransform,
           target = transform.target;
       if (target.group) {
-        //  transform pointer to target's coordinate plane
+        //  transform pointer to target's containing coordinate plane
+        //  both agree on every point
         pointer = fabric.util.transformPoint(pointer, fabric.util.invertTransform(target.group.calcTransformMatrix()));
       }
       transform.reset = false;
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -939,8 +939,12 @@
      */
     _transformObject: function(e) {
       var pointer = this.getPointer(e),
-          transform = this._currentTransform;
-
+        transform = this._currentTransform,
+        target = transform.target;
+      if (target.group) {
+        //  transform pointer to target's coordinate plane
+        pointer = fabric.util.transformPoint(pointer, fabric.util.invertTransform(target.group.calcTransformMatrix()));
+      }
       transform.reset = false;
       transform.shiftKey = e.shiftKey;
       transform.altKey = e[this.centeredKey];
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -939,8 +939,8 @@
      */
     _transformObject: function(e) {
       var pointer = this.getPointer(e),
-        transform = this._currentTransform,
-        target = transform.target;
+          transform = this._currentTransform,
+          target = transform.target;
       if (target.group) {
         //  transform pointer to target's coordinate plane
         pointer = fabric.util.transformPoint(pointer, fabric.util.invertTransform(target.group.calcTransformMatrix()));
--- a/src/mixins/itext_click_behavior.mixin.js
+++ b/src/mixins/itext_click_behavior.mixin.js
@@ -152,7 +152,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    */
   mouseUpHandler: function(options) {
     this.__isMousedown = false;
-    if (!this.editable || this.group ||
+    if (!this.editable ||
       (options.transform && options.transform.actionPerformed) ||
       (options.e.button && options.e.button !== 1)) {
       return;
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -526,7 +526,7 @@
      * aCoords are used to quickly find an object on the canvas
      * lineCoords are used to quickly find object during pointer events.
      * See {@link https://github.com/fabricjs/fabric.js/wiki/When-to-call-setCoords} and {@link http://fabricjs.com/fabric-gotchas}
-     * 
+     *
      * @param {Boolean} [skipCorners] skip calculation of oCoords.
      * @return {fabric.Object} thisArg
      * @chainable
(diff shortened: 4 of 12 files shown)


src/mixins/canvas_events.mixin.js
```diff
@@ -939,8 +939,12 @@
      */
     _transformObject: function(e) {
       var pointer = this.getPointer(e),
-          transform = this._currentTransform;
-
+          transform = this._currentTransform,
+          target = transform.target;
+      if (target.group) {
+        //  transform pointer to target's coordinate plane
+        pointer = fabric.util.transformPoint(pointer, fabric.util.invertTransform(target.group.calcTransformMatrix()));
```
asturur, line 946: yes this is exactly what i would have done to transform stuff into the group.

src/shapes/object.class.js
```diff
@@ -979,6 +979,18 @@
...
+     * Returns the object angle relative to canvas counting also the group property
+     * @returns {number}
+     */
+    getTotalAngle: function () {
+      var angle = this.angle;
+      if (this.group) {
+        angle += this.group.getTotalAngle();
+      }
+      return angle;
+    },
```
asturur, line 992: i would use the qrDecompose to find the angle.
How this angle adding plays with skew and flip we don't know.

asturur:
This is promising.
But does not solve the acoord/ocoord issue YET.
right? we are still keeping both.
I can work on that if we stabilize this.

asturur:
I don't think you should include text editing in the first PR.
The logic is much similar, but does not need to be done in the same pr.

src/controls.actions.js
```diff
@@ -231,7 +233,7 @@
         control = target.controls[transform.corner],
         zoom = target.canvas.getZoom(),
         padding = target.padding / zoom,
-        localPoint = target.toLocalPoint(new fabric.Point(x, y), originX, originY);
+        localPoint = target.normalizePoint(new fabric.Point(x, y), originX, originY);
```
asturur, line 236: i need to check if getLocalPoint gets used for other things.

src/mixins/itext_click_behavior.mixin.js
```diff
@@ -152,7 +152,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    */
   mouseUpHandler: function(options) {
     this.__isMousedown = false;
-    if (!this.editable || this.group ||
+    if (!this.editable ||
```
asturur, line 155: this is a strong shortcut to remove.
For which reason this check is removed?
To make possible use a corner of an object in a group? or use the object at all?

src/mixins/object_geometry.mixin.js
```diff
@@ -88,8 +88,15 @@
...
-      return arrayFromCoords(this._getCoords(absolute, calculate));
+    getCoords: function (absolute, calculate) {
+      var coords = arrayFromCoords(this._getCoords(absolute, calculate));
+      if (this.group) {
+        var t = this.group.calcTransformMatrix();
+        return coords.map(function (p) {
+          return util.transformPoint(p, t);
+        });
+      }
+      return coords;
```
asturur, line 99: the issue here is that if now you want to do some check in group coordinates, you can't anymore.
Is also true that maybe that is not useful.

src/shapes/group.class.js
```diff
@@ -131,7 +131,7 @@
     _updateObjectCoords: function(object, center) {
       var objectLeft = object.left,
           objectTop = object.top,
-          skipControls = true;
+          skipControls = false;
```
asturur, line 134: i think the whoe reason of skipControls was this one, probably it can be removed alltogether

asturur:
the render box/border is a bug that there is already now. let me fix it first and let's see if it also fixes your.

ShaMan123:
> This is promising. But does not solve the acoord/ocoord issue YET. right? we are still keeping both. I can work on that if we stabilize this.

Right.

@kirill-konshin made me rethink the need of relative coords

> It's very handy because if you'll need to sync a state of a group to server, then in case of relative coords you will only sync > > own group properties, not touching children.

> And in case of absolute coords all children will have to sync, so if group has 100 elements, it will be a huge update.

> _Originally posted by @kirill-konshin in https://github.com/fabricjs/fabric.js/issues/7598#issuecomment-1011351171_

src/shapes/group.class.js
ShaMan123: I thought so but was afraid to break.
I will remove.

src/mixins/object_geometry.mixin.js
ShaMan123: I am thinking of a method in `util`. Something like 
- `transformPointBetweenPlanes(from, to)`
- `getPointRelativeTo(originObject, destinationObject)` - here is it confusing because it can be relative to destination as a sibling or as a child.

src/mixins/itext_click_behavior.mixin.js
ShaMan123: use at all.
it's the same as `skipCorners`, put here not to handle the case.

src/controls.actions.js
ShaMan123: I checked (and didn't find, but didn't want to break) and there's another function that is called for a same purpose that is used by one method:
https://github.com/fabricjs/fabric.js/blob/794b6b4a9461b9e79b894784dcfbf4112c90fa7b/src/shapes/object.class.js#L1933

src/mixins/object_origin.mixin.js
```diff
@@ -129,16 +129,28 @@
...
+     * Returns the normalized point (rotated relative to center) in local coordinates
+     * @param {fabric.Point} point The point relative to canvas coordinate system
      * @param {String} originX Horizontal origin: 'left', 'center' or 'right'
      * @param {String} originY Vertical origin: 'top', 'center' or 'bottom'
      * @return {fabric.Point}
      */
-    toLocalPoint: function(point, originX, originY) {
-      var center = this.getCenterPoint(),
-          p, p2;
+    toLocalPoint: function (point, originX, originY) {
```
ShaMan123, line 138: I think this method isn't used anywhere. And I'm not sure it correct.

src/controls.actions.js
ShaMan123: > This one

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/pull/7599#r785279349_

src/mixins/object_geometry.mixin.js
ShaMan123: - `transformPointBetweenPlanes(point, originObject, destinationObject, originPlaneRelativity, destinationPlaneRelativity)`
`originPlaneRelativity` can be a boolean indicating if the point is a sibling or a child in relation to object

ShaMan123: It is a good util that will probably be useful in fabric and for devs

src/shapes/group.class.js
ShaMan123: bdbca70d

ShaMan123: I saw it's used only by StaticCanvas background/overlay Image

src/mixins/itext_click_behavior.mixin.js
ShaMan123: Tested and looks good + #7598

src/mixins/object_interactivity.mixin.js
```diff
@@ -194,11 +179,46 @@
...
 
+    /**
+     * Draws lines from a borders of an object's bounding box to controls that have `withConnection` property set.
+     * Requires public properties: width, height
+     * Requires public options: padding, borderColor
+     * @param {CanvasRenderingContext2D} ctx Context to draw on
+     * @return {fabric.Object} thisArg
+     * @chainable
+     */
+    drawControlsConnectingLines: function (ctx) {
```
ShaMan123, line 196: doesn't respect group

ShaMan123: https://github.com/fabricjs/fabric.js/blob/69020dfaeadf8c1b15b31778a0ad0238f46106d2/src/mixins/object_interactivity.mixin.js#L248
this is wrong and I can't get it fixed.
It should not skew the point and it should scale it to account for the skew.
I tried something but I can't get the translation to correct:
https://github.com/fabricjs/fabric.js/blob/9d8fce4a568efc1dafabaceb69017aba30777453/src/mixins/object_interactivity.mixin.js#L228

src/mixins/object_origin.mixin.js
asturur: well if is not used anywhere now we can trash it.

ShaMan123: d410e0da

src/mixins/object_geometry.mixin.js
ShaMan123: #7614

src/mixins/object_origin.mixin.js
ShaMan123: aacaa76a

asturur:
this PR is open into master.
That means that if we want to merge this we need to do it after the 5.0 release and give a week or two for the initial bugs that pops out on 5.0.
Can you move it on the fabric-6 branch and we merge it sooner? even right away to be honest.

ShaMan123:
merged into v6

ShaMan123:
Was I supposed to open a PR?
sorry

asturur:
Does not change much.
I can check back the diffs here and from the commit lists.
And fabric-v6 is not master.
Pr on non master branch are just for review and acknoledgment.

ShaMan123:
merged into v6.
closing this PR and will open a dedicated one for v6...

PR #7607, opened 2022-01-12
Prevent dividing by 0 in for touch gestures

rockerBOO:
In scenarios where the distance resolves to 0 as not moved from the same point. 

`start.x - self.x, start.y - self.y` where start and self are the same numbers will make distance 0. 

`start.distance = Math.sqrt(0 * 0 + 0 * 0)`


asturur:
This seems fine to me!

asturur (approved):

PR #7611, opened 2022-01-15
fix(fabric.Object) bounding box display when outside group

### LINKED ISSUE #7429: The controls for the skewed objects are in the wrong place.
Screen Shot 2021-10-12 at 16 32 08 PM

## Version
latest

## Test Case
https://jsfiddle.net/human_a/3bowe02j/1/

Nodejs or browser? Browser
Which browsers? All

## Steps to reproduce
In the fiddle select the rectangle/

## Expected Behavior
The controls must be according to the skew

I remember this was once reported in 2015, which was resolved at that time, something in the latest updates must have broken this again.

[Update]
It has started breaking in 4.6.0, if you set the version of fabric in the fiddle to 4.5.0 it works as it should

asturur:
closes #7429

The generic option angle can't be used when we draw the bounding box of an object with skewY, the reason is not super clear to me right now, but has probably to do with the fact that the calculation of the boudning box size do consider skewY while the decomposition removes skewY.

So the code is drawing a boudning box with this angle

image

the angle is wrong.
but this is because the angle is calculated for an object that has lost skewY after a reduction.

So the angle is correct for the same object, but with this different size

image

before code changes

image

after code changes

image


asturur:
image

First time i manage to use stylesOverride for something useful

asturur:
@ShaMan123 this add a test for the bug you are seeing in your PR.

asturur:
(the controls or the controlbox with skewY are not rendered correctly inside a group)

asturur:
Ideally to fix it properly, we stop to use strokeRect and we move between the 4 main coordinates with lines. period.

PR #7614, opened 2022-01-16
feat(util): transform utils

ShaMan123:
Making transformations easier and accessible
Added the following utils:
2. `sendPointToPlane`
3. `transformPointRelativeToCanvas`
4. `sendObjectToPlane`

https://codesandbox.io/s/fabric-app-forked-bipwk

TODO
- [x] tests

ezgif com-gif-maker (9)


ShaMan123:
Is this normal?
image

ShaMan123:
This is ready and tested!

ShaMan123:
I am a bit confused a regarding transformPoint util.
I run a test to make sure I didn't confuse the order of transformations

ShaMan123:
As far as I am concerned this can be merge.
Tested in live demo as well.
Tests pass de facto but have the inaccuracy of `composeMatrix/qrDecompose`

ShaMan123:
mergable

ShaMan123:
Perhaps the we could think of a better naming for the relation prop

asturur:
@ShaMan123 i m all for better utils.
Said so i want to focus on release of fabric 5.0, closing bugs, and new code refactor and your work on groups.
Try to focus on those things with me, and don't burn out with too much things.
I would eventually add more utils when those can be tree shaked by imports.

asturur:
>  is this normal?

Yes!
Math.sin and Math.cos aren't great.

When you work with 90 degree angles and you expect 1 or 0 for cos and sin, you often get either 0.0001e-10 or 0.99999999999999999 and you can't do much about it.
We wrapped cos and sin because for no angle or perfect rotations we wanted those things to do not happen.
People that were expecting top to be 3, were getting 2.99999999 that with a bit of scaling was enough to get blurry pixels.
But since the compose code uses math.atan2 and Math.PI those things definetely happen and a bit of loss of precisions happen.

ShaMan123:
@asturur I'll explain my motivation.
I have used `sendObjectToPlane` inexplicitly in the eraser PR (at least 3 times), the group PR (at least 4 times), the ICollection PR a few times and in my projects. The minion example uses it. It is very useful and abstracts a very complicated issue - transform matrices.
It is readable and easy to understand, so it's a part of the general move toward a better code base.
That's why I wrote it.

ShaMan123:
fb83a7152c8581d67b32caa9d6ba51fccfadb048

ShaMan123:
merged to v6

src/util/misc.js
```diff
@@ -805,6 +944,13 @@
         translateX: a[4],
         translateY: a[5]
       };
+      Object.defineProperty(options, 'matrix', {
+        value: a,
+        configurable: false,
+        enumerable: false,
+        writable: false
+      });
+      return options;
```
asturur, line 953: i don't think this is good.
Wha happens if i do:

```js
var a = decompose(matrix);
a.translateX = 20;
var b = compose(a);
```

if i hit the shortcut this is wrong.

asturur:
i don't mind adding those utils in the next version, but i want to work on the meaning of each and clearly make examples how those are used in applications.

src/util/misc.js
ShaMan123: Good point
fb83a7152c8581d67b32caa9d6ba51fccfadb048

ShaMan123:
I got confused and closed the wrong PR one. meant to close the group PR.

src/util/misc.js
```diff
@@ -290,8 +290,147 @@
...
+     * meaning that if `point` is equal to (0,0) it is positioned at object's center.
+     *
+     * `sibling` relation means `point` exists in the same coordinate plane as the object we relate to.
+     * In other words they both relate to the same (0,0) and agree on every point.
+     *
+     * @static
+     * @memberOf fabric.util
+     * @see {fabric.util.transformPointRelativeToCanvas} for transforming relative to canvas
+     * @param {fabric.Point} point
+     * @param {fabric.Object} sourceObject object that point currently relates to
```
asturur, line 379: in calcTransformationBetweenObjectPlanes sourceObject is optional. So it could be here too?

src/util/misc.js
```diff
@@ -290,8 +290,147 @@
...
+    getTransformMatrixByObject: function (object, relationToObject) {
+      if (relationToObject !== 'child' && relationToObject !== 'sibling') {
+        throw new Error('fabric.js: recieved bad argument ' + relationToObject);
+      }
+      return relationToObject === 'child' ?
+        object.calcTransformMatrix() :
+        object.group ?
+          object.group.calcTransformMatrix() :
+          fabric.iMatrix.concat();
+    },
```
asturur, line 319: I m confused about this utility here.
I can't match the name with the code, maybe i just don't get what `matrixByObject` means.

asturur:
remember to checkout the dist folder back to master.

asturur (dismissed):

src/util/misc.js
```diff
@@ -290,8 +290,147 @@
...
+     *
+     * `child` relation means target exists in the coordinate plane created by the object we relate to.
+     * In other words:
+     * - target is measured acoording to object's center point
+     * - target is drawn by object in the plane object creates
+     * - target is transformed by object but is unaware of it.
+     * - if target is positioned at (0,0) it is positioned at object's center.
+     *
+     * `sibling` relation means target exists in the same coordinate plane as the object we relate to.
+     * In other words they both relate to the same (0,0), agree on every point and are drawn in the same transformed plane.
```
asturur, line 337: We could add an explanation for simple people like me, making the example that you want to move one object from the canvas inside a group, or from a group to another, but without changing its perceived position and properties

src/util/misc.js
ShaMan123: We pass an object and it's relation to the plane you what and it returns the matrix of the plane.
It can be either the plane that object creates (for it's children) or the plane it is contained in (where it is a sibling)

ShaMan123:
I realized that it won't work on clip paths because clip paths aren't aware of their parent.
What do you suggest?

src/util/misc.js
ShaMan123: done

ShaMan123:
I have added `sendChildToPlane` for the sake of clipPath and other orphans
Added some examples
Ready to merge

src/util/misc.js
asturur: or maybe fabric.Object.calcPlaneTransformMatrix() for the sibling/container, while for the child we just call calcTransformMatrix()

asturur:
Merging this and fixing the unit test with a commit later.

asturur (approved):

src/util/misc.js
ShaMan123: sounds right to me

ShaMan123: I like this `fabric.Object.calcTransformMatrixByRelationship('sibling'|'child')`?

ShaMan123:
Now I think `sendObjectToPlane` and `sendChildToPlane` are pretty much the same.
For the sake of consistency I think of removing `sendObjectToPlane` and renaming `sendChildToPlane` to `sendObjectToPlane`

ShaMan123:
now, I'm refactoring

src/mixins/object_geometry.mixin.js
```diff
@@ -600,6 +600,22 @@
...
+     * 
+     * **WARNING**\
+     * Do **NOT** use with clip paths. 
+     * A clip path object isn't aware of it's containing plane (parent).
+     * This means this method will return the identity matrix, which is **WRONG**
+     * 
+     * @returns {Array} plane matrix relative to the coordinate plane created by a canvas
+     */
+    calcPlaneMatrix: function () {
+      return this.group ?
```
ShaMan123, line 615: Is there a way to fix this with clip path?

ShaMan123:
I have refactored this PR heavily because of the clip path (and other orphans) not aware of the fact they have parents.
It's sad.
It was a good util but clip path broke it.
It still has a vulnerability `getPlaneMatrix`, see my comment.
I was thinking to fix the by adding a `parent` property to clip path but then I remembered you said there is an option for objects to share a clip path so I backed off. 
Now it is more generic.

ShaMan123:
Maybe we can mark an object as a clipPath so then we can throw an error from `calcPlaneMatrix`
Here for example
https://github.com/fabricjs/fabric.js/blob/23ae1a784a4a1431c1446589487407c15b774dae/src/shapes/object.class.js#L1146

ShaMan123:
so all in all I think to remove `transformPointRelativeToCanvas` and standardize `Canvas.getPointer`
Thoughts?

src/mixins/object_geometry.mixin.js
ShaMan123: I decided to remove this method because it was safe to use because of clip path

asturur:
imho you are aiming at a moving target.
clipPath is full of edge cases, and probably should be simplified and not drive our efforts.

We should examine, does forbidding to share clipPaths block real use cases or make them extremely complicated?
if yes, can it be solved adding the clipPath to a group?

I have experienced many time following a refactor idea and the get stopped at the end by something it couldn't work because of requirement x.

Are you satisfied with this PR? should i merge it?
I would like to start looking at the group myself and extract the first parts to merge.

asturur (approved):

ShaMan123:
I am pretty satisified, perhaps I'll reworks this in the future

PR #7615, opened 2022-01-16
ci(): add auto-changelog package to quickly draft a changelog

asturur:
Add a package to use from the command line to produce new changelog lines for recent commits

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   },
   "license": "MIT",
   "scripts": {
+    "changelog": "auto-changelog -o change-output.md --unreleased-only",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -44,6 +44,7 @@
   },
   "license": "MIT",
   "scripts": {
+    "changelog": "auto-changelog -o change-output.md --unreleased-only",
```
asturur, line 47: this does not work great, or maybe is because fabric has some old untagged commits.
This command will give you a longer list with on top the latest commit.
Is still better than what i m doing now that is typing the changelog manually.

PR #7616, opened 2022-01-16
fix(svg_export): clip path

ShaMan123:
WIP

### DIFF
--- a/src/mixins/object.svg_export.js
+++ b/src/mixins/object.svg_export.js
@@ -180,8 +180,8 @@
     _createBaseClipPathSVGMarkup: function(objectMarkup, options) {
       options = options || {};
       var reviver = options.reviver,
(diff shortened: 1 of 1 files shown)


asturur:
There are 2 ways to add svg export tests for the change:

https://github.com/fabricjs/fabric.js/blob/master/test/visual/z_svg_export.js#L325-L337

or

https://github.com/fabricjs/fabric.js/blob/master/test/visual/z_svg_export.js#L303-L323

important is that you capture something that it totally breaks when the code changes.
Build a scene in which there is a clipPath situation that can't get exported today, but the difference needs to be enough to have the test failing without the builds.

asturur:
i see you have windows, let me know if you have troubles installing the canvas dependency to render under node.
Or i can write the test if you can give me the scene you want to render

PR #7623, opened 2022-01-19
Remove not working badges from README.md

jpomykala:
These badges are not working since a long time.


asturur:
Thank you @jpomykala

PR #7627, opened 2022-01-22
feat(fabric.StaticCanvas, fabric.Canvas): limit breaking changes

asturur:
- keep animation module optional
- avoid cancelling all animations of different canvases
- move cancelling to static canvas
- restore chanability for simple effects

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1038,6 +1038,7 @@
     /**
      * Centers object horizontally in the canvas
      * @param {fabric.Object} object Object to center horizontally
+     * @return {fabric.Canvas} thisArg
      */
     centerObjectH: function (object) {
       return this._centerObject(object, new fabric.Point(this.getCenter().left, object.getCenterPoint().y));
@@ -1046,6 +1047,8 @@
     /**
      * Centers object vertically in the canvas
      * @param {fabric.Object} object Object to center vertically
+     * @return {fabric.Canvas} thisArg
+     * @chainable
      */
     centerObjectV: function (object) {
       return this._centerObject(object, new fabric.Point(object.getCenterPoint().x, this.getCenter().top));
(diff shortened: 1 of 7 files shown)


asturur:
@ShaMan123 have a look

src/static_canvas.class.js
```diff
@@ -1746,6 +1762,10 @@
       }
       this.forEachObject(function(object) {
         object.dispose && object.dispose();
+        // animation module is still optional
+        if (fabric.runningAnimations) {
+          fabric.runningAnimations.cancelByTarget(object);
```
ShaMan123, line 1767: not sure this is good enough because what if there's a nested object being animated under a group?

ShaMan123: Give a few minutes I'm writing a function for canvas

ShaMan123: redundant because it's covered by `object.dispose`

ShaMan123: see #7629

asturur: oh ok i did not think of object.dispose right.

asturur: i want to puke, but if someone finds a good idea to animate an object clippath, when object is disposed that needs to be cancelled to. i ll add that.

asturur: no maybe the clippath is an edge case to explain. But if the clippath is shared across objects, its animation should be cacnceled manually.

ShaMan123: > i want to puke, but if someone finds a good idea to animate an object clippath, when object is disposed that needs to be cancelled to. i ll add that.

Funny stuff.

But I saw you merged with this line. It should be removed.

asturur:
did you not applied a commit that was deleting it?

src/static_canvas.class.js
asturur: oh ok your commit was adding the code to make this line useless.
But wasn't removing this line.
I'll do that

ShaMan123:
didn't.
opened PR #7631

PR #7630, opened 2022-01-22
BREAKING - remove deprecated code

asturur:
- selection event won't keep firing informations with the format of fabric 3 and lower ( use seelcted and deselected and not updated or target
- canvas. _addEventOptions method removed ( private and deprecated )
- events like `scaled` `skewed` `rotated` are removed. Event 'modified' will still have actionName inside
- removed deprecated and unused methods
  - Object.calcCoords
  - Object._calcDimensionsTransformMatrix
- circle startAngle and endAngle are now in angles and no more in radians
- group.realizeTransform has been removed
- utils.drawDashedLine removed
- utils.getBoundsOfArc removed
- utils.drawArc removed
- utils.fromArcToBeizers removed
- saved JSON with circles needs a reviver.


asturur:
image

seems clear to me i broke circle

asturur:
@ShaMan123 @melchiar this is the removal of things that were deprecated since version 4 or more.

PR #7631, opened 2022-01-22
fix(Animation): patch #7627

ShaMan123:
patch #7627

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1954,7 +1954,6 @@
     dispose: function () {
       if (fabric.runningAnimations) {
         fabric.runningAnimations.cancelByTarget(this);
-        this.clipPath && fabric.runningAnimations.cancelByTarget(this.clipPath);
       }
     }
   });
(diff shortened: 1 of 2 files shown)


src/shapes/object.class.js
```diff
@@ -1954,6 +1954,7 @@
     dispose: function () {
       if (fabric.runningAnimations) {
         fabric.runningAnimations.cancelByTarget(this);
+        this.clipPath && fabric.runningAnimations.cancelByTarget(this.clipPath);
```
asturur, line 1957: let's not do that.
Because otherwise we should dispose the clipPath too, while we decided that we can't decide for the developer.

ShaMan123, line 1957: @asturur do you want to use ENLIAVABLE_PROPS?

ShaMan123: Alright

ShaMan123: dismiss this then

ShaMan123:
mergable

ShaMan123:
@asturur this needs to be included in v5

ShaMan123:
> @asturur this needs to be included in v5

@asturur

PR #7633, opened 2022-01-26
feat(animation): Support a list of animation values

rockerBOO:
Allow lists of values to be supported in animate. 

```js
fabric.util.animate({
  startValue: [1, 2, 3],
  endValue: [2, 4, 6],
  onChange: function([a, b, c]) {
    canvas.zoomToPoint({x: b, y: c}, a)
    canvas.renderAll()
  }
});
```

Updated types to infer and allow the type to be passed (to allow `number | number []`)

```ts
interface IUtilAnimationOptions<T> {
	/**
	 * Starting value
	 */
	startValue?: T;
	/**
	 * Ending value
	 */
	endValue?: T;
	/**
	 * Value to modify the property by
	 */
	byValue?: T;
	/**
	 * Duration of change (in ms)
	 */
	duration?: number;
	/**
	 * Callback; invoked on every value change
	 */
	onChange?: (currentValue: T, finished: number, timePercentage: number) => void;
	/**
	 * Callback; invoked when value change is completed
	 */
	onComplete?: (endValue: T, finished: number, timePercentage: number) => void;
	/**
	 * Easing function
	 */
	easing?: Function;
}
interface IUtilAnimation {
	/**
	 * Changes value from one to another within certain period of time, invoking callbacks as value is being changed.
	 * @param [options] Animation options
	 */
	animate<T>(options?: IUtilAnimationOptions<T>): void;
}
```

### DIFF
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -6,6 +6,15 @@
   /**
    * @typedef {Object} AnimationOptions
    * Animation of a value or list of values.
+   * When using lists, think of something like this:
+   * fabric.util.animate({
+   *   startValue: [1, 2, 3],
+   *   endValue: [2, 4, 6],
+   *   onChange: function([a, b, c]) {
+   *     canvas.zoomToPoint({x: b, y: c}, a)
+   *     canvas.renderAll()
+   *   }
+   * });
    * @example
    * @property {Function} [onChange] Callback; invoked on every value change
    * @property {Function} [onComplete] Callback; invoked when value change is completed
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -153,8 +153,8 @@
           onComplete = options.onComplete || noop,
           easing = options.easing || defaultEasing,
           isMany = 'startValue' in options ? options.startValue.length > 0 : false,
-          startValue = 'startValue' in options ? (isMany ? options.startValue : options.startValue) : 0,
-          endValue = 'endValue' in options ? (isMany ? options.endValue.slice() : options.endValue) : 100,
+          startValue = 'startValue' in options ? options.startValue : 0,
+          endValue = 'endValue' in options ? options.endValue : 100,
           byValue = options.byValue || (isMany ? startValue.map(function(value, i) {
             return endValue[i] - startValue[i];
           }) : endValue - startValue);
@@ -188,7 +188,7 @@
           context.durationRate = 1;
           //  execute callbacks
           onChange(isMany ? endValue.slice() : endValue, 1, 1);
-          onComplete(isMany ? endValue.slice() : endValue, 1, 1);
+          onComplete(endValue, 1, 1);
           removeFromRegistry();
           return;
         }
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -170,7 +170,6 @@
             }) : easing(currentTime, startValue, byValue, duration),
             valuePerc = isMany ? Math.abs((current[0] - startValue[0]) / byValue[0])
               : Math.abs((current - startValue) / byValue);
-        console.log('current', current);
         //  update context
         context.currentValue = current;
         context.completionRate = valuePerc;
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -152,9 +152,9 @@
           abort = options.abort || noop,
           onComplete = options.onComplete || noop,
           easing = options.easing || defaultEasing,
-          startValue = 'startValue' in options ? options.startValue : 0,
-          endValue = 'endValue' in options ? options.endValue : 100,
-          isMany = 'startValue' in options ? startValue.length > 0 : false,
+          isMany = 'startValue' in options ? options.startValue.length > 0 : false,
+          startValue = 'startValue' in options ? (isMany ? options.startValue : options.startValue) : 0,
+          endValue = 'endValue' in options ? (isMany ? options.endValue.slice() : options.endValue) : 100,
           byValue = options.byValue || (isMany ? startValue.map(function(value, i) {
             return endValue[i] - startValue[i];
           }) : endValue - startValue);
@@ -171,29 +171,29 @@
             valuePerc = isMany ? Math.abs((current[0] - startValue[0]) / byValue[0])
               : Math.abs((current - startValue) / byValue);
         //  update context
-        context.currentValue = current;
+        context.currentValue = isMany ? current.slice() : current;
         context.completionRate = valuePerc;
         context.durationRate = timePerc;
         if (cancel) {
           return;
         }
-        if (abort(current, valuePerc, timePerc)) {
+        if (abort(isMany ? current.slice() : current, valuePerc, timePerc)) {
           removeFromRegistry();
           return;
         }
         if (time > finish) {
           //  update context
-          context.currentValue = endValue;
+          context.currentValue = isMany ? endValue.slice() : endValue;
           context.completionRate = 1;
           context.durationRate = 1;
           //  execute callbacks
-          onChange(endValue, 1, 1);
-          onComplete(endValue, 1, 1);
+          onChange(isMany ? endValue.slice() : endValue, 1, 1);
+          onComplete(isMany ? endValue.slice() : endValue, 1, 1);
           removeFromRegistry();
           return;
         }
         else {
-          onChange(current, valuePerc, timePerc);
+          onChange(isMany ? current.slice() : current, valuePerc, timePerc);
           requestAnimFrame(tick);
         }
       })(start);
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -5,14 +5,20 @@
 
   /**
    * @typedef {Object} AnimationOptions
-   * @property {Function} [options.onChange] Callback; invoked on every value change
-   * @property {Function} [options.onComplete] Callback; invoked when value change is completed
-   * @property {number | number[]} [options.startValue=0] Starting value
-   * @property {number | number[]} [options.endValue=100] Ending value
-   * @property {number | number[]} [options.byValue=100] Value to modify the property by
-   * @property {Function} [options.easing] Easing function
-   * @property {Number} [options.duration=500] Duration of change (in ms)
-   * @property {Function} [options.abort] Additional function with logic. If returns true, animation aborts.
+   * Animation of a value or list of values.
+   * @example
+   * @property {Function} [onChange] Callback; invoked on every value change
+   * @property {Function} [onComplete] Callback; invoked when value change is completed
+   * @example
+   * // Note: startValue, endValue, and byValue must match the type
+   * var animationOptions = { startValue: 0, endValue: 1, byValue: 0.25 }
+   * var animationOptions = { startValue: [0, 1], endValue: [1, 2], byValue: [0.25, 0.25] }
+   * @property {number | number[]} [startValue=0] Starting value
+   * @property {number | number[]} [endValue=100] Ending value
+   * @property {number | number[]} [byValue=100] Value to modify the property by
+   * @property {Function} [easing] Easing function
+   * @property {Number} [duration=500] Duration of change (in ms)
+   * @property {Function} [abort] Additional function with logic. If returns true, animation aborts.
    *
    * @typedef {() => void} CancelFunction
    *
@@ -122,6 +128,10 @@
    * Changes value from one to another within certain period of time, invoking callbacks as value is being changed.
    * @memberOf fabric.util
    * @param {AnimationOptions} [options] Animation options
+   * @example
+   * // Note: startValue, endValue, and byValue must match the type
+   * fabric.util.animate({ startValue: 0, endValue: 1, byValue: 0.25 })
+   * fabric.util.animate({ startValue: [0, 1], endValue: [1, 2], byValue: [0.25, 0.25] })
    * @returns {CancelFunction} cancel function
    */
   function animate(options) {
--- a/src/util/animate.js
+++ b/src/util/animate.js
@@ -187,9 +187,8 @@
           context.completionRate = 1;
           context.durationRate = 1;
           //  execute callbacks
-          var end = isMany ? endValue.slice() : endValue;
-          onChange(end, 1, 1);
-          onComplete(end, 1, 1);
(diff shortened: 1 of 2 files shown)


asturur:
I get this.
This is a relatively small change and seems fine to me.
Can you write a test to cover it?

rockerBOO:
Added test. 

---

I can't seem to get the tests running locally, but seems to pass here. 

`node 14.18.1`

`npm run test`

```
 ---
  message: "Died on test #1     at Object.test (/home/rockerboo/code/fabric.js/node_modules/qunit/qunit/qunit.js:2900:5): fabric is not defined"
  severity: failed
  actual  : null
  expected: undefined
  stack: |
    ReferenceError: fabric is not defined
        at Object.<anonymous> (/home/rockerboo/code/fabric.js/test/unit/animation.js:338:5)
  ...
  ---
  message: afterEach failed on animate with list of values: fabric is not defined
  severity: failed
  actual  : null
  expected: undefined
  stack: |
    ReferenceError: fabric is not defined
        at Object.afterEach (/home/rockerboo/code/fabric.js/test/unit/animation.js:4:20)
  ...
```

`npx testem`

```
global failure
    ✘ Error: Failed to load file test/node_test_setup.js
    ReferenceError: window is not defined

        |
          ReferenceError: window is not defined
```

and 

```
➜ npm install

up to date, audited 534 packages in 2s
```

Not a big worry but not sure why, so any hints would be helpful.

rockerBOO:
Thanks @ShaMan123 ! Got the tests to run. I think the build did it but that didn't make sense to me as it should already be built. Anyways thanks!

ShaMan123:
Happy to help.
I get the same strange behavior.

ShaMan123:
This PR makes animation even more powerful.

test/unit/animation.js
```diff
@@ -332,6 +332,24 @@
...
+  QUnit.test('animate with list of values', function(assert) {
+    var done = assert.async();
+
+    fabric.util.animate({
+      startValue: [1, 2, 3],
+      endValue: [2, 4, 6],
+      onValue: [1, 1, 1],
+      duration: 96,
+      onChange: function(currentValue) {
+        assert.equal(currentValue.length, 3)
```
ShaMan123, line 344: @rockerBOO could you add an assertion for `fabric.runningAnimations`
It's in the test file.
You should check that the values have updated accordingly there.

ShaMan123: And maybe we want to concat the array so that it isn`t the same ref used by animate.
Hmm....

ShaMan123: The same applies to the callbacks onChange etc.

ShaMan123: I would add a test:

```js
fabric.util.animate({
  startValue: [1, 2, 3],
  endValue: [2, 4, 6],
  onChange: function([a, b, c]) {
    a=200;
  },
    onComplete([a]) {
     assert.equal(a, 2);
    }
});

```

rockerBOO: Added a few more tests for checking if fabric.runningAnimations is valid and that the values are valid. The interstitial values would be hard to test but I think this checks for what could be valid values.

rockerBOO: Whoops missed some of this.

> ```js
> assert.lt(a, 2);
>     a=200;
> ```

What does this mean? I am not sure. Is what I have adequate? I don't think there should be any sharing of arrays as I'm using map to transpose the easing values.  Let me know. Thanks!

ShaMan123: As you can see here:
https://github.com/fabricjs/fabric.js/blob/7db2515f09c4950a31fe0b1db1dddedf2fcd0240/src/util/animate.js#L174

the array you've created will be set to `runningAnimations`. This isn't good, it needs to be copied. Same goes for values passed to the callbacks. Or else a change in the callback will change inner logic of the animation tick - undesirable!

That's what we need to write a test for both `runningAnimations` and the callbacks - making sure the arrays aren't equal, only `deepEqual`



> Whoops missed some of this.
> 
> > ```js
> > assert.lt(a, 2);
> >     a=200;
> > ```
> 
> What does this mean? I am not sure. Is what I have adequate? I don't think there should be any sharing of arrays as I'm using map to transpose the easing values. Let me know. Thanks!

This bit of code mutates the values passed from the animation driver to the callback.
If all goes well the mutation will live only inside the callback so the next tick will assert that `a` is `lt` (lesser than) 2.
If it's 200 then the arrays are shared.

ShaMan123:
```js
assert.equal(fabric.runningAnimations.length, 1, 'runningAnimations should not be empty');
var context=fabric.runningAnimations[0];
// check that context is deep equal to the values but that the arrays are not the same object

```

rockerBOO: Ok I think I got it all working now. Thanks!

src/util/animate.js
```diff
@@ -171,29 +171,29 @@
...
-        if (abort(current, valuePerc, timePerc)) {
+        if (abort(isMany ? current.slice() : current, valuePerc, timePerc)) {
           removeFromRegistry();
           return;
         }
         if (time > finish) {
           //  update context
-          context.currentValue = endValue;
+          context.currentValue = isMany ? endValue.slice() : endValue;
           context.completionRate = 1;
```
ShaMan123, line 187:
```suggestion
          context.completionRate = isMany ? new Array(current.length).fill(1) : 1;
```

src/util/animate.js
```diff
@@ -171,29 +171,29 @@
...
         //  update context
-        context.currentValue = current;
+        context.currentValue = isMany ? current.slice() : current;
         context.completionRate = valuePerc;
         context.durationRate = timePerc;
         if (cancel) {
           return;
         }
-        if (abort(current, valuePerc, timePerc)) {
+        if (abort(isMany ? current.slice() : current, valuePerc, timePerc)) {
```
ShaMan123, line 180: redundant.
My comment was misleading.

src/util/animate.js
```diff
@@ -171,29 +171,29 @@
...
-          onChange(endValue, 1, 1);
-          onComplete(endValue, 1, 1);
+          onChange(isMany ? endValue.slice() : endValue, 1, 1);
+          onComplete(isMany ? endValue.slice() : endValue, 1, 1);
           removeFromRegistry();
           return;
         }
         else {
-          onChange(current, valuePerc, timePerc);
+          onChange(isMany ? current.slice() : current, valuePerc, timePerc);
```
ShaMan123, line 196: redundant

src/util/animate.js
```diff
@@ -171,29 +171,29 @@
...
           //  update context
-          context.currentValue = endValue;
+          context.currentValue = isMany ? endValue.slice() : endValue;
           context.completionRate = 1;
           context.durationRate = 1;
           //  execute callbacks
-          onChange(endValue, 1, 1);
-          onComplete(endValue, 1, 1);
+          onChange(isMany ? endValue.slice() : endValue, 1, 1);
+          onComplete(isMany ? endValue.slice() : endValue, 1, 1);
```
ShaMan123, line 191: here we need to slice only once
and the second argument needs to be any array as well in case of `isMany` I think.

ShaMan123 (changes requested):
@rockerBOO sorry my comment was misleading.
> As you can see here:
> 
> https://github.com/fabricjs/fabric.js/blob/7db2515f09c4950a31fe0b1db1dddedf2fcd0240/src/util/animate.js#L174
> 
> the array you've created will be set to `runningAnimations`. This isn't good, it needs to be copied. Same goes for values passed to the callbacks. Or else a change in the callback will change inner logic of the animation tick - undesirable!
> 
> That's why we need to write a test for both `runningAnimations` and the callbacks - making sure the arrays aren't equal, only `deepEqual`
> 
> > Whoops missed some of this.
> > > ```js
> > > assert.lt(a, 2);
> > >     a=200;
> > > ```
> > 
> > 
> > What does this mean? I am not sure. Is what I have adequate? I don't think there should be any sharing of arrays as I'm using map to transpose the easing values. Let me know. Thanks!
> 
> This bit of code mutates the values passed from the animation driver to the callback. If all goes well the mutation will live only inside the callback so the next tick will assert that `a` is `lt` (lesser than) 2. If it's 200 then the arrays are shared.


Array sharing happens only between the callbacks and `fabric.runningAnimations` so we only need to slice the arrays assigned to `runningAnimations`.
The only other case is calling `onChange` and `onComplete` one after the other.
The rest are unnecessary as I see it.
I have a question: isn't `valuePerc` the same for all values? if so we don't need an array for that (and better of because if we do we need to pass array in callbacks and for `runningAnimations`

ShaMan123:
On the other hand, a better solution would be to make `runningAnimations.currentValue` a getter that returns a fresh copy of the array of values. This will keep it immutable as well.
Sounds the way to go forward. And makes all this slicing we've discussed redundant apart from the subsequent call to `onChange` `onComplete`.

rockerBOO:
> I have a question: isn't `valuePerc` the same for all values? if so we don't need an array for that (and better of because if we do we need to pass arrays for the callbacks and for `runningAnimations`)

valuePerc should be the same for each element in the array so I pick the first element to calculate it. 

Updated the requested values.

src/util/animate.js
rockerBOO: In this case we need separate because mutating the value (if shared) will mutate for the other (`onComplete`)

ShaMan123:
```suggestion
          onChange(isMany ? endValue.slice() : endValue, 1, 1);
          onComplete(endValue, 1, 1);
```

Isn't this enough?

src/util/animate.js
```diff
@@ -152,20 +152,26 @@
           abort = options.abort || noop,
           onComplete = options.onComplete || noop,
           easing = options.easing || defaultEasing,
-          startValue = 'startValue' in options ? options.startValue : 0,
-          endValue = 'endValue' in options ? options.endValue : 100,
-          byValue = options.byValue || endValue - startValue;
+          isMany = 'startValue' in options ? options.startValue.length > 0 : false,
+          startValue = 'startValue' in options ? (isMany ? options.startValue : options.startValue) : 0,
+          endValue = 'endValue' in options ? (isMany ? options.endValue.slice() : options.endValue) : 100,
```
ShaMan123, line 157: These 2 lines are redundant as well I think

src/util/animate.js
ShaMan123: ?

rockerBOO:
Ok, I think we have resolve them all now and the end value isn't mutating. Thanks for all the reviews.

ShaMan123 (dismissed):
Great!

src/util/animate.js
```diff
@@ -7,17 +7,17 @@
    * @typedef {Object} AnimationOptions
    * @property {Function} [options.onChange] Callback; invoked on every value change
    * @property {Function} [options.onComplete] Callback; invoked when value change is completed
-   * @property {Number} [options.startValue=0] Starting value
-   * @property {Number} [options.endValue=100] Ending value
-   * @property {Number} [options.byValue=100] Value to modify the property by
+   * @property {number | number[]} [options.startValue=0] Starting value
+   * @property {number | number[]} [options.endValue=100] Ending value
+   * @property {number | number[]} [options.byValue=100] Value to modify the property by
```
asturur, line 12: Should we write somewhere here or on top of the block that:
- this is done to animate arrays, in particular matrices ( unless you have other examples )
- that you have to be consistent, either all arrays or all single numbers? ( if you have to. i m not sure )

test/unit/animation.js
```diff
@@ -332,6 +332,34 @@
...
   });
 
+  QUnit.test('animate with list of values', function(assert) {
+    var done = assert.async();
+    assert.expect(52)
+
+    fabric.util.animate({
+      startValue: [1, 2, 3],
+      endValue: [2, 4, 6],
+      onValue: [1, 1, 1],
```
asturur, line 342: is this byValue?

test/unit/animation.js
```diff
@@ -332,6 +332,34 @@
     }, 1000);
   });
 
+  QUnit.test('animate with list of values', function(assert) {
+    var done = assert.async();
+    assert.expect(52)
```
asturur, line 337: what is this?

asturur:
I'm ready to merge when we address the small issues i pointed out.
I also wonder if this could replace animateColor at this point... not sure.

test/unit/animation.js
rockerBOO: This confirms the number of assertions should be run in this test, so we make sure we are running all the async tests before finishing.

test/unit/animation.js
rockerBOO: Updated to byValue.

src/util/animate.js
rockerBOO: Added an example note that should clarify that they should be consistent and also the values they can take. Currently if the start value is an array then all the other values `startValue`, `endValue`, and `byValue` should match.

ShaMan123:
This can also replace object's animation mixin

test/unit/animation.js
asturur: I had no idea this was a feature.
So this test has 52 assertion?

asturur (approved):

asturur:
Needs to be backported to 5.0.0 branch in case some middle release will come up

asturur:
(i ll do it with cherry pick)

ShaMan123:
@asturur if you're here merge #7631 as it is stabalizes the work on animations

ShaMan123:
I see tests are failing in other environments as well.

rockerBOO:
Its possible the timing would not be consistent enough. We can remove this assertion as different environments may/do support different timing structures or delays. (`requestionAnimationFrame` vs `setTimeout`)

ShaMan123:
Seems best

asturur:
i think we broke color animaton with this change.
Color gets animated with an array [r,g,b,a] but then has special rule for `a`

ShaMan123:
can you point it out ?
Looking at `animateColor` it calculates rgba string

asturur:
yes but taking internally an array, now i receive only single color channel values inside the calculate color

PR #7635, opened 2022-01-26
fix(tests): windows paths

ShaMan123:
Fixed the annoying tests that failed because of path resolution.
I've changed tests to run on the basename of the path rather than the entire ugly path.


ShaMan123:
ready

asturur:
I think i gave up with windows development of JS immediately when i discovered fabricJS.
The tooling was automatic on linux and i just didn't bother.

PR #7638, opened 2022-01-28
Fix SVG's default viewBox when width/height is specified.

TerrenceSun:
When setting width/height options for toSVG, it should use specified
value for default viewBox calculation.

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -1327,7 +1327,7 @@
     _setSVGHeader: function(markup, options) {
       var width = options.width || this.width,
           height = options.height || this.height,
-          vpt, viewBox = 'viewBox="0 0 ' + this.width + ' ' + this.height + '" ',
+          vpt, viewBox = 'viewBox="0 0 ' + width + ' ' + height + '" ',
           NUM_FRACTION_DIGITS = fabric.Object.NUM_FRACTION_DIGITS;
 
       if (options.viewBox) {
@@ -1343,8 +1343,8 @@
           viewBox = 'viewBox="' +
                   toFixed(-vpt[4] / vpt[0], NUM_FRACTION_DIGITS) + ' ' +
                   toFixed(-vpt[5] / vpt[3], NUM_FRACTION_DIGITS) + ' ' +
-                  toFixed(this.width / vpt[0], NUM_FRACTION_DIGITS) + ' ' +
-                  toFixed(this.height / vpt[3], NUM_FRACTION_DIGITS) + '" ';
+                  toFixed(width / vpt[0], NUM_FRACTION_DIGITS) + ' ' +
+                  toFixed(height / vpt[3], NUM_FRACTION_DIGITS) + '" ';
         }
       }
 


asturur:
Hi @TerrenceSun can you pleas show how the toSVG method changes from before and after the fix with this change?
thanks!

TerrenceSun:
> Hi @TerrenceSun can you pleas show how the toSVG method changes from before and after the fix with this change?
> thanks!

Usually we just call toSVG without specifying width or height, but when the content is not as same size as the canvas, it would be better to calculate a real height and width value and pass that to toSVG.

With the old code, view box stills use canvas size which is bigger than real size, and this would shrink the image in the end.

asturur:
i was asking you to add actual SVG + screenshots of some use cases

TerrenceSun:
> i was asking you to add actual SVG + screenshots 
of some use cases

Below are SVG results from a 800x800 canvas with user placed drawings. The picture size is 800x800 for 1, and 261x146.2 for 2 and 3. User may import SVG file into other document file, using 1 is not a good way for this case, as user needs to do a second resize to remove the empty space in the SVG file. 2 has the desired width and height but has small image because of the viewBox is still 800x800. 3 has the desired size and corrent content.

1. toSVG without options, the usually way.
    canvas_without_options
2. the old toSVG with width and height options of drawing contents
   canvas_old
3. result of 2. with this fix
   canvas

asturur:
I see what you want.
You want to use width and height to crop it.
Width and height in SVG do a different thing, and the way to crop it is exactly the bounding box.
Let me check if there is an option to pass a custom bounding box, if there isn't we should add it and you can open a pr with that.

TerrenceSun:
Yes, bounding box is the way of doing any crop. Changing the width and height only does a right and bottom crop, which is good enough for my case.  I will also check how to do bounding box in toSVG, thanks.

发自我的iPhone

TerrenceSun:
@asturur Would you please give me a hint on how to use bouding box to crop empty space in svg? I searched but found nothing usefull.

Thanks

asturur:
Your fix is to pass width/height to the viewbox width and height. Correct?

When you call toSVG() the options also take a `viewbox = { x, y, width, height }`  option.

The first step is that you can pass width and height also to the viewbox option and get the same result.

x and y would represent which is the point to that will be the SVG 0,0.

To make an example, if you  have a single object in the canvas and when you go to make the SVG you pass the object top and left to the viewbox  y/x you should remove the empty part to the left of the object.

In case the object is rotated you have to do some extra calculation

TerrenceSun:
> Your fix is to pass width/height to the viewbox width and height. Correct?

More precisely, if user pass width/height only, then also pass it to viewbox. This may fit one of the most usages, which is cropping the right and bottom space of the canvas. 

> When you call toSVG() the options also take a `viewbox = { x, y, width, height }` option.
>
> The first step is that you can pass width and height also to the viewbox option and get the same result.
> 

That's correct. It works this way as old code process these two options independently.

Usually we need to change width/height of viewport and viewbox at the same time. Because, only changing viewbox or viewport would lead to zoom effect.

This update only does a slight improve of the default behavior when user changing width/height without changing viewbox, not kind of any bug fix.

asturur:
Yes but this would be a behaviour change and actually removing the ability to pass width and height independently of the viewbox.

Maybe what we need is just a good explanation on how to use those options to obtain different effects

TerrenceSun:
Well, maybe we can add a new boudingbox option to toSVG, to make the cropping thing more clearly.

asturur:
the viewbox is the boundingbox, is exactly that.

TerrenceSun:
No need to do the update. Close it. Thanks.

asturur:
Did you manag to use the current options to achieve what you needed?

TerrenceSun:
@asturur 
Yes, setting both `viewBox`, `width`, and `height` can do the cropping correctly.
For say, if we want to do a cropping with bb {left, top, width, height} then call toSVG with options:
`{width: bb.width, 
  height: bb.height,
  viewBox: {
    x: bb.left,
    y: bb.right, 
    width: bb.width, 
    height: bb.height}}`

Issue #7605, opened 2022-01-12
Polygon points editing with `strokeUniform: true`


goodbot13:
Set strokeUniform: true on polygon.
Scale it.
Try to edit points - polygon moves.

Source copied from http://fabricjs.com/custom-controls-polygon

## Version
4.6.0

## Test Case
https://jsfiddle.net/5jtu48n2/

goodbot13:
> please provide an immediate repro

added. https://jsfiddle.net/5jtu48n2/

ShaMan123:
You are right.
The example didn't take that into mind.
Welcome to PR.
It is probably a confusion between canvas/object coordinate plane (many bugs like this are)
Closing.

ShaMan123:
So if understand correctly it's the calculation of `absolutePoint` and/or `setPositionByOrigin`.
Something there doesn't account for scaling.
But it's a guess.

asturur:
Issues like this can be labelled as DOCS so they don't get harvested by the stale bot

asturur:
yeah that example can use so much improvements

goodbot13:
> yeah that example can use so much improvements

can we have some published? :)

ShaMan123:
#7029 duplicate? Is this a bug?

asturur:
is labeled as docs and probably the example should be fixed for the sake of showing that it should be doable

goodbot13:
I came up with this solution.
The original source from fabric.js 4.6.0
```
/*
 * Calculate object dimensions from its properties
 * @private
 * @return {Object} .x width dimension
 * @return {Object} .y height dimension
 */
_getNonTransformedDimensions: function() {
  var strokeWidth = this.strokeWidth,
      w = this.width + strokeWidth,
      h = this.height + strokeWidth;
  return { x: w, y: h };
}
```

changed to this:
```
/*
 * Calculate object dimensions from its properties
 * @private
 * @return {Object} .x width dimension
 * @return {Object} .y height dimension
 */
_getNonTransformedDimensions: function() {
    var strokeWidth = this.strokeWidth,
        strokeUniform = this.strokeUniform,
        w = this.width + strokeUniform ? (strokeWidth / this.scaleX) : strokeWidth,
        h = this.height + strokeUniform ? (strokeWidth / this.scaleY) : strokeWidth;
    return { x: w, y: h };
},
```

And everything works fine.

ShaMan123:
[...]

PR #7639, opened 2022-01-28
fix(geometry): nonTransformedDimensions + `strokeUniform`

goodbot13:
#7605

### DIFF
--- a/src/mixins/object_geometry.mixin.js
+++ b/src/mixins/object_geometry.mixin.js
@@ -609,8 +609,9 @@
      */
     _getNonTransformedDimensions: function() {
       var strokeWidth = this.strokeWidth,
-          w = this.width + strokeWidth,
-          h = this.height + strokeWidth;
+        strokeUniform = this.strokeUniform,
+        w = this.width + strokeUniform ? (strokeWidth / this.scaleX) : strokeWidth,
+        h = this.height + strokeUniform ? (strokeWidth / this.scaleY) : strokeWidth;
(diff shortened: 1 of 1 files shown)


ShaMan123:
Could you run `npm run lint -- --fix` and commit the changes (if it makes a giant commit first stash all changes and restore - whitespace OS diffs)
We want to make contributing easier but for now it is what it is.

asturur:
This has been proposed in the past and hasn't been done because it breaks other things.
when we talk of non transformed dimensions, the scale is not taken in consideration. is the size of the object at scale 1. so dividing by scale would bring to strange results. Getting `nonTransformedDimensions` of a rect when it has scale 5 or 1 should give you the same result.

All tests break, and that could be that because they are made to reflect the old behaviour, but i m not confident to accept this change because it fixes a demo.

Does it fix something else?

Let's check it more broadly. SVG import tests could be affected by this for example.

ShaMan123:
I think there's a need to clearly define `strokeUniform`.
I understand it remains the same size no matter the scale.
If so is it part of non transformed dimensions?

asturur:
strokeUniform would mean that the stroke is not affected by the current transformation.
so if you deform a path stretching it on a single axis, the stroke will stay of the same size instead of being larger in some directions.
Is not the feature wrong but our stack of transform to find the size.

asturur:
This behaviour needs to be fixed, just to be clear, there are multiple open issues for it.

PR #7657, opened 2022-02-06
MAJOR feat(fabric) remove callbacks in for Promise support

### LINKED ISSUE #3684: Promises support
Currently, something like `fabric.Image.fromURL` only supports callbacks. Usage example from the tutorial:

```js
fabric.Image.fromURL('/assets/pug.jpg', function(img) {
  var img1 = img.scale(0.1).set({ left: 100, top: 100 });

  fabric.Image.fromURL('/assets/pug.jpg', function(img) {
    var img2 = img.scale(0.1).set({ left: 175, top: 175 });

    fabric.Image.fromURL('/assets/pug.jpg', function(img) {
      var img3 = img.scale(0.1).set({ left: 250, top: 250 });

      canvas.add(new fabric.Group([ img1, img2, img3], { left: 200, top: 200 }))
    });
  });
});
```

My request would be to add Promises support out of the box for `fromURL`, and any other asynchronous calls in the library, such that you could instead write the following:

```js
Promise.all([
    fabric.Image.fromURL('/assets/pug.jpg'),
    fabric.Image.fromURL('/assets/pug.jpg'),
    fabric.Image.fromURL('/assets/pug.jpg')
]).then(([img1, img2, img3]) => {
    let transformedImg1 = img1.scale(0.1).set({ left: 100, top: 100 });
    let transformedImg2 = img2.scale(0.1).set({ left: 175, top: 175 });
    let transformedImg3 = img3.scale(0.1).set({ left: 250, top: 250 });

    canvas.add(new fabric.Group([ transformedImg1, transformedImg2, transformedImg3], { left: 200, top: 200 }))
});
```

... or even something like this, using Bluebird:

```js
Promise.all([
    fabric.Image.fromURL('/assets/pug.jpg'),
    fabric.Image.fromURL('/assets/pug.jpg'),
    fabric.Image.fromURL('/assets/pug.jpg')
]).map((img, i) => {
    let offset = 25 + (i * 75);
    return img.scale(0.1).set({ left: offset, top: offset })
}).then(([img1, img2, img3]) => {
    canvas.add(new fabric.Group([ img1, img2, img3], { left: 200, top: 200 }))
});
```

(Of course it doesn't make much sense to load the same image three times like this, I'm just translating the tutorial example verbatim for demonstration purposes.)

__Reasoning for the feature request:__ All the usual benefits of Promises; easier composition, more reliable error handling, no chance of accidental double callbacks, and so on, and so forth.

__Implementation and consequences:__ This does not need to be a breaking change; an approach similar to Bluebird's [`nodeify`/`asCallback` ](http://bluebirdjs.com/docs/api/ascallback.html) could be taken, where a Promise is returned by default, and a callback is only called if one is specified.

I would *not* recommend actually using Bluebird in this case, for two reasons:

1. Bundle size. Bluebird is relatively big, and it probably doesn't make sense to throw in all of Bluebird for this relatively small change, especially given that fabric.js is browser-oriented.
2. Bluebird's `.asCallback` assumes an error-first callback, but `fromURL` doesn't use those; there is no first `errror` argument.

As for browser support; there are roughly two options here:

1. Include an ES6 Promises shim bundled into fabric.js.
2. Make the "nodeification" logic dependent on whether there is an ES6 Promise implementation available in the environment; if yes, return a Promise, if no, return nothing. This puts the burden on the user to polyfill `global.Promise` if they want to use Promises in all environments, without affecting non-Promise-using users.

The second option is probably preferable from a bundle size and "interference with other users" point of view, but is a bit more complex implementation-wise.

I'd say that overall implementation complexity and work required for somebody who knows the fabric.js codebase would be __very low__, since it's a generalizable conversion - but one that is much easier and more reliable to do from within the library than on the outside. Unfortunately I don't really know much about the fabric.js codebase, or I would've outright submitted a PR...

(Of course, I'd be happy to answer any questions about the Promises aspect of this, to make implementation go smoother; implementation details, common usage patterns, and so on.)

EDIT: Fixed example to actually be equivalent. More caffeine is required...

asturur:
closes #3684

This is not ready yet, all tests are broken for sure, probably the code is broken too.
I gave a first pass at swapping out callbacks for promises but there is still a lot to do.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,7 @@
     "browser": true
   },
   "globals": {
-    "ActiveXObject": true,
+    "Promise": true,
     "define": true,
     "eventjs": true,
     "exports": true,
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -559,27 +559,14 @@
    * @static
    * @memberOf fabric.Group
    * @param {Object} object Object to create a group from
-   * @param {Function} [callback] Callback to invoke when an group instance is created
+   * @returns {Promise<fabric.Group>}
    */
-  fabric.Group.fromObject = function(object, callback) {
+  fabric.Group.fromObject = function(object) {
     var objects = object.objects,
         options = fabric.util.object.clone(object, true);
     delete options.objects;
-    if (typeof objects === 'string') {
-      // it has to be an url or something went wrong.
-      fabric.loadSVGFromURL(objects, function (elements) {
-        var group = fabric.util.groupSVGElements(elements, object, objects);
-        group.set(options);
-        callback && callback(group);
-      });
-      return;
-    }
-    fabric.util.enlivenObjects(objects, function (enlivenedObjects) {
-      var options = fabric.util.object.clone(object, true);
-      delete options.objects;
-      fabric.util.enlivenObjectEnlivables(object, options, function () {
-        callback && callback(new fabric.Group(enlivenedObjects, options, true));
-      });
+    return fabric.util.enlivenObjects(objects).then(function (enlivenedObjects) {
+      return fabric.Group(enlivenedObjects, options, true);
     });
   };
 
--- a/src/shapes/image.class.js
+++ b/src/shapes/image.class.js
@@ -700,26 +700,25 @@
    * Creates an instance of fabric.Image from its object representation
    * @static
    * @param {Object} object Object to create an instance from
-   * @param {Function} callback Callback to invoke when an image instance is created
+   * @returns {Promise<fabric.Image>}
    */
-  fabric.Image.fromObject = function(_object, callback) {
+  fabric.Image.fromObject = function(_object) {
     var object = fabric.util.object.clone(_object);
-    fabric.util.loadImage(object.src, function(img, isError) {
-      if (isError) {
-        callback && callback(null, true);
-        return;
-      }
-      fabric.Image.prototype._initFilters.call(object, object.filters, function(filters) {
-        object.filters = filters || [];
-        fabric.Image.prototype._initFilters.call(object, [object.resizeFilter], function(resizeFilters) {
-          object.resizeFilter = resizeFilters[0];
-          fabric.util.enlivenObjectEnlivables(object, object, function () {
-            var image = new fabric.Image(img, object);
-            callback(image, false);
+    return fabric.util.loadImage(object.src, { crossOrigin: _object.crossOrigin })
+      .then(function(img) {
+        // this is obviously broken for now. need to refactor a bit
+        fabric.Image.prototype._initFilters.call(object, object.filters, function(filters) {
+          object.filters = filters || [];
+          fabric.Image.prototype._initFilters.call(object, [object.resizeFilter], function(resizeFilters) {
+            object.resizeFilter = resizeFilters[0];
+            fabric.util.enlivenObjectEnlivables(object, object, function () {
+              var image = new fabric.Image(img, object);
+              callback(image, false);
+            });
           });
         });
-      });
-    }, null, object.crossOrigin);
+      })
+    });
   };
 
   /**
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -1,14 +1,4 @@
 (function() {
-
-  function parseDecoration(object) {
-    if (object.textDecoration) {
-      object.textDecoration.indexOf('underline') > -1 && (object.underline = true);
-      object.textDecoration.indexOf('line-through') > -1 && (object.linethrough = true);
-      object.textDecoration.indexOf('overline') > -1 && (object.overline = true);
-      delete object.textDecoration;
-    }
-  }
-
   /**
    * IText class (introduced in <b>v1.4</b>) Events are also fired with "text:"
    * prefix when observing canvas.
@@ -507,17 +497,9 @@
    * @static
    * @memberOf fabric.IText
    * @param {Object} object Object to create an instance from
-   * @param {function} [callback] invoked with new instance as argument
+   * @returns {Promise<fabric.IText>}
    */
-  fabric.IText.fromObject = function(object, callback) {
-    parseDecoration(object);
-    if (object.styles) {
-      for (var i in object.styles) {
-        for (var j in object.styles[i]) {
-          parseDecoration(object.styles[i][j]);
-        }
-      }
-    }
-    fabric.Object._fromObject('IText', object, callback, 'text');
+  fabric.IText.fromObject = function(object) {
+    return fabric.Object._fromObject('IText', object, 'text');
   };
 })();
--- a/src/shapes/line.class.js
+++ b/src/shapes/line.class.js
@@ -285,15 +285,15 @@
    * @memberOf fabric.Line
    * @param {Object} object Object to create an instance from
    * @param {function} [callback] invoked with new instance as first argument
+   * @returns {Promise<fabric.Line>}
    */
-  fabric.Line.fromObject = function(object, callback) {
-    function _callback(instance) {
-      delete instance.points;
-      callback && callback(instance);
-    };
+  fabric.Line.fromObject = function(object) {
     var options = clone(object, true);
     options.points = [object.x1, object.y1, object.x2, object.y2];
-    fabric.Object._fromObject('Line', options, _callback, 'points');
+    return fabric.Object._fromObject('Line', options, 'points').then(function(fabricLine) {
+      delete fabricLine.points;
+      return fabricLine;
+    });
   };
 
   /**
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1979,22 +1979,20 @@
    * @constant
    * @type string[]
    */
-  fabric.Object.ENLIVEN_PROPS = ['clipPath'];
+  fabric.Object.ENLIVEN_PROPS = ['clipPath', 'path'];
 
   fabric.Object._fromObject = function(className, object, callback, extraParam) {
-    var klass = fabric[className];
-    object = clone(object, true);
-    fabric.util.enlivenPatterns([object.fill, object.stroke], function(patterns) {
-      if (typeof patterns[0] !== 'undefined') {
-        object.fill = patterns[0];
-      }
-      if (typeof patterns[1] !== 'undefined') {
-        object.stroke = patterns[1];
-      }
-      fabric.util.enlivenObjectEnlivables(object, object, function () {
-        var instance = extraParam ? new klass(object[extraParam], object) : new klass(object);
-        callback && callback(instance);
-      });
+    var klass = fabric[className], serializedObject = clone(object, true),
+        fill = serializedObject.fill, stroke = serializedObject.stroke;
+    var promises = [
+      fill && fill.source ? fabric.Pattern.fromObject(fill) : fill,
+      stroke && stroke.source ? fabric.Pattern.fromObject(stroke) : stroke,
+      fabric.util.enlivenObjectEnlivables(serializedObject, serializedObject),
+    ];
+    return Promise.all(promises).then(function(allTheThings) {
+      serializedObject.fill = allTheThings[0];
+      serializedObject.stroke = allTheThings[1];
+      return extraParam ? new klass(object[extraParam], serializedObject) : new klass(object);
     });
   };
 
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -335,20 +335,10 @@
    * @static
    * @memberOf fabric.Path
    * @param {Object} object
-   * @param {Function} [callback] Callback to invoke when an fabric.Path instance is created
+   * @returns {Promise<fabric.Path>}
    */
-  fabric.Path.fromObject = function(object, callback) {
-    if (typeof object.sourcePath === 'string') {
-      var pathUrl = object.sourcePath;
-      fabric.loadSVGFromURL(pathUrl, function (elements) {
-        var path = elements[0];
-        path.setOptions(object);
-        callback && callback(path);
-      });
-    }
-    else {
-      fabric.Object._fromObject('Path', object, callback, 'path');
-    }
+  fabric.Path.fromObject = function(object) {
+    return fabric.Object._fromObject('Path', object, 'path');
   };
 
   /* _FROM_SVG_START_ */
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1699,22 +1699,10 @@
    * @static
    * @memberOf fabric.Text
    * @param {Object} object plain js Object to create an instance from
-   * @param {Function} [callback] Callback to invoke when an fabric.Text instance is created
+   * @returns {Promise<fabric.Text>}
    */
-  fabric.Text.fromObject = function(object, callback) {
-    var objectCopy = clone(object), path = object.path;
-    delete objectCopy.path;
-    return fabric.Object._fromObject('Text', objectCopy, function(textInstance) {
-      if (path) {
-        fabric.Object._fromObject('Path', path, function(pathInstance) {
-          textInstance.set('path', pathInstance);
-          callback(textInstance);
-        }, 'path');
-      }
-      else {
-        callback(textInstance);
-      }
-    }, 'text');
+  fabric.Text.fromObject = function(object) {
+    return fabric.Object._fromObject('Text', object, 'text');
   };
 
   fabric.Text.genericFonts = ['sans-serif', 'serif', 'cursive', 'fantasy', 'monospace'];
--- a/src/util/lang_object.js
+++ b/src/util/lang_object.js
@@ -54,7 +54,7 @@
 
   /**
    * Creates an empty object and copies all enumerable properties of another object to it
-   * This method is mostly for internal use, and not intended for duplicating shapes in canvas. 
+   * This method is mostly for internal use, and not intended for duplicating shapes in canvas.
    * @memberOf fabric.util.object
    * @param {Object} object Object to clone
    * @param {Boolean} [deep] Whether to clone nested objects
@@ -63,7 +63,7 @@
 
   //TODO: this function return an empty object if you try to clone null
   function clone(object, deep) {
-    return extend({ }, object, deep);
+    return deep ? extend({ }, object, deep) : Object.assign({}, object);
   }
 
   /** @namespace fabric.util.object */
--- a/src/util/misc.js
+++ b/src/util/misc.js
@@ -457,55 +457,32 @@
     },
 
     /**
-     * Loads image element from given url and passes it to a callback
+     * Loads image element from given url and resolve it, or catch.
      * @memberOf fabric.util
      * @param {String} url URL representing an image
-     * @param {Function} callback Callback; invoked with loaded image
-     * @param {*} [context] Context to invoke callback in
-     * @param {Object} [crossOrigin] crossOrigin value to set image element to
+     * @param {Object} [options] image loading options
+     * @param {string} [options.crossOrigin] cors value for the image loading, default to anonymous
+     * @param {Promise<fabric.Image>} img the loaded image.
      */
-    loadImage: function(url, callback, context, crossOrigin) {
-      if (!url) {
-        callback && callback.call(context, url);
-        return;
-      }
-
-      var img = fabric.util.createImage();
-
-      /** @ignore */
-      var onLoadCallback = function () {
-        callback && callback.call(context, img, false);
-        img = img.onload = img.onerror = null;
-      };
-
-      img.onload = onLoadCallback;
-      /** @ignore */
-      img.onerror = function() {
-        fabric.log('Error loading ' + img.src);
-        callback && callback.call(context, null, true);
-        img = img.onload = img.onerror = null;
-      };
-
-      // data-urls appear to be buggy with crossOrigin
-      // https://github.com/kangax/fabric.js/commit/d0abb90f1cd5c5ef9d2a94d3fb21a22330da3e0a#commitcomment-4513767
-      // see https://code.google.com/p/chromium/issues/detail?id=315152
-      //     https://bugzilla.mozilla.org/show_bug.cgi?id=935069
-      // crossOrigin null is the same as not set.
-      if (url.indexOf('data') !== 0 &&
-        crossOrigin !== undefined &&
-        crossOrigin !== null) {
-        img.crossOrigin = crossOrigin;
-      }
-
-      // IE10 / IE11-Fix: SVG contents from data: URI
-      // will only be available if the IMG is present
-      // in the DOM (and visible)
-      if (url.substring(0,14) === 'data:image/svg') {
-        img.onload = null;
-        fabric.util.loadImageInDom(img, onLoadCallback);
-      }
-
-      img.src = url;
+    loadImage: function(url, options) {
+      return new Promise(function(resolve, reject) {
+        var img = fabric.util.createImage();
+        var done = function() {
+          img.onload = img.onerror = null;
+          resolve(img);
+        };
+        if (!url) {
+          done();
+        }
+        else {
+          img.onload = done;
+          img.onerror = function () {
+            reject(new Error('Error loading ' + img.src));
+          };
+          img.crossOrigin = options.crossOrigin;
+          img.src = url;
+        }
+      });
     },
 
     /**
@@ -538,46 +515,19 @@
      * Creates corresponding fabric instances from their object representations
      * @static
      * @memberOf fabric.util
-     * @param {Array} objects Objects to enliven
-     * @param {Function} callback Callback to invoke when all objects are created
+     * @param {Object[]} objects Objects to enliven
      * @param {String} namespace Namespace to get klass "Class" object from
      * @param {Function} reviver Method for further parsing of object elements,
      * called after each fabric object created.
      */
-    enlivenObjects: function(objects, callback, namespace, reviver) {
-      objects = objects || [];
-
-      var enlivenedObjects = [],
-          numLoadedObjects = 0,
-          numTotalObjects = objects.length;
-
-      function onLoaded() {
-        if (++numLoadedObjects === numTotalObjects) {
-          callback && callback(enlivenedObjects.filter(function(obj) {
-            // filter out undefined objects (objects that gave error)
-            return obj;
-          }));
-        }
-      }
-
-      if (!numTotalObjects) {
-        callback && callback(enlivenedObjects);
-        return;
-      }
-
-      objects.forEach(function (o, index) {
-        // if sparse array
-        if (!o || !o.type) {
-          onLoaded();
-          return;
-        }
-        var klass = fabric.util.getKlass(o.type, namespace);
-        klass.fromObject(o, function (obj, error) {
-          error || (enlivenedObjects[index] = obj);
-          reviver && reviver(o, obj, error);
-          onLoaded();
+    enlivenObjects: function(objects, namespace, reviver) {
+      return Promise.all(objects.map(function(obj) {
+        var klass = fabric.util.getKlass(obj.type, namespace);
+        return klass.fromObject(obj).then(function(fabricInstance) {
+          reviver && reviver(obj, fabricInstance);
+          return fabricInstance;
         });
-      });
+      }));
     },
 
     /**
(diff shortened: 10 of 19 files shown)


src/shapes/group.class.js
```diff
@@ -559,27 +559,14 @@
...
+   * @returns {Promise<fabric.Group>}
    */
-  fabric.Group.fromObject = function(object, callback) {
+  fabric.Group.fromObject = function(object) {
     var objects = object.objects,
         options = fabric.util.object.clone(object, true);
     delete options.objects;
-    if (typeof objects === 'string') {
-      // it has to be an url or something went wrong.
-      fabric.loadSVGFromURL(objects, function (elements) {
```
asturur, line 570: this feature has to go away.

src/shapes/image.class.js
```diff
@@ -700,26 +700,25 @@
...
+      .then(function(img) {
+        // this is obviously broken for now. need to refactor a bit
+        fabric.Image.prototype._initFilters.call(object, object.filters, function(filters) {
+          object.filters = filters || [];
+          fabric.Image.prototype._initFilters.call(object, [object.resizeFilter], function(resizeFilters) {
+            object.resizeFilter = resizeFilters[0];
+            fabric.util.enlivenObjectEnlivables(object, object, function () {
+              var image = new fabric.Image(img, object);
+              callback(image, false);
+            });
```
asturur, line 717: filter loading needs to be refactored as an enlivabled

src/shapes/itext.class.js
```diff
@@ -1,14 +1,4 @@
 (function() {
-
-  function parseDecoration(object) {
-    if (object.textDecoration) {
-      object.textDecoration.indexOf('underline') > -1 && (object.underline = true);
-      object.textDecoration.indexOf('line-through') > -1 && (object.linethrough = true);
-      object.textDecoration.indexOf('overline') > -1 && (object.overline = true);
-      delete object.textDecoration;
-    }
-  }
```
asturur, line 10: this was meant to aid for changes between fabric 1.x and fabric 2.0
I think it can go away now.

src/shapes/line.class.js
```diff
@@ -285,15 +285,15 @@
...
+   * @returns {Promise<fabric.Line>}
    */
-  fabric.Line.fromObject = function(object, callback) {
-    function _callback(instance) {
-      delete instance.points;
-      callback && callback(instance);
-    };
+  fabric.Line.fromObject = function(object) {
     var options = clone(object, true);
     options.points = [object.x1, object.y1, object.x2, object.y2];
```
asturur, line 292: not sure why we export with x1,x2... and we import with a points array, but we need to standardize this.

src/shapes/path.class.js
```diff
@@ -335,20 +335,10 @@
    * @static
    * @memberOf fabric.Path
    * @param {Object} object
-   * @param {Function} [callback] Callback to invoke when an fabric.Path instance is created
+   * @returns {Promise<fabric.Path>}
    */
-  fabric.Path.fromObject = function(object, callback) {
-    if (typeof object.sourcePath === 'string') {
-      var pathUrl = object.sourcePath;
-      fabric.loadSVGFromURL(pathUrl, function (elements) {
```
asturur, line 343: also this feature, away.
This is a reminescensce of pathGroup, in which if you exported to a data-less object, the path would be converted in a path-string or an svg.

src/shapes/text.class.js
```diff
@@ -1699,22 +1699,10 @@
...
-      if (path) {
-        fabric.Object._fromObject('Path', path, function(pathInstance) {
-          textInstance.set('path', pathInstance);
-          callback(textInstance);
-        }, 'path');
-      }
-      else {
-        callback(textInstance);
-      }
-    }, 'text');
```
asturur, line 1717: in theory adding 'path' to enlivable should do the trick no?

src/util/lang_object.js
```diff
@@ -63,7 +63,7 @@
 
   //TODO: this function return an empty object if you try to clone null
   function clone(object, deep) {
-    return extend({ }, object, deep);
+    return deep ? extend({ }, object, deep) : Object.assign({}, object);
```
asturur, line 66: since we removed support for ie11 and we plan to add a transpiler, i think we can leverage Object.assign for the shallow clone.

src/util/misc.js
```diff
@@ -457,55 +457,32 @@
...
+        }
+        else {
+          img.onload = done;
+          img.onerror = function () {
+            reject(new Error('Error loading ' + img.src));
+          };
+          img.crossOrigin = options.crossOrigin;
+          img.src = url;
+        }
+      });
```
asturur, line 485: crossOrigin is no more optional, is necessary everywhere nowadays, we can stop assuming is optional.

src/util/misc.js
```diff
@@ -538,105 +515,38 @@
...
-          return;
-        }
-        var klass = fabric.util.getKlass(o.type, namespace);
-        klass.fromObject(o, function (obj, error) {
-          error || (enlivenedObjects[index] = obj);
-          reviver && reviver(o, obj, error);
-          onLoaded();
+    enlivenObjects: function(objects, namespace, reviver) {
+      return Promise.all(objects.map(function(obj) {
+        var klass = fabric.util.getKlass(obj.type, namespace);
```
asturur, line 525: if we find a nice solution for filters, namespace can go away

asturur:
@ShaMan123 @melchiar When i know it actually runs i will ask you to give a look.
For now is wasted time

.eslintrc.json
```diff
@@ -3,7 +3,7 @@
     "browser": true
   },
   "globals": {
-    "ActiveXObject": true,
```
asturur, line 6: If you know what is this, you are probably 35+

src/shapes/text.class.js
ShaMan123: I think so

ShaMan123:
Holy shit this is major!
So much garbage removed! Exciting!

src/mixins/canvas_serialization.mixin.js
```diff
@@ -31,38 +30,36 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
...
-            _this.__setupCanvas.call(_this, serialized, enlivenedObjects, renderOnAddRemove, callback);
+    return fabric.util.enlivenObjects(serialized.objects || [], '', reviver)
+      .then(function(enlived) {
+        _this.clear();
+        return fabric.util.enlivenObjectEnlivables({
+          backgroundImage: serialized.backgroundImage,
+          backgroundColor: serialized.background,
+          overlayImage: serialized.overlayImage,
+          overlayColor: serialized.overlay,
+          clipPath: serialized.clipPath,
```
asturur, line 46: as soon as we have es6 in place this will look leaner.

src/util/misc.js
```diff
@@ -538,104 +515,56 @@
...
+        }
+        enliven[prop] = value;
+      });
+      var promises = Object.values(enliven);
+      var keys = Object.keys(enliven);
+      return Promise.all(promises).then(function(enlived) {
+        return keys.reduce(function(acc, element, index) {
+          acc[keys[index]] = enlived[index];
+          return acc;
+        }, {});
```
asturur, line 567: @ShaMan123  i changed your work on enlivenObjectEnlivables.
I wanted it to take care of patterns/gradients/
Can you tell me what you think?

asturur:
There is a little bit of changes here.
Apart promises, things should work in this way:
When you create an object with `var myThing = new fabric.MyThing(options)`,  i wish the standard was that you pass in the options values that are ready to use, and not the serialized part of it.
Serialized things are async and i don't like async constructors or with callback. The general consensus is that constructor shouldn't do much work.

So enliveObjectEnlivables can eventually wrap the options before you create an object, or you start from the fromObject method directly.

Indeed i removed the setBackgroundImage, setOverlayImage.
if you have a source of an image:

`fabric.Image.FromUrl(src).then((fabricImage) => canvas.backgroundImage = fabricImage)`

From a developer persepctive seems reasonable to write something like that.

ShaMan123 (changes requested):
This PR looks awesome.
Makes the codebase much more readable and elegant and easy to use and override.
I have added comments.
The main things are:
1. keeping customization in mind (e.g. introducing a param/mechanism to allow adding custom enlivened props with ease) 
2. drop namespaces altogether
3. be consist in dropping callbacks - basically drop `_fromObject` as well

src/shapes/object.class.js
```diff
@@ -1981,23 +1972,18 @@
    */
   fabric.Object.ENLIVEN_PROPS = ['clipPath'];
```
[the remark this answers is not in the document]
ShaMan123: drop this in favor of something better

src/shapes/object.class.js
```diff
@@ -1981,23 +1972,18 @@
...
-      }
-      if (typeof patterns[1] !== 'undefined') {
-        object.stroke = patterns[1];
-      }
-      fabric.util.enlivenObjectEnlivables(object, object, function () {
-        var instance = extraParam ? new klass(object[extraParam], object) : new klass(object);
-        callback && callback(instance);
-      });
+  fabric.Object._fromObject = function(className, object, extraParam) {
+    var klass = fabric[className], serializedObject = clone(object, true);
```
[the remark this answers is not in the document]
ShaMan123: +1 for dropping fabric.Object._fromObject and letting ` fabric.util.enlivenObjectEnlivables` clone the arg if needed

src/shapes/group.class.js
ShaMan123: I think this means you can remove `groupSVGElements`

src/util/misc.js
ShaMan123: We can add `objects` as well

asturur: ok we find a way to do that

src/mixins/canvas_serialization.mixin.js
```diff
@@ -31,38 +30,36 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
...
-      _this.clear();
-      _this._setBgOverlay(serialized, function () {
-        if (clipPath) {
-          _this._enlivenObjects([clipPath], function (enlivenedCanvasClip) {
-            _this.clipPath = enlivenedCanvasClip[0];
-            _this.__setupCanvas.call(_this, serialized, enlivenedObjects, renderOnAddRemove, callback);
+    return fabric.util.enlivenObjects(serialized.objects || [], '', reviver)
+      .then(function(enlived) {
+        _this.clear();
+        return fabric.util.enlivenObjectEnlivables({
```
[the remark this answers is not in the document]
asturur: We could eventually try to enlive every single prop, and not having a list.

asturur:
> be consist in dropping callbacks - basically drop _fromObject as well
What do you mean with dropping _fromObject?

src/mixins/canvas_serialization.mixin.js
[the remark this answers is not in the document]
ShaMan123: Well if it's possible and not damaging... sound not too bad

ShaMan123:
> > be consist in dropping callbacks - basically drop _fromObject as well
> > What do you mean with dropping _fromObject?

I mean I think this it is redundant - it doesn't do much but wrap the call to `fabric.util.enlivenObjectEnlivables`

src/shapes/line.class.js
melchiar: For that matter you could probably make a good argument for removing circle or even triangle as well since we have ellipse and polygon.

asturur:
> > > be consist in dropping callbacks - basically drop _fromObject as well
> > > What do you mean with dropping _fromObject?
> 
> I think it is redundant - it doesn't do much but wrap the call to `fabric.util.enlivenObjectEnlivables` Perhaps rename it as something like `prepareObject`

Well if it wraps if for 6-7 classes, is good enough.
When switching to es6 or ts, most of those things shrink in size considerably.

asturur:
So the PR would have passing tests, but something is making it stuck in the browser

.eslintrc.json
andrewzolotukhin: confirming. almost 36 and I'm aware of it :)

asturur:
@ShaMan123 help me:
```
not ok 30 Firefox 96.0 - [108 ms] - fabric.util.animate: animate with list of values
    ---
        actual: >
            null
        stack: >
            test@http://localhost:8080/node_modules/qunit/qunit/qunit.js:2900:12
            @http://localhost:8080/6514[49](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:49)99684/test/unit/animation.js:335:9
            @http://localhost:8080/6[51](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:51)44999684/test/unit/animation.js:845:3
            
        message: >
            Expected [52](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:52) assertions, but [59](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:59) were run
        negative: >
            false
        browser log: |
    ...
```
i swear i looked for this test in all my hard disk and i can't locate it

asturur:
ok somehow the test was in latest master, but still running here.
Misteries of the CI.

asturur:
ok 55 files changed is no joke to review.
I would suggest this process:
- review the tests first, that will give you an idea of what we broke.
- review the code changes and check if there is something that you necessarly want changed now, and that for some reason can't be an follow up.

asturur:
After this PR i would shift my focus onboarding on the group changes

ShaMan123:
> @ShaMan123 help me:
> 
> ```
> not ok 30 Firefox 96.0 - [108 ms] - fabric.util.animate: animate with list of values
>     ---
>         actual: >
>             null
>         stack: >
>             test@http://localhost:8080/node_modules/qunit/qunit/qunit.js:2900:12
>             @http://localhost:8080/6514[49](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:49)99684/test/unit/animation.js:335:9
>             @http://localhost:8080/6[51](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:51)44999684/test/unit/animation.js:845:3
>             
>         message: >
>             Expected [52](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:52) assertions, but [59](https://github.com/fabricjs/fabric.js/runs/5183084137?check_suite_focus=true#step:6:59) were run
>         negative: >
>             false
>         browser log: |
>     ...
> ```
> 
> i swear i looked for this test in all my hard disk and i can't locate it
> 
> When i go to `animation.js:335:9` i have the other test, the one that says animate with abort

@rockerBOO take a look 
Running on windows - might be the animation frame rate is different? seems to lack 2 calls to `onChange` which are 2*7 assertions
@asturur is this possible?
```bash
not ok 30 fabric.util.animate > animate with list of values
  ---
  message: Expected 52 assertions, but 38 were run
  severity: failed
  actual  : null
  expected: undefined
  stack: |
        at Object.test (C:\Users\DELL\Desktop\DEV\fabric.js\node_modules\qunit\qunit\qunit.js:2900:5)
        at C:\Users\DELL\Desktop\DEV\fabric.js\test\unit\animation.js:335:9
        at Object.<anonymous> (C:\Users\DELL\Desktop\DEV\fabric.js\test\unit\animation.js:845:3)
        at Module._compile (internal/modules/cjs/loader.js:1072:14)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
        at Module.load (internal/modules/cjs/loader.js:937:32)
        at Function.Module._load (internal/modules/cjs/loader.js:778:12)
        at Module.require (internal/modules/cjs/loader.js:961:19)
  ...
```

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/7633#issuecomment-1035963544_

ShaMan123:
we should simply remove the `assert.expect`

asturur:
i removed that because i did not want to deal with it.
I'm sure on browser it can be different because it can skip a frame for whatever reason. also browser run in the command line do not have a real screen connected, the refresh rate is an artifact. ( on my 144hz screen the refresh rate is 144 fps )

src/shapes/image.class.js
```diff
@@ -756,8 +739,10 @@
    */
   fabric.Image.fromElement = function(element, callback, options) {
     var parsedAttributes = fabric.parseAttributes(element, fabric.Image.ATTRIBUTE_NAMES);
-    fabric.Image.fromURL(parsedAttributes['xlink:href'], callback,
-      extend((options ? fabric.util.object.clone(options) : { }), parsedAttributes));
+    fabric.Image.fromURL(parsedAttributes['xlink:href'], Object.assign({ }, options || { }, parsedAttributes))
+      .then(function(fabricImage) {
+        callback(fabricImage);
```
ShaMan123, line 744: Is the remaining callback intentional?
Why `fromElement` is left as legacy?

src/shapes/line.class.js
ShaMan123: what's the decision regarding the export?

src/shapes/object.class.js
```diff
@@ -815,10 +815,6 @@
      */
     setOptions: function(options) {
       this._setOptions(options);
```
ShaMan123, line 817: `setOptions` doesn't do anything anymore,
maybe rename `_setOptions`?

src/shapes/object.class.js
```diff
@@ -1980,25 +1971,19 @@
...
-      }
-      if (typeof patterns[1] !== 'undefined') {
-        object.stroke = patterns[1];
-      }
-      fabric.util.enlivenObjectEnlivables(object, object, function () {
-        var instance = extraParam ? new klass(object[extraParam], object) : new klass(object);
-        callback && callback(instance);
-      });
+  fabric.Object._fromObject = function(className, object, extraParam) {
+    var klass = fabric[className], serializedObject = clone(object, true);
```
ShaMan123, line 1976: I think we should get rid of `className` and pass a live class (`klass`)

src/util/misc.js
```diff
@@ -457,55 +457,32 @@
...
-      img.src = url;
+    loadImage: function(url, options) {
+      return new Promise(function(resolve, reject) {
+        var img = fabric.util.createImage();
+        var done = function() {
+          img.onload = img.onerror = null;
+          resolve(img);
+        };
+        if (!url) {
+          done();
```
ShaMan123, line 475: isn't it better rejecting in this case?
I understand this is how it was but it's wrong.

src/util/misc.js
```diff
@@ -457,55 +457,32 @@
...
+        }
+        else {
+          img.onload = done;
+          img.onerror = function () {
+            reject(new Error('Error loading ' + img.src));
```
ShaMan123, line 480: we should pass the original error event
```diff
-            img.onerror = function () {
-            reject(new Error('Error loading ' + img.src));
+            img.onerror = reject;
```

src/util/misc.js
ShaMan123: same here regarding passing `klass` instead of namespace

src/util/misc.js
```diff
@@ -538,104 +515,52 @@
...
+     * @param {Object} object with properties to enlive ( fill, stroke, clipPath, path )
+     * @returns {Promise<object>} the input object with enlived values
      */
-    enlivenPatterns: function(patterns, callback) {
-      patterns = patterns || [];
 
-      function onLoaded() {
-        if (++numLoadedPatterns === numPatterns) {
-          callback && callback(enlivenedPatterns);
+    enlivenObjectEnlivables: function (serializedObject) {
```
ShaMan123, line 539: this seems a bit loose and error prone.
shouldn't we pass a map with keys and types?
e.g. 
```js
{
   clipPath: 'object',
   fill: 'pattern',
   path: 'object'
}
```

ShaMan123 (changes requested):
I reviewed and here are my findings:
1. namespaces - can we let them go and pass real classes?
2. canvas_serialization - did you miss these?
https://github.com/fabricjs/fabric.js/blob/fee85470b80127b43dc9e1174f4a17d0bd7951be/src/mixins/canvas_serialization.mixin.js#L99
https://github.com/fabricjs/fabric.js/blob/fee85470b80127b43dc9e1174f4a17d0bd7951be/src/mixins/canvas_serialization.mixin.js#L111
I see `test/visual/z_svg_export.js` `test/visual/toDataURL.js` use callbacks instead of promises as well
3. left many comments to change stuff
4. `misc/groupSVGElements` is dead and should be removed?
5. actually I see there are a few dead methods in misc
6. I will try to push some minor doc commits
7. what about svg parsing? I see it remained in callback form `Object.fromElement` and `loadSvgFrom..`
8. you missed `fabric.Image.filters.BlendImage.fromObject`

src/shapes/object.class.js
asturur: yes we can, but that is not fundamental for moving this forward. this is ok as an action item

src/mixins/canvas_serialization.mixin.js
```diff
@@ -205,24 +132,20 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
...
-        clone.renderAll();
-        callback && callback(clone);
-      });
-      clone.backgroundImageOpacity = this.backgroundImageOpacity;
-      clone.backgroundImageStretch = this.backgroundImageStretch;
-    }
-    else {
-      callback && callback(clone);
+      // TODO verify this
+      clone.setBackgroundImage(this.backgroundImage.src);
```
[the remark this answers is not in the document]
asturur: this should error, so we miss a test.

src/shapes/group.class.js
```diff
@@ -559,26 +559,15 @@
...
-      return;
-    }
-    fabric.util.enlivenObjects(objects, function (enlivenedObjects) {
-      var options = fabric.util.object.clone(object, true);
-      delete options.objects;
-      fabric.util.enlivenObjectEnlivables(object, options, function () {
-        callback && callback(new fabric.Group(enlivenedObjects, options, true));
+    return fabric.util.enlivenObjects(objects).then(function (enlivenedObjects) {
+      return fabric.util.enlivenObjectEnlivables(options).then(function(enlivedProps) {
+        return new fabric.Group(enlivenedObjects, Object.assign(options, enlivedProps), true);
```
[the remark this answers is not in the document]
asturur: no objects no for now.
It would required doing things recursively, i m fine with enlving the objects manually for groups and canvas right now.
The extra terse code is not giving us much

src/util/misc.js
asturur: As soon as it errors, we can find a solution.
You are not required to pass all in, but you can.
With deconstruction this becomes super easy to do.

src/util/misc.js
asturur: https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent

So the ErrorEvent  seems to be different from `new Error()` so we can't get 1:1 mapping reject and onerrorr

src/util/misc.js
asturur: Yes but i don't want to change it now. We can change it , no issues from me

src/shapes/object.class.js
ShaMan123: alright

src/util/misc.js
ShaMan123: not in this PR you mean?

src/util/misc.js
ShaMan123: strange, it seems it has not important data
```txt
Event {isTrusted: true, type: 'error', target: img, currentTarget: img, eventPhase: 2, …}
isTrusted: true
bubbles: false
cancelBubble: false
cancelable: false
composed: false
currentTarget: null
defaultPrevented: false
eventPhase: 0
path: [img]
returnValue: true
srcElement: null
target: null
timeStamp: 382917.6999999881
type: "error"
```
so ok we leave it as is

ShaMan123: The red log is what I meant important but it isn't part of the error event
image

asturur:
image

@kangax 

I'll write the test for you

src/shapes/line.class.js
asturur: I'm not sure where did i write it, but in theory, apart circle, everything else can be done at the same speed with fabric.Path, you can still have constructors for Triangle, but will create a Path.
Then you need to write setters/getters to change the path when you change width/height. Same for line.

This PR won't take care of those things, but we can have a session where we look at those smaller classes and we decided if we want to normalize them.

src/mixins/canvas_serialization.mixin.js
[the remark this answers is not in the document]
asturur: There isn't a real decision. For now a generic enlive all works, as long as you remove the resizeFilter ( thing i did in the image class ). I would keep it as it is, and we circle back when issues surface.
Compared to the previous canvas enliving it seems a step up anyway

asturur:
Ok @ShaMan123 @melchiar i think is mergeable now.
- added tests for Vibrance filter and BlendColor filters that were missing
- Fixing @ShaMan123 comments on: using the class instead of classname, blendImage fix, some remaining callbacks
- removed some more dead methods

SVG parsing will be done in a followup.
I'm not even sure i want to change it before evaluating if i should work on a different parser entirely.

ShaMan123:
I have a commit for eraser
I want to push it


בתאריך יום ג׳, 15 בפבר׳ 2022, 13:18, מאת Andrea Bogazzi ‏<
***@***.***>:

asturur:
on a PR on this branch and let's merge in on this

asturur:
oh well i see it depends also on the group work.
Well as soon as this approved i ll merge it.
Then i ll get to group rewrite understanding/commenting/collaborating.

ShaMan123:
I want to makr incremental change only for promise and commit that here
And then handle the group chsnge

בתאריך יום ג׳, 15 בפבר׳ 2022, 14:38, מאת Andrea Bogazzi ‏<
***@***.***>:

asturur:
That is fine, when you are ok with this pr, you press the approve button.
i'll wait.

src/util/misc.js
ShaMan123: already it is much better, an error has been raised in a demo I work with that never did and should have

src/shapes/group.class.js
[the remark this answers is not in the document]
ShaMan123: group is phasing out so I don't care

ShaMan123 (approved):
Eraser is committed
I've pushed some minor JSDOC changes as well
1. I want to change `enlivenObjects` not to use `getKlass` at all (we'll do that once we migrate js)
2. I do think we should consider wrapping the external methods of the svg parser with promises (and leave internals as is until you decide it's fate) in order to unify fabric. But it can wait.


Excellent job! This was a huge tedious effort.
Group is up!

melchiar (dismissed):
This is great. So much garbage removed!

src/mixins/canvas_serialization.mixin.js
```diff
@@ -3,24 +3,23 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
...
-   * canvas.loadFromJSON(json, canvas.renderAll.bind(canvas));
+   * canvas.loadFromJSON(json).then((canvas) => canvas.requestRenderAll());
    * @example <caption>loadFromJSON with reviver</caption>
-   * canvas.loadFromJSON(json, canvas.renderAll.bind(canvas), function(o, object) {
+   * canvas.loadFromJSON(json, function(o, object) {
    *   // `o` = json object
    *   // `object` = fabric.Object instance
    *   // ... do some stuff ...
+   * }).then((canvas) => {
+   *   ... your canvas callback
```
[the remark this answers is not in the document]
asturur: no we can change that too

asturur:
off i pushed away reviews because i addressed the text comment.

rhanmiano:
Hello, is this feature available already, and in which version? Thanks

ShaMan123:
v5+

rhanmiano:
Hi @ShaMan123 , thanks for the response. I just have a problem. After upgrading from `4.6.0` to `5.2.1`. I get an error similar to [this](https://github.com/fabricjs/fabric.js/issues/7830) using `loadFromJSON(...).then`. After inspecting the `dist/fabric.js` the structure of `loadFromJSON` is still using the `fn(json, callback, reviver)`. Am I missing something? Thank you.

Please see image for reference.
image

image

ShaMan123:
Did you try building fabric as suggested in the issue you referenced?

rhanmiano:
I tried rebuilding the fabric also, and the generated `fabric.min.js` still contains the structure of `fn(json, callback, reviver)` just the minified version. And still having the same issue.

ShaMan123:
I see the version has somehow got mixed with outdated code.
I suggest you install fabric's latest version directly from github.
```bash
npm i --save github:fabricjs/fabric.js
```

rhanmiano:
Thanks @ShaMan123 , that solves the issue on my side. Will wait for the next stable patch.

PR #7659, opened 2022-02-07
Update bug_report.md

ShaMan123:


asturur:
We can try,
but most of the time i see people doing CTRL+A, CANC, Write what they want to write.

ShaMan123:
so we add an angry bot
joking

PR #7672, opened 2022-02-12
docs(blendimage_filter.class.js) corrected mode options

melchiar:
Removed unsupported mode options for blendimage filter and added variable type

### DIFF
--- a/src/filters/colormatrix_filter.class.js
+++ b/src/filters/colormatrix_filter.class.js
@@ -66,8 +66,10 @@
     mainParameter: 'matrix',
 
     /**
-     * Lock the colormatrix on the color part, skipping alpha, manly for non webgl scenario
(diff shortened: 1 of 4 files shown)


melchiar:
@asturur we seem to be a bit inconsistent with our doc structure. Do you have a preferred structure for tagging property types?

@type String
or
@param {String}

asturur:
would be nicer to implement them XD

src/filters/colormatrix_filter.class.js
```diff
@@ -66,8 +66,10 @@
     mainParameter: 'matrix',
 
     /**
-     * Lock the colormatrix on the color part, skipping alpha, manly for non webgl scenario
```
asturur, line 69: Skipping alpha manly.

image

asturur (approved):

PR #7674, opened 2022-02-12
fix(fabric.Text): support text alignments in RTL text

amirhossein92:
The RTL text support was initially added in version 4.5, but it had some problem specially with cursor placement and toSVG mode. The cursor position was correct only if 'textAlign' was set to 'right'.

In this PR I considered the cursor position in 'left', 'center' and 'justify' text alignments and also considered the direction and text alignment in the toSVG method.

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1044,7 +1044,7 @@
           path = this.path,
           shortCut = !isJustify && this.charSpacing === 0 && this.isEmptyStyles(lineIndex) && !path,
           isLtr = this.direction === 'ltr', sign = this.direction === 'ltr' ? 1 : -1,
-          drawingLeft, currentDirection = ctx.canvas.getAttribute('dir');
+          drawingLeft, currentDirection = ctx.direction;
       ctx.save();
       if (currentDirection !== this.direction) {
         ctx.canvas.setAttribute('dir', isLtr ? 'ltr' : 'rtl');
@@ -1306,7 +1306,15 @@
         leftOffset = lineDiff;
       }
       if (direction === 'rtl') {
-        leftOffset -= lineDiff;
+        if (textAlign === 'right' || textAlign === 'justify'){
+          leftOffset = 0;
+        }
+        else if (textAlign === 'left') {
+          leftOffset = -lineDiff;
+        }
+        else if (textAlign === 'center') {
+          leftOffset = -lineDiff / 2;
+        }
       }
       return leftOffset;
     },
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1306,13 +1306,13 @@
         leftOffset = lineDiff;
       }
       if (direction === 'rtl') {
-        if (textAlign === 'right' || textAlign === 'justify'){
+        if (textAlign === 'right' || textAlign === 'justify' || textAlign === 'justify-right') {
           leftOffset = 0;
         }
-        else if (textAlign === 'left') {
+        else if (textAlign === 'left' || textAlign === 'justify-left') {
           leftOffset = -lineDiff;
         }
-        else if (textAlign === 'center') {
+        else if (textAlign === 'center' || textAlign === 'justify-center') {
           leftOffset = -lineDiff / 2;
         }
       }
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1044,6 +1044,8 @@
           path = this.path,
           shortCut = !isJustify && this.charSpacing === 0 && this.isEmptyStyles(lineIndex) && !path,
           isLtr = this.direction === 'ltr', sign = this.direction === 'ltr' ? 1 : -1,
+          // this was changed in the PR #7674
+          // currentDirection = ctx.canvas.getAttribute('dir');
           drawingLeft, currentDirection = ctx.direction;
       ctx.save();
       if (currentDirection !== this.direction) {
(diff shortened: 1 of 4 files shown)


src/shapes/text.class.js
```diff
@@ -1044,7 +1044,7 @@
           path = this.path,
           shortCut = !isJustify && this.charSpacing === 0 && this.isEmptyStyles(lineIndex) && !path,
           isLtr = this.direction === 'ltr', sign = this.direction === 'ltr' ? 1 : -1,
-          drawingLeft, currentDirection = ctx.canvas.getAttribute('dir');
+          drawingLeft, currentDirection = ctx.direction;
```
asturur, line 1047: Can you tell me about this change here?

asturur:
@amirhossein92  unit tests are failing, something changed in a breaking way.

On top of that, can you provide some visual test or unit test to cover the new code?

src/shapes/text.class.js
amirhossein92: I figured out that everything works correctly on initial loading of the text, but when you change some property like the `textAlign`, the `direction` is not considered anymore (because of the `currentDirection !== this.direction` condition) and the rendering is not correct anymore.

initial loading with `direction=rtl` and `textAlign=rtl`:
Screen Shot 2022-02-14 at 10 46 29 PM

after changing the text align to center and then changed it back to left:
Screen Shot 2022-02-14 at 10 47 19 PM

amirhossein92:
I created a sample for text and uploaded it as a [gist](https://gist.github.com/amirhossein92/ea7f54b745df600cfae13519a6dba6f0)

here is a video I captured, after my modifications.

amirhossein92:
@asturur I revised the code and considered different type of `justify`. Now, I expect the unit test to pass.

src/shapes/text.class.js
asturur: i still don't get it.
currentDirection is supposed to mean `direction on which the canvas is set to` and both the attribute and the ctx.direction are kept in sync. How one is different from the other?

amirhossein92: I logged their values and It seems that the `ctx.direction` is not in sync with the `ctx.canvas.dir`

Screen Shot 2022-02-24 at 10 58 43 PM

asturur: oh right, because ctx.direction is involved probably in save and restore while the canvas attribute isn't.
That is right and is a bug on our end

asturur:
As a notice, bidirectional test still is not done. So the cursor in that case won't work and multi colored text won't work either in bidirectional cases

asturur:
Can i ask you to leave a comment on top of canvas.getAttribute, in the code, saying that this was before ctx.direction and was changed in PR 7674?

asturur:
also @amirhossein92 you are changing a beahviour but not adding any test.
You should definitely write a test for the use case you are fixing, is important.
If you don't know how to do it i can definitely help

amirhossein92:
> As a notice, bidirectional test still is not done. So the cursor in that case won't work and multi colored text won't work either in bidirectional cases

Yes exactly, but I had no idea how to fix it.


> Can i ask you to leave a comment on top of canvas.getAttribute, in the code, saying that this was before ctx.direction and was changed in PR 7674?

Yes, of course. I just did it.


> also @amirhossein92 you are changing a beahviour but not adding any test.
> You should definitely write a test for the use case you are fixing, is important.
> If you don't know how to do it i can definitely help

I am not sure what kind of test should I add, it will be great if you can help me.

ShaMan123:
I can help as well.
I will try to look into bidi selection

ShaMan123:
It seems possible to use text style to explicitly mark parts as rtl.
I want to support an `auto` direction

ShaMan123:
@amirhossein92 this is real progess!
line breaking is still wrong

image

ShaMan123:
starting word on bidi #7796

ShaMan123:
Ohh no, I tried to rebase and did this disaster. @asturur please help me revert

src/shapes/text.class.js
amirhossein92: @asturur by the way, it seems that the `ctx.direction` is not supported in Firefox as stated [here](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/direction), so this is not

asturur:
@[amirhossein92](https://github.com/amirhossein92) we are sorry, but @ShaMan123 trying to update the branch messed up the history.
Would you be so kind to push force with your current local, so we get back what you did and we can proceed to merge it?

asturur:
@amirhossein92

ShaMan123:
todo:
- [ ] originX/Y start/end values

amirhossein92:
> @[amirhossein92](https://github.com/amirhossein92) we are sorry, but @ShaMan123 trying to update the branch messed up the history.
> Would you be so kind to push force with your current local, so we get back what you did and we can proceed to merge it?

@asturur Of course, I did it.

ShaMan123:
Excellent! You fixed it.
I saw it removed my commits. And now I'm too afraid I'll mess up everything again so I've created a branch with what I salveged [rtl-7674[(https://github.com/fabricjs/fabric.js/tree/rtl-7674)

ShaMan123:
And I was thinking that real rtl support should include supporting originX/Y start/end that relate to tl/tr corner of canvas/group

asturur:
> And I was thinking that real rtl support should include supporting originX start/end that relate to tl/tr corner of canvas/group according to the direction prop of the object

mmmm not sure what you mean. Please open a separate issue with that so you can explain what you mean.

src/shapes/text.class.js
asturur: yep firefox is a bit rough and is also the reason why i added canvas.attribute.dir.
But eventually we can do browser sniffing and change.

asturur (approved):

asturur:
i will add a Unit test a visual test and merge it

asturur (approved):

PR #7678, opened 2022-02-14
fix(test): Remove expect assertion

rockerBOO:
Due to different browser framerates and timing intervals, the
assertion would fail. 144fps browser, `requestAnimationFrame`, and
`setTimeout` would cause inconsistent assertions.


asturur:
Thank you @rockerBOO

PR #7669, opened 2022-02-10
feat(`V6`): Group Rewrite + Nested Selection

ShaMan123:
# Group Rewrite
This PR has been broken down into #7858  #7859 #7860 and will be left open for tracking only.

**This description is updated continuously with relevant information**
You don't need to scroll and read endless comments (unless you are a maintainer :))

## Motivation
#7473 https://github.com/fabricjs/fabric.js/pull/7316 https://github.com/fabricjs/fabric.js/issues/6776 https://github.com/fabricjs/fabric.js/discussions/7136 https://github.com/fabricjs/fabric.js/discussions/7299 https://github.com/fabricjs/fabric.js/issues/7130 https://github.com/fabricjs/fabric.js/issues/7142 #7449 and anything with a `group` tag

closes #6499, closes #3012, closes #3000, closes #7130, closes #7142, closes #7598

Nested selection is a MUST.
Group wasn't easy to use and was very limiting (e.g. doesn't support resizing logic, customize rendering - bg) and confusing (relative coordinate system and transform).

## Current State
Fixed logic across fabric for nested objects to be selectable.
Selection logic needs refactoring to support some cases of nested selection (multiselect), for now use `canvas.setActiveObject`.
Group has been completely rewritten and Layer has been introduced. As part of the rewrite group is now a layout engine, you can use it to customize layout of objects.
Group now respects  `originX` `originY` (still buggy)

### BREAKING
- all awkward methods such as `addWithUpdate` are removed

## Working Examples
- [**CodeSandbox**](https://codesandbox.io/s/jumshw) or [**JSFiddle**](http://jsfiddle.net/Lxjegow3/8/)
- [**Layer**](https://codesandbox.io/s/fabric-app-forked-cxcz3) or in [**JSFiddle**](http://jsfiddle.net/6xjszqtg/1/)

## POC
Tryouts (out of scope, and stale, here to light the way):
- [ICollection 1](https://codesandbox.io/s/yiliv)
- [ICollection 2](https://codesandbox.io/s/fabric-app-forked-cvewn)

## Alpha testing
- cdn: https://cdn.jsdelivr.net/gh/fabricjs/fabric.js@vv6!-build/dist/fabric.js
- npm: `npm i --save fabricjs/fabric.js#v6! && cd ./node_modules/fabric && npm run build`
- yarn pnp: follow the patching instructions in the POC

## To Do

- [ ] refactor selection logic (use `dblClick` to select nested objects? create mechanism for dev to customize selection)
[...]

PR #7680, opened 2022-02-15
fix(EraserBrush): migrate to v6

ShaMan123:
migrates to promises and new group
merge only after #7657 #7669 are merged
hasn't been validated, needs testing


ShaMan123:
merged some to both

PR #7696, opened 2022-02-18
chore(): Replace deprecated String.prototype.substr() with Array.prototype.slice()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/lib/event.js
+++ b/lib/event.js
@@ -237,7 +237,7 @@ var eventManager = function(target, type, listener, configure, trigger, fromOver
 		}
 		return createBatchCommands(events);
 	} else if (type.indexOf("on") === 0) { // to support things like "onclick" instead of "click"
-		type = type.substr(2);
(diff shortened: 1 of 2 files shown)


asturur (approved):

asturur:
if tests pass i m fine with this

CommanderRoot:
The failed test error looks unrelated to the changes in this PR

asturur:
inthink closing the pr was a missclick. sorry. Yes failure is unrelated

PR #7698, opened 2022-02-18
fix(`isOnScreen`): account for shadow

### LINKED ISSUE #7045: Shadow disappears when associated object goes off canvas even though it should still show based on the offset
## Version
4.4.0 (not a new regression)

## Test Case
https://jsfiddle.net/2y5v7enk/

Grab the shape in the bottom right corner and drag up and left off the screen

Nodejs and all browsers

## Steps to reproduce
1. Create an object with a shadow with a large offsetX and offsetY
2. Move object so that it is off canvas but not so far that the shadow would also be off screen

## Expected Behavior
Shadow should still show

## Actual Behavior
Shadow disappears

## Workaround
set skipOffscreen=false on the canvas. Not a great workaround for apps enjoying performance advantage of skipOffscreen

## Thanks!

ShaMan123:
closes #7045 
As always rotation makes everything so complicated.
~~I need help to get it working.... and writing a test for it.~~  **resolved**
~~The thing is the blur vector is affected by the angle, but I don't think the offset vector is.~~
Reminds me of #7478 

https://23kbgk.csb.app/

**READY** to merge


asturur:
yes rotation and shadow is hard. Is better to overshoot.
Enlarge the box in all the directions for the largest of the 2 shadow offsets and call it a day.
Don't overthink it.

ShaMan123:
I can't get it to work without annoying edge cases

ShaMan123:
This is ready to be merged

src/mixins/object_geometry.mixin.js
```diff
@@ -183,67 +202,112 @@
...
+    isShadowPartiallyOnScreen: function (calculate) {
+      if (!this.canvas || !this.shadow) {
+        return false;
+      }
+      var shadowOffsets = this.getShadowOffsets(),
+          shadowOffset = shadowOffsets.offset, blurOffset = shadowOffsets.blur;
+      //  instead of offsetting object coordinates to calculate shadows coordinates,
+      //  we translate canvas' corners to the shadow's coordinate system
+      var tl = this.canvas.vptCoords.tl.subtract(shadowOffset).subtract(blurOffset),
+          br = this.canvas.vptCoords.br.subtract(shadowOffset).add(blurOffset);
```
asturur, line 309: should this be
`this.canvas.vptCoords.br.add(shadowOffset).add(blurOffset)` ?

src/mixins/object_geometry.mixin.js
```diff
@@ -183,67 +202,112 @@
...
      * @param {Boolean} [calculate] use coordinates of current position instead of .aCoords
      * @return {Boolean} true if object is fully or partially contained within canvas
      */
-    isOnScreen: function(calculate) {
-      if (!this.canvas) {
-        return false;
-      }
-      var pointTL = this.canvas.vptCoords.tl, pointBR = this.canvas.vptCoords.br;
-      var points = this.getCoords(true, calculate);
+    isOnBoundingBox: function (tl, br, absolute, calculate) {
```
asturur, line 212: does this need to be a new public method? Do developers need it? i would prefer no.
Especially the absolute and calculate make it complicate to use as other methods we have.

src/mixins/object_geometry.mixin.js
ShaMan123: We translate vptCoords to the shadow coord system, so all points need to move by the opposite of the shadow offset.

ShaMan123: Hope this explains it well enough

src/mixins/object_geometry.mixin.js
ShaMan123: I can dump the absolute, it is useless. I just didn't know if I should break parity with the others.
I don't mind making it private but it's a good and useful method, not a private hack

src/mixins/object_geometry.mixin.js
```diff
@@ -88,10 +88,29 @@
...
 
+    /**
+     * @returns {{ offset: fabric.Point, blur: fabric.Point }}
+     */
+    getShadowOffsets: function () {
+      var blur = this.shadow.blur;
+      var shadowOffset = new fabric.Point(this.shadow.offsetX, this.shadow.offsetY),
+          blurOffset = new fabric.Point(blur, blur);
+      if (!this.shadow.nonScaling) {
+        var scaling = this.getObjectScaling(),
```
asturur, line 103: i think that on top of getObjectScaling you also need the zoom level.
Zoom will grow the shadow no matter what.

src/mixins/object_geometry.mixin.js
```diff
@@ -88,10 +88,29 @@
...
+            sx = scaling.scaleX, sy = scaling.scaleY;
+        shadowOffset.setXY(shadowOffset.x * sx, shadowOffset.y * sy);
+        blurOffset.setXY(blurOffset.x * sx, blurOffset.y * sy);
```
asturur, line 106: instead of creating the point and then set their values again, i would work on the scale factor, start at zoom level ( or 1 ) and then grows with sx and sy, and you create the points at the end.

asturur:
ok this is a bug and needs to be fixed, out of discussion.

you correctly countered the rotation issue shifting the viewport of the visible canvas in the shadow direction and accounting for bigger size for blur. Removing the issue of weird calculations.

But this require us to calculate twice once for the object and one for the shadow when the object is offscreen.
Since the goal is not stop rendering objects that may be offscreen but appear for the shadow, i would suggest we do that always when an object has shadow.

If an object has shadow, the viewport instead of translating, grow the viewport TL and BR by the amount of offset and blur, on all sides, and we do the normal calculation.

This will make not necessary to reuse the part objectIsOnBoundingBox so we remove the issue if having or not having an extra public method.

On top of that i think that zoom change shadow size and you should account that in your calculation.

ShaMan123:
If you are going in that direction I suggest we split the rendering of shadow from object so it can be rendered without the object once object is off screen and shadow is on screen.
The methods are very cheap in calculations (iterating over 4 points for object and 4 for shadow + very little calculations).
Regarding canvas zoom I am not sure, I tested in live **BUT** I need your confirmation on it.

src/mixins/object_geometry.mixin.js
ShaMan123: done

src/mixins/object_geometry.mixin.js
ShaMan123: I think you are right, pushed
[76a142d](https://github.com/fabricjs/fabric.js/pull/7698/commits/76a142d8874ab2c3920e34b23b292a99eda49594)

src/mixins/object_geometry.mixin.js
ShaMan123: made private
should I dump the `absolute` arg?

ShaMan123:
As far as I am concerned this is ready
Better to calculate if shadow is on screen than to render the object unnecessarily in terms of perf, right?

ShaMan123:
Ported to https://github.com/ShaMan123/fabric.js/pull/4
Contributions are welcome

PR #7699, opened 2022-02-18
dep(Canvas#getCenter): migrate to `getCenterPoint`

ShaMan123:
`getCenter` doesn't conform with object's `getCenterPoint` and it should.
This is non breaking, though I think we should remove `getCenter` right away.


asturur:
This is probably my fault.
Initially my first js days i thought creating an object was better than getting a new point instance.
To be honest i don't thing there is much difference, and if there is, we have to stick with it for consistency sake

asturur (approved):

Issue #7705, opened 2022-02-19
TypeError: null is not an object (evaluating 'H.scale')


andrejslogins:
```TypeError: null is not an object (evaluating 'H.scale')``` error is thrown.
variable `H` based on stacktrace seems to be `context`, likely changed due to doing a build and minimizing the code.

## Version
5.1.0

## Information about environment
**Browser**
Mobile Safari 15.3
Chrome Mobile iOS 98.0.4758
Safari 15.0 (desktop)

## Steps To Reproduce

No clear steps to reproduce at this point, though this has been popping up on sentry for my project's users.

<details><summary>Error Message & Stack Trace</summary><p>

```txt
TypeError: null is not an object (evaluating 'H.scale')
  at __initRetinaScaling(/build/main.4835ee4746e926683e9c.js:1:96381)
  at _initRetinaScaling(/build/main.4835ee4746e926683e9c.js:1:96227)
  at _initInteractive(/build/main.4835ee4746e926683e9c.js:1:123257)
  at initialize(/build/main.4835ee4746e926683e9c.js:1:122264)
  at T(/build/main.4835ee4746e926683e9c.js:1:38821)
  at ? (/build/main.4835ee4746e926683e9c.js:1:3957453)
  at i(/build/main.4835ee4746e926683e9c.js:1:1262561)
  at ? (/build/polyfills.a4bf941f0f8d5c4b9264.js:1:35310)
  at onInvokeTask(/build/main.4835ee4746e926683e9c.js:1:1124990)
  at runTask(/build/polyfills.a4bf941f0f8d5c4b9264.js:1:6605)
  at invokeTask(/build/polyfills.a4bf941f0f8d5c4b9264.js:1:12288)
```
</p></details>

## Expected Behavior
Error should not be thrown and `context.scale` should be called without issues

## Actual Behavior
In some rare cases `context.scale` is undefined

ShaMan123:
Again, I suspect you have a memory leak. Please be **100%** sure you don't.
call `fabric.runningAnimation.cancelAll()` to make sure you don't (I don't know where you should call this, at cleanup of app, and before refreshing in dev mode)
If it returns something you are leaking memory.
#7694 #7693

andrejslogins:
@ShaMan123 I'm very confident we don't have animations in our code. Never we call `.animate` for the code, so I don't see where `fabric.runningAnimation.cancelAll()` would help.
[...]

PR #7706, opened 2022-02-19
Add context check for __initRetinaScaling

andrejslogins:
Related to this https://github.com/fabricjs/fabric.js/issues/7705

In some cases, the context may not be defined.
Likely due to a function being called before the canvas is initialized, or another canvas instance is created in the old one's stead

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -235,16 +235,18 @@
         return;
       }
       var scaleRatio = fabric.devicePixelRatio;
-      this.__initRetinaScaling(scaleRatio, this.lowerCanvasEl, this.contextContainer);
+      this.__initRetinaScaling(scaleRatio, this.lowerCanvasEl, this.lowerCanvasEl.getContext('2d'));
       if (this.upperCanvasEl) {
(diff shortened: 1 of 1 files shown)


ShaMan123:
@andrejslogins I don't like this.
You don't know if the problem is at your end or not.
You'll be patching fabric with checks probably because you didn't find the source of the problem.
Investigate and return with insights and a repro or something so we can understand as well and give feedback.
This is bad practice.

> Likely due to a function being called before the canvas is initialized, or another canvas instance is created in the old one's stead
Check your cleanup methods.
Check resizing events on canvas.
Check initializing fabric more than once and throw an error if so.

asturur:
yes i would like to know how do we get to upperCanvas or lowerCanvas without the context.
When we do create them they are assigned immedialtey.

With the previous case of cacheContext that was a little bit less steady, because the cache gets created on the fly and during resizing or other events you may end up destroying and recreating it.

But this is less clear on how it happens

andrejslogins:
i'll close this for now, issue persists, but i'm not actively working on it atm

ShaMan123:
@andrejslogins try confirming your issue persists after patching fabric with #7775

PR #7709, opened 2022-02-19
chore(): Cleanup fabric.Point for v6

ShaMan123:
BREAKING Point#multiply

### DIFF
--- a/src/point.class.js
+++ b/src/point.class.js
@@ -123,9 +123,6 @@
      * @return {fabric.Point}
      */
     multiply: function (that) {
-      if (!(that instanceof fabric.Point)) {
-        throw new Error('deprecated, migrate to fabric.Point#scalarMultiply');
-      }
       return new Point(this.x * that.x, this.y * that.y);
     },
 
--- a/src/point.class.js
+++ b/src/point.class.js
@@ -118,23 +118,33 @@
     },
 
     /**
+     * Multiplies this point by another value and returns a new one
+     * @param {fabric.Point} that
+     * @return {fabric.Point}
+     */
+    multiply: function (that) {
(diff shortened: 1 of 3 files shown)


src/point.class.js
```diff
@@ -117,24 +117,34 @@
...
+    /**
+     * Multiplies this point by another value and returns a new one
+     * @param {fabric.Point} that
+     * @return {fabric.Point}
+     */
+    multiply: function (that) {
+      if (!(that instanceof fabric.Point)) {
+        throw new Error('deprecated, migrate to fabric.Point#scalarMultiply');
+      }
+      return new Point(this.x * that.x, this.y * that.y);
```
asturur, line 129: We don't have a clean way to send out dev messages about wrong usage of the api.
This would be the first one, there is a way to do so probably, with console warns more than throwing errors ( error would happen any slightly after.

Let's keep this error check out, and we can build dev directed code with the same kind of regex strip we do for svg export or import.

ShaMan123: > Let's keep this error check out, and we can build dev directed code with the same kind of regex strip we do for svg export or import.

Didn't understand...
What are you suggesting?

asturur: i suggest to not throwing error, and when we have less to do, to have a seires of console warns for clearly wrong operations on the code. This console warns would be wrapped in code comments that will allow for safe removal with regex on the minified build.
This will give us a real `developer` mode with the ability to  convey more messages.

We don't throw errors for migrations and i don't want to start now with an isolate case.

asturur (approved):
my laptop died, feel free to merge this and the scaling one once rebased on the latest change

PR #7713, opened 2022-02-19
fix(Canvas): selection + `setViewportTransform` edge case

ShaMan123:
This PR handle the following edge case:
Say an object is being actively transformed by the user and at some point a call to `setViewportTransform` changes canvas vpt.
What happens prior to this PR is that the object jumps from the mouse because of the new vpt and then jumps back once the next event fires.
This is bad.
I think that `setViewportTransform` should not affect an active object that is being transformed.
It makes no sense.

https://gkpiim.csb.app/

## Repro
http://jsfiddle.net/4zexwjo6/
1. Select and hold an object **OR** transform it **OR** edit the text. 
1. After a few seconds canvas vpt changes and the object will move to another place
2. You will experience an unwanted jump in transformation

To fix the fiddle change the cdn. 

### Before
ezgif com-gif-maker (20)

## Expected Behavior
selected object should not be transformed by canvas vpt

### After
ezgif com-gif-maker (21)


asturur:
I m not sure what is going on, the explanations are scarce.
The vieportTransform shouldn't transform anything, is just a camera view over the canvas.
What is happening that is out of ordinary?

ShaMan123:
Take a look at the demo.
I hit an edge case a while back.
If you are transforming an object (say scaling or moving) and at the same moment the canvas vpt changes it affects the active object. Resulting in a jump once the next event is triggered.

ShaMan123:
I am not sure about c1279b4
What do you think?
Anyways if we decide it's in then we need to merge #7698 first (it calls `isShadowOnScreen`)

ShaMan123:
ready

asturur:
image

Most of your demo get stuck hours like this.
I'm not sure why.

ShaMan123:
This is a repro http://jsfiddle.net/exabpkto/

ShaMan123:
> image
> 
> Most of your demo get stuck hours like this. I'm not sure why.

Probably because it's very heavy. fabric dist is copied to there

src/canvas.class.js
```diff
@@ -1349,10 +1353,25 @@
     },
 
     setViewportTransform: function (vpt) {
```
Lazauya, line 1355: Since you're working on this function it'd probably be good to add some documentation.

Lazauya (commented):
Ran locally; PR worked as intended and didn't notice any added weirdness.

src/canvas.class.js
ShaMan123: 994ef12

ShaMan123 (commented):
Fixed the main issue and found and fixed a side effect that was causing tests to fail (stale `_currentTransform` ref on canvas)

ShaMan123:
Updated description

PR #7714, opened 2022-02-20
feat(): fire `contextmenu` event

ShaMan123:
fire `contextmenu:before`, `contextmenu`
these are useful for devs


asturur:
I m ok with this, but at the next occasion update the `fires` section of the canvas documentation. (@fires contenxtMenu... ) Try to make the example of why you need both in the docs ( i m sure using prevent default in the native event does not stops firing contextMenu, so i don't see the reason for contextmenu:before )

ShaMan123:
The reason is that canvas is fired with the event first but what if you want the target to handle the event.
Then the before event is useful.

PR #7715, opened 2022-02-20
feat(Object.isType): accept multiple `type`

ShaMan123:
A simple feature testing if an object is one of an array of types
`obj.isType('rect', 'circle')`


asturur:
yeah i can agree this is useful and not adding code in a significant way.

asturur (approved):

PR #7719, opened 2022-02-20
feat(): `includeDefaultValues` arg

ShaMan123:
Support accepting `includeDefaultValues` as an argument to `toObject` methods instead of fussing with `object.includeDefaultValues`

I have been wanting this for a long time
I think we should drop `includeDefaultValues` as a prop on the object/canvas level
And I think `excludeFromExport` can go away as well if we accept a filter function
```js
toObject(props, includeDefaultValues, filter) {
// for each object
filter(obj) ? 'include': 'exclude'
}
```

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1237,13 +1237,13 @@
     /**
      * @private
      */
-    _toObject: function(instance, methodName, propertiesToInclude) {
+    _toObject: function(instance, methodName, propertiesToInclude, includeDefaultValues) {
       //If the object is part of the current selection group, it should
       //be transformed appropriately
       //i.e. it should be serialised as it would appear if the selection group
       //were to be destroyed.
       var originalProperties = this._realizeGroupTransformOnObject(instance),
-          object = this.callSuper('_toObject', instance, methodName, propertiesToInclude);
+          object = this.callSuper('_toObject', instance, methodName, propertiesToInclude, includeDefaultValues);
       //Undo the damage we did by changing all of its properties
       this._unwindGroupTransformOnObject(instance, originalProperties);
       return object;
--- a/src/gradient.class.js
+++ b/src/gradient.class.js
@@ -193,10 +193,10 @@
 
     /**
      * Returns object representation of a gradient
-     * @param {Array} [propertiesToInclude] Any properties that you might want to additionally include in the output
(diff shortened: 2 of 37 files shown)


asturur:
Shachar the PR descriptions do not match up the quality and quantity of work you are doing.
To understand what you want to do i need to read the code, and that ( because how my brain works ) makes me distracted with the code instead of the purpose of the code.

I did not ready the code yet, i would like the default behaviour was to not include default values all the time, because that creates enormous object dumps that are often not useful.
It also makes the prototype changes less effective when restoring such object states.

`includeDefaultValues` for me can be either at canvas level, library level, i agree that at object level is not that much useful.
It can also be an argument of the function toObject as you did, that is good too.
I'm ok dumping it as a property.

In general i prefer a single argument in the form of object compared to many arguments, with modern js is easier to handle.

ShaMan123:
@asturur I think `includeDefaultValues` should be false as default as well. It's very strange to put default as true.
How to move forward? Shall we dump this property? I think it's entirely redundant now because once you call `toObject` you simply pass `true` as follows `toObject(undefined, true)`
Sorry for not explaining, will do next time.

ShaMan123:
> In general i prefer a single argument in the form of object compared to many arguments, with modern js is easier to handle.

You mean `toObject({ includeDefaultValues: true, propertiesToInclude: ['stroke'] })`, that's good

asturur:
> > In general i prefer a single argument in the form of object compared to many arguments, with modern js is easier to handle.
> 
> You mean `toObject({ includeDefaultValues: true, propertiesToInclude: ['stroke'] })`? That's good

Yes but when we have es6/TS we can think of change this shape.

asturur:
For me you can move forward.
let's keep the separated arguments, get rid of the property everywhere, and let's have it as a false as default.
That will also avoid to need to check if is boolean or undefined.

ShaMan123:
@asturur what type of object is this?
https://github.com/fabricjs/fabric.js/blob/fd1b0c33f36420f343a5e917a64763a7c7b4c323/src/filters/blendimage_filter.class.js#L225

I am trying to figure out if to pass down the serialization args to it as well

ShaMan123:
Finally all tests are passing. It was horrible...

ShaMan123:
Will die out if/once removeDefaultValues is gone

PR #7720, opened 2022-02-20
fix(): add `eraser` to Object state/cache props

ShaMan123:
add `eraser` to state/cache props so object will invalidate properly


asturur (approved):

asturur:
not sure if you need more changes, if not, merge it.

PR #7710, opened 2022-02-19
MAJOR feat(): Reuse fabric.Point logic for scaling and naming consistency

ShaMan123:

PR #7728, opened 2022-02-21
chore/fix(v6): prerequisites for Group

ShaMan123:
same as #7710 

**BREAKING** `_getTransformedDimensions`, now accepts an options object.
**FIX**: `calcOCoords`


src/mixins/object_geometry.mixin.js
```diff
@@ -601,77 +601,51 @@
...
+        noSkew = options.skewX === 0 && options.skewY === 0;
 
-      if (this.strokeUniform) {
-        dimX = this.width;
-        dimY = this.height;
+      if (noSkew) {
+        finalDimensions = new fabric.Point(dimX * options.scaleX, dimY * options.scaleY);
       }
       else {
-        dimensions = this._getNonTransformedDimensions();
```
ShaMan123, line 640: @asturur was this call intentional?
Seems wrong to me because in some cases `strokeWidth` will be accounted for twice.
This could be what was causing some of the problems with `strokeUniform`

ShaMan123: my mistake

src/mixins/object_geometry.mixin.js
```diff
@@ -431,7 +431,7 @@
 
     calcLineCoords: function() {
       var vpt = this.getViewportTransform(),
-          padding = this.padding, angle = degreesToRadians(this.angle),
+          padding = this.padding, angle = degreesToRadians(this.getTotalAngle()),
```
ShaMan123, line 434: Not 100% sure about this, 90%

src/mixins/object_interactivity.mixin.js
```diff
@@ -231,28 +224,17 @@
...
       }
       this._setLineDash(ctx, styleOverride.cornerDashArray || this.cornerDashArray);
       this.setCoords();
-      if (this.group) {
-        // fabricJS does not really support drawing controls inside groups,
-        // this piece of code here helps having at least the control in places.
-        // If an application needs to show some objects as selected because of some UI state
-        // can still call Object._renderControls() on any object they desire, independently of groups.
-        // using no padding, circular controls and hiding the rotating cursor is higly suggested,
-        matrix = this.group.calcTransformMatrix();
```
ShaMan123, line 248: This hack was here because oCoords were wrong, but now they are fixed.
Why are they called oCoords????

asturur: they were called oCoords when i joined.
Is not that oCoords were wrong, is that those coords are supposed to be coordinate relative to the html element, independetly of anything else going on in fabric, so that they could match the mouse cursor easily.

This hack was here to have controls rendered correctly for people that wanted to show a selection status in a group.

src/mixins/object_geometry.mixin.js
asturur: We need to test this with skewed objects.
I'm not sure is right.

src/mixins/object_geometry.mixin.js
```diff
@@ -601,77 +605,62 @@
...
-    /*
+    /**
      * Calculate object bounding box dimensions from its properties scale, skew.
-     * @param {Number} skewX, a value to override current skewX
-     * @param {Number} skewY, a value to override current skewY
+     * @param {Object} [options]
+     * @param {Number} [options.scaleX]
+     * @param {Number} [options.scaleY]
+     * @param {Number} [options.skewX]
+     * @param {Number} [options.skewY]
```
asturur, line 623: adding skewX and skewY was an hack already.
It was added to try support skew indeed, and ask, how big would be the bounding box if skewing wasn't there.
And then use to counter calculate the growth of the object to follow a mouse movement.
Why are we adding scaleX and scaleY support to this?

src/mixins/object_interactivity.mixin.js
asturur: Sorry i miss a piece of explanation.
Before oCoords where related to controls mouse interactions and nothing else, drawing code was static and independent.
If i m not confused, now oCoords contain the point where the controls will be drawn.
So now they take care of interaction and drawing, that all in all is more correct, at least you know that where is it drawn it will be also interactive.

src/mixins/object_geometry.mixin.js
```diff
@@ -601,77 +605,62 @@
...
+        scaleY: this.scaleY,
+        skewX: this.skewX,
+        skewY: this.skewY,
+      }, options || {});
+      //  stroke is applied before/after transformations are applied according to `strokeUniform`
+      var preStroke, postStroke, strokeWidth = this.strokeWidth;
       if (this.strokeUniform) {
-        dimX = this.width;
-        dimY = this.height;
+        preStroke = 0;
```
asturur, line 637: can we call those preScalingStrokeValue and postScalingStrokeValue? ( or width )

asturur:
This change in theory breaks the control api, because is passing to the conrol position handler a different matrix depending if it is in a group or not.
I don't think that is a real issue since controls in group nowadays do not really work.

I think those changes are fine, but i want to understand why we are adding scaleX and scaleY to getTransformedDimensions.
I would be eventually more happy to remove any option from that and move the special calculations to a util.

also getTotalAngle needs a double check for edge cases.

Let me think about it.

The very important part of this PR is the matrix shuffling away from rendering and put inside calcOcoords, and that seems good to me.

src/mixins/object_interactivity.mixin.js
ShaMan123: The hack is not needed anymore

src/mixins/object_geometry.mixin.js
ShaMan123: Because I think it is needed for fixing controls rendering for nested objects

src/mixins/object_geometry.mixin.js
ShaMan123: done

ShaMan123:
It actually fixes the control api, so yes, that's breaking :).

src/mixins/object_geometry.mixin.js
ShaMan123: skew doesn't work in oCoords as well

asturur:
What is fixing that in the api?

src/mixins/object_geometry.mixin.js
asturur: at canvas level it works, but not in group currently if you have  skewY ( or skewX + skewY )

ShaMan123:
I mean that the control api was receiving bad data, wrong position of controls.
Now it is correct.

src/mixins/object_geometry.mixin.js
ShaMan123: I tried fixing oCoords to include skewing but it's above my understanding

src/mixins/object_geometry.mixin.js
asturur: if we are not using those in this PR, let's not document it

asturur:
i would expect this PR to fix this rendering here:
image

from the visual test: `controlboxes with skewY, green is wrong and needs fix`

asturur:
Or at least we should have a unit test that  changes output before and after the change.

Or someone could come in 6 months, put the matrices as they were before, thinking that are better in another position and we don't have anything that catches the regression.

If no output is changing, fine, we are just making the code more straight and more generic but we are not breaking any behaviour of fix a bug.

ShaMan123:
> If no output is changing, fine, we are just making the code more straight and more generic but we are not breaking any behaviour of fix a bug.

I didn't understand,
Did the visual test work before the change?
Are these objects inside a group?

ShaMan123:
How do I run this test? Can you try merging this on top of v6! and viewing the result?

ShaMan123:
The problem with oCoords is that the calculation doesn't take into account skewing and the bbox is calculated in 2 different ways.
I don't understand what goes where.

ShaMan123:
And I remember we said the all rendering of controls should be looked into deeply

ShaMan123:
`calcLineCoords` is broken, now I see. I doesn't take into account the group

asturur:
What i m saying is just, why no tests are failing? If no tests fails either we miss a test that cover the case we are changing, or we are not changing oCoords.

To run that test you can do:

```
npm run test:visual:single test/visual/control_rendering.js
```

you can delete `controls12.png` from the golden folder and see if it gets created equal before and after building the lib.

asturur:
> And even more we talked about dropping double coord system... I don't really understand why the calculations of aCoords and oCoords are different (except from vpt)

They are different for historical reason.
oCoords existed initially. Zoom and pan did not exists.
A developer came after a bounty to implement pan/zoom.
oCoords got adapted to work on zoom and pan.
All the functions about intersection, contain and such were adapted for the zoom.
At that point to know where an object was on the canvas you had to update oCoords, and then remove the zoom/pan from it in order to get to absolute coordinates.
At that point aCoords ( absolute coords ) were added, to represent the 4 corners of the object, in design/canvas space, without the viewport.

The 4 corners of the object and controls mostly overlapped, until in fabric 4.0 the control api was introduced that freed the controls from static position.
But mouse targeting was still relying on oCoords to find the objects, so lineCoords were added, that are the oCoords version of the aCoords.

In theory lineCoords aren't necessary anymore and aCoords should do it after a proper refactor of the pointer interaction.

Calculations are today different because oCoords are completely custom and developer driven in theory and oCoords have to support padding, that is a pain, is the equivalent of the pain strokeUniform on controls. Because padding is independent from everything ( zoom, and transformations ) but needs to be added back and pre-transformed.

ShaMan123:
Thanks for the explanation.
So I want to ask, are oCoords necessary as well? If they are good only for mouse interactions and we can transform (easily enough) the point into the canvas space is there any need for them? If they are needed shouldn't they be calculated from aCoords?
I have a more drastic idea/proposal. But a question first. Is this mess caused because of canvas transform?
I am beginning to think that canvas transform is a bad idea, too many cases to handle. However, implementing it using a Layer sounds relatively straight forward. 
What I mean to say is maybe we should rethink the design of canvas. What should it do and so on.
I believe the design spec got messed up because group was inadequate so it never came up as a solution (rather, a problem) and made everyone think of workarounds. Canvas transform seems to be such a workaround.
Think of a group/layer that goes in between the canvas and the objects as part of design. It will be in charge of transformations. The up side is that it leverages existing logic to so, object transform logic, and relieves canvas from that task.
SVG uses group for everything, it's a building block. And I think we should do that as well.
I understand the need for incremental change. It's reasonable, maintainable, steady and healthy. But I don't want us to waste time on fixing hard time-consuming logic instead of altering the design.

ShaMan123:
> This change in theory breaks the control api, because is passing to the conrol position handler a different matrix depending if it is in a group or not.

It doesn't really affect the control because I simply applied the group transform early on in calcOCoords instead of in the event handler

asturur:
oCoords are probalby not necessary anymore, but i want to take care of that part of refactor, because it requires so much old stuff that i want to be sure you don't get stuck in an endless comment ping pong.

ViewportTransform i think is not messy, is a very simple concept camera view, the issue that arise are only for things that we don't want affected by the viewport transform. padding for example and other UI elements developers want to add.

I don't want to redesign everything is an unnecessary risk and a large undertake. For developer too.

Running a test with controls and then approving ( unless my test is the failing test )

asturur:
i think the final drawing is the same:
controls13

and same test run on this code:
controls13

the drawing is broken in both cases, so we are not really breaking anything more

ShaMan123:
> oCoords are probalby not necessary anymore, but i want to take care of that part of refactor, because it requires so much old stuff that i want to be sure you don't get stuck in an endless comment ping pong.

Great

PR #7729, opened 2022-02-21
fix(): ensure scaling factor is positive for strokeUniform

ShaMan123:
remove redundant else if.
I think this might fix bugs related to `strokeUniform` because `getObjectScaling` returns abs values and here it could have been negative...


asturur:
I don't think is going to fix anything, but i agree that scaling for sizing issue is better always positive.
I added a visual test to try to cover the edge case, but there are no visual changes.

asturur:
let me change the test with dashes, maybe there something will change

asturur:
offff. there is a difference in rendering between node and browsers, well it means that least with dashes something is different.
I need to take the test image from the browser and commit that, and account for the test to be disabled under node.

ShaMan123:
BTW I am pretty sure that in some cases (e.g. shadow) the value returned from `getObjectScaling` should be abs

asturur:
i do not know because i do not know how drawing shadow works on the canvas internally.
We can test it with plain js canvas api to see how it behaves when the context is scaled negatively.

asturur (approved):

asturur:
The test i added doesn't really detect much, because the drawing is the same with or without the fix. You can merge if you are done

PR #7744, opened 2022-02-27
TEMP PR: Group collection activeselection rewrite

asturur:
This PR brings in the rewrite made by @ShaMan123 for the group and collection and active selection.
There is no new feature that can be seen, but the group class now is able to self adjust the bounding box.

### canvas_grouping.mixin.js
removed addWithUpdate and removeWithUpdate with add and remove.

### collection.mixin.js
add insertAt and remove now take an array of items rather than a single item and they take a callback.
the callback replace `_onObjectAdded` and `_onObjectRemoved` functions in the classes.
The reason for this is that a classes may want to do different things when adding or removing for different reasons.
Add Remove and InsertAt are now the `withUpdate` version, and the without update does not exist anymore.
getObjects now can take more types.

Worth to note that as of now the collection add/remove are never used directly but always wrapped in another method from 
the class that inherits from collection.

### active_selection.class.js
Not much changed. 
toGroup became exitGroup ( verify if the name of the method sound correct, together with enterGroup in group )

### object.class.js
The change isn't related to the PR and sneak in for some reason, probably because in looking at UTs changes i noticed a strong dependency between `includeDefaultValues` and `stateProperties`. Depending how many UT fails removing that change it could be worth putting it in its own commit.

### static_canvas.class.js 
add the internal implementation for add, remove and insertAt followed by a possible renderAll.
The implementation just call the collection one.
Notes:
- eventually specify the arguments in the functions.
- consider wrapping useless calls ( with no objects for example ) wit ha DEV_START regex and a console warning
- remove has a `didRemove` logic, while add and insertAt don't

### misc.js 
Removed the dependency of SVG width/height in the groupSVGElements.
Before those changes group allowed to specify which was the centerPoint for the group and run special logic on it. Was an hack from when removed the PathGroup that was a specific container for parsed SVGs.
groupSVGElements can simply group thing. The specific logic that is there about trying to keep the SVG viewport was an hack and should be implemented with a clipPath or some other render logic, or could be ignored and developers need to clean the SVG from cross border objects.

### group.class.js
Complete rewrite of the class.
Added a layout property to define the behavour of the group. More layout type will be defined in following PRs or supported as subclasses, depending on what seems to make more sense.
Added interactive property, for now disabled, used to permit to interact with objects inside the group
The initialization now attach an event listener to objects for modified, changed and selected/deselected.
When those event fire, the group will trigger a layout change.
(more to come)

### Activeselection
- removed the specific tests for add/remove/insert add and add generic one. ( tests were duplicated ? )

### Canvas
Nothing ( swapped removeWIthUpdate and AddWithUpdate with remove/add and added a couple of setCoords after a change in subTargetCheck, that is completely fine. No real functional changes, but the setCoords required needs to be noted in the changelog )

### StaticCanvas
extended the test about renderOnAddRemove, nothing else changed.

### Collection
Adapted tests to the new add/remove/insert att argument signature with a callback.
Removed the constraints about chainability

### Group
removed chainability constrains for add, remove, insertAt
Change the object shape to consider the new properties
rename destroy with removeAll

## CHANGES compared to v6!


src/mixins/collection.mixin.js
```diff
@@ -3,79 +3,69 @@
...
+   * @param {(object:fabric.Object) => any} [callback]
    */
-  insertAt: function (object, index, nonSplicing) {
-    var objects = this._objects;
-    if (nonSplicing) {
-      objects[index] = object;
-    }
-    else {
-      objects.splice(index, 0, object);
+  insertAt: function (objects, index, callback) {
```
asturur, line 38: modified to take more than one object as add and remove always did.
rare use case, maybe none, but good for consistency.
callback added because group will have 2 different equivalent of `onObjectAdded`

src/shapes/active_selection.class.js
```diff
@@ -24,58 +24,42 @@
...
+
+    /**
+     * @override
+     */
+    subTargetCheck: false,
+
+    /**
+     * @override
+     */
+    interactive: false,
```
asturur, line 40: subTargetCheck and being able to select and modify sub objects are 2 distinct use cases.

src/shapes/object.class.js
```diff
@@ -899,10 +899,9 @@
      * @param {Object} object
      */
     _removeDefaultValues: function(object) {
-      var prototype = fabric.util.getKlass(object.type).prototype,
-          stateProperties = prototype.stateProperties;
-      stateProperties.forEach(function(prop) {
-        if (prop === 'left' || prop === 'top') {
+      var prototype = fabric.util.getKlass(object.type).prototype;
+      Object.keys(object).forEach(function(prop) {
```
asturur, line 903: modified so that all properties of the exported object get compared to the prototype.

src/static_canvas.class.js
```diff
@@ -32,7 +32,8 @@
    * @fires object:added
    * @fires object:removed
    */
-  fabric.StaticCanvas = fabric.util.createClass(fabric.CommonMethods, /** @lends fabric.StaticCanvas.prototype */ {
+  // disable-eslint-next-line max-len
+  fabric.StaticCanvas = fabric.util.createClass(fabric.CommonMethods, fabric.Collection, /** @lends fabric.StaticCanvas.prototype */ {
```
asturur, line 36: collection was added at the end of the class.
Since we want to override some methods we can't do that.

ShaMan123: correct

src/shapes/active_selection.class.js
ShaMan123: I can commit this to v6! if you want, thought of it too

src/mixins/collection.mixin.js
ShaMan123: yep

src/shapes/active_selection.class.js
asturur: we will need to rebase on this once we bring another piece on master. So if this work out it will go in naturally.

asturur:
i - need - to - finish - this

asturur:
Somehow this PR breaks svg import

ShaMan123:
> Somehow this PR breaks svg import

https://github.com/fabricjs/fabric.js/pull/7669/files#diff-d2af815271fe68bc090b7f0a204a74bc05a4289fe5d8f05467fec056d757085e

https://github.com/fabricjs/fabric.js/pull/7669#issuecomment-1053704270

asturur:
i remember this discussion 3 minutes after i posted the comment. I came back now determined to fix it.

ShaMan123:
When you have time let's talk. I want to understand how you plan on moving forward.

asturur:
We can talk monday if you want, or tomorrow. But for tomorrow i have no idea when i m free.
When the baby fall asleep in the afternoon i can try to ping you and see if you are around.

The idea anyway is to merge your work in small pieces, this is the biggest piece that i couldn't really divide in sub pieces

asturur:
GroupSvgElement isn't broken at all, grouping works, but is the final position that differs.
I changed a bit the group initialization to make it work, i ll comment and ask you question.

ShaMan123:
> We can talk monday if you want, or tomorrow. But for tomorrow i have no idea when i m free. When the baby fall asleep in the afternoon i can try to ping you and see if you are around.
> 
> The idea anyway is to merge your work in small pieces, this is the biggest piece that i couldn't really divide in sub pieces

Works

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
-      options = options || {};
-      this._objects = [];
-      // if objects enclosed in a group have been grouped already,
-      // we cannot change properties of objects.
-      // Thus we need to set options to group without objects,
-      isAlreadyGrouped && this.callSuper('initialize', options);
+     * @param {boolean} [objectsRelativeToGroup] true if objects exist in group coordinate plane
+     * @return {fabric.Group} thisArg
+     */
+    initialize: function (objects, options, objectsRelativeToGroup) {
```
asturur, line 92: which are the cases in which `objectsRelativeToGroup` would be true, apart when loading from json/object?

ShaMan123:
Svg as well

בתאריך שבת, 26 במרץ 2022, 23:37, מאת Andrea Bogazzi ‏<
***@***.***>:

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
+        this.forEachObject(function (object) {
+          this.enterGroup(object, false);
+        }, this);
       }
       else {
-        this._updateObjectsACoords();
+        //  we need to preserve object's center point in relation to canvas and apply group's transform to it
+        var inv = invertTransform(this.calcTransformMatrix());
+        this.forEachObject(function (object) {
+          var center = transformPoint(object.getCenterPoint(), inv);
```
asturur, line 108: this line was:
```
var center = transformPoint(this.getCenterPoint(), t);
```

The center point was the one from the group, and the matrix was (inv * objectTransform)

now, if i m not wrong the objectTransform would move the the group center ( the 0,0 of this plane ) in the object center, taking in account also the object trasnform. But getCenterPoint is doing the same somehow.

I not 100% this change is equivalent, but fixes the svg import issue. Do you think is breaking something else?

asturur: is this supposed to take in account creating group with objects that are part of other groups and already deeply transformed?

src/shapes/group.class.js
ShaMan123: We shouldn't do that. Mainly because I want the layout mechanism to be easily overridden.

src/shapes/active_selection.class.js
ShaMan123: Why do you think it's 2 distinct use cases?

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
+          this.enterGroup(object, false);
```
ShaMan123, line 109: > is this supposed to take in account creating group with objects that are part of other groups and already deeply transformed?

This is in charge of it.

src/shapes/group.class.js
ShaMan123: I think this was a bug.
It should be the object center that is transformed into the group's plane.
Great work!

ShaMan123: I have an idea for a test to write for this. Can I commit to v6!? or should I open a PR?

ShaMan123: I want to commit this fix to v6! as well.

src/util/misc.js
```diff
@@ -607,31 +607,19 @@
...
           delete options.width;
           delete options.height;
         }
       }
-      object = new fabric.Group(elements, options);
-      if (typeof path !== 'undefined') {
-        object.sourcePath = path;
-      }
-      return object;
+      return new fabric.Group(elements, options);
```
ShaMan123, line 622:
```suggestion
      return new fabric.Group(elements, Object.assign(options || {}, { layout: 'svg' }));
```

You missed this

src/shapes/group.class.js
asturur: ok but isn't the layout mechanism unnecessary if you are reloading a group from object or a subclass? the properties of the layout are already saved in your json form

ShaMan123: In most use cases it is unnecessary.
But who knows what's the edge case.
The layout mechanism handles this case internally from `prepareInitialBoundingBox`

ShaMan123: I thought svg import uses `objectsRelativeToGroup` but I see it doesn't

src/shapes/group.class.js
asturur: we need to polish and merge this PR, either by you with a patch or we just merge this.
Changing this piece of code break a single UTs, it has some differences for nested groups.

src/util/misc.js
asturur: So the issue with the SVG is that the group creation code is perfectly able to mantain position and size, but then the position was changed from the layout strategy.
Removing options entirely works, but of course it does not work for SVG that are outside the bounding box.

src/shapes/active_selection.class.js
ShaMan123: Alright I think I understand.
e.g. ActiveSelection should not be interactive be should have subTargetCheck set to true?

src/shapes/group.class.js
ShaMan123: I'll patch and merge.
What are the diffs in tests?

ShaMan123: master:
image

before your fix:
image


with your fix:
image

I'll push a visual test for this

src/shapes/group.class.js
```diff
@@ -105,7 +105,11 @@
         //  we need to preserve object's center point in relation to canvas and apply group's transform to it
         var inv = invertTransform(this.calcTransformMatrix());
         this.forEachObject(function (object) {
-          var center = transformPoint(object.getCenterPoint(), inv);
```
ShaMan123: Why did you remove this???
It is correct

ShaMan123:
It is becoming hard to work like this, with so many branches and conflicts

src/shapes/group.class.js
asturur: i put back the original code because i have a failing UT.
I want all the test to pass and the figure out what is going on.
Now i have 2 completely unrelated UTs failing only in browser

ShaMan123: > master: image
> 
> before your fix: image
> 
> with your fix: image
> 
> I'll push a visual test for this

ShaMan123: @asturur the UT is wrong

src/shapes/group.class.js
ShaMan123: This test fails because of `group3`
Something is making it spread over the blue rect.

src/shapes/group.class.js
ShaMan123: Something is still wrong
image

ShaMan123: I have found the bug.
Give me a few minutes

ShaMan123: fixed 847284f7b93580ebe3b9fbe165bf209417ae009e

ShaMan123:
@asturur I saw you made a lot a changes to group. Why?

asturur:
I don't think i made changes willingly.
I may have tried to not bring in some parts that aren't related to the current group functionality and i may have changed some line trying to make visual test pass.

src/shapes/group.class.js
```diff
@@ -383,8 +383,7 @@
      * Execute the drawing operation for an object on a specified context
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
-    drawObject: function (ctx) {
-      this._renderBackground(ctx);
```
ShaMan123: why don't you want this?
or you want it in the next PR?
it's instead of fill property.

asturur: group does not have fill today.
I'm trying to separate refactor from new features where possible.
I'm trying to have a series of (not so) small PRs where in case of errors, we can trace back where the behavour changed and why.

If this PR can have group with the new code with the listener objects, and all the other stuff working as. it was working before, we made a huge step forward

asturur:
please don't commit on top, or i ll never get to the finish line. i know that line removed is breaking some visual test. i'll skip them.

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
+          multiplyTransformMatrices(
+            invertTransform(this.calcTransformMatrix()),
+            object.calcTransformMatrix()
+          )
+        );
+      }
+      object.setCoords();
+      object._set('group', this);
       object._set('canvas', this.canvas);
+      this._watchObject(true, object);
```
ShaMan123, line 255:
```suggestion
      this.interactive && this._watchObject(true, object);
```

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
+      var activeObject = this.canvas && this.canvas.getActiveObject && this.canvas.getActiveObject();
+      // if we are adding the activeObject in a group
+      if (activeObject && (activeObject === object || (object.contains && object.contains(activeObject)))) {
```
ShaMan123, line 258: why did you change this? bad perf

src/shapes/group.class.js
```diff
@@ -370,151 +414,436 @@
...
+    prepareInitialBoundingBox: function (layoutDirective, objects, context) {
+      var options = context.options || {},
+          hasX = typeof options.left === 'number',
+          hasY = typeof options.top === 'number',
+          hasWidth = typeof options.width === 'number',
+          hasHeight = typeof options.height === 'number';
+
+      //  performance enhancement
+      //  skip layout calculation if bbox is defined
+      if ((hasWidth && hasHeight) || context.objectsRelativeToGroup || objects.length === 0) {
```
ShaMan123, line 654: why did you change this?

ShaMan123 (commented):
@asturur it is extremely difficult for me to understand what you changed.
git diff is causing a nightmare.
I have left comments of places I notied a change that I disagree of or don't understand.
Plus I don't understand why you reintroduced the shadow logic. It is wrong and causes bugs. I verified that these shadow/group bugs were fixed once I removed the shadow methods. No need for them as far as I can tell.
Plus you can remove the `svg` layout directive logic

src/shapes/group.class.js
ShaMan123: This means that setting left/top values on group initialization will have no effect.
BUG

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
+     * @private
+     * @param {boolean} watch
+     * @param {fabric.Object} object
+     */
+    _watchObject: function (watch, object) {
+      var directive = watch ? 'on' : 'off';
+      //  make sure we listen only once
+      watch && this._watchObject(false, object);
+      // TODO check what changed ischeck
+      object[directive]('changed', this.__objectMonitor);
```
ShaMan123, line 221: changed it a textbox event

src/shapes/group.class.js
```diff
@@ -15,280 +21,318 @@
...
+    /**
+     * @private
+     * @param {boolean} watch
+     * @param {fabric.Object} object
+     */
+    _watchObject: function (watch, object) {
+      var directive = watch ? 'on' : 'off';
+      //  make sure we listen only once
+      watch && this._watchObject(false, object);
+      // TODO check what changed ischeck
```
ShaMan123, line 220:
```suggestion

```
changed is a textbox event

src/shapes/group.class.js
ShaMan123: not only is it bad perf but it is wrong as well because it checks only 1 level of nesting

src/shapes/active_selection.class.js
asturur: well once it did not need it. Now it seems to need it.
But apart from that that is minor, you may want to fire events for nested targets, but not need to select them.
We could argue that events ( mouse up, mouse down ) could fire 100% of the time since this is a breaking version, but i m not sure that would be an issue for some application that added on the mousedown/mouseup events logic that won't work on group. ( they can still filter it out checking if an object has group )

ShaMan123:
updated comments

src/shapes/group.class.js
asturur: i changed this because i was experimenting with the SVGs and this helped me fix it before i figured out the better option was not to pass in the svg options when grouping

src/shapes/group.class.js
asturur: because i did not import the other function yet. the whole descendant of is not in yet

asturur:
```
I have left comments of places I noticed a change that I disagree of or don't understand.
I don't understand why you reintroduced the shadow logic. It is wrong and causes bugs. I verified that these shadow/group bugs were fixed once I removed the shadow methods. No need for them as far as I can tell.
you can remove the svg layout directive logic
I don't understand why you removed the _renderObjects method
```

Because i wasn't done yet.
i m sure i left out rendering parts because are not related to the layout logic.
The renderObjects stuff is needed because of optimizations with the Group._activeObejects array
This PR contains only the fit-content layout, the improved geometric calculation and the watchers.

asturur:
Ok now i can document any difference and we can argue if is part of the refactor of is a new feature that we can merge on its own.

ShaMan123:
The last force push should be reverted as it reintroduced regressions

ShaMan123:
I prefer to continue in a chat

src/shapes/group.class.js
ShaMan123: Alright

src/shapes/active_selection.class.js
ShaMan123: Didn't understand

ShaMan123:
closes #7811

ShaMan123:
I think we should rebase #7761 on top of this branch and merge it here

src/util/misc.js
ShaMan123: I still see an difference in outputs margin but it doesn't surface in the visual tests

src/static_canvas.class.js
```diff
@@ -558,6 +559,43 @@
...
+     * An object should be an instance of (or inherit from) fabric.Object
+     * @param {fabric.Object|fabric.Object[]} objects Object(s) to insert
+     * @param {Number} index Index to insert object at
+     * @param {Boolean} nonSplicing When `true`, no splicing (shifting) of objects occurs
+     * @return {Self} thisArg
+     * @chainable
+     */
+    insertAt: function (objects, index) {
+      fabric.Collection.insertAt.call(this, objects, index, this._onObjectAdded);
+      this.renderOnAddRemove && this.requestRenderAll();
```
ShaMan123, line 584:
```suggestion
      objects.length > 0 && this.renderOnAddRemove && this.requestRenderAll();
```

asturur:
> I think we should rebase https://github.com/fabricjs/fabric.js/pull/7761 on top of this branch and merge it here

We can do that, i didn't have a chance to deeply study the layout strategy logic apart the early returns, so for me changing it now is the same as later.

Let me see if i m able to

asturur:
i applied the PR opening yours, copying the entire function and overwriting the one i had, since rebase got nasty quickly for me. ( plus the other minor changes to types and to initialize ).
The branch does not look bad to me at allow, it has the functionalities of basic fabricJS.
I didn't try yet a bunch of things i wanted to try, but i ll do tomorrow mornin asap.

ShaMan123:
Regarding shadow+group take a look at #7454

ShaMan123:
What about #7781 ? can we add to here as well?

src/shapes/group.class.js
ShaMan123: resolved 70bc1493a45c7d8e35d7e0fa407175249cac117c

src/shapes/group.class.js
ShaMan123: `deep` true as we discussed or simply import isDescendant of 
I would add a TODO comment so it's visible this is not complete.

ShaMan123:
It seems multi selection is working as expected once `subTargetCheck=true`

asturur:
yes the fact that it requires subTargetCheck set to true, is what i wanted to look into.

asturur:
i see what it is, is a one line fix

src/shapes/active_selection.class.js
```diff
@@ -100,6 +100,17 @@
       return false;
     },
 
+    /**
+     * @override
+     * @return {Boolean}
+     */
+    setCoords: function () {
```
ShaMan123, line 107: isn't this redundant?
group has exactly this

asturur: no it has it with subtargetcheck as an option indeed.

ShaMan123: I see.

src/shapes/group.class.js
ShaMan123: 7d793b37

src/shapes/group.class.js
ShaMan123: 7d793b37

src/shapes/group.class.js
ShaMan123: Added a TODO comment 7d793b37

PR #7758, opened 2022-03-06
fix(v6!): nested controls

ShaMan123:
Fixed nested controls.
Take a look at the goldens

I think we need to take a closer look at aCoords now after oCoords seems fixed


ShaMan123:
I want to merge into v6!

ShaMan123:
ready

ShaMan123:
Now it is abvious there's a bug in the transform event data or something.
1. skew a group
2. skew a nested object
3. rotate group
4. skew object

skewing object will make it jump to skew 0 or something like that

ShaMan123:
BTW I have a successful POC of rendering controls without oCoords.

ShaMan123:
Another thought.
I can make this mergeable into master if you prefer

ShaMan123:
And I think, with this PR, that `oCoords` can become the only set of coords if we add corner values to them.

ShaMan123:
Take a look at [5e129d1](https://github.com/fabricjs/fabric.js/pull/7758/commits/5e129d1a3839596fe6a83afca17d51cba7061a8a)
and [58e2cd5](https://github.com/fabricjs/fabric.js/pull/7758/commits/58e2cd5ea7eb03c28fb89e0668cfeb09d6ea133f)

I am no longer sure `flipX/Y` are configured correctly when rendering controls

ShaMan123:
@asturur I can rebase this PR to master. It's a prerequisite.

PR #7764, opened 2022-03-07
test(V6!): add tests

ShaMan123:
Found many layout edge cases. **Many** including origin, angle, clip path,


ShaMan123 (commented):
origin + angle goldens (angle != 0 are wrong)
golden `clipPath2` is wrong as well in position

ShaMan123:
Handled by #7744

PR #7775, opened 2022-03-08
fix(Canvas): safeguard dispose

### LINKED ISSUE #7664: Error calling toDataUrl() after upgraded to 5.0.0
## Version
5.0.0

## Test Case
https://pastebin.com/in8ySZnG

Nodejs server v16.13.2

## Steps To Reproduce

1. Create a canvas
```js
const canvas = new fabric.Canvas(null, { width: 100, height: 100 });
```
2. Create an object and add to canvas
```js
const rect = new fabric.Rect({width: 20, height: 50, fill: "#ff0000"})
canvas.add(rect)
canvas.renderAll()
```
3. Call toDataURL()
```js
const imageData = rect.toDataURL()
```
<details><summary>Error Message & Stack Trace</summary><p>

```txt
 ERROR  Cannot set property parentRule of #<CSSStyleDeclaration> which has only a getter

  at Object.setStyle (node_modules/fabric/dist/fabric.js:3158:42)
  at klass.dispose (node_modules/fabric/dist/fabric.js:10524:19)
  at klass.toCanvasElement (node_modules/fabric/dist/fabric.js:16082:14)
  at klass.toDataURL (node_modules/fabric/dist/fabric.js:16105:41)
  at server/index.js:32:29
  at klass.__setupCanvas (node_modules/fabric/dist/fabric.js:14158:17)
  at node_modules/fabric/dist/fabric.js:14124:31
  at klass._setBgOverlay (node_modules/fabric/dist/fabric.js:14175:19)
  at node_modules/fabric/dist/fabric.js:14116:13
  at node_modules/fabric/dist/fabric.js:14235:19
```
</p></details>

## Expected Behavior
`toDataURL()` would succeed at fabric 4.6.0.

## Actual Behavior
`toDataURL()` fail with error `Cannot set property parentRule of #<CSSStyleDeclaration> which has only a getter` at fabric 5.0.0

ShaMan123:
closes #7664


asturur (commented):
Can you check which of the 2 changes it the real fix?

ShaMan123:
> Can you check which of the 2 changes it the real fix?

Both are the real fix.
Added another test for setting dimensions and disposing.
Made me encounter a race condition between dispose and render.
If you look at the test you can see I had to wrap the disposing in an event to wait for after rendering.
I think this should be standard behavior handled internally by canvas, returning a promise once disposing has completed.
Thoughts?
But as you often suggest, let's limit the scope for this PR.
This on my end it's ready.

ShaMan123:
I'll explain.
Using cssText solves the problem of trying to set read only props (and is better for perf I think)
Only an interactive canvas changes lowerCanvasEl style so that's why I wrapped the style restoration as well

asturur:
weird i would imagine this
https://github.com/fabricjs/fabric.js/pull/7775/files#diff-fc54cc24e3c94d9201b66567331262bc363a509f49283a00e715aa3bf08dce35R1559-R1562
would take care of interaction between disposing and rendering

asturur (approved):

ShaMan123:
> weird i would imagine this https://github.com/fabricjs/fabric.js/pull/7775/files#diff-fc54cc24e3c94d9201b66567331262bc363a509f49283a00e715aa3bf08dce35R1559-R1562 would take care of interaction between disposing and rendering

The link doesn't open properly

Issue #7755, opened 2022-03-04
vue3/production: canvas initialization error


Zero-Qi:
在vue3下 vite打包的项目没有报错 ，开发环境也没错 
部署到项目上，无法绘制，事件监听失效，在画布上停止响应事件，捆绑到函数可以正常执行
![image](https://user-images.githubusercontent.com/50648661/156767545-8406caac-1b1b-44f8-aba1-b8c16321b197.png)

![image](https://user-images.githubusercontent.com/50648661/156767610-b183c0f8-8583-487e-8085-2c4b8123ab86.png)
 
dom异常
![image](https://user-images.githubusercontent.com/50648661/156786614-3281a768-cccb-4f4e-a50c-2b35d5fe2b65.png)
正常dom如下
![image](https://user-images.githubusercontent.com/50648661/156786697-b819dadd-764b-4286-baad-6798492b0ab7.png)

Zero-Qi:
第一次加载组件没问题，第二次的时候canvas的事件全部失效

Zero-Qi:
![Uploading image.png…]()
在这期间，浏览器只有输出这些警告

Zero-Qi:
在这期间浏览器只有输出这些警告
] Added non-passive event listener to a scroll-blocking 'mousewheel' event. Consider marking event handler as 'passive' to make the page more responsive.

Zero-Qi:
dom少绘制了一层

Zero-Qi:
封装的fabricjs组件，是通过el-dialog打开的，给canvas的id加个随机值，就解决问题了，
问题大概出在没有这个dom上，有可能是v-if的问题

ShaMan123:
@Zero-Qi please provide a working repro. And I kindly ask of you to write in English

ShaMan123:
I am able to see that the exception is very strange. There is no canvas el, only upper canvas.

Zero-Qi:
first，I thought was&nbsp;something wrong with the vite packaging，because&nbsp; in development environment is normal，no&nbsp; this bug。
[...]

PR #7776, opened 2022-03-08
fix(Canvas): Safeguard from multiple initialization

ShaMan123:
Will throw an error in this context:
```js
new fabric.Canvas(el);
new fabric.Canvas(el); // will throw
```

@asturur why was some logic set up to enable multiple canvas initialization?
~~The tests that fail, fail because of it.~~ **Updated tests**
~~I have no problem removing them and adding a multiple init test that should throw.~~

If there's a need to enable multiple init (I can't imagine why) I can patch up the PR accordingly

This will put an end to confusing bugs that are caused by this.

Motivation #7755 and more

dbl-init-error


asturur:
How do you find yourself initializing more than once?
I did it in the past but more as an escape hatch and is not defintely a feature.

ShaMan123:
I don't. I see it happening in strange bug reports. If you recall one you looked into and I closed.
Someone trying to init canvas from some event handler that runs more than once.

ShaMan123 (commented):
Added the DEV_MODE comment
Ready to merge

asturur (approved):

PR #7781, opened 2022-03-08
fix(V6!): object padding in group

### LINKED ISSUE #6989: Padding of objects in group is reset
## Version
4.2.0

## Test Case
https://jsfiddle.net/dellvolk/c8zas7th/6/

## Expected Behavior
 I need to make sure that when objects form a group, the indent for those objects remains as it was.

ShaMan123:
fixes #6989

https://jsfiddle.net/p7xzwg4o/4/


asturur:
i would hold off on this change.
padding is always external to the object, and is for controls not for the object box.
I know they seem similar as concept, but in the past removing padding from inside groups was a huge semplification for a lot of issues

ShaMan123:
Agreed

PR #7784, opened 2022-03-08
fix(itext): stop cursor on blur

### LINKED ISSUE #6370: Loosing focus on edited textbox
## Version
3.6.2

## Test Case
http://fabricjs.com/loadfonts

Nodejs or browser?
Browser

Which browsers?
Probably all

## Steps to reproduce
1. Edit text
2. While in edit mode change font using select
3. Try typing text without clicking on anything
4. Click on textbox and edit text

## Expected Behavior
Unfocusing textbox element so that the user can't see the blinking cursor.

## Actual Behavior
Textbox seems focused while in reality it's not.

ShaMan123:
closes #6370 

https://codesandbox.io/s/fabric-app-forked-6jmep4?file=/src/Basic.ts
ezgif com-gif-maker (10)

### DIFF
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -141,6 +141,7 @@
 
       this.abortCursorAnimation();
       this._currentCursorOpacity = 1;
+      clearTimeout(this._cursorTimeout2);
       if (delay) {
         this._cursorTimeout2 = setTimeout(function () {
           _this._tick();
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -141,7 +141,6 @@
 
       this.abortCursorAnimation();
       this._currentCursorOpacity = 1;
-      clearTimeout(this._cursorTimeout2);
       if (delay) {
         this._cursorTimeout2 = setTimeout(function () {
           _this._tick();
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -141,9 +141,14 @@
 
       this.abortCursorAnimation();
       this._currentCursorOpacity = 1;
-      this._cursorTimeout2 = setTimeout(function() {
-        _this._tick();
-      }, delay);
+      if (delay) {
+        this._cursorTimeout2 = setTimeout(function () {
+          _this._tick();
+        }, delay);
+      }
+      else {
+        this._tick();
+      }
     },
 
     /**
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -25,6 +25,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       fabric.document.body.appendChild(this.hiddenTextarea);
     }
 
+    fabric.util.addListener(this.hiddenTextarea, 'blur', this.blur.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'keydown', this.onKeyDown.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'keyup', this.onKeyUp.bind(this));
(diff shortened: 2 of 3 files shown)


src/mixins/itext_key_behavior.mixin.js
```diff
@@ -25,6 +25,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       fabric.document.body.appendChild(this.hiddenTextarea);
     }
 
+    fabric.util.addListener(this.hiddenTextarea, 'blur', this.blur.bind(this));
```
asturur, line 28: is this for cases when we blur out of the canvas / page but we don't actually exit editing?
or is covering some other case?

src/mixins/itext_behavior.mixin.js
```diff
@@ -141,9 +141,15 @@
 
       this.abortCursorAnimation();
       this._currentCursorOpacity = 1;
-      this._cursorTimeout2 = setTimeout(function() {
-        _this._tick();
-      }, delay);
+      clearTimeout(this._cursorTimeout2);
```
asturur, line 144: this should be taken care by `this.abortCursorAnimation`

src/mixins/itext_behavior.mixin.js
```diff
@@ -141,9 +141,15 @@
...
+      if (delay) {
+        this._cursorTimeout2 = setTimeout(function () {
+          _this._tick();
+        }, delay);
+      }
+      else {
+        this._tick();
+      }
```
asturur, line 152: i see how, with cursorDelay set to 0, the setTimeout isn't needed.
But with es6 syntax this would become 
```
his._cursorTimeout2 = setTimeout(() => this._tick(), delay);
```
And i would prefer it compare to an if.
Is the exta setTimout creating any kind of issue?

asturur:
ok so i read the original issue, and while i m fine detecting the blurs, i wonder if those happen also when we don't want.
Need to check with the chines input

src/mixins/itext_behavior.mixin.js
asturur: as discusssed @ShaMan123 please remove

ShaMan123: done 08afa25

src/mixins/itext_behavior.mixin.js
ShaMan123: As discussed this resolves a small flickering issue

ShaMan123 (commented):
ready

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -98,6 +99,10 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     this.hiddenTextarea && this.hiddenTextarea.focus();
   },
 
+  blur: function () {
+    this.abortCursorAnimation();
```
asturur, line 103: i have the last minute thought and this should be optional.
this.stopCursorOnBlur && this.abortCursorAnimation();
This is something that seems logical but only depending on what your app is doing when you are focusing out.
If i need to add, as an example, an emoji from an emoji panel, i probably want to see my cursor.

asturur: What do you think?

ShaMan123: I agree, but the dev can override easily

ShaMan123: I can add a bit about it

asturur (approved):
i still think it should be optional, because the same could be said on the other way, put a noop on the event handler, and override if you want to stop cursor animatio.

ShaMan123:
For the sake of conversation:
I expect it to follow this behavior, it's how a normal textbox behaves.
In general I think it's better for the dev to override working functionalities in order to disable them rather than trying to make something work. It's easier and doesn't require knowledge of internals.

PR #7788, opened 2022-03-09
feat(): dataURL export - filter objects

ShaMan123:
Added a nice and simple option to filter objects for canvas export

This option is shallow, it doesn't filter objects under groups, yet...


melchiar (approved):
this is great!

ShaMan123:
Thanks. So should I merge?

ShaMan123:
I am thinking of having in supported for group as well.
But it might be simpler to clone the group and run `toDataURL` on the clone and support it under the hood.
What do you think?

melchiar:
> Thanks. So should I merge?

looks good to me



> I am thinking of supporting group as well. But it might be simpler to clone the group and run `toDataURL` on the clone **NOT** supporting it under the hood. What do you think?

sounds good, as I recall that's how toDataURL works on a fabric canvas as well

ShaMan123:
> sounds good, as I recall that's how toDataURL works on a fabric canvas as well
It uses exiting objects on a new canvas

salomonsanz:
"excludeFromExport" already exists. In the future, to "centralize" all export options, in the future it might be nice to accept an object instead of a boolean

```
excludeFromExport = {
    json : true,
    dataUrl : true,
    svg : true
}
```

ShaMan123:
I didn't think of `excludeFromExport`. Good point.
I'd prefer removing this strange but useful prop in favor of filtering functions everywhere.

salomonsanz:
mmm, from my experience, most filters of that type are based on object properties, not on dynamic logic, so in the end it would be redundant to delete that property and pass everything to functions:

`objects.filter(o=>o.excludeFromExport)`

On the other hand, there is no need to choose between functions and boolean properties, in many frameworks a design pattern that allows a value or a function is common.

https://laravel.com/docs/9.x/helpers#method-value

ShaMan123:
I want to make fabric more intuitive. That's why `excludeFromExport` is on my target.
As you said, it is unclear how it should behave and some cases. Should it apply for every export? `toCanvasElement` , `toDataURL`, `toJSON`, `clone`. A filter function, however, it very clear and easy to maintain.

PR #7796, opened 2022-03-10
feat(Text): bidi intial support

ShaMan123:
First step to support bidi text ~~including `auto` dir prop~~

This is very promising.
Take a look at the state of selection thanks to this PR and to #7674 
@amirhossein92 

Didn't touch cursor/selection logic that much.
You can see textbox behavior when pressing the arrows

ezgif com-gif-maker (18)
ezgif com-gif-maker (17)

---

I have experimented with a much more elegant solution, using a `hiddenTextarea` to determine the direction using selection changes.
[POC](https://jsfiddle.net/8zdmxr9s/)
So this is even more promising than before.


src/shapes/text.class.js
```diff
@@ -879,6 +887,27 @@
...
+        return this.direction === 'auto' ?
+          fabric.util.getElementStyle(fabric.document.body, 'dir') || 'ltr' :
+          this.direction;
+      }
+      else if (charIndex === 0) {
+        var prevLine = this.__charBounds[lineIndex - 1];
+        return prevLine[prevLine.length - 1].dir;
+      }
+      else {
+        return this.__charBounds[lineIndex][charIndex - 1].dir;
```
ShaMan123, line 907: O wonder if this is the desired logic for spaces

asturur:
this would be great!
I want to sync on this i had a strategy in mind, and the implementation imho should be straight forward ( a bit of trial and error in the render loop and cursor position for sure )

ShaMan123:
I have a good POC for bidi selection

src/shapes/text.class.js
asturur: i can ask to people that are expert about this.
But in general unless we find evidence, a space is one of the nor left nor right chars, so it behaves like the others that are such, it look at the context.

src/shapes/text.class.js
```diff
@@ -879,6 +887,33 @@
...
+      }
+      else {
+        return this._getGraphemeFallbackDirection(lineIndex, charIndex);
+      }
+    },
+
+    _getGraphemeFallbackDirection: function (lineIndex, charIndex) {
+      if (lineIndex === 0 && charIndex === 0) {
+        return this.direction === 'auto' ?
+          fabric.util.getElementStyle(fabric.document.body, 'dir') || 'ltr' :
```
asturur, line 905: i think auto is fine but is not related to bidi.

src/shapes/text.class.js
```diff
@@ -72,6 +72,13 @@
      */
     _reWords: /\S+/g,
 
+    /**
+     * https://stackoverflow.com/questions/12006095/javascript-how-to-check-if-character-is-rtl/14824756#14824756
+     */
+    _reExplicitRTL: /^[\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC]+$/,
+
+    _reExplicitLTR: /^[A-Za-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02B8\u0300-\u0590\u0800-\u1FFF\u2C00-\uFB1C\uFDFE-\uFE6F\uFEFD-\uFFFF]+$/,
```
asturur, line 80: I think we can find some official source compared to stackoverflow.

asturur:
For me an initial PR that just adds dir to measurement bounding boxes for chars, is good enough.
It will be easy to review, and merged fast.

src/shapes/text.class.js
ShaMan123: surprising. I thought the 2 are 1.
I shall remove it.

ShaMan123:
Regarding spaces. From my experience they inherit the direction of the textbox unless they are in between 2 chars that are the opposite direction. I shall push a commit that addresses this.

src/shapes/text.class.js
ShaMan123: https://www.w3.org/International/questions/qa-html-dir#dirauto

ShaMan123:
I think we might die until we manage to draft an impl of the spec.
So I suggest we use the same workaround that itext uses.
We use textboxes, inject characters and use `getComputedStyle` to get the direction.
If we have 2 textboxes for each direction we can determine what character is weak.
I'll give it a try.

src/shapes/text.class.js
asturur: Once bidi support mostly work, the developer still does have to set rtl and ltr to dictate how the textbox is oriented.
With bidi not working/not done, auto is still usable to detect the user possible preference, right?
I meant those are 2 features that do not intersect. Did i get it right or am i wrong?

asturur:
using getComputedStyle at render time for strings can be slow.

I see there are strong isolate, spillover cases and specific codepoints.

Remains a fact that today i can't write correctly:

`In persian, 'XXXXX' means hello`

And writing in persian

`XXXXXXXXX 'hello'  XXXXX XXXX`

And having even just a basic implementation is a step forward for use cases.

I personally know a developer that was waiting for this feature, in any form.
(on top of some firefox old issues to solve) (can't remember the handle Avi Mistrel)

ShaMan123:
I remember you mailed me something regarding this person.
How is form.submit vs. getComputedStyle in terms of speed? Take a look:
https://jsfiddle.net/j5oshp0w/3/
I will use caching as well.

src/shapes/text.class.js
asturur: If i check the links you posted, the direction auto does not seem to infuence the character itself, but the line ( we don't even have different line support ).
So if the textbox is set to auto, the first strong character typed dictates the the line orientation.
Do i read it correctly?

asturur:
> How is form.submit vs. getComputedStyle in terms of speed? Take a look:

Then we get a strong dependence on JSDOM corectness in node.
Rendering a piece of text in node becomes more complicated than already is.

I really vote for the simple regex and a direction on the bounding box. It gives us the info that we need at render time, is easy to test,  and yes it won't be perfect but will work for most of the cases.

asturur:
@mistriel maybe you are still looking forward this feature.
Or gave up :)

src/shapes/text.class.js
ShaMan123: solved

src/shapes/text.class.js
ShaMan123: Correct

src/shapes/text.class.js
```diff
@@ -785,32 +838,31 @@
...
     /**
      * measure every grapheme of a line, populating __charBounds
      * @param {Number} lineIndex
-     * @return {Object} object.width total width of characters
-     * @return {Object} object.widthOfSpaces length of chars that match this._reSpacesAndTabs
+     * @return {Number} total width of characters
      */
     _measureLine: function(lineIndex) {
       var width = 0, i, grapheme, line = this._textLines[lineIndex], prevGrapheme,
-          graphemeInfo, numOfSpaces = 0, lineBounds = new Array(line.length),
```
ShaMan123, line 806: removed unused var

ShaMan123 (commented):
ready

ShaMan123:
This is the main point of the spec I believe
https://unicode.org/reports/tr9/#Resolving_Embedding_Levels

ShaMan123 (commented):
This is far from perfect.
but it's major progress and I'm fed up with it.
tedious work

mistriel:
> 

Hi we have created this [Doc](https://docs.google.com/document/d/1rXT4Xj2lbH5gXTSWsMgqgXD7eUzHXvGTWX07vMsCPOI/edit?usp=sharing) to track all the RTL & Bidi issues we saw in fabric. 

## Firefox 
Also one of the major issues we saw in Firefox was the lack of support of [ctx.direction](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/direction) that is not implemented in Gecko  [Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1728999) still open and not discussed 

I don't think it worth the try - but i'll try to request triage about this as well.

asturur:
i can take over, i wanted to work over this.

asturur:
I have just a question, why did you remove `widthOfSpaces` from the measurements?

ShaMan123:
I didn't see anything using it.

asturur:
> I didn't see anything using it.

is used for justify, i think it needs to be there. i ll double check from this branch

src/mixins/itext_behavior.mixin.js
```diff
@@ -473,6 +473,7 @@
         this.hiddenTextarea.selectionStart = newSelection.selectionStart;
         this.hiddenTextarea.selectionEnd = newSelection.selectionEnd;
       }
+      this.hiddenTextarea.setAttribute('dir', this.direction);
```
ShaMan123, line 476: We might want to check if the direction has changed on the object since the textarea was updated somewhere and run an update if so

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -11,6 +11,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     this.hiddenTextarea.setAttribute('spellcheck', 'false');
     this.hiddenTextarea.setAttribute('data-fabric-hiddentextarea', '');
     this.hiddenTextarea.setAttribute('wrap', 'off');
+    this.hiddenTextarea.setAttribute('dir', this.direction);
```
asturur, line 14: there was a reason why i held on this initially, but i don't remember.
I think because it was changing how selection gets calculated and was incosistente between firefox and chrome

asturur:
commit [fe3b3db](https://github.com/fabricjs/fabric.js/pull/7796/commits/fe3b3dbb96580689bfdcc4d3703c901c964e206b) is totally fine on its own.

ShaMan123:
> commit [fe3b3db](https://github.com/fabricjs/fabric.js/pull/7796/commits/fe3b3dbb96580689bfdcc4d3703c901c964e206b) is totally fine on its own.

You mean to extract to a separate PR?

src/mixins/itext_key_behavior.mixin.js
mistriel: Firefox is now getting this [direction](https://bugzilla.mozilla.org/show_bug.cgi?id=1728999) issue fixed

mistriel: Fixed indeed - verified with nightly

mistriel: https://jsfiddle.net/mistriel/4mu386t5/  works similar on both browsers !

chiragbytes:
@ShaMan123 anyone is going to approve this PR in the near future, as we want to fix issues with RTL text urgently.

ShaMan123:
We will get to after TS

mistriel:
@ShaMan123  can i assist checking it ?

ShaMan123:
ברור

mistriel:
> 

@asturur  it's highly important for me as well.

ShaMan123:
You can try out the v6-alpha branch.
Not highly recommended any longer since it has become stale in relation to master. 
The divergence is too large so I have abandoned it.

ShaMan123:
If it is any comfort we have turned a corner in the migration

chiragbytes:
> You can try out the v6-alpha branch. Not highly recommended any longer since it has become stale in relation to master. The divergence is too large so I have abandoned it.

Hello @ShaMan123 I have checked out to v6-alpha-branch in my node modules and then verified with my concerns, still I am facing the same issue.

Thank you

ShaMan123:
I have experimented with a much more elegant solution, using a `hiddenTextarea` to determine the direction using selection changes.
[POC](https://jsfiddle.net/5gLr8na1/4/)
So this is even more promising than before.

Issue #7732, opened 2022-02-22
Changing `freeDrawingBrush` `strokeDashArray` property changes `strokeDashArray` of selection


sunsetninja:
## Version
5.2.1

## Test Case
https://codesandbox.io/s/trusting-williamson-uh8rxx?file=/src/index.js

## Information about environment
Chrome 98.0.4758.102

## Steps To Reproduce
In the test case above
1. Enable free drawing mode
2. Draw something on a canvas
3. Disable free drawing mode
4. Start selection by dragging mouse

## Expected Behavior
`strokeDashArray` of selection isnt change

## Actual Behavior
`strokeDashArray` of selection is changed

<img width="514" alt="image" src="https://user-images.githubusercontent.com/13559663/155097665-a72413ba-5536-46d8-893c-bbd16ba9a224.png">

ShaMan123:
It's quite strange that `_setBrushStyles` is called outside of `_saveAndTransform`

ShaMan123:
https://codesandbox.io/embed/mystifying-goldwasser-cdfvc9?fontsize=14&hidenavigation=1&theme=dark&view=preview

PR #7802, opened 2022-03-11
feat(IText): Draggable text

ShaMan123:
### General
Adds support for native like text dragging! Out of the box!
This PR aims to support text dragging, making fabric's text objects feel/behave like a HTML textarea.
It is an expected interaction from the user's point of view.

closes #7800 

### Changes
- drag/drop event logic on canvas/text:
   -  firing DnD events with more data for the dev
   - addressed an issue with firing order on drag over targets and the selected target
   - `onDragStart` and `canDrop` are public methods for overriding that control drag and drop respectively
   - `dragStart` logic cycle - imitates spec, invokes target's `onDragStart` method, if result is `true` drag starts. Behaves like `onSelect`.
   - creating a drag image from `onDragStart`
   - `dragEnd` calls `_onMouseUp` synthetically - the browser doesn't do it and fabric needs it to finalize interaction logic
- cursor/selection rendering on text - in order to render text selection on the drag source while dragging and the cursor on the drop target while dragging over it (same as what the browser does)

### Thoughts
I think we ought to change the `selectWord` method to align with the behavior of html textarea which selects the word and all the trailing whitespace (it's important for drag and drop). 

### TO DO
- [x] **perhaps wait for #7842 for `DragEvent.dataTransfer`**
- [ ] retina scaling for drag image - couldn't manage to fix the blur (will be handled in a follow up if ever)
- [x] #7904 

ezgif com-gif-maker (11)
ezgif com-gif-maker (12)


src/mixins/itext_click_behavior.mixin.js
```diff
@@ -146,12 +149,38 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
...
+   * @private
+   * @param {DragEvent} e 
+   * @returns {boolean} should handle event
+   */
+  onDragStart: function (e) {
+    this.__dragStartFired = true;
+    if (this.__isDragging) {
+      var selectedText = this.text.slice(this.selectionStart, this.selectionEnd);
+      e.dataTransfer.setData('text/plain', selectedText);
+      //e.dataTransfer.setDragImage(this._cacheCanvas, 0,0);
```
ShaMan123, line 163: Here we should set the drag image, I need help.

ShaMan123: Seems like the drag image element needs to be in the html body for this to work

ShaMan123:
What about tests?

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,173 @@
...
+          dragImage.remove();
+        };
+        //  position drag image offsecreen
+        fabric.util.setStyle(dragImage, {
+          position: 'absolute',
+          left: -dragImage.width + 'px'
+        });
+        fabric.document.body.appendChild(dragImage);
+        var offset = correction.add(diff);
+        e.dataTransfer.setDragImage(dragImage, offset.x, offset.y);
```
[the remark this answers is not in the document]
ShaMan123: d112f3b9

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,252 @@
...
+        fill: 'transparent',
+        textBackgroundColor: 'transparent'
+      };
+      this.setSelectionStyles(styleOverride, 0, data.selectionStart);
+      this.setSelectionStyles(styleOverride, data.selectionEnd, data.text.length);
+      var dragImage = this.toCanvasElement({ enableRetinaScaling: enableRetinaScaling });
+      this.backgroundColor = bgc;
+      this.styles = styles;
+      //  handle retina scaling
+      if (enableRetinaScaling && retinaScaling > 1) {
```
Lazauya, line 462: can't we scale down too?

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,252 @@
...
+        this.enterEditing(e);
+        var isSpace = /\s/;
+        var insertingOnSpace = isSpace.test(this._text[insertAt]);
+        var insertingAfterSpace = isSpace.test(this._text[insertAt - 1]);
+        var seletionStartOffset = 0;
+        //  drag and drop in same instance
+        if (this.__dragStartSelection) {
+          var selectionStart = this.__dragStartSelection.selectionStart;
+          var selectionEnd = this.__dragStartSelection.selectionEnd;
+          if (insertAt > selectionStart && insertAt <= selectionEnd) {
```
Lazauya, line 629: See other comment on inequality order.

Lazauya (commented):
Mostly minor stuff, 1 or 2 possible bugs.

src/mixins/itext_behavior.mixin.js
ShaMan123: What do you mean?

src/mixins/canvas_events.mixin.js
```diff
@@ -207,14 +211,43 @@
...
     },
 
+    /**
+     * supports native like text dragging
+     * @private
+     * @param {DragEvent} e
+     */
+    _onDragStart: function (e) {
+      var activeObject = this.getActiveObject();
+      if (activeObject && typeof activeObject.onDragStart === 'function' && activeObject.onDragStart(e)) {
```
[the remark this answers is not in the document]
ShaMan123: true if target should be dragged

https://github.com/fabricjs/fabric.js/blob/72b9a777fa8eed146c9a6cb4c05aff665af4f4f8/src/mixins/itext_behavior.mixin.js#L516

src/mixins/itext_click_behavior.mixin.js
```diff
@@ -129,6 +129,10 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     // we want to avoid that an object that was selected and then becomes unselectable,
     // may trigger editing mode in some way.
     this.selected = this === this.canvas._activeObject;
+    // text dragging logic
+    var newSelection = this.getSelectionStartFromPointer(options.e);
+    this.__isDragging = this.isEditing && newSelection >= this.selectionStart && newSelection <= this.selectionEnd
```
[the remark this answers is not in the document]
ShaMan123: It's personal flavor. I prefer this.

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,252 @@
...
+          }
+          else if (insertAt > selectionEnd) {
+            insertAt -= selectionEnd - selectionStart;
+          }
+          this.insertChars('', null, selectionStart, selectionEnd);
+          // prevent `dragend` from handling event
+          delete this.__dragStartSelection;
+        }
+        //  handle spacing
+        if (trailingSpace && this.findLineBoundaryLeft(insertAt) < insertAt && !insertingAfterSpace
```
ShaMan123, line 641: @Lazauya Thanks! Did you live test it?
I think I encountered a bug when dragging a selected word. In some cases white space injection might be wrong.
The logic starts here

src/mixins/itext_behavior.mixin.js
Lazauya: You never check if the scaling is less than 1, unless I'm missing something.

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,252 @@
...
+        dragImage = c;
+      }
+      this.__dragImageDisposer && this.__dragImageDisposer();
+      this.__dragImageDisposer = function () {
+        dragImage.remove();
+      };
+      //  position drag image offsecreen
+      fabric.util.setStyle(dragImage, {
+        position: 'absolute',
+        left: -dragImage.width + 'px'
```
melchiar, line 478: Should we maybe have `border: none` on here?

image

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,252 @@
...
+      this.backgroundColor = bgc;
+      this.styles = styles;
+      //  handle retina scaling
+      if (enableRetinaScaling && retinaScaling > 1) {
+        var c = fabric.util.createCanvasElement();
+        c.width = dragImage.width / retinaScaling;
+        c.height = dragImage.height / retinaScaling;
+        var ctx = c.getContext('2d');
+        ctx.scale(1 / retinaScaling, 1 / retinaScaling);
+        ctx.drawImage(dragImage, 0, 0);
```
melchiar, line 468: don't forget to account for viewport transform

melchiar (changes requested):
This is great stuff! you even managed to get dragging text between objects working 👍

I've made a couple of comments inline. Also, we need to account for object scale.

src/mixins/itext_behavior.mixin.js
melchiar: also, would it not be better to keep the dragImage in the original resolution (with retinascaling) and just resize the drag image's dom element so that it stays in high res?

src/mixins/itext_behavior.mixin.js
ShaMan123: canvas handles that
https://github.com/fabricjs/fabric.js/blob/65b3731ad220c3a6fede8ec4bf88aa10eb34d43d/src/static_canvas.class.js#L226

src/mixins/itext_behavior.mixin.js
ShaMan123: Please elaborate

src/mixins/itext_behavior.mixin.js
ShaMan123: 72b9a777

src/mixins/itext_behavior.mixin.js
melchiar: I tested on a high dpi display. The drag image is at a lower dpi than the canvas.

ShaMan123: I have seen that as well.
Any suggestions?

ShaMan123:
> This is great stuff! you even managed to get dragging text between objects working 👍
> 
> I've made a couple of comments inline. Also, we need to account for object scale.

Object scale seems fine to me. vpt not.

src/mixins/itext_behavior.mixin.js
melchiar: just spent some time looking into this... your code looks correct, I'm really not sure what's causing the dragimage to have a lower res 😐

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,253 @@
...
+          {};
+        var styles = data.styles;
+        var trailingSpace = data.trailingSpace, trailing = insert[Math.max(0, insert.length - 1)];
+        this.canvas.discardActiveObject();
+        this.canvas.setActiveObject(this);
+        this.enterEditing(e);
+        var isSpace = /\s/;
+        var insertingOnSpace = isSpace.test(this._text[insertAt]);
+        var insertingAfterSpace = isSpace.test(this._text[insertAt - 1]);
+        var seletionStartOffset = 0;
```
melchiar, line 625: *selectionStartOffset

src/mixins/itext_behavior.mixin.js
```diff
@@ -412,6 +418,253 @@
...
+      this.__isDraggingOver = false;
+      // inform browser that the drop has been accepted
+      e.preventDefault();
+      var insert = e.dataTransfer.getData('text/plain');
+      if (insert) {
+        var insertAt = this.selectionStart;
+        var data = e.dataTransfer.types.includes('application/fabric') ?
+          JSON.parse(e.dataTransfer.getData('application/fabric')) :
+          {};
+        var styles = data.styles;
```
melchiar, line 617: @ShaMan123 @asturur 

Something to consider here. This currently allows dragging of inline styles from one object to another. Some applications enforce a one style per object behavior restriction, using the `fabric.disableStyleCopyPaste` property to prevent styles from being copied from one object to another. I'd be inclined to put a check for this property here as well. Maybe rename it to `useDestinationStyle`?

ShaMan123: hmmm I don't like the flag itself because it isn't flexible enough.
You mean to have a property on Text? It's so diverse, that's why I left methods public to make it easy to override.
You simply listen to the `drop:before` event and use `DragEvent.setData('application/fabric',...)` to change the data accordingly, Then the `drop` handler` won't have style data to work with, or whatever the dev does.
I think this is best practice, it's how most frameworks do it.

src/mixins/itext_behavior.mixin.js
ShaMan123: done

src/mixins/canvas_events.mixin.js
```diff
@@ -374,7 +407,9 @@
      * @param {Event} e Event object fired on mousemove
      */
     _onMouseMove: function (e) {
-      !this.allowTouchScrolling && e.preventDefault && e.preventDefault();
+      var activeObject = this.getActiveObject();
+      !this.allowTouchScrolling && (!activeObject || !activeObject.__isDragging)
+        && e.preventDefault && e.preventDefault();
```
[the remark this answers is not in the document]
ShaMan123: It's like this all over fabric so I'll leave it this way for now, but thanks

src/mixins/itext_behavior.mixin.js
ShaMan123: Or even from `dragstart` using `effectAllowed`, many possibilities

ShaMan123:
image

ShaMan123 (commented):
Well this is great!
Please review and let's get this merged.
If you have a test concept I am willing-ish.
This is really an important piece of UX.
Regarding the retina scaling issue of the drag image. If you are alright with it let's move forward and merge leaving it as is for the next iteration.

melchiar:
The pointer position still doesn't appear to be accounting for an object's scale.

Test object:

`canvas.add(new fabric.Textbox('this text is draggable', {
            left: 300,
            top: 100,
            width: 300,
            scaleX: 2,
            scaleY: 2
        }));`

ShaMan123:
> canvas.add(new fabric.Textbox('this text is draggable', { left: 300, top: 100, width: 300, scaleX: 2, scaleY: 2 }));

It doesn't seem to be related to this PR

melchiar:
> > canvas.add(new fabric.Textbox('this text is draggable', { left: 300, top: 100, width: 300, scaleX: 2, scaleY: 2 }));
> 
> It doesn't seem to be related to this PR

hmm, you're right. It looks like the issue is present in the current master.

src/mixins/itext_behavior.mixin.js
```diff
@@ -152,24 +160,23 @@
...
-      // to clear just itext area we need to transform the context
-      // it may not be worth it
-      if (shouldClear && canvas) {
-        canvas.clearContext(canvas.contextTop || canvas.contextContainer);
-      }
 
+      //  make sure we clear context even if instance is not editing
+      if (shouldClear) {
+        var ctx = this._clearContextTop();
+        ctx && ctx.restore();
```
asturur, line 178: why this lonely restore? where the save counterpar is?

asturur:
@ShaMan123 there are changes to the copy-paste logic, plus removal of the optional `disableStyleCopyPaste`.
How does relate to drag/drop changes?
If they don't restore the changes.

src/mixins/itext_behavior.mixin.js
```diff
@@ -417,6 +424,270 @@
...
+      this.__dragImageDisposer = function () {
+        dragImage.remove();
+      };
+      //  position drag image offsecreen
+      fabric.util.setStyle(dragImage, {
+        position: 'absolute',
+        left: -dragImage.width + 'px',
+        border: 'none'
+      });
+      fabric.document.body.appendChild(dragImage);
```
asturur, line 486: i wish there was a better solution for this.
Could there be weird css transform that makes this image visible in the body regardless of us setting the left offscreen?

src/shapes/itext.class.js
```diff
@@ -276,31 +282,57 @@
...
+     * clear top context
+     * @returns {CanvasRenderingContext2D|undefined} ctx
+     */
+    clearContextTop: function() {
+      if (!this.isEditing) {
+        return;
+      }
+      var ctx = this._clearContextTop();
+      ctx && ctx.restore();
+      return ctx;
```
asturur, line 319: ok clearContextTop and  _clearContextTop  is worse than before imho.
Is not fixable now because everything is interleaved and we use it i 6 places.
2 of those 6 wants to execute code between the cleaning and the restore part.

asturur: possibly in the future we want to have clearing more blunt and the places that need to execute extra code will have their own save/restore couple and transform

asturur:
Regarding the drag functionality, my issue with this PR is that  this feature shouldn't be mandatory.
Being able to create a new selection on top of the one that exist right now is a functionality that would get lost.
The native textarea behaviour is inconsistent, meaning that sometimes i can start a new selection, sometimes a drag operation starts.

src/mixins/canvas_events.mixin.js
```diff
@@ -207,40 +212,179 @@
...
+        this._dragSource = activeObject;
+        var options = { e: e, target: activeObject };
+        this.fire('dragstart', options);
+        activeObject.fire('dragstart', options);
+        addListener(this.upperCanvasEl, 'drag', this._onDragProgress);
+        return;
+      }
+      e.preventDefault();
+      e.stopPropagation();
+    },
```
asturur, line 232: i would prefer to call this `_onDragStartTextInternal` ( or similar ) and short circuit if the active object type does not inherit from text rather than checking if the activeObject has onDragStart function.

If more drag/drop functionality get added in the future, each functionality can have its own handler and they can be turned off easily by not registering the event handler or de-registering it.

src/mixins/canvas_events.mixin.js
```diff
@@ -207,40 +212,179 @@
...
+     */
+    _onDragProgress: function (e) {
+      var options = {
+        e: e,
+        dragSource: this._dragSource,
+        dropTarget: this._draggedoverTarget
+      };
+      this.fire('drag', options);
+      this._dragSource && this._dragSource.fire('drag', options);
+    },
```
asturur, line 280: while this fires like 30 times per second when you do not move the mouse, and like 120 times per second when moving the mouse. Just to say never put anything substantial.
I understand we need this to render the cursor, otherwise i would happily skip this.

src/mixins/canvas_events.mixin.js
```diff
@@ -207,40 +212,179 @@
...
+        e: e,
+        target: this._draggedoverTarget,
+        subTargets: this.targets,
+        dragSource: this._dragSource
+      });
+      //  fire dragleave on targets
+      this._fireEnterLeaveEvents(null, e);
+      //  clear targets
+      this.targets = [];
+      this._hoveredTargets = [];
```
asturur, line 352: are those necessary? isn't the `this._fireEnterLeaveEvents(null, e);` fired from dragOver good enough to fire dragenter and dragleave for those cases?
When we enter a canvas with a drag operation, we also trigger dragOver.

src/mixins/itext_behavior.mixin.js
ShaMan123: `clearContextTop` -> `_clearContextTop` -> `prepareContextTop` -> `ctx.save()`

src/mixins/itext_behavior.mixin.js
ShaMan123: I have no idea about the css.
I guess if css is something like
```css
img {
 transform: scale(5000, 5000)
}
```
The image will definitely be visible.
But it must be attached, couldn't find a workaround.
We should add a data attribute + perhaps a class to enable mitigation of such cases.
Shall I push a commit?

src/shapes/itext.class.js
ShaMan123: What are you suggesting?
I worked around with what I found.

src/mixins/canvas_events.mixin.js
ShaMan123: I added this to comply with the spec.
AFAIK it has no side effects. It's meant for the dev only, fabric doesn't use it.
The cursor is handled by the browser's drag event functionality.

src/mixins/canvas_events.mixin.js
ShaMan123: I disagree because this way any object can utilize this logic out of the box.
And the place to disable the event is from `onDragStart`  - this is from the spec.

src/mixins/canvas_events.mixin.js
ShaMan123: I added tests that show it doesn't.
There was an edge case, not sure I remember correctly, but I think it was dragging on canvas and leaving the canvas, drag would continue or something strange. Perhaps dragleave didn't fire on the drag source.

ShaMan123: I see there's `config.canvasEvtIn` `config.canvasEvtOut` as a param of `fireSyntheticInOutEvents` which I didn't know of but is unused in drag events.
We could change that but I'm not sure what would be the expected behavior. Should canvas fire `dragenter/leave` to comply with the spec or should it fire whenever a target changes? Current logic follows spec.

src/shapes/itext.class.js
ShaMan123: I can merge `clearContextTop` and `_clearContextTop` into `clearContextTop(restore: boolean)`.
That's the only difference between the two.

ShaMan123:
Regarding these:

> @ShaMan123 there are changes to the copy-paste logic, plus removal of the optional `disableStyleCopyPaste`. How does relate to drag/drop changes? If they don't please restore the changes.

These changes are similar in logic. It makes copy and paste comply with the spec as well. the spec of copy/paste and drag/drop is pretty much the same, both use dataTransfer, so the work/logic is the same. And if issues raise they should be related.


> Regarding the drag functionality, my issue with this PR is that this feature shouldn't be mandatory. Being able to create a new selection on top of the one that exist right now is a functionality that would get lost. The native textarea behaviour is inconsistent, meaning that sometimes i can start a new selection, sometimes a drag operation starts.

I disagree. The logic is sound. If you start dragging over a selected text a drag operation starts. Otherwise selection is handled by default logic.

src/mixins/canvas_events.mixin.js
asturur: But what specs? we don't care for real event specs. Our events are just a mean for devs to interven in things that happen. We don't need to mimic anything if not a generic flow that is expect.
If we can just pin down in a comment which is the edge case we are adding those 2 handlers for, we can solve it later.

ShaMan123: I mean it is expected behavior from frag/drop in HTML. That's what I was after.
And it conforms with
>  Our events are just a mean for devs to intervene in things that happen. We don't need to mimic anything if not a generic flow that is expect.

I expect canvas to fire drag enter/exit as HTML element and not for each change of target. But I don't mind supporting both. Sound reasonable.
I will try to remember what the edge case was. Why are you against solving it in the PR?

asturur:
If i want to disable accidental drags i need to be able to.
The HtmlTextarea totally allows you to restart selections on top of the existing one, without deselecting. It seems that in order to start a drag, it require your mouse down event before the movement to last at least a fraction of a second.

Text dragging is a nice addition, it adds a lot of code, but is not buildable without some sort of internal support.
If you want to merge it, please leave copy paste changes out, and make it optional somehow.

We can revisit copy paste that today works, when we cleared out our backlog

Paste is a text event, we don't support paste on canvas, that is more an application handling thing because usually in order to do so, you have to create an hidden textarea anyway to support paste event, a div or a canvas do not support the paste event if i remember correctly.

src/mixins/canvas_events.mixin.js
asturur: I don't want to open the doors to drag things that we didn't plan to support.
We are not adding text drag support, with a skeleton of infrastructure to do maybe more dragging, but we don't really know.
It is easier to add later than rollback

ShaMan123:
> If i want to disable accidental drags i need to be able to. The HtmlTextarea totally allows you to restart selections on top of the existing one, without deselecting. It seems that in order to start a drag, it require your mouse down event before the movement to last at least a fraction of a second.
> 
> Text dragging is a nice addition, it adds a lot of code, but is not buildable without some sort of internal support. If you want to merge it, please leave copy paste changes out, and make it optional somehow.

It should behave the same this:
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_textarea

I will extract copy/paste as you wish.

src/mixins/canvas_events.mixin.js
asturur: I meant the fabricJS text cursor that indicates the drop position of text. It seems to me that without this event we can't draw it, and so we need it.

ShaMan123:
Dragging selected text starts a drag. UX standard if click + drag on selected text to start a new selection

src/mixins/canvas_events.mixin.js
asturur: the change of target is somehow necessary if you want to make any change when dragging over a tearget, as you probably find out with the enterEditing/exitEditing text logic.
Same is for mousedown and mouseup, you would expect them on canvas level, but you need them on object level to do anything meaningful

My point is, add only events that are strictly necessary to text dragging implementation.

src/shapes/itext.class.js
ShaMan123: That's how it was kind of

src/mixins/itext_behavior.mixin.js
asturur: yes i found it later, at some point we need to make it more evident. Lonely restore or save are bugs ready to happen

src/mixins/itext_behavior.mixin.js
asturur: no is an edge case, i just prefer to interct with the dom as little as possible

src/mixins/itext_behavior.mixin.js
```diff
@@ -417,6 +424,270 @@
...
+        this.__isDraggingOver = true;
+        this.enterEditing(e);
+        this.__isDragging && this.abortCursorAnimation();
+      }
+      else if (this.__isDraggingOver && !canDrop) {
+        //  drop state has changed
+        this.__isDraggingOver = false;
+        !this.__isDragging && this.clearContextTop();
+        this.exitEditing();
+      }
```
asturur, line 555: couldn't this go on a dragEnter event? Wouldn't it be its natural place?

src/mixins/itext_behavior.mixin.js
```diff
@@ -417,6 +424,270 @@
...
+        //  can be dropped, inform browser
+        e.preventDefault();
+        //  inform event subscribers
+        options.canDrop = true;
+        options.dropTarget = this;
+        //  render
+        this.setCursorByClick(e);
+        this._updateTextarea();
+        this.restartCursorIfNeeded();
+        this.renderCursorOrSelection();
```
asturur, line 566: it would be nice, but can be left for later, to optimize this to run only when the cursor change position rather than as fast as the browser thinks it needs to happen.

src/mixins/itext_behavior.mixin.js
ShaMan123: It might be a good idea to mark the image but I don't care

src/mixins/itext_behavior.mixin.js
ShaMan123: agreed.
I think we can do something like
```js
runInContext(callback: (ctx: Canvas2DRenderingContext) => any)
```
I am more than willing to migrate to that right away.

src/mixins/canvas_events.mixin.js
ShaMan123: > The cursor is handled by the browser's drag event **native** functionality.

ShaMan123: We do not need it

src/mixins/canvas_events.mixin.js
ShaMan123: I will push a commit for canvas level event named `object:dragenter` or something like that.
I wish to leave canvas drag enter/exit, important for UI decisions.

src/mixins/itext_behavior.mixin.js
ShaMan123: I think I get what you mean. Checking if selection has changed and then rendering

src/mixins/itext_behavior.mixin.js
ShaMan123: I don't remember why I use the drag over event.
Perhaps it was before I fixed enter/leave firing order, so at the time it was unreliable.

ShaMan123: I think I understand now... 
Consider the edge case where an object is actively dragged over and `canDrop` is toggled during the drag over.
I think the object remains the target of the event so this case won't be handled from the `dragenter` event.

src/mixins/itext_behavior.mixin.js
ShaMan123: I have tried this. It doesn't simply work... in case we are dragging over the drag source because we need to render both cursor and selection at that case... I thought it would work. Something to do with cursor animation timer I guess.

src/mixins/itext_behavior.mixin.js
ShaMan123: I decided to add but not remove
I think is best this way
7777841e

src/mixins/canvas_events.mixin.js
ShaMan123: > are those necessary? isn't the this._fireEnterLeaveEvents(null, e); fired from dragOver good enough to fire dragenter and dragleave for those cases?
When we enter a canvas with a drag operation, we also trigger dragOver.

A better explanation:
They are necessary, it's instead of `this._simpleEventHandler('dragover', e);`

src/mixins/itext_behavior.mixin.js
ShaMan123: brush suffers from this as well BTW. There's an open issue related.

ShaMan123 (commented):
minors PR #7904 , merge at will.
I have addressed all our comments both in comments and PR. Take a look.

src/mixins/itext_behavior.mixin.js
asturur: contextTop i think is the only place where we do this in a non straight forward way.
If there is more let's find a strategy.
Can you link me the issue?

src/mixins/itext_behavior.mixin.js
asturur: it is not retina scaled because there is probably no css for the drawImage.
indeed here https://github.com/fabricjs/fabric.js/pull/7802/files#diff-af5c45c9018fd92f35ba43f5e25303b5fe3b627d0f81d2d78b328746fe015b4eR468-R473 we are dividing the natural size of the image by the retina factor, making it smaller.
Not sure if the dragged image can targeted by a selector, or can be wrapped in a div and added a class that would retina-scale it.
Is not that much important we can try later

asturur: If i remember correctly, when handling drag/drop with react.js i remember the dragged element can be anything. I believe you can remove the division by retina scaling and add a div around the canvas that will divide the size by 2 with css and so get the hiDpi effect.

src/mixins/itext_behavior.mixin.js
ShaMan123: #7732

src/mixins/itext_behavior.mixin.js
```diff
@@ -417,6 +424,270 @@
...
+        var selection = this.__dragStartSelection = {
+          selectionStart: this.selectionStart,
+          selectionEnd: this.selectionEnd,
+        };
+        var value = this._text.slice(selection.selectionStart, selection.selectionEnd).join('');
+        var data = Object.assign({ text: this.text, value: value }, selection);
+        e.dataTransfer.setData('text/plain', value);
+        e.dataTransfer.setData('application/fabric', JSON.stringify({
+          value: value,
+          styles: this.getSelectionStyles(selection.selectionStart, selection.selectionEnd, true)
```
ShaMan123, line 508: @melchiar so to make sure...
The use of this function wasn't affected by #8006 right?

ShaMan123 (commented):
merge #7904 at will

melchiar:
@ShaMan123 I think we need add some conditions for the type of data being dropped to prevent the drop cursor effect during other types of drops like files. Just did some tests and you get some strange behavior when dropping a file onto a text object.

draggable-text

ShaMan123:
Could you explain what we are seeing? What is expected and the actual behaviour?
I see a bbox that is wrong
Besides that i don't understand

melchiar:
Yes, wrong bounding box, then after selecting a different object and then trying to select the first one again you get no bounding box at all though it's still in text editing mode.

ShaMan123:
Tried to reproduce on windows Chrome without success.
Could you post the dataTransfer data so I could have a look?

melchiar:
I've simplified my test and the tiny bounding box was actually an app issue. The issue of the object being stuck in edit mode with no bounding box remains though. 

Here's my dataTransfer data:

image

Also, firefox seems render the hover image in the wrong position/size on highdpi displays (my pixel ratio is 1.5)

draggable-text-firefox

ShaMan123:
could you upload the test into a sandbox?
I don't have firefox...

ShaMan123:
@melchiar test on http://fabricjs.com/test/misc/itext.html

melchiar:
same issues running on itext.html. Drag an image file into a unselected textbox and it gets stuck in edit mode with no control box. Only way to get it back is by dragging text into it from another box.

The hover image position/size issue only occurs in firefox, chrome is fine.

draggable-text-firefox

src/mixins/itext_behavior.mixin.js
```diff
@@ -417,6 +424,270 @@
...
+        dragImage.remove();
+      };
+      //  position drag image offsecreen
+      fabric.util.setStyle(dragImage, {
+        position: 'absolute',
+        left: -dragImage.width + 'px',
+        border: 'none'
+      });
+      fabric.document.body.appendChild(dragImage);
+      e.dataTransfer.setDragImage(dragImage, offset.x, offset.y);
```
ShaMan123, line 487: @melchiar try manipulating the offset value
Perhaps firefox accounts for retina scaling?

ShaMan123:
> mplified my test and the tiny bounding box was actually an app issue. The issue of the object being stuck in edit mode with no bounding box remains though.
> 
> Here's my dataTransfer data:

This doesn't seem like the data transfer from the drop event, seems like from a drag event.

ShaMan123:
can you make sure all drag/drop events are fired in the correct order?
start
enter
over
exit
drop 
end

src/mixins/canvas_events.mixin.js
asturur: Did you use in particular? can you make a good use case? otherwise let's not add it.

asturur:
There are a bunch of things i want to experiment with before merging.
In particular i don't like:
- the use of enter/exit editing and update textarea to draw the destination cursor for the drop
- the particular meaning we attribute to dropEffect ( we read it on the event, we never assign it, meaning that someone else has assigned it and there is a contract on the meaning ).

ShaMan123:
1. Sorry about the PR description. You are right.
2. The `effectAllowed` is set on `dragStart` - the systems populates `dropEffect`, the docs saw you can change it but experimenting with it showed differently if I recall correctly, see MDN [dropEffect](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/dropEffect) [effectAllowed](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/effectAllowed)

src/mixins/object_interactivity.mixin.js
```diff
@@ -299,6 +299,17 @@
      */
     onSelect: function() {
       // implemented by sub-classes, as needed.
-    }
+    },
+
```
ShaMan123, line 303: I guess we should/could add `onDragStart: () => false` here, as `onDragStart` and `canDrop` are public methods to for overriding

asturur (approved):

src/mixins/itext_behavior.mixin.js
```diff
@@ -512,6 +506,7 @@
         e.dataTransfer.effectAllowed = 'copyMove';
         this.setDragImage(e, data);
       }
+      this.abortCursorAnimation();
```
ShaMan123, line 509: for future follow up
this line might be causing selection to be cleared on drag start

PR #7803, opened 2022-03-12
fix(IText): behaviour of move cursor down after select word

Superlin-supreme:
Hi, there

I found a problem on IText and tried to fix it

When the user selects a word and presses down key, the cursor does not move to the right position, unlike <textarea>

if it is not the last line, the cursor will move down and close to the right side, but this is not a problem, it's design to be like this
case_1
In the same case, <textarea> will move the cursor to the left
textarea_1

What's worse, when the same case happen in the last line, the cursor will not move to the end of text, and the right key doesn’t work after that
case_2

It is not difficult to fixed this problem, and I change the test unit of this case

Hope I can help 🦆

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -478,18 +478,13 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
   },
 
   /**
-   * Moves cursor up without shift
+   * Moves cursor up / down without shift
    * @param {Number} offset
    */
   moveCursorWithoutShift: function(offset) {
-    if (offset < 0) {
-      this.selectionStart += offset;
-      this.selectionEnd = this.selectionStart;
-    }
-    else {
-      this.selectionEnd += offset;
-      this.selectionStart = this.selectionEnd;
-    }
+    // moves cursor down / up both start from selectionStart
+    this.selectionStart += offset;
+    this.selectionEnd = this.selectionStart;
     return offset !== 0;
   },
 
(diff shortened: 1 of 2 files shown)


ShaMan123:
Good timing, we are working on IText :)

ShaMan123:
You should checkout the #7674 branch and see if what you experience happens after applying the fix

Superlin-supreme:
> You should checkout the #7674 branch and see if what you experience happens after applying the fix

Hi, This branch does not solve the situation I meet.

I just realize that the problem happen when an emoji 🦆 appears in the text, and I added a test of this case.

You can try it on the home page of fabricjs.com.

asturur:
hi @Superlin-supreme and thanks for your first pr, complete of unit tests.
I have been out a couple of weeks.

I ll look at this branch asap.
That textarea behaviour you pointed out, is it consistent across browsers?

Superlin-supreme:
> 

Hi @asturur ,
I tried this behavour of textarea in Chrome and Firefox, and found it to be inconsistent across browsers.
so, how do think about this behabour.
Of course, I would like to revert the change of `src/mixins/itext_key_behavior.mixin.js`, cause it will change the feature of this.
And keep the change of `src/mixins/itext_behavior.mixin.js` to fix the case happen in the last line.

Superlin-supreme:
> Hi @asturur, I tried this behavour of textarea in Chrome and Firefox, and found it to be inconsistent across browsers. So, how do think about this behabour. Of course, I would like to revert the change of `src/mixins/itext_key_behavior.mixin.js`, cause it will change the feature of this. And keep the change of `src/mixins/itext_behavior.mixin.js` to fix the case happen in the last line.

I revert the change of behavior when move cursor down.

PR #7806, opened 2022-03-14
fix(Textbox): expose methods for overrides + fix resize filckering

ShaMan123:
The first 2 methods are to make text more accessible from subclassing
1. `graphemeSplit` on instance
2. `wordSplit`

4. fixed line wrapping when word > width
**BEFORE**
ezgif com-gif-maker (14)

**FIXED**
ezgif com-gif-maker (15)

### DIFF
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -692,6 +692,7 @@
         oldWidth = target.width,
         newWidth = Math.ceil(Math.abs(localPoint.x * multiplier / target.scaleX) - strokePadding);
     target.set('width', Math.max(newWidth, 0));
+    //  check against actual target width in case `newWidth` was rejected
     return oldWidth !== target.width;
   }
 
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -690,9 +690,9 @@
         strokePadding = target.strokeWidth / (target.strokeUniform ? target.scaleX : 1),
         multiplier = isTransformCentered(transform) ? 2 : 1,
         oldWidth = target.width,
-        newWidth = Math.abs(localPoint.x * multiplier / target.scaleX) - strokePadding;
+        newWidth = Math.ceil(Math.abs(localPoint.x * multiplier / target.scaleX) - strokePadding);
     target.set('width', Math.max(newWidth, 0));
-    return oldWidth !== newWidth;
+    return oldWidth !== target.width;
   }
 
   /**
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1513,7 +1513,7 @@
     },
 
     /**
-     *
+     * Override this method to customize grapheme splitting
      * @param {string} value
      * @returns {string[]} array of graphemes
      */
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1513,8 +1513,8 @@
     },
 
     /**
-     * 
-     * @param {string} value 
+     *
+     * @param {string} value
      * @returns {string[]} array of graphemes
      */
     graphemeSplit: function (value) {
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -1513,6 +1513,15 @@
     },
 
     /**
+     * 
+     * @param {string} value 
+     * @returns {string[]} array of graphemes
+     */
+    graphemeSplit: function (value) {
+      return fabric.util.string.graphemeSplit(value);
+    },
+
+    /**
      * Returns the text as an array of lines.
      * @param {String} text text to split
      * @returns {Array} Lines in the text
@@ -1523,7 +1532,7 @@
           newLine = ['\n'],
           newText = [];
       for (var i = 0; i < lines.length; i++) {
-        newLines[i] = fabric.util.string.graphemeSplit(lines[i]);
+        newLines[i] = this.graphemeSplit(lines[i]);
         newText = newText.concat(newLines[i], newLine);
       }
       newText.pop();
--- a/src/shapes/textbox.class.js
+++ b/src/shapes/textbox.class.js
@@ -279,15 +279,16 @@
 
     /**
      * Helper function to measure a string of text, given its lineIndex and charIndex offset
-     * it gets called when charBounds are not available yet.
+     * It gets called when charBounds are not available yet.
+     * Override if necessary
+     * 
      * @param {CanvasRenderingContext2D} ctx
      * @param {String} text
      * @param {number} lineIndex
      * @param {number} charOffset
      * @returns {number}
-     * @private
      */
-    _measureWord: function(word, lineIndex, charOffset) {
+    measureWord: function(word, lineIndex, charOffset) {
       var width = 0, prevGrapheme, skipLeft = true;
       charOffset = charOffset || 0;
       for (var i = 0, len = word.length; i < len; i++) {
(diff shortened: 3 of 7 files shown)


src/controls.actions.js
```diff
@@ -690,9 +690,10 @@
         strokePadding = target.strokeWidth / (target.strokeUniform ? target.scaleX : 1),
         multiplier = isTransformCentered(transform) ? 2 : 1,
         oldWidth = target.width,
-        newWidth = Math.abs(localPoint.x * multiplier / target.scaleX) - strokePadding;
+        newWidth = Math.ceil(Math.abs(localPoint.x * multiplier / target.scaleX) - strokePadding);
     target.set('width', Math.max(newWidth, 0));
-    return oldWidth !== newWidth;
+    //  check against actual target width in case `newWidth` was rejected
+    return oldWidth !== target.width;
```
asturur, line 696: oh good point

src/shapes/text.class.js
```diff
@@ -1512,6 +1512,15 @@
...
     },
 
+    /**
+     *
+     * @param {string} value
+     * @returns {string[]} array of graphemes
+     */
+    graphemeSplit: function (value) {
+      return fabric.util.string.graphemeSplit(value);
+    },
```
asturur, line 1522: this feels extra, but i have zero issues with it.
Why not swapping fabric.util.string.graphemeSplit directly?

src/shapes/textbox.class.js
```diff
@@ -271,7 +271,7 @@
       var wrapped = [], i;
       this.isWrapping = true;
       for (i = 0; i < lines.length; i++) {
-        wrapped = wrapped.concat(this._wrapLine(lines[i], i, desiredWidth));
+        wrapped.push.apply(wrapped, this._wrapLine(lines[i], i, desiredWidth));
```
asturur, line 274: here there is a slight change.
before wrapped was always a new array, while now is mutated.
Is that something you want? or you dont care? i don't care.

src/shapes/textbox.class.js
```diff
@@ -328,14 +337,21 @@
         words.push([]);
       }
       desiredWidth -= reservedSpace;
-      for (var i = 0; i < words.length; i++) {
+      var data = words.map(function (word) {
         // if using splitByGrapheme words are already in graphemes.
-        word = splitByGrapheme ? words[i] : fabric.util.string.graphemeSplit(words[i]);
+        word = splitByGrapheme ? word : this.graphemeSplit(word);
+        var width = this._measureWord(word, lineIndex);
+        largestWordWidth = Math.max(width, largestWordWidth);
```
asturur, line 344: what is going on here? We are measuring twice, without using the offset, tha means without counting styles, and that means that the measuring can be extremely wrong.
You can access offset in that loop, but why we need to loop over words twice?

src/shapes/textbox.class.js
```diff
@@ -298,6 +298,15 @@
       return width;
     },
 
+    /**
+     *
+     * @param {string} value
+     * @returns {string[]} array of words
+     */
+    wordSplit: function (value) {
```
asturur, line 306: if this is meant to ease overrides, please let's add a comment here to suggest that to developers

src/shapes/text.class.js
asturur: if this is meant to ease overrides, please let's add a comment here to suggest that to developers

asturur:
This seems good, just give some explanation on why we are not using the offset information on the first pass of measuring.
If we do, then we don't need to remeasure later, we can use cached measures.
Or am i missing the point of that double loop?
(i think i because we want to know immediately which is the large word of all text before we take decisions on intermediate long words

src/shapes/textbox.class.js
ShaMan123: I don't care either.
What do think? Any side effects?

src/shapes/text.class.js
ShaMan123: > i think i because we want to know immediately which is the large word of all text before we take decisions on intermediate long words

Absolutely!

ShaMan123:
> i think i because we want to know immediately which is the large word of all text before we take decisions on intermediate long words

Absolutely!

src/shapes/text.class.js
ShaMan123: [f6864cb](https://github.com/fabricjs/fabric.js/pull/7806/commits/f6864cba191c8aea1629ab4dc86c9377465e8ffa)

ShaMan123:
I don't think a util should be overridden, that's why I've exposed `graphemeSplit`
Renamed `_measureWord` to `measureWord` and made it public. Crucial for overrides and completes what I was after.
But is kind of **BREAKING** (the method was private so not so much)

ShaMan123:
2 tests are failing, no idea why yet

src/shapes/textbox.class.js
asturur: if is not measurable in performance it does not matter.

src/shapes/textbox.class.js
```diff
@@ -271,23 +271,25 @@
...
+     * 
      * @param {CanvasRenderingContext2D} ctx
      * @param {String} text
      * @param {number} lineIndex
      * @param {number} charOffset
      * @returns {number}
-     * @private
      */
-    _measureWord: function(word, lineIndex, charOffset) {
+    measureWord: function(word, lineIndex, charOffset) {
```
asturur, line 292: to be 100% honest with consumer, we can leave it with the low dash, port it to 5.x ( because is a nice change ) and then we can rename it in a single commit later.

src/shapes/textbox.class.js
```diff
@@ -328,14 +340,25 @@
...
-      for (var i = 0; i < words.length; i++) {
+      // measure words
+      var data = words.map(function (word) {
         // if using splitByGrapheme words are already in graphemes.
-        word = splitByGrapheme ? words[i] : fabric.util.string.graphemeSplit(words[i]);
-        wordWidth = this._measureWord(word, lineIndex, offset);
+        word = splitByGrapheme ? word : this.graphemeSplit(word);
+        var width = this.measureWord(word, lineIndex, offset);
+        largestWordWidth = Math.max(width, largestWordWidth);
+        offset += word.length + 1;
```
asturur, line 349: why here is lenght + 1 and before was jsut lenght?

src/shapes/textbox.class.js
ShaMan123: bb262710

src/shapes/textbox.class.js
ShaMan123: I thought it might boost perf. That's why.
Leaving it as is.

src/shapes/textbox.class.js
```diff
@@ -353,10 +376,6 @@
         infixWidth = splitByGrapheme ? 0 : this._measureWord([infix], lineIndex, offset);
         offset++;
```
ShaMan123, line 377: > why here is lenght + 1 and before was jsut lenght?

Here offset is incremented. I thought it's because of the space between worsds.
That's why I thought I should increment as well. But tests fail.

src/shapes/textbox.class.js
ShaMan123: I don't really understand how style position is counted.
With spaces or without... Doesn't make sense without. 
Need help

ShaMan123: The test was wrong
image

ShaMan123: The style of the test textbox object was incomplete.
fixed by ff6b99db

src/shapes/textbox.class.js
ShaMan123: Resolved

ShaMan123:
Good on my end.
Ready to merge

asturur (approved):

asturur:
this is a good nice fix and improvement.
And of little scope, thanks.

PR #7811, opened 2022-03-15
Update canvas.class.js

ShaMan123:
change check from group to collection


asturur:
is this a bugfix? what is actually changing?

ShaMan123:
This changes makes any collection of objects support nested selection.
All it needs is `subTargetCheck = true` and `_objects`

I opened these PRs because I didn't want to commit to v6!, not to disturb your reviewing process

ShaMan123:
This is what can be achieved from the change
https://github.com/fabricjs/fabric.js/pull/7669#issuecomment-1067121710

ShaMan123:
Handled by #7744

PR #7816, opened 2022-03-16
Update itext_key_behavior.mixin.js

milonic:
Non ASCII character on line 19. Hyphen in padding-top was non standard


ShaMan123 (approved):
Thanks!

PR #7827, opened 2022-03-24
feat(): support aborting promises/requests

ShaMan123:
Motivation #7075 

Gist
https://developer.mozilla.org/en-US/docs/Web/API/AbortController
https://dev.to/abdus/terminating-dom-operations-at-will-abortcontroller-in-javascript-1icm
https://nodejs.org/api/globals.html#class-abortcontroller

In Action
https://codesandbox.io/s/mpcq6z


ShaMan123:
I think we should consider migrating `XMLHttpRequest` to `fetch`

src/mixins/canvas_serialization.mixin.js
```diff
@@ -20,8 +30,9 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
    * });
    */
   loadFromJSON: function (json, reviver) {
+    this.abortLoadingTask();
     if (!json) {
-      return;
+      return Promise.reject(new Error('fabric.js: `json` is undefined'));
```
ShaMan123, line 35: BREAKING

src/mixins/canvas_serialization.mixin.js
```diff
@@ -30,42 +41,45 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
...
-    return fabric.util.enlivenObjects(serialized.objects || [], '', reviver)
-      .then(function(enlived) {
+    return Promise.all([
+      fabric.util.enlivenObjects(serialized.objects || [], { reviver: reviver, signal: abortController.signal }),
+      fabric.util.enlivenObjectEnlivables({
+        backgroundImage: serialized.backgroundImage,
+        backgroundColor: serialized.background,
+        overlayImage: serialized.overlayImage,
+        overlayColor: serialized.overlay,
+        clipPath: serialized.clipPath,
```
ShaMan123, line 57: can I/should I add the eraser property here?

smoke (dismissed):
Awesome!

ShaMan123:
Because this means bumping node to v16 I want @asturur and @melchiar to share their thoughts

ShaMan123:
It seems the fetch API was introduced to node js in v17.5. If we want to migrate to it we can use `node-fetch` or something else. Or leave fabric.util.request for now.

.github/workflows/node-unit-tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [15.4.0, 16.x]
```
asturur, line 18: if AbortController is 15.4 and up, we can remove 14 and keep 16 only. We do odd numbers just when they are the latest

ShaMan123: 639223e4

ShaMan123:
@melchiar @asturur once this is merged (it fixes some logic related to loading from json) I want to prepare for the js/ts migration by adding a util method `registerClass` that registers a class to a registry that fabric will use to instead of `resolveNamespace`.
This will make subclassing much simpler. The dev will simply need to register the subclass. It will use the subclass' static type property. This why you could simply override core methods by subclassing `fabric.Object` for example and registering it with the same type **ending** the use of prototype direct manipulation

ShaMan123:
@melchiar @asturur once this is merged (it fixes some logic related to loading from json) I want to prepare for the js/ts migration by adding a util method `registerClass` that registers a class to a registry that fabric will use to instead of `resolveNamespace`.
This will make subclassing much simpler. The dev will simply need to register the subclass. It will use the subclass' static type property. This why you could simply override core methods by subclassing `fabric.Object` for example and registering it with the same type **ending** the use of prototype direct manipulation

asturur:
we need to decide at some point which browsers are supported by fabric 6.

Abort controller set the minimum at those

image

That is not bad.

Safari 12.1 is from 2019 and is on the newish side for me

ShaMan123:
I don't have an opinion, not very knowledgeable regarding these matters.
All I care about are good native features that make life easier 😋. But fabric should support the majority I guess.
Shit storms in the form of IE will find their right place in the previous decade

.github/workflows/node-unit-tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [16.x]
```
asturur, line 18: this is a bit extreme for my taste, node 14 is supported up to apr 2023 in general, getting rid of it way before seems bad from a support point of view.

ShaMan123: if AbortController is 15.4 and up, we can remove 14 and keep 16 only. We do odd numbers just when they are the latest

_Originally posted by @asturur in https://github.com/fabricjs/fabric.js/pull/7827#discussion_r835652818_

.github/workflows/node-unit-tests.yml
asturur: i know i said this before, i didn't check the schedule of 14. Isn't there a polyfill or something we can suggest?

.github/workflows/node-unit-tests.yml
asturur: yep i found my comments later.

asturur:
read all the PR, and the only thing i disagree is providing the abort controller ourselves.
The abort signal from the controller should be sent from outside from the developer ( that wants to use it ), this is how i interacted with aws uploads in s3 for example, they accept an abort signal from outside, but don't create one for you.
This would also remove the issue with the supported node version.

We would also not need to provide a method to interrup the current task, supporting the existance of an abort signal is all we should surface.

.github/workflows/node-unit-tests.yml
ShaMan123: https://www.npmjs.com/package/abortcontroller-polyfill

Look fine.
And polyfills `fetch` as well

src/mixins/canvas_serialization.mixin.js
```diff
@@ -30,42 +46,48 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
       : Object.assign({}, json);
 
     var _this = this,
-        renderOnAddRemove = this.renderOnAddRemove;
+        renderOnAddRemove = this.renderOnAddRemove,
+        abortController = new AbortController();
```
ShaMan123, line 50: You mean here?

src/shapes/image.class.js
```diff
@@ -373,10 +387,20 @@
      */
     setSrc: function(src, options) {
       var _this = this;
-      return fabric.util.loadImage(src, options).then(function(img) {
+      if (this.abortLoadingTask()) {
+        this.fire('loading:aborted');
+      }
+      var abortController = new AbortController();
+      this.__abortController = abortController;
```
ShaMan123, line 394: This is trickier
I think it should remain as is

ShaMan123 (commented):
> the only thing i disagree is providing the abort controller ourselves.

src/mixins/canvas_serialization.mixin.js
ShaMan123: IMO loadFromJSON is a strange thing.
It should behave like loadSVGFromXXX and let the dev add objects to canvas.
In such a case it makes sense to pass an abort controller to the loading task, and for that task to be separated from canvas state.

src/shapes/image.class.js
ShaMan123: e.g. 
```js
img.set('src', someURL)
// very little time passed
img.set('src', someOtherURL)
```

src/mixins/canvas_serialization.mixin.js
ShaMan123: ?

asturur: that is an entire different issue.
LoadFromJSON today restore a previous canvas, is not a bulk loader, while it could totaly be, but we have enliveObjects from that.

What i mean is that we should not wrap the abortController functionality with custom code.
We should let it pass down by options, arrive to the only thing that can be interrupted and is the request of an image and let the developer call it.

When a developer call set('src', '....') he will verify if there is a previous abort controller going on and can cancel it or not.

ShaMan123: How can a dev use `set('src', '....')` with an abort controller?
I don't see how it's possible.
Meaning it will make one of the possibilities die out.
It is possible using `setSrc` but it's messy to tell the dev "if you want to set `src` with an abort controller then don't use `set`"
Regarding the json task... I think all async tasks should be abortable.

ShaMan123: > LoadFromJSON today restore a previous canvas, is not a bulk loader, while it could totaly be, but we have enliveObjects for that.

OK but what if the dev calls loadFromJSON before the prev task completed. And let's say the the prev task is really heavy. If we don't abort it it will override the incoming task once it completes. That's why I've handled in such a way and that's why I think that mixing canvas state and loading tasks is bad practice.

src/shapes/image.class.js
asturur: i don't think we should guard against this internally. I want to leave loading patterns to the developer. Give me some hour, my laptop is now dead, but i was trimming down this to show you i want to support abort signal.

ShaMan123: I understand what you mean.
I simply think we should support it as long as the loading task is part of the object's state.
Because it causes bugs that are hard to trace.
It reminds me the double init of canvas bug.
Concurrent bugs are unexpected.

ShaMan123: .

src/mixins/canvas_serialization.mixin.js
```diff
@@ -20,8 +33,11 @@ fabric.util.object.extend(fabric.StaticCanvas.prototype, /** @lends fabric.Stati
    * });
    */
   loadFromJSON: function (json, reviver) {
+    if (this.abortLoadingTask()) {
+      this.fire('loading:aborted', { from: 'json' });
+    }
     if (!json) {
-      return;
+      return Promise.reject(new Error('fabric.js: `json` is undefined'));
```
asturur, line 40: good catch. this should have been committed earlier in a one line PR.

asturur (changes requested):
I insist for the Abort controller being passed from outside in the functions where we decided to generate it automatically.
That does not change the functionality you are proposing, and allows for better integration of loading tasks.
The operation can be cancelled from inside or from outside, and most importantly the same abort signal from the used abort controller can be passed to other systems that needs to be cancelled on coordination.

ShaMan123:
I still don't understand how it solves the bug I described or how I can set `src` using `img.set`. I am unconvinced.
And regarding passing the signal down - it's how I wrote it - it does that exactly.

asturur:
i don't see `set('src', value)` anywhere defaulting to setSrc, where that happens?

ShaMan123:
> `img.set('src',...)`

It doesn't which IMO should be viewed as a bug

asturur:
indeed src is not an image of the property, setSrc is like, a, change the element. It does not even update it, it totally changes it ( and that is fine because the previous one could be a canvas element ).
I m not even sure if setSrc is a good idea in general, it looks like a wrapper for setElement with a loader, but on top of that then changes the image size, while setElement does't.
No one of them is resetting cropX and cropY, creating potentially weird visual bugs when cropX is bigger than the new width and similar for cropY.

ShaMan123:
Good point
So perhaps `src` should only exist on from/to object

I now understand your point

asturur:
> > `img.set('src',...)`
> 
> It doesn't which IMO should be viewed as a bug

probably the bug is the name of the method. Probably that is setElementFromSrc
fabric.Image does not have a src, it has a drawable element, we have utils to load new drawable elements ( like fabric.util.loadImage ) and we can improve how they are swapped the element.

ShaMan123:
Yeah we should rename and add logic
I think we should rename Image as well because it can be Video as well

.github/workflows/node-unit-tests.yml
ShaMan123: I'm leaving tests on v16 because they use AbortController

ShaMan123 (commented):
READY!

asturur (approved):

PR #7825, opened 2022-03-22
feat(ci): better cli

ShaMan123:
This PR aims to make local dev of fabric easier for the new comer and the seasoned maintainer. So the community can take part easily.
Test suite is quite horrible in terms of DX, this should improve it drastically.
1. Refactored scripts/cmds to be os agnostic (some didn't work on windows). They should all work, please confirm on MacOS
2. Added a debug flag for visual tests to recreate a golden file for the dev to compare changes (this means we need to relay of a [PR I made to qunit](https://github.com/qunitjs/qunit/pull/1681) or simply use my branch)
3. Added a `start` script that handles what's needed to dev via fabricjs.com
4. Consolidated all scripts under one executable

### In Action

https://user-images.githubusercontent.com/34343793/160217304-c2e225d8-7a80-4044-9042-818716d23017.mp4

https://user-images.githubusercontent.com/34343793/160217309-43f0d649-0279-4a48-8047-505a85e8ff74.mp4

PR #7833, opened 2022-03-26
fix(ci): CLI `debug` and `recreate` options

ShaMan123:
Finalizes #7825 

> Excellent.
> The only think I'm thinking of is this:
> > Added a debug flag for visual tests to recreate a golden file for the dev to compare changes (this means we need to relay of a https://github.com/qunitjs/qunit/pull/1681 or simply use my branch)

> I can simply add a env var instead of waiting for quint to allow other flags

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/7825#issuecomment-1079698430_

**DONE**


ShaMan123 (commented):
ready

asturur (approved):

asturur:
there was a little bug i noticed using the new ci, i think that sometimes the full suite does not start if i select it from the menu.
But overall is a great improvement.

PR #7839, opened 2022-03-29
chore(v6): TypeScript proto

ShaMan123:
Introduce a transformer that is able to convert most of the files under `src` into es6/ts and produces a good diff on git when preserving the file names (changing the extension is OK by git).

### Relevant changes
- `scripts/transform_file.js` - the transformer logic
- `scripts/index.js` - expose cmd for cli (`npm run dev transform -- --help`)
- `src` files were committed to show the diff - they shouldn't be accepted into master yet
- `build.js` - 3e8cf5f7ed8a43c0f8a88a87ae90c9d5558964cb reverted the following:
   The build file can do the following:
   1. compile ts/js files from `src` to `dist`
   2. build `fabric.js` from compiled files.
- `src/util/anim_ease.ts` is a manual converted file

build needs to be replaced and babel introduced


package.json
```diff
@@ -82,11 +83,11 @@
     "pixelmatch": "^4.0.2",
     "qunit": "^2.17.2",
     "testem": "^3.2.0",
-    "uglify-js": "^3.3.28"
+    "typescript": "^4.5.4",
+    "uglify-js": "^3.15.3"
```
asturur, line 87: those this work with es6?

asturur:
@ShaMan123 as soon as the group pr is merged, we can start this.
My plan is:
first pr: 
- convert 1 file that has only functions ( easing file is a good example )
- get a new eslint config that support ts syntax and tslint config
- choose the basic transpile target ( cant' be es5 )
- choose the new bundler / packager ( because the old build script won't work since wrapping function won't be there and so it will pollute the global space, or modify the current one to work.
- no classes, no mixins nothing

then we rotate a class per pr, possibly let's involve whoever wants to convert, so that everyone can get dirty hands and get comfortable with the process. I want to do at least one.

package.json
ShaMan123: typescript for sure, uglify I think so

ShaMan123:
sounds good to me

ShaMan123:
should use tsc --watch option as it is incremental

ShaMan123 (commented):
@asturur @melchiar  take a look! I managed to build a transformer that works pretty good. I'm really happy about it (accept for the horrible time sink)! 
And I saw something very exciting in VSCODE:

https://user-images.githubusercontent.com/34343793/173205378-7bb9f23a-10a9-4476-84ee-f6233d776159.mp4


Then human work can be much easier:
- static properties - too hard to impl in transformer
- imports/export/build/
- sub/super class
- things missed out by the transformer (see 
[dump.txt](https://github.com/fabricjs/fabric.js/files/8884768/transformer.txt))

 
**OHHH**
Forgot this:
Run `npm run dev transform`

ShaMan123:
Transforming PRs is now possible!
I have tested the following:
- mock merging a PR into master (merged `deprecate-extend` into `test-pr-transform`)
- ran `npm run dev transform -- -ts -o --diff`
- committed changes 0b6da98f
- merged branch into mock transformed master 656bc9d8 (merged `test-pr-transform` into `ts-proto`)
- resolved conflicts

The conflicts were simple and weren't affected by the conversion.

ShaMan123 (commented):
The transformer is in a good state
We can revert the changes to `src` and `dist` and merge this PR as it only changes scripts

ShaMan123:
updated description

ShaMan123:
closing in favor of #8020

PR #7841, opened 2022-03-30
chore(ts-prep): refactor: move `_setPositionDimensions` path method

ShaMan123:
Preparing for typescript.
This methods belongs on path.

### DIFF
--- a/src/shapes/path.class.js
+++ b/src/shapes/path.class.js
@@ -64,7 +64,7 @@
         Array.isArray(path) ? path : fabric.util.parsePath(path)
       );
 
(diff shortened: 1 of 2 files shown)


ShaMan123:
I thought polyline is a subclass of path but it isn't.
TS migration - we may want them both subclassing an abstract class

asturur:
tbh i think polygon, polyline, triangle rect should be just paths. 
Exception made for cirle maybe.
They could mantain some special setter/getter as example for the triangl setWidth and setHeight, and setting the points on a polygon would then modify the resuling path.

ShaMan123:
I agree

PR #7842, opened 2022-03-30
feat(Text): condensed styles structure

melchiar:
This condenses text styles into a better optimized structure at serialization, and expands the structure again at initialization. The new structure uses ranges rather than having a style object for each character, resulting in significantly smaller style strings when the same style is used for a range of characters.

Here's a comparison for a textbox with 110 characters and 2 unique styles.

**Current styles structure:**

`'{"0":{"0":{"fill":"red"},"1":{"fill":"red"},"2":{"fill":"red"},"3":{"fill":"red"},"4":{"fill":"red"},"5":{"fill":"red"},"6":{"fill":"red"},"7":{"fill":"red"},"8":{"fill":"red"},"9":{"fill":"red"},"10":{"fill":"red"},"11":{"fill":"red"},"12":{"fill":"red"},"13":{"fill":"red"},"14":{"fill":"red"},"15":{"fill":"red"},"16":{"fill":"red"},"17":{"fill":"red"},"18":{"fill":"red"},"19":{"fill":"red"},"20":{"fill":"red"},"21":{"fill":"red"},"22":{"fill":"red"},"23":{"fill":"red"},"24":{"fill":"red"},"25":{"fill":"red"},"26":{"fill":"red"},"27":{"fill":"red"},"28":{"fill":"red"},"29":{"fill":"red"},"30":{"fill":"red"},"31":{"fill":"red"},"32":{"fill":"red"},"33":{"fill":"red"},"34":{"fill":"red"},"35":{"fill":"red"},"36":{"fill":"red"},"37":{"fill":"red"},"38":{"fill":"red"},"39":{"fill":"red"},"40":{"fill":"red"},"41":{"fill":"red"},"42":{"fill":"red"},"43":{"fill":"red"},"44":{"fill":"red"},"45":{"fill":"red"},"46":{"fill":"red"},"47":{"fill":"red"},"48":{"fill":"red"},"49":{"fill":"red"},"50":{"fill":"red"},"51":{"fill":"red"},"52":{"fill":"red"},"54":{"fill":"blue"},"55":{"fill":"blue"},"56":{"fill":"blue"},"57":{"fill":"blue"},"58":{"fill":"blue"},"59":{"fill":"blue"},"60":{"fill":"blue"},"61":{"fill":"blue"},"62":{"fill":"blue"},"63":{"fill":"blue"},"64":{"fill":"blue"},"65":{"fill":"blue"},"66":{"fill":"blue"},"67":{"fill":"blue"},"68":{"fill":"blue"},"69":{"fill":"blue"},"70":{"fill":"blue"},"71":{"fill":"blue"},"72":{"fill":"blue"},"73":{"fill":"blue"},"74":{"fill":"blue"},"75":{"fill":"blue"},"76":{"fill":"blue"},"77":{"fill":"blue"},"78":{"fill":"blue"},"79":{"fill":"blue"},"80":{"fill":"blue"},"81":{"fill":"blue"},"82":{"fill":"blue"},"83":{"fill":"blue"},"84":{"fill":"blue"},"85":{"fill":"blue"},"86":{"fill":"blue"},"87":{"fill":"blue"},"88":{"fill":"blue"},"89":{"fill":"blue"},"90":{"fill":"blue"},"91":{"fill":"blue"},"92":{"fill":"blue"},"93":{"fill":"blue"},"94":{"fill":"blue"},"95":{"fill":"blue"},"96":{"fill":"blue"},"97":{"fill":"blue"},"98":{"fill":"blue"},"99":{"fill":"blue"},"100":{"fill":"blue"},"101":{"fill":"blue"},"102":{"fill":"blue"},"103":{"fill":"blue"},"104":{"fill":"blue"},"105":{"fill":"blue"},"106":{"fill":"blue"},"107":{"fill":"blue"},"108":{"fill":"blue"},"109":{"fill":"blue"}}}'`

**Condensed styles structure**

`'[{"start":0,"end":53,"style":{"fill":"red"}},{"start":54,"end":110,"style":{"fill":"blue"}}]'`


melchiar:
@asturur, @ShaMan123 I still need to fix the tests, but if you could take a look a look and let me know your thoughts on this approach that'd be appreciated.

ShaMan123:
@melchiar why not migrate all logic to your new approach?
It seems better

melchiar:
> @melchiar why not migrate all logic to your new approach? It seems better

Ultimately it came down to complexity and performance. 

The current structure, while more verbose, is fast since it uses line numbers as keys so the style for each line can be accessed directly. The new structure uses an array so finding the style for a specific character means iterating through the style groups looking for that index. Since styles are in ranges, changing the style for a subset of characters means going through and modifying any overlapping groups. Every text change made would also require going through every style collection in the array to update the index numbers. 

By using this structure only for serialization we avoid these issues, and also avoid a major breaking change since this approach is backwards compatible.

Thoughts?

ShaMan123:
Sounds solid to me. Winning on all fronts. perf in runtime and storage size after serializing.

ShaMan123:
What do think about renaming? So it's clearer what the methods do.
Something like getExportedStyles/stylesToObject/serializeStyles and their counterparts..

melchiar:
> What do think about renaming? So it's clearer what the methods do. Something like getExportedStyles/stylesToObject/serializeStyles/inflateStyles and their counterparts..

Yeah, I had trouble with the names. I initialize went with compressStyles/decompressStyles but that's not really accurate since there is no compression. I think getCondensedStyles is pretty clear, but there isn't really a good antonym for condense so I just went with getExpandedStyles. Not perfect I agree. The methods only return the new styles though and don't actually assign them so I do think they should be named "get...something" (this was intentional so that condensed styles can easily be obtained without modifying the object).

ShaMan123:
My point is that the purpose of getting a condensed style is unclear and that's why I think it should be renamed so that a dev seeing it for the first time will have a feeling when to call it (preferably without reading the JSDOC description)

melchiar:
Fair point. Do you think getStylesObject/getStylesArray would work?

ShaMan123:
How about toStyleObject/fromStyleObject?

melchiar:
or maybe stylesToArray/stylesFromArray?

ShaMan123:
yeah, those are fine

src/shapes/text.class.js
```diff
@@ -373,6 +373,7 @@
         this.setPathInfo();
       }
       this.__skipDimension = false;
+      this.styles = this._stylesFromArray();
```
ShaMan123, line 376: I think it's best to pass the style array and not set it on the instance
It doesn't seem safe

src/mixins/text_style.mixin.js
```diff
@@ -319,6 +319,86 @@
...
+      return stylesArray;
+    },
+
+    /**
+     * Returns the object form of the styles property where styles are assigned per
+     * character rather than grouped by range. This format is more verbose, and is
+     * only used during runtime (not for serialization/storage)
+     * @return {Object}
+     */
+    _stylesFromArray: function() {
```
ShaMan123, line 375: I would pass styles array and text array as arg here as well.

src/mixins/text_style.mixin.js
```diff
@@ -319,6 +319,86 @@
...
+      if (!Array.isArray(this.styles)) {
+        return this.styles;
+      }
+      var textLines = this.text.split('\n'),
```
ShaMan123, line 379: Does this need to happen before text is parsed or can it happen right after? 
Saving some computation effort

src/mixins/text_style.mixin.js
```diff
@@ -319,6 +319,86 @@
...
+    /**
+     * Returns the array form of an object's styles property where styles are grouped in ranges
+     * rather than per character. This format is less verbose, and is better suited for storage
+     * so it is used in serialization (not during runtime).
+     * @return {Array}
+     */
+    _stylesToArray: function() {
+      // clone style structure to prevent mutation
+      var styles = fabric.util.object.clone(this.styles, true),
+          textLines = this.text.split('\n'),
```
ShaMan123, line 333: same here regarding computation
can you use `this._textLines` instead? and maybe pass this as an arg as well making these functions utils not relating to the instance?

src/mixins/text_style.mixin.js
```diff
@@ -319,6 +319,86 @@
...
+        for (var c = 0; c < textLines[i].length; c++) {
+          charIndex++;
+          //check if there's a style collection that includes the current character
+          if (this.styles[styleIndex]
+            && this.styles[styleIndex].start <= charIndex
+            && charIndex < this.styles[styleIndex].end) {
+            //create object for line index if it doesn't exist
+            stylesObject[i] = stylesObject[i] || {};
+            //add a style entry for this character's index
+            stylesObject[i][c] = this.styles[styleIndex].style;
```
ShaMan123, line 393: Isn't this a shared object?
I think you should add a test making sure it isn't (I think it will fail)
So use `Object.assign({}...`)

ShaMan123 (changes requested):
Seems good apart from a few minor changes (see comments) and a test that needs to be added

src/mixins/text_style.mixin.js
melchiar: Using `this_textLines` would work for the i-text and text classes, but textbox lines have autowrapping so the `_textLines.lines` property treats autowrapped lines as their own lines (also these autowrapped line divisions aren't serialized, so they're not yet available during initialization until after initDimensions is called)

But yes, these methods should work fine as utils, I just thought it made sense to attach them to the instance since _stylesToArray uses `_hasStyleChangedForSvg` which is also attached to the instance.

src/mixins/text_style.mixin.js
melchiar: Hmm...not sure I follow. Where are you suggesting `text.split('\n')` could go instead?

src/mixins/text_style.mixin.js
melchiar: right, nice catch!

melchiar:
@ShaMan123 @asturur I've moved the new style methods to `util/misc.js`, and also combined `styleHasChanged` and `styleHasChangedForSvg` into a single util since `styleHasChangedForSvg` is no longer just for svg.

Unless you spot any other issues I can port this over to master.

src/shapes/text.class.js
```diff
@@ -373,6 +373,7 @@
         this.setPathInfo();
       }
       this.__skipDimension = false;
+      this.styles = fabric.util.stylesFromArray(this.styles, this.text);
```
ShaMan123, line 376: As mention earlier I don't think it's safe to set styles before they are parsed (I would remove L368) and combine it here as follows.
I doesn't seem that logic uses it up to here
```suggestion
      this.styles = fabric.util.stylesFromArray(options ? (options.styles || { }) : { }, this.text);
```

ShaMan123 (changes requested):
Looks good.
One comment regarding safety and best practices - to avoid strange bugs.

src/shapes/text.class.js
melchiar: good call

src/util/misc.js
```diff
@@ -1215,5 +1215,111 @@
...
     },
+
+    /**
+     * @memberOf fabric.util
+     * @param {Object} prevStyle first style to compare
+     * @param {Object} thisStyle second style to compare
+     * @param {boolean} includeLineStyles whether to check overline, underline, and line-through properties
+     */
+    hasStyleChanged: function(prevStyle, thisStyle, includeLineStyles) {
+      var includeLineStyles = includeLineStyles || false;
```
ShaMan123, line 1226:
```suggestion
      includeLineStyles = includeLineStyles || false;
```

src/util/misc.js
```diff
@@ -1215,5 +1215,111 @@
...
+              prevStyle.fontStyle !== thisStyle.fontStyle ||
+              prevStyle.deltaY !== thisStyle.deltaY) ||
+              (includeLineStyles &&
+                (prevStyle.overline !== thisStyle.overline ||
+                prevStyle.underline !== thisStyle.underline ||
+                prevStyle.linethrough !== thisStyle.linethrough));
+    },
+
+    /**
+     * Returns the array form of an text object's inline styles property with styles grouped in ranges
```
ShaMan123, line 1242:
```suggestion
     * Returns the array form of a text object's inline styles property with styles grouped in ranges
```

src/util/misc.js
```diff
@@ -1215,5 +1215,111 @@
...
+     * rather than per character. This format is less verbose, and is better suited for storage
+     * so it is used in serialization (not during runtime).
+     * @memberOf fabric.util
+     * @param {fabric.Object} styles per character styles for a text object
+     * @param {String} text the text string that the styles are applied to
+     * @return {Array}
```
ShaMan123, line 1248: Perhaps better typed array?

ShaMan123, line 1248:
```suggestion
     * @return {{start: number, end: number, style: object}[]}
```

Or even better to define style but not crucial I guess

src/util/misc.js
```diff
@@ -1215,5 +1215,111 @@
...
+                prevStyle.underline !== thisStyle.underline ||
+                prevStyle.linethrough !== thisStyle.linethrough));
+    },
+
+    /**
+     * Returns the array form of an text object's inline styles property with styles grouped in ranges
+     * rather than per character. This format is less verbose, and is better suited for storage
+     * so it is used in serialization (not during runtime).
+     * @memberOf fabric.util
+     * @param {fabric.Object} styles per character styles for a text object
```
ShaMan123, line 1246:
```suggestion
     * @param {object} styles per character styles for a text object
```

src/util/misc.js
```diff
@@ -1215,5 +1215,111 @@
...
+            }
+          }
+          prevStyle = thisStyle || {};
+        }
+      }
+      return stylesArray;
+    },
+
+    /**
+     * Returns the object form of the styles property with styles are assigned per
```
ShaMan123, line 1288:
```suggestion
     * Returns the object form of the styles property with styles that are assigned per
```

ShaMan123 (changes requested):
My only last concern is about splitting text in this method.
I am strongly against.
1. perf reasons
2. What if a dev customizes the line breaks regex?

Can the style parsing use the instance's data instead of splitting by itself?
> Using `this_textLines` would work for the i-text and text classes, but textbox lines have autowrapping so the `_textLines.lines` property treats autowrapped lines as their own lines (also these autowrapped line divisions aren't serialized, so they're not yet available during initialization until after initDimensions is called)

> _Originally posted by @melchiar in https://github.com/fabricjs/fabric.js/pull/7842#discussion_r841292067_

Isn't there another property that can be used?
And if not maybe use `_reNewline`?
Maybe make the data available at `_splitTextIntoLines`?

Apart from it's great.

src/util/misc.js
melchiar: can you explain?

ShaMan123: This

https://github.com/fabricjs/fabric.js/pull/7842/files/06285f5384019ab6da58634f1accb8535d909291#r843663506

melchiar:
> Isn't there another property that can be used? And if not maybe use `_reNewline`? Maybe make the data available at `_splitTextIntoLines`?

Splitting by `_reNewline` would make sense since it seems linebreaks can sometimes appear as `\r\n`. Some more refactoring would be order though since `_reNewline` is a property of `fabric.Text`, while we've moved the new methods over to utils. We could just use the `lines` prop of the `_splitTextIntoLines` method in `fabric.Text` (the `_splitTextIntoLines` of `fabric.Textbox` wouldn't work here since it separates lines with autowrapping), though calling `_splitTextIntoLines` would slow performance since it does a whole bunch of unnecessary stuff compared to just using the `split()` approach. 

I'd be inclined to make `_reNewline` a property of the `fabric` namespace and then just call `split(fabric._reNewLine)` in `_stylesToArray`. This seems to make the most sense from a performance standpoint.

ShaMan123:
> We could just use the `lines` prop of the `_splitTextIntoLines` method in `fabric.Text` (the `_splitTextIntoLines` of `fabric.Textbox` wouldn't work here since it separates lines with autowrapping), though calling `_splitTextIntoLines` would slow performance since it does a whole bunch of unnecessary stuff compared to just using the `split()` approach.

I don't understand why it would slow down things. 
It's called anyways `initialize > initDimensions > _splitText > _splitTextIntoLines` both in Text and Textbox.

https://github.com/fabricjs/fabric.js/blob/887a745b5943492478b12a79519f381a23758e4d/src/shapes/text.class.js#L433
https://github.com/fabricjs/fabric.js/blob/887a745b5943492478b12a79519f381a23758e4d/src/shapes/textbox.class.js#L97

ShaMan123:
and maybe `_generateStyleMap` is now redundant

melchiar:
> I don't understand why it would slow down things. It's called anyways `initialize > initDimensions > _splitText > _splitTextIntoLines` both in Text and Textbox.

We could move `_splitText()` from `initDimensions()` to `initialize()`, then pass the `textLines` property as an argument to the `stylesToArray` method. Doing that though would defeat the purpose of making `stylesToArray` a util since you'd still need a text instance to call `_splitText()` on to send to the util.

Also, the textbox class still presents problems since the `_splitTextIntoLines` method performs wrapping.

src/util/misc.js
```diff
@@ -1215,5 +1215,111 @@
...
       return new fabric.Group([a], { clipPath: b, inverted: inverted });
     },
+
+    /**
+     * @memberOf fabric.util
+     * @param {Object} prevStyle first style to compare
+     * @param {Object} thisStyle second style to compare
+     * @param {boolean} includeLineStyles whether to check overline, underline, and line-through properties
+     */
+    hasStyleChanged: function(prevStyle, thisStyle, includeLineStyles) {
```
asturur, line 1225: let's call `includeLineStyles` something like forSvg or forTextSpans.
lineStyles is specific and does not cover what the difference was there for.

src/shapes/text.class.js
```diff
@@ -365,14 +365,14 @@
...
     initialize: function(text, options) {
-      this.styles = options ? (options.styles || { }) : { };
       this.text = text;
       this.__skipDimension = true;
       this.callSuper('initialize', options);
       if (this.path) {
         this.setPathInfo();
       }
       this.__skipDimension = false;
+      this.styles = fabric.util.stylesFromArray(options ? (options.styles || { }) : { }, this.text);
```
asturur, line 375: i think this conversion should go in the `fromObject` function more than initialize.
Reason: if the runtime style format for now stays the same, we want to still be able to initialize a new text copying the style from another.
If we go though the clone function nothing will change since we use toObject and fromObject.
In general since this is an object dump change, i would like to handle into the toObject/fromObject pair

asturur:
regargin the splitting text in the serialization, yes apart customization we need to check how we split lines in general.
If i remember correctly textbox splits differently but then has a map to go back to normal textlines in itext mode.
That would mean that how itext breaks line is always good, maybe we can just call that method instead of using the split.

Also maybe is possible to change the runtime format in something that is a single long array instead of an array of array.
That would make inserting and deleting easier, and we can track which is the start index of each line easily,

src/shapes/text.class.js
melchiar: done.

melchiar:
> regargin the splitting text in the serialization, yes apart customization we need to check how we split lines in general. If i remember correctly textbox splits differently but then has a map to go back to normal textlines in itext mode. That would mean that how itext breaks line is always good, maybe we can just call that method instead of using the split.

@asturur The would work, though I've currently got the stylesFromArray/stylesToArray methods as utils rather than attached to an instance. Would you suggest we attach those methods back to the text class so that the _splitText method can be called on the instance?

ShaMan123:
If it's meant to be used in `fromObject` it can't be attached to instance.

melchiar:
@asturur Just wanted to follow up on this. As @ShaMan123 has pointed out, the new methods can't be called from fromObject if they're attached to an instance. Doesn't keeping to the current method make the most sense?

ShaMan123:
I agree with @asturur. However because text splitting occurs after initialization I agree with you as well.

melchiar:
@asturur do you have a minute to take a look at this?

asturur:
I think this is fine, and also that we should advertise this change as much as possible and see if a boolean for compatibility is necessary for someone that may have been invested in the style analisys in the serialized objects.

Also as soon as we add the object spread operator let's remove that full clone so that we can clone the actual objects we make copy of in the toArray util.

PR #7844, opened 2022-03-30
feat(ci): CLI logging options

ShaMan123:
added 2 options:
1. qunit `filter` option
2. disabled logging passing tests by default, `show` option overrides this

cli


ShaMan123 (commented):
The package-lock is node16+
So if you don't want to commit it you'll need to push a revised one

PR #7850, opened 2022-04-01
fix(Polyline): safegurad `_setPositionDimensions`

ShaMan123:
before
```js
poly._setPositionDimensions() //throws error
poly._setPositionDimensions({}) //ok
```

now both are ok

### DIFF
--- a/src/shapes/polyline.class.js
+++ b/src/shapes/polyline.class.js
@@ -83,6 +83,7 @@
     },
 
     _setPositionDimensions: function(options) {
+      options || (options = {});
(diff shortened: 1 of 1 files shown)


melchiar (approved):
for sure, this is a no brainer.

asturur (approved):

asturur:
for things like this with another approval you don't have to wait me, you should be able to merge

asturur (approved):

ShaMan123:
> for things like this with another approval you don't have to wait me, you should be able to merge

I wasn't here

PR #7859, opened 2022-04-04
feat(Group): 2nd Patch of New Group! 🎉

ShaMan123:
This PR makes nested selection work out of the box.

Group's selection observers are now almost completely redundant (d8f8251d0b18d8d1c95805e8e5a4122712ba3020) if I am not mistaken. The only use I can think of is when active object is not active selection. **Consider** using `canvas.getActiveObjects().includes(obj)` instead.

Logic has been adapted and now active selection almost completely supports selecting nesting objects (multiple selection is handled by #7882 ) I believe the only thing from perfect impl is `containsPoint`

## To Do - done on #7900 
- [x] `canEnter` - https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846827347
    - [x] naming
    - [x] throwing or what? https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846754553 Errors will raise in current code after warning is logged. Meaning it shouldn't warn, severity is error.
- [x] Group renderBackground - migrate to https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846930838
- [x] collection add/insertAt/remove return values
- [ ] minor dead code in `text.class` https://github.com/fabricjs/fabric.js/pull/7859#issuecomment-1088259426 - this can be removed only if group set coords on it's object. The question is what would be less damaging to perf... leaving as is or setting coords on all nested objects? Or a third option would be for group to check if object is on screen before it renders it
- [ ] `containsPoint` should be calculated relative to canvas coordinate plane
- [ ] remove object monitors?

@asturur 
- [ ] agree on `isInFrontOf`
- [ ] verify edge case with this._objectsToRender invalidation
- [ ] find a different solution for _getTransformedDimensions, @asturur improve my understanding of the group layout needs for it

### Changed Files

- `src/brushes/spray_brush.class.js` - adapt to new group
- `src/canvas.class.js` - return top most target including sub targets in case target is interactive
- `src/mixins/animation.mixin.js` - fix animations to be relative to canvas
- `src/mixins/canvas_events.mixin.js` - port missing logic from v6!
- `src/mixins/canvas_grouping.mixin.js` - port missing logic from v6! accounting for nested objects 
    - make sure not to create a circular call when adding an object to active selection
    - create active selection using isInFrontOf
- `src/mixins/eraser_brush.mixin.js` - adapt to new group (follow up will be in a separate PR)
- `src/mixins/itext_behavior.mixin.js` - add/remove handlers from canvas events (not group)
- `src/mixins/itext_click_behavior.mixin.js` - enable text editing under group
- `src/mixins/object_ancestry.mixin.js` - port logic from v6!
- `src/mixins/object_geometry.mixin.js` - port logic from v6!
    - account for nested objects (getCoords+ _getTransformedDimensions)
    - expose methods on object that interact with canvas coordinate plane
- `src/mixins/object_stacking.mixin.js` - exposed `isInFrontOf` to determine which object is in front of the other in the entire tree
- `src/shapes/active_selection.class.js` - adapt to restore objects to their owning group
- `src/shapes/group.class.js` - tidy code for easier subclassing + support `backgroundColor`
- `src/static_canvas.class.js` 
   - pass target for added/removed events (used by IText)
   - properly unset canvas to the entire object tree and not only to top most object
   - center object in canvas plane even if it is nested
- `test/unit/group.js` - fix failing test
- `test/unit/itext_click_behaviour.js` - adapt tests for nested itext
- `test/unit/object.js`
   - test canvas added/removed event object
   - `isDescendantOf`
   - `findCommonAncestors`
   - `isInFrontOf`
- `test/visual/group_layout.js` - reactivate tests

### Things Left Out
- `Layer`  - created a dedicated PR for it #7860
- `src/mixins/canvas_grouping` - I suspect it needs additional work. Committed some fixes and dumped the file from v6!.
- `Group#_renderObjects` - we need to discuss this, opened dedicated PR #7863 
- Caching - discussed in #7874

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -892,7 +892,7 @@
      */
     searchPossibleTargets: function (objects, pointer) {
       var target = this._searchPossibleTargets(objects, pointer);
-      return target;
+      return /*this.targets[0] ||*/ target;
     },
 
     /**
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -892,7 +892,7 @@
      */
     searchPossibleTargets: function (objects, pointer) {
       var target = this._searchPossibleTargets(objects, pointer);
-        return target && target.interactive && this.targets[0] ? this.targets[0] : target;
+      return target && target.interactive && this.targets[0] ? this.targets[0] : target;
     },
 
     /**
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -892,7 +892,7 @@
      */
     searchPossibleTargets: function (objects, pointer) {
       var target = this._searchPossibleTargets(objects, pointer);
-      return this.targets[0] || target;
+        return target && target.interactive && this.targets[0] ? this.targets[0] : target;
     },
 
     /**
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -441,7 +441,7 @@
         }
         objsToRender.push.apply(objsToRender, activeGroupObjects);
       }
-      //  in case a single object is selected render it's entire above the other objects
+      //  in case a single object is selected render it's entire parent above the other objects
       else if (!this.preserveObjectStacking && activeObjects.length === 1) {
         var target = activeObjects[0], ancestors = target.getAncestors(true);
         var topAncestor = ancestors.length === 0 ? target : ancestors.pop();
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -892,7 +892,7 @@
      */
     searchPossibleTargets: function (objects, pointer) {
       var target = this._searchPossibleTargets(objects, pointer);
-      return /*this.targets[0] ||*/ target;
+      return this.targets[0] || target;
     },
 
     /**
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -712,10 +712,13 @@
           }
         }
       }
-      this._objectsToRender = this._chooseObjectsToRender();
+      var invalidate = shouldRender || shouldGroup;
+      //  we clear `_objectsToRender` in case of a change in order to repopulate it at rendering
+      //  run before firing the `down` event to give the dev a chance to populate it themselves
+      invalidate && (this._objectsToRender = undefined);
       this._handleEvent(e, 'down');
       // we must renderAll so that we update the visuals
-      (shouldRender || shouldGroup) && this.requestRenderAll();
+      invalidate && this.requestRenderAll();
     },
 
     /**
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -712,6 +712,7 @@
           }
         }
       }
+      this._objectsToRender = this._chooseObjectsToRender();
       this._handleEvent(e, 'down');
       // we must renderAll so that we update the visuals
       (shouldRender || shouldGroup) && this.requestRenderAll();
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -904,7 +904,9 @@
           target = transform.target,
           //  transform pointer to target's containing coordinate plane
           //  both pointer and object should agree on every point
-          localPointer = fabric.util.sendPointToPlane(pointer, null, target.group);
+          localPointer = target.group ?
+            fabric.util.sendPointToPlane(pointer, null, target.group.calcTransformMatrix()) :
+            pointer;
 
       transform.reset = false;
       transform.shiftKey = e.shiftKey;
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -905,7 +905,7 @@
           //  transform pointer to target's containing coordinate plane
           //  both pointer and object should agree on every point
           localPointer = fabric.util.sendPointToPlane(pointer, null, target.group);
-      
+
       transform.reset = false;
       transform.shiftKey = e.shiftKey;
       transform.altKey = e[this.centeredKey];
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -900,13 +900,17 @@
      */
     _transformObject: function(e) {
       var pointer = this.getPointer(e),
-          transform = this._currentTransform;
-
+          transform = this._currentTransform,
+          target = transform.target,
+          //  transform pointer to target's containing coordinate plane
+          //  both pointer and object should agree on every point
+          localPointer = fabric.util.sendPointToPlane(pointer, null, target.group);
+      
       transform.reset = false;
       transform.shiftKey = e.shiftKey;
       transform.altKey = e[this.centeredKey];
 
-      this._performTransformAction(e, transform, pointer);
+      this._performTransformAction(e, transform, localPointer);
       transform.actionPerformed && this.requestRenderAll();
     },
 
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -54,7 +54,7 @@
     _updateActiveSelection: function(target, e) {
       var activeSelection = this._activeObject,
           currentActiveObjects = activeSelection._objects.slice(0);
-      if (activeSelection.contains(target)) {
+      if (target.group === activeSelection) {
         activeSelection.remove(target);
         this._hoveredTarget = target;
         this._hoveredTargets = this.targets.concat();
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -13,8 +13,16 @@
      */
     _shouldGroup: function(e, target) {
       var activeObject = this._activeObject;
+      // first 2 lines are the pre-requisites for selecting.
+      // !!activeObject an activeObjectExists and the user is pressing the selectionKey
+      // on top of that the user also has to hit a target that is selectable.
+      // this.selection is the canvas switch to support multi selection.
       return !!activeObject && this._isSelectionKeyPressed(e)
         && !!target && target.selectable && this.selection
+        // if all pre-requisite pass, the target is either something different from the current
+        // activeObject or if an activeSelection already exists
+        // TODO at time of writing why `activeObject.type === 'activeSelection'` matter is unclear.
+        // is a very old condition uncertain if still valid.
         && (activeObject !== target || activeObject.type === 'activeSelection')
         //  make sure `activeObject` and `target` aren't ancestors of each other
         && !target.isDescendantOf(activeObject) && !activeObject.isDescendantOf(target)
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -13,8 +13,8 @@
      */
     _shouldGroup: function(e, target) {
       var activeObject = this._activeObject;
-      return activeObject && this._isSelectionKeyPressed(e)
-        && target && target.selectable && this.selection
+      return !!activeObject && this._isSelectionKeyPressed(e)
+        && !!target && target.selectable && this.selection
         && (activeObject !== target || activeObject.type === 'activeSelection')
         //  make sure `activeObject` and `target` aren't ancestors of each other
         && !target.isDescendantOf(activeObject) && !activeObject.isDescendantOf(target)
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -13,8 +13,12 @@
      */
     _shouldGroup: function(e, target) {
       var activeObject = this._activeObject;
-      return activeObject && this._isSelectionKeyPressed(e) && target && target.selectable && this.selection &&
-            (activeObject !== target || activeObject.type === 'activeSelection') && !target.onSelect({ e: e });
+      return activeObject && this._isSelectionKeyPressed(e)
+        && target && target.selectable && this.selection
+        && (activeObject !== target || activeObject.type === 'activeSelection')
+        //  make sure `activeObject` and `target` aren't ancestors of each other
+        && !target.isDescendantOf(activeObject) && !activeObject.isDescendantOf(target)
+        && !target.onSelect({ e: e });
     },
 
     /**
@@ -80,17 +84,19 @@
       this._fireSelectionEvents(currentActives, e);
     },
 
+
     /**
      * @private
      * @param {Object} target
+     * @returns {fabric.ActiveSelection}
      */
     _createGroup: function(target) {
-      var objects = this._objects,
-          isActiveLower = objects.indexOf(this._activeObject) < objects.indexOf(target),
-          groupObjects = isActiveLower
-            ? [this._activeObject, target]
-            : [target, this._activeObject];
-      this._activeObject.isEditing && this._activeObject.exitEditing();
+      var activeObject = this._activeObject;
+      var groupObjects = target.isInFrontOf(activeObject) ?
+        [activeObject, target] :
+        [target, activeObject];
+      activeObject.isEditing && activeObject.exitEditing();
+      //  handle case: target is nested
       return new fabric.ActiveSelection(groupObjects, {
         canvas: this
       });
--- a/src/mixins/canvas_grouping.mixin.js
+++ b/src/mixins/canvas_grouping.mixin.js
@@ -13,12 +13,10 @@
      */
     _shouldGroup: function(e, target) {
       var activeObject = this._activeObject;
-      // first 2 lines are the pre-requisites for selecting.
-      // !!activeObject an activeObjectExists and the user is pressing the selectionKey
-      // on top of that the user also has to hit a target that is selectable.
-      // this.selection is the canvas switch to support multi selection.
-      return !!activeObject && this._isSelectionKeyPressed(e)
-        && !!target && target.selectable && this.selection
+      // check if an active object exists on canvas and if the user is pressing the `selectionKey` while canvas supports multi selection.
+      return !!activeObject && this._isSelectionKeyPressed(e) && this.selection
+        // on top of that the user also has to hit a target that is selectable.
+        && !!target && target.selectable
         // if all pre-requisite pass, the target is either something different from the current
         // activeObject or if an activeSelection already exists
         // TODO at time of writing why `activeObject.type === 'activeSelection'` matter is unclear.
@@ -26,6 +24,7 @@
         && (activeObject !== target || activeObject.type === 'activeSelection')
         //  make sure `activeObject` and `target` aren't ancestors of each other
         && !target.isDescendantOf(activeObject) && !activeObject.isDescendantOf(target)
+        //  target accepts selection
         && !target.onSelect({ e: e });
     },
 
--- a/src/mixins/itext_behavior.mixin.js
+++ b/src/mixins/itext_behavior.mixin.js
@@ -25,8 +25,9 @@
      */
     initAddedHandler: function() {
       var _this = this;
-      this.on('added', function() {
-        var canvas = _this.canvas;
+      this.on('added', function (opt) {
+        //  make sure we listen to the canvas added event
+        var canvas = opt.target;
         if (canvas) {
           if (!canvas._hasITextHandlers) {
             canvas._hasITextHandlers = true;
@@ -40,8 +41,9 @@
 
     initRemovedHandler: function() {
       var _this = this;
-      this.on('removed', function() {
-        var canvas = _this.canvas;
+      this.on('removed', function (opt) {
+        //  make sure we listen to the canvas removed event
+        var canvas = opt.target;
         if (canvas) {
           canvas._iTextInstances = canvas._iTextInstances || [];
           fabric.util.removeFromArray(canvas._iTextInstances, _this);
--- a/src/mixins/itext_click_behavior.mixin.js
+++ b/src/mixins/itext_click_behavior.mixin.js
@@ -152,7 +152,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    */
   mouseUpHandler: function(options) {
     this.__isMousedown = false;
-    if (!this.editable || this.group ||
+    if (!this.editable ||
       (options.transform && options.transform.actionPerformed) ||
       (options.e.button && options.e.button !== 1)) {
       return;
--- a/src/mixins/itext_click_behavior.mixin.js
+++ b/src/mixins/itext_click_behavior.mixin.js
@@ -153,6 +153,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
   mouseUpHandler: function(options) {
     this.__isMousedown = false;
     if (!this.editable ||
+      (this.group && !this.group.interactive) ||
       (options.transform && options.transform.actionPerformed) ||
       (options.e.button && options.e.button !== 1)) {
       return;
--- a/src/mixins/object_ancestry.mixin.js
+++ b/src/mixins/object_ancestry.mixin.js
@@ -39,7 +39,7 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
   /**
    *
    * @param {fabric.Object} other
-   * @returns {{ index: number, otherIndex: number, ancestors: fabric.Object[] }} ancestors may include the passed objects if one is an ancestor of the other resulting in index of -1
+   * @returns {{ index: number, otherIndex: number, ancestors: fabric.Object[] } | boolean} ancestors may include the passed objects if one is an ancestor of the other resulting in index of -1
    */
   findCommonAncestors: function (other) {
     if (this === other) {
--- a/src/mixins/object_ancestry.mixin.js
+++ b/src/mixins/object_ancestry.mixin.js
@@ -23,8 +23,10 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
 
   /**
    *
+   * @typedef {fabric.Object[] | [...fabric.Object[], fabric.StaticCanvas]} Ancestors
+   * 
    * @param {boolean} [strict] returns only ancestors that are objects (without canvas)
-   * @returns {(fabric.Object | fabric.StaticCanvas)[]} ancestors from bottom to top
+   * @returns {Ancestors} ancestors from bottom to top
    */
   getAncestors: function (strict) {
     var ancestors = [];
@@ -39,23 +41,28 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
   /**
    *
    * @param {fabric.Object} other
-   * @returns {{ index: number, otherIndex: number, ancestors: fabric.Object[] } | boolean} ancestors may include the passed objects if one is an ancestor of the other resulting in index of -1
+   * @param {boolean} [strict] finds only ancestors that are objects (without canvas)
+   * @returns {{ index: number, otherIndex: number, ancestors: Ancestors } | undefined} ancestors may include the passed objects if one is an ancestor of the other resulting in index of -1
    */
-  findCommonAncestors: function (other) {
+  findCommonAncestors: function (other, strict) {
     if (this === other) {
-      return true;
+      return {
+        index: 0,
+        otherIndex: 0,
+        ancestors: this.getAncestors(strict)
+      };
     }
     else if (!other) {
-      return false;
+      return undefined;
     }
-    var ancestors = this.getAncestors();
+    var ancestors = this.getAncestors(strict);
     ancestors.unshift(this);
-    var otherAncestors = other.getAncestors();
+    var otherAncestors = other.getAncestors(strict);
     otherAncestors.unshift(other);
     for (var i = 0, ancestor; i < ancestors.length; i++) {
       ancestor = ancestors[i];
       for (var j = 0; j < otherAncestors.length; j++) {
-        if (ancestor === otherAncestors[j] && !(ancestor instanceof fabric.StaticCanvas)) {
+        if (ancestor === otherAncestors[j]) {
           return {
             index: i - 1,
             otherIndex: j - 1,
@@ -69,9 +76,10 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
   /**
    *
    * @param {fabric.Object} other
+   * @param {boolean} [strict] checks only ancestors that are objects (without canvas)
    * @returns {boolean}
    */
-  hasCommonAncestors: function (other) {
-    return !!this.findCommonAncestors(other);
+  hasCommonAncestors: function (other, strict) {
+    return !!this.findCommonAncestors(other, strict);
   }
 });
--- a/src/mixins/object_ancestry.mixin.js
+++ b/src/mixins/object_ancestry.mixin.js
@@ -28,10 +28,10 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
    */
   getAncestors: function (strict) {
     var ancestors = [];
-    var parent = this.group || (!strict ? this.canvas : undefined);
+    var parent = this.group || (strict ? undefined : this.canvas);
     while (parent) {
       ancestors.push(parent);
-      parent = parent.group || (!strict ? parent.canvas : undefined);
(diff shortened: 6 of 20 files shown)


ShaMan123:
@asturur What is `!this.group` doing here?
https://github.com/fabricjs/fabric.js/blob/4eff9e270b2918964b3749ebcd4e55013b072842/src/shapes/text.class.js#L1525

src/canvas.class.js
```diff
@@ -892,7 +892,7 @@
      */
     searchPossibleTargets: function (objects, pointer) {
       var target = this._searchPossibleTargets(objects, pointer);
-      return target;
+      return target && target.interactive && this.targets[0] ? this.targets[0] : target;
```
asturur, line 895: note for myself: check what happens with nested groups

src/mixins/canvas_grouping.mixin.js
```diff
@@ -13,8 +13,12 @@
...
     _shouldGroup: function(e, target) {
       var activeObject = this._activeObject;
-      return activeObject && this._isSelectionKeyPressed(e) && target && target.selectable && this.selection &&
-            (activeObject !== target || activeObject.type === 'activeSelection') && !target.onSelect({ e: e });
+      return !!activeObject && this._isSelectionKeyPressed(e)
+        && !!target && target.selectable && this.selection
+        && (activeObject !== target || activeObject.type === 'activeSelection')
+        //  make sure `activeObject` and `target` aren't ancestors of each other
+        && !target.isDescendantOf(activeObject) && !activeObject.isDescendantOf(target)
+        && !target.onSelect({ e: e });
```
asturur, line 21: ok this boolean condition was unaceppable before and is now too. we need to write this differently or with a lot of comments. I want to remember what is going on here and write it down in a paragraph

src/mixins/itext_behavior.mixin.js
```diff
@@ -25,8 +25,9 @@
      */
     initAddedHandler: function() {
       var _this = this;
-      this.on('added', function() {
-        var canvas = _this.canvas;
+      this.on('added', function (opt) {
+        //  make sure we listen to the canvas added event
```
asturur, line 29: is this a comment of somthing to fix or is this comment explains what the line below is doing? ( same for the below instance )

src/shapes/group.class.js
```diff
@@ -224,23 +220,53 @@
...
+        return false;
+      }
+      else if (object.group && object.group === this) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Group: duplicate objects are not supported inside group, this call has no effect');
+        /* _DEV_MODE_END_ */
+        return false;
+      }
+      return true;
+    },
```
asturur, line 243: is this safeguarding from user mouse usage or developer misuse of the api?

src/shapes/group.class.js
```diff
@@ -385,6 +410,7 @@
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
     drawObject: function(ctx) {
+      this._renderBackground(ctx);
```
asturur, line 413: mmmmm i really don't want this. can the group be a container only?

src/mixins/canvas_grouping.mixin.js
asturur: this is a task for me.

asturur:
the changes here are not extremely extensive. thanks.

src/mixins/itext_behavior.mixin.js
ShaMan123: explanation

src/shapes/group.class.js
ShaMan123: Could you explain why you so against??
I really need/want it

src/shapes/group.class.js
ShaMan123: dev misuse

src/canvas.class.js
ShaMan123: If I understand it correctly it iterates over all objects from last to first, wee it hits an objects it walks down it's objects as well so I expect targets[0] to be the top most target to accept the pointer.

src/mixins/itext_behavior.mixin.js
asturur: well so i don't get it.
We are on the added event handler so we are listenting to it.
And we are putting the canvas in a variable from the add event rather than the _this context.
Why does it get that comment?

src/shapes/group.class.js
asturur: because a group should be a container, that handles layout. if you want a colored background you add a colored rect in this case.
But then....
can i have a gradient? can i have a picture?  can the background be a pattern?
And why there is no stroke?

This goes downhill fast.

How are you using it? that would clarify a lot.
In general background color for objects was a mistake imho.

ShaMan123:
I am concerned about `containsPoint`
I think I missed it when I migrated all methods from relative plane to canvas plane.
Could you take a look and tell me what you think?

src/mixins/itext_behavior.mixin.js
ShaMan123: Because now that group fires `added|removed` these callbacks got triggered once iText entered/exited group which is undesriable

src/shapes/group.class.js
ShaMan123: Well I have a solution.
The thing is I need to define an object that spreads over it's containing parent.
I can adapt the layout engine to check for objects with a flag that indicates that and it solves all issues I guess.
Thoughts?

ShaMan123: Same goes IMO for canvas' bg/overlay props

ShaMan123:
Another thing I want to discuss is `isCacheDirty`
Very damaging for performance.
Why do we need to check if objects are dirty when invalidating group? Especially when Object._set method does it for us?

asturur:
> Another thing I want to discuss is `isCacheDirty` Very damaging for performance. Why do we need to check if objects are dirty when invalidating group? Especially when Object._set method does it for us?
> 
> We should avoid iterations on `_objects`

I don't remember why we do that, but let's not make changes again to this branch to address performance issues.
With those large changes can come performance regressions.
In general large nested groups will be slow.

Eventually there has been something that required us inspect all objects because there could be something that invalidate the cache outside set
Example, a lef and top change in a group?

src/shapes/group.class.js
asturur: yes they are, more the overlay image than the background.
Especially also the option they have to ignore the zoom.

But said so, can you explain your use case for group background?

src/mixins/canvas_events.mixin.js
```diff
@@ -712,6 +712,7 @@
           }
         }
       }
+      this._objectsToRender = this._chooseObjectsToRender();
```
asturur, line 715: mmm Isn't this done before rendering? why should we do this on mouse down?

src/mixins/itext_behavior.mixin.js
asturur: maybe then added on canvas and added on group shouldn't be the same event. They are pretty much different.
Why did we add the added to group event?
A user of an app does not have a way to add things in a group without the developer writing some code . Why instead of using the event the developer can't add the desired code in his function?

Usually we add events only for things that are out of the developer control

src/mixins/itext_click_behavior.mixin.js
```diff
@@ -152,7 +152,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
    */
   mouseUpHandler: function(options) {
     this.__isMousedown = false;
-    if (!this.editable || this.group ||
+    if (!this.editable ||
```
asturur, line 155: should we keep here  `|| ( this.group && !this.group.interactive )` ?

src/mixins/object_ancestry.mixin.js
```diff
@@ -1,10 +1,31 @@
...
+        return false;
+      }
+      parent = parent.group || parent.canvas;
+    }
+    return false;
+  },
+
+  /**
+   *
+   * @param {boolean} [strict] returns only ancestors that are objects (without canvas)
```
asturur, line 26: i wonder how we will find ourselves with the following:
typescript and getAncestor returning an Object or a Canvas
the other ancestry methods not having the strict option.

src/mixins/object_geometry.mixin.js
```diff
@@ -630,9 +738,12 @@
         scaleY: this.scaleY,
         skewX: this.skewX,
         skewY: this.skewY,
+        width: this.width,
+        height: this.height,
+        strokeWidth: this.strokeWidth
```
asturur, line 748: this addition seems not used.
i also remember this was introduced to work around skewX and skewY difficulties nad is already an hack.
Why do we want to exented it to width/height/strokeWidth?

src/mixins/object_stacking.mixin.js
```diff
@@ -76,5 +76,48 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
...
+  },
+
+  /**
+   *
+   * @param {fabric.Object} other object to compare against
+   * @returns {boolean | undefined} if objects do not share a common ancestor or they are strictly equal it is impossible to determine which is in front of the other; in such cases the function returns `undefined`
+   */
+  isInFrontOf: function (other) {
+    if (this === other) {
+      return undefined;
```
asturur, line 88: warn and return false?

src/mixins/object_stacking.mixin.js
```diff
@@ -76,5 +76,48 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
       this.canvas.moveTo(this, index);
     }
     return this;
+  },
+
+  /**
+   *
+   * @param {fabric.Object} other object to compare against
+   * @returns {boolean | undefined} if objects do not share a common ancestor or they are strictly equal it is impossible to determine which is in front of the other; in such cases the function returns `undefined`
```
asturur, line 84: not sharing a common ancestor may only mean theyare on different canvases?
In that case can we print a log warn and return false rather than returning true/false/undefined?

src/mixins/object_stacking.mixin.js
```diff
@@ -76,5 +76,48 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
...
+   */
+  isInFrontOf: function (other) {
+    if (this === other) {
+      return undefined;
+    }
+    var ancestors = this.getAncestors().reverse().concat(this);
+    var otherAncestors = other.getAncestors().reverse().concat(other);
+    var i, j, found = false;
+    //  find the common ancestor
+    for (i = 0; i < ancestors.length; i++) {
```
asturur, line 94: reversed ancestors have as first element the canvas, always.
Isn't this loop finding, immediately a solution at the first loop in any case?

src/mixins/object_stacking.mixin.js
```diff
@@ -76,5 +76,48 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
...
+    for (i = 1; i < Math.min(tree.length, otherTree.length); i++) {
+      a = tree[i];
+      b = otherTree[i];
+      if (a !== b) {
+        parent = tree[i - 1];
+        return parent._objects.indexOf(a) > parent._objects.indexOf(b);
+      }
+    }
+    //  happens if a is ancestor of b or vice versa
+    return tree.length > otherTree.length;
```
asturur, line 121: seems to me that the first part of this function is `findCommonAncestors`.
Once a common ancestor is found, to me it seem that all that matter is position index of the next acenstor after teh common one. Who is front of who at this level will also dictacte who is i front later.
In fewer words i don't see the reason for the second loop.
I'll paste here in the comment a tenative replacement tomorrow morning ( now too tired )

src/mixins/canvas_events.mixin.js
ShaMan123: 3e091142

src/mixins/itext_click_behavior.mixin.js
ShaMan123: af3430ce

src/mixins/object_stacking.mixin.js
ShaMan123: I think that it should remain.
The result of checking which object is in front of the other is really undefined. false means it is not.

ShaMan123: > I think that it should remain. The result of checking which object is in front of the other is really undefined in this case. false means it is not.

It may also mean that the ancestors to belong to any canvas yet.

src/mixins/object_stacking.mixin.js
ShaMan123: > I think that it should remain. The result of checking which object is in front of the other is really undefined in this case. false means it is not.

src/mixins/object_geometry.mixin.js
ShaMan123: It is used by Group/Layer layout logic

src/mixins/object_ancestry.mixin.js
ShaMan123: checkout [typescript playground](https://www.typescriptlang.org/play?#code/FAQwRgzgLgTiDGUAE8A2IISQMXDAlvAPJgBWApokgN7BJLQhSFIzkgAmA9gHaoCeSKPwAO5AFwNY+HgHM6DclADCXLjA4QAFAEoawBQF9gx4GgxYASpWTkAHlHI9NOPIRIUqteo2bxW7Nx8gsJikgDkbIjhJgbmmEjK+DBo5Ej2js5YuGAExGQ2+j5QTCxsnLwCQqISSOHwyakxpvFYAApMABbpDk4uOXkehd5Spf7lQVWhteEiXc1x6AnKIDwAbhhFAc7kMACCqKi6kmtc+Bz6RrFmS1hDiHs88OTQMIIjskqPz9DqEAA8ABUepkXGA1Kh2DwAHxaV6EKAAfkkgJ0KJBfSwsAArmlEa5cu4CogANoAXSQkhJADpaQMiZ4oOSADSJVYbCAUkb0eC8aBIEBIAC8SB45AA7khrIhdABuBQ8vnIMDC0USxKNSFyhUoJUoVViyUdKCdbX0RU8fnwdmbEWGtnrDBm81sKDYmA8KR5ZD4kkgZlgZnwClU-2B+BBm2c2VIAD0sdFXCQ53YSHFnSY4Sw1uxEBksiEnXwWF2MHUV3onyg3xeUD+ACYgRiskhwVxIatYfDEMikKj0RlMUIYLikPj6flGeTKUgaXS3JObCyHRyuTreZbkIK7erpVBnRb+Sqd5KkiktTp5ebdZv9SekMbTZf13rrY6sPeVu+D3HYwABKAIAAWnwWQeHUcgdVdd1PW7H1ZzDIMQwQgMg0jd8ySvJBjFMe17mrJ5azeXRqSrGtfhgbQdGpEQuBEXRfwJQZiWQAAfFdNnY7EdgAMxkcgOGAPCWPI2B+BIsjCIo7QcXIajaPovR4yYhlCi43j+MEoT1Xw0TiOoySfjrSitFk+S6N0RFqQgJRVHUTQ5SAA)

src/mixins/itext_click_behavior.mixin.js
ShaMan123: af3430ce

src/mixins/object_ancestry.mixin.js
ShaMan123: hover on top of the results to see the types

src/mixins/object_stacking.mixin.js
ShaMan123: Added the `strict` option and tests
678f528f16d515cc353c6a6014e207da8f1f9769
3b8b393f740a2bfb81c9eb0737c1a8a335c76118

ShaMan123: The second loop walks down the common ancestors from the top most one until it finds a fork in the tree and then checks the index of each ancestor at the fork and returns.

src/mixins/canvas_events.mixin.js
asturur: Ok now i read the comment, in thise case then, shouldn't we do the same on mouse up?
there is an option to change the selection on mouseUp too rather than on mouse down.

asturur: Now i remember we are caching the objects to render, every time we change what is selected we should also invalidate that. Right? i ll see if i spot more places.

src/mixins/object_stacking.mixin.js
ShaMan123: No, see other comment

ShaMan123:
> > Another thing I want to discuss is `isCacheDirty` Very damaging for performance. Why do we need to check if objects are dirty when invalidating group? Especially when Object._set method does it for us?
> > We should avoid iterations on `_objects`
> 
> I don't remember why we do that, but let's not make changes again to this branch to address performance issues. With those large changes can come performance regressions. In general large nested groups will be slow.
> 
> Eventually there has been something that required us inspect all objects because there could be something that invalidate the cache outside set Example, a lef and top change in a group?

I think it's dead code. You said yourself that Group wasn't touched since 2014.
Investigation shows that you committed https://github.com/fabricjs/fabric.js/commit/700ee14ff5eb2297642747a5e424d49dc4df643d in 2016.
It seems to be logic prior to enforcing using `set` on `Object` so there wasn't a reliable way of knowing if a nested object was invalidated. This is not the case any longer.
Let's continue this conversation here #7874

src/mixins/object_stacking.mixin.js
asturur: i still don't get it.
findCommonAncestors seems to return the closest common ancestor.
If this function was:

```js
  isInFrontOf: function (other) {
    if (this === other) {
      return undefined;
    }
    var data = this.findCommonAncestors(other)
    if (!data) {
      return undefined;
    }
    parent = data.ancestors[0];
    // this contains other, is not in front of it
    if (parent === this) return false;
    // other contains this, so this is drawn on top of it
    if (parent === other) return true;
    return parent._objects.indexOf(this) > parent._objects.indexOf(other);
  }
```

What i don't get is why findCommonAncestors and this function need to work in a similar logic but reversed.
findCommonAncestors doesn't reverse the ancestors order and so doesn't need to walk the three for a fork, the common ancestor returned, the [0] of it, is the closest container of both objects.
There may be a hole in my logic, but i don't see it.

ShaMan123 (commented):
@asturur please review my comments.
The main issue we need to address is `containsPoint`. 
I have adapted `Collection#remove` in light of our previous discussions.

+ https://github.com/fabricjs/fabric.js/pull/7859#issuecomment-1088259426

src/mixins/canvas_events.mixin.js
ShaMan123: correct.
Mouse up might be. Isn't there some flag that makes selection run on mouseup? That's when we need to do it.

src/mixins/object_stacking.mixin.js
ShaMan123: Think of an asymmetric case.
- canvas
  - a
  - b
     - c
        - d

d.findCommonAncestors(a)

src/shapes/group.class.js
asturur: words! i have no idea what that drawing does and which is the background of the group and which is the group.
If if a secret, dump me an email, i just want to understand.

src/mixins/object_geometry.mixin.js
asturur: i see where this is used. Letting go for now, but i ll try my best to remove the need for this, right after this Pr is merged. I don't think this method should exist in this way.

src/mixins/object_stacking.mixin.js
asturur: My reason to think is not true, is that isInFrontOf is added to support preserveObjectStacking: false with nested selections. is not a generic utility supposed to do help and behave correctly.
We need to support a use case, and we get to the length to say, what if developers uses it for other context and need to know if is true/false/undefined.

What if is private and no one can use it? simpler?

asturur:
no please if you continue to add commits i have to review again and again, i will never get to the end of it. If we found a bug let's keep it and address it later. The pr can't be a moving target.

asturur:
(getting out a bit not being able to answer for a bit )

src/shapes/group.class.js
ShaMan123: canvas bg is the light blue color.
group bg is white.
You see objects (paths, rect, circle) that are part of groups.

ShaMan123:
Sorry. Those were the last changes for this PR. **LOCKED UNTIL MERGE**

src/shapes/group.class.js
```diff
@@ -224,23 +220,53 @@
...
+        console.warn('fabric.Group: trying to add group to itself, this call has no effect');
+        /* _DEV_MODE_END_ */
+        return false;
+      }
+      else if (object.group && object.group === this) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Group: duplicate objects are not supported inside group, this call has no effect');
+        /* _DEV_MODE_END_ */
+        return false;
+      }
```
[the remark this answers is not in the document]
ShaMan123: handled by
[3df9435](https://github.com/fabricjs/fabric.js/pull/7859/commits/3df9435e0861c0c16edf7d2c4b1f90c3eb451061)

src/mixins/object_stacking.mixin.js
ShaMan123: I think it should be exposed publicly. That's why I made it work for all cases.
It's useful and will come to play in many place IMO

ShaMan123: perf-wise it's supposed to be great

ShaMan123:
@asturur are you aware that github exposed a filter on commits in the diff view? It makes it easier to check against changes and once I push stuff they add a `view changes` button next to the commit log on the PR.
But anyways **LOCKDOWN** applied

src/mixins/canvas_grouping.mixin.js
```diff
@@ -13,8 +13,19 @@
      */
     _shouldGroup: function(e, target) {
       var activeObject = this._activeObject;
-      return activeObject && this._isSelectionKeyPressed(e) && target && target.selectable && this.selection &&
-            (activeObject !== target || activeObject.type === 'activeSelection') && !target.onSelect({ e: e });
+      // check if an active object exists on canvas and if the user is pressing the `selectionKey` while canvas supports multi selection.
+      return !!activeObject && this._isSelectionKeyPressed(e) && this.selection
```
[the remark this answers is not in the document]
asturur: this.selection should block you from creating multi selections.
It is true that it would be more logic to be constrained to the group selector only, since setting the selectionKey to something that does not exist would allow you to block shift+click and you would keep granularity in choices.
I remember i went over this topic more than once.
image

this is old old code, we can eventually change it, we can open an issue or discussion for it.
I'm not attached to it and does not change the complexity in a significant way.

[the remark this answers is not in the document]
ShaMan123: So we leave it as is for now, right?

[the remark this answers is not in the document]
asturur: yes that change needs to be a PR with that change only because is totally feasible is a beahviour change, has its own test changes, can't be mixed and is completely tangential to this PR

src/shapes/group.class.js
```diff
@@ -224,23 +220,53 @@
...
+    /**
+     * Checks if object can enter group and logs relevant warnings
+     * @private
+     * @param {fabric.Object} object
+     * @returns
+     */
+    canEnter: function (object) {
+      if (object === this || this.isDescendantOf(object)) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Group: trying to add group to itself, this call has no effect');
```
ShaMan123, line 232: This is wrong.
The call has an effect. At the time this is called `object` is already in the array.
This is bad and causes maximum stack called errors.
That's why I suggest:
1. Throwing
2. move `canEnter` to collection mixin but I fear it will damage perf (more iteration, but I'm not sure it's the case)

ShaMan123:
I'm out until the evening

src/mixins/object_stacking.mixin.js
asturur: Yes it can be useful, but if you are using it wrongly i m not helping you.
We are covering:
- Object not on any canvas
-  me checking if myself is in front of me
that are in my opinion things that we can warn the developer about but not writing logic for.

We don't need to solve it now, but i ll get back to it.

src/shapes/group.class.js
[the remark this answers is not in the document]
asturur: regarding [3df9435](https://github.com/fabricjs/fabric.js/pull/7859/commits/3df9435e0861c0c16edf7d2c4b1f90c3eb451061) 

First of all i m not sure if you thing the issue is cleare or not.
I see the issue for direct descendant, so for example.
c is a group included in B that is a group included in a, i try to check if `c.canLetEnter(a)`,  in that case running a.isDescendantOf(c) return false and we add it and we create a circular dependency in the._objects array.

Correct?

Can we check both?

Anyway performances cost outside the render loop do not matter in a meaningful way.

src/shapes/group.class.js
asturur: i don't know, let's keep it.
But there will be time in which i will tell you `I TOLD YOU`

asturur:
> I'm out until the evening

have fun.

asturur:
tasks:

## before merging:
i need to figure out why isInFrontOf cant't reuse findCommonAncestors. it has to.
this riddle: https://github.com/fabricjs/fabric.js/pull/7859/files#r846734964 didn't help me understand.

## after merging:
verify edge case with this._objectsToRender invalidation
solve the add/remove returned values inconsistency
find a different solution for _getTransformedDimensions, improve my understanding of the group layout needs for it

src/shapes/group.class.js
ShaMan123: Alright I give in.
I trust you nd your experience and I can sense the downhill.
But can we agree on supporting `layout: 'fill-parent'` instead in a dedicated PR?

src/mixins/object_stacking.mixin.js
ShaMan123: I didn't understand the last comment.
Take a closer look at the tests I wrote that cover this method.
I think that will explain best.

src/shapes/group.class.js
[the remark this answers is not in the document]
ShaMan123: > regarding [3df9435](https://github.com/fabricjs/fabric.js/pull/7859/commits/3df9435e0861c0c16edf7d2c4b1f90c3eb451061)
> 
> First of all i m not sure if you think the issue is cleared or not. I see the issue for non direct descendant, so for example. c is a group included in B that is a group included in a, i try to check if `a.canLetEnter(c)`, in that case running a.isDescendantOf(c) return false and we add it and we create a circular dependency in the._objects array.
> 
> Correct?

No because there are other forces at play. Part of the `enterGroup` logic removes c from b and only then truly accepts c into a.

The question is are there any more cases....
- if c is completely strange to a then it can enter
- if c is descendant of a before entering a then it will be removed from it's direct parent and will enter a
- if c is a direct descendant of a `canEnter` returns false.

Did I miss something? The rest of the possibilities aren't real. I mean these 3 conditions won't allow them to happen.


Regarding naming
`canLetEnter` is bad as well. What about `canEnterGroup` or `grantEntrance` or `isEnteringAllowed` or `assertEnterObject` (implying it throws)

ShaMan123:
> tasks:
> 
> ## before merging:
> i need to figure out why isInFrontOf can't reuse findCommonAncestors. it has to. this riddle: https://github.com/fabricjs/fabric.js/pull/7859/files#r846734964 didn't help me understand.

You made me laugh aloud. A damn riddle it is. Sorry!
Look at the tests. It should cover all edge cases.

> 
> ## after merging:
> add/remove returned values inconsistency 

I have changed collection.

src/mixins/object_stacking.mixin.js
ShaMan123: I will try to explain the riddle by countering your code.
```
d !== a1
data = {index: 2 ,otherIndex: 1, ancestors:[canvas]}
parent !== this
parent !== other
-1 > -1
```

ShaMan123: updated

src/shapes/group.class.js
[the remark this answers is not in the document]
ShaMan123: `allowEntrance`?

ShaMan123:
The discussion is getting hectic.
I will update the description with a todo list

src/shapes/group.class.js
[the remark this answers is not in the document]
ShaMan123: This is related as well https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846754553

ShaMan123:
updated description

src/shapes/group.class.js
ShaMan123: I have found an elegant solution!
Layer is configured to fill parent at layout.
The solution is then:
1. wrap the object you want to act as bg/overlay with Layer 
2. listen to the layer's `layout` event or override the `onLayout` method (depends when you hook onto it)
3. change bg/overlay size from within the event!

https://user-images.githubusercontent.com/34343793/162664147-1282f3aa-c06a-453f-88c2-6c6470c5f313.mp4

Tell me if I can dump backgroundColor support

ShaMan123: Once Layer is merged we can open a PR to dump bg/overlay and perhaps add a bit of code in `loadFromJSON` that supports loading them and parsing them into a layer for backward comptibility.
Thoughts?

ShaMan123: I love this.
Thanks for challenging me not to accept `backgroundColor | fill | stroke`

asturur:
the only thing to do is me to dig isFrontOf, i have seen there are a lot of tests done so i can probably just use those to test my assumptions.
I will open issues for the other follow ups.

For add and remove, i just want to change so that they behave same, they return what they need to return, but they return the same thing. ( both the length, or the list of added remove objects, or both true / false )

ShaMan123:
I can now offer a mature explanation as to where your logic is missing.
Testing `isInFrontOf` need the first fork of ancestors from the top and `findCommonAncestors()[0]` will return the parent holding the fork. Meaning it will not provide sufficient data for us to conclude anything about the fork. Only that it happens below this first common ancestor.

ShaMan123:
I am satisfied with the explanation. Are you?

asturur:
i spent 30 minutes changing the function as i see it fit, then stuck hours to make the test pass. Something makes the deepEquals fails for me, even when those are identical.

asturur:
I pushed up my branch `discuss-common-ancenstor` the idea is that i empowered findCommonAncestor to return useful data for solving isFronOf ( the rest of the ancestors fork rather than the indexes ). I m sure tomorrow fresh mind i can fix the tests. Now the test file is pretty much messed up.

ShaMan123:
opened #7900 for minors

src/shapes/group.class.js
ShaMan123: solved by #7901

ShaMan123:
@asturur I have encountered an inconsistency I would like to fix.
The question is who should conform to the other? Group or Canvas?

```js
let obj;
obj.on('removed', opt => {
   obj.canvas // still referenced
   obj.group // not referenced
  let parent = opt.target  // group or canvas
})
canvas.add(obj);
canvas.remove(obj);
group.add(obj);
group.remove(obj);
```

ShaMan123:
because of the naming `removed` I think we should dereference canvas.

asturur:
i should be able to merge this tomorrow morning.

ShaMan123:
Great. Do you plan on merging #7900 first or right after?

src/shapes/group.class.js
asturur: Didn't read all of this till now.
In my idea everything that is not the standard layout or the clipPath one will be reworked to be a plugin or a layout engine.
I'm not sure what a layout  is yet, but the core fabricJS should do the minimum possible. I ll get there on patch 3 or 4 i suppose

asturur (approved):
I ll follow up with add/remove/insertAt ( maybe insertAt can be removed? ) consistency chnages and i need to check that getTransformedDimensions overrides.

asturur:
Need to get a good run to the master branch, this is a first very large change also in behaviour

ShaMan123:
Excellent! And big step forward! Many thanks for your work Andrea.

PR #7860, opened 2022-04-04
feat(Layer): 3rd Patch of Group Rewrite ♦️

### LINKED ISSUE #3000: Upvote: Layers
At the Wiki there is "_Support for multiple layers (for performance)"_ on the Roadmap.

I would like to upvote this feature. 

It is not only helpfull for performance reasons, but can be very useful for any  kind of drawing application. Basically it offers a powerull way to organize the drawings, steers which objects are currently visible/selectable/changeable.

For those, who are not familiar with the concept of layers, here is a short introduction: 
https://www.youtube.com/watch?v=CuKNa4KBC8s
This video shows some powerful examples how layers can be used:
https://www.youtube.com/watch?v=tEzBAVPQUF0
(_Edit: I updated the video-links to better ones. This was the original link_: https://www.youtube.com/watch?v=mE82knjW7IA)

I thought a bit about the functionality of a layer feature in fabrics. If you are interested, we can discuss it here.

PS: I appreciate the faster release policy very much.

### LINKED ISSUE #7692: multiple canvas layer support?
hey, is it possible to create multiple canvas layer in fabric to speed up performance? 

```javascript
// eg
const canvas = new fabric.Canvas('canvas'); // now it will create a default canvas layer and topCanvas

const newLayer1 = new fabric.Layer(...)
                                   
const newLayer2 = new fabric.Layer(...)

const rect = new fabric.Rect({
  top : 100,
  left : 100,
  width : 60,
  height : 70,
  fill : 'red'
});

canvas.add(newLayer1)
canvas.add(newLayer2)
```

https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas

as we know , using multiple layer to render some complex scenes is a efficient way. so I wonder this feature is under ur plans or it is out of ur plans

ShaMan123:
# Layer

This is pretty straight forward.
Subclass of Group.
Adapted layout logic to fit Layer's needs
Added logic in Group to support nesting Layer. 
This is **AMAZING** because Layer can now be used to boost perf. We can split a canvas/group's objects into 2 sub collections, one for active objects and one for non active ones. We can then use Layer to render all the non actives and have it cached while setting group not to cache. 

## Gist
Layer will layout itself once it is added to a collection and by listening to it's parent layouts
- canvas - `resize` event (from the `setDimensions` method - should we listen to anything else? `MutationObserver` works both on web/node)
- group - `layout` event

## Dev

~~Merge the 2nd patch, update from master and then merge this one.~~

~~View the [diff](https://github.com/fabricjs/fabric.js/compare/v6-group-patch2...v6-layer)~~

closes #3000 
closes #7692

https://fx32zv.csb.app/

## Cutting Edge
Read the comments below to see what can be achieved using Layer:
- https://github.com/fabricjs/fabric.js/pull/7860#issuecomment-1091817465
- https://github.com/fabricjs/fabric.js/pull/7860#issuecomment-1094545074
- https://github.com/fabricjs/fabric.js/pull/7860#issuecomment-1094571580


ShaMan123:
Layer can now be nested under Group!!
This is superb [c16bdfb](https://github.com/fabricjs/fabric.js/pull/7860/commits/c16bdfb1a9d7ff1bfd87e41ac05ac9a0761958da)

ShaMan123:
> This is **AMAZING** because Layer can now be used to boost perf. We can split a canvas/group's objects into 2 sub collections, one for active objects and one for non active ones. We can then use Layer to render all the non actives and have it cached while setting group not to cache.


I am experimenting with this approach
See how promising it is, and this is without implementing render batching

https://user-images.githubusercontent.com/34343793/162223662-2221ebe0-f7d8-4a22-812c-0788ce3d109f.mp4

ShaMan123:
I have found an elegant solution for bg/overlay in group/canvas!
Layer is configured to fill parent at layout.
The solution is then:
1. wrap the object you want to act as bg/overlay with Layer 
2. listen to the layer's `layout` event or override the `onLayout` method (depends when you hook onto it - if you add an event listener before adding layer to group/canvas it should do)
3. change bg/overlay size from within the event!

This is the same for canvas so I vote for dumping bg/overlay from canvas

In the video below you can see the yellow bg with a green dashed stroke that is in fact a simple Rect.

https://user-images.githubusercontent.com/34343793/162664147-1282f3aa-c06a-453f-88c2-6c6470c5f313.mp4

Tell me if I can dump backgroundColor support

[Edit fabric_app (forked)](https://codesandbox.io/s/fabric-app-forked-xifpd7?fontsize=14&hidenavigation=1&module=%2Fsrc%2FLayer.tsx&theme=dark)

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846930838_

ShaMan123:
My next idea/suggestion is ground breaking as well.
Layer is kind of a virtual canvas. And with caching it's not even virtual.
It's much more flexible.
So when caching is refactored properly I think that Layer can replace the need for top canvas entirely.
Then brushes/selection/whatever goes up there can be split into dedicated layers (by the dev even), perf undamaged and new possibilities unlocked (e.g. having 2+ active brushes without clearing each other, #7175 is then obsolete cause you would simply add a clip path to layer that renders the brush).
@asturur You said top canvas needs refactoring, maybe this is the solution.
In terms of possibilities this is mind blowing. Very rich UI/UX can be achieved.

asturur:
I m not sure if i resolved the conflicts correctly, you should have a look @ShaMan123 .
I see there is new code in the group to handle the case in which you add a layer inside a group. Am i correct?
I don't want to complicate the group more than it is now, and i see the layer as a special case more than the norm.

What works and what doesn't if we add the layer and we don't modify the group?
If is not easy to explain i can just try by myself.

asturur:
> My next idea/suggestion is ground breaking. Layer is kind of a virtual canvas. And with caching it's not even virtual. It's much more flexible than canvas. So when caching is refactored properly I think that Layer can replace the need for top canvas entirely. Then brushes (#7880)/selection/text editing/whatever goes up there can be split into dedicated layers (assigned to by the dev even), perf undamaged and new possibilities unlocked (e.g. having 2+ active brushes without clearing each other, #7175 is then obsolete 'cause you would simply add a clip path to the layer that renders the brush). @asturur You said top canvas needs refactoring, maybe this is the solution. In terms of possibilities this is mind blowing. Very rich UI/UX can be achieved. And if I think of filtering/masking... **WOW 🤯**.

I m not sure i want to change that much.
I ll try to understand the layer better.

ShaMan123:
> I see there is new code in the group to handle the case in which you add a layer inside a group. Am i correct?

Yes

There is a bit more logic in group to support layout of Layer under group. That's more or less it. Layer is resized according to Group so it should not make Group resize. And Therefore, when an object under layer is modified it should trigger layout on group.

I want it in. It is very useful. Take into account #7901

ShaMan123:
> Layer can now be nested under Group!! This is superb [c16bdfb](https://github.com/fabricjs/fabric.js/pull/7860/commits/c16bdfb1a9d7ff1bfd87e41ac05ac9a0761958da)

These are the changes to logic to allow layer to nest under group. As you can see it is minor and #7901 utilizes the same approach.

ShaMan123:
I have done my best to view the changes. But I fear I have missed something out.
What is the best way to view the diff from before the merge compared to after?

ShaMan123:
https://github.com/fabricjs/fabric.js/pull/7860/commits/dc6d32b72f9990571982ead582ef461e750492aa seems alright.
If you can point out things you are not sure about that could help.

ShaMan123 (commented):
I have reviewed the Group changes again.
And I think we should/could merge them separately before Layer.
They are crucial fixes for other PRs as well.

ShaMan123 (commented):
MAJOR update from master
[7860.diff.txt](https://github.com/fabricjs/fabric.js/files/9569414/7860.diff.txt) from before

PR #7861, opened 2022-04-04
fix(v6): 4th PR of Group Rewrite 🎛️ nested controls 😜

ShaMan123:
This PR fixes control rendering.
There are a few unresolved comments in the [original PR](#7758), please take a look.

### Main Changes
[commit](https://github.com/fabricjs/fabric.js/pull/7861/commits/3632fd0df9bc23ccbf908bde7cd9114c8575f6cc)

This PR ports #7758 to master

## Visuals
[controls12.png](https://github.com/fabricjs/fabric.js/blob/5e286aa15c279172a2630343e7e18d6eac114fb9/test/visual/golden/controls12.png)
[controls13.png](https://github.com/fabricjs/fabric.js/blob/5e286aa15c279172a2630343e7e18d6eac114fb9/test/visual/golden/controls13.png)


src/mixins/collection.mixin.js
```diff
@@ -3,79 +3,69 @@
...
-    if (nonSplicing) {
-      objects[index] = object;
-    }
-    else {
-      objects.splice(index, 0, object);
+  insertAt: function (objects, index, callback) {
+    var args = [index, 0].concat(objects);
+    this._objects.splice.apply(this._objects, args);
+    if (callback) {
+      for (var i = 2, length = args.length; i < length; i++) {
```
melchiar, line 42:
```suggestion
      for (var i = 2; i < args.length; i++) {
```

src/mixins/collection.mixin.js
```diff
@@ -3,79 +3,69 @@
...
    */
-  remove: function() {
+  remove: function(objectsToRemove, callback) {
     var objects = this._objects,
         index, somethingRemoved = false;
 
-    for (var i = 0, length = arguments.length; i < length; i++) {
-      index = objects.indexOf(arguments[i]);
-
+    for (var i = 0, length = objectsToRemove.length; i < length; i++) {
```
melchiar, line 59:
```suggestion
    for (var i = 0; i < objectsToRemove.length; i++) {
```

src/mixins/object_geometry.mixin.js
```diff
@@ -529,18 +637,22 @@
 
     /**
      * calculate rotation matrix of an object
+     * @param {boolean} [absolute] true means angle is measured relative to canvas, false means angle is measured realtive to parent
```
melchiar, line 640:
```suggestion
     * @param {boolean} [absolute] true means angle is measured relative to canvas, false means angle is measured relative to parent
```

src/mixins/object_geometry.mixin.js
```diff
@@ -529,18 +637,22 @@
...
-      return util.calcRotateMatrix(this);
+    _calcRotateMatrix: function (absolute, accountForFlipping) {
+      var angle = absolute ? this.getTotalAngle() : this.angle;
+      accountForFlipping && this.flipX && (angle -= 180);
+      return util.calcRotateMatrix({ angle: angle });
     },
 
     /**
      * calculate the translation matrix for an object transform
+     * @param {boolean} [absolute] true means translation is relative to canvas, false means realtive to parent
```
melchiar, line 651:
```suggestion
     * @param {boolean} [absolute] true means translation is relative to canvas, false means relative to parent
```

src/mixins/collection.mixin.js
ShaMan123: better commit this change into #7859

src/mixins/collection.mixin.js
ShaMan123: better commit this change into #7859

src/mixins/object_stacking.mixin.js
```diff
@@ -76,5 +76,48 @@ fabric.util.object.extend(fabric.Object.prototype, /** @lends fabric.Object.prot
...
+   * @param {fabric.Object} other object to compare against
+   * @returns {boolean | undefined} if objects do not share a common ancestor or they are strictly equal it is impossible to determine which is in front of the other; in such cases the function returns `undefined`
+   */
+  isInFrontOf: function (other) {
+    if (this === other) {
+      return undefined;
+    }
+    var ancestors = this.getAncestors().reverse().concat(this);
+    var otherAncestors = other.getAncestors().reverse().concat(other);
+    var i, j, found = false;
```
melchiar, line 92:
```suggestion
    var ancestors = this.getAncestors().reverse().concat(this),
        otherAncestors = other.getAncestors().reverse().concat(other),
        i, j, found = false;
```

src/shapes/active_selection.class.js
```diff
@@ -126,13 +167,13 @@
...
-      if (typeof childrenOverride.hasControls === 'undefined') {
-        childrenOverride.hasControls = false;
-      }
-      childrenOverride.forActiveSelection = true;
+      var options = Object.assign(
+        { hasControls: false },
+        childrenOverride,
+        { forActiveSelection: true }
+      );
       for (var i = 0, len = this._objects.length; i < len; i++) {
```
melchiar, line 175:
```suggestion
      for (var i = 0; i < this._objects.length; i++) {
```

src/shapes/group.class.js
```diff
@@ -15,280 +21,344 @@
...
+     * List of properties to consider when checking if state
+     * of an object is changed (fabric.Object#hasStateChanged)
+     * as well as for history (undo/redo) purposes
+     * @type string[]
+     */
+    stateProperties: fabric.Object.prototype.stateProperties.concat('layout'),
+
+    /**
+     * Used to optimize performance
+     * set to `false` if you don't need caontained objects to be target of events
```
melchiar, line 61:
```suggestion
     * set to `false` if you don't need contained objects to be target of events
```

ShaMan123:
@melchiar the changes you re suggesting belong to #7859

src/shapes/object.class.js
```diff
@@ -1403,11 +1402,11 @@
...
       ctx.rotate(degreesToRadians(this.group ? options.angle : this.angle));
-      if (styleOverride.forActiveSelection || this.group) {
-        drawBorders && this.drawBordersInGroup(ctx, options, styleOverride);
+      if (drawBorders && (styleOverride.forActiveSelection || this.group)) {
+        this.drawBordersInGroup(ctx, options, styleOverride);
       }
-      else {
-        drawBorders && this.drawBorders(ctx, styleOverride);
+      else if (drawBorders) {
+        this.drawBorders(ctx, styleOverride);
```
melchiar, line 1409: can we put L1405 to L1409 inside a single `if (drawBorders)` block?

src/mixins/collection.mixin.js
```diff
@@ -3,79 +3,69 @@
...
    */
-  add: function () {
-    this._objects.push.apply(this._objects, arguments);
-    if (this._onObjectAdded) {
-      for (var i = 0, length = arguments.length; i < length; i++) {
-        this._onObjectAdded(arguments[i]);
+  add: function (objects, callback) {
+    var size = this._objects.push.apply(this._objects, objects);
+    if (callback) {
+      for (var i = 0, length = objects.length; i < length; i++) {
```
melchiar, line 23:
```suggestion
      for (var i = 0; i < objects.length; i++) {
```

src/shapes/group.class.js
```diff
@@ -328,18 +398,19 @@
...
+      return this.ownCaching || (!!this.group && this.group.isOnACache());
     },
 
     /**
      * Execute the drawing operation for an object on a specified context
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
     drawObject: function(ctx) {
+      this._renderBackground(ctx);
       for (var i = 0, len = this._objects.length; i < len; i++) {
```
melchiar, line 414:
```suggestion
      for (var i = 0; i < this._objects.length; i++) {
```

src/shapes/group.class.js
```diff
@@ -370,151 +441,463 @@
...
-      this._objects = [];
     },
 
     /**
-     * make a group an active selection, remove the group from canvas
-     * the group has to be on canvas for this to work.
-     * @return {fabric.ActiveSelection} thisArg
-     * @chainable
+     * initial layout logic:
+     * calculate bbox of objects (if necessary) and translate it according to options recieved from the constructor (left, top, width, height)
```
melchiar, line 491:
```suggestion
     * calculate bbox of objects (if necessary) and translate it according to options received from the constructor (left, top, width, height)
```

src/shapes/group.class.js
```diff
@@ -370,151 +441,463 @@
...
+            clipPathCenter = transformPoint(clipPathCenter, inv);
+          }
+          return {
+            centerX: clipPathCenter.x,
+            centerY: clipPathCenter.y,
+            width: clipPathSizeAfter.x,
+            height: clipPathSizeAfter.y,
+          };
+        }
+        else if (!clipPath.absolutePositioned) {
```
melchiar, line 615: This condition is checked twice (here and L600). Can we nest L600 to L615 in a `if (clipPath.absolutePositioned)` block and change L615 to just `else`?

melchiar:
> @melchiar the changes you are suggesting belong to #7859

ugh...just saw this now

ShaMan123:
Continue there, I'll patch up what you've suggested later on.
Regarding nesting if statements I don't really like it. You prefer it because of perf?

melchiar:
> Continue there, I'll patch up what you've suggested later on. Regarding nesting if statements I don't really like it. You prefer it because of perf?

I try to avoid multiple checks for the same conditions.

ShaMan123:
I dislike nesting logic, I prefer it flat as possible

melchiar:
fair enough. this is excellent by the way 👍

asturur:
Catching up with this.

src/mixins/object_geometry.mixin.js
```diff
@@ -643,18 +643,22 @@
...
     /**
      * calculate rotation matrix of an object
+     * @param {boolean} [absolute] true means angle is measured relative to canvas, false means angle is measured relative to parent
      * @return {Array} rotation matrix for the object
      */
-    _calcRotateMatrix: function() {
-      return util.calcRotateMatrix(this);
+    _calcRotateMatrix: function (absolute, accountForFlipping) {
+      var angle = absolute ? this.getTotalAngle() : this.angle;
+      accountForFlipping && this.flipX && (angle -= 180);
```
asturur, line 651: i wonder how nested flippings behave.

src/mixins/object_geometry.mixin.js
```diff
@@ -572,14 +572,14 @@
...
-          finalMatrix = multiplyMatrices(finalMatrix, [1 / vpt[0], 0, 0, 1 / vpt[3], 0, 0]),
-          dim = this._calculateCurrentDimensions(),
+      var vpt = this.getViewportTransform(),
+          tMatrix = this._calcTranslateMatrix(true),
+          rMatrix = this._calcRotateMatrix(true, !!this.group),
+          positionMatrix = multiplyMatrices(tMatrix, rMatrix),
+          startMatrix = multiplyMatrices(vpt, positionMatrix),
+          finalMatrix = multiplyMatrices(startMatrix, [1 / vpt[0], 0, 0, 1 / vpt[3], 0, 0]),
+          transformOptions = this.group ? fabric.util.qrDecompose(this.calcTransformMatrix()) : undefined,
+          dim = this._calculateCurrentDimensions(transformOptions),
```
asturur, line 582: since the changes we are doing to rotate matrix and translate matrix are used here only, i wouldn't extend the api.
Let's handle the difference in this function.
Something like:
```js
    var vpt = this.getViewportTransform(),
          center = this.getCenterPoint(),
          angle = this.getTotalAngle(),
          finalAngle = !!this.group && this.flipX ? angle - 180 : angle;
          rMatrix = util.calcRotateMatrix({ angle: finalAngle });
    rMatrix[4] = center.x; rMatrix[5] = center.y;
    var startMatrix = multiplyMatrices(vpt, rMatrix),
          finalMatrix = multiplyMatrices(startMatrix, [1 / vpt[0], 0, 0, 1 / vpt[3], 0, 0]),
          transformOptions = this.group ? fabric.util.qrDecompose(this.calcTransformMatrix()) : undefined,
          dim = this._calculateCurrentDimensions(transformOptions),
```

and avoid adding the changes to calcRotationMatrix and calcTranslateMatrix and extend the api surface adding an option that is unclear when it should be used.

For `_calculateCurrentDimensions` i m sure we can find a better option, and i remember is said i would do. That does not bother me for now

ShaMan123: 5deefc15

src/mixins/object_geometry.mixin.js
ShaMan123: I wonder/forgot why I accountd for flipX only for nested objects, why not for all objects?

ShaMan123: It's bad. Good catch.
Isn't flipX/flipY a z plane transformation? If so wouldn't it be better to multiply 3d matrices?

ShaMan123: rotation as used by fabric is a 2d plane transformation. So it shouldn't relate to z plane transformations. I think this is the cause of the problem

ShaMan123: What about [`ml-matrix`](https://mljs.github.io/matrix/)?
https://mljs.github.io/matrix/classes/QrDecomposition.html

ShaMan123: https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Matrix_math_for_the_web

src/mixins/object_geometry.mixin.js
```diff
@@ -571,15 +571,16 @@
...
+    calcOCoords: function () {
+      var vpt = this.getViewportTransform(),
+          center = this.getCenterPoint(),
+          tMatrix = [1, 0, 0, 1, center.x, center.y],
+          rMatrix = util.calcRotateMatrix({ angle: this.getTotalAngle() - (!!this.group && this.flipX ? 180 : 0) }),
+          positionMatrix = multiplyMatrices(tMatrix, rMatrix),
+          startMatrix = multiplyMatrices(vpt, positionMatrix),
+          finalMatrix = multiplyMatrices(startMatrix, [1 / vpt[0], 0, 0, 1 / vpt[3], 0, 0]),
+          transformOptions = this.group ? fabric.util.qrDecompose(this.calcTransformMatrix()) : undefined,
+          dim = this._calculateCurrentDimensions(transformOptions),
```
asturur, line 583: Just to be clear, i m sure there are bugs with angle compensation but is not important. This PR improves the situation anyway.

asturur:
For the reader of the future.
If the controlbox of an object appear different when is inside a group from when is outside, that is probably possible and is not considered a bug. As long as controls works in both situation that is considered an acceptable quirk.

asturur (approved):

src/mixins/object_geometry.mixin.js
ShaMan123: image

image

src/mixins/object_geometry.mixin.js
ShaMan123: `(!!this.group && this.flipX ? 180 : 0)`
I can't imagine this is good enough for sure.

ShaMan123:
> For the reader of the future. If the controlbox of an object appear different when is inside a group from when is outside, that is probably possible and is not considered a bug. As long as controls works in both situation that is considered an acceptable quirk.

😨

I see it as a bug in qrDecompose. Fixing it means moving to 3d transformation matrix or dropping flipX/Y for good (I think that's not a bad idea - very strange feature)

For the reader of the future - We are sorry this is how it is. **STAY AWAY** from `flipX/Y`.

src/mixins/object_geometry.mixin.js
ShaMan123: @asturur Thoguhts?

ShaMan123: https://codepen.io/ktamas77/pen/AXmmeX?editors=0010

ShaMan123: I guess we can use matrix projection from 3d -> 2d...

ShaMan123: But it open up 3d and I'm not sure we want that.
So I vote to remove flip props as they are too broken

ShaMan123: https://en.wikipedia.org/wiki/Orthographic_projection

asturur:
flipX and flipY are totally normal, and are just an abstraction for a negative scale that is totally normal too.
You can't propose someone to have nested groups of different types but not being able to flip an image, is a basic feature imho.
qrDecompose isn't bugged in a strict sense, it just does what it can do.
As we talked before we have more trasform variable  than 6 numbers and we don't have as many relationship between numbers.

is like when you get a 6, was it 3 x 2 or 6 x 1? you can't know.

I m not sure a different matrix would help, unless rotation and scale occupy different cells in the matrix.

[scaleX * cosA] [ -sinA ] [translateX]
[sinA] [scaleY * cosA] [translateY]

idependently from the fact that the mulitplication between trasnformation make everything even more complicated, the biggest lie is in the fact that the position 'a' and 'd' of the transform matrix are occupied by scale ( and so flip ) and a part of the rotation. with the cosine that can be negative or positive and the scale too, once you have those multipled together, if you decompose a matrix without knowing how the object ( and the parents ) are flipped or not flipped, you can't get out of it. 
Oh and skewX and skewY are also there somewhere together with rotation. I just see more variables than equations between them i can think of.

ShaMan123:
> flipX and flipY are totally normal, and are just an abstraction for a negative scale that is totally normal too.

Is it negative scale or 180deg rotation around the x/y axis? Or are the 2 cases the same thing?

> Oh and skewX and skewY are also there somewhere together with rotation

Confusing indeed. However the characteristic polynomial might serve as an answer in this case (not sure about it though) 'cause if a matrix is multiplied by a rotation matrix it doesn't have a solution for the characteristic polynomial. Meaning that we can check to see if there is a solution and if so we now the angle was 0. And I read about extracting a rotation matrix be it's heavy stuff.
https://github.com/ShaMan123/fabric.js/commit/22e472bcc5243733eaed96fad6a4932da25eab06

Did you look at `ml-matrix`?

> I just see more variables than equations between them i can think of.

Yes, Is true.

PR #7863, opened 2022-04-04
feat(Group): render selected objects

ShaMan123:
This is here to hold the `_renderObjects` logic in case we shall need it
View [diff](https://github.com/fabricjs/fabric.js/compare/v6-group-patch2...v6-group-patch5)


ShaMan123:
closing this in favor of a simpler approach
```diff
- this._objects[i].render(ctx);
+ this._objects[i].group === this && this._objects[i].render(ctx);
```

PR #7864, opened 2022-04-05
fix(IText): some RTL selection issues

ShaMan123:
Renamed relevant left/right methods to start/end (I am quite sure I missed some, logic is quite messy). This will make things clearer.
Adapted the move cursor logic. It is still far from perfect but is much better.


asturur:
We have:
- a wide rename of functions
- adding new option to align values
- some fixes

I don't want to use the time to demonstrate that they could have split, but reviewing and understanding the pr in this way is way more complicated it could be.

ShaMan123:
Sorry, it was a mess to work on this going back and forth between branches.

ShaMan123:
Isn't the description informative enough?

asturur:
The description is ok, it does not change that the PR is huge and could be not.

A large pr where you say `i just renamed methods for consistency in this PR` gets read scrolling fast and merged.

you-have-to-train-to-do-smaller-prs.

And if is too slow, is too slow, i can't do much about it.

ShaMan123:
Alright, I will break it down if you haven't read through it already.
have you?

igalya:
@ShaMan123 hi there, I'm working with RTL direction in textBox, In version 4.6 it has problems with selection when text is aligned to center or right.
I checked 5.2.4 and it seems to be all messed up.
Am I missing something? 
Here's an example.
http://jsfiddle.net/Igal_lib/bxm0j2Lv/7/

ShaMan123:
This PR isn't merged so it is obvious your fiddle is broken.
http://jsfiddle.net/7f916mpj/1/

igalya:
oh, thanks. 
 is there a merged version that implements RTL text selection ?

ShaMan123:
`Home` `End` still broken

faizux:
@ShaMan123 Many thanks for the RTL update. I successfully integrated fabric-rtl@5.1.0 into my project, and it's working well. However, I encountered some issues after downgrading from fabric@5.3.0, which I was previously using. I wonder if you have a similar implementation for fabric@5.3.0 that can be used? TY

ShaMan123:
I plan to migrate it to v6.
Hopefully published in 6.1 but can't promise
I recommend all devs to migrate to v6. Too many fixes to miss out

oustr:
> I plan to migrate it to v6. Hopefully published in 6.1 but can't promise I recommend all devs to migrate to v6. Too many fixes to miss out

I still encountered text selection problems when direction is "rtl" and textAlign is "right", that I couldn't select any part of the text, and when I click in the middle of the text the cursor is always at the right. My version of fabricjs is "^6.0.0-beta19", and I wonder if this bug is solved or not as all the rtl issues and discussions I found here on github still confuse me a lot😥 @ShaMan123 Seems the fix is still not merged?

PR #7865, opened 2022-04-05
feat(IText): deprecate `KeyboardEvent#keyCode` in favor of `key`

ShaMan123:
- deprecate `keyCode` (mdn states it is) in favor of `key | code`
- support passing functions to keyMap
- fix: modifier functions prior to non modifier in calling logic


src/mixins/itext_key_behavior.mixin.js
```diff
@@ -116,11 +120,24 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
...
-    if (e.keyCode in keyMap) {
-      this[keyMap[e.keyCode]](e);
+    var func;
+    if (e.key in keyMap) {
+      func = keyMap[e.key];
+    }
+    else if (e.keyCode in keyMap) {
+      func = keyMap[e.keyCode];
+    }
+    else if ((e.key in this.ctrlKeysMapDown) && (e.ctrlKey || e.metaKey)) {
```
ShaMan123, line 130: Shouldn't the ctrl maps come before the general key maps?

ShaMan123 (commented):
nice and simple

src/mixins/itext_key_behavior.mixin.js
asturur: eventually yes. Probably there was not overlapping?

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -116,11 +120,24 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
...
     }
     else if ((e.keyCode in this.ctrlKeysMapDown) && (e.ctrlKey || e.metaKey)) {
-      this[this.ctrlKeysMapDown[e.keyCode]](e);
+      func = this.ctrlKeysMapDown[e.keyCode];
+    }
+    if (typeof func === 'string' && typeof this[func] === 'function') {
+      func = this[func];
+    }
+    if (typeof func === 'function') {
+      func.call(this, e);
```
asturur, line 140: no let's do that if someone needs a function that does not exist, it adds it to the class and continue to call it by name.

asturur:
keyCode is probably deprecated, important is that key is fully available in all supported browser.
Probably ie11 was the reason we were on keyCode.

Can we verify that? in case we don't need to support keyCode anymore.
Do the values between key and keyCode differ?

src/mixins/itext_key_behavior.mixin.js
ShaMan123: Why should it add it to the class?
That's strange.

src/mixins/itext_key_behavior.mixin.js
ShaMan123: No overlapping in fabric. I had in mind the dev wanting to override what happens when a ctrl modifier is active.
Current logic doesn't allow.

ShaMan123:
They are totally different.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key#browser_compatibility
All browsers support it and MDN officially suggests using it instead of the rest.
image

ShaMan123:
Now if I want to remove keyCode from this PR I will get a conflict with #7864 (This is the part I dislike about splitting PRs - but I do undersrand it's a must)

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -116,11 +120,24 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
       return;
     }
     var keyMap = this.direction === 'rtl' ? this.keysMapRtl : this.keysMap;
-    if (e.keyCode in keyMap) {
-      this[keyMap[e.keyCode]](e);
+    var func;
```
melchiar, line 123:
```suggestion
var keyMap = this.direction === 'rtl' ? this.keysMapRtl : this.keysMap,
    func;
```
maybe comma separate these declarations?

src/mixins/itext_key_behavior.mixin.js
asturur: well right now the map is:
```
{
  keycode: functionName
}
```
after can be:

```
{
  keycode: functionName
  otherkeycode: (e) => { ...doSomething }
}
```

So you have:

IText.prototype.ctrlKeysMapUp.otherKeyCode that is a function.  You added it to the class. But as name it has the keycode.

today you have:

IText.prototype.ctrlKeysMapUp.'65' = 'selectAll' and then you have IText.prototype.selectAll

the function is added to the prototype and its name to the map.
Those approach are totally similar, one is not more dynamic of the other, and so i don't see the reason to have both options.

asturur:
i think that if we want to duplicate keycode in favor of key, we should change the maps to have the key and we can leave keycode in with a dev warning.

PR #7866, opened 2022-04-05
fix(Canvas): safeguard canvas add

ShaMan123:
removes object from previous canvas when adding to a new one

### DIFF
--- a/src/static_canvas.class.js
+++ b/src/static_canvas.class.js
@@ -569,6 +569,13 @@
      */
     _onObjectAdded: function(obj) {
       this.stateful && obj.setupState();
+      if (obj.canvas && obj.canvas !== this) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Canvas: trying to add an object that belongs to a different canvas.\n' +
+          'Resulting to default behavior: removing object from previous canvas and adding to new canvas');
+        /* _DEV_MODE_END_ */
+        obj.canvas.remove(obj);
+      }
       obj._set('canvas', this);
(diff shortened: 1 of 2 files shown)


src/static_canvas.class.js
```diff
@@ -569,6 +569,13 @@
      */
     _onObjectAdded: function(obj) {
       this.stateful && obj.setupState();
+      if (obj.canvas && obj.canvas !== this) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Canvas: trying to add an object that belongs to a different canvas.\n' +
+          'Resulting to default behavior: removing object from previous canvas and adding to new canvas');
+        /* _DEV_MODE_END_ */
+        obj.canvas.remove(obj);
+      }
```
asturur, line 578: i think this is totally fair, because the old reference in the canas would be just a broken reference.

i used this a couple of times to render objects in different canvases for copy purposes, the thing i was doing was to swap the canvas reference manually, while now i should go through the add/remove process and avoid extra renders.

asturur (approved):

src/static_canvas.class.js
ShaMan123: Do we want to check for duplicates as well? I think it's damaging for perf and I don't see too much advantage for it

asturur: if we do we can do in the dev mode braket and remove it in prod build.

ShaMan123:
This exposed a regression for toCanvasElement.
PRing soon.

PR #7867, opened 2022-04-06
fix(Group): 6th Patch of New Group 🛠️ layout, angle and origin ⚡

ShaMan123:
Group now respects originX/Y + angle/skewing!
It is a minor fix (to the layout engine) but a significant one.
I have committed all binaries in a single commit so you can view the diff without them.
View the [diff](https://github.com/fabricjs/fabric.js/compare/v6-group-patch2...v6-group-patch6-fix-origin-angle-layout)
~~Merge #7859, update from master and merge this one~~
There are some conflicts with #7860 so we better merge that first (because conflicts are **resolved** there) or perhaps extract group logic and merge that.


ShaMan123:
updated description

ShaMan123:
Closed because history got polluted (I need to get past my git history limitations)
Rebased to show the correct commit history
Replaced by #8004

PR #7869, opened 2022-04-07
using indexed property setter is not supported

rubenlagatie:
When fabric is used on a canvas that is part of an angular app, you get the following error when the canvas is being disposed:
"Failed to set an indexed property on 'CSSStyleDeclaration': Indexed property setter is not supported."
This is because angular replaces some setters on the style object for binding.
Using the setProperty method fixes the problem and is supported in all browsers.

### DIFF
--- a/src/util/dom_style.js
+++ b/src/util/dom_style.js
@@ -26,7 +26,7 @@
         var normalizedProperty = (property === 'float' || property === 'cssFloat')
           ? (typeof elementStyle.styleFloat === 'undefined' ? 'cssFloat' : 'styleFloat')
           : property;
-        elementStyle[normalizedProperty] = styles[property];
+        elementStyle.setProperty(normalizedProperty, styles[property]);
       }
     }
     return element;


willlahr (approved):

asturur:
Sorry i lost this PR in the cracks of the many emails.
Seems good to me.

asturur:
@ShaMan123 @melchiar any doubt about this?

ShaMan123:
seems fine
https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty

asturur (approved):

dennisideaonce:
I am also facing the same issue!

DinosaurDad:
I'm experiencing this same issue in a Vue app that moved from Fabric 4.6.0 to 5.2.1. Is 5.2.2 inbound?

ShaMan123:
Did this fix solve your issue?

solomax:
@ShaMan123 this one looks like the issue I'm facing (not using Vue or Angular)
I'll try to test this fix locally (hopefully will be able to build :)))

Thanks for the pointer :)

DinosaurDad:
> Did this fix solve your issue?

@ShaMan123 yes it did. Rather than rebuild fabric, I just assigned `fabric.util.setStyle` to the updated version from this PR. It's working great.

solomax:
Works for me as expected :)
Thanks for the fix!

Any ETA for release? :)))

ShaMan123:
v6 beta should be released soon.
v5 I'm not sure...
@asturur ?
I'll merge this into v5 as well.

solomax:
@asturur  `5.2.1` was released in Februrary
maybe it's not that bad idea to do another release? :))

solomax:
ping? :))

zshno1:
Any ETA for release?

solomax:
@ShaMan123 @asturur could you please share the plans regarding the next release date? :)

ShaMan123:
I can't answer that, I can only guess. @asturur can.

ShaMan123:
I can tell you we are focused on v6 and a release should be available soon as well. But that isn't what you need nor asked.

solomax:
I'll try to migrate to v6 :)
hopefully it wouldn't be too much work :)

ShaMan123:
It's work, lot's of breaking changes that can go beneath the radar from what I've experienced. But if we manage to migrate to TS in this version it should be easier.

asturur:
if you need this fix backported in a 5.3 that can be done quickly

solomax:
Thanks @asturur !
We are at v5 now, so would love to have new version with this fix released :)

solomax:
@asturur could you please release `5.2.2/5.3.0`?
peeease :))

solomax:
seems `5.x` release can't be done quickly after all ... :((

melchiar:
I believe asturer said the fix can be backported quickly, that doesn't mean the next release is ready yet. Why not just create a new build yourself?

solomax:
Hello @melchiar,
I actually have this fix in our code base (which is a bit dirty .... :( )
I can't create custom build since we are open-source project - have to be compilable from sources :)

DinosaurDad:
Copy the code from this gist:

[fabric 5.2.1 style fix](https://gist.github.com/DinosaurDad/6fae60f833a6580f5d166ba8fe1437ef)

Paste it in to your project. It will patch fabric without needing to maintain a build. Easy to add now, easy to remove later.

If it helps these guys keep building v6 then I'm all for a temporary solution.

solomax:
@DinosaurDad already have this piece of code copy/pasted :(
I would prefer to have fixed library :)
And I believe other might also like to use new, fixed release :)

ShaMan123:
v5.2.4 is published

solomax:
Thanks a million @ShaMan123 !
seems to work for me as expected :)

ShaMan123:
Great
I did a mistake not bumping a minor cause there's a breaking change in `Text#toJSON` so look out for that

PR #7875, opened 2022-04-10
chore(ci): rename option

ShaMan123:
Minor rename.

I would love to ditch QUnit in favor of something better like chai

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -143,7 +143,7 @@ function exportToWebsite(options) {
 /**
  * 
  * @param {string[]} tests file paths
- * @param {{debug?:boolean,recreate?:boolean,show?:boolean,filter?:boolean}} [options] 
+ * @param {{debug?:boolean,recreate?:boolean,verbose?:boolean,filter?:boolean}} [options] 
  */
 function test(tests, options) {
     options = options || {};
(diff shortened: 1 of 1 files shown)


asturur:
> I would love to ditch QUnit in favor of something better like chai

Does it run in the browser?
We have 2 sets of test, unit tests and visual one.
Between the unit tests there are some that are more meaningful in the browser ( dom elements and events ).

I would love to have Jest and node only for all the things that are pure functional testing, and have something that can run in the browser for the rest.
QUnit is a bit old and less known, but to be realistic, does it make sense to use time on  tests? the website and docs could be at least 10x time better

asturur (approved):

ShaMan123:
https://github.com/mochajs/mocha
https://github.com/chaijs/chai
https://github.com/facebook/jest

https://github.com/sinonjs/sinon - stubs!


All these support node/browser.
Jest has more downloads but is under facebook... It can turn into a disadvantage as soon as they loose interest in it. Though it looks very solid. Same as React.
I don't have any preferences.
What I do want is a framework with excellent docs, an easy API and a relatively easy migration (very obvious...).
I think we need something that can mock browser events as well (#7802 needs that for tests). `sinon` might be able to that, not sure.

ShaMan123:
> QUnit is a bit old and less known, but to be realistic, does it make sense to use time on tests? the website and docs could be at least 10x time better

10x for sure.

> does it make sense to use **time** on tests?

typo? I don't understand.

ShaMan123:
BTW we are compromised in the visual tests because of pixel comparison. Empty pixels are acounted for so the percentage is wrong. If you run `npm test -- -s visual -r` you will how different visuals are in passing tests.
So we should address that as well.

PR #7877, opened 2022-04-15
feat(Object Stacking): 🔙 refactor logic to support Group 🔝

ShaMan123:
Refactored logic to support object stacking operations on Group by moving relevant logic from StaticCanvas to Collection.
Renamed methods to enable the following:
```js
group.sendToBack();   // will send group to back in it's parent
group.sendObjectToBack(group.item(5));   // will move the object at index 5 to index 0
```

I think we should remove object stacking methods (`src/mixins/object_stacking.mixin.js`) from `Object`. What happens when an object is part of `ActiveSelection` and an object stacking method is called on it is strange and confusing.
Consider the following:
```js
let obj, canvas;
canvas.getActiveObject().item(0) === obj //   true
canvas.getActiveObject().item(0).bringToFront()  // mutates canvas' stack
canvas.getActiveObject().item(0) === obj //   true
```

With this PR the dev can achieve the following:
```js
obj.bringToFront() // acts on the origin stack manager, not on the ActiveSelection 
canvas.getActiveObject().bringObjectToFront(obj) // acts on the ActiveSelection 
```

Should we drop object stacking methods on Object and have the dev decide which stack to mutate?
```js
canvas.getActiveObject().bringObjectToFront(obj)
// in case obj is part of some group and is in active selection it's group prop is overridden by the ActiveSelection
// parent is canvas or group, not AS
obj.getParent().bringObjectToFront(obj) 
```

### BREAKING
canvas object stacking methods were renamed (`sendToBack` -> `sendObjectToBack` etc.)


ShaMan123 (commented):
updated from master

ShaMan123:
This PR will be ported to the `collection-stacking` branch.

PR #7878, opened 2022-04-15
fix(ActiveSelection): 🔃 `preserveObjectStacking` 📌

### LINKED ISSUE #4529: Preserve order of selected objects in multiple selection
Feature request:

Currently when user selects multiple objects (holding down Shift key) the code in the library converts single selection to multiple selection when selecting the second object with the following code:
```javascript
_createGroup: function(target) {
  var objects = this.getObjects(),
      isActiveLower = objects.indexOf(this._activeObject) < objects.indexOf(target),
      groupObjects = isActiveLower
        ? [this._activeObject, target]
        : [target, this._activeObject];
  this._activeObject.isEditing && this._activeObject.exitEditing();
  return new fabric.ActiveSelection(groupObjects, {
    canvas: this
  });
}
```

As implemented the order of two first items in the multiselection depends on the Z-order of these two first objects. As such, the library prevents application developer from knowing the order in which objects were selected. Such knowledge is important to implement group operations (like align for example) where first selected object is used as a "reference".

https://stackoverflow.com/questions/47614090/preserve-the-order-of-selected-objects

### LINKED ISSUE #6840: Object order changes when grouping items selected individually (shift-clicked) with preserveObjectStacking: true
## Version
4.3.1

## Test Case (use "current module view" to see graphics)
https://codesandbox.io/s/amazing-borg-5srzj?file=/src/index.js

browser
Which browsers: chrome and firefox

## Steps to reproduce
In the test case, there are 3 overlapping colored rectangles - green on top, which overlaps blue, which overlaps red.  When you select the green one first, then shift-click the blue one, then shift-click the red one they are all members of an `activeSelection`.  The button on the bottom will change the `activeSelection` into a `group`.  Press the button and when the `activeSelection` converted into a `group`, the red rectangle changes position and now overlaps the blue rectangle - the order for display has changed

## Expected Behavior
After they are selected as described above and the button is pressed, the 3 rectangles should now become a `group` and the display order should not change.   It does not change when `preserveObjectStacking` is `false`, only when it is `true`

## Actual Behavior
When converted into a `group`, the red rectangle changes position and now overlaps the blue rectangle - the order for display has changed.  But this only happens when `preserveObjectStacking` is `true`

ShaMan123:
closes #6840 
closes #4529 

~~First merge #7859, update from master and then merge this
[diff](https://github.com/fabricjs/fabric.js/compare/v6-group-patch2...v6-group-patch7)~~

We should consider a use case that we want a shift click to disregard object stacking.
For now the dev can do one of the following:
-  listen to the `object:added` event (*we need to keep a constant ref to ActiveSelection for events*)  - **DONE** by #7882 
```js
const activeSel = canvas.getActiveSelection();
activeSel .on('object:added', (object) => {  activeSel.moveObjectToFront(object); });
```
- override `_onAfterObjectsChange` **OR**  listen to `selection:updated`  and move the object to the front of the stack 

## TODO

- [x] add tests


ShaMan123 (commented):
updated from master

src/shapes/active_selection.class.js
```diff
@@ -98,6 +98,7 @@
      * @param {fabric.Object[]} targets
      */
     _onAfterObjectsChange: function (type, targets) {
+      this.callSuper('_onAfterObjectsChange', type, targets);
```
ShaMan123, line 101: ported

ShaMan123 (commented):
apart for some insignificant tests in group this PR is ported to #8665

PR #7879, opened 2022-04-16
feat(V6): interactive clip path

ShaMan123:
interactive clip path
[Edit fabric_app](https://codesandbox.io/s/fabric-app-forked-b22m4p?fontsize=14&hidenavigation=1&module=%2Fsrc%2FManyLayouts.tsx&theme=dark)

[diff](https://github.com/fabricjs/fabric.js/compare/v6-group-patch2...v6-interactive-clip-path)


asturur:
This is old now and obsolete i think we can close this since the tweak it was doing are probably not more applicable with layout manager.
If there is something to salvage from here let's reopen and port it

ShaMan123:
It is not obsolete and can be turned into a demo at minimum

PR #7885, opened 2022-04-18
fix(Canvas): `dispose` race condition

### LINKED ISSUE #7876: Error "Cannot read properties of null (reading 'save')" for `renderTopLayer` method calling
## Version
4.6.0

## Information
In some cases we get error "Cannot read properties of null (reading 'save')".
I don't know why it is, but as i see, we don't check type of object `ctx` in the `renderTopLayer` method.

Let's insert type guard there or will review calls to `renderTopLayer` and insert condition there.

image (49)
image (50)

ShaMan123:
calling static dispose before interactive dispose to prevent race condition as described in #7876 
should close #7876 and some other strange bugs.

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1273,21 +1273,24 @@
      * @chainable
      */
     dispose: function () {
-      var wrapper = this.wrapperEl;
+      var wrapperEl = this.wrapperEl,
+          lowerCanvasEl = this.lowerCanvasEl,
+          upperCanvasEl = this.upperCanvasEl,
+          cacheCanvasEl = this.cacheCanvasEl;
       this.removeListeners();
-      wrapper.removeChild(this.upperCanvasEl);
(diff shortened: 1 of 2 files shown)


asturur:
The PR seems fine to me, and seem innocent enough to merge it without digging deeper.
But would be nice to understand where the race condition was and caused by which operation.
Is because only the staticCanvas dispose call for a cancel render?

asturur (approved):

asturur:
Is approved, so you can merge whenever. But try to explain

ShaMan123:
> Is because only the staticCanvas dispose call for a cancel render?

Yes. What happened was that Canvas#dispose nullified topContext and there was a small gap of time that rendering could occur before StaticCanvas#dispose terminated requested rendering.

Hew007:
Why hasn't it merged and fixed？

PR #7882, opened 2022-04-18
fix(ActiveSelection): 🚀 multiple nested selection 🗂️

ShaMan123:
# Multiple Nested Selection
This PR finalizes (so it seems) the group rewrite!

## Gist

- refactored `_chooseObjectsToRender`
- made canvas hold a constant ref to `ActiveSelection` instead of creating new ones all time (important for devs wanting to attach events to it for customization, see #7878)
- introduced `Group#invalidate` - to handle children/ActiveSelection state changes that should propagate to group + an option for layout customization (`progress` - very experimental and unstable at the moment, remains optional)
- invalidate owning group of object under ActiveSelection 

The `progess` layout trigger is designed to enable updating group's size dynamically but it introduces unwanted changes and craziness. e.g. when object is rotating. We should consider if and how to support it. I think it should remain as is, the dev can override `getLayoutStrategyResult` and do what they want.

## TODO
- [x] figure out what to do with exporters (toDataUrl, toCanvas) on group while some objects are selected and `canvas.preserveObjectStacking = false` and therefore not rendered by group - **DONE** 79db7d5b
- [ ] `test/visual/golden/group-layout/selected_object.png` - seems the selected rect is doubled in opacity...
- [ ] rethink ActiveSelection caching - why not?

## Dev

Video was heavy so visuals are poor but perf is amazing
ezgif com-gif-maker (1)

closes #6776

[Edit fabric_app (forked)](https://codesandbox.io/s/fabric-app-forked-i29woj?fontsize=14&hidenavigation=1&module=%2Fsrc%2FManyLayouts.tsx&theme=dark)

<details> <summary>test code to paste in the kitchensink demo</summary>

```js

// clear canvas
canvas.clear();
 canvas.preserveObjectStacking = false;
    const circle = new fabric.Circle({ left: 100, top: 50, radius: 50 });
    const text = new fabric.Text('', { evented: false });
    const itext = new fabric.IText('Edit me\nfit-content layout', { left: 100, top: 150 });
    const g = new fabric.Group([
      new fabric.Rect({
        top: 200,
        width: 50,
        height: 50,
        fill: 'red',
        opacity: 0.3
      }),
      circle,
      itext
    ], {
      backgroundColor: 'blue',
      subTargetCheck: true,
      interactive: true
      //objectCaching: false
    });
    canvas.add(
      text,
      g
    );
    g.clone().then(clone => {
      clone.item(2).set({ text: 'Edit me\nclip path layout' });
      clone.set({
[...]

PR #7893, opened 2022-04-20
fix(V6): nested selection 🔧 fix canvas exporting 🧩

ShaMan123:
## Gist
Solves an edge case for #7882. 
- [ ] Merge that and update this from master.
- [x] merge #7907 and fix canvas export `filter` option
- [ ] It might be that propagating `forClipping` down via the rendering context is wrong... need to confirm

Group doesn't render selected objects if `preserveObjectStacking` is false. This poses a problem when calling `isTargetTransparent` and `toCanvasElement`.
This PR solves these cases.

- the rendering cycle now passes a context object that can be used for various cases. Very useful. I needed it to inform group to render itself with all its objects. This is used by `isTargetTransparent` and `toCanvasElement` and even enhances the `filter` option of `toCanvasElement` that is now deep.
- adapted logic
- added tests

View the [diff](https://github.com/fabricjs/fabric.js/compare/v6-nested-selection...v6-nested-selection-fix-export)


ShaMan123 (commented):
updated from #7882

PR #7900, opened 2022-04-22
fix(Group): patch2 minors

ShaMan123:
## To Do
- [x] `canEnter` - https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846827347
    - [x] naming
    - [x] throwing or what? https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846754553 Errors will raise in current code after warning is logged. Meaning it shouldn't warn, severity is error.
- [ ] @asturur agree on `isInFrontOf`
- [ ] ~~minor dead code in `text.class` https://github.com/fabricjs/fabric.js/pull/7859#issuecomment-1088259426~~
- [x] Group renderBackground - migrate to https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846930838
- [ ] `containsPoint` should be calculated relative to canvas coordinate plane
- [x] collection add/insertAt/remove return values
- [ ] remove object monitors?
- [ ] verify edge case with this._objectsToRender invalidation
- [ ] find a different solution for _getTransformedDimensions, @asturur improve my understanding of the group layout needs


src/shapes/group.class.js
```diff
@@ -226,18 +226,12 @@
...
      * @returns
      */
-    canEnter: function (object) {
+    canEnterGroup: function (object) {
       if (object === this || this.isDescendantOf(object)) {
-        /* _DEV_MODE_START_ */
-        console.warn('fabric.Group: trying to add group to itself, this call has no effect');
-        /* _DEV_MODE_END_ */
-        return false;
+        throw new Error('fabric.Group: trying to add group to itself');
```
asturur, line 231: we have to stop going back and fort on this.
Let's not throw errors for now. swapping all the console.error with throw.error is a search/replace away.

src/shapes/group.class.js
```diff
@@ -410,7 +404,6 @@
      * @param {CanvasRenderingContext2D} ctx Context to render on
      */
     drawObject: function(ctx) {
-      this._renderBackground(ctx);
```
asturur, line 413: any reason why you let this go at the end?

ShaMan123: https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846930838

src/shapes/group.class.js
ShaMan123: Did you read this? https://github.com/fabricjs/fabric.js/pull/7859#discussion_r846754553
If these conditions are met fabric will hit an exception anyways but it won't be clear as to why it happened. That's why I think throwing is a MUST.

src/shapes/group.class.js
ShaMan123: solved by #7901

ShaMan123:
@asturur I am waiting upon your decisions.
I have stated my opinions on the matters and I wish to move forward... I am OK with what you decide. I/You can revert/change what's needed.
Let's get this in and merge the 2nd patch.
After that I will resolve some conflict and prepare the rest to be merged. They are much more contained in scope.

src/shapes/group.class.js
asturur: we can use a console.error, and the exception will throw for us?
So what returning false does if not avoiding this object to enter the group?

asturur: ok so canEnterGroup is, couldHaveEnteredGroup.
Because is called in the callback via enterGroup.
The solution here is to move canEnterGroup out of the callback, in the add function of the group to gatekeep from entering.

asturur: opening a PR on this branch

ShaMan123: > ok so canEnterGroup is, couldHaveEnteredGroup. Because is called in the callback via enterGroup. The solution here is to move canEnterGroup out of the callback, in the add function of the group to gatekeep from entering.

Exactly

ShaMan123: The advantage of moving it to add/insertAt is clear. That's how it should be. The disadvantage is a drop in perf for large groups because we add an iteration over objects. Unless we manage to change collection logic to run all logic on object in one single iteration

ShaMan123:
closed in favor of #7916

PR #7903, opened 2022-04-25
fix(IText): clipboard events

ShaMan123:
continues #7802

Refactors itext clipboard events to utilize browser events instead of synthetic events generated by fabric.
This PR removes all limitations to the clipboard events and conforms with HTML standards.
Previously for example you couldn't copy text between different pages with fabric based apps. Now it's supported out of the box.

### DIFF
--- a/HEADER.js
+++ b/HEADER.js
@@ -119,15 +119,6 @@ fabric.charWidthsCache = { };
 fabric.textureSize = 2048;
 
 /**
- * When 'true', style information is not retained when copy/pasting text, making
- * pasted text use destination style.
- * Defaults to 'false'.
- * @type Boolean
- * @default
- */
-fabric.disableStyleCopyPaste = false;
-
-/**
  * Enable webgl for filtering picture is available
  * A filtering backend will be initialized, this will both take memory and
  * time since a default 2048x2048 canvas will be created for the gl context
--- a/src/mixins/itext_key_behavior.mixin.js
+++ b/src/mixins/itext_key_behavior.mixin.js
@@ -30,7 +30,7 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
     fabric.util.addListener(this.hiddenTextarea, 'keyup', this.onKeyUp.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'input', this.onInput.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'copy', this.copy.bind(this));
-    fabric.util.addListener(this.hiddenTextarea, 'cut', this.copy.bind(this));
+    fabric.util.addListener(this.hiddenTextarea, 'cut', this.cut.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'paste', this.paste.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'compositionstart', this.onCompositionStart.bind(this));
     fabric.util.addListener(this.hiddenTextarea, 'compositionupdate', this.onCompositionUpdate.bind(this));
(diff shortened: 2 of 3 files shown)


src/mixins/itext_key_behavior.mixin.js
```diff
@@ -140,15 +137,9 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
 
   /**
    * Handles keyup event
-   * We handle KeyUp because ie11 and edge have difficulties copy/pasting
-   * if a copy/cut event fired, keyup is dismissed
    * @param {Event} e Event object
    */
   onKeyUp: function(e) {
-    if (!this.isEditing || this._copyDone || this.inCompositionMode) {
```
asturur, line 148: the fact that there is inCompositionMode is alarming.
We need to verify why we were blocking paste/copy in composition mode.

HEADER.js
```diff
@@ -118,15 +118,6 @@ fabric.charWidthsCache = { };
...
 fabric.textureSize = 2048;
 
-/**
- * When 'true', style information is not retained when copy/pasting text, making
- * pasted text use destination style.
- * Defaults to 'false'.
- * @type Boolean
- * @default
- */
-fabric.disableStyleCopyPaste = false;
```
asturur, line 128: i think we need this and i think is not hard to keep it in place. Flexible or not. Can be moved to canvas level if it helps

src/mixins/itext_key_behavior.mixin.js
```diff
@@ -277,39 +260,78 @@ fabric.util.object.extend(fabric.IText.prototype, /** @lends fabric.IText.protot
...
+    if (this.setClipboardData(e)) {
+      //  remove selection
+      this.insertChars('', null, this.selectionStart, this.selectionEnd);
+      this.selectionEnd = this.selectionStart;
+      this.hiddenTextarea && (this.hiddenTextarea.value = this.text);
+      this._updateTextarea();
+      this.fire('cut', { e: e });
+      this.fire('changed', { index: this.selectionStart, action: 'cut' });
+      this.canvas.fire('text:changed', { target: this });
+      this.canvas.requestRenderAll();
```
asturur, line 305: this before was happening from the cut event naturally? why that is not more true?

src/mixins/itext_key_behavior.mixin.js
ShaMan123: I am guessing here but it the copy was synthetic so when run with `inCompositionMode` it hit some edge case with `ctrl + v/c`

ShaMan123: Now browser is in charge of deciding what the action is so less edge cases for all. Hopefully

HEADER.js
ShaMan123: I don't care about this. It is dead code AFAIC.

ShaMan123: It is not used anywhere after this PR

src/mixins/itext_key_behavior.mixin.js
ShaMan123: Ohhh I am not certain anymore

src/mixins/itext_key_behavior.mixin.js
ShaMan123: Was it? you mean the browser did it?

src/mixins/itext_key_behavior.mixin.js
ShaMan123: > We need to verify why we were blocking paste/copy in composition mode.

It is blocked for all mouse up when not in editing mode. Maybe because of the char suggestion toolbar?

ShaMan123: If so I should revert

src/mixins/itext_key_behavior.mixin.js
ShaMan123: I see. It did but I've decided to prevent default and handle it internally. Makes more sense IMO. And crucial for customization/overriding.

PR #7907, opened 2022-04-26
fix(canvas export): regression caused by safegurading

ShaMan123:
#7866 exposed a regression when using `toCanvasElement`
This commit fixes it by working around `canvas.add` and fixes another issue:
`object.toCanvasElement()` would fire an added event which is absolutely undesired

### DIFF
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1738,7 +1738,7 @@
       if (originalGroup) {
         this.group = originalGroup;
       }
-      this.set(origParams).setCoords();
+      this.set(origParams);
       // canvas.dispose will call image.dispose that will nullify the elements
       // since this canvas is a simple element for the process, we remove references
       // to objects in this way in order to avoid object trashing.
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1729,12 +1729,8 @@
         canvas.backgroundColor = '#fff';
       }
       this.setPositionByOrigin(new fabric.Point(canvas.width / 2, canvas.height / 2), 'center', 'center');
-
-      var originalCanvas = this.canvas;
-      canvas.add(this);
-      var canvasEl = canvas.toCanvasElement(multiplier || 1, options);
+      var canvasEl = canvas.toCanvasElement(multiplier || 1, Object.assign(options, { objects: [this] }));
       this.shadow = originalShadow;
-      this.set('canvas', originalCanvas);
       if (originalGroup) {
         this.group = originalGroup;
       }
--- a/src/shapes/object.class.js
+++ b/src/shapes/object.class.js
@@ -1729,7 +1729,10 @@
         canvas.backgroundColor = '#fff';
       }
       this.setPositionByOrigin(new fabric.Point(canvas.width / 2, canvas.height / 2), 'center', 'center');
+      var originalCanvas = this.canvas;
+      this.set('canvas', canvas);
       var canvasEl = canvas.toCanvasElement(multiplier || 1, Object.assign(options, { objects: [this] }));
+      this.set('canvas', originalCanvas);
       this.shadow = originalShadow;
       if (originalGroup) {
         this.group = originalGroup;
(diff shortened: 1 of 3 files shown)


src/shapes/object.class.js
```diff
@@ -1729,12 +1729,11 @@
         canvas.backgroundColor = '#fff';
       }
       this.setPositionByOrigin(new fabric.Point(canvas.width / 2, canvas.height / 2), 'center', 'center');
-
       var originalCanvas = this.canvas;
-      canvas.add(this);
-      var canvasEl = canvas.toCanvasElement(multiplier || 1, options);
-      this.shadow = originalShadow;
+      this.set('canvas', canvas);
```
asturur, line 1733: switching from add to set('canvas') don't we miss a setCoords on the object that could be necessary for proper rendering?

asturur: i would add a setCoords here, just to be sure we change with something is equivalent

asturur:
What was the new issue?
I understand there was a past issue that was firing an extra added event.
is the new issue that now we are actually removing an object from the canvas, so going back to the original one is not working anymore?

src/shapes/object.class.js
asturur: Also, what about we do:
canvas.objects = [this] since this is a trash-away canvas and we avoid adding an extra options to toDataUrl?

When adding this option here:
https://github.com/fabricjs/fabric.js/pull/7907/files#diff-6748525f6b3662ac6540af0ceab4cbcd5e5d44069b54ec0b1c3d9edab1e894ecR64

Is unclear that passing the objects in, those have to have a canvas reference properly set and setCoords done.
Is just an open door for misunderstandings. let's solve it inside the object.toDataUrl function and leave the rest as it is.

ShaMan123: Ok this is good.
Regarding setCoords - why would we need it? 
https://github.com/fabricjs/fabric.js/pull/7907/files#diff-de494ae6278212520cae6374e5a8eecaa8d08248ab53f6f30c167949186292caR1726

ShaMan123:
The new issue was that the safeguarding logic removed the canvas ref from the StaticCanvas#add method and caused a bug when using with `toCanvasElement`

src/shapes/object.class.js
asturur: i think i would needed because if the object in question has offscreen coords, it will eventually not be drawn.
Since it needs to be drawn ( the whole reason of the dataurl export ) and coords can influence a drawing skip, and since we were implicitly setting them using add, i would continue setting them.
The coords are indeed reset at the end of the function. If we don't change them, there would be at least no reason to set them again.

asturur:
yes but which is the bug ? I think it is that it would missing in the object array of the original canvas, is it that?

src/shapes/object.class.js
ShaMan123: > de494ae6278212520cae6374e5a8eecaa8d08248ab53f6f30c167949186292caR1726

`skipOffscreen: false,`

ShaMan123:
It surfaced when I merged #7802 with #7859 I think. The drag image uses `toCanvasElement`  - it was there.

ShaMan123 (commented):
ready

src/shapes/object.class.js
```diff
@@ -1729,16 +1729,16 @@
...
+      canvas._objects = [this];
+      this.set('canvas', canvas);
       var canvasEl = canvas.toCanvasElement(multiplier || 1, options);
-      this.shadow = originalShadow;
       this.set('canvas', originalCanvas);
+      this.shadow = originalShadow;
       if (originalGroup) {
         this.group = originalGroup;
       }
-      this.set(origParams).setCoords();
```
asturur, line 1741: please let's leave both setCoords.
Now it should not matter, but we know correct coords are necessary for renders.
if something changes later or in an after:render loop, since this object to canvas export is a bit of an hack in which the object gets added to a canvas to be rendered, let's be sure we set coords correctly when we finish the procedure.
Let's not optimize when we know we are fragile in those kinds of operations. The example is the exact regression we are fixing, we introduced because we did not remember of this.
Let's not overdo and let's leave setCoords in place.

ShaMan123: Good point

ShaMan123: a2294f9b

asturur (approved):

Issue #7596, opened 2022-01-10
IMPORTANT: Migrating to modern javascript/typescript


asturur:
Hello everyone!

FabricJS is nearly 12 years old.

The library has still its place around the internet, is good for many things, is never perfect, but i know that for who approach rich canvas based interactive apps, this is a nice piece of software to leverage.

While our main fault are still lack of great docs and updated and clarifying examples, now also the javascript code looks like a lot obsolete.

There are things we could leverage like Promises, default value for arguments, fat arrow for loops and readibility, native classes, proper getters and setters.

We would like to take the occasion to revisit the shape of the codebase to something more modern.
This would include proably a bundler, proper import statements instead of a single file that gets built like lego blocks, maybe tree shaking for who import just what he wants to import.

This would make the custom built unnecessary, using other libraries easier including touch interactions.

So a move to es201X is coming anyway.

The question is if we want to add type descriptors or if we want to make a TS migration right away.

The benefits of consuming TS are clear to me, the benefit of writing typescript a little less, i'm not a great fan and also i m not great a TS.
Is very easy to use TS in the wrong way and produce subpar results.

We are open to suggestions and understand what is the best way to make everyone happy.
external type definition in form of d.ts files are a good middle way if there is a way to enforce correctness at build time.
A new lint profile is required.
A new build tool too.
Legacy JS support is not going away.
A bundler will produce the correct code for older browsers.
But probably IE11 is out of the game anyway. I would love to have 2 builds, a legacy one and a modern one ( the classic last 2 versions of each browser )

So which are your reason to list TS on the pro/cons of this change?
I understand you may not like TS. that is fine. I don't like it either.
But if is terse, with types out of the way, and a transpiled build is available, what is your main point against or for it?

melchiar:
@jimmywarting Typescript support is one of our most commonly requested features. Feel free to elaborate if you have specific objections to adding support for it.

jimmywarting:
esm is also a desirable request, working with typescript and esm is a total nightmare. beside you can have a type safety system with vanila js + jsdoc and writing the code in such a manner that it don't require annotation... like using # instead of private, use classes instead of prototype, avoid using promise constructor so the IDE can really figure out what something return and so on.

it's also possible to generate d.ts files from js also

i created https://jimmywarting.github.io/you-might-not-need-typescript/ to this end...

andrewzolotukhin:
[...]

PR #7914, opened 2022-04-30
refactor(controls): deprecate 🚫 SHARED 🚫 controls

ShaMan123:
This PR refactors `Object#controls` from a shared object attached to the prototype into a unique object attached to an instance.

## Motivation
https://github.com/fabricjs/fabric.js/issues/7596#issuecomment-1104981511
I hate the shared controls and the shared concept altogether.
It's extremely unintuitive and many issues were raised because of this. I have spoken to fabric devs and they all mentioned this.
IMO sharing instances should never be done in fabric (and maybe even at all - except for mega perf enhancements) and anyways rewriting JS to use `class` will and **MUST** deprecate this bad practice in favor of `static` properties/methods or simply bound properties/methods.

## Gist
- created `ObjectControls`, `TextboxControls` classes that hold the default controls.
- `controls` must have an `attach` method - best practice is to subclass `ObjectControls | TextboxControls`. When migrating to TS we could/should use an abstract class to define this class that the dev will have to subclass.
- **BREAKING** control API since `fabricObject` is no longer needed as it is referenced in the `control.object`

- [ ] I have no idea why this test fails.
image

## Migration
If a dev needs to change controls across many objects they can subclass or they can pass a controls object in initialization. But it cannot be **SHARED**

## Post PR
We should consider moving control logic from `src/mixins/object_interactivity.mixin.js` to `ObjectControls`


src/shapes/object.class.js
```diff
@@ -1002,6 +1004,9 @@
       else if (key === 'shadow' && value && !(value instanceof fabric.Shadow)) {
         value = new fabric.Shadow(value);
       }
+      else if (key === 'controls') {
+        value.attach(this);
```
ShaMan123, line 1008: not sure we need this because I couldn't get `controls` to be set in `setOptions` without breaking all tests

ShaMan123 (commented):
1 failing test - why does it fail?
1 comment

Besides this is good to go.

asturur:
I don't know custom controls works, this a change in name of best practices of which i see very little utility.
We have a TON of breaking changes, if we are not fixing anything i don't want to change this.

ShaMan123:
> I don't know custom controls works, this a change in name of best practices of which i see very little utility.

Didn't understand

> We have a TON of breaking changes, if we are not fixing anything i don't want to change this.

Absolutely. And the TS rewrite will introduce even more. 
That's why I think we should release v6 with this in it because it already requires a major adjustment. Plus, once we migrate this will happen naturally. It will require a very awkward work around to continue sharing objects between instances when written with proper classes.

src/mixins/default_controls.js
```diff
@@ -7,108 +7,127 @@
...
+      changeWidth = controlsUtils.changeWidth;
 
-  objectControls.ml = new fabric.Control({
-    x: -0.5,
-    y: 0,
-    cursorStyleHandler: scaleSkewStyleHandler,
-    actionHandler: scalingXOrSkewingY,
-    getActionName: scaleOrSkewActionName,
-  });
+  class ObjectControls {
```
asturur, line 14: since we are not migrating yet, this should still be the function + prototype class.

asturur:
To be clear i see this is more polished.
I don't think the current control api is broken or terrible.
It allows you to plug in your functions.
Even if we want to move to class system, there is no need to use attach and to move render function in.
Let's mitigate the impact, we can still get rid of the shared object ( that would be mitigated anyway when you go and read how it works ) but we don't need to break the api.
I have a bunch of custom controls and they just works.

Also i have questions, if i have a control group class for each object, as soon as i want to add a single control to all objects, like a delete button, how do i do for the instantiated objects? I have to loop and reassign controls.

What is good of the current pattern is that you may have multiple control set ready to be used by object, for example:
- locked object controls
- cropping picture controls
- normal controls
And just swap them when needed.
In this case you would have to write a class, and instantiate it each time.

> I don't know custom controls works, this a change in name of best practices of which i see very little utility.

With this i meant, i don't like to fix what is not broken, in the name of `this is a better practice`.
Was late and i wrote hastly.

ShaMan123:
I don't think the control API is horrible at all.
I simply dislike the shared object practice. Someone new to fabric will bash their head against the wall until they understand that's the origin of what they are experiencing as did I. Numerous issues have been raised and the devs I speak to dislike it as well (dislike is a soft term).
I do think it makes sense to iterate over objects to set controls over them. Entirely on the dev. 
And if I look from above it adds a logical layer to fabric that IMO should not exist.

Anyways you indulge me with the things I want/need/think are important so now it's my turn.

PR #7916, opened 2022-04-30
fix(Group): patch2 minors

asturur:
This PR rename canEnter in canEnterGroup and move canEnterGroup to an higher level of the logic so that it can stops adding not supported objects to groups.

Background color has been left in for group since the layers in group concept may be slow to enter fabricjs.

### DIFF
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -141,8 +141,8 @@
      * @param {...fabric.Object} objects
      */
     add: function () {
-      var possibleObjects = Array.from(arguments).filter(function(object) {
-        return this.canEnterGroup(object);
+      var _this = this, possibleObjects = Array.from(arguments).filter(function(object) {
+        return _this.canEnterGroup(object);
       });
       fabric.Collection.add.call(this, possibleObjects, this._onObjectAdded);
       this._onAfterObjectsChange('added', possibleObjects);
--- a/src/shapes/group.class.js
+++ b/src/shapes/group.class.js
@@ -593,7 +593,7 @@
         return this.prepareBoundingBox(layoutDirective, addedObjects, context);
       }
       else if (layoutDirective === 'fit-content' || layoutDirective === 'fit-content-lazy'
-          || (layoutDirective === 'fixed' && context.type === 'initialization')) {
+          || (layoutDirective === 'fixed' && (context.type === 'initialization' || context.type === 'imperative'))) {
         return this.prepareBoundingBox(layoutDirective, objects, context);
       }
       else if (layoutDirective === 'clip-path' && this.clipPath) {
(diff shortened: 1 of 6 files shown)


asturur:
replace #7900 
@ShaMan123 have a look at this solution to catching the issue before need to throw

ShaMan123 (approved):
This is much better no doubt.

However:

> The advantage of moving it to add/insertAt is clear. That's how it should be. The disadvantage is a drop in perf for large groups because we add an iteration over objects. Unless we manage to change collection logic to run all logic on object in one single iteration

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/pull/7900#discussion_r862421564_

But I think we should move forward

ShaMan123:
> Background color has been left in for group since the layers in group concept may be slow to enter fabricjs.

Could you explain why Layer will take time to merge into master?
I am fine with dropping background. I have worked around it in my projects.

ShaMan123 (changes requested):
Please push this
https://github.com/fabricjs/fabric.js/pull/7900/files#diff-216444c1bc3837d560c8241c4d9643b6de8c147095081c5b38715b9e058c24cbR587

asturur:
> > Background color has been left in for group since the layers in group concept may be slow to enter fabricjs.
> 
> Could you explain why Layer will take time to merge into master? I am fine with dropping background. I have worked around it in my projects.

Because i have seen the PR, and i don't like the idea of having special logic to allow layers to react to both canvas and group  container. I want to simplify things not make more complicated. The layer per se is ok, a dimensionless group always interactive. fine.
the gymmics in the layout management less.
But we can talk there, or maybe we could make an hands on and see how we can make it simpler.

asturur:
The only large group i know of is the spray brush, can you think of others?
maybe the spray brush shouldn't be a group with thons of rect or circles, it should be just a function able to render a spray effect.

ShaMan123:
I guess you are right. I don't use huge groups in my projects.

ShaMan123:
This next? Don't forget my review

asturur:
Applied fixes and merging

src/shapes/group.class.js
```diff
@@ -226,16 +230,17 @@
      * @param {fabric.Object} object
      * @returns
      */
-    canEnter: function (object) {
+    canEnterGroup: function (object) {
       if (object === this || this.isDescendantOf(object)) {
         /* _DEV_MODE_START_ */
-        console.warn('fabric.Group: trying to add group to itself, this call has no effect');
+        console.error('fabric.Group: trying to add group to itself, this call has no effect');
```
ShaMan123, line 236: so we should only warn now, right?

ShaMan123 (commented):
What about `insertAt` - we should filter that as well

PR #7918, opened 2022-05-01
feat(path): `getRegularPolygonPath`

### LINKED ISSUE #2437: Add RegularPolygon object type
Sorry, this is my first time using GitHub, and I'm not sure if this is the proper forum to get this to you. 

It seems that just having a triangle primitive type is insufficient. What if you want a square, pentagon, hexagon etc? I have written a small method that returns a path for a regular polygon with the specified number of sides which I would like to submit for inclusion in fabricjs. I don't think I'm competent enough to make the change in fabric source directly, but please consider creating a RegularPolygon object type.

The function I wrote below takes arguments of x, y, radius, numVertexes and returns an SVG path string for a regular polygon centered at x,y, with vertex points at radius from the center, and with numVertexes corners. 

``` javascript
function polarToCartesian(centerX, centerY, radius, angleInDegrees) {
  // This function converts polar coordinates to cartesian coordinates

  var angleInRadians = (angleInDegrees-90) * Math.PI / 180.0;

  return {
    x: centerX + (radius * Math.cos(angleInRadians)),
    y: centerY + (radius * Math.sin(angleInRadians))
  };
}

function getRegularPolygonPath(x,y, radius, numVertexes){
  // This function returns a path for a regular polygon centered at x,y with a radius specified with numVertexes sides

  var interiorAngle = 360/numVertexes;

  // rotationAdjustment rotates the path by 1/2 the interior angle so that the polygon always has a flat side on the bottom
  // This isn't strictly necessary, but it's how we tend to think of and expect polygons to be drawn
  var rotationAdjustment = 0;
  if (numVertexes % 2 == 0){
    rotationAdjustment = interiorAngle/2;
  }

  var d = [];
  for (var i=0; i<numVertexes; i++){
    // var coord = coordList[i];
    var coord = polarToCartesian(x, y, radius, i*interiorAngle + rotationAdjustment);

    if (i==0){
      d.push('M ');

      // If an odd number of vertexes, add an additional point at the top of the polygon-- this will shift the calculated center
      // point of the shape so that the centerpoint of the polygon is at x,y (otherwise the center is mis-located)
      if (numVertexes %2 == 1){
        d.push(0);
        d.push(radius);
      }

      d.push('M');

    } else {
      d.push(" L ");
    }

    d.push(coord.x);
    d.push(coord.y);
  }

  d.join(" ");

  d+= " Z";

  return d;

}

```

The current code is used like this:

``` javascript
    var pathStr = getRegularPolygonPath(50,25, 10, 8);

    var polygon = new fabric.Path(pathStr,
    {
      stroke: 'black',
      strokeWidth: 1,
      fill: 'blue',
      originX: 'center',
      originY: 'center',
    });

    // draws an octagon in this case
    canvas.add(polygon);
```

Note: I drew a separate circle object at the x,y location to show the origin point
octagon

Similarly we can draw other polygons as well

numVertexes = 5
image

numVertexes = 6
image

numVertexes = 14
image

I think it would be cleaner if we could just create a new RegularPolygon object and have it internally use the path, but I don't know how to do this.

``` javascript
    var polygon = new fabric.RegularPolygon(
    {
      left: 50,
      top: 25,
      radius: 10,
      vertexCount: 8,
      stroke: 'black',
      strokeWidth: 1,
      fill: 'blue',
      originX: 'center',
      originY: 'center',
    });
```

Something would probably need to be done to handle the left, top coordinate transform, but I don't know how that works internally in fabric.

Thanks for your consideration!

ShaMan123:
a util curtesy of @nickvans
closes #2437

https://jsfiddle.net/tqrms0ae/

### DIFF
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -813,8 +813,8 @@
 
   /**
    * Returns an array of path commands to create a regular polygon
-   * @param {number} radius 
-   * @param {number} numVertexes 
+   * @param {number} radius
+   * @param {number} numVertexes
    * @returns {(string|number)[][]} An array of SVG path commands
    */
   function getRegularPolygonPath(radius, numVertexes) {
@@ -822,19 +822,19 @@
     // rotationAdjustment rotates the path by 1/2 the interior angle so that the polygon always has a flat side on the bottom
     // This isn't strictly necessary, but it's how we tend to think of and expect polygons to be drawn
     var rotationAdjustment = -Math.PI / 2;
-    if (numVertexes % 2 == 0) {
+    if (numVertexes % 2 === 0) {
       rotationAdjustment += interiorAngle / 2;
     }
     var d = [];
     // If an odd number of vertexes, add an additional point at the top of the polygon in order to shift the calculated center
     // point of the shape so that the center point of the polygon is at 0,0 (otherwise the center is mis-located)
-    if (numVertexes % 2 == 1) {
+    if (numVertexes % 2 === 1) {
       d.push(['M', 0, radius]);
     }
     for (var i = 0, rad, coord; i < numVertexes; i++) {
       rad = i * interiorAngle + rotationAdjustment;
       coord = new fabric.Point(Math.cos(rad), Math.sin(rad)).scalarMultiplyEquals(radius);
-      d.push([i == 0 ? 'M' : 'L', coord.x, coord.y]);
+      d.push([i === 0 ? 'M' : 'L', coord.x, coord.y]);
     }
     d.push(['Z']);
     return d;
--- a/src/util/path.js
+++ b/src/util/path.js
@@ -826,11 +826,6 @@
       rotationAdjustment += interiorAngle / 2;
     }
     var d = [];
-    // If an odd number of vertexes, add an additional point at the top of the polygon in order to shift the calculated center
-    // point of the shape so that the center point of the polygon is at 0,0 (otherwise the center is mis-located)
-    if (numVertexes % 2 === 1) {
-      d.push(['M', 0, radius]);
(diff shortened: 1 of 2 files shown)


src/util/path.js
```diff
@@ -811,6 +811,30 @@
...
+    }
+    var d = [];
+    for (var i = 0, rad, coord; i < numVertexes; i++) {
+      rad = i * interiorAngle + rotationAdjustment;
+      coord = new fabric.Point(Math.cos(rad), Math.sin(rad)).scalarMultiplyEquals(radius);
+      d.push([i === 0 ? 'M' : 'L', coord.x, coord.y]);
+    }
+    d.push(['Z']);
+    return d;
+  }
```
asturur, line 836: yes this is nice, very useful.

src/util/path.js
```diff
@@ -811,6 +811,30 @@
     });
   }
 
+  /**
+   * Returns an array of path commands to create a regular polygon
+   * @param {number} radius
+   * @param {number} numVertexes
+   * @returns {(string|number)[][]} An array of SVG path commands
+   */
+  function getRegularPolygonPath(radius, numVertexes) {
```
asturur, line 820: can we swap and have numVertexes first? radius can be defaulted to something when we have es6

asturur:
I remember this function from that issue.
Can we add a simple test for 5 and 6 points to match a pre-defined string?
so that if the function gets changed the test needs to be updated and we can stop and think

src/util/path.js
ShaMan123: 99d5183e

asturur (approved):

PR #7920, opened 2022-05-01
Remove stateful concept ( without passing for deprecation first ).

asturur:
Stateful requires cloning and keep track of copies of properties.
It extends to
- statefulCache
- fire object:modified for state changes
- text size invalidation

Now statefulCache requires a lot of deep checks and is a bit slow, is disabled by default since long and is easily not needed if the developer uses the set method to set props and values. An easy let go.

object:modified was weird.
You would set fill = red from blue. Nothing would happen.
Then if you would execute a transformation, with no results color change would still be detected and object modified would fire. Was partial and a bit useless.
As for the cache, we could fire `object:modifed` when using set method over the sateProperties ( to be discussed )

For invalidating text size changes we will be clear in docs that now using the set method is necessary, and we will use the same logic we use for dirty prop.

## Missing

[ ] adapt tests
[ ] Write text alternative


src/shapes/group.class.js
```diff
@@ -417,29 +417,6 @@
...
-          if (this._cacheCanvas) {
-            // if this group has not a cache canvas there is nothing to clean
-            var x = this.cacheWidth / this.zoomX, y = this.cacheHeight / this.zoomY;
-            this._cacheContext.clearRect(-x / 2, -y / 2, x, y);
-          }
-          return true;
-        }
-      }
-      return false;
-    },
```
asturur, line 441: Why this goes all away?
https://github.com/fabricjs/fabric.js/pull/7920/files#diff-216444c1bc3837d560c8241c4d9643b6de8c147095081c5b38715b9e058c24cbL427
if is not statefulCache, return false, stop there.
The rest of the code was executed only if the group had `statefullCache` activated. Since it is no more possible to activate it, all of this function becomes just a callSuper, and so it can be removed.

ShaMan123: I don't understand and I don't even know what `statefulCache` is about

asturur: statefulCache determine that there is a cache invalidation comparing all the properties of the object with a copy of those when the cache was created. It is slow when you have gradients or large groups or a long path.

ShaMan123: I'm not sure I need to answer this question.
But group invalidation is done in `Object#_set`, this piece of code is an artifact, and a damaging one too.

asturur: it wasn't a question :D it was more what i was trying to answer to with the comment.
I would expect a reader does not get automatically why a large chunk of code is deleted

ShaMan123: OK now I get it
Question:
> Why this goes all away?

Answer:
> https://github.com/fabricjs/fabric.js/pull/7920/files#diff-216444c1bc3837d560c8241c4d9643b6de8c147095081c5b38715b9e058c24cbL427

ShaMan123:
#8573

PR #7932, opened 2022-05-10
refactor(object_interactivity): draw borders

ShaMan123:
This PR refactors logic to enable easier overriding + maintenance.

- exposes private `_drawBorders`
- exposes public `strokeBorders` for easier overriding

**BREAKING** d257c99:
- removed `drawBordersInGroup`
- changed `drawBroders` signature

### DIFF
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -172,7 +172,7 @@
             new fabric.Point(0, 0).scalarAddEquals(this.canvas.getZoom()) :
             new fabric.Point(options.scaleX, options.scaleY),
           stroke = strokeFactor.scalarMultiplyEquals(this.strokeWidth),
-          size = bbox.scalarAddEquals(stroke).scalarAddEquals(this.borderScaleFactor);
+          size = bbox.addEquals(stroke).scalarAddEquals(this.borderScaleFactor);
       this._drawBorders(ctx, size, styleOverride);
       return this;
     },
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -146,33 +146,24 @@
      * Requires public properties: width, height
      * Requires public options: padding, borderColor
      * @param {CanvasRenderingContext2D} ctx Context to draw on
-     * @param {Object} styleOverride object to override the object style
-     * @return {fabric.Object} thisArg
-     * @chainable
-     */
-    drawBorders: function (ctx, styleOverride) {
-      var size = this._calculateCurrentDimensions().scalarAddEquals(this.borderScaleFactor);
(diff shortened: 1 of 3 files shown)


src/util/misc.js
```diff
@@ -1154,10 +1154,7 @@
             }],
           transformMatrix = fabric.util.calcDimensionsMatrix(options),
           bbox = fabric.util.makeBoundingBoxFromPoints(points, transformMatrix);
-      return {
-        x: bbox.width,
-        y: bbox.height,
-      };
+      return new fabric.Point(bbox.width, bbox.height);
```
asturur, line 1157: change the jsdoc please

src/mixins/object_interactivity.mixin.js
```diff
@@ -108,73 +108,63 @@
...
 
     /**
-     * Draws borders of an object's bounding box.
-     * Requires public properties: width, height
-     * Requires public options: padding, borderColor
-     * @param {CanvasRenderingContext2D} ctx Context to draw on
-     * @param {Object} styleOverride object to override the object style
-     * @return {fabric.Object} thisArg
-     * @chainable
+     * @public override if necessary
```
asturur, line 111: we could say that this function is here exactly for the scope of overriding. Otherwise someone could argue, why having a method on a class that does not need `this` at all.
I would just say override this function in order to draw a different control box. example: rounded corners, different border style.
We should also inform whihc is the state of the context in that situation to give the dev some extra information.
(rotated aligned with object base, centered on the center of the object, scaled to absorb object scale and container scale and zoom )

ShaMan123: The ctx is in the same state as in the object's render method right?

ShaMan123: Or maybe not. I am confused

src/util/misc.js
ShaMan123: e65cb50a

src/mixins/object_interactivity.mixin.js
asturur: the ctx is not in the same state, because when you scale the object the controls do not grow.
i m also wrong the ctx is only centered and rotated, and is not scaled at all.

I m pointing out that `override in case of necessity` can be writte on top of any function or none.
Since you made this refactor with a keen eye on having it overridable for some reason, just indicate the reason as an example for future reader

ShaMan123: If I recall correctly it is rotated and translated to center. But not scaled or skewed...

ShaMan123: 36b7608e

ShaMan123:
READY

asturur (approved):

PR #7933, opened 2022-05-11
feat(invert_filter): Alpha support

GM-Alex:
This implements a option flag for the invert filter to also invert the alpha values by the filter.


asturur:
h @GM-Alex this pr is welcome.
2 things:
1) you shouldn't submit built dist folder when opening a pr.
2) you can do this keeping a single shader and adding an extra line there. Do you think you can do that?

I can try to make an example if is not clear.
We can send an extra uniform exactly as uInvert, called uAlpha and add an if in the code.
Let me know what you think, should be less code overall.

GM-Alex:
@asturur The dist folder was added by accident. 🙈 I know how the shader code should look like but since I'm not familiar with the shader logic I don't know how to pass the right values to the shader code.

asturur:
Let's do it together, if it looks nicer we keep it, or we revert back to this version.

asturur:
At the end of the file you see this 2 function to which yuo can add your parameter
```diff
    /**
     * Return WebGL uniform locations for this filter's shader.
     *
     * @param {WebGLRenderingContext} gl The GL canvas context used to compile this filter's shader.
     * @param {WebGLShaderProgram} program This filter's compiled shader program.
     */
    getUniformLocations: function(gl, program) {
      return {
        uInvert: gl.getUniformLocation(program, 'uInvert'),
+       uAlpha: gl.getUnifromLocation(program, 'uAlpha'),
      };
    },
    /**
     * Send data from this filter to its shader program's uniforms.
     *
     * @param {WebGLRenderingContext} gl The GL canvas context used to compile this filter's shader.
     * @param {Object} uniformLocations A map of string uniform names to WebGLUniformLocation objects
     */
    sendUniformData: function(gl, uniformLocations) {
      gl.uniform1i(uniformLocations.uInvert, this.invert);
+     gl.uniform1i(uniformLocations.uAlpha, this.alpha);
    },
```

This would bring the uAlpha in the shader availabile.

At that point you can just pull it off in the original shader

```diff
    fragmentSource: 'precision highp float;\n' +
      'uniform sampler2D uTexture;\n' +
+     'uniform int uAlpha;\n' +
      'uniform int uInvert;\n' +
      'varying vec2 vTexCoord;\n' +
      'void main() {\n' +
        'vec4 color = texture2D(uTexture, vTexCoord);\n' +
        'if (uInvert == 1) {\n' +
          'gl_FragColor = vec4(1.0 - color.r,1.0 -color.g,1.0 -color.b,color.a);\n' +
        '} else {\n' +
          'gl_FragColor = color;\n' +
        '}\n' +
      '}',
```
And then you can decide how to change the rest.

To revert the extra dist folder, just run `git checkout master dist`

GM-Alex:
@asturur Sorry took me some time to get my hands on this again. Could you have a look again.

src/filters/invert_filter.class.js
```diff
@@ -48,6 +60,20 @@
...
+     * @param {WebGLRenderingContext} options.context The GL context used for rendering.
+     * @param {Object} options.programCache A map of compiled shader programs, keyed by filter type.
+     */
+    retrieveShader: function(options) {
+      var cacheKey = this.type + '_' + (this.alpha ? 'alpha' : 'normal');
+      if (!options.programCache.hasOwnProperty(cacheKey)) {
+        options.programCache[cacheKey] = this.createProgram(options.context, this.fragmentSource);
+      }
+      return options.programCache[cacheKey];
+    },
```
asturur, line 75: We don't need this anymore now!
The shader is just one.

asturur:
i looks better imho! There is an unused function now that can be removed.

asturur (approved):

PR #7934, opened 2022-05-12
fix(): assigning canvas to collections

ShaMan123:
Collections could be used as clip paths.
This PR fixes canvas assignment in such a case.
Because Group assigns canvas only if the property changed this should not affect performance.


asturur:
The only thing we need to be careful of is determine what `_set` does and what `set` does.
Is fine to have one that is a bit higher level than another, important that is clear when to use one and another.

asturur (approved):

ShaMan123:
I assumed `set` is a simple wrapper that accepts objects to iterate over and pass down to `_set`.

PR #7941, opened 2022-05-14
test(freedrawing): test enhancement

ShaMan123:
This PR is a byproduct of the next PR that will add erasing visual tests.
I added logic to freedrawing tests to prepare for testing erasing and decided (with your voice guiding me :)) to split up the PR into dedicated changes.
This PR adds another layer of tests to free drawing by testing visuals before and after mouseup and comparing between them.
I think it's relevant and to the point.
It is proven by the tests failing (freedrawing4 - `Simple free drawing, with opacity`) - ~~you might want to output the mesh result to a different file to compare changes. It might be desirable in general so tell me if to commit a different golden name for the mesh test.~~ **Shadow + opacity cause this** it seems


ShaMan123:
@asturur please take a look at c6f33d2c4a6c1670107f2329e649002b1842429c
I reverted because so many tests began to fail... I am clueless as to why. I tried to refactor the code w/o changing logic to expose a function to compare to visual objects I want for freedrawing/erasing

ShaMan123:
Regarding Eraser tests...
If I add them I need to export to the website a build including the Eraser which makes a lot of unit tests fail because of `Object#erasable = true` by default...
What do you suggest?

test/visual/freedraw.js
```diff
@@ -27,178 +27,2225 @@
...
+    },
+    {
+      "x": 163.347,
+      "y": 24.06
+    },
+    {
+      "x": 164.343,
+      "y": 23.064
+    }
+  ];
```
asturur, line 2052: this pr seems all good, can you just keep this in a line for scrolling sanity only?
Even if we make a string and then we parse it each time.

ShaMan123: The IDE I use (VSCODE) allows me to collapse it with ease. That's one reason I changed it.
I prefer to extract it to a dedicated json file. Do you have any objections? It's better IMO. Once we migrate importing will be simple.

ShaMan123:
What about the failing test?

test/visual/freedraw.js
asturur: i have VSCODE too, but yes it can be on its own file, too, as every fixture usually is.

asturur:
Is the broken test a new test?
i see it has opacity and shadow, was the golden built with node?

asturur:
The website visual test got broken, probably the qunit update.
I need to fix that so i can see the difference in chrome/firefox

ShaMan123:
> i see it has opacity and shadow, was the golden built with node?

The test was built in node. It is a new test. It's the upper canvas test.

asturur:
We can fix the the display in browser, take the golden from there, and disable the node test once we understand how much the difference is. With shadow and opacity is mostly normal

ShaMan123:
I have just updated ruby so I can't serve the website as well.
Makes sense to disable the test on node as it is interactive visuals only.

ShaMan123:
I have managed to serve the website

expected is golden generated in node 
actual is generated in browser (chrome)
image

ShaMan123:
If I understand correctly node canvas doesn't apply opacity to the shadow...
What is the expected behavior?

asturur:
we can take the browser one that is most common case ( especially for free drawing ) and we can disable the test when we detect node with fabric.isLikelyNode and move forward.
Did you manage to fix the qunit display in the website?
I remember i worked on it really recently and i fixed it.

asturur:
image

i fixed it here, then:

image

i broke it there and i didn't verify it.

I saw we updated qunit in fabric and i thought it was a good idea to have the same version

asturur:
When hiding passed tests is still broken, anyway, there are 22 failing tests in 5.2.0 mostly on svg import, i wonder what happened

ShaMan123:
I don't see any of it...
Did you run the `start` cmd from fabric.js dir?


image

ShaMan123:
> image
> 
> i fixed it here, then:
> 
> image
> 
> i broke it there and i didn't verify it.
> 
> I saw we updated qunit in fabric and i thought it was a good idea to have the same version

Where is this? In the website? I have the website up and running locally

ShaMan123:
All set

asturur (approved):

ShaMan123:
I forget about the json...
I guess it will wait for es6

PR #7954, opened 2022-05-20
fix(): stop transforming on `discardActiveObject`

### LINKED ISSUE #7953: `discardActiveObject` is not prevent moving of object
When i start move object with mouse and call `discardActiveObject`, the active object are reset, but i still move object.
I can't stop moving object programmatically.

## Version
5.0.0

## Test Case
Browser: https://jsfiddle.net/aLuvj3hp/

## Steps To Reproduce

1. Start move red object
2. Wait 3 second, to selection will reset

## Expected Behavior

- Active object are reseted
- Moving of object are stopped

## Actual Behavior

- Active object are reseted
- User still can move object

ShaMan123:
Motivation 
closes #7953

### DIFF
--- a/src/canvas.class.js
+++ b/src/canvas.class.js
@@ -1229,7 +1229,7 @@
     /**
      * This is a private method for now.
      * This is supposed to be equivalent to discardActiveObject but without firing
-     * any events. There is commitment to have this stay this way.
+     * any selection events ( can still fire object transformation events ). There is commitment to have this stay this way.
      * This is the functional part of discardActiveObject.
      * @param {Event} [e] Event (passed along when firing "object:deselected")
      * @param {Object} object to set as active
@@ -1244,9 +1244,7 @@
           return false;
         }
         if (this._currentTransform && this._currentTransform.target === obj) {
-          this._finalizeCurrentTransform(e);
-          this._currentTransform = null;
-          obj.isMoving = false;
+          this.endCurrentTransform(e);
         }
         this._activeObject = null;
       }
--- a/src/mixins/canvas_events.mixin.js
+++ b/src/mixins/canvas_events.mixin.js
@@ -566,6 +566,22 @@
     },
 
     /**
+     * End the current transfrom.
+     * You don't usually need to call this method unless you are interupting a user initiated transform
+     * because of some other event ( a press of key combination, or something that block the user UX )
+     * @param {Event} [e] send the mouse event that generate the finalize down, so it can be used in the event
+     */
+    endCurrentTransform: function(e) {
+      var transform = this._currentTransform;
+      this._finalizeCurrentTransform(e);
+      if (transform && transform.target) {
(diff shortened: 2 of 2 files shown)


src/canvas.class.js
```diff
@@ -1243,6 +1243,10 @@
         if (obj.onDeselect({ e: e, object: object })) {
           return false;
         }
+        if (this._currentTransform && this._currentTransform.target === obj) {
```
ShaMan123, line 1246: pretty sure `this._currentTransform.target === obj` is useless and harmless

src/canvas.class.js
```diff
@@ -1243,6 +1243,11 @@
...
+          this._finalizeCurrentTransform(e);
+          this._currentTransform = null;
```
ShaMan123, line 1248: should we call `this._resetTransformEventData();`?

asturur:
I m ok with this PR but i would like there was a canvas method that says `(cancel/stop/end)CurrentTransform` and does what it needs to do to to cancel the current transform, and we use it on discardActiveObject

asturur:
i grouped the 2 calls in a method. i m ok merginging this.
The only doubt i have left and we can solve whenever is if that the discard of active objects belong to _discardActiveObject or discardActiveObject.

for example on object removed calls _discardActiveObject, and it could call _discardActiveObject and endCurrentTransform, instead of relying on the inner part, focused on discardind, now also fires an object modified event ( eventually ).

Is not a big deal, can probably be changed later.
Too much automatic behaviours usually brought to issues in the past

asturur (approved):

ShaMan123:
> for example on object removed calls _discardActiveObject, and it could call _discardActiveObject and endCurrentTransform, instead of relying on the inner part, focused on discardind, now also fires an object modified event ( eventually ).

Sounds a good point to me.

I am not sure but I think I walked into a bug with this PR somewhere...
Well, I'll try reproduce and PR a fix is so.

asturur:
is ok to merge this part as it is, is reasonably correct and help us moving forward

asturur:
is small, is focused, has a test, has docs, can be merged imho.

src/canvas.class.js
ShaMan123: @asturur this is not outdated
Thoughts?

ShaMan123: It's really `Canvas._target` , I think it should be reset but I'm not sure

ShaMan123: I have encountered the case that demands to nullify `_target`
When an object is being moved by the user and discarded meanwhile.
Currently the object will continue moving and will be inactive only on mouseup.

ShaMan123: Note for future self

setting `canvas._target` on `mousedown:before` hijacks the event and fires it on the hijacker

ShaMan123: #8335

PR #7961, opened 2022-05-22
fix(Point): safeguard initialization

ShaMan123:
This PR fixes Point so that the following
```js
new fabric.Point().scalarAdd(5)
```

will result in Point(5, 5) instead of Nan


melchiar (approved):

asturur (approved):
yes this makes sense. I would also say people have to read docs and not pass undefined in a point, but also is so little change that is fine by me

PR #7981, opened 2022-05-30
feat(Textbox): min/max width

### LINKED ISSUE #2376: Entering text in a Textbox object expands it when there aren't newlines

### LINKED ISSUE #5114: Support for word-breaking on Textbox
## Version
2.3.3

## Test Case
http://jsfiddle.net/fabricjs/Da7SP/

## Expected Behavior
Textbox's `maxWidth` didn't work in my case. Because I usally have long words.
I need word breaking on special characters too. ex) `-`,`/`,`_`

Is there a way to break a line on special characters too?
Or just word breaking is enough.

### LINKED ISSUE #5911: How can we set MaxWidth in fabric.js canvas Textbox ?
Hi , Need a small help. How can we set the MaxWidth and dynamic width while adding text using **fabric.Textbox** method.

If we can't use MaxWidth then can anyone help me with any alternative of this.

### LINKED ISSUE #7888: How to limit the maxWidth of textbox？
I have a textbox with determined width and height，want to limit the input text to not exceed the set width, so I will give an error prompt and give up the content entered this time.

I have tried whether I can monitor text changes, or size changes, and handle them in events, but I haven't found a suitable event.

https://codesandbox.io/s/still-haze-zys42f?file=/src/App.tsx

ShaMan123:
### EDITED

This PR supports `minWidth` `maxWidth`. 
I have refactored logic not to care if  `maxWidth < minWidth`. Values are not changed by fabric and everything works!
`maxWidth` will be overriden by `minWidth` and by `_actualMaxWidth` (min values + largest word) in case of conflict.

I have removed the redundant extend Observable code

### splitByGrapheme + maxWidth

https://user-images.githubusercontent.com/34343793/173128644-76ff3024-bdd8-4708-862a-4caf544de091.mp4

### maxWidth

https://user-images.githubusercontent.com/34343793/173128681-de8311c4-5c18-45bc-bf55-37db4f23d32b.mp4

### onInput + maxWidth
https://user-images.githubusercontent.com/34343793/173128702-0191fda3-e4c3-4844-8ea9-5154473bb2d9.mp4

closes #7888 
closes #7891
closes #5911
related #1184
closes #2376 
closes #5114
closes #1162


src/shapes/textbox.class.js
```diff
@@ -31,6 +31,14 @@
      */
     minWidth: 20,
 
+    /**
+     * Maximum width of textbox, in pixels.
+     * Use the `resizing` event to change this value on the fly
+     * @type {Number | undefined}
+     * @default
+     */
+    maxWidth: undefined,
```
asturur, line 40: i prefer to have Infinity rather than undefined, so we avoid type checking and type complications?

src/shapes/textbox.class.js
```diff
@@ -78,6 +86,33 @@
...
+     * @param {*} value
+     */
+    _set: function (key, value) {
+      if (key === 'width') {
+        if (typeof this.maxWidth === 'number' && this.maxWidth < value) {
+          value = this.maxWidth;
+        }
+        if (typeof this.minWidth === 'number' && this.minWidth > value) {
+          value = this.minWidth;
+        }
```
asturur, line 102: also infinity enables this:
```suggestion
value = Math.max(this.minWidth, Math.min(value, this.maxWidth))
```

src/shapes/textbox.class.js
asturur, line 40:
```suggestion
    maxWidth: Infinity,
```

src/shapes/textbox.class.js
```diff
@@ -78,6 +86,33 @@
...
+      if (key === 'maxWidth' && this.minWidth > value) {
+        value = this.minWidth;
+      }
+      this.callSuper('_set', key, value);
+      if (key === 'minWidth' && this.maxWidth < value) {
+        this._set('maxWidth', value);
+      }
+      if ((key === 'maxWidth' && this.width > value) || (key === 'minWidth' && this.width < value)) {
+        this._set('width', value);
+      }
```
asturur, line 113: mmmm not my favorite side effect.
This a new behavior, makes it breaking ( fine ) and should be written on top of minWidth and maxWidth.
Warning, setting this will also immediately affect width if condition apply.
Consider that setting this without re-rendering could be a recipe for disaster.
A valid alternative would be the opposite, if you are setting maxWidth or minWidth out of range, console warn and do nothing.
I think i prefer the do nothing option.

src/shapes/textbox.class.js
ShaMan123: I never worked with infinity. If it works... alright.

src/shapes/textbox.class.js
ShaMan123: I agree. It's better to let the dev set width as well.

src/shapes/textbox.class.js
```diff
@@ -78,6 +86,33 @@
...
+          value = this.minWidth;
+        }
+      }
+      if (key === 'maxWidth' && this.minWidth > value) {
+        value = this.minWidth;
+      }
+      this.callSuper('_set', key, value);
+      if (key === 'minWidth' && this.maxWidth < value) {
+        this._set('maxWidth', value);
+      }
```
asturur, line 105: this is:
If you want to set maxWidth below minWidth i m going to stop you to minWidth.
If you set up minWidth over maxWidth that is allowed, but  then i will rise maxWidth accordingly.

I see how minWidth should be the boss to avoid creating zero width objects and mess up when the user is driving the action. But this needs to be described too in the properties, let's not have the developers discover it by trial and error.

src/shapes/textbox.class.js
```diff
@@ -349,7 +380,10 @@
         offset += word.length + 1;
         return { word: word, width: width };
       }.bind(this));
-      var maxWidth = Math.max(desiredWidth, largestWordWidth, this.dynamicMinWidth);
+      var maxWidth = Math.max(desiredWidth, largestWordWidth, this.getMinWidth());
+      if (typeof this.maxWidth === 'number' && maxWidth > this.maxWidth) {
```
asturur, line 384: we can remove this extra typeof too.

asturur:
This PR is good! But if you had described the behaviour in the PR description instead of saying `properly supports`, the review would have been super fast.

asturur:
So once you set minWidth and maxwidth to, for example, 300, and you have a word that is long 360, what happens to wrapping?

ShaMan123:
> There is more work to be done in order to adapt onInput functionality to support min/max width
I have a commit that almost does it. I'll push when I'm more satisfied

I wanted to revisit it.... but I should have.
I remember there's a bug regarding the input and I lost the stash of the branch :(
I'll open PR

src/shapes/textbox.class.js
```diff
@@ -78,6 +86,28 @@
      */
     splitByGrapheme: false,
 
+    /**
+     * While editing handle differently
```
asturur, line 90: this comment makes little sense.
If it means `this logic doesn't apply while editing because ...` let's write that

src/shapes/textbox.class.js
ShaMan123: We can warn as well...
What do you prefer?

ShaMan123:
But I see logic doesn't support that.
I'll take a closer look.

ShaMan123:
@asturur I am confused and need help in defining how it should work.
I think that once the largest word is measured it should override maxWidth. Am I correct?

src/shapes/textbox.class.js
ShaMan123: resolved

src/shapes/textbox.class.js
ShaMan123: done

ShaMan123 (commented):
I have edited the PR description
Firefox tests seem to be broken with no reason.
I use a linux distro now and it takes a decade to load the website so I can't tell if the tests pass in the browser but they do on node.
Besides that - Ready to merge.

asturur:
i m reading the linked issues:

https://github.com/fabricjs/fabric.js/issues/7888
https://github.com/fabricjs/fabric.js/discussions/7891
Wants to limit text input

https://github.com/fabricjs/fabric.js/issues/5911
Wants to avoid the textbox going out of canvas when inputing text

https://github.com/fabricjs/fabric.js/issues/5114
Wants to wrap text when reach a certain width becase of long words

It seems that maxWidth should indeed break the words.
so while starting to implement it is fine, i don't think it closes the related issues.

ShaMan123:
If you set `splitByGrapheme` to true it works.
I set it after initialization so it starts with words and then when I change it it splits graphemes
Word breaking can be achieved by subclassing I believe

lachezarR:
Hello, when this is ready do you plan to apply it to version ^4.5.7, or will it be available only for latest?

ShaMan123:
latest v6

ShaMan123:
F**** this PR when trying to port it to v6
#8470 will take over.

PR #7991, opened 2022-06-06
fix(EraserBrush): visual trace while erasing

### LINKED ISSUE #7984: Stroke appearing while erasing in free drawing mode
## Version
5.2.1

## Test Case
- [ ] [Browser Issue Template](https://jsfiddle.net/Lcp2h3nv/)
- [ ] [Node Issue Template](https://codesandbox.io/s/exciting-browser-ytb701)

Nodejs or browser?
Which browsers?
Tested in following browsers
- Chrome for windows version 101
- Chrome for Android  Version 102
- Firefox Developer Edition for Windows Version 101

## Steps To Reproduce

1.  Create a canvas with free Drawing option enabled
2. Set Eraser Brush as canvas.freeDrawingBrush 

<details><summary>Error Message & Stack Trace</summary><p>

```txt
<!-- Provide a log message if relevant -->
```
</p></details>

## Expected Behavior
Eraser Should clear the brush path without any strokes / border 
## Actual Behavior
The eraser path has a tiny white border while erasing
can be reproduced in the fabric eraser demo: [Eraser Demo](http://fabricjs.com/erasing)

ShaMan123:
A **hack** that reduces the path width while clipping the main context, resulting in a better visual overlap of both contexts
The issue's cause is unknown at time of writing - I don't understand why this is happening.

Added another minor change of the masking logic. Instead of source-over I reversed the order to use destination-in. Shouldn't affect anything and is simpler to grasp.

closes #7984

### DIFF
--- a/src/mixins/eraser_brush.mixin.js
+++ b/src/mixins/eraser_brush.mixin.js
@@ -495,7 +495,7 @@
       _saveAndTransform: function (ctx) {
         this.callSuper('_saveAndTransform', ctx);
         this._setBrushStyles(ctx);
-        ctx.globalCompositeOperation = ctx === this.canvas.getContext() ? 'destination-out' : 'source-over';
+        ctx.globalCompositeOperation = ctx === this.canvas.getContext() ? 'destination-out' : 'destination-in';
       },
 
       /**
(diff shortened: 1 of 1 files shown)


ShaMan123 (commented):
This is ready.
Anyways it is not breaking because I've exposed the prop `erasingWidthAliasing` that can be set to 0 to revert to previous behavior.

asturur (approved):

PR #7994, opened 2022-06-07
feat(observable): BREAKING return disposer instead of context for chaining

ShaMan123:
This PR makes observable easier to use by changing it's methods return values:
- return disposer for on/once
- void for fire/off

This means this PR is **BREAKING**. 
As we discussed chainable objects are overrated and IMO are bad design because then a simple change such as this proposal becomes breaking.

So instead of:
```js
const cb = () => {
 ...
}
obj.on('event', cb)
const dispose = () => obj.off('event', cb);
...

dispose()

```

This is possible

```js
const dispose = obj.on('event', () => {
  ...
})
...

dispose()
```

For the dev (me for example) it saves a lot of repetitive code. It makes things easier.

### DIFF
--- a/src/mixins/observable.mixin.js
+++ b/src/mixins/observable.mixin.js
@@ -24,8 +24,7 @@
    * @alias on
    * @param {String|Object} eventName Event name (eg. 'after:render') or object with key/value pairs (eg. {'after:render': handler, 'selection:cleared': handler})
    * @param {Function} handler Function that receives a notification when an event of the specified type occurs
-   * @return {Self} thisArg
-   * @chainable
+   * @return {Function} disposer
    */
   function on(eventName, handler) {
     if (!this.__eventListeners) {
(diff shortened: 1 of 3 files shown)


src/mixins/observable.mixin.js
```diff
@@ -89,33 +97,30 @@
       }
     }
     // one object with key/value pairs was passed
-    else if (arguments.length === 1 && typeof arguments[0] === 'object') {
+    else if (typeof eventName === 'object' && typeof handler === 'undefined') {
```
asturur, line 100: is this better in anyway or you just like it more?

ShaMan123: `arguments.length === 1` is wrong because if you pass undefined as the second arg in will return false

ShaMan123:
> More than overrated chaining, the issue is that once you opt for chaining, you can't return anything anymore, and so you cut away a bunch of comfortable patterns. Having 10% of method chainable is not really an advantage then.
We should make clear somewhere that chaining is a leftover and developers shouldn't count on it

YES!!

asturur (approved):

PR #8000, opened 2022-06-10
fix(Group): patch #7916

ShaMan123:
Hopefully this is all we missed out in #7916 .
Resolves https://github.com/fabricjs/fabric.js/pull/7916#pullrequestreview-1002987203


ShaMan123 (commented):
Ready

ShaMan123 (commented):
updated from master
READY TO MERGE!

src/static_canvas.class.js
```diff
@@ -625,9 +625,9 @@
      * @param {fabric.Object} obj Object that was removed
      */
     _onObjectRemoved: function(obj) {
+      obj._set('canvas', undefined);
       this.fire('object:removed', { target: obj });
       obj.fire('removed', { target: this });
-      obj._set('canvas', undefined);
```
asturur, line 630: is there any bug with removing the canvas before or after?
We could pass a reference to the canvas in the event itself as oldCanvas?
In general what are the reason for this line swap?

ShaMan123: https://github.com/fabricjs/fabric.js/pull/7859#issuecomment-1109377515

In terms of design and naming it's `removed` so I think canvas should not be referenced anymore once the event triggers (this is how group behaves) so it is consistent across fabric

ShaMan123: And it is passed as the target

asturur: oh right this is the canvas not the object. i didn't think of it.

asturur (approved):

PR #8001, opened 2022-06-12
chore(): move away from extend/clone

ShaMan123:
A semantic PR that replaces `fabric.util.object.extend` or the shallow version of `fabric.util.object.clone` with `Object.assign` to prepare for the js/ts migration.
I replaced relevant code after using the transformer (It uses extend as a search value).
The deep version of clone is used sparsely and in the future could be replaced by [`lodash.cloneDeep`](https://www.npmjs.com/package/lodash.clonedeep) (~44KB vs. ~4KB of `lang_object` 😬). BTW I'm not sure it is really necessary in most cases.

### DIFF
--- a/src/controls.actions.js
+++ b/src/controls.actions.js
@@ -32,10 +32,8 @@
 
   function fireEvent(eventName, options) {
     var target = options.transform.target,
-        canvas = target.canvas,
-        canvasOptions = fabric.util.object.clone(options);
-    canvasOptions.target = target;
-    canvas && canvas.fire('object:' + eventName, canvasOptions);
+        canvas = target.canvas;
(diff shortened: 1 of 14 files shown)


ShaMan123 (commented):
Take a close look at `src/gradient.class.js` tom make sure the cloning is really redundant

asturur (approved):

asturur:
this is fine. I'll merge it now.
Let's try to not open more PRs before we can migrate.

ShaMan123:
Agreed.
I opened this to make the migration easier. Less complication

בתאריך יום א׳, 12 ביוני 2022, 15:00, מאת Andrea Bogazzi ‏<
***@***.***>:

PR #8004, opened 2022-06-13
fix(Group): 🛠️ layout, angle and origin ⚡

ShaMan123:
This PR replaces the polluted #7867 

Group now respects originX/Y + angle/skewing!
It is a minor fix (to the layout engine) but a significant one.
I have committed all binaries in a single commit so you can view the diff without them.

There are some conflicts with #7860 so we better merge that first (because conflicts are **resolved** there) or perhaps extract group logic and merge that.


asturur:
Those tests are useful, but we need them to have less transparent pixels if we want the detection to work. We need either to scale up the group and have it fill up more image

test/visual/group_layout.js
```diff
@@ -323,37 +359,39 @@
...
+                        canvas.add.apply(canvas, createObjectsForOriginTests(ox, oy, { angle }));
+                        canvas.setViewportTransform([1, 0, 0, 1, -50, 0]);
                         canvas.renderAll();
                         callback(canvas.lowerCanvasEl);
                     },
                     golden: `group-layout/origin-${ox}-${oy}-${angle}deg.png`,
-                    percentage: 0.06,
-                    width: 300,
-                    height: 300
+                    percentage: 0.001,
```
ShaMan123, line 388: I calculated this percent to be exactly the red/blue origin indicators.
Now after resizing it will be even more sensitive. Which is good.

READY TO MERGE!

ShaMan123 (commented):
READY!

asturur:
i wanted to make the drawing bigger, so that the eyes could catch easily what is wrong ( if any ) and we don't have to rely on few pixel differences of different aliasing.
But i m sure that if i look at the test i can do that easily later

ShaMan123:
for sure.

PR #8006, opened 2022-06-16
feat(Text): condensed styles structure v6

melchiar:
Styles structure rewrite for v6 (ported from https://github.com/fabricjs/fabric.js/pull/7842)

### DIFF
--- a/src/shapes/itext.class.js
+++ b/src/shapes/itext.class.js
@@ -531,6 +531,7 @@
    * @returns {Promise<fabric.IText>}
    */
   fabric.IText.fromObject = function(object) {
+    object.styles = fabric.util.stylesFromArray(object.styles, object.text);
     return fabric.Object._fromObject(fabric.IText, object, 'text');
   };
 })();
--- a/test/unit/itext.js
+++ b/test/unit/itext.js
@@ -44,7 +44,7 @@
     skewX:                    0,
     skewY:                    0,
     charSpacing:              0,
-    styles:                   { },
+    styles:                   [],
     strokeUniform:            false,
     path:                     null,
     direction:                'ltr',
@@ -105,7 +105,10 @@
       assert.ok(typeof fabric.IText.fromObject === 'function');
       fabric.IText.fromObject(ITEXT_OBJECT).then(function(iText) {
         assert.ok(iText instanceof fabric.IText);
-        assert.deepEqual(ITEXT_OBJECT, iText.toObject());
+        // change styles from array to object for comparison
+        var object = iText.toObject();
+            object.styles = {};
+        assert.deepEqual(ITEXT_OBJECT, object);
         done();
       });
     });
@@ -131,22 +134,38 @@
     });
 
     QUnit.test('toObject', function(assert) {
-      var styles = {
+      var stylesObject = {
         0: {
           0: { fill: 'red' },
           1: { textDecoration: 'underline' }
         }
       };
+      var stylesArray = [
+        {
+          start: 0,
+          end: 1,
+          style: { fill: 'red' }
+        },
+        {
+          start: 1,
+          end: 2,
+          style: { textDecoration: 'underline' }
+        }
+      ];
       var iText = new fabric.IText('test', {
-        styles: styles
+        styles: stylesObject
       });
       assert.equal(typeof iText.toObject, 'function');
       var obj = iText.toObject();
-      assert.deepEqual(obj.styles, styles);
-      assert.notEqual(obj.styles[0], styles[0]);
-      assert.notEqual(obj.styles[0][1], styles[0][1]);
-      assert.deepEqual(obj.styles[0], styles[0]);
-      assert.deepEqual(obj.styles[0][1], styles[0][1]);
+      assert.deepEqual(obj.styles, stylesArray);
+      assert.notEqual(obj.styles[0], stylesArray[0]);
+      assert.notEqual(obj.styles[1], stylesArray[1]);
+      assert.notEqual(obj.styles[0].style, stylesArray[0].style);
+      assert.notEqual(obj.styles[1].style, stylesArray[1].style);
+      assert.deepEqual(obj.styles[0], stylesArray[0]);
+      assert.deepEqual(obj.styles[1], stylesArray[1]);
+      assert.deepEqual(obj.styles[0].style, stylesArray[0].style);
+      assert.deepEqual(obj.styles[1].style, stylesArray[1].style);
     });
 
     QUnit.test('setSelectionStart', function(assert) {
--- a/test/unit/text.js
+++ b/test/unit/text.js
@@ -51,7 +51,7 @@
     skewX:                      0,
     skewY:                      0,
     charSpacing:                0,
-    styles:                     {},
+    styles:                     [],
     path:                       null,
     strokeUniform:              false,
     direction:                  'ltr',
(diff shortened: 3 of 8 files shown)


test/unit/itext.js
```diff
@@ -105,7 +105,10 @@
       assert.ok(typeof fabric.IText.fromObject === 'function');
       fabric.IText.fromObject(ITEXT_OBJECT).then(function(iText) {
         assert.ok(iText instanceof fabric.IText);
-        assert.deepEqual(ITEXT_OBJECT, iText.toObject());
+        // change styles from array to object for comparison
+        var object = iText.toObject();
+            object.styles = {};
+        assert.deepEqual(ITEXT_OBJECT, object);
```
ShaMan123, line 111: Very minor and insignificant... but if you want
```suggestion
        assert.deepEqual(object, ITEXT_OBJECT);
```

ShaMan123 (approved):
Looks good

test/unit/itext.js
melchiar: 👍

ShaMan123:
Why are unit tests failing?

melchiar:
> Why are unit tests failing?

Just fixed the unit/textbox.js test. Don't know why the firefox visual test is failing but it seems unrelated to this pr.

image

ShaMan123:
> > Why are unit tests failing?
> 
> Just fixed the unit/textbox.js test. Don't know why the firefox visual test is failing but it seems unrelated to this pr.
> 
> image

It does that all the time. The unit tests were what I intended to ask about.

ShaMan123 (approved):

ShaMan123:
conflicts @melchiar

ShaMan123:
simple conflicts as I see it. All should be resolved to the master tag

melchiar:
> simple conflicts as I see it. All should be resolved to the master tag

should be good now

ShaMan123 (changes requested):
I think we should use Object.assign in all fromObject methods so we don't mutate the arg

melchiar:
> I think we should use Object.assign in all fromObject methods so we don't mutate the arg

Object.assign is already used in the `stylesFromArray` method though (line 1300 in `src/util/misc.js`). Where exactly are you seeing a risk of mutation?

src/shapes/itext.class.js
```diff
@@ -531,6 +531,7 @@
    * @returns {Promise<fabric.IText>}
    */
   fabric.IText.fromObject = function(object) {
+    object.styles = fabric.util.stylesFromArray(object.styles, object.text);
```
ShaMan123, line 534: Here for example

src/shapes/itext.class.js
```diff
@@ -531,6 +531,7 @@
...
     return fabric.Object._fromObject(fabric.IText, object, { extraParam: 'text' });
```
ShaMan123, line 535:
```suggestion
    return fabric.Object._fromObject(fabric.IText, Object.assign({}, object, {style: style}), { extraParam: 'text' });
```

src/shapes/itext.class.js
ShaMan123:
```suggestion
    var styles = fabric.util.stylesFromArray(object.styles, object.text);
```

src/shapes/itext.class.js
melchiar: what I mean is the object returned by `stylesFromArray` is already composed of copied properties, so shouldn't it already be safe from mutation?

ShaMan123: I'm talking about the object passed to the fromObject function.
Test a case which you create a text object from an object and use the same object to create another text object.
It should break if you use styles, shouldn't it? The object will be mutated

ShaMan123: You can/should test it by comparing the styles property of the passed object before and after you initialized the text object

melchiar: dangit, you're right 😆

src/shapes/itext.class.js
```diff
@@ -531,6 +531,9 @@
    * @returns {Promise<fabric.IText>}
    */
   fabric.IText.fromObject = function(object) {
-    return fabric.Object._fromObject(fabric.IText, object, { extraParam: 'text' });
+    var styles = fabric.util.stylesFromArray(object.styles, object.text);
+    //copy object to prevent mutation
+    var objCopy = Object.assign({}, object, { styles: styles });
```
melchiar, line 536: looks like some more tests are breaking with this change 😑

ShaMan123 (commented):
@melchiar I do not understand the tests... Something seems off

test/unit/text.js
```diff
@@ -207,9 +210,12 @@
         width: 8,
         height: 18.08,
         fontSize: 16,
-        originX: 'left'
+        originX: 'left',
       });
-      assert.deepEqual(text.toObject(), expectedObject, 'parsed object is what expected');
+      var object = text.toObject();
+      // change styles from array to object for comparison
+      object.styles = expectedObject.styles = {};
```
asturur, line 217: i don't understand what we are doing here. is deepEqual the issue?

asturur:
i think we need to solve the mistery of line 193 and 217 in unit/text.js
those assignments shouldn't be there.
Weren't needed before, shouldn't be needed now

melchiar:
> i think we need to solve the mistery of line 193 and 217 in unit/text.js those assignments shouldn't be there. Weren't needed before, shouldn't be needed now

Should be good now. Turns out that a unit test issue I had tried to "fix" was actually related to the mutation bug that shachar noticed.

ShaMan123 (commented):
Looks good.
I would like to ask a final question from a previous discussion but I think it's important.

```js
let data// has styles array set
fabric.Text.fromObject(data) // respects styles
new fabric.Text(data) // doesn't respect styles

```

Am I right? I don't like that. I think it should respect both use cases

asturur:
I don't think that is a real issue, exactly as you can't instantiate an obejct with a serialized pattern, gradient or element.
The serailization has a shape and the object options/data a different one.
FromObject is there to fill the gap.

You don't exepct an js object to be initialized with its JSON.stringify form without passing into JSON.parse, and this is exactly the same

asturur (approved):

ShaMan123:
Well I still dislike it.
Fabric has too many unintuitive functions you can call only in a certain way that beg to be called in other contexts.
It makes for bad DX and a feel that the repo is broken. Though it is not.
Bad PR for fabric IMO - to many dos and don'ts and cases and allowed combinations and disallowed combinations. e.g. you can't set path on path directly but if you do you must call `_setPositionDimesions`. Now styles is part of this notorious gang.

melchiar:
I don't really have a strong opinion on it either way. 

I agree we should be trying to make the dev experience as intuitive as possible and not force a heavy reliance on docs to show the correct way to call each method. I also feel though that methods like `new fabric.Text` are better suited to creating basic objects that inherit an app's default properties, while `fromObject` is the better and more logical method to use when deserializing previously created objects and already have complete properties. If we're going to try to make both approaches equally flexible then we would also need to fix the other things lacking in the `new fabric.` method that Andrea mentioned.

asturur:
@ShaMan123 is an issue with serialization, toObject and toJSON are called when you have to convert objects to strings to store them somewhere.
Those are strings now and not functional pieces of code.
Is the same for gradients, patterns, and rects.
You can't add to canvas a rect that has been serialized right? and that feels normal.

If you call toObject on something, to use it again you have to call fromObject.

ShaMan123:
I understand your point but I still think it's too fragile.
But let's move forward

asturur:
To move forward please dismiss your request for changes that have been addressed.
If you don't want to expliitly approve that is fine, but at least remove the change requested flag

ShaMan123 (approved):

PR #8013, opened 2022-06-19
ci(): switch the old custom build for rollup

asturur:
This test works, but while the code changes seems large and aren't ( all whitespace and mandatory function wrapping ),
this change itself doesn't give us much.

I made this to have a small step transition that could allow tracing back issues to particular commits, and i would like the classes to transition one at one to be able to use git commit meaningfully.

If from this point modules can be migrated one by one ( starting from leaf nodes going up to the basic one ) ok then we can merge it, otherwise is pointless.

I ll experiment more and check

To look at PR:
https://github.com/fabricjs/fabric.js/pull/8013/files?diff=unified&w=1

most of the files will have changes like this.
image


asturur:
ok browser import is broken

ShaMan123:
@asturur why don't you want to use #7839 ?

asturur:
i m trying to add rollup without changing much else.
We can then use the transforme to chagne from the actual code to new code.

asturur:
i ll make a branch from here to test partial swapping of a single file or function to a different export method. But seems to work.

ShaMan123:
I have tested the new build.
My apps works with it without a problem.
Good job!

asturur:
in a different branch i will convert a single function of the misc utils to TS, and see if that + ts + rollup-ts-whatever works
if it works it means this combination is sustainable

ShaMan123:
I checked the old build size with and without svg and found
| file | svg  | no svg | diff
| ------------- | ------------- | ------------- | ------------- |
| fabric.js | 1039 KB  | 993 KB | 46 KB
| fabric.min.js  | 312 KB | 296 KB | 16 KB

Is it a big enough difference to support svg as an opt in?

asturur:
In theory we can move all the svg related functions in a different file and developers will import them when needed.
Hopefully the big fabric.* objects goes in disuse quickly.

of course 16kb is not that big, but if you are not doing anything with svg import and export, maybe you don't want to load that code anyway.

When everything is on import/export it should be very easy to do tests.

asturur:
ok here we are

ShaMan123:
If you need help with resolving conflict I know what to do with:
- scripts/index.js
- src/mixins/canvas_gestures.mixin.js

asturur:
ok now the PR is green, i won't touch it.
Let's see if there is some large PR that we don't want to conflict because is too complex to fix, and then we can merge this.

ShaMan123:
I gave a look at my open PRs and I marked 3 huge ones:
- #7802 
- #7827 
- ~#7719~ **low priority**

asturur:
I re read the draggable text pr, and while i m sure is a good feature we want to add, i currently don't  like how we attach it to the events, i would like to write down all the pr is doing and see if we can break the code differently.  Let me look the promises pr and let's get that in, then let's scehdule some time for the draggable text after i m sure i can expose my point of view and comunicate it properly.

test/unit/canvas.js
```diff
@@ -105,7 +105,6 @@
    */
   QUnit.assert.sameImageObject = function (actual, expected) {
     var a = {}, b = {};
-    expected = expected || REFERENCE_IMG_OBJECT;
```
ShaMan123, line 108: ?

test/unit/group.js
```diff
@@ -889,7 +889,7 @@
 
     group1.addRelativeToGroup(rect5);
     var t = group1.calcTransformMatrix();
-    var pos = fabric.util.transformPoint(new fabric.Point(rect5.left, rect5.top), t);
```
ShaMan123, line 892: I guess I wanted to check something here and didn't complete

src/util/dom_style.js
```diff
@@ -35,7 +36,7 @@
   var parseEl = fabric.document.createElement('div'),
       supportsOpacity = typeof parseEl.style.opacity === 'string',
       supportsFilters = typeof parseEl.style.filter === 'string',
-      reOpacity = /alpha\s*\(\s*opacity\s*=\s*([^\)]+)\)/,
```
ShaMan123, line 38: are you sure about this?

src/typedefs.ts
```diff
@@ -0,0 +1,11 @@
+interface NominalTag<T> {
+  'nominalTag': T;
```
ShaMan123, line 2:
```suggestion
  'nominalTag'?: T;
```

https://github.com/fabricjs/fabric.js/pull/8019#discussion_r903596787

ShaMan123 (commented):
Looks good

PR #8019, opened 2022-06-22
Adding TS to rollup

asturur:
Didn't test yet, just push up for visibility


rollup.config.js
```diff
@@ -16,4 +17,9 @@ export default {
         plugins: [terser()],
     },
   ],
+  plugins: [
+		ts({
```
ShaMan123, line 21: Is this the place to add tsconfig? or does the plugin read it under the hood

ShaMan123:
@asturur you should write about this in the tracker issue

src/typedefs.ts
```diff
@@ -0,0 +1,11 @@
...
+  'nominalTag': T;
+}
+
+type Nominal<Type, Tag> = NominalTag<Tag> & Type;
+
+const enum Degree {}
+const enum Radian {}
+
+export type TDegree = Nominal<number, Degree>;
+export type TRadian = Nominal<number, Radian>;
```
asturur, line 11: This is soething i have seen doing to other devs, is like a away to have 2 numbers that do not mix together.
This is good for radians vs degree and little else probably in our case.

src/util/cos.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const cos = (angle: TRadian): number => {
+  if (angle === 0) { return 1; }
+  var angleSlice = Math.abs(angle) / halfPI;
+  switch (angleSlice) {
+    case 1: case 3: return 0;
+    case 2: return -1;
+  }
+  return Math.cos(angle);
+};
```
asturur, line 21: i like the idea of one function per file and a file of constants.
I m also fine with a file grouping for example cos, sin, degreeToRadians and radiansToDegree, but not much else.

But since rollup has zero footprint on file numbers for me a function per file gives me  easy accessibility to code, i can find the function both in the search and in the file search and i don't have to scroll files of thousands of lines searching the utils.

src/typedefs.ts
ShaMan123: I like this

src/util/cos.ts
ShaMan123: I don't mind if a file for each of grouping.
We should then put cos, sin ,rad/deg utils into a seprate folder IMO called `trig` or something

src/typedefs.ts
ShaMan123: If I add an optional mark to the `nominalTag` https://github.com/fabricjs/fabric.js/commit/304767d210aef7c3c9ebcbac051d9258a0394c6a it accepts `number` or `TRadian` but not `TDegree`:
image

image

asturur: oh that would be fine, important is that they don't mix when they come out of a function that specifically outputs either radians or degree, i'll make that change.

Typescript should have this partitioned number somehow, even if is not a specific TS use case, but is a nice to have functionality.

src/util/cos.ts
asturur: For top leveI functions is just an aestetic preference, i m not sure if there is a meaningful difference in functionality or speed. If there was i think it would be well chatted on the internet.

src/typedefs.ts
ShaMan123: very nice to have

asturur:
ok this works, tests pass, i add it to the general build process pr.
Next is eslint update

PR #8024, opened 2022-06-25
ci(): Add linting

asturur:
- update eslint to latest to support the typescript
- create a new command and configuration to lint ts files in order to slowly eliminate the old one
- pickup the default configuration ( that we probably will going to slighty change depending on preferences )
- fix the necessary changes in the test files in order for the new lint to pass


src/util/cos.ts
```diff
@@ -12,7 +12,7 @@ import { halfPI } from '../constants';
  */
 export const cos = (angle: TRadian): number => {
   if (angle === 0) { return 1; }
-  var angleSlice = Math.abs(angle) / halfPI;
+  const angleSlice = Math.abs(angle) / halfPI;
```
asturur, line 15: first lint request, at leat  it works.

src/util/dom_style.js
```diff
@@ -1,5 +1,6 @@
 (function (global) {
   var fabric = global.fabric;
+  /* this file needs to go away, cross browser style support is not fabricjs domain.
```
asturur, line 3: i didn't know we had this, but we need to remove it. Is probably for old days.
Safari eventuall doesn't support yet opacity in stile, meaning that opacity in style shouldn't be used imho.

src/util/dom_style.js
```diff
@@ -35,7 +36,7 @@
   var parseEl = fabric.document.createElement('div'),
       supportsOpacity = typeof parseEl.style.opacity === 'string',
       supportsFilters = typeof parseEl.style.filter === 'string',
-      reOpacity = /alpha\s*\(\s*opacity\s*=\s*([^\)]+)\)/,
+      reOpacity = /alpha\s*\(\s*opacity\s*=\s*([^)]+)\)/,
```
asturur, line 39: another lint caught thing

asturur:
@ShaMan123 @melchiar have a look if you see something noteworthy.
In theory old lint does not change ( no changes to the code indeed apart a bug that i committed on its own ), new lint is whatever it is, test lint was and is still a mess.

src/util/cos.ts
ShaMan123: fabulous

src/util/dom_style.js
ShaMan123: I am shocked that opacity isn't supported... WHY?
Well, how does it affect fabric?

src/util/dom_style.js
ShaMan123: It seems that linting now works for us

ShaMan123 (approved):
I gave this a quick look.
Seems fine. I didn't drill down to every change and regarding rules - I'm not very knowledgeable and I understand we'll get there in the near future

.eslintrc_tests
```diff
@@ -10,6 +10,18 @@
     "pixelmatch": true
```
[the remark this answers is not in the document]
asturur: it was setup like this ages ago, in theory we want to add to the normal config a bunch of globals for qunit and nothing else, but that requires extensive reformatting of tests files.
But yes we want extend the tests config from the source configuration

src/util/dom_style.js
asturur: it doesn't. I think that if you pass a style object to the canvas, we transfer it on the container using this setStyle function.
And if for some weird reason you want your canvas with the opacity style property.
But really is not our problem to fix.

src/util/dom_style.js
asturur: it was working even before, but our lint config lacked lint-recommended plugin, and had just the old formatting rules.

Issue #8030, opened 2022-06-28
Scaling up very small image behaves differently in browser vs Node


savy:
## Version
Fabric 5.0.0
Node 18.3.0
Chrome Version 103.0.5060.53 (Official Build) (x86_64)
Macbook Pro, Big Sur, Retina Display

## Test Case
- [X] [Browser Issue Template](https://jsfiddle.net/5qc6dvgx/)
- [X] [Node Issue Template](https://codesandbox.io/s/winter-currying-0nrsl4)

## Information about environment
Nodejs or browser? Issue is in Node
Which browsers? Chrome Version 103.0.5060.53 (Official Build) (x86_64)

## Steps To Reproduce

1. Create 1x1 solid color png
2. Use Image.fromURL() to load image onto canvas. In the test code, I converted the png to base64 format to isolate the example, however the issue occurs if you retrieve the image file via http as well.
3. Generate png of the canvas using toDataURL() on Node

## Expected Behavior
I expect both the browser and Node to generate the same solid image like below:
![download (15)](https://user-images.githubusercontent.com/749574/176298794-4f3207a4-185e-4abb-9071-aa81849c8a89.png)

## Actual Behavior
Chrome generates the expected solid image, but Node generates the following image:
![download (14)](https://user-images.githubusercontent.com/749574/176299294-e7ac1573-f07b-42f1-b852-102bdc2c5782.png)

I tried various combinations of enabling/disabling imageSmoothing, objectCaching, and enableRetinaScaling, since I know some defaults are different on Node, but I wasn't successful at getting Node to output the same image as the browser.

I also investigated to see if this was due to Fabric using node-canvas on Node, however if I make a similar example directly via node-canvas, the output is a solid color as expected [node-canvas example](https://codesandbox.io/s/exciting-shirley-eilvn8?file=/src/index.js)

ShaMan123:
[node-canvas example](https://codesandbox.io/s/awesome-perlman-lyf2cb?file=/package.json) - fixed deps.

Excellent issue!

ShaMan123:
![image](https://user-images.githubusercontent.com/34343793/176311763-a407da15-f85c-4afa-9486-1dce4205f1a0.png)

I have managed to reproduce on node-canvas with the following 
https://codesandbox.io/s/gifted-perlman-jw6o4h

ShaMan123:
It is a strange bug 
Try 
```diff
-  ctx.drawImage(image, 0, 0, 1, 1);
[...]

PR #8032, opened 2022-06-29
fix(node-canvas): imageSmoothing setter

ShaMan123:
fixes https://github.com/fabricjs/fabric.js/issues/8030
override default `patternQuality` from `good` to `best` for node canvas
https://github.com/Automattic/node-canvas/blob/0bccc05b384cc1f3e969cd9ed7859a95376ec88c/src/CanvasRenderingContext2d.cc#L1479

Should we expose a way to control `patternQuality`? Should it be set on if imageSmoothing is enabled?

Seems this prop doesn't exist on the browser. The only thing I found that comes near is https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/imageSmoothingQuality


src/static_canvas.class.js
```diff
@@ -755,6 +755,7 @@
       this.calcViewportBoundaries();
       this.clearContext(ctx);
       fabric.util.setImageSmoothing(ctx, this.imageSmoothingEnabled);
+      ctx.patternQuality = 'best';
```
ShaMan123, line 758: Maybe fast as default?

ShaMan123: fast isn't good enough
look at the diff between `best` and `fast`
image

ShaMan123: objectsInActiveSelections


objectsInActiveSelections

asturur:
is a node-canvas non standard property, that control some cairo parameters.
I m fine leaving it in a way that satisfy most expectations, but i don't want to add logic around that ( when to set best or fast )

asturur (approved):

PR #8034, opened 2022-06-29
fix(Group): remove added object from canvas

ShaMan123:
This PR safeguards `group.add` from a case which the added object belongs to a canvas and vice versa

ActiveSelection should conform to this logic IMO, though currently this bug is prevented because the selected objects are filtered out from the rendered objects in the render cycle.

~Meaning a case in which an object is shared between 2 canvases and is selected by one is a hole in the logic. I can code something to patch this but I don't think it's needed 'cause were talking about a super edge case (which I think states the dev is handling heavy customization so tough luck for them)~ **EDITED** OK that's wrong. An object won't be selected if it doesn't belong to canvas or a group in it  (only programmatically) . So this isn't really a case


ShaMan123:
It's very nice that tests simply pass (without firefox failing)

src/static_canvas.class.ts
```diff
@@ -601,21 +620,26 @@ import { pick } from './util/misc/pick';
...
-      obj._set('canvas', this);
-      obj.setCoords();
-      this.fire('object:added', { target: obj });
-      obj.fire('added', { target: this });
+      else if (object.canvas) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Canvas: removing object from canvas before entering another', object, object.canvas, this);
+        /* _DEV_MODE_END_ */
+        object.canvas.remove(object);
+      }
```
asturur, line 638: in my opinion this should be executed always, not just with an else if.
What if it was in a group, on another canvas?
In that case the older canvas has a reference to an invalid object.
if we are safeguarding, we should remove from group and from canvas.

src/shapes/group.class.ts
```diff
@@ -260,8 +260,17 @@ import { Point } from '../point.class';
...
+        console.warn('fabric.Group: removing object from group before entering another', object, object.group, this);
+        /* _DEV_MODE_END_ */
         object.group.remove(object);
       }
+      else if (object.canvas) {
+        /* _DEV_MODE_START_ */
+        console.warn('fabric.Group: removing object from canvas before entering group', object, object.canvas, this);
+        /* _DEV_MODE_END_ */
+        object.canvas.remove(object);
+      }
```
asturur, line 273: as the comment below, this check seems necessary in every case

asturur:
Forgot about this small pr.
There are 2 comments but otherwise looks good

src/static_canvas.class.ts
ShaMan123: > What if it was in a group, on another canvas?

I that case the first if block will catch it and `Group#remove` will clear refs.
Else if is necessary since we are checking for a parent and because an object always has a ref to canvas (even if not a direct child of canvas) we call tell if canvas is the parent or not only by checking if the `group` ref isn't empty.
I introduced `getParent` in a pending group PR for this purpose exactly.
Then it can become
```js
object.getParent()?.remove(object)

```

ShaMan123:
up

ShaMan123 (commented):
merge conflict artifact found

ShaMan123 (commented):
Updated from master and resolved conflicts

ShaMan123:
@asturur up

asturur:
those fix where committed with other PRs, i see them inside the codebase

src/shapes/group.class.ts
```diff
@@ -296,7 +292,21 @@ export class Group extends FabricObject {}
...
+          );
           object.group.remove(object);
+        } else if (object.canvas) {
+          console.warn(
+            'fabric.Group: removing object from canvas before entering group',
+            object,
+            object.canvas,
+            this
+          );
+          object.canvas.remove(object);
```
ShaMan123, line 309: This is not in master @asturur

ShaMan123:
This Pr needs porting but should enter rc ASAP

PR #8047, opened 2022-07-03
add fabric.Arrow

QJesus:
image

### DIFF
--- a/build.js
+++ b/build.js
@@ -193,6 +193,7 @@ var filesToInclude = [
   'src/shapes/rect.class.js',
   'src/shapes/polyline.class.js',
   'src/shapes/polygon.class.js',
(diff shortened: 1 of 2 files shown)


asturur:
Hi @QJesus , we are in the middle of a typescript rewrite, and we can't merge PRs like this right now.
With the new build system any new subclass that is ideally a leaf node only ( with that i mean that does not require changes to other classes to work ) can be merged.

We don't know where exactly, and how, but we will give instructions.
So leave the PR open, we are going to discuss it soon

asturur:
Also in general, a bunch of introuction around what are you doing, why and how on your pr is good.
Which features this arrow has?
How does caching work?
Does it work with every fabric feature?

Don't ask the reviewer to discover that looking at the code

PR #8063, opened 2022-07-14
Changed calculation to support better rtl

angeloba-microsoft:
The calculation changed avoids issues with the position of the background while using line text on RTL

**Before:** 

image

**After:** 

image

### DIFF
--- a/src/shapes/text.class.js
+++ b/src/shapes/text.class.js
@@ -680,7 +680,7 @@
         if (currentColor && !path) {
           drawStart = leftOffset + lineLeftOffset + boxStart;
           if (this.direction === 'rtl') {
-            drawStart = this.width - drawStart - boxWidth;
+            drawStart = leftOffset + lineLeftOffset - boxWidth;
           }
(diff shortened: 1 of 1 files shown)


asturur:
Hi @angeloba-microsoft , thanks for the PR.
Can you please make an example with actual RTL text, and alos add a visual test of the kind of the one that already exists for RTL?
thanks

angeloba-microsoft:
> Hi @angeloba-microsoft , thanks for the PR. Can you please make an example with actual RTL text, and alos add a visual test of the kind of the one that already exists for RTL? thanks

Sure, I can make a code pen, for the visual test is there an example to base the test on that? Not sure how to do those.

asturur:
Hi again @angeloba-microsoft add some days off didn't open the notebook.
https://github.com/fabricjs/fabric.js/blob/master/test/visual/text.js

Look at this file, and check if the test structure is clear, you can add one there.
Running the text one time will generate the golden image.

PR #8065, opened 2022-07-17
Changes for text dragging

asturur:
This PR makes the first change for the text drag pr.
It removes the need to enter and exit editing the textboxes in order to make the drag effect work.
I m actually comparing right now if i introduced some regressions.
One regression for sure is there, and is the fact that if i drag/drop in the same text box the dragStart selection isn't visible.

Why this change?
EnterEdit and exitEdit do change the status of the textbox, changing their selection start and selection end, plus is an expensive operation since we create a textarea and we add it to the dom, we style it, and we manipulate it, on top of that we are making state changes to the textboxes in order to render a cursor, instead of just rendering a cursor.

The rendering needs to be slow down anyway, i see too many render loop in one second, it must be capped at 60fps OR it must be called once for each time the drop position actually change.

I plan on finishing this tomorrow.

Will follow other small tenatives to reduce the code of this feature, while maintaining all the functionalities.
- defer to requestAnimation frame to render the drag drop effects
- remove the boolean _isDraggingOver ( should be possible ).
- reduce the event handlers to simple event handlers and call the effect explicitly instead of delegating them to firing events., avoiding that a simple text.off('drop') will eventually break the functionality.
- check if there is a way to reuse current canvas pixels for
For merging the main branch, this is the only PR that is important to me, the other are small details that can follow if they require more than on day each.

Performances:
Original PR, 5 frames of dragover
image

this PR, 5 frames of dragover
image

after additional changes doesn't seem we lost or gained much
image


asturur:
This PR solves a bug by chance.
The original PR using the actual enterEditing functionality was considering the modifier keys when dragging over, making the drop position a text selection instead of a text cursor. The drop effect would then occur on the cursor and not on the selection.

There is also another small bug in the original PR, and is not a regression ( maybe is a generic bug in fabricJS right now ), sometimes the style of the dropped text is applied uncorrectly, shifting right by one char or by as many char as the newlines contained in the selection.
Does not happen in all cases.

ShaMan123:
> The original PR using the actual enterEditing functionality was considering the modifier keys when dragging over, making the drop position a text selection instead of a text cursor. The drop effect would then occur on the cursor and not on the selection.

I am not sure I understand what you mean. HTML drop selects the text that was dropped to the drop target. This PR changed it and I disagree with it.


See this gif showing regressions:
- no cursor showing on selected drop target
- wrong selection on drop (wrong selection if` dragSource ===dropTarget`, no selection elsewise) - should be changed to the dropped text
- selection not showing if `dragSource ===dragOverTarget` (as you mentioned)

ezgif com-gif-maker (1)

src/mixins/itext_behavior.mixin.js
```diff
@@ -562,26 +560,21 @@
...
-        this.__isDragging && this.abortCursorAnimation();
       }
       else if (this.__isDraggingOver && !canDrop) {
         //  drop state has changed
         this.__isDraggingOver = false;
         !this.__isDragging && this.clearContextTop();
-        this.exitEditing();
       }
-      if (this.__isDraggingOver) {
+      if (this.__isDraggingOver && canDrop) {
```
ShaMan123, line 569: are you sure about this?

src/mixins/itext_behavior.mixin.js
```diff
@@ -696,12 +685,9 @@
         options.dropTarget = this;
         //  finalize
         this.insertChars(insert, styles, insertAt);
-        this.selectionStart = Math.min(insertAt + selectionStartOffset, this._text.length);
```
ShaMan123, line 699: I disagree with this deletion

src/mixins/itext_behavior.mixin.js
```diff
@@ -562,26 +560,21 @@
...
         options.canDrop = true;
         options.dropTarget = this;
-        //  render
-        this.setCursorByClick(e);
-        this._updateTextarea();
-        this.restartCursorIfNeeded();
-        this.renderCursorOrSelection();
+        // find cursor under the drag part.
+        var dragSelection = this.getSelectionStartFromPointer(e);
+        this.renderCursorAt(dragSelection);
```
ShaMan123, line 577: If I understand correctly we should render the selection here in case `dragSource===dragOverTarget`

ShaMan123: https://github.com/fabricjs/fabric.js/blob/2db76ec809ef3daf33c37e44f40f23ddb038fb42/src/shapes/itext.class.js#L335

src/mixins/itext_behavior.mixin.js
asturur: i think this is a consequence of the previous ifs, and i made explicit.
I didn't think of it too much, could be wrong.

src/mixins/itext_behavior.mixin.js
asturur: In the pr description i wrote that:
```
One regression for sure is there, and is the fact that if i drag/drop in the same text box the dragStart selection isn't visible.
```

I know why this is broken, and is because of the double clear textarea, i was supposed to fix it yesterday but then i couldn't.

asturur:
> HTML drop selects the text that was dropped to the drop target. This PR changed it and I disagree with it

So HTML has only textareas that are permanently in edit state.
This concept does not apply to fabric text rendering.

I'm happy adding a feature that allow for drag/drop text, but not adding a full flow that put the developer in a rigid user experience. We implement the minimal and we let them use the events to build the experience they prefer using the drop event or drag end event (enterEdit after drop, select after drop, or just place the cursor at the end after drop)

src/mixins/itext_behavior.mixin.js
asturur: oh i see is not the same because this just an if and not an else if.
needs to be removed

src/mixins/itext_behavior.mixin.js
ShaMan123: At least set the selection value to the drop cursor value
Though i still think i should remain as is and the dev csn override it. It's the expected behavior IMO

src/shapes/itext.class.js
```diff
@@ -487,24 +516,34 @@
...
+
     /**
      * Renders text selection
      * @private
      * @param {{ selectionStart: number, selectionEnd: number }} selection
      * @param {Object} boundaries Object with left/top/leftOffset/topOffset
      * @param {CanvasRenderingContext2D} ctx transformed context to draw on
      */
-    _renderSelection: function (selection, boundaries, ctx) {
+    _renderSelection: function (ctx, selection, boundaries, ctx) {
```
melchiar, line 546: @asturur why two `ctx` arguments?

melchiar, line 546:
```suggestion
    _renderSelection: function (ctx, selection, boundaries) {
```

src/shapes/itext.class.js
```diff
@@ -446,6 +470,8 @@
         - charHeight * (1 - this._fontSizeFraction);
 
       if (this.inCompositionMode) {
+        // TODO: investigate why there isn't a return inside the if,
+        // and why can't happe top of the function
         this.renderSelection(boundaries, ctx);
```
melchiar, line 475: need to swap these

src/shapes/itext.class.js
```diff
@@ -332,15 +333,34 @@
...
+    renderCursorAt: function(selectionStart) {
+      var ctx = this._clearContextTop();
+      if (!ctx) {
+        return;
       }
+      this._renderCursorAt(selectionStart);
       ctx.restore();
     },
 
+    _renderCursorAt: function(selecetionStart) {
```
melchiar, line 359:
```suggestion
    _renderCursorAt: function(selectionStart) {
```

src/shapes/itext.class.js
```diff
@@ -434,7 +454,11 @@
      * @param {CanvasRenderingContext2D} ctx transformed context to draw on
      */
     renderCursor: function(boundaries, ctx) {
```
melchiar, line 456: need to swap arguments here

src/shapes/itext.class.js
```diff
@@ -332,15 +333,34 @@
...
+      if (!ctx) {
+        return;
       }
+      this._renderCursorAt(selectionStart);
       ctx.restore();
     },
 
+    _renderCursorAt: function(selecetionStart) {
+      var boundaries = this._getCursorBoundaries(selectionStart, true);
+      this._renderCursor(ctx, boundaries, selectionStart);
```
melchiar, line 361: ctx is missing here

src/shapes/itext.class.js
```diff
@@ -462,23 +488,26 @@
      * @param {Object} boundaries Object with left/top/leftOffset/topOffset
      * @param {CanvasRenderingContext2D} ctx transformed context to draw on
      */
-    renderSelection: function (boundaries, ctx) {
+    renderSelection: function (ctx, boundaries) {
```
asturur, line 491: i moved those argument orders, in general, because we put ctx as first argument when we render.

src/mixins/object_interactivity.mixin.js
```diff
@@ -311,5 +311,13 @@
...
     },
 
+    renderDragSourceEffect: function() {
+      // for subclasses
+    },
+
+    renderDropTargetEffect: function(/* e, dontClear */) {
+      // for subclasses
+    },
+
```
asturur, line 321: this make clear that if you want to support drag/drop you have to define at least those 3 ( i will try to move the event handlers on Object interactivity ( another week )

src/shapes/itext.class.js
```diff
@@ -158,23 +158,13 @@
...
-    /**
-     * @private
-     */
-    _abortCursorAnimation: false,
-
-    /**
-     * @private
-     */
-    __widthOfSpace: [],
-
```
asturur, line 177: completely unused

src/mixins/itext_behavior.mixin.js
asturur: this is outdated, not sure why github doesn't get it

src/shapes/itext.class.js
asturur: because i was sleepy. And also the second one was killing the first one, magic JS, even if undefined.

asturur:
oh the UTs

asturur:
image

and lint fixed

src/mixins/itext_behavior.mixin.js
```diff
@@ -564,17 +560,14 @@
       else if (this.__isDraggingOver && !canDrop) {
         //  drop state has changed
         this.__isDraggingOver = false;
-        !this.__isDragging && this.clearContextTop();
       }
-      if (this.__isDraggingOver && canDrop) {
+      if (this.__isDraggingOver) {
```
ShaMan123, line 564: Now I recall this logic
`canDrop` => if another object has accepted the event it has become the drop target and has called `e.preventDefault()` to inform the browser so this object should not try to handle the event. And of course if it `canDrop(e)` is false...

Is there anything I am missing out?

ShaMan123 (commented):
I like the approach of the last commits `renderDropEffect` etc.
I will test behavior on browser and comment if I find regressions.

src/shapes/itext.class.js
```diff
@@ -510,24 +502,38 @@
...
+          ctx,
           this.__dragStartSelection,
-          this._getCursorBoundaries(this.__dragStartSelection.selectionStart, true),
-          ctx
+          this._getCursorBoundaries(this.__dragStartSelection.selectionStart, true)
         );
       }
     },
 
+    renderDropTargetEffect: function(e, dontClear) {
```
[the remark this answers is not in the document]
asturur: i tried it but didn't work, there was a case in which i was killing the source effect anyway.

What we can do is trying to transform ctx here, and pass it along and make that the effects receive a ctx.

move clearContextTop from text to object

clearContextTop for source, for target,
then draw them in sequence.

This will mean that we change the signature of the functions source and target to accept e and ctx.
is that ok?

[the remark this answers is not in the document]
ShaMan123: sounds better IMO

ShaMan123:
There's a small regression on dragstart. It clears the selection and should not.

src/mixins/itext_behavior.mixin.js
ShaMan123: needs to be restored IMO

asturur: i removed this, i m not sure why gitbud doesn't get it

asturur: i mean there is nothing to revert, the line is white

image

ShaMan123: Ohh

ShaMan123: OK

src/mixins/object_interactivity.mixin.js
```diff
@@ -279,6 +279,32 @@
...
 
+    /**
+     * Clears the canvas.contextTop on top of the object.
+     * This function is used to clear pieces of contextTop where we render ephemeral effects
+     * Example: blinking cursror text selection.
+     * // TODO: discuss swapping restoreManually with a renderCallback, but think of async issues
+     * @param {Boolean} [restoreManually] true to don't restore the context after clear, in order to draw something else.
+     * @returns {CanvasRenderingContext2D|undefined} canvas.contextTop transformed on the object center
+     */
+    clearContextTop: function(restoreManually) {
```
asturur, line 290: this for me solves what i didn't like of the ctx = this.prepareContextTop() and also streamline a bit the code that was divied in 3 methods

asturur: i know that restoreManually seems a bit like dontClear but isn't exactly as that.

ShaMan123: Ok
It is better
The callback approach sounds good
Let's merge

ShaMan123: Maybe make jsdoc a bit better?

src/mixins/itext_behavior.mixin.js
```diff
@@ -172,7 +172,7 @@
 
       //  make sure we clear context even if instance is not editing
       if (shouldClear) {
-        var ctx = this._clearContextTop();
+        var ctx = this.clearContextTop(true);
```
ShaMan123, line 175: Ctx is restored on the next line so can we let the method restore it?

src/mixins/itext_behavior.mixin.js
```diff
@@ -605,7 +605,7 @@
             this._updateTextarea();
           }
           else {
-            var ctx = this._clearContextTop();
+            var ctx = this.clearContextTop(true);
```
ShaMan123, line 608: graphics-et-al

src/mixins/object_interactivity.mixin.js
```diff
@@ -279,6 +279,32 @@
       return this;
     },
 
+    /**
+     * Clears the canvas.contextTop on top of the object.
+     * This function is used to clear pieces of contextTop where we render ephemeral effects
+     * Example: blinking cursror text selection.
+     * // TODO: discuss swapping restoreManually with a renderCallback, but think of async issues
+     * @param {Boolean} [restoreManually] true to don't restore the context after clear, in order to draw something else.
```
ShaMan123, line 287: Remove 'to'

src/mixins/object_interactivity.mixin.js
```diff
@@ -279,6 +279,32 @@
       return this;
     },
 
+    /**
+     * Clears the canvas.contextTop on top of the object.
```
ShaMan123, line 283: Unclear

src/mixins/object_interactivity.mixin.js
```diff
@@ -279,6 +279,32 @@
...
+     * This function is used to clear pieces of contextTop where we render ephemeral effects
+     * Example: blinking cursror text selection.
+     * // TODO: discuss swapping restoreManually with a renderCallback, but think of async issues
+     * @param {Boolean} [restoreManually] true to don't restore the context after clear, in order to draw something else.
+     * @returns {CanvasRenderingContext2D|undefined} canvas.contextTop transformed on the object center
```
ShaMan123, line 288: On object center is weird english

ShaMan123 (approved):
Some minor stuff

src/mixins/object_interactivity.mixin.js
asturur: some function hasn't be jsdoced, i know. but i really need to stop. Let me open an issue to come back to those things.

src/mixins/itext_behavior.mixin.js
asturur: because i did a blind search/replace 😄 
correcting this

src/mixins/object_interactivity.mixin.js
asturur: true

src/mixins/object_interactivity.mixin.js
```diff
@@ -280,12 +280,14 @@
...
+     * This function is used to clear pieces of contextTop where we render ephemeral effects on top of the object.
+     * Example: blinking cursror text selection, drag effects.
      * // TODO: discuss swapping restoreManually with a renderCallback, but think of async issues
-     * @param {Boolean} [restoreManually] true to don't restore the context after clear, in order to draw something else.
-     * @returns {CanvasRenderingContext2D|undefined} canvas.contextTop transformed on the object center
+     * @param {Boolean} [restoreManually] When true won't restore the context after clear, in order to draw something else.
+     * @return {CanvasRenderingContext2D|undefined} canvas.contextTop that is either still transformed
+     * with the object transformMatrix, or restord to neutral transform
      */
     clearContextTop: function(restoreManually) {
```
ShaMan123, line 292: I can live with this though I think it is better to have the arg be `restore` instead of `dontRestore`
I will create a branch of that for a quick merge

src/mixins/object_interactivity.mixin.js
```diff
@@ -279,6 +279,34 @@
...
 
+    /**
+     * Clears the canvas.contextTop in a specific area that correspond to the object's bounding box
+     * that is in the canvas.contextContainer.
+     * This function is used to clear pieces of contextTop where we render ephemeral effects on top of the object.
+     * Example: blinking cursror text selection, drag effects.
+     * // TODO: discuss swapping restoreManually with a renderCallback, but think of async issues
+     * @param {Boolean} [restoreManually] When true won't restore the context after clear, in order to draw something else.
+     * @return {CanvasRenderingContext2D|undefined} canvas.contextTop that is either still transformed
+     * with the object transformMatrix, or restord to neutral transform
```
ShaMan123, line 290:
```suggestion
     * with the object transformMatrix, or restored to neutral transform
```

src/mixins/object_interactivity.mixin.js
asturur: we can have restore = true as soon as we use es6 syntax

asturur:
follow up https://github.com/fabricjs/fabric.js/issues/8095

PR #8073, opened 2022-07-20
add a undo_redo lib 

QJesus:
1. fabric.Text.text cannot be null\undefined
2. undo_redo extension

### DIFF
--- a/lib/undo_redo.js
+++ b/lib/undo_redo.js
@@ -13,8 +13,7 @@
 
     const _initialize = fabric.Canvas.prototype.initialize;
     fabric.Canvas.prototype.initialize = function () {
-        const t = _initialize.call(this);
-
+        const inited = _initialize.call(this);
         this.clearUndoRedo();
         this.on('object:added', e => {
             if (this.undoredo.escape || !e.target || e.target.excludeFromExport) {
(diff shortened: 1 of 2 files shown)


ShaMan123:
> fabric.Text.text cannot be null\undefined

Can be a standalone PR

lib/undo_redo.js
```diff
@@ -0,0 +1,121 @@
...
+            Object.keys(e.transform.original).forEach(k => destination[k] = e.target[k], this);
+            this.undoredo.stack[++this.undoredo.index] = { kind: 'modified', action: e.action, target: e.target, original: e.transform.original, destination };
+            this.undoredo.stack.length = this.undoredo.index + 1;
+        });
+        this.on('canvas:cleared', _ => this.clearUndoRedo());
+        return t;
+    }
+
+    const ___setupCanvas = fabric.canvas.prototype.__setupCanvas;
+    fabric.canvas.prototype.__setupCanvas = function (serialized, enlivenedObjects, renderOnAddRemove, callback) {
```
melchiar, line 47:
```suggestion
    const ___setupCanvas = fabric.Canvas.prototype.__setupCanvas;
    fabric.Canvas.prototype.__setupCanvas = function (serialized, enlivenedObjects, renderOnAddRemove, callback) {
```

Nice lib idea! `fabric.canvas` should have uppercase C though

melchiar (changes requested):
Can we improve some of these variable names. Names like t, v, and h aren't very human readable.

ShaMan123:
I don't think history should be part of core.
IMO history is something that is heavily customized between different apps.
If it is hard to build such functionality I would suggest PRing to make fabric easier to develop such functionalities. That makes sense to me.
Regardless, `saveObjectTransform` isn't good enough. What about skewing? cacheProperties? stateProperties?
This re-introduces state management which we have decided to deprecate in #7920

ShaMan123:
@QJesus I am very for contributions, hope I didn't put you down.
We are migrating to TS if you want to join the effort. And that should eventually open up a better way to write extensions for fabric.

asturur:
Actually we wanted to think about a more simple way to tackle this, both on core and an external lib. We have to move forward with our schedule and as soon as the rollup pr is merged, we can get back to this.

melchiar:
@QJesus Thanks for the pr, however as asturur explained this isn't quite the direction we want to take with an undoredo library. I'm going to close this PR for now, but we can revisit adding an official undoredo component once v6 is out of the gate.

PR #8081, opened 2022-07-24
minor(Canvas.getPointer): `recalculate` arg

ShaMan123:
This PR is very minor and add a `recalculate` arg to `getPointer`
It is needed in some edge cases when calling `getPointer` and for some reason the caches still exist.

I am not sure about the PR cause it seems there is a bug somewhere else not invalidating caches.
It happens when I draw with a brush and then use canvas.getPointer on a drop event
calling `resetTransformData` does the trick of course


ShaMan123:
found the bug

Issue #8025, opened 2022-06-25
bug(Path): stroke projection on bbox


ShaMan123:
I see I didn't impl  https://github.com/fabricjs/fabric.js/pull/7478 on path which is strange.
I'll wait until we migrate and decide the fate of all the shapes that should subclass Path.
We should discuss how to impl this on Path at that point.
And remove the flag `exactBoundingBox`

ShaMan123:
Could it be that path accounts for stroke acute angles?

ShaMan123:
Take a look at this image that corresponds to `projectStrokeOnPoints`:
![stroke_projection](https://raw.githubusercontent.com/fabricjs/fabric.js/e133fa0ede73c7364d8cdc890fd8b256f1713ab4/src/util/stroke_projection.svg)

With `strokeUniform` we need to project a different vector because the plane is transformed with inverse scaling

This confusing part is that it isn't visible to our eyes, only in the eyes of the object's plane

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/7847#issuecomment-1173779421_



The blue opaic lines are probably how the object sees the vectors that create the bisector... which mean we ought to rotate the vectors in the angle diff (tried that without luck)... BUT it makes no sense to me and is pretty impossible to debug

![stroke_projection](https://user-images.githubusercontent.com/34343793/177159031-f85aa98e-f6cd-43ca-acb8-8e1bedc2d49f.svg)

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/7847#issuecomment-1173787679_

luizzappa:
@ShaMan123 If you plot the points it's easier to understand.

Plotting a triangle with the points:
A (0, 0)
B (50, -200)
C (100, 0)

![image](https://user-images.githubusercontent.com/65685842/179652258-349b5e5f-1fbc-4523-962b-263f2ed053d0.png)

The angle of the bisector is alpha.

But when I apply a scaling only on the X axis of 2, the angle of the bisector changes (beta < alpha), hence the miter vector too. To calculate this new angle we must apply the scaling on the points to find the new vectors that form A, B and C.

![image](https://user-images.githubusercontent.com/65685842/179652978-2d782545-7762-4708-84cb-68eb82e7fbce.png)

scaleX = 2 ; scaleY = 1

scaled A (0*scaleX , 0*scaleY) --> (0, 0)
[...]

Issue #7847, opened 2022-03-31
Stroke for triangles and polygons get cut off


isaac-khooyt-shopee:
## Version
5.2.1
4.5.0

## Test Case
https://codepen.io/zaccccc/pen/OJzjGmW

## Information about environment
Nodejs or browser? browser
Which browsers? all

## Steps To Reproduce

1. create triangle or polygon of about 100x100
2. add stroke information and large strokeWidth (i.e. 100)
3. stroke is cut off, stroke behaves weirdly when scaling

## Expected Behavior
stroke does not get cut off.
stroke does not leave artifacts / show unexpected borders when scaled-flipped

## Actual Behavior
stroke gets cut off
stroke shows unexpected artifacts and borders when scaled or scaled-flipped

![image](https://user-images.githubusercontent.com/102722105/161001247-b8ce6876-9bb7-4983-89b0-a3978419baea.png)
![Screen Shot 2022-03-31 at 3 29 28 PM](https://user-images.githubusercontent.com/102722105/161001599-22bf0c6d-57d9-41d1-9467-9c27aad03c77.png)

ShaMan123:
This fix is part of v4.6+
https://codepen.io/shaman123/pen/bGaoRXe

However I forgot triangles out. So best use a polygon.
It was decided to make this opt in by setting `exactBoundingBox:true` and default in a future release.

isaac-khooyt-github-coding:
I cant seem to replicate this in my personal project using v5.2.1, but it works fine in the codepen.

Also i noticed that `exactBoundingBox:true` doesn't work if it is set using the set method after it has been initialized.
https://codepen.io/zaccccc/pen/jOYGZQo

ShaMan123:
If you look at the source ode it's straight forward
you need to call `_setPositionDimensions`
https://codepen.io/shaman123/pen/vYpeRzb

isaac-khooyt-github-coding:
i see, thank you!

ShaMan123:
@isaac-khooyt-github-coding the plan is to deprecate triangle in favor of a subclass of Path/Polyline. This will solve your issue out of the box.
If you want to PR I can support you.

luizzappa:
If strokeUniform property is equal to true, the stroke gets cut off:

![image](https://user-images.githubusercontent.com/65685842/177061800-cf86e40d-1ebf-4177-a288-b83c8835a398.png)

https://codepen.io/Theluiz-eduardo/pen/MWVwMyy

[...]

PR #8040, opened 2022-06-30
fix(Path): bbox stroke projection

ShaMan123:
closes #8025 
closes #8036

And tests fail because Path has broken it's positioning logic somehow... I don't yet understand how or why.

### TODO

- [x] fix positioning

PR #8083, opened 2022-07-26
fix(polyline): stroke bounding rect calculation

luizzappa:
# Related Issues
- https://github.com/fabricjs/fabric.js/issues/8025
- https://github.com/fabricjs/fabric.js/issues/7847
- #7478

# Proposed changes

- ~~Created `acos` function in `fabric.util` to ensure the value is in the range `[-1, 1]` and the `calcAngleBetweenVectors` function is using it. This was causing a bug ([see comment on commit bc93beb](https://github.com/fabricjs/fabric.js/commit/bc93beba924147d42af963bbdfaca17a2cd640d6#commitcomment-79407686))~~
- ~~In `calcAngleBetweenVectors` function, replaces `Math.hypot(x, y)` with `Math.sqrt(x*x + y*y)` to ensure equal results in both engines: Firefox and Chrome. ([see comment on commit bc93beb](https://github.com/fabricjs/fabric.js/commit/bc93beba924147d42af963bbdfaca17a2cd640d6#commitcomment-79407698))~~
- ~~In `getBisector` function, checks if `ro` is close to zero instead of being strictly equal.~~
- In `calcAngleBetweenVectors` function, calculate angle between two vector using atan2 instead of dot product. So eliminated the need to check the angle direction in `getBisector` function, 
- Changed the `_setPositionDimensions` function in the polyline class to adjust the width/height by the scale in case of uniform stroke.
- Changed the `projectStrokeOnPoints` function to cover all stroke line join cases and also to uniform stroke. This function deserves further comment:

In the case of **open path**, the stroke is rendered differently, so I calculated with the vectors orthogonal to the stroke.
image

In the case of the **bevel**, they are the points formed by the orthogonal to the vertex
image

In the case of the **round**, ~~if the angle is less than or equal to 90°,~~ the projection is calculated with two vectors parallel to the X and Y axes. ~~In the case of larger angles, it is calculated using the orthogonals.~~
image

In the case of the **miter**, is the same calculation commented here: https://github.com/fabricjs/fabric.js/issues/8025

For a better view, see this link (you can change the vectors direction by holding the red circles) : https://hypertolosana.github.io/efficient-webgl-stroking/index.html

# Screenshots

Working example here: https://codesandbox.io/s/project-stroke-points-with-context-to-trace-b8jc4j?file=/src/index.js

image

The red dotted vector is the bisector returned by the getBisector function. The green dotted vector is in the opposite direction and is what is used to determine which side to plot the projections. The projections are the colored vectors, one color for each point.
To make debugging easier, you can configure the parameters (stroke line join type, strokeUniform, miterLimit, etc..)

## openPath = false and strokeUniform = true

| Angle   |      Miter      |      Round    |      Bevel     |
|----------|:-------------:|:-------------:|:-------------:|
| less than 90° |  image | image | image
| greater than 90° |    image   |   image | image
| equal to 180° | image | image | image |


luizzappa (commented):
**I don't know if this can generate a bug in case of very small angles**

ShaMan123:
Amazing work.
I will review it later on today.
Meanwhile, we can rethink `calcAngleBetweenVectors` and use `atan2` diff instead of dot product. That will give use the exact angle with the correct direction, eliminating the need to check the direction ourselves, hence resolving the `ro` issue.
How strange the `acos` gets values out of bounds.

luizzappa:
> Amazing work. I will review it later on today. Meanwhile, we can rethink `calcAngleBetweenVectors` and use `atan2` diff instead of dot product. That will give use the exact angle with the correct direction, eliminating the need to check the direction ourselves, hence resolving the `ro` issue. How strange that `acos` gets values out of bounds.

Great idea! I just committed this change.

src/util/misc.js
```diff
@@ -221,50 +237,185 @@
...
+                hatOrthogonalVector.x * s * strokeUniformScalar.x, 
+                hatOrthogonalVector.y * s * strokeUniformScalar.y 
+              );
+
+          coords.push(A.add(orthogonalVector));
+          coords.push(A.subtract(orthogonalVector));
+          return;
+        }
+
+        if (openPath && index === points.length - 1) {
```
ShaMan123, line 279: This is redundant
Tet's merge the 2 cases `openPath && (index === 0 || index === points.length - 1)`
We can do something like
```js
var D = index === 0 ? C : B

```

src/util/misc.js
```diff
@@ -221,50 +237,185 @@
...
         }
         else {
           B = points[index - 1];
           C = points[index + 1];
         }
-        var bisector = fabric.util.getBisector(A, B, C),
-            bisectorVector = bisector.vector,
+
+        if (openPath && index === 0) {
+          var scaledA = new fabric.Point(A.x * options.scaleX, A.y * options.scaleY),
```
ShaMan123, line 261: **code styling**
add at the top of the function
```
var scale = new fabric.Point(options.scaleX, options.scaleY);
```
should be:
```suggestion
          var scaledA = new fabric.Point(A).multiply(scale),
```

And apply to all the rest
There's `Point#scalarMultiply(scalar:number)` as well

src/util/misc.js
```diff
@@ -221,50 +237,185 @@
...
+
+          coords.push(A.add(orthogonalAC));
+          coords.push(A.subtract(orthogonalAC));
+
+          return;
+        }
+
+        if (options.strokeLineJoin === 'round') {
+
+          if (alpha > PiBy2) {
```
ShaMan123, line 379: This seems redundant as well
Seems like it belongs to the previous block

And we should use `else if` for readability if possible

src/util/misc.js
```diff
@@ -190,19 +192,33 @@
...
 
     /**
-     * Project stroke width on points returning 2 projections for each point as follows:
+     * @static
+     * @memberOf fabric.util
+     * @param {Point} vector 
+     * @param {Boolean} counterClockwise the direction of the orthogonal vector 
+     * @returns {Point} the unit orthogonal vector
+     */
+    getOrthogonalUnitVector: function(vector, counterClockwise = true) {
```
ShaMan123, line 210: `counterClockwise = true` will probably cause lint to go mad cause fabric's js is so old.
Fortunately, the next PR to enter fabric is TS so that won't be a problem.
But I do intent on merging this PR before that.
We could negate the arg to `clockwise` - but let's leave as is and TS will fix it. We can live with that.

src/util/misc.js
```diff
@@ -176,7 +178,8 @@
      * @returns {Point} vector representing the unit vector of pointing to the direction of `v`
      */
     getHatVector: function (v) {
-      return new fabric.Point(v.x, v.y).scalarMultiply(1 / Math.hypot(v.x, v.y));
+      var hypot = Math.sqrt(v.x * v.x + v.y * v.y);
+      return new fabric.Point(v.x, v.y).scalarMultiply(1 / hypot);
```
ShaMan123, line 182: removed the remaining use of `Math.hypot`

ShaMan123 (commented):
**This is excellent work. Thorough and precise.**
Most of my comments are related to styling.
I would like to eliminate the use of duplicate code as much as possible and utilize `fabric.Point` methods such as `multiply` `scalarMultiply` making the code more readable.

I will commit visual tests with the examples you have created.
Once I do that it will be easy for you to add more. We should cover all cases and as many edge cases as possible.
I will explain how to run the tests as well.

I would never have figured this out:
> In the case of the round, if the angle is less than or equal to 90°, the projection is calculated with two vectors parallel to the X and Y axes. In the case of larger angles, it is calculated using the orthogonals.

Thank you for this awesome contribution!

src/util/misc.js
ShaMan123: The commit message is wrong. I thought firefox calculated it wrong but it's chrome.
I wonder if there are other browser core methods that cause strange bugs in fabric.
I would never even think of checking `Math.hypot` or `Math.acos` assuming for sure they are correct.
Alarming notion.

src/util/misc.js
ShaMan123: Rethinking, we can use `fabric.util.rotateVector` instead. Better I think.

ShaMan123:
@luizzappa I have created a PR with tests

To run tests on fabric:
1. `npm run build -- -w`
2. `npm test`: select `visual/generic-rendering`

This will run visual tests on node. Which isn't enough.

To run in the browser you need to clone fabricjs.com. See the [contribution guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#tests) for how to.

ShaMan123:
Visual testing has exposed a bug in `stokreMiterLimit`

src/util/misc.js
ShaMan123: Coming to think of it, Tylor is probably how `hypot` is calculated. And then it is a question of the error precision.
But I'm guessing here

ShaMan123:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit

> image

I wonder what is *the inner corner of the miter*
> The ratio of miter length (distance between the outer tip and the inner corner of the miter) to [stroke-width](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-width)

luizzappa:
> @luizzappa I have created a PR #8087 with tests that is forked from this one. I don't want to merge it here to avoid the mess of so many files.
> 
> To run tests on fabric:
> 
> 1. `npm run build -- -w`
> 2. `npm test`: select `visual/generic_rendering`
> 
> This will run visual tests on node. Which isn't enough.
> 
> To run in the browser you need to clone fabricjs.com. See the [contribution guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#tests) for how to.
> 
> And don't hesitate to reach out if it isn't clear enough.

I didn't quite understand what you're comparing to in these tests. Is the percentage of the object that is outside the blue rectangle?

Running on node there were no failures.

image

I must be doing something wrong to run in the browser. I cloned fabricjs.com, copied the tests, but they don't show up.

image

luizzappa:
> Visual testing has exposed a bug in `strokeMiterLimit`

Would you mind reproducing a buggy case here: https://codesandbox.io/s/fabricjs-project-stroke-working-forked-gem8z9?file=/src/index.js

I tried to reproduce but I didn't find the bug

Golden:
image

Codesandbox:
image

ShaMan123:
Forgive me, I forgot a command... To start the fabricjs.com in dev mode run from fabric.js `npm start`
Then it should show up.

The tests run and compare the output with an image ref stored in `test/visual/golden/**`.
The percentage is the pixel diff between the two.
There will be no diff because I generated the refs from the current state of the code. Meaning it will output the same image.
Navigate to `test/visual/golden/stroke-projection/miter/` and see that there are refs that have a wrong bbox due to the miter limit calculation.

ShaMan123:
image

ShaMan123:
My bad for the second time!
I have discovered a bug in node canvas miter limit calculation.
As you can see the left side image is the ref image I produced (node) and the right side is the image generated by the browser test
I will update the refs

src/util/misc.js
```diff
@@ -158,15 +157,18 @@
...
      * @param {Point} a
      * @param {Point} b
      * @returns the angle in radian between the vectors
      */
     calcAngleBetweenVectors: function (a, b) {
-      return Math.acos((a.x * b.x + a.y * b.y) / (Math.hypot(a.x, a.y) * Math.hypot(b.x, b.y)));
+      var dot = a.x * b.x + a.y * b.y,
+				det = a.x * b.y - a.y * b.x;
+    
+      return Math.atan2(det, dot);
```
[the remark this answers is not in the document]
luizzappa: But you understood? I found out by searching an efficient way to calculate the angle between two vectors. Anyway, here's the explanation:

dot = |A·B| = |A| |B| COS(θ)
det = |A×B| = |A| |B| SIN(θ)

tan θ = |A| |B| SIN(θ) / |A| |B| COS(θ)
tan θ = det / dot

atan2 is more or less this (actually we should apply a correction when x < 0):
atan2(y, x) = arctan(y/x)

y = det, x = dot

atan2(det, dot)

src/util/misc.js
luizzappa: I also thought of using the `rotateVector` function. But since the orthogonal vector is just swapping the coordinates, it seemed to me to be less prone to accuracy issues. 

I made a commit removing the optional parameter https://github.com/fabricjs/fabric.js/pull/8083/commits/b554b6d64de1529ef31f095ea5c3d7d2c254fa03

However, if you think it's better to use the `rotateVector` function, I'll refactor.

luizzappa:
hmm looks like I messed up during the refactoring and it broke something. Tomorrow I'll take a closer look.

ShaMan123 (approved):
I have done the following:
- linted code
- reverted the change to `getOrthogonalHatVector`
- added `hypot` as a util
- deprecated `exactBoundingBox`
- changed the `if` block order

@luizzappa as far as I am concerned this can be merged once you give the green light that your tests pass

ShaMan123:
@asturur @melchiar I want to merge this PR
I thought I could merge it before TS because it is just about ready.
But many tests fail because of this fix...
Should I wait? I guess I should.

ShaMan123:
@luizzappa I have discovered a bug I think. I will elaborate once I triage it

ShaMan123:
Should we call `_setPositionDimensions` when a stroke prop changes?

ShaMan123:
found a bug:
- [Edit fabricjs-project-stroke (working) (forked)](https://codesandbox.io/s/fabricjs-project-stroke-working-forked-5yeipk?fontsize=14&hidenavigation=1&theme=dark)
- and I think something with `_setPositionDimensions` in case there are less the 1 points

ShaMan123:
I think I understand the first bug (the one with the repro).
The original thought I had was to project stroke to both possible sides because I didn't want to calculate which side is the side to project to, but it is flawed.
I thought of simple polygons and small miters, in those cases it works.
However, a self-intersecting polygon or a polyline both with a very large miter break this concept. The projected miter of the inner side exits the shape and makes the bbox larger.
This means we must do intersection tests with the miter vectors and the rest of the polygon (without the current A vertex) using the raw casting algorithm to detect which vector is the vector pointing outwards.
I can do that unless you want to.

So we have a bit more work meaning I will not be able to merge this before TS but that's OK.
I will push this forward to the top of my list.
This makes me think I will want to merge #8040 after this PR. I could use your help there as well once this is merged.

@luizzappa what are your thoughts?

ShaMan123:
Anyways the PR is a giant leap forward so we might merge it as is and open a follow up.
Cause we might hit perf issues for the detection...
@asturur thoughts?

luizzappa:
> 

@ShaMan123 you're right. But I think there is a simpler solution.

I believe this bug only occurs with `miter`. 
For `round` and `bevel` strokes, the projections cannot exceed the correct side projections of the other points.

The `getBisector` function always points inwards. So we don't need to calculate both sides, just one.

Removing that line seems to solve it. I tried to simulate other cases that could break, but I didn't find.
```
 if (fabric.util.hypot(miterVector.x, miterVector.y) / s <= strokeMiterLimit) {
            coords.push(A.add(miterVector));
            // coords.push(A.subtract(miterVector));
            return;
          }
```

Take a look: https://codesandbox.io/s/fabricjs-project-stroke-working-forked-z36cho?file=/src/index.js

What do you think?

ShaMan123:
I can think of a case which `strokeWidth` is very big and the shape is small/thin so the bevel/round projection will be off as well

ShaMan123:
The complication is that for non convex polygons we don't have enough context to decide which side to project to

ShaMan123:
הורדה

ShaMan123:
הורדה (1)

ShaMan123:
Our tests are not comprehensive enough
I will add self-intersecting polygons and very big values

ShaMan123:
the miter vector points to the opposite direction of BC, right?

ShaMan123:
I like this work!

ShaMan123:
### Brain Storming
We could check if `A` is inside the polygon created by the rest of the vertexes.
If it is we reverse the miter vector's direction.
fabric has an `Intersection.intersectsLinePolygon` method for that. Or we could simply check x,y values somehow

luizzappa:
> The complication is that for non convex polygons we don't have enough context to decide which side to project to I will add some visuals

It seems to me that the miter vector that matters is always the one that points out.

Look at an extreme case of a concave polygon where the miter vector extends beyond the stroke of the right side:
image


The stroke would be something like this:
image


Playing on canvas: https://codesandbox.io/s/fabricjs-project-stroke-working-forked-h9pve9?file=/src/index.js

image

ShaMan123:
https://codepen.io/shaman123/pen/PoROMOV?editors=0010
Sorry for using codepen... I adapted http://fabricjs.com/custom-controls-polygon

image

---

This simple polygon disproves your theorem.
The vector pointing to the top left corner is in the same direction of the bisector vector


> 2 vertexes pointing to different sides
> 
> הורדה (1)

luizzappa:
I didn't commit this change, the build you used in the codepan is not up to date.

image

Proposed:
```
 if (fabric.util.hypot(miterVector.x, miterVector.y) / s <= strokeMiterLimit) {
            coords.push(A.add(miterVector));
            // coords.push(A.subtract(miterVector));
            return;
          }
```

I did a local build and uploaded it to codesandbox. Using the same example posted by you, it seems to be working: https://codesandbox.io/s/fabricjs-project-stroke-working-forked-f5uzoh

image

image

If it looks OK, I apply the same logic to the other cases (bevel and round) and commit.

luizzappa:
This polygon controls is really come in handy!!

There is a weird stroke behavior that at first glance appeared to be a bug in the calculation of the projection of the stroke points. But investigating more deeply, I don't think is related to this PR.

When using a very high `strokeMiterLimit` and small angles, by varying the canvas zoom, the `stroke line join` switches from `miter` to `bevel`. In theory, the zoom wasn't meant to change that, right? This messes up the bounding rect calculation, since the miter limit has not been reached, it doesn't consider a bevel. Take a look:
bug_miterLimit

I thought it might even be a small width/height that could be cutting off part of the object (although it doesn't make much sense). I forced a much larger width/height, and even then the behavior remained the same:

bug_miterLimit2

I have no idea what could be causing this. But leaving the `strokeMiterLimit` to max up to 90 seems to be a safe value for this not to happen:

miterLimit_max90

ShaMan123:
good
I think we need to use `object.getTotalObjectScaling()` intead of `object.scaleX/Y` but it means that chaning the viewport requires a recalculation of the stroke projection...
fabric is very good at that kind of hook

I will merge master

ShaMan123:
Hmm I am not sure
I suggest we focus on the rest and leave this edge case for now

ShaMan123:
It looks great
https://codesandbox.io/s/fabricjs-project-stroke-working-forked-j4ug1k?file=/src/index.js

I left the bbox visible

Now, thanks to your example I understand we should only add and not subtract or need to calc a direction, cause the entire projection is mirrored so the miter vector always points to the opposite side of the bisector!

Nice

I will add more tests I guess but let's move forward.
What else do we need to cover for this to be complete? Do you want to try and tackle this edge case?
I

ShaMan123:
`strokeUniform` is still buggy

https://codesandbox.io/s/fabricjs-project-stroke-working-forked-od8xt7?file=/src/index.js:664-771


scale the object with the controls

ShaMan123:
and we should test it inside a group... to see if the relative plane messes up calculations

luizzappa:
This library is **AWESOME**!

Using the custom controls code, I added the vectors that are projected to each point (and the bisectors). This helped me to fix a bug in the `round` stroke. I think it will also help us understand `strokeUniform`. For that I needed to add a new optional argument in the `projectStrokeOnPoints` function to also returns the points and bisector that originated the projection. If you think it doesn't make sense just revert the commit https://github.com/fabricjs/fabric.js/pull/8083/commits/a960b45ca0fbab1e7e4fd99ced37bc44ef78ea38

Take a look: 

The red dotted vector is the bisector returned by the `getBisector` function. The green dotted vector is in the opposite direction and is what is used to determine which side to plot the projections. The projections are the colored vectors, one color for each point.

traceProjection

To make debugging easier, you can configure the parameters (stroke line join type, strokeUniform, miterLimit, etc..): https://codesandbox.io/s/project-stroke-points-with-trace-h8ltdf

luizzappa:
Regarding the `strokeUniform` is `true`, unfortunately, I believe we will have to call the `_setPositionDimensions` function every time a scaling occurs... I can't see any other way out, here's what happens:

When `strokeUniform` is `false`, a resize on the X axis only changes the width of the object.

traceProjection_not_strokeUniform

However, when `strokeUniform` is `true`, a resize on the X axis changes both the width and height of the object. See the two miter vectors coming out of the figure:
traceProjection_strokeUniform

Due to this behavior, we will always have to recalculate the projections, to find the new height and width. I can't imagine another alternative, what do you think?

**Note**: in my opinion the behavior of `strokeUniform` when equals to `false` is weird and when it is `true` makes more sense.. When you scale a triangle only on the X axis, its angle of the bisector changes, if this angle becomes more acute, the `miter vector` will grow. But see that in the first gif, the `miter vector` remains the same size. Seems like incorrect behavior to me, it distorts the triangle in a weird way. Anyway, this is canvas implementation...

luizzappa:
> I will add more tests I guess but let's move forward. What else do we need to cover for this to be complete? Do you want to try and tackle this edge case? I

I can try to find a solution for when the miter limit is high and the behavior of miter/bevel varies according to the zoom level. But will it be cool for the user experience?

If you always work with that zoom it makes sense. But imagine using the fabricjs library in a context like Canva.com or Figma, that your slide has a "real size" at 100% zoom, and you zoom out to be able to better align the objects on the slide. But when you zoom out the object changes size, it will look really weird.

src/util/misc.ts
```diff
@@ -195,56 +233,145 @@ import { cos } from './index.ts';
...
-          };
-      if (points.length <= 1) {return coords;}
+            new fabric.Point(1 / options.scaleX, 1 / options.scaleY) :
+            new fabric.Point(1, 1);
+
+      function scaleHatVector(hatVector, scalar) {
+        return hatVector.multiply(strokeUniformScalar).scalarMultiply(scalar);
+      }
+
+      function storeTraceProjections(projection, origin, bisector) {
```
ShaMan123, line 256: I don't mind returning more context
but let's do it always without the `traceProjection` flag
Also, point will become a class today so maybe we return an object instead of mutating it (TS will not like that and I agree)
```ts
{ projectedPoint, originPoint, miterVector, bisectorVector, ...context }
```

src/util/misc.ts
```diff
@@ -169,21 +189,39 @@ import { cos } from './index.ts';
     getBisector: function (A, B, C) {
```
ShaMan123, line 189: should we change the signature to reflect that A is between B and C?

src/util/misc.ts
```diff
@@ -195,56 +233,145 @@ import { cos } from './index.ts';
...
-            bisectorVector.y * scalar * strokeUniformScalar.y
-          );
-          if (Math.hypot(miterVector.x, miterVector.y) / s <= options.strokeMiterLimit) {
-            coords.push(A.add(miterVector));
-            coords.push(A.subtract(miterVector));
+          miterVector = scaleHatVector(bisectorVector, scalar);
+
+          var strokeMiterLimit;
+          if (options.strokeUniform) {
+            var miterLimitVector = scaleHatVector(bisectorVector, options.strokeMiterLimit * s);
```
ShaMan123, line 325: this is tricky logic as well
do you want to explain it so we discuss and see we understand and agree on it?

src/util/misc.ts
```diff
@@ -37,6 +37,23 @@ import { cos } from './index.ts';
...
+     * Chrome implements `Math#hypot` with a calculation that affects percision so we hard code it as a util
+     * @see https://stackoverflow.com/questions/62931950/different-results-of-math-hypot-on-chrome-and-firefox
+     * @static
+     * @memberOf fabric.util
+     * @param {...number}
+     * @returns {number}
+     */
+    hypot: function () {
+      var sumOfSquares = 0;
+      for (var i = 0; i < arguments.length; i++) {
```
ShaMan123, line 51: es6 now!
I am on it

ShaMan123 (commented):
I love the example you made
So we focus on `strokeUniform`
I will push some cleanup due to es6

src/util/misc.ts
ShaMan123: [813358e](https://github.com/fabricjs/fabric.js/pull/8083/commits/813358e8f12de02ffe30b47f7a512543172b3645)

ShaMan123:
> I can try to find a solution for when the miter limit is high and the behavior of miter/bevel varies according to the zoom level. But will it be cool for the user experience?
> 
> If you always work with that zoom it makes sense. But imagine using the fabricjs library in a context like Canva.com or Figma, that your slide has a "real size" at 100% zoom, and you zoom out to be able to better align the objects on the slide. But when you zoom out the object changes size (in the case of the gif, it decreases by half), it will look really weird.

I didn't get what you mean
And if I did that's exactly what `strokeUniform` flags

ShaMan123:
BTW it is important you get familiar with `_getTransformedDimensions` if you haven't yet
It plays a big role in bbox logic behind scenes... and it might be that an issue lies there as well

luizzappa:
> Regarding the `strokeUniform` is `true`, unfortunately, I believe we will have to call the `_setPositionDimensions` function every time a scaling occurs... I can't see any other way out, here's what happens:
> 
> When `strokeUniform` is `false`, a resize on the X axis only changes the width of the object.
> 
> traceProjection_not_strokeUniform 
> However, when `strokeUniform` is `true`, a resize on the X axis changes both the width and height of the object. See the two miter vectors coming out of the bounding rect: traceProjection_strokeUniform
> 
> Due to this behavior, we will always have to recalculate the projections, to find the new height and width. I can't imagine another alternative, what do you think?
> 
> **Note**: in my opinion the behavior of `strokeUniform` when equals to `false` is weird and when it is `true` makes more sense.. When you scale a triangle only on the X axis, its angle of the bisector changes, if this angle becomes more acute, the `miter vector` will grow. But see that in the first gif, the `miter vector` remains the same size. Seems like incorrect behavior to me, it distorts the triangle in a weird way. Anyway, this is canvas implementation...

@ShaMan123 did you understand this?

I will try to elaborate better. When the stroke is uniform and the object is scaled, the angle of the bisector vector changes. So we need to scale the points to find this new angle and correctly calculate the miter vector.

An example:
image


Scaling twice on the X axis, makes the angle of the bisector reduce (beta is less than
alpha), this makes the miter vector increase in size:
image

And why doesn't this same logic work when `strokeUniform` is `false`? That's what I said here ( https://github.com/fabricjs/fabric.js/pull/8083#issuecomment-1203478625 ). Scaling the object with this type of stroke has a different behavior, it just applies a resize on the x-axis only, there is no miter vector recalculation.

Something more or less like this happens: 
When `strokeUniform` is `false`, it only will apply a scaling on the X axis of point M. When `strokeUniform` is `true`, it recalculates the miter vector, hence it changes the y axis of point M as well.
image




Because of this behavior, for `strokeUniform` equals `false`, you only calculate the dimensions of the object once, and any scaling, the bounding box will be correct. This is not true for when the stroke is uniform (the miter vector changes so I need to recalculate the projections )
Take a look at the two gifs in this comment ( https://github.com/fabricjs/fabric.js/pull/8083#issuecomment-1203478625 ) and you will understand. 

For the case of `strokeUniform` equals `true`, you would need something like this:
```
 canvas.on('object:scaling', (e) => {
     const obj = e.target;
     var oldLeft = obj.left, oldTop = obj.top;
     obj._setPositionDimensions();
     obj.left = oldLeft;
     obj.top = oldTop;
 })
 ```

**Comment by luizzappa** [comment]
> > I can try to find a solution for when the miter limit is high and the behavior of miter/bevel varies according to the zoom level. But will it be cool for the user experience?
> > If you always work with that zoom it makes sense. But imagine using the fabricjs library in a context like Canva.com or Figma, that your slide has a "real size" at 100% zoom, and you zoom out to be able to better align the objects on the slide. But when you zoom out the object changes size (in the case of the gif, it decreases by half), it will look really weird.
> 
> I didn't get what you mean And if I did that's exactly what `strokeUniform` flags

Sorry, we're talking about two subjects at the same time, it confused you. This comment of mine was related to this other bug here: https://github.com/fabricjs/fabric.js/pull/8083#issuecomment-1200318054 . Not related to strokeUniform

File: src/util/misc.ts (hunk 9 above)
Comment (luizzappa) [reply src/util/misc.ts:311]: This part is kind of weird. It seems to me that the canvas recalculates the strokeMiterLimit when scaled with uniform stroke to maintain a proportion with the value initially set. 

When I kept the same strokeMiterLimit it didn't work.

File: src/util/misc.ts (hunk 8 above)
Comment (luizzappa) [reply src/util/misc.ts:176]: I hadn't thought of that, but it might be more intuitive

File: src/util/misc.ts [hunk 11]
```diff
@@ -195,56 +233,145 @@ import { cos } from './index.ts';
...
+          s = options.strokeWidth / 2,
+          scale = new fabric.Point(options.scaleX, options.scaleY),
           strokeUniformScalar = options.strokeUniform ?
-            new fabric.Point(1 / options.scaleX, 1 / options.scaleY) : new fabric.Point(1, 1),
-          getStrokeHatVector = function (v) {
-            var scalar = s / (Math.hypot(v.x, v.y));
-            return new fabric.Point(v.x * scalar * strokeUniformScalar.x, v.y * scalar * strokeUniformScalar.y);
-          };
-      if (points.length <= 1) {return coords;}
+            new fabric.Point(1 / options.scaleX, 1 / options.scaleY) :
```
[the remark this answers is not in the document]
luizzappa: why?

ShaMan123:
Let's focus on `strokeUniform` and then tackle the other one. Agreed?
I get what you are saying about the scale.
I figure we can try to override `_getTransformedDimensions` and recalculate the projection from there:
```ts
_getTransformedDimensions(...args) {
this.strokeUniform && this._setPositionDimension();
return this.callSuper('_getTransformedDimensions', ...args);
}

```

But to me it seems that applying a transform matrix on the miter vector should work (https://github.com/fabricjs/fabric.js/issues/7847#issuecomment-1173779421) and will be extremely more performant

ShaMan123:
It is proportional to scaling but matrix-wise I think
try 
```ts
fabric.util.transformPoint(miterVector, [1 / scaleX ,0 ,0 ,1 / scaleY],true)
```

luizzappa:
Got your point. 

I think an adjustment in the `_getTransformedDimensions` function so that both width and height are dependent on scaleX and scaleY when `strokeUniform` is `true` should do the trick. I just need to think better how to calculate this ratio...

ShaMan123:
updated from master
hope I didn't cause you a headache

ShaMan123:
I regret the update
I won't do that again until we are done

luizzappa:
I tried it here, it's more complicated than I thought...

For the case of `bevel`, it doesn't seem possible to create a transformation matrix. When scaling an object, the `strokeMiterLimit` can be reached and the `miter` stroke will become `bevel`. A transformation matrix would not handle this. In this case, with every scaling, `projectStrokeOnPoints` must be called to recalculate.

For the case of `round`, it is working correctly. There is no need to recalculate the projections. This is because it is easy to remove any scaling effect from projections (it's inversely proportional to scaling). So, when resizing it works without needing any kind of adjustment.
image

For the `bevel` case, I tried to remove all scaling effects from the projections, but with no success, mathematically it's complex for me.. If we could multiply each of the axes by a factor that removed the scale from the equations and multiply by that same factor in `_calcDimensions` and `_getTransformedDimensions` it would work. 

image

luizzappa:
I'm going this way...

> I figure we can try to override `_getTransformedDimensions` and recalculate the projection from there:
> 
> ```ts
> _getTransformedDimensions(...args) {
> this.strokeUniform && this._setPositionDimension();
> return this.callSuper('_getTransformedDimensions', ...args);
> }
> ```

Something along these lines:

```ts
_getTransformedDimensions(...args) {
    this.strokeUniform && this.strokeLineJoin !== 'round' && this._setPositionDimension();
    return this.callSuper('_getTransformedDimensions', ...args);
}
```
I pulled the last commits but the build no longer works.. Is it something related to typescript (I don't have much experience)?

image

ShaMan123:
try again now [e4d5a74](https://github.com/fabricjs/fabric.js/pull/8083/commits/e4d5a74110599dcf931938606d648025b4eefc4e)

ShaMan123:
Is build passing now?

luizzappa:
Same error... 

The circular dependency seems to continue:
image

luizzappa:
@ShaMan123, looks like it's working now. However, it is still not good to call the `_setPositionDimensions` function in the current way. It's called many times unnecessarily..

For example, while you are moving an object, you would not need to recalculate the bbox, but it will recalculate (call `_setPositionDimensions` function) not only for the moving object, but also for all other objects on canvas..

If it's to stay like this, it seems more performant to treat on the application side with something like this:
```
 canvas.on('object:scaling', (e) => {
     const obj = e.target;
     var oldLeft = obj.left, oldTop = obj.top;
     obj._setPositionDimensions();
     obj.left = oldLeft;
     obj.top = oldTop;
 })
```

luizzappa:
I'll be on vacation for 14 days, feel free to continue, or we'll wait for my return. When I return, I'll create some test cases (I haven't had time to look into this yet)

ShaMan123:
I have a few ideas
I'll try some out on different branches
This PR is inspired by you #8135 - hope you'll enjoy it once it's merged
Enjoy your vacation
I really appreciate your fine work

asturur:
hi @luizzappa meanwhile thanks for all this hard work.
We are a bit busy with the TS migration, that needs to be done in order to move forward, we will get back soon to this

luizzappa:
Instead of overriding the `_getTransformedDimensions` method, I did it in the `_set` method. In this way, the calculation of the projection of the stroke points happens only when resizing the object. It's much more performative.

```
    _set(key, value) {
      var output = this.callSuper('_set', key, value);
      (key === 'scaleX' || key === 'scaleY') && this.strokeUniform && this.strokeLineJoin !== 'round' && this._setPositionDimensions();
      return output; 
    }
```

What do you think? For me it's good to go.. I look forward to your review when you have time

src/shapes/polyline.class.ts
```diff
@@ -78,15 +78,23 @@
...
+          correctSizeY = this.exactBoundingBox
+            ? this.strokeUniform
+              ? this.strokeWidth / this.scaleY
+              : this.strokeWidth
+            : 0;
+      this.width = calcDim.width - correctSizeX;
+      this.height = calcDim.height - correctSizeY;
       if (!options.fromSVG) {
         correctLeftTop = this.translateToGivenOrigin(
           {
```
ShaMan123, line 95: let's make this a Point

src/shapes/polyline.class.ts
```diff
@@ -101,8 +109,8 @@
         this.top = options.fromSVG ? calcDim.top : correctLeftTop.y;
       }
       this.pathOffset = {
```
ShaMan123, line 111: same - make a Point

src/shapes/polyline.class.ts
```diff
@@ -118,7 +126,7 @@
      */
     _calcDimensions: function() {
 
-      var points = this.exactBoundingBox ? this._projectStrokeOnPoints() : this.points,
+      var points = this.exactBoundingBox ? this._projectStrokeOnPoints().map(elem => elem.projectedPoint) : this.points,
```
ShaMan123, line 129: So now after your elegant fix is there a need to return anything else than the projected points?

src/shapes/polyline.class.ts
```diff
@@ -134,6 +142,16 @@
       };
     },
 
+    /**
+     * After setting scale, recalculates dimensions when the stroke is uniform and stroke line join is bevel or round
+     * @private
+     */
+    _set(key, value) {
+      var output = this.callSuper('_set', key, value);
```
ShaMan123, line 150: `output` equals `this` currently
But maybe it will change.
Good enough, let's leave as is.

src/util/misc.ts
```diff
@@ -197,56 +219,156 @@ import { cos } from './cos';
...
+              ),
+              hatOrthogonalVector = fabric.util.getOrthogonalUnitVector(vector),
+              orthogonalVector = scaleHatVector(hatOrthogonalVector, s);
+
+          var proj1 = A.add(orthogonalVector),
+              proj2 = A.subtract(orthogonalVector);
+
+          [proj1, proj2].forEach(proj => {
+            coords.push({ 
+              'projectedPoint': proj,
```
ShaMan123, line 274: code style: remove `'`

src/util/misc.ts
```diff
@@ -197,56 +219,156 @@ import { cos } from './cos';
...
+              'originPoint': A,
+              'bisector': bisector
```
ShaMan123, line 276: es6 now allows use to write just `bisector`

src/util/misc.ts
```diff
@@ -197,56 +219,156 @@ import { cos } from './cos';
...
+            });
+          });
+
+          return;
+        }
+
+        var bisector,
+            scaledA,
```
ShaMan123, line 284: I prefer that from this point we use a single set of variables intead of `A`, `B`, `C` and `scaledA`, `scaledB`, `scaledC`.

src/util/misc.ts
```diff
@@ -171,21 +177,37 @@ import { cos } from './cos';
...
       var AB = fabric.util.createVector(A, B), AC = fabric.util.createVector(A, C);
       var alpha = fabric.util.calcAngleBetweenVectors(AB, AC);
-      //  check if alpha is relative to AB->BC
-      var ro = fabric.util.calcAngleBetweenVectors(fabric.util.rotateVector(AB, alpha), AC);
-      var phi = alpha * (ro === 0 ? 1 : -1) / 2;
       return {
-        vector: fabric.util.getHatVector(fabric.util.rotateVector(AB, phi)),
-        angle: alpha
+        vector: fabric.util.getHatVector(fabric.util.rotateVector(AB, alpha / 2)),
+        angle: Math.abs(alpha)
```
ShaMan123, line 182: For the sake of the pure method, not taking into consideration how we use it, I think we shouldn't return an abs angle. 
Am I missing something?
Why is it abs? I saw where it is used, maybe we should abs it there.
Thoughts?

ShaMan123 (changes requested):
You have done it!
I thought we had another big challenge to overcome but you nailed it.
I left some comments regarding styling. And a few more.

Meanwhile we managed to migrate quite a lot of fabric to TS/es6.
So let's do it here as well.
No vars, spread object etc.
`Point` can be used more easily `new Point(p)` instead of `new Point(p.x,p.y)`
Imports and exports
But first we need to update from master, so keep on reading.

The file structure has changed as well.
So I would like you to refactor the code so it suits master.
I don't know what is best. 
Creating a new branch from master (`origin/master` for your fork) and moving all this work to there and refactoring it and then you merge into this. Or running an update from master, resolving conflicts (`src/util/misc.ts` for example doesn't exist anymore I think) and then refactoring the code.
Anyways I would create a branch that will save the current state in case 💩 hits the fan so we can have something to revert to.
Vector methods have been moved to [`src/util/misc/vectors.ts`](https://github.com/fabricjs/fabric.js/blob/master/src/util/misc/vectors.ts)
Stroke projection has been moved to [`src/util/misc/projectStroke.ts`](https://github.com/fabricjs/fabric.js/blob/master/src/util/misc/projectStroke.ts) 
and more.
If you want me to give a hand just say so.

After that I think we can focus on testing.

In addition, I would like you to write meaningful comments for devs in the code, where ever you think is needed (I might comment where I think should be an explanation) so people can understand the logic.
If you feel it is an overkill maybe we leave as is and point to the PR.
What do you think?

src/shapes/polyline.class.ts
```diff
@@ -142,6 +142,16 @@
...
+      (key === 'scaleX' || key === 'scaleY') && this.strokeUniform && this.strokeLineJoin !== 'round' && this._setPositionDimensions();
```
ShaMan123, line 151: does `skew` affect as well?

ShaMan123: Seems we didn't take that into account.
Want to focus on it in a follow up or here?

ShaMan123: skew bbox is inaccurate regardless I think

lovegnep:
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Title</title>
    <style type="text/css">
        body {
            background-color: ivory;
        }

        canvas {
            border: 1px solid red;
        }
    </style>
    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.9.1/jquery.min.js"></script>
</head>
<body>
    <div>Resize left canvas</div>
    <input id=myslider1 type=range><br>
    <div>Resize right canvas</div>
    <input id=myslider2 type=range><br>
    <h4 id=mouse>Mouse coordinates:</h4>
    <canvas id="canvas1"  width=300 height=300></canvas>
    <canvas id="canvas2" width=300 height=300></canvas>

    <script>
        const BRUSH_COLOR = '#ffcc00bb'
        let obj = {lineGroup:[]}
        var canvas1 = document.getElementById("canvas1");
        var ctx1 = canvas1.getContext("2d");
        var canvas2 = document.getElementById("canvas2");
        var ctx2 = canvas2.getContext("2d");
        var originalWidth = canvas1.width;
        var originalHeight = canvas1.height;

        var scale1 = 1;
        var scale2 = 1;

        $myslider1 = $('#myslider1');
        $myslider1.attr({min: 50, max: 200}).val(100);
        $myslider1.on('input change', function () {
            var scale = parseInt($(this).val()) / 100;
            scale1 = scale;
            redraw(ctx1, scale);
        });
        $myslider2 = $('#myslider2');
        $myslider2.attr({min: 50, max: 200}).val(100);
        $myslider2.on('input change', function () {
            var scale = parseInt($(this).val()) / 100;
            scale2 = scale;
            redraw(ctx2, scale);
        });



        // draw(ctx1);
        // draw(ctx2);

        function redraw(ctx, scale) {

            // Resizing the canvas will clear all drawings off the canvas
            // Resizing will also automatically clear the context
            // of all its current values and set default context values
            ctx.canvas.width = originalWidth * scale;
            ctx.canvas.height = originalHeight * scale;

            ctx.save()
            // context.scale will scale the original drawings to fit on
            // the newly resized canvas
            ctx.scale(scale, scale);

            drawLines(ctx, obj.lineGroup)

            // always clean up! Reverse the scale
            // ctx.scale(-scale, -scale);
            ctx.restore();

        }

        function draw(ctx) {
            // note: context.scale causes canvas to do all the rescaling
            //       math for us, so we can always just draw using the
            //       original sizes and x,y coordinates
            ctx.beginPath();
            ctx.moveTo(150, 50);
            ctx.lineTo(250, 150);
            ctx.lineTo(50, 150);
            ctx.closePath();
            ctx.stroke();
            ctx.fillStyle = 'skyblue';
            ctx.beginPath();
            ctx.arc(150, 50, 20, 0, Math.PI * 2);
            ctx.closePath();
            ctx.fill();
            ctx.stroke();
            ctx.beginPath();
            ctx.arc(250, 150, 20, 0, Math.PI * 2);
            ctx.closePath();
            ctx.fill();
            ctx.stroke();
            ctx.beginPath();
            ;
            ctx.arc(50, 150, 20, 0, Math.PI * 2);
            ctx.fill();
            ctx.stroke();
        }

        $("#canvas1, #canvas2").mousemove(function (e) {
            onmousemove(e)
            handleMouseMove(e);
        });

        $("#canvas1, #canvas2").mousedown(function (e) {
            obj.pain = true
            onmousedown(e)
        });

        $("#canvas1, #canvas2").mouseup(function (e) {
            obj.pain = false
        });
        var $mouse = $('#mouse');
        function mouseXY(ev) {
            console.log("mouseXY: clinetXY:", ev.clientX, ev.clientY);
            var bb = ev.target.getBoundingClientRect();
            console.log("mouseXY: bb left top:", bb.left, bb.top);

            let mouseX = parseFloat(ev.clientX - bb.left);
            let mouseY = parseFloat(ev.clientY - bb.top);
            console.log("mouseXY: result xy:", mouseX, mouseY);

            return { x: mouseX, y: mouseY }
        }
        function onmousedown(e){
            let lineGroup = []
            if (obj.lineGroup.length > 0) {
                lineGroup = [...obj.lineGroup]
            }
            lineGroup.push({ size: 10, pts: [mouseXY(e)] })
            obj.lineGroup = lineGroup;
            console.log("onmousedown:lineGroup:", lineGroup);

            drawLines(ctx1, lineGroup)
        }
        function onmousemove(e){
            if (obj.lineGroup.length <= 0 || obj.pain ===false){
                return;
            }
            const lineGroup = [...obj.lineGroup]
            lineGroup[lineGroup.length - 1].pts.push(mouseXY(e))
            obj.lineGroup = lineGroup;
            console.log("onmousemove:lineGroup:", lineGroup);
            drawLines(ctx1, lineGroup)
        }
        function drawLines(ctx,lines,color = BRUSH_COLOR) {
            ctx.strokeStyle = color
            ctx.lineCap = 'round'
            ctx.lineJoin = 'round'

            lines.forEach(line => {
                if (!line.pts.length || !line.size) {
                    return
                }
                ctx.lineWidth = line.size
                ctx.beginPath()
                ctx.moveTo(line.pts[0].x, line.pts[0].y)
                line.pts.forEach(pt => {
                    ctx.lineTo(pt.x, pt.y);
                    console.log(`lineto:(${pt.x},${pt.y})`)
                })
                ctx.stroke()
            })
        }

        function handleMouseMove(e) {

            // tell the browser we're handling this event
            e.preventDefault();
            e.stopPropagation();


            var bb = e.target.getBoundingClientRect();
            mouseX = parseInt(e.clientX - bb.left);
            mouseY = parseInt(e.clientY - bb.top);

            if (e.target.id == 'canvas1') {
                $mouse.text('Mouse1: ' + mouseX / scale1 + ' / ' + mouseY / scale1 + ' (scale:' + scale1 + ')');
            } else {
                $mouse.text('Mouse2: ' + mouseX / scale2 + ' / ' + mouseY / scale2 + ' (scale:' + scale2 + ')');
            }

        }
    </script>
</body>
</html>

ShaMan123:
@lovegnep you should provide an explanation and a summary, no one should guess why you posted this code here

ShaMan123:
@luizzappa ?

luizzappa:
@ShaMan123 ,  I'm working on it. It's more difficult than I initially imagined... 

I got the `miter` and `bevel` cases, but the `round` is still missing. 

When `skew` is applied, the semicircle becomes an ellipse, and we need to find the farthest points from that ellipse. 

Picture1


I'm going to take this weekend to review some concepts of geometry and calculus, I believe that with the ellipse equation we can take the derivative and find the tangents parallel to the x-axis and y-axis that are candidates for more distant points.

ShaMan123:
Alright
I am thinking we could merge it with skew broken and return to that (something I learned lately from @asturur...)
This PR, your work, is a huge leap forward in bbox calculations and accuracy.
It is important to feel accomplishment as well and merging is a good and decent way to do that. It will give more motivation to tackle the next issue.
I am more focused on updating from master, es6, types etc. It is more important IMO at this point.
If you'd like I can migrate it and produce a diff for you to merge into the PR (you deserve the credit)
Open to your thoughts and will.

ShaMan123:
Just airing some thoughts regarding `round` + `skew`
Your code assumes that both axis are orthogonal. That is why it is enough to add the radius of each axis to a point, getting 2 projected points covering the intersection of the tangent lines of the curve.
And that is the problem when skew is applied, making the axis not orthogonal, am I right so far?

ShaMan123:
If that is the case why don't we intersect the 2 vectors after they are transformed by the plane?

ShaMan123:
Ok, now I understand why you want to look for the axis tangents.
Intersection isn't accurate enough... but the tangents are.

Isn't the bisector orthogonal to the "tip" of the semicircle?
Could it mean that the "tip" point is our point of interest regardless of the shear transform?
If this is correct (I am quite positive I have a huge hole in my reasoning) once we transform that point with the shear matrix we should get the farthest point.

Another direction:
Look at the path util function `getBoundsOfCurve`. We can use that (calculating the control point doesn't sound too complex)

luizzappa:
> Alright I am thinking we could merge it with skew broken and return to that (something I learned lately from @asturur...) This PR, your work, is a huge leap forward in bbox calculations and accuracy. It is important to feel accomplishment as well and merging is a good and decent way to do that. It will give more motivation to tackle the next issue. I am more focused on updating from master, es6, types etc. It is more important IMO at this point. If you'd like I can migrate it and produce a diff for you to merge into the PR (you deserve the credit) Open to your thoughts and will.

I'm in the flow, give me a few more days, if I can't solve this problem, I focus on refactoring and PR. I will appreciate your help, I have little experience in typescript

luizzappa:
> Ok, now I understand why you want to look for the axis tangents. Intersection isn't accurate enough... but the tangents are.
> 
> Isn't the bisector orthogonal to the "tip" of the semicircle? Could it mean that the "tip" point is our point of interest regardless of the shear transform? If this is correct (I am quite positive I have a huge hole in my reasoning) once we transform that point with the shear matrix we should get the farthest point.

I tried using the point that is in the direction of the bisector before the transformation and then applying the skew (I tried the opposite too), but as the new position of the X depends on the Y height the point is at, it will not necessarily be the solution. Take a look:

skew with bisector

> Another direction: Look at the path util function `getBoundsOfCurve`. We can use that (calculating the control point doesn't sound too complex)

I'll take a look, thanks! Mathematically I've already calculated the points that have tangents parallel to the X and Y axes. Now it's translating, I believe I'm close to solving

image

luizzappa:
I  adjusted the `round` line join with skew. However, it only works when `skewX` and `skewY` are applied separately. If the two are applied together, it's still buggy. But I wonder what must be happening. Let's stop here, continue with the PR and then I'll think about this case again

image

It's a little different than what I initially imagined. The semicircle does not become the segment of an ellipse. The logic behind it is to work with the equation that generates the position of the point after the skew is applied. For example, `skewX`, the final point X will be:

$X_f = X + Y * Tan(skewX)$

Limited by the equation (the circle of the `round` line join):

$Y^2 = r^2 - X^2$

Substituting the second equation into the first e and derivative:

$X = \pm \frac{r}{\sqrt{1 + Tan^2(skewX)}}$

This is the value of X before applying `skewX` which will be the furthest distance after applying skewX. The same logic follows for Y.

luizzappa:
My code is a mess. Tomorrow I will adjust it, and commit to this PR to have a valid version. Then I open a new PR by cloning from the master and make the adjustments.

ShaMan123:
Outstanding.

I will take the advantage and use this PR for learning a bit if you don't mind.
So I will ask a few question to see if I understand.

Are you looking for a derivative for 2 variables?

Is it correct to assume that when `d/dy` equals 0 it is an asymptote for `d/dx` and vice versa? If so that can simplify calculations

luizzappa:
> Outstanding.
> 
> I will take the advantage and use this PR for learning a bit if you don't mind. So I will ask a few question to see if I understand.
> 
> Are you looking for a derivative for 2 variables?
> 
> Is it correct to assume that when `d/dy` equals 0 it is an asymptote for `d/dx` and vice versa? If so that can simplify calculations
> 
> **Correction** Is it correct to assume that when `d/dy` equals 0 it is an asymptote for the function plotting `x` as well as to `d/dx` and vice versa? If so that can simplify calculations since finding the asymptote sounds pretty straight forward in a root function.
> 
> Are these functions inverse to each other?

No, the equation is one-variable. I don't use implicit differentiation.

$f(X) = X \pm \sqrt{r^2 - X^2} * Tan(skewX)$

And are inverse:

image

luizzappa:
I'll take the opportunity to detail it better for future documentation.

Consider this case:
image



There is a point $(X_i, Y_i)$ that will be the furthest X distance from the origin after applying skewX. This is our point of interest.
image

We have:

**[eq.1]** $X_{afterSkewX} = X_i + Y_i * Tan(skewX)$

And we know that $Y_i$ is a function of $X_i$:

$Y_i^2 + X_i^2 = r^2$ 
**[eq.2]** $Y_i = \pm \sqrt{r^2 - X_i^2}$

Substituting eq.2 in eq.1 we have:

$X_{afterSkewX} = X_i \pm \sqrt{r^2 - X_i^2} * Tan(skewX)$

So we have an equation in terms of just X:

$f(X_i) = X_i \pm \sqrt{r^2 - X_i^2} * Tan(skewX)

The plus/minus sign represent both sides of the circle after skewX:
image

The furthest X distance from the origin are the maximum/minimun of $f(X_i)$:
image

Computing $f'(X_i)=0$ for both cases, we find:

$X_i = \pm \frac{r}{\sqrt{1 + Tan^2(skewX)}}$

Substituting $X_i$ into the circle equation, we find $Y_i$:

$Y_i = \pm \sqrt{r^2 - X_i^2}$

That's our solution:

image

I'm considering the two points for projection, this can be a problem when the control points are too close. I'll adjust to consider only the point belonging to the yellow segment.

ShaMan123:
Got it.

> The furthest X distance from the origin are the maximum/minimun of $f(X_i)$:

```diff
- The furthest X distance from the origin are the maximum/minimun of $f(X_i)$:
+ The furthest Y distance from the origin are the maximum/minimun of $f(X_i)$:
```

???

luizzappa:
@ShaMan123 

I started working on `projectStroke.ts`.

I don't know much about typescript. In the `projectStrokeOnPoints` function you are indicating that the function returns an array of Point, right? 

```
export const projectStrokeOnPoints = (
  points: Point[],
  {
    strokeWidth,
    strokeLineJoin,
    strokeMiterLimit, // https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit
    strokeUniform,
    scaleX,
    scaleY,
  }: projectStrokeOnPointsOptions,
  openPath: boolean
): Point[] => {
```

I changed the return to a context:
```
coords.push({
                            'projectedPoint': applySkew(A.add(vector)),
                            'originPoint': A,
                            'bisector': bisector
                        });
(...)
return coords
```

How would I indicate this to typescript? Googling I think it would be something like this:
```
export const projectStrokeOnPoints = (
//(...)
):  Array<{projectedPoint: Point, originPoint: Point, bisector: Point}> => {
```

ShaMan123:
TS can handle that under the hood. I don't bother adding a return value unless TS is confused about (some devs like it because it makes the signature strict).
If you hover over it TS should inform you what it asserts it is.
So I would say focus on args. But someone might call me out on that. So for now whatever you choose.

What you wrote is fine, a syntax alias to `Array<T>` is `T[]` and we prefer that but not a big deal

ShaMan123:
I referred you to the new guides, right? #8189 

Take a look at setting up prettier if you want it seamlessly integrated into the IDE (instead of calling `npm run prettier:write`)
https://github.com/fabricjs/fabric.js/blob/899178603653794f811d9c20ac3e387c07d8cc71/CONTRIBUTING.md#-guidelines

luizzappa:
> Got it.
> 
> > The furthest X distance from the origin are the maximum/minimun of f(Xi):
> 
> ```diff
> - The furthest X distance from the origin are the maximum/minimun of $f(X_i)$:
> + The furthest Y distance from the origin are the maximum/minimun of $f(X_i)$:
> ```
> 
> ???

The text got a little confusing because of the axes of the graph. In terms of the graph it is the farthest Y, but Y = X after skew X.

The graph plots on the X-axis the Xi value (X before skewX), and on the Y-axis the X value after skewX. So the max/min of $f(X_i)$ is the X after the skewX furthest from the origin in the "X-axis direction".

Perhaps a better wording would be: "The furthest X after skewX distance from the origin are the maximum/minimun of f(Xi)".

ShaMan123:
I have developed an apatite for your explanations.
If you'd like to look at `qrDecompose` and make that available to the a mass I know by fact that @asturur would love that and so would I. 
I googled it and dived into a maze of linear algebra, gram-schmidt, eigen values/vectors, orthonormal sets and came out a few hours later as entering.

I love [khanacademy](https://www.khanacademy.org/math/linear-algebra/alternate-bases/orthonormal-basis/v/linear-algebra-introduction-to-orthonormal-bases)
but haven't found out enough yet

luizzappa:
I'll take a look later. What I know of linear algebra is from college (engineering). Depending on the field, there are things that I never studied.. But is there a problem with it or is it just understanding how it works?

ShaMan123:
It works and is correct
We face accuracy issues but I don't think there is much to do

luizzappa:
@ShaMan123, I'm not finding hypot in util, is it implemented? If not, tell me where I put it

There was that precision problem:

```
/**
 * Returns the square root of the sum of squares of its arguments\
 * Chrome implements `Math#hypot` with a calculation that affects percision so we hard code it as a util
 * @see https://stackoverflow.com/questions/62931950/different-results-of-math-hypot-on-chrome-and-firefox
 * @static
 * @memberOf fabric.util
 * @param {...number}
 * @returns {number}
 */
function hypot(...values) {
    let sumOfSquares = 0;
    for (let i = 0; i < values.length; i++) {
        sumOfSquares += values[i] * values[i];
    }
    return Math.sqrt(sumOfSquares);
}
```

ShaMan123:
Yes
It is something we fixed here. I did add a lint rule to block using Math.hypot.
Under util/misc you can add a dedicated file or a math file. Whatever.
@asturur refactored utils as is so feel free to remove things.
All logic under projectStroke should be dumped and replaced by your work

luizzappa:
@ShaMan123, As I had to start from scratch from master, I opened a new PR with the branch I worked on. If you think it's better to continue with this PR, I'll close this new one

ShaMan123:
I am fine with what you decide.
I can use github api to copy the thread. Or it can be turned into a wiki. I like that idea. An md file with all this crucial knowledge.

ShaMan123:
But isn't a must

PR #8087, opened 2022-07-27
test(#8083)

ShaMan123:


luizzappa:
@ShaMan123 As the canvas node has a bug with the stroke miter limit calculation, I created this condesandbox to generate the goldens via browser: https://codesandbox.io/s/generate-goldens-stroke-projection-tcn10x . It generates a zipped file.

I added concave cases to the test too

Since I can't commit to this PR, I commited on the `pull_8087` branch of this fork here: https://github.com/luizzappa/fabricjs-project-uniform-stroke/tree/pull_8087

A strange behavior is occurring in the tests...

If I just run the `stroke projection` module tests, they all pass: 
http://127.0.0.1:4000/test/visual/?coverage&moduleId=a40b012b

image

But when I run **all modules** tests at the same time, some 'stroke projection' tests fail:
http://127.0.0.1:4000/test/visual/?coverage&hidepassed

Analyzing the difference in the images, there is a slight distortion at rectangle edges that **doesn't happen when we run all the tests**:
image


Zooming:
image

The curious thing is that this distortion is present in goldens and also when we only run the tests of the `stroke projection` module, so these tests do not fail. But when we run all tests of all modules at the same time, this distortion disappears from the canvas, that makes the test fail. It is as if there is some property from a previous test that is spilling over to the `stroke projection` test.

ShaMan123:
@luizzappa glad to see you back
Hope you had a great vacation
I will look into your comment tomorrow
And with your permission I will update from master (I have upgraded the test suite so it now generates goldens from the browser without hassle - sorry for the work you put into that, hope it wasn't a time sink)

luizzappa:
Ah cool! I hadn't seen your update on the generation of goldens (actually now that I've looked at the code more calmly, and I've seen that if the golden doesn't exist, it generates it for you.. I thought they were generated externally)

Feel free to update. I ended up splitting the test into concave and convex cases.

Tomorrow we'll take care of it

luizzappa:
> (I have upgraded the test suite so it now generates goldens from the browser without hassle - sorry for the work you put into that, hope it wasn't a time sink)

@ShaMan123, did this update remain after migrating to typescript? I'm trying to run the tests, but it seems to me that the goldens were generated by node (which has that problem with miter-limit)

image

ShaMan123:
Did you open a new PR?

luizzappa:
> Did you open a new PR?

I was working locally. Just opened it here: https://github.com/fabricjs/fabric.js/pull/8374

PR #8097, opened 2022-07-28
chore(`Object#clearContextTop`): negate `manuallyRestore`

ShaMan123:
Instead of `clearContextTop(manuallyRestore)`  => `clearContextTop(restore)` 
Follow up for #7802 

closed #8096


ShaMan123:
Will be done by es6

PR #8100, opened 2022-07-31
chore(TS): main patch

ShaMan123:
Mainly es6 rework. a bit of typescript, very little

Includes:
- `shapes` - all passing except `Image` because it needs filters to pass
- `mixins` - without eraser
- `brushes`
- `filters`
- `color`
- `static_canvas`
- `point`, `intersection` et al.
- `parser`

The `context` module needs work (refactor from `HEADER`)
I couldn't get a conditional import to work

TESTS FAIL
I need help
TOO MUCH WORK


ShaMan123:
Maybe for a proper diff we should merge a pr with the transfomer output
and then patch on top of that so git understands the files are the same

ShaMan123:
fed fabric from this, but now it has become stale
so no more

PR #8101, opened 2022-07-31
Convert Rect to fake ts

asturur:


src/shapes/Rect.ts
```diff
@@ -0,0 +1,178 @@
...
+/* _FROM_SVG_END_ */
+
+/**
+ * Returns {@link fabric.Rect} instance from an object representation
+ * @static
+ * @memberOf fabric.Rect
+ * @param {Object} object Object to create an instance from
+ * @returns {Promise<fabric.Rect>}
+ */
+Rect.fromObject = function(object) {
```
ShaMan123, line 174: make these static on the class

ShaMan123 (commented):
looks OK

PR #8102, opened 2022-07-31
ci(scripts) patch

ShaMan123:
build fast option
patch transformer indexing


ShaMan123:
WHY the *** are tests failing?

asturur (approved):

PR #8103, opened 2022-07-31
chore(TS): Observable

ShaMan123:
- fixes observable off - now removes the handler instead of removing the handler after `fire`
- ~~`observable` tests pass accept 2 on active selection stuff~~ **FIXED `once` migration error**
- ~~I can't manage to extend Object/StaticCanvas with the class/proto/instance of class `Observable`. Nothing works.~~
- make it a class so it becomes a super class for Object/Canvas for overriding needs etc.


src/shapes/object.class.js
```diff
@@ -1905,9 +1907,13 @@
...
 
-  fabric.util.createAccessors && fabric.util.createAccessors(fabric.Object);
+  // fabric.util.createAccessors && fabric.util.createAccessors(fabric.Object);
 
-  extend(fabric.Object.prototype, fabric.Observable);
+  fabric.Object.prototype.on = Observable.prototype.on;
+  fabric.Object.prototype.once = Observable.prototype.once;
+  fabric.Object.prototype.off = Observable.prototype.off;
+  fabric.Object.prototype.fire = Observable.prototype.fire;
+  fabric.Object.prototype._removeEventListener = Observable.prototype._removeEventListener;
```
asturur, line 1916: Why not Object.assign(fabric.Object.prototype, fabric.Observable) ?

src/util/lang_object.js
```diff
@@ -72,5 +72,5 @@
     extend: extend,
     clone: clone
   };
-  fabric.util.object.extend(fabric.util, fabric.Observable);
+  // fabric.util.object.extend(fabric.util, fabric.Observable); //snot sure why this is here
```
asturur, line 75: just trash it probably a relic of the past

src/mixins/observable.mixin.js
```diff
@@ -1,145 +0,0 @@
...
-   * @namespace fabric.Observable
-   * @tutorial {@link http://fabricjs.com/fabric-intro-part-2#events}
-   * @see {@link http://fabricjs.com/events|Events demo}
-   */
-  fabric.Observable = {
-    fire: fire,
-    on: on,
-    once: once,
-    off: off,
-  };
```
asturur, line 144: Observable was an object with 4 methods. Why does to become a class?

src/mixins/Observable.ts
```diff
@@ -0,0 +1,142 @@
+//@ts-nocheck
+
+type EventRegistryObject = { [eventName: string]: Function };
```
asturur, line 3: I think there is a a TS util, Record, that makes this more clear.

test/unit/observable.js
```diff
@@ -1,19 +1,19 @@
 QUnit.module('fabric.Observable');
 
-QUnit.test('fabric.Observable exists', function(assert) {
+QUnit.test('fabric.Observable exists', function (assert) {
   assert.ok(fabric.Observable);
-  assert.ok(fabric.Observable.fire);
-  assert.ok(fabric.Observable.on);
-  assert.ok(fabric.Observable.once);
-  assert.ok(fabric.Observable.off);
+  var o = new fabric.Observable();
```
asturur, line 5: i undertstand this looks better, but is a mixin really supposed to to be initialized on its own?

index.js
```diff
@@ -1,100 +1,105 @@
...
-import './src/mixins/itext.svg_export.js'; // optional itext
-import './src/shapes/textbox.class.js'; // optional textbox
-import './src/mixins/default_controls.js'; // optional interaction
+import { fabric } from './HEADER';
+// import './lib/event'), // optional gestures
+import './src/mixins/collection.mixin';
+
+// import './src/mixins/Observable';
+import { Observable } from './src/mixins/Observable';
+fabric.Observable = Observable;
```
asturur, line 7: i think is safe to bring all of those up, Observable does not seem to need to come after collection.mixin.
I would also rather do the assign to fabric inside the file itself.

ShaMan123: > I would also rather do the assign to fabric inside the file itself.

I am against for a few reasons:
- after we complete the migration we won't be assigning to fabric
- limiting the mess to the index which will change a lot anyways

src/shapes/object.class.js
ShaMan123: I tried that but it doesn't work

src/mixins/observable.mixin.js
ShaMan123: my intention is to make CommonMethods extend Observable

src/mixins/Observable.ts
ShaMan123: great

test/unit/observable.js
ShaMan123: > my intention is to make CommonMethods extend Observable

src/shapes/object.class.js
ShaMan123: because methods aren't enumerable on class proto

src/mixins/observable.mixin.ts
```diff
@@ -1,146 +1,157 @@
...
+      const eventName = arg0;
       if (!this.__eventListeners[eventName]) {
         this.__eventListeners[eventName] = [];
       }
       this.__eventListeners[eventName].push(handler);
+      return () => this.off(eventName, handler);
+    }
+    else {
+      // noop
+      return () => false;
```
ShaMan123, line 47: not sure we need this
just in case a dev does something illegal

ShaMan123 (commented):
READY

test/unit/observable.js
ShaMan123: line node's `EventEmitter`

ShaMan123:
closed by #8322

PR #8107, opened 2022-08-02
Convert rect simple

asturur:
Rect simple changes to adapt to import/export and es6


src/shapes/rect.class.ts
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Returns {@link fabric.Rect} instance from an object representation
+ * @static
+ * @memberOf fabric.Rect
+ * @param {Object} object Object to create an instance from
+ * @returns {Promise<fabric.Rect>}
+ */
+Rect.fromObject = (object) => fabric.Object._fromObject(Rect, object);
+
+fabric.Rect = Rect;
```
ShaMan123, line 178: do you mind if we assign this in the index?

asturur: i prefer to just delete this when the index goes away.
Otherwise while working on different files we do useless merge conflicts between each other

ShaMan123 (dismissed):

PR #8108, opened 2022-08-02
chore(TS): convert file ext

ShaMan123:
- convert file ext under `src`
- `@ts-nocheck` for build to pass
- build script patch - to kill build if a new build is about to start in watch mode


ShaMan123:
it's just lint that is failing because TS

.eslintrc.json
```diff
@@ -160,6 +160,14 @@
       {
         "ignoreComments": true
       }
-    ]
+    ],
+
+
+    //  TODO: remove the following for TS, just temporary 08/22
```
ShaMan123, line 166: these actually do a really good job
they convert whatever they can from var to let/const
I didn't commit to leave it clean

ShaMan123 (commented):
couldn't get lint to pass no matter what

asturur (approved):

PR #8110, opened 2022-08-02
dep(): `includeDefaultValues`

ShaMan123:
## Gist
This PR aims to move the responsibility of removing the  default values of `toObject` return value from fabric to the dev.
It is actually a pretty good idea IMO
**We gain not needing default values to check against**

1. For that I have exposed utils (Let the dev handle it!!)
- `removeDefaultValues(object, defaults)`
- `removeObjectDefaultValues(object, defaults)` 

2. exposed `toDefaultObject` method that is called by `toObject` (this way the dev can override the method and do magic and hurt themselves in `clone`)

3. We can do a more aggressive change by making `toObject(propsToInclude)` strict - meaning it will return **ONLY** an object with `propsToInclude` as keys and nothing else. We can do that later on

I just want the annoyance of default values to leave us alone a bit at least on this end

**BREAKING**


src/shadow.class.ts
```diff
@@ -45,13 +45,6 @@
      */
     affectStroke: false,
 
-    /**
-     * Indicates whether toObject should include default values
-     * @type Boolean
-     * @default
-     */
-    includeDefaultValues: true,
```
ShaMan123, line 53: b

src/shapes/group.class.ts
```diff
@@ -831,30 +831,37 @@
      * @returns {fabric.Object[]} serialized objects
      */
     __serializeObjects: function (method, propertiesToInclude) {
-      var _includeDefaultValues = this.includeDefaultValues;
```
ShaMan123, line 834: b

src/shapes/object.class.ts
```diff
@@ -370,128 +374,121 @@
...
      * @default
      */
-    perPixelTargetFind:       false,
-
-    /**
-     * When `false`, default object's values are not included in its serialization
-     * @type Boolean
-     * @default
-     */
-    includeDefaultValues:     true,
```
ShaMan123, line 430: b

src/static_canvas.class.ts
```diff
@@ -76,14 +76,6 @@
      */
     overlayImage: null,
 
-    /**
-     * Indicates whether toObject/toDatalessObject should include default values
-     * if set to false, takes precedence over the object value.
-     * @type Boolean
-     * @default
-     */
-    includeDefaultValues: true,
```
ShaMan123, line 85: b

test/unit/activeselection.js
```diff
@@ -97,39 +97,6 @@
     assert.ok(group.getObjects()[0] !== clone.objects[0], 'should produce different objects in array');
   });
 
-  QUnit.test('toObject without default values', function(assert) {
```
ShaMan123, line 100: b

test/unit/canvas_static.js
```diff
@@ -1117,16 +1115,6 @@
     assert.equal(canvas.toObject().objects[0].type, rect.type);
   });
 
-  QUnit.test('toObject non includeDefaultValues', function(assert) {
```
ShaMan123, line 1120: b

test/unit/group.js
```diff
@@ -218,40 +218,6 @@
     assert.ok(group.getObjects()[0] !== clone.objects[0], 'should produce different objects in array');
   });
 
-  QUnit.test('toObject without default values', function(assert) {
```
ShaMan123, line 221: b

test/unit/object.js
```diff
@@ -26,7 +26,6 @@
     assert.ok(cObj.constructor === fabric.Object);
 
     assert.equal(cObj.type, 'object');
-    assert.equal(cObj.includeDefaultValues, true);
```
ShaMan123, line 29: checkout

test/unit/path.js
```diff
@@ -162,12 +162,11 @@
     });
   });
 
-  QUnit.test('toObject', function(assert) {
```
ShaMan123, line 165: checkout

test/unit/rect.js
```diff
@@ -204,13 +204,6 @@
     assert.equal(svg, '<g transform=\"matrix(1 0 0 1 50.5 50.5)\"  >\n<rect style=\"stroke: none; stroke-width: 1; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(0,0,0); fill-rule: nonzero; opacity: 1;\"  paint-order=\"stroke\"  x=\"-50\" y=\"-50\" rx=\"0\" ry=\"0\" width=\"100\" height=\"100\" />\n</g>\n');
   });
 
-  QUnit.test('toObject without default values', function(assert) {
```
ShaMan123, line 207: checkout

test/unit/shadow.js
```diff
@@ -173,19 +173,6 @@
     assert.deepEqual(object, object2);
   });
 
-  QUnit.test('toObject without default value', function(assert) {
```
ShaMan123, line 176: checkout

test/visual/svg_import.js
```diff
@@ -23,8 +23,6 @@
           // something is disabling objectCaching and i cannot find where it is.
           var group = fabric.util.groupSVGElements(objects, options);
           canvas.setDimensions({ width: group.width + group.left, height: group.height + group.top });
-          group.includeDefaultValues = false;
```
ShaMan123, line 26: checkout

ShaMan123 (commented):
breaking

ShaMan123:
we decided not to break
DIE! until we bring you back to life

PR #8111, opened 2022-08-02
fix(): weird `toJson` bug

ShaMan123:
JSON.stringify passes an empty string as first param
It causes a bug in `toObject` as it passes a string as `propsToInclude`
So I removed the params from the signature
No need for them anyways, and if someone does need them they should use `toObject`

DISABLED `lint` test

**QUICK MERGE**

### DIFF
--- a/src/static_canvas.class.ts
+++ b/src/static_canvas.class.ts
@@ -990,6 +990,26 @@
       return this._toObjectMethod('toObject', propertiesToInclude);
     },
 
+
+    /**
+     * Returns Object representation of canvas
+     * this alias is provided because if you call JSON.stringify on an instance,
+     * the toJSON object will be invoked if it exists.
+     * Having a toJSON method means you can do JSON.stringify(myCanvas)
+     * @return {Object} JSON compatible object
+     * @tutorial {@link http://fabricjs.com/fabric-intro-part-3#serialization}
+     * @see {@link http://jsfiddle.net/fabricjs/pec86/|jsFiddle demo}
+     * @example <caption>JSON without additional properties</caption>
+     * var json = canvas.toJSON();
+     * @example <caption>JSON with additional properties included</caption>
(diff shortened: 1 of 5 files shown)


ShaMan123 (commented):
good to go

asturur:
I wonder why this does not happen in master but just on prs.
anyway here is the spec:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

image

this the part of the spec we are interested in.

Thanks for this investigation

asturur:
i wonder why was never a problem before today. We had zero transpilation before

asturur (approved):

src/static_canvas.class.ts
```diff
@@ -1688,25 +1708,6 @@
...
-   * @see {@link http://jsfiddle.net/fabricjs/pec86/|jsFiddle demo}
-   * @example <caption>JSON without additional properties</caption>
-   * var json = canvas.toJSON();
-   * @example <caption>JSON with additional properties included</caption>
-   * var json = canvas.toJSON(['lockMovementX', 'lockMovementY', 'lockRotation', 'lockScalingX', 'lockScalingY']);
-   * @example <caption>JSON without default values</caption>
-   * canvas.includeDefaultValues = false;
-   * var json = canvas.toJSON();
-   */
-  fabric.StaticCanvas.prototype.toJSON = fabric.StaticCanvas.prototype.toObject;
```
asturur, line 1709: this was probably my fault.

asturur: image

well not this time, but this looks totally like something i would do.

asturur: 11 years this bug and no one was armed

asturur: then comes @ShaMan123 ...

ShaMan123: 🤪

PR #8113, opened 2022-08-02
chore(): refactor remove default values

ShaMan123:
instead of deprecating (#8110 ) simply refactors the logic.


ShaMan123:
maybe naming is wrong

PR #8114, opened 2022-08-03
ci(): adapt build script to rollup

ShaMan123:
change the script a bit to use rollup watch because it is 10 times (4s vs 13-17s) faster than using fs and rebuilding!

### DIFF
--- a/scripts/index.js
+++ b/scripts/index.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 const fs = require('fs-extra');
+const os = require('os');
 const _ = require('lodash');
 const path = require('path');
 const cp = require('child_process');
@@ -10,6 +11,8 @@ const chalk = require('chalk');
 const moment = require('moment');
 const Checkbox = require('inquirer-checkbox-plus-prompt');
 const commander = require('commander');
+// const rollup = require('rollup');
+// const loadConfigFile = require('rollup/loadConfigFile');
 const program = new commander.Command();
 
 const { transform: transformFiles, listFiles } = require('./transform_files');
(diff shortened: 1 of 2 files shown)


scripts/index.js
```diff
@@ -80,7 +83,37 @@ class ICheckbox extends Checkbox {
 }
 inquirer.registerPrompt('test-selection', ICheckbox);
 
+// async function rollupBuild(options = {}, onComplete) {
```
ShaMan123, line 86: we can use the custom watcher in the future if we need

scripts/index.js
```diff
@@ -90,23 +123,41 @@ function build(options = {}) {
...
     if (require(path.resolve(websiteDir, 'package.json')).name !== 'fabricjs.com') {
         console.log(chalk.red('Could not locate fabricjs.com directory'));
     }
-    cp.spawn('npm', ['run', 'start:dev'], {
+    const args = ['run', 'start:dev'];
+
+    //  WSL ubuntu
+    // https://github.com/microsoft/WSL/issues/216 
+    // os.platform() === 'win32' && args.push('--', '--force_polling', '--livereload');
+    if (os.platform() === 'win32') {
```
ShaMan123, line 154: poor suckers
me part of them

scripts/index.js
```diff
@@ -364,18 +432,14 @@ program
 program
     .command('build')
     .description('build dist')
-    .option('-f, --fast')
+    .option('-f, --fast', 'skip minifying')
```
ShaMan123, line 435: we can negate this to minify
but who cares

ShaMan123 (commented):
**QUICK MERGE**

Thanks!

scripts/index.js
asturur: Did you try wls2?  Windows linux subsystem...

asturur: Ok is written one line below

asturur (approved):

scripts/index.js
ShaMan123: crossing between kernels is the slowest thing ever 
or something there that slowed down my amazing computer to a halt
I think I should remove it though besides the perf damage it has caused it is really good

PR #8115, opened 2022-08-04
chore(TS): Color

ShaMan123:
`Color` class

I have created a patch index for util so we can start importing from there 

I don't like that the history isn't kept between `src/color.class.ts` -> `src/color/color.class.ts` so I think we should open a PR that only moves the file and then merge this one and do that in future PRs

**EDITED**
**Merge #8116 BEFORE** this PR, update from master and merge this


src/util/index.ts
```diff
@@ -1,2 +1,162 @@
...
+    isTouchEvent,
+    sendPointToPlane,
+    radiansToDegrees,
+    setStyle,
+    calcRotateMatrix,
+    makeBoundingBoxFromPoints,
+    composeMatrix,
+    rotatePoint
+};
+export default fabric.util;
```
asturur, line 162: let's not use default exports, is a mess when we mix default and named exports.
Let's merge this PR. But in general this kind of extraction i think is actual code more than a bunch of import/exports that goes away at rollup time. I would like to use named exports only unless we have some kind of concerns internally.

asturur (approved):

asturur (approved):

asturur (approved):

src/util/index.ts
ShaMan123: For sure
I want named exports as well. 
Just for hacking the build

ShaMan123: well maybe we should remove this line
nothing should use it

PR #8120, opened 2022-08-04
chore(TS): migrate Point

ShaMan123:
nothing interesting

### DIFF
--- a/src/point.class.ts
+++ b/src/point.class.ts
@@ -1,3 +1,4 @@
+import { fabric } from '../HEADER';
 
 interface IPoint {
   x: number
@@ -68,10 +69,10 @@ export class Point {
 
   /**
    * Subtracts another point from this point and returns a new one
-   * @param {IPoint} that
+   * @param {Point} that
    * @return {Point} new Point object with subtracted values
    */
-  subtract(that: IPoint): Point {
+  subtract(that: Point): Point {
     return new Point(this.x - that.x, this.y - that.y);
   }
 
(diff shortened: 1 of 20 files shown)


src/point.class.ts
```diff
@@ -1,344 +1,348 @@
...
+  x: number
+
+  y: number
+
+  type = 'point'
+
+  static toPoint(from: IPoint) {
+    return new Point(from.x, from.y);
+  }
+
```
asturur, line 22: is this new? what is it about?

asturur:
multiplyEquals and divideEquals aren't available.
Now, i was thinking or we do full set of function that return a new point and mutate te point.

Or we do only return new.

src/point.class.ts
asturur: When @ShaMan123 says `nothing interesting` he is hiding a new motehod.

asturur: Can't we have the class that can initialized both from 2 numbers or one object?
Is there a multiple pattern calls for constructors?

ShaMan123:
I wanted to discuss that as well
I think no mutating is better. less errors in the long run
So dump all XXXEquals methods

src/point.class.ts
ShaMan123: good point!
on it

ShaMan123: > When @ShaMan123 says `nothing interesting` he is hiding a new motehod.

Haha

ShaMan123 (commented):
I removed the use of XXXEquals methods in fabric and marked them as deprecated
It was simple removing these methods which means they are not needed

asturur (approved):

PR #8121, opened 2022-08-04
Chore(TS): migrate Intersection

ShaMan123:
~~first merge #8120 and update from master~~

code remarks:
Should we create classes instead of the static methods?
`new LinePolygonIntersection(...args).status`

## Changes
- **fixes** coincident logic
- **exposed** `line` and `segment` (`line` is infinite) - **BREAKING**, older versions should rename `Line` to `Segment` (what was named `Line` was a finite line which is now called a segment)

## ??
Is `segment` the best naming? Interval seems to be the correct math naming...


src/intersection.class.ts
```diff
@@ -1,68 +1,85 @@
...
+
+/**
+ * **Assuming `T`, `A`, `B` are points on the same line**,
+ * check if `T` is contained in `[A, B]` by comparing the direction of the vectors from `T` to `A` and `B`
+ * @param T 
+ * @param A 
+ * @param B 
+ * @returns true if `T` is contained
+ */
+const isContainedInInterval = (T: Point, A: Point, B: Point) => {
```
ShaMan123, line 17: I am afraid future misuse of the method because it is meant to be used in a very specific context as mentioned in the comment

Any suggestions?

src/intersection.class.ts
```diff
@@ -1,68 +1,85 @@
...
-     */
-    appendPoint: function (point) {
+  /**
+   * Appends a point to intersection
+   * @param {Point} point
+   * @return {Intersection} thisArg
+   * @chainable
+   */
+  appendPoint(point) {
+    if (!this.contains(point)) {
```
ShaMan123, line 50: append only unique points

src/intersection.class.ts
```diff
@@ -1,68 +1,91 @@
...
-   * Checks if one line intersects another
-   * TODO: rename in intersectSegmentSegment
+   * Appends points to intersection
+   * @param {Point[]} points
+   * @return {Intersection} thisArg
+   * @chainable
+   */
+  appendPoints(points) {
+    this.points = this.points.concat(points.filter(point => {
+      return !this.contains(point);
```
ShaMan123, line 64: same unique

src/intersection.class.ts
```diff
@@ -1,68 +1,91 @@
...
+   * @param {boolean} [aIinfinite=true] check intersection by passing `false`
+   * @param {boolean} [bIinfinite=true] check intersection by passing `false`
+   * @return {Intersection}
    */
-  fabric.Intersection.intersectLineLine = function (a1, a2, b1, b2) {
-    var result,
-        uaT = (b2.x - b1.x) * (a1.y - b1.y) - (b2.y - b1.y) * (a1.x - b1.x),
-        ubT = (a2.x - a1.x) * (a1.y - b1.y) - (a2.y - a1.y) * (a1.x - b1.x),
-        uB = (b2.y - b1.y) * (a2.x - a1.x) - (b2.x - b1.x) * (a2.y - a1.y);
+  static intersectLineLine(a1, a2, b1, b2, aIinfinite = true, bIinfinite = true) {
```
ShaMan123, line 80: added option for infinite line check

src/intersection.class.ts
```diff
@@ -1,68 +1,91 @@
...
+    let result;
+    const uaT = (b2.x - b1.x) * (a1.y - b1.y) - (b2.y - b1.y) * (a1.x - b1.x),
+      ubT = (a2.x - a1.x) * (a1.y - b1.y) - (a2.y - a1.y) * (a1.x - b1.x),
+      uB = (b2.y - b1.y) * (a2.x - a1.x) - (b2.x - b1.x) * (a2.y - a1.y);
```
ShaMan123, line 84: who ever wrote this knows their algerba
I can't understand it exactly, just the main concept

src/intersection.class.ts
```diff
@@ -72,95 +95,145 @@ import { Point } from './point.class';
     }
     else {
       if (uaT === 0 || ubT === 0) {
-        result = new Intersection('Coincident');
+        const segmentsCoincide = aIinfinite || bIinfinite
+          || isContainedInInterval(a1, b1, b2) || isContainedInInterval(a2, b1, b2)
+          || isContainedInInterval(b1, a1, a2) || isContainedInInterval(b2, a1, a2);
+        result = new Intersection(segmentsCoincide ? 'Coincident' : undefined);
```
ShaMan123, line 101: If segments check that coincident happens in range
fix!

src/intersection.class.ts
```diff
@@ -72,95 +95,145 @@ import { Point } from './point.class';
...
 
-    for (i = 0; i < length; i++) {
+    for (let i = 0, b1, b2, inter; i < length; i++) {
       b1 = points[i];
       b2 = points[(i + 1) % length];
-      inter = Intersection.intersectLineLine(a1, a2, b1, b2);
-
+      inter = Intersection.intersectLineLine(a1, a2, b1, b2, infinite, false);
+      if (inter.status === 'Coincident') {
+        return inter;
```
ShaMan123, line 156: fixed detection coincident (someone left a note about it)

src/intersection.class.ts
```diff
@@ -72,95 +95,145 @@ import { Point } from './point.class';
...
+        result.appendPoints([a1, a2]);
+      }
+      else {
+        result.appendPoints(inter.points);
+      }
+    }
 
-      result.appendPoints(inter.points);
+    if (coincidents.length > 0 && coincidents.length === points1.length && coincidents.length === points2.length) {
+      return new Intersection('Coincident');
```
ShaMan123, line 206: same fix of TODO comment

ShaMan123 (commented):
OK
This demanded much more work than I thought

Fixed 2 bugs regarding coincident

**READY**

I am very happy about this

src/intersection.class.ts
ShaMan123: some matrix det/cross clever calculation
well done who ever you are

src/intersection.class.ts
```diff
@@ -1,166 +1,226 @@
...
 
   /**
-   * Checks if one line intersects another
-   * TODO: rename in intersectSegmentSegment
+   * Appends points of intersection
+   * @param {...Point[]} points
+   * @return {Intersection} thisArg
+   * @chainable
+   */
+  private append(...points) {
```
ShaMan123, line 49: refactored `appendPoint` `appendPoints` to `append` es6!
I don't consider it breaking as this method is private

src/intersection.class.ts
```diff
@@ -124,10 +124,14 @@ export class Intersection {
 
   /**
    * Checks if line intersects polygon
+   * 
+   * @todo account for stroke
```
ShaMan123, line 128: #8083

ShaMan123 (commented):
reviewed again and fixed some more stuff

**READY** set go

src/intersection.class.ts
asturur: is this hot code? why does it matters? If this is used during render i don't really want to do it.

ShaMan123: hot code? What's that?

ShaMan123: How damaging is this loop?
we can make it a map instead but consider that most case points intersection is <=2

ShaMan123: only 2 polygons that are almost identical will have <= n-1

ShaMan123: > hot code? What's that?

ahh code used by the dev?
yes, getting intersection points in crucial for ui displaying these points

asturur: with hot code i mean code we use in tight loops during rendering. Goint to check how many intersections we use

asturur: Ok is fine, the Intersection is just a fancy container for points, the math to find the point didn't change. I'm good with this.

src/intersection.class.ts
asturur: Some of those can be probably grouped in smaller const, i m not sure the extra const are worth the subtraction saved.

asturur (approved):

src/intersection.class.ts
ShaMan123: I thought of it too.
Wasn't sure how to name what...

PR #8122, opened 2022-08-04
chore(TS): migrate parser

ShaMan123:
dreadful work
forgive me


ShaMan123 (commented):
Nothing changed
split everything to modules
moved consts from HEADER to parser

asturur:
> Nothing changed split everything to modules moved consts from HEADER to parser

I will trust you because i can't read this thing. Tests pass, is good.

> forgive me

**No**, because you were perfectly aware that after you put hours in something i let go because i don't want to waste people time.

ShaMan123:
I understand your view and respect it.
I became impatient and wanted to move forward not being stuck on objects because of the parser. I wanted to fix all imports on objects files.
So I did this. I should have left it to you.

PR #8123, opened 2022-08-04
chore(ts): WIP - Convert utils

asturur:
This will be a partial conversion of utils, i will stop at the misc file and review/merge.
All together is a mess, but luckily we can also swap the imports at that point.


src/util/misc/radiansDegreesConversion.ts
```diff
@@ -0,0 +1,20 @@
...
+export const degreesToRadians = (degrees: TDegree): TRadian => (degrees * PiBy180) as TRadian;
+
+/**
+ * Transforms radians to degrees.
+ * @static
+ * @memberOf fabric.util
+ * @param {TRadian} radians value in radians
+ * @return {TDegree} value in degrees
+ */
+export const radiansToDegrees = (radians: TRadian): TDegree => (radians / PiBy180) as TDegree;
```
ShaMan123, line 20: beautiful

src/util/misc/sin.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param {TRadian} angle the angle
+ * @return {Number} the sin value for angle.
+ */
+export const sin = (angle: TRadian): number => {
+  if (angle === 0) { return 0; }
+  const angleSlice = angle / halfPI;
+  let value = 1;
+  if (angle < 0) {
+    // sin(-a) = -sin(a)
+    value = -1;
```
ShaMan123, line 19: make a const?
```suggestion
    value = Math.sign(angle);
```

asturur: i always forget about Math.sign

asturur:
i know what i broke, i ll try to fix it today.
I messed up the imports/exports

asturur:
@ShaMan123 i had to revert a couple of imports.
The color class importing some util and the animate color util importing the color class was creating a circular dependency.

src/util/internals/removeFromArray.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Removes value from an array.
+ * Presence of value (and its position in an array) is determined via `Array.prototype.indexOf`
+ * @static
+ * @memberOf fabric.util
+ * @param {Array} array
+ * @param {*} value
+ * @return {Array} original array
+ */
+export const removeFromArray = (array: any[], value: any): any[] => {
```
asturur, line 10: i m sure there is a way to tell Typescript that the returned array is the same type of the input array

asturur:
Ok something broke with the import system and i had to patch it. I have no idea why.

src/util/internals/removeFromArray.ts
ShaMan123: Generic types

```suggestion
export function removeFromArray<T>(array: T[], value: T): T[] {
```

src/mixins/observable.mixin.ts
```diff
@@ -1,6 +1,7 @@
 //@ts-nocheck
 (function(global) {
   var fabric = global.fabric;
+  console.log(fabric)
```
ShaMan123, line 4: Hi!

ShaMan123 (commented):
Regarding the circular import
I have seen it as well and managed to fix it in a different way without reverting
Try importing utils directly and not from the `util` index OR make `animate_color` or whatever a module (even as a hack).
That should fix it.
rollup seems to be a decent tool

src/mixins/observable.mixin.ts
asturur: yeah i was investigating why fabric was undefined

src/util/internals/removeFromArray.ts
asturur: and of course T can be a mix of types. ( even if in our case often isn't )

asturur: fat arrows aren't good for TS as functions?

ShaMan123:
Can't be generic i thinl

בתאריך יום ד׳, 10 באוג׳ 2022, 19:28, מאת Andrea Bogazzi ‏<
***@***.***>:

src/util/misc/vectors.ts
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Creates a vetor from points represented as a point
+ * @static
+ * @memberOf fabric.util
+ *
+ * @param {Point} from
+ * @param {Point} to
+ * @returns {Point} vector
+ */
+export const createVector = (from: IPoint, to: IPoint): Point => new Point(to).subtract(from);
```
asturur, line 32: i don't get this. if the type was Point , way we changed it to interface? we expect devs to use points or objects with x/y?

ShaMan123: You are right
I committed it and felt the same
Reverr this 
And safeguard createVector instead

asturur:
note for later self: `const foo = <T,>(x: T) => x;`

PR #8133, opened 2022-08-07
fix(TS): color merge conflict

ShaMan123:
fix c6b6a5faca0d48c855112f38b43a9d028d63e812 that is wrong
and do some more es6/ts


ShaMan123 (commented):
Added a bit a types
Fixed the attribution link
Cleanup a bit

asturur (approved):

PR #8139, opened 2022-08-10
Migrate circle to es6

avra-m3:


src/shapes/circle.class.ts
```diff
@@ -1,201 +1,213 @@
...
+  if (!isValidRadius(parsedAttributes)) {
+    throw new Error('value of `r` attribute is required and can not be negative');
+  }
 
-})(typeof exports !== 'undefined' ? exports : window);
+  const left = (parsedAttributes.left || 0) - parsedAttributes.radius;
+  const top = (parsedAttributes.top || 0) - parsedAttributes.radius;
+
+  callback(new fabric.Circle({
+    ...options, ...parsedAttributes,
```
avra-m3, line 186: Went with the approach of keeping options consistent with Rect in this pr, it now accepts and correctly uses options.

asturur: this may have unintended consequences. options are the values present in the svg tag, and are meant to either pass in width/height or other small things.
I wouldn't add them here.

asturur:
Aapart a small comment seems good to me.

ShaMan123 (changes requested):
I wish to transform 
```diff
- method: function(...args) {
+ method(...args) {
```

asturur:
> I wish to transform
> 
> ```diff
> - method: function(...args) {
> + method(...args) {
> ```

This is a good point, but we didn't do in Rect PR and we asked him to look at that as a reference.
We can fix this later with a search/replace

asturur (approved):

avra-m3:
oh wow - saw these messages and went to go fix the above comments but looks like I was too slow 😅

asturur:
Sorry i just didn't want to drag you along forever with this. If you want to do some other class like ellipse or polyline or what else is still es5 go on

PR #8141, opened 2022-08-10
fix(GuideLines): clear top context condition

### LINKED ISSUE #6954: Selection rectangle not appearing
## Version
4.3.1

## Test Case
https://jsfiddle.net/hamzawain7/45d2sr3y/

Nodejs or browser?
chrome

## Steps to reproduce
Try to create a selection box by click and drag. It doesn't appear because the initAligningGuidelines library keeps on clearing contextTop

## Expected Behavior
Group selection rectangle should be visible

## Actual Behavior
Group selection rectangle isn't be visible when using the aligning_guidelines.js or centering_guidelines.js libraries

ShaMan123:
closes #6954 
it is perhaps not the best fix but is enough for most use cases
The problem was clearing content top all the time so of course group selector was cleared


lib/aligning_guidelines.js
```diff
@@ -52,11 +52,13 @@ function initAligningGuidelines(canvas) {
...
 
   var verticalLines = [],
-      horizontalLines = [];
+      horizontalLines = [],
+      touchInProgress = false;
 
   canvas.on('mouse:down', function () {
     viewportTransform = canvas.viewportTransform;
     zoom = canvas.getZoom();
+    touchInProgress = true;
```
asturur, line 61: I would be carefull calling this `touch` if is on mouse events and touch events.
Maybe we can find a name that is neutral between mouse and touch events.

ShaMan123: DONE

PR #8148, opened 2022-08-15
chore(TS): more conversion of utils

asturur:
Convert a bit more of utils, and do ad hoc replacemente when the build break.

### DIFF
--- a/src/typedefs.ts
+++ b/src/typedefs.ts
@@ -5,10 +5,16 @@ interface NominalTag<T> {
 
 type Nominal<Type, Tag> = NominalTag<Tag> & Type;
 
-// eslint-disable-next-line no-unused-vars
 const enum Degree {}
-// eslint-disable-next-line no-unused-vars
 const enum Radian {}
 
 export type TDegree = Nominal<number, Degree>;
 export type TRadian = Nominal<number, Radian>;
+
+export const enum StrokeLineJoin {
+  miter = 'miter',
+  bevel = 'bevel',
+  round = 'round',
+}
+
+export type TMat2D = [number, number, number, number, number, number];
--- a/src/util/animate.ts
+++ b/src/util/animate.ts
@@ -1,5 +1,5 @@
 //@ts-nocheck
-import { Point } from '../point.class';
+import { extend } from './lang_object';
 
 (function (global) {
   /**
@@ -32,7 +32,7 @@ import { Point } from '../point.class';
    * @type {AnimationContext[]}
    */
   var fabric = global.fabric, RUNNING_ANIMATIONS = [];
-  fabric.util.object.extend(RUNNING_ANIMATIONS, {
+  extend(RUNNING_ANIMATIONS, {
 
     /**
      * cancel all running animations at the next requestAnimFrame
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -33,18 +33,7 @@ type TComposeMatrixArgs = TTranslateMatrixArgs & TRotateMatrixArgs & TScaleMatri
  * @param  {Boolean} [ignoreOffset] Indicates that the offset should not be applied
  * @return {Point} The transformed point
  */
-export const transformPoint = (p: Point | IPoint, t: TMat2D, ignoreOffset: boolean): Point => {
-  if (ignoreOffset) {
-    return new Point(
-      t[0] * p.x + t[2] * p.y,
-      t[1] * p.x + t[3] * p.y
-    );
-  }
-  return new Point(
-    t[0] * p.x + t[2] * p.y + t[4],
-    t[1] * p.x + t[3] * p.y + t[5]
-  );
-};
+export const transformPoint = (p: Point | IPoint, t: TMat2D, ignoreOffset: boolean): Point => p.transform(t, ignoreOffset);
 
 /**
  * Invert transformation t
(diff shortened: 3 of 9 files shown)


src/util/misc/matrix.ts
```diff
@@ -0,0 +1,207 @@
...
+    return new Point(
+      t[0] * p.x + t[2] * p.y,
+      t[1] * p.x + t[3] * p.y
+    );
+  }
+  return new Point(
+    t[0] * p.x + t[2] * p.y + t[4],
+    t[1] * p.x + t[3] * p.y + t[5]
+  );
+};
```
asturur, line 45: Probably a point.tansform is a good idea too.

ShaMan123: yes!

src/util/misc/matrix.ts
```diff
@@ -0,0 +1,209 @@
...
+type TRotateMatrixArgs = {
+  angle?: TDegree;
+}
+
+type TTranslateMatrixArgs = {
+  translateX?: number;
+  translateY?: number;
+}
+
+type TScaleMatrixArgs = {
```
ShaMan123, line 17: type TScaleMatrixArgs = Patial<{ ... }>

https://www.typescriptlang.org/docs/handbook/utility-types.html

src/util/misc/projectStroke.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param {number} options.scaleX
+ * @param {number} options.scaleY
+ * @param {boolean} [openPath] whether the shape is open or not, affects the calculations of the first and last points
+ * @returns {Point[]} array of size 2n/4n of all suspected points
+ */
+
+type projectStrokeOnPointsOptions = {
+  strokeWidth: number;
+  strokeLineJoin: StrokeLineJoin;
+  strokeMiterLimit: number; // https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit
```
ShaMan123, line 27: if the comment is `/** **/` on top of the prop then it is live and not static text

ShaMan123 (commented):
should matrix be a class extending Array?
It makes sense IMO and will allow a drop replacement for matrix in the future
I will carefully merge stroke projection into #7475 when it is done

src/util/misc/projectStroke.ts
ShaMan123: and perhaps it should be a `Pick` type from `TStrokeOptions`

src/util/animate.ts
```diff
@@ -32,7 +32,7 @@ import { Point } from '../point.class';
    * @type {AnimationContext[]}
    */
   var fabric = global.fabric, RUNNING_ANIMATIONS = [];
-  fabric.util.object.extend(RUNNING_ANIMATIONS, {
+  extend(RUNNING_ANIMATIONS, {
```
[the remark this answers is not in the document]
asturur: Doesn't have to be a subclass of array, it can be a class that has an array referenced. It doesn't really matter too much to me. Here i just swapped fabric.util.object because i had issues with circular imports

src/util/misc/matrix.ts
asturur: Why you prefer Partial here?

src/util/misc/projectStroke.ts
asturur: i ll recheck this, wasn't really a deep though.

src/util/animate.ts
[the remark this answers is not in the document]
ShaMan123: even better

src/util/misc/matrix.ts
```diff
@@ -0,0 +1,209 @@
...
+ * Invert transformation t
+ * @static
+ * @memberOf fabric.util
+ * @param {Array} t The transform
+ * @return {Array} The inverted transform
+ */
+export const invertTransform = (t: TMat2D): TMat2D => {
+  const a = 1 / (t[0] * t[3] - t[1] * t[2]),
+        r = [a * t[3], -a * t[1], -a * t[2], a * t[0], 0, 0] as TMat2D,
+        { x, y } = transformPoint({ x: t[4], y: t[5] }, r, true);
```
ShaMan123, line 59: we should decide about the fate of Point and IPoint

src/typedefs.ts
```diff
@@ -5,10 +5,16 @@ interface NominalTag<T> {
...
 export type TDegree = Nominal<number, Degree>;
 export type TRadian = Nominal<number, Radian>;
+
+export const enum StrokeLineJoin {
+  miter = 'miter',
+  bevel = 'bevel',
+  round = 'round',
+}
+
+export type TMat2D = [number, number, number, number, number, number];
```
ShaMan123, line 20: maybe 
```ts
export type TMat2D =  [number, number, number, number] | [number, number, number, number, number, number];
```

asturur: i made this initially but then i had to go back because it makes more complicated to write code checking for 4 or 6 variant of the array.

src/util/misc/matrix.ts
asturur: The reason why i didn't overuse the point years ago is because i thought a plain object would be more performant than a function or a class. But i think this is not true.
I never benchmarked it, and i struggle to understand why fabric starts to cycle through an handful of hundreds objects.

I like that a Point can be a Point or its basic interface because it allows you to reuse the data you have without transforming it.

ShaMan123: I think we need to aim for flexibility but tend to use Point in fabric and not a simple object.
I don't think there a real diff in perf.
All major repos use classes so it has to be a good idea and the browser as well

PR #8152, opened 2022-08-17
docs(Collection): JSDOC fix `item` return type

Mansehej:
[Collection item should return an object, not it's own type](http://fabricjs.com/docs/fabric.Collection.html#.item)


ShaMan123 (approved):

ShaMan123:
TS will make this fix redundant but it is fine
thanks!

PR #8154, opened 2022-08-18
chore(TS): migrate gradient

ShaMan123:
- ~~merge #8155~~ 
- ~~update from master~~
- ~~resolve conflicts `ours` strategy (ping me if you are not sure and request my review to be certain)~~

**READY**


ShaMan123 (commented):
All tests pass on my machine... 

image

CI has gone mad saying everything fails

**READY** on my end

src/gradient/gradient.class.ts
```diff
@@ -1,493 +1,342 @@
-//@ts-nocheck
 
-import { Color } from "../color";
 
+import { fabric } from "../../HEADER";
+import { Color } from "../color";
+import { iMatrix } from "../constants";
+import { parseTransformAttribute } from "../parser/parseTransformAttribute";
+import { matrixToSVG, populateWithProperties } from "../util";
```
asturur, line 7: please import thos from the respective files, wherever are them now, if are not exported yet keep using fabric.util...

src/gradient/gradient.class.ts
```diff
@@ -0,0 +1,342 @@
...
+        ' cx="', needsSwap ? x1 : x2,
+        '" cy="', needsSwap ? y1 : y2,
+        '" r="', needsSwap ? r1 : r2,
+        '" fx="', needsSwap ? x2 : x1,
+        '" fy="', needsSwap ? y2 : y1,
+        '">\n'
+      );
+      if (needsSwap) {
+        // svg goes from internal to external radius. if radius are inverted, swap color stops.
+        colorStops = colorStops.reverse();
```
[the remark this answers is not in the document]
asturur: I'm mostly sure the old svg export code was correct.
Why you changed it is uncleare there are no explanations.
If you don'y like the way it kept from mutating the original, do it differently, but the behavour shouldn't change

src/gradient/typedefs.ts
```diff
@@ -0,0 +1,104 @@
+import { Percent } from "../typedefs";
+
+export type GradientUnits = 'pixels' | 'percentage';
+
+export type GradientType = 'linear' | 'radial';
```
asturur, line 5: shouldn't those be const enums?

src/gradient/typedefs.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export type GradientValue = number | Percent | string;
```
asturur, line 7: is this a value for the gradient coordinate?

src/gradient/gradient.class.ts
```diff
@@ -0,0 +1,342 @@
...
+  // isType<S extends GradientType>(type: S): this is Gradient<S> {
+  //   return (this.type as GradientType) === type;
+  // }
+
+  /**
+   * Adds another colorStop
+   * @param {Record<string, string>} colorStop Object with offset and color
+   * @return {Gradient} thisArg
+   */
+  addColorStop(colorStops: Record<string, string>) {
```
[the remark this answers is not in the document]
asturur: is probably just for the developer.

src/gradient/gradient.class.ts
```diff
@@ -0,0 +1,342 @@
...
+   *  OR
+   *
+   *  <radialGradient id="radialGrad2">
+   *    <stop offset="0" stop-color="rgb(255,255,255)" />
+   *    <stop offset="0.5" stop-color="rgb(0,0,0)" />
+   *    <stop offset="1" stop-color="rgb(255,255,255)" />
+   *  </radialGradient>
+   *
+   */
+  static fromElement(el: SVGGradientElement, instance: FabricObject, svgOptions: SVGOptions): Gradient<GradientType> {
```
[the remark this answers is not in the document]
asturur: this is ok if we fix the generic svg parser too, fromElement is not supposed to be used directly from the developer

src/typedefs.ts
```diff
@@ -1,16 +1,25 @@
...
+// eslint-disable-next-line no-unused-vars
+const enum Radian { }
 
 export type TDegree = Nominal<number, Degree>;
 export type TRadian = Nominal<number, Radian>;
 
+export type TSize = {
+  width: number;
+  height: number;
+}
```
asturur, line 20: I added something similar in the utils,
I wonder if TSize is an interface more than a type, meaning that anything that has a width and height can be used in something that accepts a size

src/gradient/gradient.class.ts
```diff
@@ -1,493 +1,342 @@
...
-  function getColorStop(el, multiplier) {
-    var style = el.getAttribute('style'),
-        offset = el.getAttribute('offset') || 0,
-        color, colorAlpha, opacity, i;
+  /**
+   * Horizontal offset for aligning gradients coming from SVG when outside pathgroups
+   * @type Number
+   * @default 0
+   */
+  offsetX
```
asturur, line 28: isnt' there a way with class definition to declare the default value in the code, with the property?

src/gradient/gradient.class.ts
```diff
@@ -1,493 +1,342 @@
...
+    type = 'linear' as T,
+    gradientUnits = 'pixels',
+    coords,
+    colorStops = [],
+    offsetX = 0,
+    offsetY = 0,
+    gradientTransform,
+    id
+  }: GradientOptions<T>) {
+    const uid = fabric.Object.__uid++;
```
asturur, line 81: in general, this part has been removed
```
      // sets everything, then coords and colorstops get sets again
      Object.keys(options).forEach(function(option) {
        _this[option] = options[option];
      });
```

This is something we do often with shapes.
Is TS or classes going in our way here?

src/gradient/gradient.class.ts
```diff
@@ -0,0 +1,342 @@
...
+    this.id = id ? `${id}_${uid}` : uid;
+    this.type = type;
+    this.gradientUnits = gradientUnits;
+    this.gradientTransform = gradientTransform || null;
+    this.offsetX = offsetX;
+    this.offsetY = offsetY;
+    this.coords = {
```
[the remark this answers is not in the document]
asturur: i would definitely have a const radialCoords and linearCoords, predefined in a constant file, and i would spread the const depending on the type

asturur:
I m falling asleep while reading, i ll re-read tomorrow, do not take my comments as final.
In general i would like to discuss:
- string unions vs enums for use cases like linear/radial or miter/bevel/round
- type names, i started with a T prefix for type, you added I for interface and there is no prefix for const enums ( but your types in this PR miss the prefix )

src/typedefs.ts
ShaMan123: How do you see the difference between interface and type?

src/gradient/gradient.class.ts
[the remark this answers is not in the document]
ShaMan123: done that [here](https://github.com/fabricjs/fabric.js/pull/8154/files#diff-77607bb54fb0f19b2fbd4190bfc701d2c7bece016ad943ef5cc687141f8b448bR82)

src/gradient/gradient.class.ts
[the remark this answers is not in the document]
ShaMan123: > 
> https://github.com/fabricjs/fabric.js/blob/54c152ef0f95d4f1d671b7366a5b366af3ac7653/src/gradient.class.ts#L287
> 

Ahh https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reverse I thought `reverse` returns a new array

src/gradient/gradient.class.ts
ShaMan123: What's wrong?

src/gradient/gradient.class.ts
ShaMan123: It can be done
In shapes it is done in `set` only right?

src/gradient/gradient.class.ts
```diff
@@ -1,493 +1,342 @@
...
+   * Adds another colorStop
+   * @param {Record<string, string>} colorStop Object with offset and color
+   * @return {Gradient} thisArg
+   */
+  addColorStop(colorStops: Record<string, string>) {
+    for (const position in colorStops) {
+      const color = new Color(colorStops[position]);
+      this.colorStops.push({
+        offset: parseFloat(position),
+        color: color.toRgb(),
```
ShaMan123, line 120: shouldn't this be `color.toRgba()`?

src/gradient/typedefs.ts
ShaMan123: done

ShaMan123:
- **string unions or enums** - I don't have an opinion. I have used both in the past. Not sure if one is better than the other
- **T/I prefix** - I am against prefixing. When I use fabric in an app and I need a type/interface for typedefs/annotation I don't know/care if it's an interface or a type. I just want to import/use it with ease (using autofill, not knowing exactly the name)

ShaMan123 (commented):
- renamed `GradientValue`
- offsetX/Y = 0 class default value
- `src/gradient/constants.ts` - default coord values
- 378ab7291ebf565d2fb9774b2c28f43bc1daa4f2 - safegurad `colorStops` from mutation
- removed `ifNaN` - I found out that `NaN || value === value`

src/gradient/gradient.class.ts
asturur: i think is in the basic setOptions of object.
Is not a big deal for gradients because gradients are often just used for fills and not to carry any logic, but sometimes devs add random properties to object without subclassing

src/gradient/gradient.class.ts
asturur: somehow it does create issues when rollup build, mostly because point is referenced inside of some util and so when transposing some utils to TS, the import order change and the build break

On top of that i don't know if someone using the TS file directly will have tree shaking available, and i don't want them to import the whole util folder if they are trying to import only the utils they need

src/gradient/gradient.class.ts
```diff
@@ -1,493 +1,334 @@
...
+        '" r="', needsSwap ? r1 : r2,
+        '" fx="', needsSwap ? x2 : x1,
+        '" fy="', needsSwap ? y2 : y1,
+        '">\n'
+      );
+      if (needsSwap) {
+        // svg goes from internal to external radius. if radius are inverted, swap color stops.
+        colorStops.reverse(); //  mutates array
+        colorStops.forEach(colorStop => {
+          colorStop.offset = 1 - colorStop.offset;
```
asturur, line 207: i would argue this is a mutation too.
As it was in the old code and i wonder how it does not break the code gradient after an svg export.
i will try to write a tests that make this fail, and then fix it.

i think the safest one would be:
`colorStops  = colorStops.map((colorStop) => ({ ...colorStop, offset: 1 -colorStop.offset }))`

src/gradient/gradient.class.ts
```diff
@@ -1,493 +1,334 @@
...
+    gradientTransform,
+    id
+  }: GradientOptions<T>) {
+    const uid = fabric.Object.__uid++;
+    this.id = id ? `${id}_${uid}` : uid;
+    this.type = type;
+    this.gradientUnits = gradientUnits;
+    this.gradientTransform = gradientTransform || null;
+    this.offsetX = offsetX;
+    this.offsetY = offsetY;
```
asturur, line 88: ok so is there any reason why we want to specify those manually while before we weren't?

Coords and colorStops make sense, but the rest i would rather assign or take the default from class declaration.

```js
constructor({
    coords,
    colorStops = [],
    id,
    ...restOfOptions,
  }: GradientOptions<T>) {

....
Object.assign(this, restOfOptions)

```

why going for the explicit way?

asturur:
I have one outstanding comment, all the rest seems good to me.
The more restrictive percentage check for text and gradients does not seem to work well when running coverage

src/gradient/gradient.class.ts
ShaMan123: > As it was in the old code and i wonder how it does not break the code gradient after an svg export.

I agree.

> i would argue this is a mutation too.

Disagree since `colorStops` is a local copy

src/gradient/gradient.class.ts
ShaMan123: I don't have a good reason

ShaMan123: accept that it keeps junk passed down in options out of the class

ShaMan123:
> I have one outstanding comment, all the rest seems good to me. The more restrictive percentage check for text and gradients does not seem to work well when running coverage

Because the test actually fails, see #8168. Something weird is happening with text and gradients/patterns. Sometimes it renders perfectly and sometimes it renders nothing.

src/gradient/gradient.class.ts
asturur: is exactly what we do with objects on the otherside.
We allow devs to pass id, name, and other things they may want to use.
For me it needs to be clarified, not  blocked

src/gradient/gradient.class.ts
asturur: is still mutation because colorStops is a local copy of the array, not of the object contained inside. Those are still references to the original objects.

asturur (approved):

src/gradient/gradient.class.ts
ShaMan123: Ohh shit

src/gradient/gradient.class.ts
ShaMan123: I disagree and TS will disagree as well.
This is the exact use case of a subclass.
Stricter js is part of modern js.
It's how I see it.

src/gradient/gradient.class.ts
ShaMan123: Why didn't you commit?

PR #8156, opened 2022-08-18
fix(): add pointer data to drop event

ShaMan123:
the need for the pointer in a drop event is a must and because `mouse:move` fires just after, something very weird happens with cached pointers and I get a bad pointer... 
Related #8081
I guess I should add for all drag events as well... (in follow up)

### DIFF
--- a/src/mixins/canvas_events.mixin.ts
+++ b/src/mixins/canvas_events.mixin.ts
@@ -381,7 +381,10 @@
      * @param {Event} e
      */
     _onDrop: function (e) {
-      var options = this._simpleEventHandler('drop:before', e, { dragSource: this._dragSource });
+      var options = this._simpleEventHandler('drop:before', e, {
+        dragSource: this._dragSource,
+        pointer: this.getPointer(e)
+      });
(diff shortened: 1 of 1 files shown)


asturur:
No issue with this PR, but what actually happens with the pointer?
I m not sure what difference there is between calculating the pointer here in the code or on your event handler.
Did you manage to understand where the difference come from?

asturur (approved):

ShaMan123:
Not sure but I am guessing what happens is when I call `getPointer` it might be that a `mouse:move` event has fired filling the pointer cache and then when I can the method (perhaps is an `async` handler) it returns something wrong that was cached after the event was fired.
It is a guess... 
Some race condition if not this one

asturur:
weird, it should be all syncronous here, there shouldn't be time for the mousemove event to start running fabricjs code.

How much was the difference? 1px at most or more?

ShaMan123:
a lot more
Ohh I remember
It happened when I dragged something on to canvas and for some reason the mouse down got cached and wasn't cleared when I dropped

PR #8164, opened 2022-08-21
chore(TS): Move more utils to TS

asturur:
Moving more utils


src/util/misc/misc.ts
```diff
@@ -541,35 +425,6 @@ import { clone, extend } from '../lang_object';
...
-      }
-      //divide align in alignX and alignY
-      alignX = align !== 'none' ? align.slice(1, 4) : 'none';
-      alignY = align !== 'none' ? align.slice(5, 8) : 'none';
-      return {
-        meetOrSlice: meetOrSlice,
-        alignX: alignX,
-        alignY: alignY
-      };
-    },
```
asturur, line 571: this has been rewrote to be TS friendly and readable.
I had to re-read the spec because i couldn't read anymore what this code was doing.

src/typedefs.ts
```diff
@@ -17,4 +17,25 @@ export const enum StrokeLineJoin {
...
+export const enum SupportedSVGUnit {
+  mm = 'mm',
+  cm = 'cm',
+  in = 'in',
+  pt = 'pt',
+  pc = 'pc',
+  em = 'em',
+}
+
 export type TMat2D = [number, number, number, number, number, number];
```
ShaMan123, line 41:
```tsx

export type TMat2DBasic = [number,number,number,number];
export type TMat2DAffine = [number,number,number,number,number,number];

export type TMat2D<T> = TMat2DAffine |TMat2DBasic;

export type...
```

[Playground Link](https://www.typescriptlang.org/play?#code/LAKApgHgDg9gTgFwAQIJ5TEgKgWQIYIBMAIgEJ4DOAlgMZIC8SA2gHYCuAtgEZhwA07brwGce-QWIC6AblCRYiFOky4CJAIIAzTVRaZGrUcInGj4syKHmrM0HOjxkaDNnxFiAHiwA+Bq7XEWjp6SAA+qu7k1DSyIPYKTspIAAowusiMAN6gSLlIEABcSCZwsXlIqEUlsQC+diCabCw0CFQwLChweCwUmvAcqekAFLDpBYMsCHwcBAURGtq6YOFuJFG0AJTZcSB1OwhdPX1wA2mTQ5mFAAx8lVc1fEwAjHw3N0+SG7EH3b39EwgLtdbgV7o8Xm8+C8AKw3WGfaRAA)

src/parser/parseSVGDocument.ts
```diff
@@ -1,7 +1,6 @@
 //@ts-nocheck
 
 import { fabric } from '../../HEADER';
-import { toArray } from '../util';
```
ShaMan123, line 4: why not import directly from the file `lang_array` and not from the index?
That will solve the circular dep

ShaMan123 (commented):
I think `src/util/misc/svgParsing.ts` should move to `parser` folder

src/parser/parseSVGDocument.ts
asturur: Is what i was doing i removed for now all imports from the main index in util, and where possible i m linking to direct files.
Is probably better to leave the util/index.ts  just to create the external facing export for who will use fabric.util

asturur:
> I think `src/util/misc/svgParsing.ts` should move to `parser` folder

Yes, and there are also some duplicated functions here and there.
Let me see if i can move it now.

asturur:
rolling back to before moving svgParsing. can't unbreak what i broke

PR #8135, opened 2022-08-08
ci(): codesandbox

ShaMan123:
- [x] **FIRST** merge #8290 and update from master

add [codesandbox ci](https://github.com/apps/codesandbox-ci) to fabric for issues/PR and for **prototyping** (even on your local machine!)
https://codesandbox.io/docs/ci

~~I think I managed to install the github ci app on fabric..~~
if not please do

The issue you (@asturur ) had of codesandbox not opening should be resolved by using this tool because it should build the repo behind scenes with their engine and expose it for sandboxes to use (instead of containing a build file inside the sandbox)

- `.codesandbox`
   - `ci.json` - generic config 
   - `templates` - a folder containing templates to create a sandboxes from
   - `deploy.js` - deploys the template to codesandbox

- `scripts` - exposed `sandbox` cmd with `build` `start` `deploy`

```shell

npm run sandbox deploy /path/to/sandbox

npm run sandbox deploy -- --template node
> created codesandbox https://codesandbox.io/s/fgh476

npm run sandbox build node
> building node sandbox at ~/.fabric/vanilla

npm start
> start a sandbox from a list of available templates

```

## Try It
- [node](https://codesandbox.io/s/fgh476)
- [vanilla](https://codesandbox.io/s/emou3z)
- [Gitpod Ready-to-Code](https://gitpod.io/#https://github.com/fabricjs/fabric.js/pull/8135)

image

PR #8172, opened 2022-08-22
ci(): `kitchensink` sandbox

ShaMan123:
1. Merge #8135 
2. Update from master
3. Resolve conflicts - shouldn't be any. Ping me if any arises

This PR refactors the kitchensink code into a codesandbox template.
https://github.com/fabricjs/fabric.js/compare/sandbox-ci...kitchensink

### TO DO
- ~~patterns have a bug (could be in fabric)~~
- angular version is very old... (remove entirely on next iteration?)
- ~~deploying fails due to max size limit, too many assets~~ deploying woroks but codesandbox has some issue (seems to be on their side, maybe a different version of parcel)

```bash
npm run sandbox build kitchensink ../kitchensink
```


asturur:
not sure we need this

ShaMan123:
Why are you against?
It is another useful sandbox. Good for quick testing.
I have seen you using the kitchensink a lot for testing. Am I wrong?

ShaMan123:
IMO it is proved as useful since I found bugs in `Pattern`

ShaMan123:
I am thinking of a git action that posts a comment on an issue/PR with all sandboxes

ShaMan123 (commented):
I am happy with this and want it in
If you don't want all the assets I am fine with dropping them

ShaMan123 (commented):
updated from `sandbox-ci`

PR #8175, opened 2022-08-24
chore(TS): fabric version 

ShaMan123:
exposes `VERSION` as const from `package.json`
replaces all `fabric.version` with const


ShaMan123 (commented):
READY

asturur (approved):

PR #8179, opened 2022-08-24
fix(IText): regain focus on mouse move

### LINKED ISSUE #3661: Can't delete text after marking all
## Version
1.7.3

## Test Case
https://jsfiddle.net/noeytn5x/

## Steps to reproduce
Mark all the text inside the textfield and click the delete key (or any key really)

## Expected Behavior
The text is deleted

## Actual Behavior
Nothing happens

It however works fine if I mark all by double-clicking

### LINKED ISSUE #8177: Other <input> elements lose focus whenever there is a mouseout on the canvas while in text edit mode
## Version
5.2.1

https://jsfiddle.net/6hmuzak0/1/

Any browser

## Steps To Reproduce

1. Enter text edit mode on the i-text
2. Click into the text input
3. Move your mouse out of the text input, then back into it
4. Type a character

## Expected Behavior
The new character should go into the text input

## Actual Behavior
The new character goes into the i-text

## Notes
Looks like the code to refocus was added because of https://github.com/fabricjs/fabric.js/issues/3661

Essentially, the code added makes up for the lack of a click event if you start your click inside of the canvas but finish your click outside of the canvas. The click event is what fabric relies on to focus back into the hidden text area. The focus is lost while the mouse is down because to the browser that is effectively a mousedown event outside of the hidden text area. This manifests itself in things like not being able to type/delete while you are dragging the mouse to change the selection in a text area.

Some ideas I for possible solutions:

- Immediately refocus the hidden text area when the focus is lost to a non-input element while you are in text edit mode, remove the existing focus code in mouse out. Being able to type/delete while selecting could be a good thing, and I'm guessing fabric would be able to handle it.
- Only focus the hidden text area on mouse out if the mouse is currently down
- Move the logic from the textbox onClick function to something that runs on mouse up, remove the existing focus code in mouse out
- Add a 'focusout' event to the textbox that keeps focus when the focus is transferring to the body, remove the existing focus code in mouse out

ShaMan123:
fixes #8177 caused by #3759 (that was hacky) and fixes #3661 that was the motivation for #3759

This issue was caused because the hiddenTextarea looses focus after mousedown.
So I have added a focus call in mousemove that fixes both issues.

### DIFF
--- a/src/mixins/itext_behavior.mixin.ts
+++ b/src/mixins/itext_behavior.mixin.ts
@@ -398,6 +398,9 @@ import { removeFromArray } from '../util/internals';
         return;
       }
 
+      // regain focus
+      document.activeElement !== this.hiddenTextarea && this.hiddenTextarea.focus();
+
       var newSelectionStart = this.getSelectionStartFromPointer(options.e),
           currentStart = this.selectionStart,
           currentEnd = this.selectionEnd;
--- a/src/mixins/canvas_events.mixin.ts
+++ b/src/mixins/canvas_events.mixin.ts
@@ -158,14 +158,6 @@
         nestedTarget && nestedTarget.fire('mouseout', { e: e });
       }, this);
       this._hoveredTargets = [];
-
-      if (this._iTextInstances) {
-        this._iTextInstances.forEach(function(obj) {
-          if (obj.isEditing) {
-            obj.hiddenTextarea.focus();
-          }
-        });
-      }
     },
 
     /**


asturur:
We need to backport this to 5.x too

asturur (approved):

asturur:
To be clear https://github.com/fabricjs/fabric.js/pull/3759 was as hacky as this newer fix.
It restored focus on mouseout. this restore focus on mousemove.
I fail to understand how this is more solid that the other.
The old code was losing focus on textinput because hovering on textinput equal to a mouseout from canvas and so it was triggering.
With this new fix hovering over the itext will loose focus anyway, or am i wrong?

src/mixins/itext_behavior.mixin.ts
```diff
@@ -398,6 +398,9 @@ import { removeFromArray } from '../util/internals';
         return;
```
ShaMan123, line 398: this is why it is more solid

ShaMan123:
> We need to backport this to 5.x too

why?

PR #8180, opened 2022-08-24
chore(TS) convert more utils

asturur:
Plain conversion of add/remove transforms to object and make bounding box from points

### DIFF
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -23,7 +23,7 @@ type TScaleMatrixArgs = {
   skewY?: TDegree;
 }
 
-type TComposeMatrixArgs = TTranslateMatrixArgs & TRotateMatrixArgs & TScaleMatrixArgs;
+export type TComposeMatrixArgs = TTranslateMatrixArgs & TRotateMatrixArgs & TScaleMatrixArgs;
 /**
  * Apply transform t to point p
  * @static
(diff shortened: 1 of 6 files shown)


src/util/misc/matrix.ts
```diff
@@ -33,7 +33,7 @@ type TComposeMatrixArgs = TTranslateMatrixArgs & TRotateMatrixArgs & TScaleMatri
  * @param  {Boolean} [ignoreOffset] Indicates that the offset should not be applied
  * @return {Point} The transformed point
  */
-export const transformPoint = (p: Point | IPoint, t: TMat2D, ignoreOffset: boolean): Point => new Point(p).transform(t, ignoreOffset);
+export const transformPoint = (p: Point | IPoint, t: TMat2D, ignoreOffset?: boolean): Point => new Point(p).transform(t, ignoreOffset);
```
ShaMan123, line 36: aren't we moving this to point? same as rotatePoint?

src/util/misc/objectTransforms.ts
```diff
@@ -0,0 +1,91 @@
+import { Point } from "../../point.class";
+import { TMat2D } from "../../typedefs";
+import { invertTransform, multiplyTransformMatrices, qrDecompose } from "./matrix";
+import type { TComposeMatrixArgs } from './matrix';
+
+type FabricObject = any;
```
ShaMan123, line 6: Maybe we should put this in typedefs so it will be easy to replace...
I didn't put too much thought into it when I did it

src/util/misc/matrix.ts
asturur: is already on point, this is the old util that uses the point method

src/util/misc/objectTransforms.ts
asturur: me neither. is just to remember we need a generic type that accept any fabricObject interface.

src/util/misc/matrix.ts
ShaMan123: sorry I didn't notice

PR #8183, opened 2022-08-25
chore(TS): `ElementsParser` => `parser/ElementsParser`

ShaMan123:
move under `parser` and update some imports

PR #8184, opened 2022-08-25
dep(): fabric.console

ShaMan123:
remove dead code fabric.console

#8183 removes the last remaining usage


asturur:
this is fine,
if we ever need special logging we can add it.

asturur (approved):

ShaMan123:
bye bye then

PR #8185, opened 2022-08-25
chore(TS): move control files under `controls` folder

ShaMan123:
A first patch moving control files under a designated folder and fixing imports
after this is merged a full migration will be done in a separate PR to keep history


src/mixins/canvas_events.mixin.ts
```diff
@@ -1126,7 +1130,9 @@
     /**
      * @private
      */
-    _fire: fabric.controlsUtils.fireEvent,
+    _fire: function (eventName, options) {
+      return fireEvent(eventName, options);
+    },
```
asturur, line 1135: does have it to be a delegate? can't be continue to be an alias?

asturur (approved):

PR #8188, opened 2022-08-25
fix(ci): report build script failure + revert node test runner to QUnit

ShaMan123:
same as #8187 for build script (propagate exit code to main process for ci)
refactor the test function a bit
revert to running node tests with QUnit instead of testem, dedicated to @asturur


ShaMan123 (commented):
READY!

ShaMan123 (commented):
removed your cmds

let's get rid of `build.js` if we are at it

asturur (dismissed):

asturur (approved):

PR #8192, opened 2022-08-26
Update fabric.js

milonic:
bad hyphen character for padding-top


ShaMan123:
I saw a PR about this.
It will not occur on v6... seems to be a problem with the old build script

ShaMan123:
done!

Issue #8178, opened 2022-08-24
Image resizes slightly after loadFromJSON


AlvesJorge:
Fiddle:
https://jsfiddle.net/qm30r1bp/33/

Image changes size slightly when loaded from JSON. 
I tried checking image scaledheight before and after and it says it is still the same height.
Steps to reproduce:
Notice how the image covers the window top to bottom
Open fiddle
Click on the photo to clear it and load it from json
Notice the red bar under the photo

ShaMan123:
http://fabricjs.com/fabric-gotchas

NUM_FRACTION_DIGITS

```js
fabric.Object.NUM_FRACTION_DIGITS=15

```

https://jsfiddle.net/wj9omh8q/

AlvesJorge:
Worked like a charm.
Thank you so much, didn't think that I would need to look into the caveat section so early on.

ShaMan123:
Good point...
We are working on v6, making fabric modular so ugly global/shared constants like `NUM_FRACTION_DIGITS` will hopefully fade out in favor of better practice.

PR #8194, opened 2022-08-26
chore(TS): config

ShaMan123:
~~@asturur  What do you think about this approach?
If you approve I will replace all occurrences of the configured files~~

Extract config from HEADER to a dedicated class


src/config.ts
```diff
@@ -0,0 +1,107 @@
...
+    minCacheSideLimit = 256
+
+    /**
+     * if webgl is enabled and available, textureSize will determine the size
+     * of the canvas backend
+     * @since 2.0.0
+     * @type Number
+     * @default
+     */
+    textureSize = 2048
```
asturur, line 58: we should bump this to 4096.
there was ie11 on some kind of intel gpu that was freaking out at more than 2248.
We can leave a note that if they want to support older hardware devs should pull this down to 2048

src/config.ts
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Skip performance testing of setupGLContext and force the use of putImageData that seems to be the one that works best on
+     * Chrome + old hardware. if your users are experiencing empty images after filtering you may try to force this to true
+     * this has to be set before instantiating the filtering backend ( before filtering the first image )
+     * @type Boolean
+     * @default false
+     */
+    forceGLPutImageData = false
+
+    NUM_FRACTION_DIGITS = 2
```
asturur, line 96: We should leave a note here, that this represent the number of digits when exporting to json. Is a good idea to bump this to 4 or more for designs app, and that is kept to 2 otherwise testing the library is a mess

asturur:
for me is fine, those things had to go out from header ( probably all header will go away )

src/config.ts
ShaMan123: I think we should remove this strange const and introduce it in the options object passed to `toObject` and `toSVG`
It is really bad IMO, e.g. #8178

ShaMan123: And fix tests accordingly

ShaMan123:
header go away!

src/config.ts
ShaMan123: DONE

src/config.ts
ShaMan123: bumped to 4

HEADER.js
```diff
@@ -71,95 +77,25 @@ fabric.SHARED_ATTRIBUTES = [
...
-
-/**
- * Lowest pixel limit for cache canvases, set at 256PX
- * @since 1.7.14
- * @type Number
- * @default
- */
-fabric.minCacheSideLimit = 256;
+config.configure({
+  devicePixelRatio: fabric.window.devicePixelRatio ||
```
ShaMan123, line 89: once the window context is exported we should move this back to `config.ts`

src/config.ts
```diff
@@ -0,0 +1,171 @@
...
+     * If disabled boundsOfCurveCache is not used. For apps that make heavy usage of pencil drawing probably disabling it is better
+     * @default true
+     */
+    cachesBoundsOfCurve = true
+
+    /**
+     * Map of font files
+     * Map<fontFamily, pathToFile> of font files
+     */
+    fontPaths: Record</** fontFamily */ string, /** pathToFile */ string> = {}
```
ShaMan123, line 120: I don't really understand for this is used for
Can it be made a local var in `canvas#toSVG`?
It's the only place it is used....
I thought exposing `registerFonts` that will add fonts to `fontPaths` and call register fonts on node canvas.
??

ShaMan123 (commented):
I really dislike `NUM_FRATION_DIGITS` and think it should be replaced with an option
But for now I am done

ShaMan123:
Most of these values can be localized I guess...

HEADER.js
```diff
@@ -71,95 +77,25 @@ fabric.SHARED_ATTRIBUTES = [
...
+    fabric.window.webkitDevicePixelRatio ||
+    fabric.window.mozDevicePixelRatio ||
```
asturur, line 91: i think we can verify which browser still need those, and we can removed if is something super old.

HEADER.js
```diff
@@ -71,95 +77,25 @@ fabric.SHARED_ATTRIBUTES = [
...
- * - Edge: 1.75
- * - Firefox: 0.9
- * - Safari: 0.95
- *
- * @since 2.0.0
- * @type Number
- * @default 1
- */
-fabric.browserShadowBlurConstant = 1;
+fabric.charWidthsCache = {};
```
asturur, line 98: i m going to take care of this later for the charWidthsCache. I want to change it to a Map.

src/config.ts
asturur: this can be made local to the svg export function yes. But for this PR we should leave it here i guess.

src/config.ts
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * Map<fontFamily, pathToFile> of font files
+     */
+    addFonts(paths: Record</** fontFamily */ string, /** pathToFile */ string> = {}) {
+        this.fontPaths = {
+            ...this.fontPaths,
+            ...paths
+        };
```
asturur, line 148: Object.assign(this.fontPaths, paths) in general seems nicer to me, but i don't care.

src/typedefs.ts
```diff
@@ -55,4 +55,20 @@ export type TransformEvent<T> = TEvent & T & {
...
+export const enum TWebGLPrecision {
+  low = 'lowp',
+  medium = 'mediump',
+  high = 'highp'
+}
+
+/**
+ * @todo remove once rollup supports transforming enums... ANNOYING!
+ * https://github.com/rollup/plugins/issues/463
+ */
```
asturur, line 69: what is the the output so far? isn't TS that supports enum?

asturur (approved):

HEADER.js
ShaMan123: excellent

but change it in #8198

src/config.ts
ShaMan123: This is really something that should be local IMO

src/typedefs.ts
ShaMan123: rollup isn't doing it right and not building an object out of the enum as TS does

src/config.ts
ShaMan123: Only used in canvas.toSVG

src/config.ts
ShaMan123: only used by canvas.toSVG

PR #8196, opened 2022-08-26
fix(fabric.Gradient): Guard against deep mutation on svg export for color exports

asturur:
This was an edge case, but since the fix is actually trivial and in a non hot code path, is still worth fixing.


asturur:
The bug was added in the migration, but a test was missing anyway.
The reason for the bug was the removal of a deep clone statement.

ShaMan123:
Was there a deep clone statement? I think this bug existed regardless

asturur:
https://github.com/fabricjs/fabric.js/pull/8001/files#diff-2e1e3a2e5dc880cba81870b75734070b74d9282630f4b85b3f923999045fc110L222

wasn't the migration indeed, was before

PR #8197, opened 2022-08-27
chore(TS): extract const `reNonWord` from HEADER

ShaMan123:
cleanup header as part of #8194

### DIFF
--- a/src/mixins/itext_behavior.mixin.ts
+++ b/src/mixins/itext_behavior.mixin.ts
@@ -2,6 +2,8 @@
 import { Point } from '../point.class';
 import { removeFromArray } from '../util/internals';
 
+const reNonWord = /[ \n\.,;!\?\-]/;
+
 (function(global) {
   var fabric = global.fabric;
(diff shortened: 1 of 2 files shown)


src/mixins/itext_behavior.mixin.ts
```diff
@@ -288,9 +290,7 @@ import { removeFromArray } from '../util/internals';
     searchWordBoundary: function(selectionStart, direction) {
       var text = this._text,
           index     = this._reSpace.test(text[selectionStart]) ? selectionStart - 1 : selectionStart,
-          _char     = text[index],
-          // wrong
```
ShaMan123, line 292: what is this???

ShaMan123 (commented):
check out the comment

src/mixins/itext_behavior.mixin.ts
asturur: a bad comment that lost meaning.

asturur (approved):

PR #8198, opened 2022-08-27
chore(TS): extract caches from HEADER

ShaMan123:
first merge #8194 and update from master

extracts caches to a dedicated file

we can migrate the objects to Map if we want... Has `set` `get` `has` `clear`, useful for caches


ShaMan123 (commented):
`arcToSegmentsCache` is unused across fabric... 
I guess it was used before?

asturur (approved):

PR #8199, opened 2022-08-27
chore(TS): extract `initFilterBackend` from HEADER

ShaMan123:
update from master
if conflicts - ping me if needed

move `initFilterBackend` to `src/filters/webgl_backend.class.ts` where it belongs

moved `maxTextureSize`, `webGLPrecision` from `config` because they are not configurable, they are browser consts (as is DPR) => refactored the initFilterBackend logic into `WebGLProbe` which is in charge of getting these 2 values (lazy init)

I will not rename `WebglFilterBackend.tileSize` to `WebglFilterBackend.textureSize` though we should for **naming** in a follow up PR (I am really holding myself back)


asturur:
i don't remember  everything because the webgl part was really done fast.
But the tileSize is not the texture size, those are fine with a different name and value for now.

asturur (approved):

HEADER.js
```diff
@@ -89,14 +89,4 @@ config.configure({
     1
 });
 
-fabric.initFilterBackend = function() {
-  if (config.enableGLFiltering && fabric.isWebglSupported && fabric.isWebglSupported(config.textureSize)) {
```
ShaMan123, line 93: I lost `enableGLFiltering`

PR #8201, opened 2022-08-27
chore(TS): es6 for  util/path.ts and more utils converted

asturur:
What did i change?

In the path file i optimized some small algebra and reused some code.
in particular there were things like `2 * a + 2 * b + 2 * c` and i just made it `2 * ( a + b + c )`

I managed to reuse some code between the path bounding box and the path length calculation
in particular this:
https://github.com/fabricjs/fabric.js/pull/8201/files#diff-9492e624a90f62f4e673ae457b41a1b67ad6fee58c8ee2f8011e88d17e90450dL183-L189
became:
https://github.com/fabricjs/fabric.js/pull/8201/files#diff-9492e624a90f62f4e673ae457b41a1b67ad6fee58c8ee2f8011e88d17e90450dR187-R190

Where iterator already existed.

managed to reuse a couple of times more the QB or CB functions.

in dom_style file i mostly removed everything and extended the test, even if that function will probably go away

in dom_event addEventListener and removeEventListener, without ie11 support are now plain wrapper and are deprecated.

Other functions have been only translated to es6.


src/util/path.ts
```diff
@@ -1,859 +1,817 @@
...
-    else {
-      return 2 * Math.PI - (ta - tb);
-    }
+  return result;
+}
+
+/*
+  * Private
+  */
+const calcVectorAngle = (ux, uy, vx, vy) => {
```
ShaMan123, line 107: should be extract to vectors

ShaMan123 (commented):
Are you planning to split `src/util/path.ts` and make it a folder?
I am for

src/util/path.ts
asturur: yes i m sure this is duplicated i noticed as well

asturur:
once all the folder is done and imported correctly moving files with vscode should be easy, now renaming is a pain. In path there is so much stuff yes it can be divided.

asturur:
now what did i touch that a gradient visual test is failing?

asturur:
@ShaMan123  this is finally green, please have a look if you see anything that needs to fixed.
Types are not here yet, i wasted too much time on es6 things i shouldn't have probably.

src/gradient/gradient.class.ts
```diff
@@ -131,7 +133,7 @@ export class Gradient<S, T extends GradientType = S extends GradientType ? S : '
       gradientUnits: this.gradientUnits,
       gradientTransform: this.gradientTransform ? this.gradientTransform.concat() : this.gradientTransform
     };
-    populateWithProperties(this, object, propertiesToInclude);
+    fabric.util.populateWithProperties(this, object, propertiesToInclude);
```
ShaMan123, line 136: why revert?

src/util/dom_request.ts
```diff
@@ -1,75 +1,68 @@
...
-(function(global) {
-  var fabric = global.fabric;
-  function addParamToUrl(url, param) {
-    return url + (/\?/.test(url) ? '&' : '?') + param;
-  }
+import { noop } from '../constants';
 
-  function emptyFn() { }
+function addParamToUrl(url, param) {
+  return url + (/\?/.test(url) ? '&' : '?') + param;
```
ShaMan123, line 5: This can/should be replace with `URLSearchParams`
Supported by node as well: https://nodejs.org/api/url.html#class-urlsearchparams

ShaMan123: And can then be removed since is used only once

src/util/dom_style.ts
```diff
@@ -1,72 +1,23 @@
...
+  if (typeof styles === 'string') {
+    element.style.cssText += ';' + styles;
+    return element;
   }
-
-  fabric.util.setStyle = setStyle;
-
-})(typeof exports !== 'undefined' ? exports : window);
+  Object.entries(styles).forEach(
+    ([property, value]) => elementStyle.setProperty(property, value)
```
ShaMan123, line 20: #7869 good!

src/util/dom_style.ts
```diff
@@ -1,72 +1,23 @@
...
-    }
+/* this file needs to go away, cross browser style support is not fabricjs domain.
+/**
+ * wrapper for setting element's style
+ * @memberOf fabric.util
+ * @param {HTMLElement} element
+ * @param {Object | string} styles
+ * @return {HTMLElement} Element that was passed as a first argument
+ */
+export function setStyle(element, styles) {
```
ShaMan123, line 10: Why does it return the element? Pointless and no used

src/util/path.ts
```diff
@@ -1,861 +1,810 @@
...
+    cp2X, cp2Y,
+    toX, toY
+  ];
+};
+
+/* Adapted from http://dxr.mozilla.org/mozilla-central/source/content/svg/content/src/nsSVGPathDataParser.cpp
+  * by Andrea Bogazzi code is under MPL. if you don't have a copy of the license you can take it here
+  * http://mozilla.org/MPL/2.0/
+  */
+const arcToSegments = (toX, toY, _rx, _ry, large, sweep, rotateX) => {
```
ShaMan123, line 50: can we rename the arg `rx` and the internal var `_rx` instead?

src/util/path.ts
ShaMan123: what about this?

ShaMan123 (commented):
Looks good.
Left a few comments.
Didn't really review `src/util/path.ts` as diff is dreadful.

src/gradient/gradient.class.ts
ShaMan123: Will be handled by #8202

src/util/path.ts
ShaMan123: let's leave this for #8083

ShaMan123 (approved):

src/gradient/gradient.class.ts
asturur: reverted because i didn't conver it yet so is not importable. That was the reason. The original index.ts is removed for now because it causes me issues while converting

PR #8202, opened 2022-08-28
dep(): `populateWithProperties` => `pick`

ShaMan123:
signature changed

`pick(source, keys) => new object`


ShaMan123 (commented):
READY

asturur:
no leave the util folder aloneeeee i m working on that

ShaMan123:
OK
but please make this function so that it returns a new object and doesn't mutate anything!

asturur:
the pr is good do it when i l done there. I have 2 files left

asturur:
> OK but please make this function so that it returns a new object and doesn't mutate anything!

The old populateWithProperties was mutating objects that were created to be returned.
It wasn't bad per se. I agree that pick may be more useful and reusable than populateWithProperties that was a pick + assign.

asturur:
The typing for pick is a bit too strict, as it is toObject.
We don't need to revert it now, but you have to consider a common thing that happen is that devs may want to add id, name or other values that are not part of the class, and export it.

src/static_canvas.class.ts
```diff
@@ -1027,30 +1028,19 @@ import { removeFromArray } from './util/internals';
...
+      const clipPathData = clipPath && !clipPath.excludeFromExport ?
+        this._toObject(clipPath, methodName, propertiesToInclude) :
+        null;
+      return {
         version: VERSION,
-        objects: this._toObjects(methodName, propertiesToInclude),
+        ...pick(this, propertiesToInclude),
+        objects: this._objects
+          .filter((object) => !object.excludeFromExport)
+          .map((instance) => this._toObject(instance, methodName, propertiesToInclude)),
```
asturur, line 1040: not sure this is an improvement whatsover. is a bit harsh to read

asturur (approved):
I ll open a task to revert the strict typing.

ShaMan123:
I don't understand your comment regarding the typing of pick.
What is strict about it?

PR #8203, opened 2022-08-28
feat(): I/O Registry

ShaMan123:
```ts

import { registry } from 'fabric';

class Mip {
    type = 'mip'
    static async fromObject(d) {
        return new Mip();
    }
    static fromElement(a, b) {
        return new Mip();
    }
}

registry.registerClass(Mip);

```

```ts
const mipInstance = registry.fromJSON({ type: 'mip', ...})
```

### DIFF
--- a/src/Registry.ts
+++ b/src/Registry.ts
@@ -8,7 +8,7 @@ type TJSONData = {
     type: string
 }
 
-export type TClassIO<T = unknown> = Function & {
+export type TClassIO<T = unknown> = FunctionConstructor & {
     fromObject(data: TJSONData): T | Promise<T>;
     fromElement?(svgEl: SVGElement, instance: T, options: unknown): T | Promise<T>;
 }
--- a/src/Registry.ts
+++ b/src/Registry.ts
@@ -1,18 +1,16 @@
 
 export type TRegistry<T = unknown, D = unknown, S extends SVGElement = SVGElement, O = unknown> = {
-    fromJSON(data: D): T | Promise<T>;
-    fromSVG(svgEl: S, options: O): T | Promise<T>;
+    json(data: D): T | Promise<T>;
+    svg?(svgEl: S, instance: T, options: O): T | Promise<T>;
 }
 
-type TClassIdentifier = {
-    prototype: {
-        type: string
-    }
+type TJSONData = {
+    type: string
 }
 
-export type TClassIO<T = unknown, D = unknown, S extends SVGElement = SVGElement, O = unknown> = TClassIdentifier & {
-    fromObject(data: D): T | Promise<T>;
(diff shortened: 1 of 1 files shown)


ShaMan123:
I think we can drop `type` in favor of `klass.name` and do a switch for SVG

ShaMan123:
The other issue is how to customize/subclass the registry

src/Registry.ts
```diff
@@ -0,0 +1,49 @@
...
+        return this.registry.get(data.type)?.fromJSON(data);
+    }
+
+    resolveSVGType(el: SVGElement) {
+        return el.tagName.replace('svg:', '');
+    }
+
+    fromSVG<T extends SVGElement, S>(el: T, options: S) {
+        return this.registry.get(this.resolveSVGType(el))?.fromSVG(el, options);
+    }
```
asturur, line 46: Why the registry can't just answer to the question:
what can i enlive for object.type 'rect'
what can i enlive fro svg element 'rect'
?

Why does it have to call the methods too?

asturur: something like backward:

```js
this.register(klass, {
            type: klass.type,
            svgElement: 'rect'
        });
```

I think registering the methods unless has extra benefits i don't get withut an explanation, is just an extra thing a developer has to parse/learn and follow when is debugging code

ShaMan123: I thought of hybrid things that don't conform to how objects/classes are defined in fabric

ShaMan123: This is good

> something like backward:
> 
> ```js
> this.register(klass, {
>             type: klass.type,
>             svgElement: 'rect'
>         });
> ```
>

ShaMan123: Is there a use case that an svg tag maybe parsed by fabric to 2+ different objects?
Like a group containing elements, mask etc.
That is why

asturur: I don't know seems an unnecessary new thing.
A fixed map is bad becuase it makes you import everything.
A map on the go is fine, while you import you register, when you create your class you can override an import with your.

I can see someone wanting to parse and svg text with textbox or itext more than text for example, or a special path of some sort ( an editable path ).

I understand that the relationship between className and type is very rudimental and should be removed, but i really don't want anything that is not as simple as registrable map.

Given a type you give me the class, given the svg tag you give me the class.
Anything above this should have tangible benefits for current use cases.

ShaMan123: You stated the benefits.
Just to make sure I understand:
"Given a type you give me the function (toJSON in most cases), given the svg tag you give me the function (toSVG)"
falls under your definition or not?

ShaMan123:
closed in favor of #8282

PR #8138, opened 2022-08-09
ci(test): migrate test cmd

ShaMan123:
This PR changes the `test` cmd to use `testem`.
This is done so we can stop using `fabricjs.com` for local browser testing and discontinue any associations between the repos.

I decided to keep going with `testem` for now.
We can always migrate to `karma` if we want to.

Later on we can remove the commands of the website... ~~or should I do that here and now?~~
~~Should we move all `testem` files to `tests`? I think it's better order. Their location is used only by `scripts/index`~~ **DONE**

## Files
- `workflows` - semantic cmd change
- `scripts/index` - refactor to support testem + put all local cache files in `.fabric` folder
- `test/GoldensServer.js` - extract from `scripts/index`, now runs when testem start visual tests
- `testem*` - adapted config/test page, moved under `test` and refactored from `json` to `js`
- `test/lib/*` - adapted methods
- `rollup.test.config.js` - need help here, can't get the tests to build properly

Logging has become silly because the program spawns tests in parallel (node, chrome, firefox) so it logs a mess when `--verbose` option is set. Well actually it is not that bad because ci runs only one context at a time so there will be no parallel logging there. Only locally, so I think it is fine. The `-c,--context` option can be used to run a single context. So I can't care less about this. And the test results are dumped to separate files under `.fabric` so the dev can view it there nice and orderly

## Changes

- tests log only errors (exposed `--verbose` flag)
image

- renamed `dev` cmd to `fabric`

### What about the rest of the interactive tests on fabricjs.com?

Started another effort migrating benchmarks, will PR in the near future

PR #8206, opened 2022-08-28
test() When a test fails in node, dump the result in cli_output

asturur:


test/lib/visualTestLoop.js
```diff
@@ -106,6 +106,31 @@
...
+      var ctx = largeCanvas.getContext('2d');
+      ctx.drawImage(original, 0, 0);
+      ctx.putImageData(difference, original.width, 0);
+      ctx.drawImage(golden, original.width + difference.width, 0);
+      var dataUrl = largeCanvas.toDataURL().split(',')[1];
+      console.log('dumping failed test', testName);
+      const fileName = localPath('../../cli_output', `${testName.replaceAll(' ', '_')}.png`);
+      fs.writeFileSync(fileName.replace('file://', ''), dataUrl, { encoding: 'base64' });
+    }
+    // else if (original) {
```
ShaMan123, line 123: leave this please
I use it, git shows a very convenient diff.
And this way I just commit if I need to

image


image

test/lib/visualTestLoop.js
```diff
@@ -175,8 +200,13 @@
               testName + ' [' + golden + '] has too many different pixels ' + differentPixels + '(' + okDiff + ') representing ' + percDiff + '% (>' + (percentage * 100) + '%)'
             );
             if (!isOK) {
-              var stringa = imageDataToChalk(output);
-              console.log(stringa);
+              // var stringa = imageDataToChalk(output);
+              // console.log(stringa);
+              try {
+                dumpFailedTest(testName, renderedCanvas, canvas, output);
```
ShaMan123, line 206: you can move this into the if block below

ShaMan123 (changes requested):

test/lib/visualTestLoop.js
asturur: that is not the one you use. That is a copy that i commented because i didn't write for browser

asturur: https://github.com/fabricjs/fabric.js/pull/8206/files#diff-e236a8dadb89494460b0e0c2d0018e76d602cb4bf08e1f52d3e673b7709a03ebR71-R79

this is the one you mean

test/lib/visualTestLoop.js
asturur: i prefer it to happen every run, in case i can link the results out in PRs

test/lib/visualTestLoop.js
ShaMan123: > https://github.com/fabricjs/fabric.js/pull/8206/files#diff-e236a8dadb89494460b0e0c2d0018e76d602cb4bf08e1f52d3e673b7709a03ebR71-R79
> 


The link pints me to a strange place. Add a comment in the code so I can see


> that is not the one you use. That is a copy that i commented because i didn't write for browser

I wrote that in #8138

asturur:
```
I have disabled golden debugging/recreation from node because I understood that it is not accurate enough because of node-canvas
```
Yes node-canvas is not accurate enough, but i generate the golden from node anyway because they help finding out issues earlier. When a test is too much different in node, we disable the test for node ( we need to do that anyway or it will fail ).

```
Don't you enjoy using it this way?
```

I think the old was the best, i deleted a golder and i would get it back just running the tests, no flags.

test/lib/visualTestLoop.js
ShaMan123: I looked into that as well
But you said that node visual test results are not accurate.
That is why I disabled the option the debug visuals in node tests to begin with and moved to browser outputs.

ShaMan123: I can be done if we post the results somewhere.
Github doesn't allow creating a comment with images using actions/bots but we can easily add links to posted images.
Maybe we create a test dumps repo under `fabricjs` org and commit to there

ShaMan123: we create a branch, push images and remove the branch once the pr is closed

ShaMan123: BTW you should use the method `generateGolden` by changing the path output instead of adding more logic IMO

test/lib/visualTestLoop.js
ShaMan123: > I think the old was the best, i deleted a golden and i would get it back just running the tests, no flags.

It still works like that.
The `-d` flag will recreate the golden **ONLY** if the test failed
The `-r` flag will recreate the golden regardless

ShaMan123: I don't mind switching the `-d` flag on as default

ShaMan123: For me it is annoying running a test, seeing it fails, navigating to the file, deleting it and all over again each time I change something in the code and want to test the diff

I flag `-d` and it happens.
Then I can discard changes in github desktop.
No hassle.

ShaMan123: Perhaps I didn't explain how to use it properly until now.
It should make your life much easier.

test/lib/visualTestLoop.js
ShaMan123: How about the following:
CI runs visual tests with `-d` flag, meaning that failing tests update golden refs.
Then it creates a branch with these changes and posts a comment with the diff between the branches.
It can prefix the branch name with something and delete it once the pr is closed.
Thoughts?

ShaMan123: #8210

asturur: I didn't add a generate golden method, the method is differnet and produces a collage of the 3 images.
This method replaced the imageDataToChalk function that was not really usefull.
I want to take care of this, i felt the need for it, i started it, i can finish it. Then if is really ugly we can improve it.


> we create a branch, push images and remove the branch once the pr is closed

This seems an overkill of resourcers.

There are actions that archive artifacts of pr runs, and keep them somewhere in github.

ShaMan123: I looked at artifacts...
Am not sure how to display the results though
This on the other hand is simple: https://github.com/fabricjs/fabric.js/compare/ci-visual-test-action...ci-pr%238210-visuals-firefox

ShaMan123: Maybe is an overkill, but it is on github actions...
This is the result:
https://github.com/fabricjs/fabric.js/pull/8210#issuecomment-1229892565

.github/workflows/visual-test-chrome.yml
```diff
@@ -23,3 +23,9 @@ jobs:
       uses: GabrielBB/xvfb-action@v1
       with:
         run: npm run test -- -s visual -p 8080 -c chrome
+    - name: Archive artifacts
+      if: always()
+      uses: actions/upload-artifact@v3
+      with:
+        name: chrome-failing-visual-tests
+        path: cli_output/failed_visual_tests/chrome/*
```
asturur, line 31: this doesn't work yet

.github/workflows/visual-test-firefox.yml
```diff
@@ -23,3 +23,9 @@ jobs:
       uses: GabrielBB/xvfb-action@v1
       with:
         run: npm run test -- -s visual -p 8080 -c firefox
+    - name: Archive artifacts
+      if: always()
+      uses: actions/upload-artifact@v3
+      with:
+        name: firefox-failing-visual-tests
+        path: cli_output/failed_visual_tests/firefox/*
```
asturur, line 31: this doesn't work yet

ShaMan123:
I suggest you merge #8227 first.
It will be simpler working on one file/workflow

ShaMan123:
I am closing in favor of #8402 

If you disagree reopen

PR #8209, opened 2022-08-29
fix(): `window` => `fabric.window`

ShaMan123:
This patch is entirely redundant because the method is called only in the browser and because of #8208 
But for the sake of code styling.


asturur:
yep this is correct

asturur (approved):

PR #8210, opened 2022-08-29
ci(): visual test results action

ShaMan123:
WIP

This is cool but as @asturur suggested artifacts are probably more straight forward and less messy
Check out the first comment made by the bot that points to visual diffs from ci test results (see **Visual Test Result Diffs**)


ShaMan123:
https://github.com/fabricjs/fabric.js/compare/master...ci-pr%238210-visuals-firefox

ShaMan123:
This is great!
I will add an action that deletes the ci branches when the pr closes and update the coverage report action to include the diff urls

PR #8212, opened 2022-08-29
test(): Restore old test commands for logging

asturur:
This PR restores old test commands in order to keep logging in CI.
All the script made by @ShaMan123 is still there, accessible with `test:new` and `test:unit-browser` or `test:visual-browser` to use the script, the browsers and the command line options.


asturur:
@ShaMan123 somehow this rollback now does not work on firefox.
Can you spot what is it?
It speaks of some JSON.parse, but i m not sure which invalid json is looking for

PR #8218, opened 2022-08-30
fix(): abort concurrent rendering

ShaMan123:
This fix introduces the ability to abort a concurrent render cycle by calling `abortRendering`

Probably fixes a lot of issues that were reported on `calling "XXX" on null ctx` that were silenced because of qunit

It committed whitespace without my intention. And I don't want to pick out the trouble. Isn't there a git cmd to revert whitespace? Sound useful


src/static_canvas.class.ts
```diff
@@ -703,7 +687,23 @@ import { removeFromArray } from './util/internals';
      */
     requestRenderAll: function () {
       if (!this.isRendering) {
-        this.isRendering = fabric.util.requestAnimFrame(this.renderAndResetBound);
+        new Promise((resolve, reject) => {
+          const controller = new AbortController();
```
ShaMan123, line 692: polyfill needed for node<16

asturur:
I don't like this change.
A render is not abortable.
A render either is not started yet, or is started. And once is started, it will finish because there is no time to interrupt it.
cancelAnimationFrame should work, and if it doesn't work and this change does, is probably because we are adding extra code in front of it that is giving time to execute things, while we want just to render as fast as possible.

> This fix introduces the ability to abort a concurrent render cycle by calling `abortRendering`
> 
> Probably fixes a lot of issues that were reported on `calling "XXX" on null ctx` that were silenced because of qunit
> 
> It committed whitespace without my intention. And I don't want to pick out the trouble. Isn't there a git cmd to revert whitespace? Sound useful

There is an option to hide the whitespace changes that is good enough.
Prettier or lint are exactly to avoid those things

src/static_canvas.class.ts
```diff
@@ -748,7 +763,6 @@ import { removeFromArray } from './util/internals';
      */
     renderCanvas: function(ctx, objects) {
       var v = this.viewportTransform, path = this.clipPath;
-      this.cancelRequestedRender();
```
ShaMan123, line 751: This is an important fix
Before, if you would called `Canvas#toCanvasElement` it would have cancelled requested renders. Or was it intentional.
I guess now there could be some nasty race condition.

ShaMan123 (commented):
Seems ready to me
Please look at the second comment closely

ShaMan123:
we had a concurrent review it seems

FYI it did abort it. Not only took more time.
But you convinced me.

PR #8220, opened 2022-08-30
fix(): dispose-render concurrency

ShaMan123:
## Description

Fixes a race condition between requested rendering and disposing

## Reproduce

1. `requestRenderAll` is called
2. dispose is called - frees resources (and tries to cancel the animation frame)
3. rendering is invoked from the requestAnimFrame callback
4. rendering tries to access a freed resource (ctx)
5. error

## Changes

- rename `isRendering` => `nextRenderHandle`
- introduce `destroy` acting instead of `dispose`
- refactor `dispose` to call `destroy` after rendering is finished - it has become an async method that return true if disposing has completed, false if the canvas was already destroyed or throws an error if it was aborted by a consequent call
- introduced `destroyed`, `disposed` flag that blocks rendering calls from executing
- fix(): `canvas#toCanvasElement` used to kill requested renders b12491dbbb725b00943629bb3e7d4917188cdb29
- added tests for edge cases
- extracted all canvas disposing tests to a dedicated file (I would have split it up more but we decided to do as less as possible to tests)
- adapt existing tests

replaces #8218


ShaMan123 (commented):
Viewed the test logs
It is fixed!

ShaMan123:
need to add more tests

ShaMan123:
We need to document that now if animations aren't aborted the canvas will never dispose

ShaMan123 (commented):
Updated description
Added extensive tests
Ready!!

ShaMan123:
I earned 0.1% of testing coverage in the PR 🤓

src/static_canvas.class.ts
```diff
@@ -1609,16 +1619,49 @@ import { pick } from './util/misc/pick';
...
+          resolve(true);
+        }
+        task.kill = reject;
+        if (this.__cleanupTask) {
+          this.__cleanupTask.kill('aborted');
+        }
+        
+        if (this.destroyed) {
+          resolve(false);
+        }
```
asturur, line 1640: What are we doing with those?
i see you want to resolve with true/false in case this is an effective dispose or not.
But why? what does it give us this extra information?

asturur:
This seems good to me and less invasive than the original.
I just need some time to read the tests, but definitely the code is good.
Just unclear for me that difference between disposed and destroyed and on top of that i wonder, should a render request started before a dispose call end correctly?
I would say yes, but we  skip it if we are disposing.
Thoughts?

src/static_canvas.class.ts
ShaMan123: Perhaps someone needs to wait until dispose is done to trigger some task that shouldn't  start if not disposed

ShaMan123:
I wanted to abort the rendering once disposed is called. That is why I tried the first approach.
For me it makes less sense waiting for the render to finish. But it doesn't matter too much. 
I am not sure I understood what you meant
> I would say yes, but we skip it if we are disposing.

What happens now with `useEffect`? It will try to dispose the canvas, but that is async and will tear the hook cycle apart, won't it? If it tries to dispose, and then hot reload we will get a multiple canvas init error I think.


`destroy` is the actual logic that kills the canvas (should be `protected`). `dispose` is only in charge of calling `destroy` safely.
Subclass should override `destroy` only.

ShaMan123:
> I would say yes, but we skip it if we are disposing.

You mean maybe we should let the rendering happen even if the the `disposed` flag is set?
We can't. 'Cause if there's an infinite animation the concurrency error will surface.

ShaMan123:
What if we wrap `cancelAnimFrame` so if it is called it flags something for `requestAnimFrame` so in case the timeout is > 16ms it will not invoke the callback?

ShaMan123:
> What happens now with useEffect initializing the canvas? It will try to dispose the canvas, but that is async and will tear the hook cycle apart, won't it? If it tries to dispose, and then hot reload we will get a multiple canvas init error I think.

This makes me prefer the previous approach

src/static_canvas.class.ts
```diff
@@ -703,8 +703,8 @@ import { pick } from './util/misc/pick';
      * @chainable
      */
     requestRenderAll: function () {
-      if (!this.isRendering) {
-        this.isRendering = fabric.util.requestAnimFrame(this.renderAndResetBound);
+      if (!this.nextRenderHandle && !this.disposed && !this.destroyed) {
```
asturur, line 706: maybe i got confused the first time i read it, this is correct, if i m disposing you can't request more renders.

src/static_canvas.class.ts
```diff
@@ -747,12 +747,17 @@ import { pick } from './util/misc/pick';
...
      */
-    renderCanvas: function(ctx, objects) {
+    renderCanvas: function (ctx, objects) {
+
+      if (this.destroyed) {
+        return;
+      }
+
       var v = this.viewportTransform, path = this.clipPath;
-      this.cancelRequestedRender();
```
asturur, line 752: this can't go away, if something renders it needs to kill the queued render. it can be moved in `renderAll` it will still keep the functionality intact but needs to be specified for both static_canvas and canvas.

test/unit/canvas_dispose.js
```diff
@@ -0,0 +1,252 @@
...
+
+    QUnit.test('dispose edge case: rendering after dispose', async function (assert) {
+        const done = assert.async();
+        const canvas = new klass(null, { renderOnAddRemove: false });
+        let called = 0;
+        assert.ok(await canvas.dispose(), 'should dispose');
+        canvas.on('after:render', () => {
+            called++;
+        })
+        canvas.fire('after:render');
```
asturur, line 76: i would add a line here that already tests
```js
assert.equal(called, 1, 'should not have rendered');
```

and i would rename the others as `still equal 1`

test/unit/canvas_dispose.js
```diff
@@ -0,0 +1,252 @@
...
+        assert.equal(canvas.nextRenderHandle, undefined);
+        canvas.requestRenderAll();
+        assert.equal(canvas.nextRenderHandle, undefined, '`requestRenderAll` should have no affect');
+        assert.equal(called, 1, 'should not have rendered');
```
asturur, line 80: this assertion here is not really valid since requestRenderAll is async, the called could have been incremented before or after.

src/static_canvas.class.ts
ShaMan123: > fix(): canvas#toCanvasElement used to kill requested renders https://github.com/fabricjs/fabric.js/commit/b12491dbbb725b00943629bb3e7d4917188cdb29

It fixes a bug

So you mean the case in which there is a requested render and the dev call `renderAll` right?

ShaMan123:
@asturur please comment on the following before we proceed



> > What happens now with useEffect initializing the canvas? It will try to dispose the canvas, but that is async and will tear the hook cycle apart, won't it? If it tries to dispose, and then hot reload we will get a multiple canvas init error I think.
> 
> This makes me prefer the previous approach

src/static_canvas.class.ts
ShaMan123: DONE 0ea2761d22bbc7668f042f44ba33c28c2c16a3b5

test/unit/canvas_dispose.js
ShaMan123: 118c8613

test/unit/canvas_dispose.js
ShaMan123: 118c8613

ShaMan123:
21246b2 safeguards `renderAll` from running after disposed, something that could happen in infinite animation

asturur:
> @asturur please comment on the following before we proceed
> 
> > What happens now with useEffect initializing the canvas?
> 
> Let's say we trigger a hot reload. `useEffect` will try to dispose the canvas but that is async. Will `useEffect` reinitialize the canvas before `dispose` will complete? If so we will get a multiple canvas init error
> 
> I have to say that the prev approach is legit. You read it but I refactored it afterwards.

I m not sure what will happen with a react binding made to dispose a canvas on element unmount for an hot reload.
That depends on the app and react hook setup. Can useEffect wait an effect to resolve?

Also, if you are disposing a canvas, you should not reinitialize the same. Dispose is mostly to trash the instance and clean te resources. If you dispose on unmount, you are probably adding a new instance on mount?

asturur (approved):

ShaMan123:
Correct. 
But adding a new instance on the same canvas element before it is disposed will throw an "already initialized canvas" error.
I am not sure if useEffect can wait or not, nothing in react docs saying so

PR #8227, opened 2022-09-01
ci(): refactor test workflows

ShaMan123:
use `matrix` to consolidate all test workflows


ShaMan123 (commented):
~~I can consolidate all tests into a single file using `if`~~

asturur (approved):

PR #8230, opened 2022-09-01
chore(TS): utils

asturur:
- removed getById in favor of normal browser js
- removed toArray in favor of Array.from
- removed addClass in favor of classList.add
- moved the rotation point logic all in the point class
- removed makeElement
- removed setImageSmoothing


src/static_canvas.class.ts
```diff
@@ -752,7 +752,7 @@ import { pick } from './util/misc/pick';
       this.cancelRequestedRender();
       this.calcViewportBoundaries();
       this.clearContext(ctx);
-      fabric.util.setImageSmoothing(ctx, this.imageSmoothingEnabled);
+      ctx.imageSmoothingEnabled = this.imageSmoothingEnabled;
       ctx.patternQuality = 'best';
```
ShaMan123, line 756: should this [`patternQuality`](https://github.com/Automattic/node-canvas#canvasrenderingcontext2dpatternquality) be set on Image as well? And all cache canvases?
And maybe [`quality`](https://github.com/Automattic/node-canvas#canvasrenderingcontext2dquality) as well?

src/util/animate_color.ts
```diff
@@ -31,49 +24,45 @@ import { Color } from "../color";
    * @param {Function} [options.abort] Additional function with logic. If returns true, onComplete is called.
    * @returns {Function} abort function
    */
-  function animateColor(fromColor, toColor, duration, options) {
-    var startColor = new Color(fromColor).getSource(),
+export function animateColor(fromColor, toColor, duration, options = {}) {
```
ShaMan123, line 27: duration can be passed in options

src/util/animate_color.ts
```diff
@@ -31,49 +24,45 @@ import { Color } from "../color";
...
-          ? options.colorEasing(currentTime, duration)
-          : 1 - Math.cos(currentTime / duration * (Math.PI / 2));
-        return calculateColor(startValue, byValue, posValue);
-      },
-      // has to take in account for color restoring;
-      onComplete: function(current, valuePerc, timePerc) {
-        if (originalOnComplete) {
-          return originalOnComplete(
-            calculateColor(endColor, endColor, 0),
+  return fabric.util.animate(Object.assign(options, {
```
ShaMan123, line 33: spread object?

ShaMan123: better of
I think this object should be a class that handles the animation state, constructor accepting options.

src/util/misc/mergeClipPaths.ts
```diff
@@ -0,0 +1,45 @@
...
+      b.calcTransformMatrix()
+    )
+  );
+  //  assign the `inverted` prop to the wrapping group
+  const inverted = a.inverted && b.inverted;
+  if (inverted) {
+    //  case (1)
+    a.inverted = b.inverted = false;
+  }
+  return new fabric.Group([a], { clipPath: b, inverted: inverted });
```
ShaMan123, line 44:
```suggestion
  return new fabric.Group([a], { clipPath: b, inverted });
```

src/util/misc/mergeClipPaths.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export const mergeClipPaths = (c1: any, c2: any) => {
+  let a = c1, b = c2;
+  if (a.inverted && !b.inverted) {
+    //  case (2)
+    a = c2;
+    b = c1;
+  }
+  //  `b` becomes `a`'s clip path so we transform `b` to `a` coordinate plane
+  applyTransformToObject(
```
ShaMan123, line 31: can use instead:
`sendObjectToPlane(b, b.calcTransformMatrix(), a.calcTransformMatrix()`

src/util/misc/vectors.ts
```diff
@@ -11,14 +9,7 @@ import { TRadian } from '../../typedefs';
  * @param {Number} radians The radians of the angle for the rotation
  * @return {Point} The new rotated point
```
ShaMan123, line 10: `@deprecated`?

src/util/anim_ease.ts
```diff
@@ -1,399 +1,323 @@
...
+ * Backwards easing in
+ * @memberOf fabric.util.ease
+ */
+export const easeInBack = (t, b, c, d, s = 1.70158) => c * (t /= d) * t * ((s + 1) * t - s) + b;
+
+/**
+ * Backwards easing out
+ * @memberOf fabric.util.ease
+ */
+export const easeOutBack = (t, b, c, d, s = 1.70158) => c * ((t = t / d - 1) * t * ((s + 1) * t + s) + 1) + b;
```
ShaMan123, line 246: I see this number a lot. Maybe define it as const at top of file?

src/util/anim_ease.ts
```diff
@@ -1,399 +1,323 @@
...
+  return c / 2 * ((t -= 2) * t * (((s *= (1.525)) + 1) * t + s) + 2) + b;
+}
+
+/**
+ * Bouncing easing out
+ * @memberOf fabric.util.ease
+ */
+export const easeOutBounce = (t, b, c, d) => {
+  if ((t /= d) < (1 / 2.75)) {
+    return c * (7.5625 * t * t) + b;
```
ShaMan123, line 266: more magic numbers that can/should be defined at top of file?

src/util/dom_misc.ts
```diff
@@ -1,301 +1,144 @@
...
-   * @function
-   * @memberOf fabric.util
-   * @param {HTMLElement} element Element to get offset for
-   * @return {Object} Object with "left" and "top" properties
-   */
-  function getElementOffset(element) {
-    var docElem,
-        doc = element && element.ownerDocument,
-        box = { left: 0, top: 0 },
+  return { left, top };
```
ShaMan123, line 56: can this be a point as well?

src/util/dom_misc.ts
```diff
@@ -1,301 +1,144 @@
...
 
-    fabric.util.makeElementUnselectable = makeElementUnselectable;
-    fabric.util.makeElementSelectable = makeElementSelectable;
-  })(typeof exports !== 'undefined' ? exports : window);
+  const scrollLeftTop = getScrollLeftTop(element);
 
-  function getNodeCanvas(element) {
-    var impl = fabric.jsdomImplForWrapper(element);
-    return impl._canvas || impl._image;
+  return {
```
ShaMan123, line 92: Point?

ShaMan123: better off for rtl support as well?

src/util/dom_misc.ts
```diff
@@ -1,301 +1,144 @@
...
+  return element;
+}
+
+/**
+ * Makes element selectable
+ * @memberOf fabric.util
+ * @param {HTMLElement} element Element to make selectable
+ * @return {HTMLElement} Element that was passed in
+ */
+export function makeElementSelectable(element) {
```
ShaMan123, line 118: merge `makeElementSelectable` and `makeElementUnselectable` to one method with a flag?

src/cache.ts
```diff
@@ -44,6 +45,21 @@ export class Cache {
...
+     * Given current aspect ratio, determines the max width and height that can
+     * respect the total allowed area for the cache.
+     * @memberOf fabric.util
+     * @param {number} ar aspect ratio
+     * @return {number[]} Limited dimensions X and Y
+     */
+    limitDimsByArea(ar: number) {
+      const { perfLimitSizeTotal } = config;
+      const roughWidth = Math.sqrt(perfLimitSizeTotal * ar);
+      // we are not returning a point on purpose, to avoid circular dependencies
```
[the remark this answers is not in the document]
asturur: no is not hacky and not everything should be a point.
Before was x and y but wasn't a point.

src/canvas.class.ts
```diff
@@ -1082,9 +1083,9 @@ import { Point } from './point.class';
       if (this.wrapperEl) {
         return;
       }
-      this.wrapperEl = fabric.util.wrapElement(this.lowerCanvasEl, 'div', {
-        'class': this.containerClass
-      });
+      const container = fabric.document.createElement('div');
+      container.classList.add(this.containerClass);
+      this.wrapperEl = fabric.util.wrapElement(this.lowerCanvasEl, container);
```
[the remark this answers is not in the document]
asturur: wrapElement probably don't belong in utils exported object indeed. will be just a function imported here

src/cache.ts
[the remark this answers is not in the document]
ShaMan123: The reason for the circular dep cycle is just because of the assignment `fabric.Point` that will be removed anyways.
If you assign it in index it will not happen.

src/canvas.class.ts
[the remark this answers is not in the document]
ShaMan123: it is used only here across fabric, right?

src/constants.ts
```diff
@@ -1,8 +1,9 @@
 import { TMat2D } from "./typedefs";
 
 export { version as VERSION } from '../package.json';
-export const noop = () => {};
+export function noop() {};
```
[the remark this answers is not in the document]
asturur: There is a difference that a fat arrow can't be bound and can't be used as a class member.
So since noop is used in createClass as an empty method, just to be sure i made it a function

src/point.class.ts
```diff
@@ -378,4 +386,6 @@ export class Point {
   }
 }
 
+const originZero = new Point(0, 0);
```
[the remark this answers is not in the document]
asturur: i think is more readable defined as a const rather than in the function definition.
Does it get defined every function run if is in the function signature?

src/util/anim_ease.ts
asturur: yes there are a bunch of numbers. i wanted to define them, then i realized i didn't know how to call them and moved on for now

src/point.class.ts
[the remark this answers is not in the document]
ShaMan123: good question
no idea

src/util/dom_misc.ts
asturur: not sure, this is style related, it does not need to be transformed or anything else.

src/util/anim_ease.ts
ShaMan123: I totally get that

src/cache.ts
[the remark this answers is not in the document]
ShaMan123: Don't you agree this is out of pattern?

src/util/dom_misc.ts
asturur: not the kind of refactors or changes i m looking to now.
There is still a lot to do, those are really minors utils, i can merge them yes, if i manage to make everything else work.

src/static_canvas.class.ts
asturur: Those are node-canvas specific things i m ignoring on purpose. Are their default not good?

I want to have a clean canvas interface as much as possible in case we move to skia-canvas or maybe we have both as an option.

src/canvas.class.ts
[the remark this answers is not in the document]
asturur: yes i think only here to add a div around the canvas

src/static_canvas.class.ts
ShaMan123: good is default
I changed that in a PR to fix a visual test I recall

ShaMan123: but missed `quality`

asturur:
This is the list of utils from the utils folder, with some ideas of what to continue to expose and what to expose somewhere else as deprecated ( and to become internals )

```js
fabric.util = {
  cos,
  sin,
  rotateVector,
  createVector,
  calcAngleBetweenVectors,
  getHatVector,
  getBisector,
  degreesToRadians,
  radiansToDegrees,
  rotatePoint,
  getRandomInt, // hide
  removeFromArray, // hide
  projectStrokeOnPoints,
  transformPoint,
  invertTransform,
  composeMatrix,
  qrDecompose,
  calcDimensionsMatrix,
  calcRotateMatrix,
  multiplyTransformMatrices,
  stylesFromArray,
  stylesToArray,
  hasStyleChanged, // hide
  object: {
    clone, // hide
    extend, // hide
  },
  createCanvasElement,
  createImage,
  copyCanvasElement,
  toDataURL,
  toFixed,
  matrixToSVG, // hide
  parsePreserveAspectRatioAttribute, // hide
  groupSVGElements, // hide
  parseUnit, 
  getSvgAttributes,
  findScaleToFit,
  findScaleToCover,
  capValue, // hide
  saveObjectTransform,
  resetObjectTransform,
  addTransformToObject,
  applyTransformToObject,
  removeTransformFromObject,
  makeBoundingBoxFromPoints,
  sendPointToPlane,
  transformPointRelativeToCanvas, // this has to be renamed imho
  sendObjectToPlane,
  string: {
    camelize, // hide
    capitalize, // hide
    escapeXml, // hide
    graphemeSplit, // this needs a setter and getter to be overridden
  },
  getKlass, // hide
  loadImage,
  enlivenObjects,
  enlivenObjectEnlivables,
  array: {
    min, // hide
    max, // hide
  },
  pick, // hide
  joinPath,
  parsePath,
  makePathSimpler,
  getSmoothPathFromPoints,
  getPathSegmentsInfo,
  getBoundsOfCurve,
  getPointOnPath,
  transformPath,
  getRegularPolygonPath,
  request, // hide
  setStyle,
  isTouchEvent,
  getPointer,
  removeListener, // hide
  addListener, // hide
  wrapElement, // hide
  getScrollLeftTop, // hide
  getElementOffset, // hide
  getNodeCanvas,
  cleanUpJsdomNode,
  makeElementUnselectable, // hide
  makeElementSelectable, // hide
  isTransparent,
  sizeAfterTransform,
  mergeClipPaths,
  ease,
  animateColor,
  animate,
  requestAnimFrame,
  cancelAnimFrame,
};
```

ShaMan123:
`transformPointRelativeToCanvas` has to be renamed for sure and needs work I think, doesn't account for canvas offset.
Should redo it based on the logic in canvas, extract it from there...

`matrixToSVG` and `parsePreserveAspectRatioAttribute` are exposed in `parser` right? So a dev can use them if needed to create custom parsing, right?

Looks good to me

src/util/dom_misc.ts
asturur: i think those are just for calculating the canvas position for some sort of padding/border issue.
I m not sure we ever need a different measurement.

I also think modern browsers solves this for us giving us an extra pair of mouse event coordinates relative to the canvas element that would be exactly what we need

asturur: yes i think that i should look into moving mouse events with offsetX and offsetY and trash all of this.

asturur: i ll open a task

asturur:
> transformPointRelativeToCanvas has to be renamed for sure and needs work I think, doesn't account for canvas offset.
Should redo it based on the logic in canvas, extract it from there...

Ok so i m not sure i understaood that function. to me it just seems it converts from viewportCoordinates to htmlElement coordinates. What does it do on top of that?

src/cache.ts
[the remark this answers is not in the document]
asturur: is a function about caching, i don't think returning an array to return 2 values is better or worse than returnin an object and calling it a point. I just really need the 2 values out of there.

src/util/animate_color.ts
asturur: why? Animations has a lot of options yes, but that does not mean the thosee needs to be a class. if we es6fy it better it i will probably be less complex

src/util/misc/mergeClipPaths.ts
asturur: i tried already, it seems the same but it isn't.

asturur:
The takeaway from this PRs are:
- removed old dom related functions ( around 100 lines of code )
- figured out that animate color is broken and needs to be verified #8241
- consolidate 4 calls to getComputedStyle to 1 ( probably no differences in performance since those are cached )
- Create a task to look into new pointer coordinates that can remove more code #8237
- all the rest has been moved in new files, es6-fied or typed

There are more improvements that can be made to this code probably, but the intention of this PR was to move it to new import/exports and remove the old es5 code hacks.

If there are no objects i would like to merge this tomorrow and move to the class/function discussion

ShaMan123:
I can review later
Or merge and we move forward

ShaMan123 (commented):
I still disapprove of `limitDimsByArea` return type.
Think of a dev subclassing caching.
But let's move on.

General To Do
- cache can become standalone exports instead of a class once we make everything a map
- add `onStart` to animate - since it complements `delay`
- `sendPointRelativeToCanvas` - rename to `sendPointToWindow` and `sendPointToCanvas`? should utilize `getPointer` code from canvas.
-

src/util/misc/mergeClipPaths.ts
ShaMan123: should be
`sendObjectToPlane(b, b.group?.calcTransformMatrix(), a.calcTransformMatrix()`

ShaMan123: I/We should add a test for `mergeClipPaths`

ShaMan123:
> > transformPointRelativeToCanvas has to be renamed for sure and needs work I think, doesn't account for canvas offset.
> > Should redo it based on the logic in canvas, extract it from there...
> 
> Ok so i m not sure i understaood that function. to me it just seems it converts from viewportCoordinates to htmlElement coordinates. What does it do on top of that?

I should do that but it doesn't account for canvas offset but it should

src/util/misc/mergeClipPaths.ts
ShaMan123: 5ce2a09

src/util/animate.ts
```diff
@@ -1,267 +1,170 @@
...
+  fabric.window.cancelAnimationFrame || fabric.window.clearTimeout;
+
+/**
+ * requestAnimationFrame polyfill based on http://paulirish.com/2011/requestanimationframe-for-smart-animating/
+ * In order to get a precise start time, `requestAnimFrame` should be called as an entry into the method
+ * @memberOf fabric.util
+ * @param {Function} callback Callback to invoke
+ * @param {DOMElement} element optional Element to associate with animation
+ */
+export function requestAnimFrame(...args) {
```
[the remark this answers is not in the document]
asturur: maybe, but since the function is just a passtrough for some reason, i keep it generic

src/util/animate.ts
```diff
@@ -1,267 +1,170 @@
...
+        return;
+      }
+      else {
+        onChange(current, valuePerc, timePerc);
+        requestAnimFrame(tick);
+      }
+    })(start);
+  };
+
+  setTimeout(() => requestAnimFrame(runner), delay);
```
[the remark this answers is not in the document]
asturur: is 0 by default is not undefined !
i thought the code semplification was worth a setTimeout 0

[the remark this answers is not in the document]
ShaMan123: I didn't notice that.
I am not sure is worth...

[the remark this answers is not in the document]
asturur: this change can be reverted to:

```js
if (delay > 0 ) {
  setTimeout(() => requestAnimFrame(runner), delay);
} else {
  requestAnimFrame(runner)
}
```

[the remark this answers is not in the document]
ShaMan123: `undefined` should be not `setTimeout`
`0` we need to decide if behaves as `undefined` or as `setTimeout` 0 does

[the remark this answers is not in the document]
asturur: i removed delay undefined. delay is a number, it will be 0 or a number bigger than 0.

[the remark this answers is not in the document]
ShaMan123: > this change can be reverted to:
> 
> ```js
> if (delay > 0 ) {
>   setTimeout(() => requestAnimFrame(runner), delay);
> } else {
>   requestAnimFrame(runner)
> }
> ```

Great, even just `if(delay)`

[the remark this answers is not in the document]
ShaMan123: I get why >0.

asturur:
@ShaMan123 i reverted your commit.

this is sendObjectToPlane:

```js
  const t = multiplyTransformMatrices(invertTransform(to), from);
  applyTransformToObject(
    object,
    multiplyTransformMatrices(t, object.calcOwnMatrix())
  );
```
it has obect.calcOwnMatrix and 2 matrices to and from.

ShaMan123:
That is why the following is correct.

```diff
- sendObjectToPlane(b, b.calcTransformMatrix, a.calcTransformMatrix())
+ sendObjectToPlane(b, b.group?.calcTransformMatrix, a.calcTransformMatrix())

```

Taking b`s containing plane matrix and multiplying with b`s own matrix gives us `b.calcTransformMatrix()`

asturur:
> * `sendPointRelativeToCanvas` - rename to `sendPointToWindow` and `sendPointToCanvas`? should utilize `getPointer` code from canvas.

I m not sure i get it. There are no window coordinates ever, we either use design coordinate ( your fabricjs canvas creation, or the htmlCanvas element coordinates ( the distance between the point and the left top HTMLCanvas corner ))

asturur:
> That is why the following is correct.
> 
> ```diff
> - sendObjectToPlane(b, b.calcTransformMatrix, a.calcTransformMatrix())
> + sendObjectToPlane(b, b.group?.calcTransformMatrix, a.calcTransformMatrix())
> ```
> 
> Taking b's containing plane matrix and multiplying with b's own matrix gives us `b.calcTransformMatrix()`

I don't think so, i think you are inverting a multiplcation order.
If there is no group, you send null ( let's say is conditionally an IMatrix ).
In the case the group exist you are going to multiply forst for the group matrix and then for the object one,
in the case as it is today, you are multiplying for the ( OBJECT x CONTAINER ).

I ll write a failing unit test, and then you'll owe me a cocacola first time we ever meet.

ShaMan123:
We are on! When will we meet are where?

ShaMan123:
Let's agree that if we meet in Israel it will something much tastier. Accepted?

ShaMan123:
> I m not sure i get it. There are no window coordinates ever, we either use design coordinate ( your fabricjs canvas creation, or the htmlCanvas element coordinates ( the distance between the point and the left top HTMLCanvas corner ))

I wrote `sendPointRelativeToCanvas` so a dev could interact with HTML objects in the dom from a nested object for example. We can trash it completely and I'll redo it later.

ShaMan123:
If I win what happens?

ShaMan123:
I don't like cocacola

asturur:
I added the test in master, as soon as this is merged you can verify if the swap works, and in case do it.

asturur:
> I don't like cocacola

Any food in budget for the loser is fine.

asturur:
> I wrote `sendPointRelativeToCanvas` so a dev could interact with HTML objects in the dom from a nested object for example. We can trash it completely and I'll redo it later.

Ok i didn't know or i didn't remember this.
Offering an anchor point for html elements is doable as long there is a strategy before ( like where do we expect the dev to append the nodes? because if they are inside the canvas container there is need of any transform )

src/util/animate.ts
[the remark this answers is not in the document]
asturur: this has been changed back to the more verbose version

ShaMan123:
> Offering an anchor point for html elements is doable as long there is a strategy before ( like where do we expect the dev to append the nodes? because if they are inside the canvas container there is need of any transform )

Good point
I'll reiterate

PR #8242, opened 2022-09-02
ci(): install prettier

asturur:
This PR adds Prettier and the commands to use it, but do not actually format any code.
Actual code formatting can be discussed, i have no strong optinions apart the 2 i wrote in the configuration file.


asturur:
To me the stupid vscode doesn't work.
Only from npm command

asturur:
image

ok it was this setting here

ShaMan123:
I think you should commit the `.vscode` folder and then the config will be set locally no matter what

asturur:
will follow PR that formats all src folder.

asturur:
> I think you should commit the `.vscode` folder (if existing) and then the config will be set locally no matter what

that will force everyone to have the same options? with overlaps?
We could have an example file for vscode anyway.

We can enforce linting with a push or commit hoot

ShaMan123:
I think it will

ShaMan123:
https://code.visualstudio.com/docs/getstarted/settings#_workspace-settings

ShaMan123:
#8281

PR #8246, opened 2022-09-03
discussion: class vs functions for inheritance

asturur:
@ShaMan123 

I have been playing around a bit with raw classes and functions for the sake of:
- trying to make possible to work with classes as requested by @ShaMan123 
- trying to avoid loosing functionality and flexibility compared to current setup

Of all the things there are 2 functionality i don't want to loose:
- the ability to know which is it the default value for a property of a class
- the ability to modify some parts of the prototype at runtime ( like originX and originY to make an example )
- keep the code organized in mixins

The following code seem to satisfy all requirements, while yes, not being 100% pretty, but i really don't care.

A **mixin** would be defined as a simple object containing method or properties
```js
const sumMixin = {
  sum() {
    console.log('result is 2');
  }
}
```

A class would be defined as normal, but without static properties
```js
class testObject {
  constructor(opts) {
    // strict vs shallow assignment is out of scope of this discussion
    Object.assign(this, opts);
  }
  log() {
    console.log(this.prop);
  }
}
```

After the class declaration mixins and defautl properties can be added:

```js
Object.assign(testObject.prototype, {
  type: 'objectClass',
  color: 'blue',
  stroke: 'red',
  prop: 'logging base clas',
});
Object.assign(testObject.prototype, sumMixin);
```

Subclasses follow the same principle, declared as normal classes
```js
class testRect extends testObject {
  constructor(opts = {}) {
    super(opts);
  }
  test() {
    console.log('break?')
  }
  log() {
    super.log();
  }
}

Object.assign(testRect.prototype, {
  type: 'fRect',
  color: 'yellow',
  prop: 'a rect thing to log',
  height: 100,
});

class testCube extends testRect {
  log() {
    console.log('pre');
    super.test();
    super.log();
    console.log('post');
  }
}

Object.assign(testCube.prototype, {
  type: 'fCube',
  color: 'yellow',
  prop: 'a cube thing to log',
  depth: 300,
});
```

Now copy pasting all this code ( you can take it from the files changed section ) in the chrome console, behave as follow:

- prototype inheritance works as expected
- default values are correctly initialized but aren't copied in the instance, they surface because of inheritance
- it is possible to compare a property of the class testCube with the property of its prototype using the `__proto__` property
- super calls works ok


asturur:
I ll start commenting tomorrow in detail what seems wrong to me, now is late. Please do not commit over this branch.

ShaMan123 (commented):
can you write what isn't good?
Is it this?
```ts
class {
 prop = 'initialValue'
}
```

I still don't see why we should care so much about the default props.

asturur:
Because it would be a breaking change i don't want to break.
It require to rethink some parts to get exactly the same functionality with a different syntax.
It looks like that the `properties` ( that i understand now are not properties ) added in the class declaration are called `public fields` and are just a different thing from values added to the prototype.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields
Those are meant to be replicated in each instance.

For things like array literals, or concatednated arrays  added in the class declaration ( cacheProperties for example ) those get replicated per instance instead of being only referenced.
And while i can skip the worry on memory consumption, adding a property to one of those arrays becomes just harder and confusing.

for a class declared like this:

```js
class myClass {
  defaultSize = [ 10, 10 ]
  functionality() {
    console.log('hello world');
  }
}
```

the behaviour is as follow

```js
var myInstance = new myClass();
var myInstance2 = new myClass();

myInstance.__proto__.functionality = function() { ...modifiedCode } // will change for all classes and instances
myInstance2.__proto__.otherFunctionality = function() { ...moreCode } // will add to all classes and instances
myInstance.__proto__.newProp = { x, y } // will add to all classes and instances

// WHILE

myIstance.__proto__.defaultSize // is undefined
myIstance.__proto__.defaultSize = [newData] // will do nothing
```

Is the difference behavior for public fields and prototypes additions from mixins that i don't want to have headaches with.
If we need to have properties that aren't overridable we can do exactly as it was done before the introduction of public fields, and that is adding them in the constructor, that is also what transpilers default to when the browsers do not have the latest classes support:

```js
class myClass {
  constructor() {
    this.defaultSize = [ 10, 10 ];
  }
  functionality() {
    console.log('hello world');
  }
}
```

ShaMan123:
I am not following.
What is the difference between?
1. `myInstance.__proto__.newProp = { x, y } // will add to all classes and instances`
2. `myIstance.__proto__.defaultSize = [newData] // will do nothing`

I understand that public properties are assigned in the constructor anyways... so why do we need to do that ourselves?

ShaMan123:
```js
class myClass {
  constructor() {
    this.defaultSize = [ 10, 10 ];
  }
  functionality() {
    console.log('hello world');
  }
}
```

```

class myClass {
  defaultSize = [ 10, 10 ];
  constructor() {
    console.log(this.defaultSize) // should output [10,10]
  }
  functionality() {
    console.log('hello world');
  }
}
```

asturur:
The difference is that for originX and originY for example, or rx, ry, or other small things that you may want to set up for your project you don't want to be bound to subclassing.
If we really need something to be defined at class level, for each instance, we can do it in the constructor.
Doing it in the declaration will probably just create confusion on 2 different way to add properties vs fields.

We can make a rule, that in calss bodies you can add just functions and that is easy to spot during PRs

> I am not following. What is the difference between?
> 
> 1. `myInstance.__proto__.newProp = { x, y } // will add to all classes and instances`
> 2. `myIstance.__proto__.defaultSize = [newData] // will do nothing`
> 
> I understand that public properties are assigned in the constructor anyways... so why do we need to do that ourselves?

Oh the difference was that `defaultSize` was defined in the class declaration, so was a public field and not just a value on the prototype.

ShaMan123:
So what's the direction? no public fields?

At the risk of repeating myself I am for aligning with classes completely.
Not static/proto properties at all.
Could you give a use case accept remove defaults that needs them?
Sunclassing is the way it should be done by the dev IMO and by fabric

ShaMan123:
ok I see you continued this discussion in the main thread

ShaMan123:
Could you elaborate more? here or there?
Why is it important for the proto fields and why shouldn't subclassing become citizen A of mutations?

asturur:
> So what's the direction? no public fields?
> 
> At the risk of repeating myself I am for aligning with classes completely. Not static/proto properties at all. Could you give a use case accept remove defaults that needs them? Sunclassing is the way it should be done by the dev IMO and by fabric

The use cases i see are
- default values
- changes to default behaviour

But you can advocate for subclassing and do it too.
You can write examples with subclassing only, you can use public static fields in project and example subclasses.
It does not mean that internally the library can't make use of simple properties instead of fields.

That does not mean you have to force people into subclassing for the minimal deviation to fabricJS default code.
If TS wouldn't have allowed for plugging in JS, i wouldn't have accepted it either.
Embracing classes does not mean either:
- stop using the prototype ( class have one )
- reworking the code to work on one specific set of class features
- introducing more breaking changes than actually necessary


So to question `no public fields`, apart being able to say, i adopted the public fields, what else public fields are going to give us in practice?

PR #8247, opened 2022-09-04
fix(): use `sendObjectToPlane` in `mergeClipPaths`

ShaMan123:
**I win 😋**

---

> That is why the following is correct.
> 
> ```diff
> - sendObjectToPlane(b, b.calcTransformMatrix(), a.calcTransformMatrix())
> + sendObjectToPlane(b, b.group?.calcTransformMatrix(), a.calcTransformMatrix())
> ```
> 
> Taking b's containing plane matrix and multiplying with b's own matrix gives us `b.calcTransformMatrix()`

I don't think so, i think you are inverting a multiplcation order.
If there is no group, you send null ( let's say is conditionally an IMatrix ).
In the case the group exist you are going to multiply forst for the group matrix and then for the object one,
in the case as it is today, you are multiplying for the ( OBJECT x CONTAINER ).

I ll write a failing unit test, and then you'll owe me a cocacola first time we ever meet.

_Originally posted by @asturur in https://github.com/fabricjs/fabric.js/issues/8230#issuecomment-1236135459_


src/util/misc/planeChange.ts
```diff
@@ -104,10 +113,8 @@ export const transformPointRelativeToCanvas = (
  * @param {Matrix} [to] destination plane matrix to contain object. Passing `null` means `object` should be sent to the canvas coordinate plane.
  * @returns {Matrix} the transform matrix that was applied to `object`
  */
-export const sendObjectToPlane = (object: any, from: TMat2D = iMatrix, to: TMat2D = iMatrix): TMat2D => {
-  //  we are actually looking for the transformation from the destination plane to the source plane (which is a linear mapping)
-  //  the object will exist on the destination plane and we want it to seem unchanged by it so we reverse the destination matrix (to) and then apply the source matrix (from)
-  const t = multiplyTransformMatrices(invertTransform(to), from);
+export const sendObjectToPlane = (object: TObject, from: TMat2D = iMatrix, to: TMat2D = iMatrix): TMat2D => {
```
ShaMan123, line 116: I want to add a signature that accepts objects once we migrate `Object` so I can do `instanceof`

ShaMan123 (commented):
Refactored `src/util/misc/planeChange.ts` a bit, to make more readable

ShaMan123 (commented):
Naming: `calcChangePlaneMatrix` or `calcPlaneChangeMatrix`?

ShaMan123:
fixes passing `null` as matrix
https://github.com/fabricjs/fabric.js/blob/f9761ddc1a5504c550d6c60aa553680c5115ad3d/src/mixins/canvas_events.mixin.ts#L1097

src/util/misc/planeChange.ts
```diff
@@ -3,12 +3,23 @@ import { invertTransform, multiplyTransformMatrices, transformPoint } from "./ma
...
 
+/**
+ * We are actually looking for the transformation from the destination plane to the source plane (change of basis matrix)\
+ * The object will exist on the destination plane and we want it to seem unchanged by it so we invert the destination matrix (`to`) and then apply the source matrix (`from`)
+ * @param [from]
+ * @param [to] 
+ * @returns 
+ */
+export const calcPlaneChangeMatrix = (from?: TMat2D | null, to?: TMat2D | null) =>
+  multiplyTransformMatrices(invertTransform(to || iMatrix), from || iMatrix);
```
asturur, line 21: please let's keep iMatrix as the defaul value for arguments.
Those are my following reasons:
- default valules are a comfortable es6 feature
- i really can't understnad why we have to add `null` to typings for things that can just be undefined when omitted and be optional arguments

src/util/misc/mergeClipPaths.ts
```diff
@@ -20,21 +21,15 @@ import { multiplyTransformMatrices, invertTransform } from './matrix';
...
   }
   //  `b` becomes `a`'s clip path so we transform `b` to `a` coordinate plane
-  applyTransformToObject(
-    b,
-    multiplyTransformMatrices(
-      invertTransform(a.calcTransformMatrix()),
-      b.calcTransformMatrix()
-    )
-  );
+  sendObjectToPlane(b, b.group?.calcTransformMatrix(), a.calcTransformMatrix());
```
asturur, line 32:
```suggestion
  sendObjectToPlane(b, b.group?.calcTransformMatrix() ?? IMatrix, a.calcTransformMatrix());
```
and let's remove the null usecase.

asturur:
Yes you win, and now i see where i m wrong.

the multiplication of calcTransformMatrix is from the outside to the inside, while i was sure it was from the inside to the outside!!

src/util/misc/mergeClipPaths.ts
ShaMan123: this resolves to undefined so it is covered

ShaMan123 (commented):
READY
If you are satisfied with naming merge it

src/util/misc/mergeClipPaths.ts
asturur: oh i thought it resolved to null, i guess too much nulls in TS in the project where i work on

asturur (approved):

asturur:
maybe this wasn't a fix(), but only an optimization or a refactor

ShaMan123:
it fixed the `null` arg usage

PR #8255, opened 2022-09-05
chore(TS): migrate `Pattern`

ShaMan123:
Migrate Pattern to TS
I thought it had default values in export, that's why I left it but it doesn't so I migrated it.

### DIFF
--- a/src/pattern.class.ts
+++ b/src/pattern.class.ts
@@ -1,181 +1,153 @@
-//@ts-nocheck
-
-
-import { pick } from "./util/misc/pick";
+import { fabric } from "../HEADER";
 import { config } from "./config";
+import { TCrossOrigin, TMat2D } from "./typedefs";
+import { loadImage } from "./util/misc/objectEnlive";
+import { pick } from "./util/misc/pick";
+import { toFixed } from "./util/misc/toFixed";
 
+export type TPatternRepeat = 'repeat' | 'repeat-x' | 'repeat-y' | 'no-repeat';
 
-(function(global) {
-  var fabric = global.fabric, toFixed = fabric.util.toFixed;
+/**
+ * Pattern class
+ * @class fabric.Pattern
+ * @see {@link http://fabricjs.com/patterns|Pattern demo}
+ * @see {@link http://fabricjs.com/dynamic-patterns demo}
+ */
+export class Pattern {
 
   /**
-   * Pattern class
-   * @class fabric.Pattern
-   * @see {@link http://fabricjs.com/patterns|Pattern demo}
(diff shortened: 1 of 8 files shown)


src/pattern.class.ts
```diff
@@ -1,196 +1,188 @@
...
-        imageOptions = Object.assign({}, options, { crossOrigin: object.crossOrigin });
-    return fabric.util.loadImage(object.source, imageOptions)
-      .then(function(img) {
-        patternOptions.source = img;
-        return new fabric.Pattern(patternOptions);
-      });
-  };
-})(typeof exports !== 'undefined' ? exports : window);
+  toSVG({ width, height }: TSize) {
+    const patternSource = this.source,
```
ShaMan123, line 159: 482ded8b

ShaMan123 (commented):
I can't get `.npmignore` to checkout...
Besides that standard es6/TS, except the commented line which is a cleanup for an old PR

src/pattern.class.ts
```diff
@@ -1,196 +1,188 @@
...
+   */
+  isImageSource(): this is TImageSource {
+    return typeof this.source.src === 'string';
+  }
+
+  /**
+   * @returns true if {@link source} is a <canvas> element
+   */
+  isCanvasSource(): this is TCanvasSource {
+    return typeof this.source === 'object' && this.source.toDataURL;
```
ShaMan123, line 105: Is this an assertion for a canvas element? Or for a fabric Object? Or both?

src/pattern.class.ts
```diff
@@ -1,196 +1,188 @@
...
+    ].join('\n');
+  }
+  /* _TO_SVG_END_ */
+
+  static async fromObject({ source, ...serialized }: TPatternSerialized, options: TPatternHydrationOptions) {
+    const img = await loadImage(source, {
+      ...options,
+      crossOrigin: serialized.crossOrigin
+    })
+    return new Pattern({ ...serialized, source: img });
```
asturur, line 183: everywhere else we use the promise syntax, let's leave that one

asturur:
> Migrate Pattern to TS I thought it had default values in export, that's why I left it but it doesn't so I migrated it.

it could have them :) but patterns are usually small.

asturur:
This is fine for where we are at.
The issue with TS and Classes and that it makes harder to `add whatever i want` to an object ( shadow, fabricObject, pattern gradient ) and export/import it.

Types makes it hard at dev time, classes makes it hard at runtime.

Now Patterns and Gradients are usually not really where people add businnes logic ( names or special props or special behavior ) so those are fine to be standard classes, with or without the default values issue

asturur (approved):

PR #8263, opened 2022-09-08
refactor(Gradient): transform

ShaMan123:
This PR refactors `Gradient`'s transform logic, 
stops using pattern to achieve transform.


ShaMan123:
it seems that gradients are rotated in 90 deg for some reason

asturur:
This is larger than i thought and while i can't even read it now, at a glance seems correct ( extracting filler in its own space and using the setTransform from patterns now that modern browsers supports it ).

Please stop large refactors like this one, this fix shouldn't have anything to do with the external interface, and will wait when 6.0 is complete.

no refactors that aren't necessary to TS adoption should be done now or this will take forever.
For how correct this can be those changes needs time to sink in and to wrap our head around on how we investigate fixes and use the code.

ShaMan123 (commented):
I have made major progress but I am stuck in the transform matrix calculation.
Help is needed

PR #8266, opened 2022-09-10
ci(): migrate scripts to es6 modules

ShaMan123:
https://nodejs.org/api/esm.html#enabling

migrate scripts to es6


package.json
```diff
@@ -51,17 +51,17 @@
   },
   "license": "MIT",
   "scripts": {
+    "fabric": "node ./scripts/index.mjs",
```
asturur, line 54: can we call this CLI or something else that is not fabric? sounds confusing to me.

scripts/index.mjs
```diff
@@ -13,25 +13,31 @@
...
+import inquirer from 'inquirer';
+import Checkbox from 'inquirer-checkbox-plus-prompt';
+import killPort from 'kill-port';
+import _ from 'lodash';
+import moment from 'moment';
+import path from 'node:path';
+import process from 'node:process';
+import { fileURLToPath } from 'node:url';
+import os from 'os';
+import { listFiles, transform as transformFiles } from './transform_files.mjs';
```
asturur, line 30: can MJS import normal commonJS modules if needed?

package.json
ShaMan123: done

scripts/index.mjs
ShaMan123: seems possible
there's also an option to import require and use that, see [SO](https://stackoverflow.com/a/56350495/9068029)

https://nodejs.org/api/esm.html#differences-between-es-modules-and-commonjs

asturur (approved):

PR #8269, opened 2022-09-11
BREAKING refactor(util): `boundingBoxFromPoints`, removed transform

ShaMan123:
- accept `n` points, not only 4
- remove optional `transform` arg (resolves safeguarding `points` from mutation)
- add test


asturur:
This is a change i wanted to do as well.
The issue here is to verify if is better the reduce on the point logic or the more basic map with math.max and math.min single call on the resulting array.

Should be able to benchmark quicjly

asturur:
definitely this is better with a single loop vs 2 necessary loops on the other way.

asturur (approved):

PR #8271, opened 2022-09-11
test(): add `globalCompositeOperation` tests

ShaMan123:
Add tests based on https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation#operations

`canvas-node` fails to render some of the operations on my machine... strange


test/visual/globalCompositeOperation.js
```diff
@@ -0,0 +1,149 @@
...
+        ctx.lineTo(oleft + r, otop + r);
+        ctx.lineTo(oleft + r + 6, otop);
+        ctx.fillStyle = gradient;
+        ctx.fill();
+        ctx.translate(oleft + r, otop + r);
+        ctx.rotate(rotate);
+        ctx.translate(-(oleft + r), -(otop + r));
+    }
+    return c;
+};
```
asturur, line 71: for the sake of tests stability, should we make an image of this and load it once at tests start?
Here we are testing low level canvas code, that under node could break.

test/visual/globalCompositeOperation.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+QUnit.module('globalCompositeOperation');
+
+OPERATIONS.map(operation => {
+    return {
+        test: operation,
+        code: (canvas, callback) => {
+            canvas.backgroundImage = createExisting();
+            canvas.add(createNew(operation), createPreview(operation));
```
asturur, line 141: this will break as soon as we fix globalCompositeOperation to be group scoped.
We should make the tests already aware of group, so adding all those elements, background, previews and circles in the same group as siblings.

asturur:
> Add tests based on https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation#operations
> 
> `canvas-node` fails to render some of the operations on my machine... strange

node-canvas has broken copy operation.
image

I remember i noticed it and wrote tests for it

Other return sligthly different results:

image

test/visual/globalCompositeOperation.js
ShaMan123: please explain

test/visual/globalCompositeOperation.js
ShaMan123: done

test/visual/globalCompositeOperation.js
asturur: global composite operation reacts differently if you are using caching or not.
so if a group is cached a circle inside a group will stop his global composite operation effect on the group itself and won't reach the canvas.
This isn't explained in fabricJS is just a side effect of caching.

global composite operation should introduce a rule like we have for shadow.
If a global composite operation is detected, caching is forced on.
Of course this will bring to some perf issues on edge cases, like a group with shadow and global composite operation.

There is also an added quirk.
I can add global composite operation on groups as a property.
What does it mean? in theory nothing, because it would be then overridden by the value of the objects inside.
But if we have a documented rule, that fabric objects do have global composite operation work from themselves on their immediate container, things becomes predictable for the developer.

ShaMan123: so the tests are aligned to your comment:
> If a global composite operation is detected, caching is forced on, and the effect of global composite operation is from an object to its container.

asturur: ok then, at worst some bit will change and we will verify why and if should chnage

ShaMan123:
Ooops sorry

בתאריך יום ב׳, 12 בספט׳ 2022, 14:34, מאת Andrea Bogazzi ‏<
***@***.***>:

ShaMan123:
Shall I make a fixture then?

PR #8274, opened 2022-09-11
refactor(util): remove lang_array since there are no more use cases

asturur:
One of the reasons i wanted that refactor of boundingbox was to support polyline and delete fabric.util.array.min/max.

Those are outdate and lodash/underscore style of utilities that we should not offer in our interface


asturur:
Ready

ShaMan123 (approved):
good riddance

asturur:
later the conversion between objects and points will be done in the initilization part.
for now is just an error guard.

PR #8276, opened 2022-09-11
chore(): prettify all source code

asturur:
In theory no changes here.


asturur:
If there is a stlye rule ( those are all defaults from prettier ) that you don't like this is the moment to talk about it.

src/color/color.class.ts
```diff
@@ -34,18 +31,16 @@ export class Color {
...
-      || Color.sourceFromRgb(color)
-      || Color.sourceFromHsl(color)
-      // color is not recognize let's default to black as canvas does
-      || [0, 0, 0, 1]
+    const source =
+      color === 'transparent'
+        ? [255, 255, 255, 0]
+        : Color.sourceFromHex(color) ||
+          Color.sourceFromRgb(color) ||
+          Color.sourceFromHsl(color) || [0, 0, 0, 1]; // color is not recognize let's default to black as canvas does
```
ShaMan123, line 43: I don't like this

ShaMan123 (commented):
I don't like 
```js
condition
? value
: other
```

and prefer
```js
condition ?
   value :
   other;
```

BUT as long as VSCODE formats it for me f*** it

ShaMan123:
I wonder how hard it will be to rebase PRs from before the migration

src/color/color.class.ts
asturur: see if is configurable and change it

asturur:
> I wonder (and dread) how hard it will be to rebase PRs from before the migration

Copy and reapply the changes. Indeed we waited too much to do this.

ShaMan123:
which do you prefer? https://eslint.org/docs/latest/rules/operator-linebreak

ShaMan123:
What is the difference between this https://github.com/prettier/eslint-plugin-prettier and https://github.com/prettier/eslint-config-prettier?

Anyways I managed to get it more or less as I prefer but only when I run lint. prettier doesn't do it

ShaMan123:
It seems that https://github.com/prettier/eslint-config-prettier disables the rule I added when running with prettier

image

asturur:
the issue is that that is a Lint rule. Prettier just does formatting and has a set of lint rules with the intent of being only disabled, to aovid overlapping.

[eslint-config-prettier](https://github.com/prettier/eslint-config-prettier): Disable eslint rules that overlap with prettier.
[eslint-plugin-prettier](): Run prettier as a lint rule

I would prefer separation of concern between formatting and lintable bugs

asturur:
ok maybe we have to swallow it. This is what a popular tool offer and debating where the question mark goes is wasted time.
I didn't like the else break in more lines,
```js
}
else {
```
But i had to keep it for 7 long years

ShaMan123:
Yeah yeah sure.

ShaMan123 (approved):

ShaMan123:
Isn't `.eslintrc.json` a dead file?

asturur:
> Isn't `.eslintrc.json` a dead file?

Could be that without JS files anymore is dead.

PR #8277, opened 2022-09-12
chore(lint): disallow `Math.hypot`, `window`, `document`

ShaMan123:
I found out how to use [no-restricted-syntax](https://eslint.org/docs/latest/rules/no-restricted-syntax) to match `Math.hypot` and [no-restricted-globals](https://eslint.org/docs/latest/rules/no-restricted-globals) to match `window` and `document`

## Motivation
https://stackoverflow.com/questions/62931950/different-results-of-math-hypot-on-chrome-and-firefox

## In Action

image
image
image
image
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,22 +18,22 @@ module.exports = {
       'error',
       {
         name: 'window',
-        message: 'Use fabric window or other utils instead'
+        message: 'Use fabric window or other utils instead.'
       },
       {
         name: 'document',
-        message: 'Use fabric document or other utils instead'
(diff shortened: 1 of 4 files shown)


ShaMan123 (commented):
Why doesn't the lint job detect the error?

ShaMan123:
image

ShaMan123 (commented):
The `window` rule will fail as long as there are existing cjs export directives
`})(typeof exports !== 'undefined' ? exports : window);`

asturur:
ok, is fine anyway, even if window fails for now.

asturur (approved):

PR #8292, opened 2022-09-14
task(IText): rename `fabric-data` attr

ShaMan123:
Align with the rest of HTML elements

### DIFF
--- a/src/mixins/itext_key_behavior.mixin.ts
+++ b/src/mixins/itext_key_behavior.mixin.ts
@@ -16,7 +16,7 @@ import { config } from '../config';
         this.hiddenTextarea.setAttribute('autocorrect', 'off');
         this.hiddenTextarea.setAttribute('autocomplete', 'off');
         this.hiddenTextarea.setAttribute('spellcheck', 'false');
-        this.hiddenTextarea.setAttribute('data-fabric-hiddentextarea', '');
+        this.hiddenTextarea.setAttribute('data-fabric', 'textarea');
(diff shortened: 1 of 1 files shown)


src/mixins/itext_key_behavior.mixin.ts
```diff
@@ -16,7 +16,7 @@ import { config } from '../config';
         this.hiddenTextarea.setAttribute('autocorrect', 'off');
         this.hiddenTextarea.setAttribute('autocomplete', 'off');
         this.hiddenTextarea.setAttribute('spellcheck', 'false');
-        this.hiddenTextarea.setAttribute('data-fabric-hiddentextarea', '');
+        this.hiddenTextarea.setAttribute('data-fabric', 'textarea');
```
asturur, line 19: i have no much to say about this change, is neutral, but i want to point out a few things.
data-* does dover both data-fabric and data-fabric-attribute.
The assumed consistency gain is killed by the required change of code of whoever is using this.
This is really just a string, an identified, we can change it yes.
But i do think is a waste of energy.

asturur (approved):

src/mixins/itext_key_behavior.mixin.ts
ShaMan123: point taken
I came across it by chance
I view it as something less to document
Need we can say "use the querySelector `data-fabric` to get fabric elements", the value is one of...
I didn't think of the option `data-fabric-*`...

PR #8295, opened 2022-09-16
fix(`Object#_fromObject`): `extraParam` should not be passed to options

### LINKED ISSUE #8294: bug(): options overriding args
## Version

5.2.4

## Test Case

https://codepen.io/snap/pen/poVRgvJ?editors=1010

All major browsers (tested in Chrome, Firefox, Safari on MacOS)

## Steps To Reproduce

1. Add extended Textbox (as Textarea in my case) to the canvas
2. Save the canvas to JSON
3. Load the JSON back to the canvas
4. Create and add another extended Textbox instance (with another text) to the canvas
5. The added instance "inherits" the text from the loaded instance instead of the passed text

## Expected Behavior

After saving and loading an extended Textbox instance from JSON, and creating a new instance passed with a text, it should set and display the passed text.

- Without instance in loaded JSON it works as expected: https://codepen.io/snap/pen/RwyKrgE?editors=0010
- Without saving and loading from JSON it works as expected: https://codepen.io/snap/pen/bGMgVXy?editors=0010

## Actual Behavior

When adding a new extended Textbox instance after loading another instance from JSON, it will "inherit" the text of the loaded instance instead of the newly passed text.

Demo: https://codepen.io/snap/pen/poVRgvJ?editors=0010

## Workaround

Setting the text after calling super on "initialize" fixed it for me:

```js
this.callSuper("initialize", text, options);
this.set("text", text);
```

Fixed: https://codepen.io/snap/pen/PoeWPOe?editors=0010

ShaMan123:
fixes #8294 
but I think we should make assigning strict

### DIFF
--- a/src/shapes/object.class.ts
+++ b/src/shapes/object.class.ts
@@ -2061,6 +2061,8 @@ import { enlivenObjectEnlivables } from '../util/misc/objectEnlive';
   ) {
     return enlivenObjectEnlivables(clone(object, true), options).then(
       (enlivedMap) => {
+        // from the resulting enlived options, extract options.extraParam in arg0
+        // to avoid accidental ovverrides later
         const { [extraParam]: arg0, ...rest } = { ...options, ...enlivedMap };
         return extraParam ? new klass(arg0, rest) : new klass(rest);
       }
--- a/src/shapes/object.class.ts
+++ b/src/shapes/object.class.ts
@@ -2054,14 +2054,17 @@ import { enlivenObjectEnlivables } from '../util/misc/objectEnlive';
    * @param {AbortSignal} [options.signal] handle aborting, see https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal
    * @returns {Promise<fabric.Object>}
    */
-  fabric.Object._fromObject = function (klass, object, { extraParam, ...options } = {}) {
-    return enlivenObjectEnlivables(clone(object, true), options)
-      .then((enlivedMap) => {
+  fabric.Object._fromObject = function (
+    klass,
+    object,
+    { extraParam, ...options } = {}
+  ) {
+    return enlivenObjectEnlivables(clone(object, true), options).then(
+      (enlivedMap) => {
         const { [extraParam]: arg0, ...rest } = { ...options, ...enlivedMap };
-        return extraParam
-          ? new klass(arg0, rest)
-          : new klass(rest);
-      });
+        return extraParam ? new klass(arg0, rest) : new klass(rest);
+      }
+    );
   };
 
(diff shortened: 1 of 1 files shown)


ShaMan123 (commented):
The following shapes use `extraParams`:
- Polyline
- Polygon
- Path
- Text, IText, Textbox

Path is the only one safeguarded from this case

ShaMan123:
I am for deprecating `extraParam` and consolidating all shapes to
```
new Shape({...whatever})
```

src/shapes/object.class.ts
```diff
@@ -2053,16 +2054,17 @@ import { runningAnimations } from '../util/animation_registry';
...
+  fabric.Object._fromObject = function (
+    klass,
+    object,
+    { extraParam, ...options } = {}
+  ) {
+    return enlivenObjectEnlivables(clone(object, true), options).then(
+      (enlivedMap) => {
+        const { [extraParam]: arg0, ...rest } = { ...options, ...enlivedMap };
+        return extraParam ? new klass(arg0, rest) : new klass(rest);
+      }
```
asturur, line 2066:
```suggestion
      (enlivedMap) => {
        // from the resulting enlived options, extract options.extraParam in arg0
        // to avoid accidental ovverrides later
        const { [extraParam]: arg0, ...rest } = { ...options, ...enlivedMap };
        return extraParam ? new klass(arg0, rest) : new klass(rest);
      }
```

asturur: @ShaMan123 please add the comment and we can merge this

asturur:
> I am for deprecating `extraParam` and consolidating all shapes to
> 
> ```
> new Shape({...whatever})
> ```

This could be an idea, but i dont' want to add it on top of all the rest is changing. Is like if group would have objects in the options. There is a clearly distintive trait of each object that was added there.
Of course something was missed. Circle would have been radius, Rect would have been width AND height. ellipses the 2 radius and so on.

We can think of adding it with backward compatibility after 6.0 is released, and deprecate for 7.0, as point does.

asturur (approved):

ShaMan123:
I have another point for the class discussion regarding safegurading `set` during initialization.
Because the options object is iterated over in a given order it is possible to encounter uninitialized errors in side effects trying to run on the instance before it is hydrated.

ShaMan123:
I propose the `set` method returns a side effect function.
Then we can accumulate them in the constructor and run them after hydration.
It is also good as a perf enhancement for setting an object that run several side effects, some might be the same (e.g. setting `fontSize`, `fontFamily` triggers dimension calculation on Text)
Thoughts?

PR #8297, opened 2022-09-18
chore(TS): animation

### LINKED ISSUE #8241: Animate color seems broken in current master
The recent addition fo animation to handle arrays have taken over the color animation that is represented with an array of [r, g, b, a].
Unit test pass for some reason.

- investigate actual situation of animateColor
- if broken, write a unit test that breaks because it is broken
- with a broken unit test, fix animate color so that the test start to pass.

Lazauya:
## Changes
- Refactor animation logic into classes:
  - `AnimationBase` contains all the logic
  - Subclasses handle initialization logic and impl `calculate` that returns the current value of the animation
- animate color
   - closes #8241 
   - align class and signature with peers
      -  `animateColor(from ,to ,duration, options)` => `animateColor(options)`
      - `colorEasing` integrated into default easing of `ColorAnimation`
   - fixed `byValue` option
- fix end value calculation: 
there was a conflict happening in the cases:
   -  passing byValue without endValue since defaults are assigned to both, endValue and byValue diverged
   - passing diverged byValue, endValue
   - this fix uncovered bugs in easing functions that I fixed  
- Color:
  - accept more param types (color source, color instance) in constructor and deprecate `fromSource` as it has become redundant
- **BREAKING**: 
return animation instance from animate instead of a cancel function and remove `findAnimationByXXX` from `AnimationRegistry`

## Gist

@asturur there are too many threads in this PR. All are pretty much stale accept https://github.com/fabricjs/fabric.js/pull/8297#pullrequestreview-1200467329 that I would like you to comment on.

## Proposals
- What about making `AnimationBase` extend Observable and fire state events?
```ts
const context = animate({ ... })
context.on('change', () => {})

```

- It is possible to make `animate` accept also a color animation options object 
```ts
animate({
  startValue: [r,g,b,a] // TS will find this hard to handle since it is an array and will fall into the array animation case
  type: 'color' // this will do the job
})
```

- accept color values in color animation
```ts
animate({
  startValue: 'colorString' // TS will find this easy to understand and handle as `ColorAnimation`
})
animate({
  startValue: new Color('colorString') // even better, making passing `type: color` redundant
})
```

- accept color space option for color animation 
```ts
animate({
  startValue: new Color('colorString') // even better, making passing `type: color` redundant
  colorSpace: 'hsl'
})
```


src/util/animate.ts
```diff
@@ -1,34 +1,124 @@
...
- * @typedef {(AnimationOptions & AnimationCurrentState & { cancel: CancelFunction }} AnimationContext
+ * Callback called every frame
+ * @param t current "time"/ms elapsed. multivalue
+ * @param valueRatio ratio of current value to animation max value. [0, 1]
+ * @param timeRatio ratio of current ms to animation duration. [0, 1]
  */
+export type OnChangeCallback = (
+  t: number | number[],
+  valueRatio: number,
+  timeRatio: number
```
ShaMan123, line 16: I think we should inline the comments so TS will show them when it expands the param

```ts

  /**
   * current "time"/ms elapsed. multivalue
   * */
  t: number | number[],
```

src/util/animate.ts
```diff
@@ -1,34 +1,124 @@
...
+) => void;
```
ShaMan123, line 17: What about making this a generic returning `T` defaulting to `unknown`?
Then you don't need `WithReturnType`

src/util/animate.ts
```diff
@@ -1,34 +1,124 @@
...
+  /**
+   * Called after the last frame of the animation
+   */
+  onComplete: OnChangeCallback;
+
+  /**
+   * Easing function
+   * @default [defaultEasing]
+   */
+  easing: EasingFunction;
```
ShaMan123, line 58: optional

src/util/animate.ts
```diff
@@ -1,34 +1,124 @@
...
+
+  /**
+   * Function called at each frame.
+   * If it returns true, abort
+   */
+  abort: AbortCallback;
```
ShaMan123, line 64: optional

src/util/animate.ts
ShaMan123: naming: `durationRatio`? What do you think?
We should consolidate everything to the same naming, I am open to what is best

src/util/animate.ts
ShaMan123: all options are optional I think

Ahh I see you use Partial

src/util/animate.ts
```diff
@@ -71,8 +163,9 @@ export function animate(options = {}) {
     delay = 0,
   } = options;
 
-  const context = {
+  const context: Partial<AnimationContext> = {
     ...options,
+    cancel: noop, // placeholder
```
ShaMan123, line 168: can we put the cancel logic here? Instead of assinging it later on

src/util/animate_color.ts
```diff
@@ -46,40 +62,46 @@ const defaultColorEasing = (currentTime, duration) =>
...
-    ...restOfOptions,
+    target,
     duration,
+    abort,
     startValue: startColor,
     endValue: endColor,
     byValue: endColor,
     easing: (currentTime, startValue, byValue, duration) =>
-      calculateColor(startValue, byValue, colorEasing(currentTime, duration)),
+      colorEasing(currentTime, duration),
```
ShaMan123, line 92: this is a change, not just types

I am fine with it as it aligns with animation core

src/constants.ts
```diff
@@ -1,7 +1,7 @@
 import { TMat2D } from './typedefs';
 
 export { version as VERSION } from '../package.json';
-export function noop() {}
+export function noop(...args: unknown[]): any {}
```
ShaMan123, line 4: necessary for types?

ShaMan123 (changes requested):
I left a few comments, nothing major
Checkout `package-lock.json` from master (upstream probably) so it is removed from the PR diff
And run prettier

src/constants.ts
asturur: Is typescript complaining that a function is either a function or default value noop?

src/util/anim_ease.ts
```diff
@@ -1,10 +1,18 @@
...
+ * @param b start/"base" value
+ * @param c increment/"change"/"completion rate"/magnitude
+ * @param d "duration"
+ * @returns next value
+ */
+export type EasingFunction = (
+  t: number,
+  b: number,
+  c: number,
+  d: number
```
asturur, line 15: this is a downgrade in clarity.
The function has 4 arguments, that for brevity years ago have been called t,b,c,d.
Now is extremely unclear what they do, and so in the type function those have been given a clear name.
Isn't this enough to have the clear name show in vscode? i would really leave the type as it is, and add the documentation on top if is necessary for some other tool

asturur: Please restore this type as it was, both with the T and the names.

src/util/animate.ts
asturur: i think timeRatio is more something like, time elapsed in %. Not sure how can it be described properly as a parameter

src/util/animate.ts
```diff
@@ -89,15 +182,15 @@ export function animate(options = {}) {
   };
   runningAnimations.push(context);
 
-  const runner = function (timestamp) {
+  const runner = function (timestamp: number) {
     const start = timestamp || +new Date(),
       finish = start + duration,
       isMany = Array.isArray(startValue),
```
asturur, line 188: sometimes i see this done:

function isValueMany(value: number | number[]): value is number[] {
  return Array.isArray(value);
}

We could do `isMany = isValueMany(startValue) && isValueMany(endValue)`

And this could avoid having to then do the as number[] in the ternary.
It would clarify we verified those are arrays

src/util/animate_color.ts
```diff
@@ -46,40 +62,46 @@ const defaultColorEasing = (currentTime, duration) =>
...
-            calculateColor(current, current, 0),
-            valuePerc,
-            timePerc
-          );
-        }
-        onChange(current, valuePerc, timePerc);
-      }
-    },
+    onChange: (current, valuePerc, timePerc) =>
+      onChange?.(
```
asturur, line 97: we slipped in a bunch of optional chaining operator, but those are not supported before chrome 80.
We didn't fully decide yet which browsers are we going to support, in this case better leave the code as it was.

src/util/animate_color.ts
asturur: Not in a typing pr and not with a clear explanation of what is changing and how.

src/util/animate.ts
```diff
@@ -1,34 +1,124 @@
...
+export type CancelFunction = VoidFunction;
+
+/**
+ * Animation of a value or list of values
+ */
+export interface AnimationOptions {
+  /**
+   * The object this animation is being performed on
+   */
+  target?: unknown;
```
asturur, line 37: is this referenced anywhere? i can't find it used in the code

src/constants.ts
Lazauya: IDK but I did this so that I could use `noop` for functions that took an arbitrary number of params and returned anything. Specifically I needed it for the onChange type funcs in the animation.

src/util/animate_color.ts
```diff
@@ -46,40 +62,46 @@ const defaultColorEasing = (currentTime, duration) =>
...
+    abort?: AbortCallback;
+    target?: unknown;
   } = {}
-) {
+): CancelFunction {
   const startColor = new Color(fromColor).getSource(),
     endColor = new Color(toColor).getSource();
   return animate({
-    ...restOfOptions,
+    target,
```
asturur, line 85: can you explain this change?
target seem to have been introduced in this PR, and removing ...restOfOptions cuts away using animate_color as animate ( delay and onStart are missing to make an example ) requiring changes in 2 places when something else gets added to animate

asturur:
there are at least 3 behavor change i would like to see reverted, and it seems to me you added a `target` new option that should be proposed on a separate PR.

src/util/animate_color.ts
Lazauya: target was not introduced here, it was just apparently never documented. I found that out when I ran the tests. Look at the animation.minix.ts _animate function and the animation unit tests, as well as the animation_registry.ts.

src/util/animate_color.ts
Lazauya: I can remove the chaining ops, but typescript compiles it down into unchained anyway so I don't get why we should remove them

src/util/animate.ts
Lazauya: It is referenced in animation.minix.ts _animate function and the animation unit tests, as well as the animation_registry.ts.

src/util/anim_ease.ts
Lazauya: I will restore the names, I started writing this a while ago and I didn't realize that the easing had been changed when I merged from master, derp my bad.
Still if the names change then I think it would be a good idea to change the easing function signatures because now they don't match anymore.

Lazauya:
> there are at least 3 behavor change i would like to see reverted, and it seems to me you added a `target` new option that should be proposed on a separate PR.

The only ACTUAL behavior change is that the currentState of a color animation is now a TColorAlphaSource instead of a color string. See my other comments about the target stuff. I can change it but the problem is its SUPER hard to generify the `animate` util function to work with "normal" animations and color animations. It took me a long time to realize that just changing `animateColor` match `animate` would be more clean in the end. I can make another PR specifically for that I suppose if that would be better.

src/util/animate.ts
Lazauya: yeah I think so

src/util/animate.ts
Lazauya: Hm I tried this and it seems that it only confuses the typing system more because it doesn't clarify that if they aren't arrays they are numbers, at least according to typescript. Or maybe I did it wrong lol.

src/util/anim_ease.ts
Lazauya: Also I don't think clear names are enough, currentTime for instance could mean a lot of things or be in a lot different units. IMO specifying units in documentation is SUPER important

ShaMan123: Why not naming with units?
e.g. `durationMs`

src/util/animate.ts
ShaMan123: I wanted to suggest that too but couldn't get it to work as well.
Type assertion don't negate unfortunately so you will need to write 2 type assertion functions.

src/util/animate.ts
ShaMan123: @Lazauya is correct, it is used by the animation registry to `cancelByTarget`

ShaMan123: types should be adapted there as well and liked to here

src/util/animate.ts
ShaMan123: I got it wrong before
now is good

```tsx
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

function isMany<T, S extends T[]>(t:T|S): t is S {
    return Array.isArray(t)
}


function go(b:number|number[]) {
    if(isMany(b)){
        b.push(1)
    } else {
        b+1;
    }
}
```

[Playground Link](https://www.typescriptlang.org/play?ts=4.8.2#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC2AZuur5IxBtVDE4AWRqYAPAAqADSgAMqgOAAeltSscKCBANoAugB8ABSQCIEAPmEAlAigkJ4JEQDetqA1oMaQ6oQeAILSlJhoXq2E7VkFtgC+tsM2js6u7qC0DBl8CNTqSnw4hLkLSyupBaBVNrWe9hlevtSYswUFu-v7fGgwmuAZAIz9e7UDkeRwODvV1zV8ADUTwA3H8akMbJCgA)

src/util/animate.ts
```diff
@@ -176,10 +178,6 @@ export function animate(
     return index > -1 && runningAnimations.splice(index, 1)[0];
```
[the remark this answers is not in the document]
ShaMan123: Actually I think it should be handled by the registry

src/util/animate_color.ts
ShaMan123: d2075e6

ShaMan123 (commented):
typed animation registry to avoid confusion

src/util/animate.ts
ShaMan123: d2075e6

src/util/animate.ts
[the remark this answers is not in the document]
ShaMan123: Rethinking
The context should hold a `cancelled` flag instead of the `cancel` variable
Then registry should expose `register(context): CancelFunction`
That makes more sense IMO

src/constants.ts
asturur: yes so is a typescript thing.
This thing of TS is bothering a lot to me, if you provided an hardcoded default value in the code, and it does not match with the other values...
I wonder why noop is not a recognized type of thing, since is an universal way to replace a function that does nothing.
Especially since onChange does not return anything

src/util/anim_ease.ts
asturur: i m not against providing extra documentation, i just don't want to call those t,b,c,d.
Changing the easing function signatures could be done, in a separate PR or will bump up this in number of change and make hard to read what is changed.
Changes that are large and non significative for functionality should be done in isolation so that is easy to focus on changes.

src/util/animate.ts
asturur: well then typescript sucks.
It can't done both are arrays, and it can't do is array or number depending on this boolean.
Ts sucks

ShaMan123: Even better


```tsx
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

function isMany<T, S extends T extends unknown[]?T:T[]>(t:T|S): t is S {
    return Array.isArray(t)
}


function go(b:number|number[]) {
    if(isMany(b)){
        b.push(1)
    } else {
        b+1;
    }
}

```

[Playground Link](https://www.typescriptlang.org/play?ts=4.8.2#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC2AZuur5IxBtVDE4AWRqYAPAAqADSgAMqgOAAeltSscKCBkTE4cQlOQtSy1ADaALoA-IEIgfkAfAAUkCUAPmEAlAigkJ4JEQDetqDdoMaQ6oQeAILSlJhoXiOEY1X1tgC+tks2js6u7qC0DBV8CNTqSnw4hDX7h8f59aCdNj2e9hVevtSYO-X1N3d3fGgwmuAVACMc1uPXmkXIcBw1y6X26fAA1ICANyw7qLGwYoA)
```

[Playground Link](https://www.typescriptlang.org/play?ts=4.8.2#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC2AZuur5IxBtVDE4AWRqYAPAAqADSgAMqgOAAeltSscKCBkTE4cQlOQtSy1ADaALoA-IEIgfkAfAAUkCUAPmEAlAigkJ4JEQDetqDdoMaQ6oQeAILSlJhoXiOEY1X1tgC+tks2js6u7qC0DBV8CNTqSnw4hDX7h8f59aCdNj2e9hVevtSYO-X1N3d3fGgwmuAVACMc1uPXmkXIcBw1y6X26fAA1ICANyw7qLGwYoA)

File: src/util/animate_color.ts (hunk 10 above)
Comment (asturur) [reply src/util/animate_color.ts:97]: because typescript compile down a lot of stuff that we don't let it do.
For now we let the browsersupport of rollup compile the stuff for us.
I couldn't get Typescript to choose browser granularity in chosing what to compile to, and at the end what you have to support is browsers.

File: src/util/animate_color.ts (hunk 12 above)
Comment (asturur) [reply src/util/animate_color.ts:85]: I get target now, but the rest of the question still stands.
Why did you choose to remove ...restOfOptions ? animate color takes every option than animate does, and so we should pass down with a transparent pass through

File: src/util/animate.ts (hunk 9 above)
Comment (ShaMan123) [reply src/util/animate.ts:187]: it's not the first time we comment in a race condition... 😅

File: src/util/animate.ts (hunk 11 above)
Comment (asturur) [reply src/util/animate.ts]: > > there are at least 3 behavor change i would like to see reverted, and it seems to me you added a `target` new option that should be proposed on a separate PR.
> 
> The only ACTUAL behavior change is that the currentState of a color animation is now a TColorAlphaSource instead of a color string. See my other comments about the target stuff. I can change it but the problem is its SUPER hard to generify the `animate` util function to work with "normal" animations and color animations. It took me a long time to realize that just changing `animateColor` match `animate` would be more clean in the end. I can make another PR specifically for that I suppose if that would be better.

I think the 3 behavor change are:

- removal of `...restOptions`
- remove of calculateColor where @ShaMan123 pointed out
- on the onChange part assuming that the value is always an array and removing the array check

Arent' those changes?

Adding in pr description `modified the color animate function to be more generic.` doesn't help much in understanding why are you changing things, on top of that those changes should be factored away from a typing pr.

**Comment by asturur** [comment]
> > there are at least 3 behavor change i would like to see reverted, and it seems to me you added a `target` new option that should be proposed on a separate PR.
> 
> The only ACTUAL behavior change is that the currentState of a color animation is now a TColorAlphaSource instead of a color string. See my other comments about the target stuff. I can change it but the problem is its SUPER hard to generify the `animate` util function to work with "normal" animations and color animations. It took me a long time to realize that just changing `animateColor` match `animate` would be more clean in the end. I can make another PR specifically for that I suppose if that would be better.


I think the 3 behavor change are:

removal of ...restOptions
remove of calculateColor where @ShaMan123 pointed out
on the onChange part assuming that the value is always an array and removing the array check
Arent' those changes?

Adding in pr description modified the color animate function to be more generic. doesn't help much in understanding why are you changing things, on top of that those changes should be factored away from a typing pr.

File: src/util/animate_color.ts (hunk 12 above)
Comment (Lazauya) [reply src/util/animate_color.ts:85]: I chose to remove it because I didn't know it was intended for the object to just be all other animation options. I can take a look and refactor it.  I'm thinking that for this PR I'm going to take animate_color out. I'll submit another PR since the changes are too big.
Like I said AFAIK there's no easy way to generify `animate` so that it works well with the current animate color.

Comment (asturur) [reply src/util/animate_color.ts:85]: there is no easy way no, and that is why it was a different function.
Normal easing functions do not really make sense for colors.
The issue is that i still convinced that since we introduced array support for animate we did break animate_color

**Comment by Lazauya** [comment]
I went ahead and removed animate color stuff. I'm putting that in a new PR where I'll clean it up.

**Comment by Lazauya** [comment]
@ShaMan123 @asturur Any outstanding issues with this?

File: src/util/animate.ts (hunk 9 above)
Comment (ShaMan123) [reply src/util/animate.ts:187]: @Lazauya I would like implemented

Comment (ShaMan123) [reply src/util/animate.ts:187]: I don't mind doing it.
And make all the interfaces generic `AnimationOptions<T exnteds number | number[]>`

**Review by ShaMan123 — COMMENTED** [review]
What is the status of the isMany typing?
I would like that in this PR

**Comment by ShaMan123** [comment]
In order to pass tests:
- Add a changelog entry describing changes
- run `npm run prettier:write` and commit

**Comment by Lazauya** [comment]
> What is the status of the isMany typing?
> I would like that in this PR

The issue that I'm facing (maybe there's a smart solution) is that if you do something like

```
const isMany = isManyFunc(start) && isManyFunc(end);
```

Typescript can easily see that any branch where `isMany` is true is means that `start` and `end` are both arrays. But if `isMany` is false it doesn't necessarily mean that both `start` and `end` are numbers, just that they aren't BOTH arrays. Like I said, maybe there's a better solution. I think maybe to have some kind of assert that throws an error for invalid param types would be nice, but I'm not sure if that's in the scope for this PR.

**Comment by ShaMan123** [comment]
```tsx
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

function isMany<T, S extends T[]>(t:T|S): t is S {
    return Array.isArray(t)
}


function go(b:number|number[]) {
    if(isMany(b)){
        b.push(1)
    } else {
        b+1;
    }
}
```

[Playground Link](https://www.typescriptlang.org/play?ts=4.8.2#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC2AZuur5IxBtVDE4AWRqYAPAAqADSgAMqgOAAeltSscKCBANoAugB8ABSQCIEAPmEAlAigkJ4JEQDetqA1oMaQ6oQeAILSlJhoXq2E7VkFtgC+tsM2js6u7qC0DBl8CNTqSnw4hLkLSyupBaBVNrWe9hlevtSYswUFu-v7fGgwmuAZAIz9e7UDkeRwODvV1zV8ADUTwA3H8akMbJCgA)

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/pull/8297#discussion_r973896434_

Lazauya:
> ```tsx
> ```tsx
> // https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates
> 
> function isMany<T, S extends T[]>(t:T|S): t is S {
>     return Array.isArray(t)
> }
> 
> 
> function go(b:number|number[]) {
>     if(isMany(b)){
>         b.push(1)
>     } else {
>         b+1;
>     }
> }
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> Playground Link
> Originally posted by @ShaMan123 in #8297 (comment)
> ```

Right but I'm talking about [this](https://www.typescriptlang.org/play?ts=4.8.2#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC2AZuur5IxBtVDE4AWRqYAPAAqADSgAMqgOAAeltSscKCBANoAugB8ABSQCIEAPmEAlAigkJ4JEQDetqA1oMaQ6oQeAILSlJhoXq2E7VkFtgC+tsM2js6u7qC0DBl8CNTqSnw4hLkLSyupoZTzi8ur6-upBaBVNrWe9hlevtSYsycAZI9lt-eUBQVnFxd8aDCacAZACMBQA3NUfjVKP9ASDwZCagNIuQ4DhToifnwANSUCHnWpDGxEhxOFxuDzTUSzXYbA57TYpba0-ZrBmEY4Ygk1YhXG5+B5fTG-WFwIGg-FQ6Gi8UI7mgZE4VHo75SnF4zFEgZAA):

```
function isMany<T, S extends T[]>(t:T|S): t is S {
    return Array.isArray(t)
}


function go(b:number|number[], a:number|number[]) {
    if(isMany(b) && isMany(a)){
        b.push(1);
        a.push(1);
    } else {
        b+a;
    }
}

function go2(b:number|number[], a:number|number[]) {
    if(isMany(b)){
        b.push(1);
        a.push(1);
    } else {
        b+a;
    }
}
```

Both of these result in errors, and this is basically what happens in `animate` with `startValue` and `endValue`.

Lazauya:
As an addendum, maybe we can do something like [this](https://www.typescriptlang.org/play?ts=4.8.2#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC2AZuur5IxBtVDE4AWRqYAPAAqADSgAMqgOAAeltSscKCBANoAugB8ABSQCIEAPmEAlAigkJ4JEQDetqA1oMaQ6oQeAILSlJhoXq2E7VkFtgC+tsM2js6u7qC0DBl8CNTqSnw4hLkLSyupoZTzi8ur6-upBaBVNrWe9hlevtSYsycAZI9lt-eUBQVnFxd8aDCacAZACMBQA3NUfjVKP9ASDwZCagNIuQ4DhLhkAIQ3PwPUDPUDYny4j4nb5Q0B8ADUlAh51qyJwqPR5KhkHAMmQoAAotImBkAETeYhRbSgGBSShKBJOODqGAwJiWVgC-r0pGDIA):

```
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

function isMany<T, S extends T[]>(t:T|S): t is S {
    return Array.isArray(t)
}


function go(b:number|number[], a:number|number[]) {
    if(isMany(b) && isMany(a)){
        b.push(1);
        a.push(1);
    } else if(!isMany(b) && !isMany(a)) {
        b+a;
    } else {
        throw Error("Mixing params unsupported")
    }
}
```

But I think adding errors like this might go outside the scope of a typing PR. Dunno.

ShaMan123:
OK I see

Ignore naming...


```ts
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

type Dump<T> = {a:T,b:T}

function areMany<T extends number, S extends T[], Arr extends Dump<S>>(x:Dump<T>|Arr): x is Arr {
    return Array.isArray(x.a)
}


function go(x:{b:number, a:number}|{b:number[], a:number[]}) {
    if(areMany(x)){
        x.b.push(1);
        x.a.push(1);
    } else {
        x.a+x.b
    }
}

```



**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC39OUABF1SmAB4AKgD5QAXlAAb0oETwAaPlCAX1tbADN1anxIYgZqUCkcAFkaTC9QHAAPS2pWOFBqVz4cQjDQAGUC4pxS8s8AbQBdOoBBaSaSsudXD3rvbwAKQoQXNy9vAB8+wgBKBFBC0GJy5aDbUAPQY0h1QnTlykw0bYvMKbRKFdsYu1eEpJS00FoGKYRAyKVJTVWoZBBAkFRBYA8FVGpdOohCHwzpRFZ7GyHLZxCaZHLUO6FFYrQL7LGHQpoPhoGCacATACMKwA3GTyRsHjS6YyWWzQFECuQ4I5SZj2RzKABqSl8PkvF5AA)

ShaMan123:
T doesn't need to extend number. I can say generic

ShaMan123:
```ts
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

type Dump<T> = Record<string,T>

function areMany<T, Arr extends Dump<T[]>>(x:Dump<T>|Arr): x is Arr {
    return Array.isArray(x.a)
}


function go(x: Dump<number>|Dump<number[]>) {
    if(areMany(x)){
        x.b.push(1);
        x.a.push(1);
    } else {
        x.a+x.b
    }
}

```



```ts
"use strict";
// https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates
function areMany(x) {
    return Array.isArray(x.a);
}
function go(x) {
    if (areMany(x)) {
        x.b.push(1);
        x.a.push(1);
    }
    else {
        x.a + x.b;
    }
}

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/PTAEAsBdIBwZwFwgO6oHSQJ4wKZwMYBOAljJADYCGAdgOZoD2htwAJg-nMODawEYMGAa2AAmYNUqFCDZMTpooAW3IBiAK5x5tALRZcOmIRyti+SpDwBYAFC39OUABF1SmAB4AKgD5QAXlAAJRx8JlZ3OEgSOgAaH1tbADN1anxIYgZqUCkcAFkaTC8Y0ABBaVAcAA9LalY4Z1cPTwBtAF1vbwAKSoQXNy9vAB8ywgBKBFBK0GJ6kdAAb1tQZdBjSHVCLJHKTDQZ7cxutEpR2wBfBLsbZNT0zNBaBm6Jvo9qVz4cQiHX93elT6ENreUYLJYrYiJTo5fLUQ6VUajRY2FaoyZoPhoGCacCdACMowA3OC0ejKFicfiiSTQGcKuQ4I5kaTlpVjgBqNl8GkXGy8oA)

Lazauya:
ok @ShaMan123 please take a look. Not sure if you you'll like it (I'm not super satisfied but I couldn't find a better way to do it). I tried implementing the `areMany` function that you showed. I called it `isMulti`. The biggest issue that I found was that I couldn't use array's `.map` function because somehow using a nested function like that confuses typescript and makes it lose the type info obtained from `isMulti`. Otherwise I generified `AnimationOptions` and I'm pretty happy with it now since it should be hard to mix and match arrays and number values now.

ShaMan123 (changes requested):
@Lazauya I don't like what happened to `src/util/animate.ts`, it is a mess.

Before stepping into isMany I suggest we make files more readable by extracting to files and nesting all under a directory:

https://github.com/Lazauya/fabric.js/compare/laz-animate-typing...fabricjs:fabric.js:ts-animation

If we decide to that in a follow up I suggest you checkout  `src/util/animate.ts` and start again. That's what I did in `ts-animation`
I will look into isMany on that branch

Lazauya:
@ShaMan123 Ok, I don't have an issue with splitting things up more. However, I noticed a really big mistake that I was making that you should be weary of. If you do `type MyType<T extends number | number[] = number | number[]> = { a: T; b: T }` this allows you to mix and match numbers and number arrays which is the exact behavior we need to avoid. Instead you need to do something like `type MyTypeBase<T> = { a: T; b: T }; type MyType = MyTypeBase<number> | MyTypeBase<number[]>`. Like when you do `export function animate<T>(options: Partial<AnimationOptions<T>> = {}) {`, you are allowing mixing and matching of `number` and `number[]` which is incorrect and also means that isMulti can't tell which type everything is.

ShaMan123:
Good point.
I noticed that too.
If you agree you an merge that branch here and we continue isMany here

Lazauya:
@ShaMan123 OK I tried another iteration. This one has even more code duplication than the last one, which is not good. I still can't figure out a way to make it type safe without code duplication with the current code. I think it would be really easy to avoid duplication if we change all the internals to use ONLY arrays instead of values or arrays. That way we can treat everything the same and we don't need branching. That's my only idea. Perhaps a smarter person can chime in.

src/util/animation/AnimationBase.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (this.state === 'aborted') {
+      return;
+    }
+    if (this._abort(value, this.valueRate, this.durationRate)) {
+      this.state = 'aborted';
+      this.unregister();
+    }
+    if (durationMs > boundDurationMs) {
+      // TODO this line seems redundant
+      this.durationRate = this.valueRate = 1;
```
ShaMan123, line 117: this line seems redundant

src/util/animation/ColorAnimation.ts
```diff
@@ -0,0 +1,65 @@
...
+      ) as TColorAlphaSource,
+      easing,
+      onChange: wrapColorCallback(onChange),
+      onComplete: wrapColorCallback(onComplete),
+      abort: wrapColorCallback(abort),
+    });
+    this.colorEasing = colorEasing;
+  }
+  protected calculate(currentTime: number) {
+    const changeRate = this.colorEasing(currentTime, this.duration);
```
ShaMan123, line 50: I think it makes sense moving the color easing logic to the default easing function
Then ColorAnimation is identical to the rest

src/util/animation/animate.ts
```diff
@@ -0,0 +1,64 @@
...
+ * });
+ *
+ */
+export const animate = (options: AnimationOptions | ArrayAnimationOptions) =>
+  (isArrayAnimation(options)
+    ? new ArrayAnimation(options)
+    : new Animation(options)
+  ).start();
+
+export const animateColor = (
```
ShaMan123, line 53: I want to change the signature to match `animate`

src/util/animation/types.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * An easing function
+ * @param currentTime ms elapsed
+ * @param startValue
+ * @param byValue
+ * @param duration in ms
+ * @returns next value
+ */
+export type TEasingFunction<T = unknown> = T extends any[]
```
ShaMan123, line 29: I wonder if there's a better option for this, optional `index` arg somehow

ShaMan123 (commented):
I have refactored animation logic into classes.
TS and classes work nicely together.

- I believe I fixed color animation (not tested)
- I would like to align color animation with its peers, changing the default easing function to include colorEasing and dropping that from the class as well as the signature of `animateColor`. It is breaking, not heavy though.

Please review, I plan on merging this soon.

ShaMan123 (dismissed):
IMO this work will promote #6390

src/util/animation/ColorAnimation.ts
ShaMan123: fd763b7a8

src/util/animation/animate.ts
ShaMan123: 139ee11f8

src/util/animation/AnimationBase.ts
ShaMan123: 41adf1248

src/util/animation/types.ts
ShaMan123: This might not be such a good idea
I thought it was good to add an index for color easing to know if r/g/b/a
But this opens up a hatch for the easing function to be non linear relative to its runs on other indexes. This will create unexpected % value

test/unit/animation.js
```diff
@@ -191,6 +191,23 @@
     }, 1000);
   });
 
+  QUnit.test('byValue takes precedence over endValue', function (assert) {
```
ShaMan123: This test will fail

ShaMan123: I removed it
but we should address this strange glitch

ShaMan123: wasn't able to do it with TS

ShaMan123: fixed

ShaMan123 (approved):
I am very happy with how this turned out
Updated description

ShaMan123 (dismissed):
- It is possible to make `animate` accept also a color animation options object 
```ts
animate({
  startValue: 'colorString' // TS will find this easy to handle
})
animate({
  startValue: [r,g,b,a] // TS will find this hard to handle since it is an array and will fall into that case
  type: 'color' // this will do the job
})
```
- What about making `AnimationBase` extend Observable and fire state events?
```ts
const context = animate({ ... })
context.on('change', () => {})

```

Lazauya:
Since this turned into a bigger overhaul, I slipped in a small change where I turned value/durationRate to value/durationRatio. It's better grammar.

src/util/animation/AnimationBase.ts
```diff
@@ -0,0 +1,153 @@
...
+  AnimationState,
+} from './types';
+
+export abstract class AnimationBase<T extends number | number[]> {
+  readonly startValue: T;
+  readonly byValue: T;
+  readonly duration: number;
+  readonly delay: number;
+  protected readonly easing: TEasingFunction<T>;
+  private readonly _onStart: VoidFunction;
```
Lazauya, line 21: Should we make these javascript private using `#`? like `readonly #onStart`?

src/util/animation/types.ts
```diff
@@ -0,0 +1,134 @@
...
+           * Difference between the start value(s) to the end value(s)
+           * Overrides `endValue`
+           * @default [endValue - startValue]
+           */
+          byValue: T;
+          endValue?: never;
+        }
+    );
+
+export type TAnimationOptions<T, C = T, E = T> = Partial<
```
Lazauya, line 122: I'm putting this here but it applies to the others as well: I think we should name the type parameters. I was using things like `State` or `Return`. I think stuff like this is just unnecessarily confusing for newer people when it's more complicated types where the meaning of `T` isn't so obvious. For something like an array or a map, `T` is OK, but it's not obvious the type of which property is being parameterized in a `TAnimationOperations`.  I know that this stuff is mostly internal, but I still think it's a good idea to use as obvious names as possible.

src/util/animation/types.ts
Lazauya: well since this has become almost a total overhaul, you could always make `valueRatio` a `number | number[]`

src/util/animation/AnimationBase.ts
ShaMan123: We haven't discussed the `#` private pattern.
So there is no fabric official answer.

I am against this pattern.
The `private` directive tells the dev they shouldn't access it and hides it but if they want to for some reason to override it they can. TS will complain but who cares.
Everyone understands that it might break everything but it is still an option.

I have wanted to do a very small change to something `#private` a couple of times that would have made my code/solution much simpler but couldn't. So I had tp revert to patching packages, building, changing scripts. A nightmare. And for fabric and doesn't give anything.

What are the advantages? Why should we?

src/util/animation/types.ts
ShaMan123: OK, good point.
I will try naming.

src/util/animation/types.ts
ShaMan123: I am not sure I understand

ShaMan123: I want the easing signature to run on an single value.
Changing that is a breaking change I don't want to do and IMO should be done.

ShaMan123: > This might not be such a good idea I thought it was good to add an index for color easing to know if r/g/b/a But this opens up a hatch for the easing function to be non linear relative to its runs on other indexes. This will create unexpected % value

Actually it is not that crucial.
True `valueRatio` might be inaccurate but it is used in the callbacks so the dev can use `durationRatio` which is linear. Problem solved sort of.

ShaMan123 (commented):
@Lazauya I need some help figuring out why `onChange` isn't inferred correctly

[TS Play](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgSQM4FEAeBDAxjAGjm1VQFNY4BfOAMyghDgHJcIA7AE2BmA+wA2AWhgBPMGSG4AFmVwBrVMwDcAWABQGsRLgAhEmQA8AFThlMMMl1Rx2AVxAAjCnAA+th86gBtALoA+OABeNw04cLgACgQwiLirTgA1QTsyAH4ALjhjAli48MdRZIFUzNsyADcKXPU4qlc8xEb4rmLSrPZK6uaIwrb0rJzGqgBKOAAyRpja-PCOAGFpbHYAcwGo7EGiRw7ParhcXacKMaDAiohgTmGNW-VaO3Z8PnZidlBsSxMzCwSbfXIhnsxyggXcAKMwK8fn8-kiEDAvA4qEGY2mcSgZBgdigr3RszMrRSZF2AgExBsj3k7AgAHdXiRsjd1FQ7ssPpZoo0Ev0sgBmGqjNTqdkgT5kLkzApFYlZABMgpGwoA9Mq4AABGCoITmCT4HVQBhQOAcASiE2dE20KIAAz6xJtbjgNp5DrGwBsggEdLI1xF7zFnPx4VdJRJcAFjXtYfliuFGjY7FQ8ETlgswTeHIlwbgyewsF5cAAjDUWklZRHSxEFktVhKeuFsEQG3BHFXZrhGmi4C2SORYIY0Fg8DBDNoyBBrU2PCDYZEYFBUkre6QKKOhzh8GPxBPrW2Z145wul8KCRT++uMJvR+PJwciFCKEfF2Rl1LqBpRndE8niIbsKI8wcGm8AhKK4qSnEeYFhW3gAKy+O2rYyjGcDeEWiGNDWyxrJEvZIXEbYtp275oiuF6DleI7bhId7To+PgBHCx6vqeBJ9mulHDlut57g+eyMc+J7kZxG7Ubx94HhQMLMS+b51J+b5AA)

kristbp2:
> @Lazauya I need some help figuring out why `onChange` isn't inferred correctly when passing array values
> 
> [TS Play](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgSQM4FEAeBDAxjAGjm1VQFNY4BfOAMyghDgHJcIA7AE2BmA+wA2AWhgBPMGSG4AFmVwBrVMwDcAWABQGsRLgAhEmQA8AFThlMMMl1Rx2AVxAAjCnAA+th86gBtALoA+OABeNw04cLgACgQwiLirTgA1QTsyAH4ALjhjAli48MdRZIFUzNsyADcKXPU4qlc8xEb4rmLSrPZK6uaIwrb0rJzGqgBKOAAyRpja-PCOAGFpbHYAcwGo7EGiRw7ParhcXacKMaDAiohgTmGNW-VaO3Z8PnZidlBsSxMzCwSbfXIhnsxyggXcAKMwK8fn8-kiEDAvA4qEGY2mcSgZBgdigr3RszMrRSZF2AgExBsj3k7AgAHdXiRsjd1FQ7ssPpZoo0Ev0sgBmGqjNTqdkgT5kLkzApFYlZABMgpGwoA9Mq4AABGCoITmCT4HVQBhQOAcASiE2dE20KIAAz6xJtbjgNp5DrGwBsggEdLI1xF7zFnPx4VdJRJcAFjXtYfliuFGjY7FQ8ETlgswTeHIlwbgyewsF5cAAjDUWklZRHSxEFktVhKeuFsEQG3BHFXZrhGmi4C2SORYIY0Fg8DBDNoyBBrU2PCDYZEYFBUkre6QKKOhzh8GPxBPrW2Z145wul8KCRT++uMJvR+PJwciFCKEfF2Rl1LqBpRndE8niIbsKI8wcGm8AhKK4qSnEeYFhW3gAKy+O2rYyjGcDeEWiGNDWyxrJEvZIXEbYtp275oiuF6DleI7bhId7To+PgBHCx6vqeBJ9mulHDlut57g+eyMc+J7kZxG7Ubx94HhQMLMS+b51J+b5AA)
> 
> Take a look at `test/ts/animation.ts`

OK so I'm pretty sure TS can only infer the type param `T` of `animate` based on the parameters `animate` is called with; it CANNOT infer the type param with a partially specified type. IDK how to "fix" it but personally I don't think it's a big deal to just specify everything.

Lazauya:
> @Lazauya I need some help figuring out why `onChange` isn't inferred correctly when passing array values
> 
> [TS Play](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgSQM4FEAeBDAxjAGjm1VQFNY4BfOAMyghDgHJcIA7AE2BmA+wA2AWhgBPMGSG4AFmVwBrVMwDcAWABQGsRLgAhEmQA8AFThlMMMl1Rx2AVxAAjCnAA+th86gBtALoA+OABeNw04cLgACgQwiLirTgA1QTsyAH4ALjhjAli48MdRZIFUzNsyADcKXPU4qlc8xEb4rmLSrPZK6uaIwrb0rJzGqgBKOAAyRpja-PCOAGFpbHYAcwGo7EGiRw7ParhcXacKMaDAiohgTmGNW-VaO3Z8PnZidlBsSxMzCwSbfXIhnsxyggXcAKMwK8fn8-kiEDAvA4qEGY2mcSgZBgdigr3RszMrRSZF2AgExBsj3k7AgAHdXiRsjd1FQ7ssPpZoo0Ev0sgBmGqjNTqdkgT5kLkzApFYlZABMgpGwoA9Mq4AABGCoITmCT4HVQBhQOAcASiE2dE20KIAAz6xJtbjgNp5DrGwBsggEdLI1xF7zFnPx4VdJRJcAFjXtYfliuFGjY7FQ8ETlgswTeHIlwbgyewsF5cAAjDUWklZRHSxEFktVhKeuFsEQG3BHFXZrhGmi4C2SORYIY0Fg8DBDNoyBBrU2PCDYZEYFBUkre6QKKOhzh8GPxBPrW2Z145wul8KCRT++uMJvR+PJwciFCKEfF2Rl1LqBpRndE8niIbsKI8wcGm8AhKK4qSnEeYFhW3gAKy+O2rYyjGcDeEWiGNDWyxrJEvZIXEbYtp275oiuF6DleI7bhId7To+PgBHCx6vqeBJ9mulHDlut57g+eyMc+J7kZxG7Ubx94HhQMLMS+b51J+b5AA)
> 
> Take a look at `test/ts/animation.ts`

OK so I'm pretty sure TS can only infer the type param `T` of `animate` based on the parameters `animate` is called with; it CANNOT infer the type param with a partially specified type. IDK how to "fix" it but personally I don't think it's a big deal to just specify everything.

ShaMan123:
READY to merge

ShaMan123 (dismissed):
DONE

src/util/animation/AnimationBase.ts
Lazauya: The advantages are very situational. I actually have benefitted from using `#` in the past. I was just wondering since it's a stronger form of private methods then the TS `private`.

Lazauya:
I want to add some more documentation, which I should get to by Friday if you guys don't mind waiting to merge.

ShaMan123:
You could always open another PR with the same branch

ShaMan123 (approved):
Updated from master
Typed `src/mixins/object_animation.mixin.ts` and `src/color/color.class.ts`
I have this idea creating a class to run an array of animations together.
That can be done later on.

src/color/color.class.ts
```diff
@@ -2,8 +2,14 @@ import { ColorNameMap } from './color_map';
...
 
+/**
+ * RGB format
+ */
 export type TColorSource = [number, number, number];
 
+/**
+ * RGBA format
+ */
 export type TColorAlphaSource = [number, number, number, number];
```
ShaMan123, line 13: we can rename these actually to `TRGBColorSource` and `TRGBAColorSource`

src/util/animation/AnimationBase.ts
```diff
@@ -17,42 +17,66 @@ const defaultAbort = () => false;
...
+   * @param startValue
+   * @param byValue
+   * @param duration
+   * @param delay
+   * @param easing
+   * @param onStart
+   * @param onChange
+   * @param onComplete
+   * @param abort
+   * @param target
```
ShaMan123, line 79: These `@param` directives are incorrect
You can do `@param {TAnimationBaseOptions} arg0` I think but I feell it is pointless. I don't mind

src/util/animation/AnimationRegistry.ts
```diff
@@ -22,7 +26,8 @@ class AnimationRegistry extends Array<AnimationBase> {
   }
 
   /**
-   * cancel all running animations attached to canvas at the next requestAnimFrame
+   * Cancel all running animations attached to a Canvas at the next requestAnimFrame
```
ShaMan123, line 29:
```suggestion
   * Cancel all running animations attached to a Canvas at the next frame
```
??

src/util/animation/AnimationRegistry.ts
```diff
@@ -38,7 +43,8 @@ class AnimationRegistry extends Array<AnimationBase> {
   }
 
   /**
-   * cancel all running animations for target at the next requestAnimFrame
+   * Cancel all running animations for target at the next requestAnimFrame
```
ShaMan123, line 46: same as prev

src/util/animation/AnimationBase.ts
```diff
@@ -17,42 +17,66 @@ const defaultAbort = () => false;
 export abstract class AnimationBase<
   T extends number | number[] = number | number[]
 > {
+  /**
+   * @see TAnimationBaseOptions
```
ShaMan123, line 21: why this?

src/util/animation/AnimationBase.ts
```diff
@@ -17,42 +17,66 @@ const defaultAbort = () => false;
...
+   */
   readonly startValue: T;
   readonly byValue: T;
   readonly endValue: T;
   readonly duration: number;
   readonly delay: number;
   protected readonly easing: TEasingFunction<T>;
+  /**
+   * @private
+   * @see TAnimationCallbacks
```
ShaMan123, line 31: same as prev

ShaMan123 (commented):
take a look @Lazauya

src/util/animation/AnimationBase.ts
Lazauya: What do you mean they are incorrect? The types are optional on javadoc param directives, and since we use typescript I don't see a point to include them

src/util/animation/AnimationBase.ts
Lazauya: It improves clarity IMO, but if you think it's too verbose it's not strictly necessary

src/util/animation/easing.ts
```diff
@@ -1,10 +1,16 @@
 /**
  * Easing functions
- * @see {@link http://gizma.com/easing/ Easing Equations by Robert Penner}
+ * See <a href="http://gizma.com/easing/">Easing Equations by Robert Penner</a>
```
ShaMan123, line 3: `@see {@link http://gizma.com/easing/ Easing Equations by Robert Penner}` is better

src/util/animation/AnimationBase.ts
ShaMan123: We haven't looked into the pipeline that will generate docs to the website yet so we don't know if we will need JSDOC or not.
Anyways there are cases we do document essentials in JSDOC.
This isn't the case.
The syntax is wrong.
```ts
* @param {MyType} options
* @param [number] options.delay
```

No need for here.

src/util/animation/AnimationBase.ts
ShaMan123: Is it visible anywhere for the dev accept here?
If so I don't see how helpful it is.
If the dev is here they can see the typedefs themselves and use VSCODE to navigate to definitions.

Lazauya: That's true, but I just like having extra pointers that make stuff more obvious. I'll keep it unless you specifically want it removed.

src/util/animation/AnimationBase.ts
Lazauya: I'm sorry but I'm still confused. Do I need to change something? My understanding is that the types for the param directives are optional, and since the code has the typings it shouldn't be necessary. Is this incorrect?

src/mixins/object_animation.mixin.ts
```diff
@@ -94,42 +108,50 @@ export abstract class AnimatableObject<
 
     const animationOptions = {
       target: this,
-      startValue: options.from ?? currentValue,
+      startValue:
+        options.startValue ??
+        // backward compat
+        (options as any).from ??
```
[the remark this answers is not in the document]
Lazauya: The comment says `// backward compat` so it looks intentional? But it's probably so old it's safe to remove now.

src/mixins/object_animation.mixin.ts
```diff
@@ -78,58 +88,67 @@ export abstract class AnimatableObject<
...
-    to = to.toString();
-    if (!propIsColor) {
-      if (~to.indexOf('=')) {
-        to = currentValue + parseFloat(to.replace('=', ''));
-      } else {
-        to = parseFloat(to);
-      }
+    if (!propIsColor && typeof to === 'string') {
+      to = to.includes('=')
+        ? currentValue + parseFloat(to.replace('=', ''))
```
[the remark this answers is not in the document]
Lazauya: My understanding is that this logic is fine. But it *really* needs to be documented better, because it took my way too long to understand how this worked. It's fine though because the `+=50` syntax is for animating a property (specifically not a color) `50` units (pixels, I think) in whatever relevant direction. The tests show stuff like `.animate('left', '+=50')` which is supposed to shift the object 50 pixels to the left. So add 50 to the current left, as the code says.

`byValue` is more more about how much a value should be incremented every tick if I understand correctly. In the `Animation` constructor, the default of `byValue` is `endValue - startValue` which would make our animation 1 tick.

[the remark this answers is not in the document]
Lazauya: Please double and check and correct me if I'm wrong, though. I wrote comments and stuff with that understanding.

src/util/animation/AnimationBase.ts
ShaMan123: So it is extra code without a gain
I prefer removed

src/util/animation/AnimationBase.ts
ShaMan123: > So it is extra code without a gain I prefer removed

this too

src/mixins/object_animation.mixin.ts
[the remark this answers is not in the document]
ShaMan123: > `byValue` is more more about how much a value should be incremented every tick if I understand correctly

not every tick, but for the whole animation. It doesn't make sense to increment by a fixed value each tick because the frames aren't guaranteed to flush in a fixed delta so you would get uncertain values.

Comments seem correct, but as I wrote we have `byValue` for that so I will remove this option once @asturur approves

src/util/animation/AnimationBase.ts
ShaMan123: I will push the fix

src/mixins/object_animation.mixin.ts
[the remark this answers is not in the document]
ShaMan123: I put that comment there

src/color/color.class.ts
```diff
@@ -223,20 +226,14 @@ export class Color {
...
-    const result = [],
-      alpha = this.getAlpha(),
+    const [r, g, b, alpha] = this.getSource(),
       otherAlpha = 0.5,
-      source = this.getSource(),
-      otherSource = otherColor.getSource();
-
-    for (let i = 0; i < 3; i++) {
-      result.push(
-        Math.round(source[i] * (1 - otherAlpha) + otherSource[i] * otherAlpha)
```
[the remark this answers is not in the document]
ShaMan123: We could expose an option for color interpolation, the `calculate` method is there for such subclassing.
But I want to finalize this PR.
So if you want to propose something fork this branch and start another PR, that's what I do.

src/util/animation/AnimationBase.ts
ShaMan123: actually I am ok with this.
I will push the fix including this

src/util/animation/AnimationBase.ts
ShaMan123: For me the important thing is the autocomplete/discovery docs VSCODE can offer.

https://user-images.githubusercontent.com/34343793/206888412-615a08bb-6a22-410c-83a6-afa254e5003e.mp4

Since it is able to extract comments from types seems to me it makes a lot of JSDOC redundant.
I prefer types being documented where possible instead of signatures, but there are signatures that must be typed for clarity as well. It is not strict, which is good.

We are still in search of the best practice so it remains to be resolved in the future.

ShaMan123: a89ce3592

src/color/color.class.ts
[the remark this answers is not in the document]
ShaMan123: I would change `ColorAnimation` to accept color string values, that way it can accept any color space.
Then with a color space option it becomes very powerful

src/color/color.class.ts
```diff
@@ -399,14 +391,11 @@ export class Color {
...
-   * @memberOf Color
+   * @deprecated use `new Color(source)` instead
    * @param {TColorSource | TColorAlphaSource} source
    * @return {Color}
    */
-  static fromSource(source: TColorSource | TColorAlphaSource): Color {
-    const oColor = new Color();
-    oColor.setSource(source);
-    return oColor;
+  static fromSource(source: TColorSource | TColorAlphaSource) {
```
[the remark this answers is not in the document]
ShaMan123: @Lazauya  I left this comment and didn't remove for @asturur to review.
This is how I communicate the PR for a review.

ShaMan123 (approved):
@asturur there are too many threads in this PR. All are pretty much stale accept https://github.com/fabricjs/fabric.js/pull/8297#pullrequestreview-1200467329 that I would like you to comment on.

Updated description, fixed merge conflict

ShaMan123 (approved):
updated from master
seems good to go

ShaMan123:
@asturur let's merge this one
Too long has it been waiting

asturur:
> @asturur let's merge this one Too long has it been waiting

We had a list of priorities i m going in order.
I have to re-catch up because i have no idea what is this anymore.
Re-reading now.

asturur:
Ok fixed merge conflicts that is step 1

asturur:
notes for me since this is a long review
- check the regex match in color. https://github.com/fabricjs/fabric.js/pull/8297/files#diff-52c2e42c3516c3b6ccf6a73c0fe7f280f5dc9500283b88f6a0555defb0eacf7dR291

src/util/animation/Animation.ts
```diff
@@ -0,0 +1,30 @@
+import { AnimationBase } from './AnimationBase';
+import { AnimationOptions } from './types';
+
+export class Animation extends AnimationBase<number> {
```
asturur, line 4: Can we call this ValueAnimation? If we have ArrayAnimation we need a name that is a sibling of ArrayAnimation.
`Animation` only isn't clear enough.

asturur:
this build
313352 Dec 29 22:31 fabric.min.js
master build
311733 Dec 29 22:37 fabric.min.js

1.6KB extra, not sure why, they usually reduce, but not an issue

src/util/animation/AnimationBase.ts
```diff
@@ -0,0 +1,167 @@
...
+        this.durationRatio
+      );
+      this._state = 'completed';
+      this._onComplete(endValue, this.valueRatio, this.durationRatio);
+      this.unregister();
+    } else {
+      this._onChange(value, this.valueRatio, this.durationRatio);
+      requestAnimFrame(this.tick);
+    }
+  };
```
asturur, line 153: There is clear indication on MDN that fat arrows shouldn't be used for class methods.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

Unless there is a strong reason around not following that advice, this should be a standard method and an explicit bind in the constructor, would work correctly from the prototype AND bound out of context.

asturur:
Will take me tomorrow to finish reading this.
I wish there was a little bit more explanation on the pr message.
Is it clear that the external interface isn't changed ( apart for the returned value ) but a bit more of text on AnimationBase and Animation, ArrayAnimation and ColorAnimation would have been nice.
You have to consider the order you get the files to review is not the execution order and you start reading having no idea what is going to happen.

src/util/animation/Animation.ts
ShaMan123: sure
makes sense

src/util/animation/AnimationBase.ts
ShaMan123: I read that post and wasn't convinced that was what it said...
But I don't mind binding

src/shapes/Object/AnimatableObject.ts
```diff
@@ -78,58 +94,70 @@ export abstract class AnimatableObject<
...
-    if (!propIsColor) {
-      if (~to.indexOf('=')) {
-        to = currentValue + parseFloat(to.replace('=', ''));
-      } else {
-        to = parseFloat(to);
-      }
+    if (!propIsColor && typeof to === 'string') {
+      // check for things like +=50
+      // which should animate so that the thing moves by 50 units in the positive direction
+      to = to.includes('=')
```
ShaMan123, line 109: can we dump this?

src/util/animation/Animation.ts
ShaMan123: done

src/util/animation/AnimationBase.ts
ShaMan123: done

ShaMan123:
> You have to consider the order you get the files to review is not the execution order and you start reading having no idea what is going to happen.

Point taken

asturur:
I think this PR is good, thanks both.
I want to simplify something, maybe i can do that following up with a different pr.
Example:
- object animatabale could not use util animate but just start the animation on its own
- get rid of all custom syntax `+=` syntax for animation.
- remove the FX_DURATION concepts and all the 4 premade effects and relegate them to a folder called examples.

I have noticed there are type tests, and i see they can't run for now. is that correct?

asturur (approved):

ShaMan123:
> object animatabale could not use util animate but just start the animation on its own

didn't get what you meant.

The rest I agree.

> I have noticed there are type tests, and i see they can't run for now. is that correct?

Correct. #8394 handles the test framework.

PR #8298, opened 2022-09-18
BREAKING: refactor object rendering

ShaMan123:
## Motivation

This PR refactors object rendering which was IMO extremely unclear due to it's cyclic nature. I have read it more than I would have liked and never did I manage to wrap my head around it.
It was also confusing to use. Setting `objectCaching` to true or false would have made a difference only in some cases.

## Description
The main change that is introduced is a concept change:
Caching and rendering isolation are separate things.
Caching is something the dev decides about for perf. 
Rendering isolation is something fabric takes care of so clipping can occur for example.

In addition to the refactoring and renaming I have exposed a rendering context that is passed down the rendering process. It is an excellent thing allowing lots of stuff including the mentioned `Achievements & Fixes`, better cache management/interrogation etc.  
This change will solve most of the needs Group presents

## Achievements & Fixes
- fix shadow casting of nested object on cache
- fix `absolutePositioned` clip path nested in clip path
- feat(): possibility to support shadow casting direction/origin by using the same logic that fixes shadow to add more rotation/translation to the cache

## Changes
- `shouldCache` returns a simple flag indicating if what the dev wants
- renamed `needsItsOwnCache` => `shouldRenderInIsolation` for accuracy
- renamed `_drawClipPath` => `drawClipPath`
- removed `drawClipPathOnCache` - duplicate logic
- merged `_removeCacheCanvas`, `isCacheDirty` into `prepareCache`
- removed `drawClipPathOnCanvas` from StaticCanvas
- Group does not cache by default and should not be cached unless forced to in extreme edge cases such as a non interactive group with 10K objects. In other words when you need to create an image out of objects.
- Group renders in isolation by default
- removed `isOnACache` and group logic for shadow checks
- deprecated  `_transformDone`: `Object#transform` now accepts a second arg `needFullTransform` instead of running checks
- propagate `dirty` flag from objects to group regardless of caching state
- `RenderingContext`: holds context related to the current rendered tree
- added rotation to cache for shadow (needs more work to solve deep levels of nesting)
- **BREAKING**: clip path is now aware of it's ancestors. Prior, clip paths weren't aware of their container or that they were clip paths (probably because of stateful blocking circular refs). This was more than awkward when calling `calcTransformMatrix`. The expected value is the complete matrix from the canvas/viewer perspective but it gave the matrix from the container's perspective. Now it is aligned with expected behavior.
- calling `calcTransformMatrix` on `absolutePositioned` object will return the correct matrix which is object's own matrix. I wonder if it should return the full matrix in case `skipGroup===false`

## Rendering Process
- `render`
- validate rendering context (adding an entry if needed)
- `prepareCache`
- `drawObject`
- `drawClipPath`: calls `render`

## To Do
- [ ] work on group - shadow/isolation
- [x] shadow
- [x] canvas provider
- [ ] document `absolutePositioned` in plane change


asturur:
@ShaMan123 i tried to be clear that i m not accepting any more refactors before 6.0
Too many changes i don't want to look at before a basic transition to ts is finished.
Then after that whichever refactor we want to do that adds performances or unblock features can be discussed.

ShaMan123:
You were clear
I had an idea I wanted to see through
Once I have a drive I operate on it. It is all muse and moyivatiin for me. That is how I do best. I don't do well otherwise.

ShaMan123 (commented):
The last commits seem to have solved cached shadow casting.
Regarding if we should allow the dev to catch a group that has shadow - I saw yes. Their consideration.
This fix opens up an interesting possibility to support shadow casting origin and direction. Sounds cool. 
A big step forward

ShaMan123:
And actually it means that caching bugs are now **SOLVED** if I am not mistaken

ShaMan123:
1clipPath#calcTransformMatrix`
It is a long time I have been thinking about the fact the calling `calcTransformMatrix` on a clip path is completely out line. Because it is not aware of it's container object it returns a relative transform matrix.
This is something that we should eradicate. It is **very breaking** because it is a subtle change in a specific use case instead of a wide change to all calls. That is why I suggest adding a warning for it.

ShaMan123:
This also means we can change `sendObjectToPlane` to `sendObjectToPlane(object, to?: TMat2D | TObject)`
If `to` is an object it will become the container of `object`

ShaMan123:
The last few days I have been thinking of what this PR unlocks. Interesting possibilites.

I am know on your side, @asturur, regarding `absolutePositioned`. 
Up till now it was impossible for the dev to achieve this effect because of the tree cut off at clip paths, there was no way the walk up. But this is not the case anymore.
If we fire `before:render` on object with the rendering context the dev can handle it with ease
```js
object.on('before:render', ({ctx}) =>
 ctx.transform(...calcPlaneChangeMatrix(object.group.calcTransformMatrix()))

```

As you mentioned here https://github.com/fabricjs/fabric.js/discussions/7874#discussioncomment-2538429 I can now see how things become unclear with it.
e.g. `calcTransformMatrix(true)` what should it return for an `absolutePositioned` clip path? What should `calcOwnMatrix` return? Should it return the own matrix or should it return the parent's inverse transform?? It depends on the use case so we can't abstract it properly.
What happens when group is widely adopted and people start asking for `absolutePositioned` objects under group... Same thing. I do think it is viable design but with this PR fabric doesn't need to handle the overhead

Let's dump it and suggest a migration path.

ShaMan123:
## Group caching enabled

https://user-images.githubusercontent.com/34343793/193415261-af453af1-8c3f-423c-902c-5f094db93d46.mp4

## Group caching disabled, fallback to isolation

https://user-images.githubusercontent.com/34343793/193415269-6d1f43ac-b73a-4698-97f3-c3ab8c3de3c8.mp4

asturur:
@ShaMan123 there is a lot of work here, but i want to make clear we can't make those kind of large changes in the near future. We are moving slowly, and our priority should be finish, stabilize and have a great website.

We should do a meeting hopefully soon in which we put on table what we want to change in the rendering logic ( i have the cache with setters, the masks, the canvas rotation, partial rendering ) and see how all those changes can fit together with your changes on the rendering logic.

But we can't do anything of this if we don't release 6.0 with types and a clear list of breaking changes

ShaMan123 (commented):
After giving more thought the `fork` method is cumbersome.
There is no need to fork the context since rendering is a sync process.
Instead I will pop the context once an object finishes rendering.
This is good if we want to have a global context for use cases such as `willDrawShadow`

src/util/CanvasProvider.ts
```diff
@@ -0,0 +1,51 @@
...
+  locked: boolean;
+  constructor() {
+    this.canvas = createCanvasElement();
+    this.locked = false;
+  }
+  lock = () => {
+    this.locked = true;
+  };
+  unlock = () => {
+    this.locked = true;
```
ShaMan123, line 17:
```suggestion
    this.locked = false;
```

ShaMan123:
Though this was a great effort I have found that the dirty rect rendering optimization along with a rendering loop and context isolation are all that an app needs to perform complex rendering with good performance results.
It seems wise to handle rendering of the tree at top level.

PR #8301, opened 2022-09-19
fix(`WebGLProbe`): regression `enableGLFiltering` config + init bug

ShaMan123:
This PR fixes 2 regression from #8199 
- `enableWebGLFiltering` see https://github.com/fabricjs/fabric.js/pull/8199/files#r973981355
- bug in negation

and adds tests


src/filters/WebGLProbe.ts
```diff
@@ -57,10 +58,10 @@ class WebGLProbe {
    * @returns config object if true
    */
   private queryWebGL() {
-    if (this.initialized || !fabric.isLikelyNode) {
+    if (this.initialized || fabric.isLikelyNode) {
```
asturur, line 61: mmmm when this changed?

asturur (approved):

asturur:
Add the changelog please!

src/filters/WebGLProbe.ts
ShaMan123, line 61: 😵
😲
regression found

ShaMan123 (commented):
fixed regressions from #8199 
- `enableWebGLFiltering`
- bug in negation
- added tests

## To Do
Once my ci PRs are in I can test the WebGLProbe with sinon
In a next PR we should make config changes restored after each test

src/filters/WebGLProbe.ts
ShaMan123: a typo

asturur:
what do you mean `Once my ci PRs are in I can test the WebGLProbe with sinon` ? We can test webGlProbe right now probably launching a browser with acceleration disabled.

I'm not familiar with sinon anymore, but i would avoid adding more testing frameworks right now. if i have to differentiate testing i want to put down all the choices and pick reasonably. ( node specific and browser specific )

ShaMan123:
sinon is good for both... it is meant for stubbing and spying after objects, classes, modules etc.
Without it I can't access the webGLProbe because it isn't and shouldn't be exported by fabric

ShaMan123:
I remember adding sinon to a ci pr

ShaMan123 (commented):
added changelog

asturur:
i used sinon a lot, i don't use it anymore, i m not sure i want to re-learn it.
Adding Sinon and committing to it is something that needs to be weighted against moving to jest for unit testing and something else for browser-needed testing.

> I remember adding sinon to a ci pr

Well is not in master yet.

asturur (approved):

PR #8302, opened 2022-09-19
ci(): Add a check for changes to changelog

asturur:
This pr adds an additional check to verify if the user add the correct changes to the changelog.
The reason for this is that we often forget to add it.
Better some extra conflict on PRs rather than an empty changelog, there are automated tools that could pull out a changelog out of commit messages, we have one, but i personally do not use commits in a meaningful way and i use them more as save points rather than concise small changes. And i don't rebase, because i often mess up and i have better things to do.
So i squash and i do write manually a meaningful changelog line


ShaMan123:
I prefer the same and use committing the same
So the changelog becomes purely manual or a dev that uses commit messages can use the automated tool and commit the output?

ShaMan123:
For future reference I would like you to write how to use the automated tool

asturur:
npm run changelog is what we have right now, but is based on master changes, so is not valid for PRs changelog.
Is just a tool to backfill the changelog when we forget to add it

ShaMan123:
updated contributing 695d2d2

asturur:
@ShaMan123 is there something you want to think about or can you review this?

.github/workflows/build.yml
```diff
@@ -53,3 +53,20 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           delete-old-comments: true
           update-comment: true
+  changelog:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Code Coverage Report Updater
```
ShaMan123, line 61: remove this line, copied by mistake

ShaMan123:
```suggestion

```

.github/workflows/build.yml
```diff
@@ -53,3 +53,20 @@ jobs:
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: 16.x
+      - name: Changelog check
+        uses: Zomzog/changelog-checker@v1.2.0
+        with:
+          fileName: CHANGELOG.md
+          noChangelogLabel: Missing update to CHANGELOG.md
```
ShaMan123, line 69: The naming of this option is confusing
From the docs:
> The check can be disabled for a PR with a label. This label can be customized with the `noChangelogLabel` parameter.
https://github.com/Zomzog/changelog-checker/blob/master/action.yml#L12

So no need for it
I agree they should have added an option to customize the title

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## [next]
 
+- ci(): Add a git hub action check for adding changes to CHANGELOG.md [#8302](https://github.com/fabricjs/fabric.js/pull/8302)
```
ShaMan123, line 7: typo `git hub`

ShaMan123 (changes requested):
see comments

ShaMan123:
I have another thing I want ci to do in this scope.
Once published it should modify the changelog by replacing the `next` tag with the published tag and add a new `next` section.
That will safeguard forgetting to do it before releasing, which will happen for sure.

ShaMan123:
If we are going in that direction why not have an action run on PR merge to master that writes to the changelog the PR commit message which is meaningful always instead of part of the PR?

.github/workflows/build.yml
asturur: mmmm but i have seen that going out in the console? let me check again

asturur: i definetely got confused with this:
image

i simply copied their example and smashed in, i didn't even search for a doc page

asturur: initially i tought this would be the message that goes on the github check section, when failed. Then i see it wasn't and moved on

asturur:
> If we are going in that direction why not have an action run on PR merge to master that writes to the changelog the PR commit message which is meaningful always instead of doing it as part of the PR?

I don't like automated commits. Is a changelog you have to add in the file, we should just do

> I have another thing I want ci to do in this scope. Once published it should modify the changelog by replacing the `next` tag with the published tag and add a new `next` section. That will safeguard forgetting to do it before releasing, which will happen for sure.

You can do it, this is more a daily issue.
But even if you forget to update the version in the changelog nothing happens.

ShaMan123:
came across https://github.com/marketplace/actions/automatic-releases

ShaMan123:
OK I understand the missing piece.
We should use versioning differently.
A release should end the life of a version and not start it.
It is confusing that there's a release for 5.1.0 and master is still on that tag.
We should increment the version **after** the release and not before. Thus, starting a new version.
Then the changelog could have the current tag as the header instead of `next`.
What do you think?
Well I do get that incrementing a major version is done hindsight...
What is the flow of versioning and releasing? Let's plan it and make CI take over.

ShaMan123:
I see that CodeQL is failing due to syntax
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-a-custom-configuration-file

asturur:
well since i applied all the requested changes i m using the admin power and merging.
You are talking about unrelated ideas to this PR

asturur:
> OK I understand the missing piece. We should use versioning differently. A release should end the life of a version and not start it. It is confusing that there's a release for 5.1.0 and master is still on that tag. We should increment the version **after** the release and not before. Thus, starting a new version. Then the changelog could have the current tag as the header instead of `next`. What do you think? Well I do get that incrementing a major version is done hindsight... What is the flow of versioning and releasing? Let's plan it and make CI take over.

The flow till now is that i would increment the version at the moment of tagging, because i didn't want devs to read a new version number that didn't exist yet. Also until is release time you don't know which release number you will get, unless you release every single commit that for me is an overkill unless we talk about bug fixes.

Fabric development has always been a gut feeling thing, today i fix this, today i add that, so one day you could have a patch or a minor, and always tried to keep at minimum the major releases

now changing development flow is entirely possible, i don't like to release each commit and i don't like to have many active branches either.
It is also fine to release with a regular cadence, immediately for bugfixes with no refactors, and on a regular cadence for the other things ( like once a month ).

asturur:
we can swap the master version to 5.99 now just so that we don't get confused

ShaMan123:
I found a bot that will comment if the changelog hasn't been updated and is more configurable and verbose
https://github.com/behaviorbot/update-docs

PR #8305, opened 2022-09-20
refactor(): group `toSVG` logic

ShaMan123:
This PR is a refactor to the svg export logic of group.
The motivation for the PR is that the svg export logic uses `calcTransformMatrix` in case of a group that is a clip path in order to obtain the transform of the underlying objects since `<g>` can't be inside a `<clipPath>`  (god know why and JSDOM agrees with me and supports it). In other words it multiplies the group transform with the object transform and sets that to the object.
Anyways since I plan to fix `calcTransformMatrix` so that clip paths are aware of the object tree and will return the entire transform including the object they clip this PR is a prerequisite.
I have learned a bit how to PR @asturur 😉

## Changes
- expose `toClipPathSVGDef` on object, overridden by group (couldn't think of a better name since `toClipPathSVG` is taken and I decided not to break)


ShaMan123 (commented):
READY

I know it is a refactor and we agreed not to do anymore.
So I leave it to you to merge it whenever.

ShaMan123:
I can be a semantic f*** and say that it promotes a fix to shadow under cache and absolutePositioned clip path under clip path so it should be merged 😏

asturur:
wait there is nothing to fix in calcTransformMatrix, you may want to change something, but is not broken and so it does not need a fix

asturur:
Regarding the changes, SVG output is a feature as a whole, the single pieces that build it are not reusable alone nor are documented or useful. So it is ok to change the the signature during a major release. I ll read the changes as a refactor per se, it doesn't mean i understand or i agree for the preequisiste story

test/unit/group.js
```diff
@@ -485,7 +485,7 @@
...
+    var expectedSVG = '<g transform=\"matrix(1 0 0 1 90 130)\" clip-path=\"url(#CLIPPATH_0)\"  >\n<clipPath id=\"CLIPPATH_0\" transform=\"matrix(1 0 0 1 90 130)\" >\n\t\t<rect transform=\"matrix(1 0 0 1 25 -25)\" x=\"-15\" y=\"-5\" rx=\"0\" ry=\"0\" width=\"30\" height=\"10\" />\n\t\t<rect transform=\"matrix(1 0 0 1 -35 10)\" x=\"-5\" y=\"-20\" rx=\"0\" ry=\"0\" width=\"10\" height=\"40\" />\n</clipPath>\n<g style=\"\"   >\n\t\t<g transform=\"matrix(1 0 0 1 25 -25)\"  >\n<rect style=\"stroke: none; stroke-width: 0; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(0,0,0); fill-rule: nonzero; opacity: 1;\"  x=\"-15\" y=\"-5\" rx=\"0\" ry=\"0\" width=\"30\" height=\"10\" />\n</g>\n\t\t<g transform=\"matrix(1 0 0 1 -35 10)\"  >\n<rect style=\"stroke: none; stroke-width: 0; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(0,0,0); fill-rule: nonzero; opacity: 1;\"  x=\"-5\" y=\"-20\" rx=\"0\" ry=\"0\" width=\"10\" height=\"40\" />\n</g>\n</g>\n</g>\n';
```
asturur, line 488: so i do remember that the reason why i added clipPath inside groups instead of in the defs tag, is becaue they were inheriting or not inheritng the group transform. this is adding a transform to an element that didn't have it before. How can it be the same thing? Do we have any visual export with a group that isn't on top left 0/0?

asturur:
> The **motivation** for the PR is that the svg export logic uses `calcTransformMatrix` in case of a group that is a clip path in order to obtain the transform of the underlying objects since `<g>` can't be inside a `<clipPath>` (god knows why and JSDOM agrees with me and supports it).

This is the spec


A [‘clipPath’](https://www.w3.org/TR/SVG11/masking.html#ClipPathElement) element can contain [‘path’](https://www.w3.org/TR/SVG11/paths.html#PathElement) elements, [‘text’](https://www.w3.org/TR/SVG11/text.html#TextElement) elements, [basic shapes](https://www.w3.org/TR/SVG11/shapes.html) (such as [‘circle’](https://www.w3.org/TR/SVG11/shapes.html#CircleElement)) or a [‘use’](https://www.w3.org/TR/SVG11/struct.html#UseElement) element. If a [‘use’](https://www.w3.org/TR/SVG11/struct.html#UseElement) element is a child of a [‘clipPath’](https://www.w3.org/TR/SVG11/masking.html#ClipPathElement) element, it must directly reference [‘path’](https://www.w3.org/TR/SVG11/paths.html#PathElement), [‘text’](https://www.w3.org/TR/SVG11/text.html#TextElement) or [basic shape](https://www.w3.org/TR/SVG11/shapes.html) elements. Indirect references are an error (see [Error processing](https://www.w3.org/TR/SVG11/implnote.html#ErrorProcessing)).

Is just a spec, and probably it was done because at the time of svg 1.1 it was easier to do so with rendereres.

ShaMan123:
What I what to fix in the next PR:
```js
clipPath.calcTransformMatrix() 
```

## Current Behavior
```js
clipPath.calcTransformMatrix() === clipPath.calcOwnMatrix()
```

## ExpectedBehavior
```js
clipPath.calcTransformMatrix() === multiplyMatrices(clippedObject.calcTransformMatrix(), clipPath.calcOwnMatrix())
```

That is why I refactored the SVG code not to use `calcTransformMatrix` because after changing this behavior it will be broken.
The SVG spec is not related. I was just unaware of it while working on this PR so I've added a comment.

ShaMan123:
What do you suggest regarding the naming of `toClipPathSVGDef`?
It is a method the returns the entire clip path markup.

asturur:
> The SVG spec is not related. I was just unaware of it while working on this PR so I've added a comment.

Was clear to me, i was just pointing out that the reason why a group can't enter a clipPath is becuase of specs, not because of some logic issue in the svg

If this PR improves any svg export, can we add a test for regressions that we can compare with the previous one?

asturur:
> What do you suggest regarding the naming of `toClipPathSVGDef`? It is a method the returns the entire clip path markup.

No issue with the naming, is ok.

asturur:
You can call it like `toClipPathSVGMarkup` if you prefer, since <defs> is an svg tag could be confusing to people who knows svg tags

ShaMan123:
will do

test/unit/group.js
ShaMan123: Isn't clippath-9?

ShaMan123 (commented):
- added visual tests with transformed group
- moved goldens to `clipping` folder and renamed `clippath-9` to `handbag`
- fixed visual tests not to create a missing golden when running from CI
- renamed `toClipPathSVGDef` => `toClipPathSVGMarkup`

ShaMan123:
I have also noticed that exporting an inverted clip path to SVG is not supported by fabric.
Moving to masks should do the trick.

ShaMan123 (commented):
If you want me to extract the CI fix to a different PR just say so.
I meant to but then part of the changes conflicted between the PRs.
So I can do it and we merge that before this.
Whatever you decide

asturur:
I will try to extract the test changes and commit them to master before the pr, so i can verify.
Adding tests only, and refactor options/positions later would have made it a bit easier, but i ll manage to do it.
I'll do it today

ShaMan123:
Ok
I see your point.
So we can split this pr into 3
1. Tests
2. Refactor svg logic
3. Fix visual ci

I can extract the ci logic tomorrow if you prefer.

The tests are identical on both files.
I really wanted to extract them and import them to both files intead of copying.

ShaMan123 (commented):
extracted the ci logic to #8311 

I want to create another PR enabling module import in testem and then we can use that here

ShaMan123:
couldn't manage modules in testem

ShaMan123:
I wonder what makes node18 visual tests fail

ShaMan123:
I tried reproducing node18 failing test in gitpod https://gitpod.io#snapshot/522e46b3-952b-4f4a-9143-2097cbd45d76

PR #8311, opened 2022-09-23
ci(): visual test suite adaptions

ShaMan123:
# Changes

- refactored `test/lib/visualTestLoop.js`: expose `assert.visualEqual` so we can use QUnit directly instead of calling the visual test loop
- tests with missing golden running from CI should fail c14e307c8b679304126aebcf520583a7a8f2938f
- `todo` tests should not recreate the golden
- refactored `test/lib/visualCallbackQunit.js` => `test/lib/appendTestResults.js`. Instead of holding the canvases in memory, the suite dumps them to disk and loads them back as images (exposed `dumpResults`). 
The dumped results are used by #8402 as well. 
They are dumped to disk **only** if a browser has been launched for visual tests **OR** CI is running visual tests (only failed tests are dumped).
This should enhance test perf.
- remove isChrome check in mustache file


ShaMan123 (commented):
READY

ShaMan123 (commented):
Updated from mater

READY

PR #8312, opened 2022-09-23
fix(`Object#_scaleTo(Width/Height)`): safeguard division by zero

big-nacho:
`Object`'s `scaledToWidth` and `scaledToHeight` allow for potential division by zero, which later on causes `NaN` to creep into object's coords.

[Demo](https://codesandbox.io/s/goofy-swanson-08y0c3)

### DIFF
--- a/src/mixins/object_geometry.mixin.ts
+++ b/src/mixins/object_geometry.mixin.ts
@@ -561,9 +561,14 @@ import { Point } from '../point.class';
        * @chainable
        */
       scaleToWidth: function (value, absolute) {
+        var scaledWidth = this.getScaledWidth();
+        if (!this.width || !scaledWidth) {
+          return this;
+        }
+
         // adjust to bounding rect factor so that rotated shapes would fit as well
         var boundingRectFactor =
-          this.getBoundingRect(absolute).width / this.getScaledWidth();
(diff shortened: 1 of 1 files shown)


src/mixins/object_geometry.mixin.ts
```diff
@@ -561,9 +561,14 @@ import { Point } from '../point.class';
...
+        var scaledWidth = this.getScaledWidth();
+        if (!this.width || !scaledWidth) {
+          return this;
+        }
+
         // adjust to bounding rect factor so that rotated shapes would fit as well
         var boundingRectFactor =
-          this.getBoundingRect(absolute).width / this.getScaledWidth();
+          this.getBoundingRect(absolute).width / scaledWidth;
         return this.scale(value / this.width / boundingRectFactor);
```
ShaMan123, line 572:
```suggestion
        return this.scale(ifNaN(value / this.width / boundingRectFactor, 0));
```

ShaMan123 (commented):
Checking if width === 0 covers also if scaledWidth === 0
But I prefer you use the internal util `ifNan`

ShaMan123 (changes requested):
So I would revert 245bd90

PR #8322, opened 2022-09-27
chore(TS) Converting Object to a class

asturur:
This PR propose a way to convert fabric obects to classes while maintaining flexibility on the prototype.
This PR also includes the work from @ShaMan123 around converting the observable in a class and is merged using his own commits.

# TLDR;

The class is declared as an es6 class, but the fields are just declared as types.

A separate object contains all the class default values and those get added on the prototype with an assignment.

The class is fully typed, and the default values are type checked using an utilty type function.

# Changes
You should skip all files but object.class.js, observable.class.js, and typdefs.ts

The changes in object.class.ts are unreadable, but what happened is this:

The class declaration is:

```js
class FabricObject {
  let: number;
  top: number;
  constructor(options) {
    this.setOptions(options);
  }
  initialize(options) {
    ... for compatibility reason, removed as soon as we changed everything to class
  }
  ... all the old methods, no properties...
}
```

The methods changed from:
```js
setOptions: function() { code },
```
to
```js
setOptions() { code }
```
Nothing else has been touched

All the properties are at the end of the file and are assigned like this:

```js
Object.assign(
  FabricObject.prototype,
  fabric.CommonMethods, 
  fabricObjectDefaultValues,
```

The observable has ben converted to a class plain and simple no strange changes of any kind.

I updated all the reference not because it was necessary but because i made an error in the constructor and nothing was working, and it took me a while to see the error.

# Type checking
This utility i found on stackoverflow removes from a type all parts that are functions.
This allow us to know from the class definition, by difference, which are the properties we expect in the defaul object values, and allow us to have typeChecking
```ts
// eslint-disable-next-line @typescript-eslint/ban-types
type TNonFunctionPropertyNames<T> = { [K in keyof T]: T[K] extends Function ? never : K }[keyof T];
export type TClassProperties<T> = Pick<T, TNonFunctionPropertyNames<T>>;
```

# Todos
- [ ] get agreement on this pattern -> NOT done, won't be done
- [x] resolve the type TDegree and number
- [x] add the optional typing on the class for all the items we add at runtime on the FabricObject and that are undocumented ( _cacheCanvas to make an example ).
- [ ] get the object.class.ts file type check and linted -> DONE partially
- [X] remeber how to do checklists in MD on github


src/mixins/observable.mixin.ts
```diff
@@ -142,4 +144,10 @@
     once: once,
     off: off,
   };
+  fabricHeader.Observable = {
+    fire: fire,
+    on: on,
+    once: once,
+    off: off,
+  }
```
asturur, line 152: this is some bullshit i did when i was half asleep.
I didn't even remember doing it.
Going to delete

src/gradient/gradient.class.ts
```diff
@@ -93,7 +90,7 @@ export class Gradient<
     gradientTransform,
     id,
   }: GradientOptions<T>) {
-    const uid = fabric.Object.__uid++;
+    const uid = FabricObject.__uid++;
```
asturur, line 93: this is not the place for the __uid. i remember @ShaMan123 already worked to put it somewhere else

src/mixins/observable.mixin.ts
ShaMan123: I saw that in #8103 and removed that there as well

src/parser/setStrokeFillOpacity.ts
```diff
@@ -19,10 +19,10 @@ export function setStrokeFillOpacity(attributes) {
     }
 
     if (typeof attributes[attr] === 'undefined') {
-      if (!fabric.Object.prototype[attr]) {
+      if (!FabricObject.prototype[attr]) {
         continue;
       }
-      attributes[attr] = fabric.Object.prototype[attr];
+      attributes[attr] = FabricObject.prototype[attr];
```
ShaMan123, line 25: I think this part can be removed in the next iteration.
Why would we need a default value if parsing will eventually init a new class that already has defaults and handles that internally.

src/util/lang_class.ts
```diff
@@ -74,12 +74,10 @@ export function createClass(...args) {
   }
 
   klass.superclass = parent;
-  klass.subclasses = [];
 
   if (parent) {
     Subclass.prototype = parent.prototype;
     klass.prototype = new Subclass();
-    parent.subclasses.push(klass);
```
ShaMan123, line 82: I am guessing this makes a lot of tests fail
Textbox etc.

ShaMan123 (commented):
As mentioned previously `cacheProperties` and `stateProperties` should be static IMO. Is there any need for them on the proto?
All calls to `FabricObject.prototype[method]` should be replaced with `super[method]`

Where do you assign a prototype value to the class?

src/mixins/observable.mixin.ts
asturur: seems like i added it here?

asturur:
> As mentioned previously `cacheProperties` and `stateProperties` should be static IMO. Is there any need for them on the proto? All calls to `FabricObject.prototype[method]` should be replaced with `super[method]`
> 
> Where do you assign a prototype value to the class?

Maybe and yes.
But those are not the kind of comments/changes we are looking at now.

Let's look a the general approach.

To solve the first issue, later i should try to add here that kind of mixin function you shown me yesterday so that it can take in account also properties where possible.

`cacheProperties` and `stateProperties` are good to be accessed with `this` once they are accessed with this, each class can change its own per class. Otherwise you have to refer to Rect.cacheProperties and Circle.cacheProperties? how do you handle the `i look at the nearest one in the prototype chain` if are static methods?

src/util/lang_class.ts
asturur: i din't see the word `subclasses` used in the codebase outside this line, so i tried to get it away. I ll look better

ShaMan123:
- There is no static inheritance AFAIK.
- I would argue that anything shared (such as cache/state props) should be static and anything at the instance level shouldn't be shared as a strict design rule. I understand that referencing `this` is more convenient but if you're using `this`, you've imported the class so it doesn't make that much of a difference.

src/mixins/observable.mixin.ts
ShaMan123: The diff is too harsh.

src/parser/setStrokeFillOpacity.ts
ShaMan123: good point

src/util/lang_class.ts
ShaMan123: I think (not sure) that callSuper uses something of that logic

asturur:
the fact is that teh cache properties may be definied per class, but accessed by the base class.

src/util/lang_class.ts
Lazauya: @ShaMan123 callSuper uses superclasses

asturur: i think it was unused

src/shapes/object.class.ts
```diff
@@ -1,2049 +1,2619 @@
...
+ * @fires mousewheel
+ * @fires mousedblclick
+ *
+ * @fires dragover
+ * @fires dragenter
+ * @fires dragleave
+ * @fires drop
+ */
+
+const fabricObjectDefaultValues = {
```
Lazauya, line 59: maybe give this object a type? like `const fabricObjectDefaultValues: FabricObject` or `const fabricObjectDefaultValues: Partial<FabricObject>`

asturur: Yes, that is where i m stuck now.
Type verification of this object.
What i did was copy pasting a utitlity from stack overflow that is this one:

https://github.com/fabricjs/fabric.js/pull/8322/commits/f05ee9cfa64fc1a255afd81845af5dfed7b257b7#diff-a2ca4295c4b9b0a677c07db27810483a1abf0d671387d14ab4258f6bdfd72cfaR21-R22

And i m trying to extract the types from the class declaration and enforce the default values to respect that.
But isn't really working for now.

asturur: oh well :)

image

it was working, is just that i put back the ts-nocheck comment to avoid too many errors.

src/shapes/object.class.ts
```diff
@@ -1,2049 +1,2023 @@
...
+   * @memberOf fabric.Object
+   * @type Number
+   */
+  static __uid = 0;
+
+  /**
+   * Constructor
+   * @param {Object} [options] Options object
+   */
+  constructor(options: FabricObjectDefaultValues & Record<string, unknown>) {
```
ShaMan123, line 651:
```suggestion
  constructor(options: typeof FabricObjectDefaultValues & Record<string, unknown>) {
```

ShaMan123 (commented):
**I will first state my thoughts and then suggest a solution.**

### Thoughts and bias
I think this PR is an anti pattern to modern classes and it doesn't comply with the [guidelines](https://github.com/fabricjs/fabric.js/issues/7596#issuecomment-1201570973) I suggested for a solution.
In other words I don't like it at all.
And it seems to me we should discuss why this is needed in the first place. We didn't discuss it, just moved on and said it is a must have. I disagree. 
We can list all occurrences and discuss other solutions.
I know this must be irritating... but I really don't understand for what reason we are doing this workaround.

---

### Moving forward
I suggest the following:
1. declare an interface with all the options an object accepts (`TOptions`), JSDOC should be added here only.
2. extend `TOptions` with more esoteric properties like cacheProps etc. (`TExtendedOptions`)
3. use class implements:
```ts
class FabricObject extends Observable implements TExtendedOptions {
  ...
}
```
TS will fill out all the class props by itself using `quick fix: implement interface....` 
image

image

4. Cast the defaultProps object as `TOptions` or `TextendedOptions`
5. Subclass should extend the interface and implement it as described in step 3

This makes sense for developing flow and TS.
Those types that pick stuff (picking class properties)... there's always something that should fit and doesn't and vice versa.

- We should avoid using assign for TS if possible, if not I guess we can cast the prototype with `TExtendedOptions`
- Why do we need to set the proto if we have a defaults object? Why not use that instead?

asturur:
> **I will first state my thoughts and then suggest a solution.**
> 
> ### Thoughts and bias
> I think this PR is an anti pattern to modern classes and it doesn't comply with the [guidelines](https://github.com/fabricjs/fabric.js/issues/7596#issuecomment-1201570973) I suggested for a solution. In other words I don't like it at all. And it seems to me we should discuss why this is needed in the first place. We didn't discuss it, just moved on and said it is a must have. I disagree. We can list all occurrences and discuss other solutions. I know this must be irritating... but I really don't understand for what reason we are doing this workaround.

I m sure we talked about it, it can be that you don't consdier that a reason.
The reason is that i want to still be able to change and detect default values of a class at runtime.
I'm not ok with giving up features or flexibility because TS makes more verbose to describe an object.
- we are not adding extra code
- we are not removing a feature
- we are obtaining embedded types and type safety

Class public fields behave differently from what fabricJS was doing till now.

So this is the reason ( discussed also in the migration thread )

Let's move on step by step. Which is the code in this PR that determine an anti pattern and which is the anti pattern, which are the side effects

asturur:
(we can't use implements because implements forces you to actuallt implement the fields in the class, and not doing so is the point of this PR )

src/shapes/object.class.ts
asturur: yes this is wrong for now, and will stay an any, for now till we can clearly separate options from internal values that are optionals and **should not be used as options** ( for example _cacheCanvas is a valid value but absolutely not an option )

ShaMan123:
```ts
interface TOptions {
    foo: number
}

interface TExtendedOptions extends TOptions {
    bar: string
}

class Parent {
    x(){
        return true
    }
}

class Impl extends Parent implements TExtendedOptions {
bar: string
foo: number

}
```



```ts
"use strict";
class Parent {
    x() {
        return true;
    }
}
class Impl extends Parent {
}

```


</details>


```json
{
  "compilerOptions": {
    "strictPropertyInitialization": true
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?strictPropertyInitialization=false#code/JYOwLgpgTgZghgYwgAgCoHkAOZgHsQDOyA3gLABQyVyMuuAXMiAK4C2ARtBQL4UWiRYiFKgCiAD0ggAJhGlYc+IhEkQZRDNjyESFasnZwojAmCigA5jz7kEAGzgEiABSNqwuytXEAKAJRkXvpUUBBgzFAgyGbMEHrUvOSJFPaORACSrJh2yCpS0i5u4MjAWXYQrO4aEvlyCtpEgYbGyKbmIFbktAxMbJxQNtxAA)

ShaMan123:
Where do you set the default values on the instance?
Shouldn't it be??:
```ts
constructor(options) {
this.setOptions({...defaults,...options})
}

```

ShaMan123:
I think that only options with defaults should belong to the defaults type/object.
The rest such as `_cacheCanvas` should remain as class fields, probably private fields...

ShaMan123:
**After mutating**



```ts


class Parent {
    x(){
        return true
    }

    static getDefaults(){
        return {
            foo: true
        }
    }
}

class Impl extends Parent {
  bar?: string
  foo: number

  constructor(options: Partial<ReturnType<typeof Impl['getDefaults']>>){
   super();
   const values={...options,...Impl.getDefaults()};
   this.foo=values.foo;
  }

  static getDefaults(){
        return {
            foo: 1,
            bar: 'baz'
        }
    }
}

Impl.getDefaults=()=>({
      foo: 5,
      bar: 'baz'
})

console.log(new Impl({bar:'pip'}));
```



```ts
"use strict";
class Parent {
    x() {
        return true;
    }
    static getDefaults() {
        return {
            foo: true
        };
    }
}
class Impl extends Parent {
    constructor(options) {
        super();
        const values = Object.assign(Object.assign({}, options), Impl.getDefaults());
        this.foo = values.foo;
    }
    static getDefaults() {
        return {
            foo: 1,
            bar: 'baz'
        };
    }
}
Impl.getDefaults = () => ({
    foo: 5,
    bar: 'baz'
});
console.log(new Impl({ bar: 'pip' }));

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/LAKFGMBsEMGdYAQAVoCcCmA7ALgg3qAkQgB4AUAlASMbQhtgK6qYLaqPqHEC+o3RWNmjYAluAQBzdNgAi6AGbRGkbLErU6xBs1aatdBQHsjALjYcuNA32tFbtiDHgIAkgFsADpAToS2LAATRBQMHHxuACM0AH5zIVRRTEluYzMETEZ3SPRUfmtwI0wExnBsI1QyI08xIthzULFoSAAeACUZXQAVAE9PdBbsPvQjBTcvSABtAHJpOUVlVVhpgF0APjWqAVhGfsqKAG4BQuLcADdmzlgAXjwAOgfq2uKAGge7j287ufklFTVKDwjnZsAALUSwO5pa4XSBXKEmYH2fKCYRiCQ-Bb-dRbOx0HQsCJ4gwINLmACMLwEJIQ0VQ5mm0QAXtNqbRbOzQI4QJ9IN8ZL9FmprpRrmsyPpaGSEABWKnE2loBnM1kgHgUfInWBGSDoO6QIySMiYdAAd3G3gldNM008ok803VhyAA)

ShaMan123:
This approach is much more agreeable IMO

asturur:
wait too much stuff.
I want to understan the anti pattern first because otherwise i don't know how to change in something that is better for you.

asturur:
> Where do you set the default values on the instance? Shouldn't it be the following?
> 
> ```ts
> constructor(options) {
>    this.setOptions({ ...defaults, ...options })
> }
> ```

no assigning them in the constructor would make every instance have a copy of the default value as an own property.
is exactly what the public field do and what i want to avoid

ShaMan123:
> no assigning them in the constructor would make every instance have a copy of the default value as an own property.
is exactly what the public field do and what i want to avoid

This is an anti pattern, is it not?
What are the advantages? Why do you want/need it?

asturur:
> > no assigning them in the constructor would make every instance have a copy of the default value as an own property.
> > is exactly what the public field do and what i want to avoid
> 
> This is an anti pattern, is it not? What are the advantages? Why do you want/need it?

Again i want to be able, as today, to change default values at runtime.
assigning all defaults in the constructor ( aka public fields ) breaks it.

ShaMan123:
`_set` is another anti pattern since it runs in the constructor and executes side effects so if you subclass `_set` will run in `super()` calls and try to execute side effects on uninitialized class properties

ShaMan123:
You are talking about reactivity. Like [mobx](https://mobx.js.org/README.html). Fabric should not do that.

ShaMan123:
In other words, to make myself clear, a shifting/shared class state is an anti pattern

asturur:
Fabric won't use it in the codebase.
The shared state is still there, the class prototype with the functions to which you can do what you want, with or without our change.
Allowing it also for default values  on top of functions is just less confusing

asturur:
> `_set` is another anti pattern since it runs in the constructor and executes side effects so if you subclass `_set` will run in `super()` calls and try to execute side effects on uninitialized class properties

but set is not giving any functionality that can't be obtained with similar code.
And that can be done without having all developers need to adapt.

Also the fact that yes, properties needs to be initialized and do not exist when you initialize the class in the super(), is very annoying and another point on the side of not using public fields and constructor initialization.

ShaMan123:
I will reply in our chat!

asturur:
# UPDATE

After some chit chat around good patterns and not good patterns, while perfectly in the realm of opinions, we decided to give a go to @ShaMan123 idea of not adding items to the prototype.

This won't be done in this PR.

The list of work items is:
- have this PR converting object into class with TS complaining not that much ( but complaining )
- have a rendevouz over object geometry mixin and move there everything that is a dependency so that it can be its own class, have object extend the geometry part ( so that the common base for any object is Geometry + observable + common methods and that is not configurable )
- move away from object SVG specific methods ( just 2 of them in the svg parser logic )
- convert the other shapes to classes in this way ( are all easy, while Text is another big chunk of work )

then have a PR ( or more than one ) that move the defaults out of the prototype so that breaking changes and diffs are captured and isolated there.

src/shapes/object.class.ts
```diff
@@ -1,2049 +1,2132 @@
...
+
+  /**
+   * Quick access for the _cacheCanvas rendering context
+   * This is part of the objectCaching feature
+   * since 1.7.0
+   * @type boolean
+   * @default undefined
+   * @private
+   */
+  _cacheContext: CanvasRenderingContext2D | null = null;
```
asturur, line 655: this is a public field, this thing is either null or not and does not need to be configured ever

src/shapes/object.class.ts
```diff
@@ -1,2049 +1,2132 @@
...
+   */
+  absolutePositioned: boolean;
+
+  /**
+   * A Reference of the Canvas where the object is actually added
+   * @type StaticCanvas | Canvas;
+   * @default undefined
+   * @private
+   */
+  canvas?: StaticCanvas | Canvas;
```
asturur, line 645: all the types below this one are new and added reading the code of object.class

src/shapes/object.class.ts
ShaMan123: Is public for what reason
Context is needed

ShaMan123 (dismissed):
Looks good

- `_render` is a good candidate to become abstract.
- Shared methods have a number of functions that all come down to set. We can use signature overload. I guess you want to avoid that for now.

src/shapes/object.class.ts
```diff
@@ -1,2049 +1,2136 @@
...
 import { runningAnimations } from '../util/animation_registry';
 import { enlivenObjectEnlivables } from '../util/misc/objectEnlive';
+import { clone } from '../util/lang_object';
+import { toFixed } from '../util/misc/toFixed';
+import { capitalize } from '../util/lang_string';
+import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
+import { createCanvasElement } from '../util/misc/dom';
+
+type StaticCanvas = any;
+type Canvas = any;
```
ShaMan123, line 20: Use the Canvas type from `__types__`?

src/shapes/object.class.ts
```diff
@@ -1,2049 +1,2136 @@
...
+
+const ALIASING_LIMIT = 2;
```
ShaMan123, line 22: extract to config? or constants?

ShaMan123 (approved):
Forgot these

src/shapes/object.class.ts
asturur: aliasing limit is not configurable, is just  an extra pixel left and right to avoid cutting away a line drawn on the edge of the cache.
If we round wrong, is going to give us the missing pixel. And shouldn't be more than one pixel per side ever

src/shapes/object.class.ts
asturur: as soon as Canvas is converted, we can trash those all away

src/shapes/object.class.ts
asturur: i mean is not public, it can be private. But it does need to be in the prototype ever

Issue #8331, opened 2022-09-30
fabric.util.animate  custom  easing ,  onComplete  Incorrect return value 


3400442579:
```
  let duration = 100;
  let endValue = 50;
  let startValue = 0;

function easingBack(t, b, c, d) {
            let a = t / d;
            if (a > 1 / 2)
                a = 1 - a;
            return c * a * 2 + b;
        }

      fabric.util.animate({
            startValue: startValue,
            endValue: endValue,
            duration: duration,
            easing: **easingBack**,
            onComplete: function (v1, t, s) {
                console.info("onComplete：   v=" + v1);
            }
        })

        let v2 = easingBack(duration, 0, endValue - startValue, duration)
        console.info("easingBack：  v=" + v2);

```
 run time == duration (100)  
 v1=0
 v2=50

 **v1 != v2** 

![image](https://user-images.githubusercontent.com/34226979/193245210-6922c200-6acf-440f-9116-759891dd2d34.png)

ShaMan123:
It seems to your easing function is wrong
It should complete at end value but it doesn't

ShaMan123:
looks like you are trying to achieve something like interpolation

PR #8333, opened 2022-09-30
easing  custom onComplete bug

3400442579:
[#8331](https://github.com/fabricjs/fabric.js/issues/8331)


ShaMan123:
see my comments on #8331

PR #8335, opened 2022-10-01
fix(): clear `_target` on discard

ShaMan123:
Handles the case in which an object is being moved by the user and discarded meanwhile.

Currently the object will continue moving and will be inactive only on mouseup.

Exposes `setTransformEventTarget` to hijack the event from the target

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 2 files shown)


ShaMan123:
closing in favor of the proposed solution at https://github.com/ShaMan123/fabric.js/pull/2

PR #8337, opened 2022-10-02
docs(): `qrDecompose`

ShaMan123:
@luizzappa here is my tryout to understand what happens in `qrDecompose` and why, all guesses.
@asturur shared a link somewhere on QR decomposition/factorization, I'll find that and comment it here.

https://www.youtube.com/watch?v=51MRHjKSbtk

QR decomposition on ml-matrix:
https://github.com/mljs/matrix/blob/f198bebeca2d0d5719313058f805149175f748fc/src/dc/qr.js

### DIFF
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -101,6 +101,8 @@ export const qrDecompose = ([a, b, c, d, e, f]: TMat2D): Required<
   // | a  c  e |
   // | b  d  f |
 
+  // assuming v1 is the first vector of our base then:
+
   // rotation of v1 relative to (1, 0)
   const rotation = Math.atan2(b, a),
     denom = Math.pow(a, 2) + Math.pow(b, 2),
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -111,9 +111,10 @@ export const qrDecompose = ([a, b, c, d, e, f]: TMat2D): Required<
     // || v1 || = the size of (a, b)
     scaleX = Math.sqrt(denom),
     // the area scale factor of A equals to |det(A)|
-    // scaleY is the area scale factor of A normalized by || v1 ||
+    // scaleY is the area scale factor of A (keeping its sign), normalized by || v1 ||
     scaleY = det / scaleX,
-    // skewX is the scalar factoring v1 into the projection of v2 onto v1
+    // skewX is the scalar such that:
+    // skewX * v1 = the projection of v2 onto v1
     // see https://www.khanacademy.org/math/linear-algebra/alternate-bases/orthonormal-basis/v/linear-algebra-the-gram-schmidt-process
     skewX = Math.atan2(dot, denom);
   return {
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -81,7 +81,7 @@ export const multiplyTransformMatrices = (
   ] as TMat2D;
 
 /**
- * Decomposes standard 2x2 affine matrix into transform components
+ * Decomposes a 2x2 affine matrix into transform components
  *
  * @static
  * @memberOf fabric.util
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -97,6 +97,9 @@ export const qrDecompose = ([a, b, c, d, e, f]: TMat2D): Required<
   // let A be the 2x2 transform matrix (v1, v2):
   // | a  c |
   // | b  d |
+  // let A' be the affine matrix of A:
+  // | a  c  e |
+  // | b  d  f |
 
   // rotation of v1 relative to (1, 0)
   const rotation = Math.atan2(b, a),
--- a/src/util/misc/matrix.ts
+++ b/src/util/misc/matrix.ts
@@ -81,28 +81,46 @@ export const multiplyTransformMatrices = (
   ] as TMat2D;
 
 /**
- * Decomposes standard 2x3 matrix into transform components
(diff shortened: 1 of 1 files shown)


src/util/misc/matrix.ts
```diff
@@ -81,28 +81,52 @@ export const multiplyTransformMatrices = (
   ] as TMat2D;
 
 /**
- * Decomposes standard 2x3 matrix into transform components
+ * Decomposes a 2x2 affine matrix into transform components
```
asturur, line 84: it is 2x3, why 2x2?

asturur:
maybe we should add a clarification to this documentation.
This decomposition works because it plays nicely on how fabricJS renders.

FabricJS has top,left ( translate ) and skew, scale, rotation.

fabricjs first translate the canvas in the object center in the center as first step of the canvas transformation.
( so [1 0 0 1 left top ] )
then at that point it scales and rotate on the center
then it apply skewing

if we would have done things differently ( like skewing first, then rotating, then apply top and left ) this process probably wouldn't have been possible.

this is the first step of the video:
image

if we would have skewed and then rotate, we would have needed those matrix with swapped position and this process wouldn't have been helpful at all.

luizzappa:
I think now I'm more grounded to comment... I'll take the opportunity to discuss a proposal for a new implementation to decode an matrix.

# Comments about qrDecompose

First some comments about `qrDecompose` (sorry if I'm stating the obvious, just so you can follow the reasoning and point out any flaws). 

As I understand it, using QR decomposition is merely because some mathematical properties fit the way fabricjs works. The "problem" is that regardless of whether we use qr decomposition, or any other strategy, we will be in a situation with more variables than equations available, which makes it impossible to discover all the variables (in the case of qrDecompose it assumes skewY=0 to be able to find the rest).

# What's going on behind

The system we are trying to solve becomes more visible if we multiply the matrices until we find the final transformation matrix:

image

Basically what we have then are the values ​​of the transformation matrix `A[0], A[1], A[2] and A[3]`, and the variables `a, b, c, d and theta (angle of rotation)`. What is impossible to solve (there are multiple solutions), we have 5 unknown variables and 4 equations.

Just to be clearer, an example. An object with the following options results in the following transformation matrix:

image

That is, we are conditioned to solve this system to find a, b, c, d, and theta.

image

In order to reach a solution, the qr decomposition considers b = 0 (skewY=0) and finds the other values of this system. But that doesn't always correspond to reality.

# Proposal

I don't see any other way out than to provide one of the variables (`a, b, c, d or theta`) so that we can find the others, that way we can correctly go back to the original `skew`/`scale` values ​​without incorrect assumptions.

My proposal would be create a `decodeTransformMatrix` function that would receive the transformation matrix and the angle, and would return the correct options for the object. The angle would be an optional parameter, if it is not provided, we use `qrDecompose` (it will be the "best guess" for the lack of information provided). 

Mathematically implementing this is quite easy, just solve the system of equations. My question is whether this format would be suitable for the internals of fabricjs. This function would have to be called recursively, for example if there is a group nested inside another group, and both with rotation, I would need to come down the chain and compute the values ​​for each angle. I believe the `calcTransformMatrix` function should do something similar, no?

@ShaMan123 @asturur do you think a solution in this way is viable?

ShaMan123:
`calcTransformMatrix` is recursive, no problem implementing the same way. A very deep tree is 10 levels deep, so the recursion is very neglegant.

The proposition is interesting.
In skewing controls we know rotation so that can be suitable.
I wonder about QR decomposition. The Q matrix is a rotation matrix and if I get it it is the rotation to apply to the space in order for the 2 column vector of R to orthonormal, I have no idea what that actually means, I can't visualize the 2 column vectors in fabric context. What does the the theta value represent in our case?

luizzappa:
> `calcTransformMatrix` is recursive, no problem implementing the same way. A very deep tree is 10 levels deep, so the recursion is very negligent.
> 
> The proposition is interesting. In skewing controls we know rotation so that can be suitable.

In fact, this function could take any of the variables (`skewX, skewY, scaleX, scaleY or angle`) and supply the others. I proposed `angle` just to simplify, but if you find it interesting to make it flexible we can extend it to any of these variables later. I'm going to set up a PoC just getting the `angle` and then we'll think about it.

> I wonder about QR decomposition. The Q matrix is a rotation matrix and if I get it it is the rotation to apply to the space in order for the 2 column vectors of R to be orthonormal, I have no idea what that actually means, I can't visualize the 2 column vectors in fabric context. 

The matrix Q is a matrix of orthonormal vectors. I don't think they make R orthonormal... It's just a factorization into two matrices, one with orthonormal (Q) vectors and another connecting matrix (R). (Here's a more detailed lesson https://www.youtube.com/watch?v=0MtwqhIwdrI . By the way, his linear algebra course is very good, I have already attended some classes, it is open: https://ocw.mit.edu/courses/18-06-linear-algebra-spring-2010/)
Since every unit vector in R2 can be written in terms of sin and cos, and the rotation matrix is ​​orthonormal, Q happened to be the rotation matrix. 

R represents our transformation matrix by purging the effect of rotation.

Our transformation matrix with all of transformation together separating just the rotation:

image

On the other hand, doing the QR factorization of the transformation matrix:

image

Equating the two equations:

image

The elements of the matrix R are the ones calculated according to the video. Solving this, we come to the options.

> What does the theta value of Q represent in our case?

The theta is the angle of the object.

luizzappa:
I opened a PR https://github.com/fabricjs/fabric.js/pull/8425. I had said that we could use any variable (`angle, skew or scale`) to find the other options, but I changed my mind. Because of the `flipX` and `flipY`, only knowing the `angle` can we distinguish the `flip`.

PR #8344, opened 2022-10-03
fix(polyline): stroke bounding rect calculation typescript

luizzappa:
I migrated the PR https://github.com/fabricjs/fabric.js/pull/8083/ code to typescript and corrected the calculation when there is skew. 

closes #8083

# Related Issues
- https://github.com/fabricjs/fabric.js/issues/8025
- https://github.com/fabricjs/fabric.js/issues/7847
- #7478

# Proposed changes
- In `calcAngleBetweenVectors` function, calculate angle between two vector using atan2 instead of dot product. So eliminated the need to check the angle direction in `getBisector` function, 
- Changed the `_setPositionDimensions` function in the polyline class to adjust the width/height by the scale/skew.
- Changed the `projectStrokeOnPoints` function to cover all stroke line join cases and also to uniform stroke. 

# Working example

https://codesandbox.io/s/project-stroke-points-with-skew-correction-yz7c77

image

The red dotted vector is the bisector returned by the getBisector function. The green dotted vector is in the opposite direction and is what is used to determine which side to plot the projections. The projections are the colored vectors, one color for each point.
To make debugging easier, you can configure the parameters (stroke line join type, strokeUniform, miterLimit, etc..)

# Documentation
This section details the calculation of projections for each type of stroke line join.

**Summary**
* [1. Open path](#1-open-path)
    * [1.1 Butt](#1-1-butt)
    * [1.2 Round](#1-2-round)
    * [1.3 Square](#1-3-square)
* [2. Closed path](#2-closed-path)
    * [2.1 Miter](#2-1-miter)
    * [2.2 Bevel](#2-2-bevel)
    * [2.3 Round](#2-3-round)
        * [2.3.1 Round without skew](#2-3-1-round-without-skew)
        * [2.3.2 Round with skew](#2-3-2-round-skew)

Links:
* MDN: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin
* Spec: https://svgwg.org/svg2-draft/painting.html#StrokeLinejoinProperty
* Playground to understand how the line joins works: https://hypertolosana.github.io/efficient-webgl-stroking/index.html

## 1. Open path <a id='1-open-path'></a>
These cases differ in relation to the line-cap property ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap))

image

### 1.1 Butt <a id='1-1-butt'></a>

The projections are calculated by two vectors orthogonal to the stroke.

image

**Scaling effect**: only when the `stroke` is `uniform` that scaling affects the calculation of projections. It changes the position of the points ([see more details here](#scaling-effect)), so we must take it into account to have the new arrangement of the points.

**Skew effect**: just apply distortion after projection calculation.

### 1.2 Round <a id='1-2-round'></a>
WIP

### 1.3 Square  <a id='1-3-square'></a>
WIP

## 2. Closed path  <a id='2-closed-path'></a>
There are three types of stroke line join ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin))
image

### 2.1 Miter  <a id='2-1-miter'></a>
The corner is formed by extending the outer edges of the stroke at the tangents of the path segments until they intersect.

stroke_projection

**Stroke miter limit**: when two line segments meet at a sharp angle, it is possible for the join to extend, far beyond the thickness of the line stroking the path. The stroke-miterlimit imposes a limit on the extent of the line join, if so the join is converted from a miter to a [bevel](#2-2-bevel). ([MDN reference](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit))

image

The ratio of `miter length` (distance between the outer tip and the inner corner of the miter) to stroke-width is directly related to the angle (`alpha`) of the `bisector`:

$strokeMiterLimit = \frac{miterLength}{strokeWidth} = \frac{1}{sin(\alpha)}$

Note: when the `stroke` is `uniform`, the arrangement of the points is changed ([see explanation below](#scaling-effect)), consequently the strokeMiterLimit. In this case we must recalculate this limit.

**Scaling effect** <a id='scaling-effect'></a>: scaling changes the arrangement of points, hence the bisector calculation. Take a look:

Plotting a triangle with the points:
```
A (0, 0)
B (50, -200)
C (100, 0)
```
image

The angle of the bisector is `alpha`.

But when I apply a scaling of 2 only on the X axis, the angle of the bisector changes (`beta` < `alpha`), hence the miter vector too. To calculate this new angle we must apply the scaling on the points to find the new vectors that form `A`, `B` and `C`.

image

```
scaleX = 2 ; scaleY = 1

scaled A (0 * scaleX , 0 * scaleY) --> (0, 0)
scaled B (50 * scaleX, -200 * scaleY) --> (100,-200)
scaled C (100 * scaleX, 0 * scaleY) --> (200,0)
```

With these vectors we were able to correctly calculate the miter vector.

**Skew effect**: just apply distortion after projection calculation.

### 2.2 Bevel  <a id='2-2-bevel'></a>

The projections are both orthogonal to the vertex:

image

**Scaling effect**: only when the `stroke` is `uniform` that scaling affects the calculation of projections. It changes the position of the points ([see more details here](#scaling-effect)), so we must take it into account to have the new arrangement of the points.

**Skew effect**: just apply distortion after projection calculation.

### 2.3 Round <a id='2-3-round'></a>
This case has calculation of projections differently if there is or not the presence of skew.

#### 2.3.1 Round without skew <a id='2-3-1-round-without-skew'></a>

The projections are the two vectors parallel to the X and Y axes.

image

**Scaling effect**: does not directly affect projections (only stroke size).

#### 2.3.2 Round with skew <a id='2-3-2-round-skew'></a>

Projections are the points furthest from the vertex in relation to the direction of the X and Y axis respectively. See the example below.

Consider this case:
image

There is a point $(X_i, Y_i)$ that will be the furthest X distance from the origin after applying `skewX`. This is our point of interest.
image

We have:

**[eq.1]** $X_{afterSkewX} = X_i + Y_i * Tan(skewX)$

And we know that $Y_i$ is a function of $X_i$:

$Y_i^2 + X_i^2 = r^2$ 
**[eq.2]** $Y_i = \pm \sqrt{r^2 - X_i^2}$

Substituting eq.2 in eq.1 we have:

$X_{afterSkewX} = X_i \pm \sqrt{r^2 - X_i^2} * Tan(skewX)$

So we have an equation in terms of just X:

$f(X_i) = X_i \pm \sqrt{r^2 - X_i^2} * Tan(skewX)$

The plus/minus sign represent both sides of the circle after skewX:
image

The furthest `X after skewX` distance from the origin are the maximum/minimun of $f(X_i)$:
image

Computing $f'(X_i)=0$ for both cases, we find:

$X_i = \pm \frac{r}{\sqrt{1 + Tan^2(skewX)}}$

Substituting $X_i$ into the circle equation, we find $Y_i$:

$Y_i = \pm \sqrt{r^2 - X_i^2}$

That's our solution:

image

**Scaling effect**: does not directly affect projections (only stroke size).

# TO DO
- [x] Bring here all the documentation that is in the PR https://github.com/fabricjs/fabric.js/pull/8083/
- [ ] Handle the case when `skewX` and `skewY` are applied at the same time when stroke line join is `round` (still buggy)
- [x] Handle open path with `strokeLinecap` equals `round` and `square`
- [ ] Check if in the `_set` function when `skew` is applied, I need to invoke dimensioning for all line type joins
- [ ] Adjust http://fabricjs.com/custom-controls-polygon to work correctly with skew (another PR)
- [ ] handle 1 point edge case: `round` projection?


luizzappa:
@ShaMan123,

I had resolved the need to recalculate BBOX to only when `_set` method is invoked. However, this causes a point of attention. When a value is directly assigned to any of the properties (`skewX`, `skewY`, `scaleX`, `scaleY`), eg `canvas.getActiveObject().scaleX =2`, the `_set` method is not invoked, and the bbox is not recalculated. (I don't think there's a way to deal with it)

ShaMan123:
That is how fabric behaves.
It is still ambiguous but already there are features that depend on the dev using the set method.
It was known as a best practice but will become a must.
We might use setters to overcome this limitation as part of the TS migration.
So no worries.

ShaMan123:
For example text and group already depend on using set

ShaMan123:
Regarding polygon controls.
The community will appreciate that as well. There are many issues open asking to fix that.
I heard chat of making it part of core.
@asturur ?
v6 has changed some internals that broke it.

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+      let adjustedStrokeMiterLimit;
+      if (strokeUniform) {
+        // TODO: As it is in the MDN, the miterLimit can also be calculated as 1/sin(alpha/2) , I believe it is more performant than the current implementation.
+        const miterLimitVector = scaleHatVector(bisectorVector, strokeMiterLimit * s);
+        adjustedStrokeMiterLimit = hypot(miterLimitVector.x, miterLimitVector.y) / s;
+      } else {
+        adjustedStrokeMiterLimit = strokeMiterLimit
+      }
+
+      if (hypot(miterVector.x, miterVector.y) / s <= adjustedStrokeMiterLimit) {
```
ShaMan123, line 161: I think we can expose hypot on the Point class, naming it hypot or size

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+    /* MITER 
+      Calculation: the corner is formed by extending the outer edges of the stroke at the tangents of the path segments until they intersect
+      Visually detailed here: https://github.com/fabricjs/fabric.js/issues/8025
+    */
+    if (strokeLineJoin === 'miter') {
+      const scalar = -s / Math.sin(alpha / 2),
+        miterVector = scaleHatVector(bisectorVector, scalar);
+
+      // When two line segments meet at a sharp angle, it is possible for the join to extend,
+      // it is possible for the join to extend far beyond the thickness of the line stroking
```
ShaMan123, line 148: rephrase this with the line before?

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
-    coords.push(A.subtract(miterVector));
+    /* ROUND (without skew)
+      Calculation: the projections are the two vectors parallel to X and Y axes
+      Visually detailed here: https://github.com/fabricjs/fabric.js/pull/8083
+    */
+    if (strokeLineJoin === 'round' && skewX === 0 && skewY === 0) {
+      // correctSide is used to only consider projecting for the outer side 
+      const correctSide = new Point(
+          Math.abs(Math.atan2(bisectorVector.y, bisectorVector.x)) >= PiBy2 ? 1 : -1, 
+          Math.abs(Math.atan2(bisectorVector.x, bisectorVector.y)) >= PiBy2 ? 1 : -1
```
ShaMan123, line 179: utilize `calcAngleBetweenVectors`?

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+          proj1 = applySkew(A.add(orthogonal));
+        coords.push({
+          "projectedPoint": proj1,
+          "originPoint": A,
+          "bisector": bisector
+        });
+      });
+
+      // The points furthest from the vertex in the direction of the X and Y axes after distortion
+      // TODO: consider only projections that are inside the beginning and end of the circle segment
```
ShaMan123, line 215: not clear to me

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+        coords.push({
+          "projectedPoint": applySkew(A.subtract(vector)),
+          "originPoint": A,
+          "bisector": bisector
+        });
+      })
+    }
+    else {
+      /* BEVEL OR MITER GREATER THAN STROKE MITER LIMIT
+        Calculation: the projection points are formed by the orthogonal to the vertex.
```
ShaMan123, line 244:
```suggestion
        Calculation: the projection points are formed by the vector orthogonal to the vertex.
```

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
       B = points[index - 1];
       C = points[index + 1];
     }
-    const bisector = getBisector(A, B, C),
+
+    //  safeguard in case `points` are not `Point`
+    B = new Point(B.x, B.y);
+    C = new Point(C.x, C.y);
+
+    /* OPEN PATH
```
ShaMan123, line 97:
```suggestion
    /* OPEN PATH START/END
```

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+      TODO: take into account line-cap (I believe that should change the projections)
+    */
+    if (openPath && (index === 0 || index === points.length - 1)) {
+      var D = index === 0 ? C : B, 
+        // When the stroke is uniform, scaling affects the arrangement of points. So we must take it into account.
+        vector = createVector(strokeUniform ? A.multiply(scale) : A, strokeUniform ? D.multiply(scale) : D), 
+        hatOrthogonalVector = getOrthogonalUnitVector(vector), 
+        orthogonalVector = scaleHatVector(hatOrthogonalVector, s);
+
+      // We must bounce the projection to both sides, since we don't know which one is the outer edge
```
ShaMan123, line 110: This opens up the bug that we discussed, the case of a big `strokeWidth` value.
The wrong projection might cross over the shape's other side.
We should use intersection IMO.
`intersectLinePolygon` should do the job

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
     scaleY,
+    skewX,
+    skewY
   }: projectStrokeOnPointsOptions,
   openPath: boolean
-): Point[] => {
-  const coords: Point[] = [],
+): {projectedPoint: Point, originPoint: Point, bisector?: { vector: Point, angle: number }}[] => {
+
+  const coords: {projectedPoint: Point, originPoint: Point, bisector?: { vector: Point, angle: number }}[] = [],
```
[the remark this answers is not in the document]
ShaMan123: Or define a type for the return type and import it.

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+
+    //  safeguard in case `points` are not `Point`
+    B = new Point(B.x, B.y);
+    C = new Point(C.x, C.y);
+
+    /* OPEN PATH
+      If open path, no matter the type of the line join, the line is always the same
+      Calculation: to find the projections, just find the points orthogonal to that stroke 
+      Visually detailed here: https://github.com/fabricjs/fabric.js/issues/8025
+      TODO: take into account line-cap (I believe that should change the projections)
```
ShaMan123, line 101: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/lineCap#comparison_of_line_caps

for sure

image

ShaMan123 (commented):
What do you think about splitting up the code and creating a function for each case?
If you don't like that I think we should define A, B, C, AB, AC at the top if possible. 

- I saw some `var`. They should all become `const` or `let`. VSCODE has an option to quick fix all occurrences if you hover over one.

src/util/misc/projectStroke.ts
ShaMan123: Then using the ray algorithm concept, `intersectedPoints % 2`

luizzappa:
> What do you think about splitting up the code and creating a function for each case? If you don't like that I think we should define A, B, C, AB, AC at the top if possible.

The code will be cleaner, I will refactor

src/shapes/polyline.class.ts
```diff
@@ -142,6 +142,46 @@ import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
...
+            preScalingStrokeValue = 0;
+            postScalingStrokeValue = strokeWidth;
+        }
+        else {
+            preScalingStrokeValue = strokeWidth;
+            postScalingStrokeValue = 0;
+        }
+        const dimX = this.width + preScalingStrokeValue, 
+          dimY = this.height + preScalingStrokeValue, 
+          finalDimensions = new Point(dimX * this.scaleX, dimY * this.scaleY);
```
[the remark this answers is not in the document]
luizzappa: No, in the calculation of projections the `skew` effect is computed individually for each one (as it depends on X, Y). So the width/height are already taking into account the skew

```
applySkew = (v: Point) => {
      let vector = new Point(v);
      vector.y += vector.x * Math.tan(degreesToRadians(skewY));
      vector.x += vector.y * Math.tan(degreesToRadians(skewX));
      return vector
    }
```

It is a little different from what happens with the `scale` that produces the same effect (only a multiplication) for all projections. This allows you to purge it from the projections and then multiply the width/height by the scale to find the correct values

```  
finalDimensions = new Point(dimX * this.scaleX, dimY * this.scaleY);
```

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
+      From here on, we will deal with closed path cases, which have different projections according to the line join type.
+      MDN: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin
+      Spec: https://svgwg.org/svg2-draft/painting.html#StrokeLinejoinProperty
+      Playground to understand how the line joins works: https://hypertolosana.github.io/efficient-webgl-stroking/index.html
+      View the calculated projections here for each of the control points: https://codesandbox.io/s/project-stroke-points-with-context-to-trace-b8jc4j?file=/src/index.js
+    */
+
+    // First we calculate the bisector between the points. Used in round and miter cases
+    // When the stroke is uniform, scaling changes the arrangement of the points, so we have to take it into account
+    const bisector = strokeUniform
```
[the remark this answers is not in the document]
luizzappa: outdated after refactoring

src/util/misc/projectStroke.ts
luizzappa: This is the case explained here: https://github.com/fabricjs/fabric.js/pull/8083#issuecomment-1264450123

`I'm considering the two points for projection, this can be a problem when the control points are too close. I'll adjust to consider only the point belonging to the yellow segment.`

src/util/misc/projectStroke.ts
luizzappa: It's a little different here, I'm not calculating the angle between vectors.. I'm just checking which quadrant the bisector belongs to to find out which of the two projections will be chosen.

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +49,216 @@ export const projectStrokeOnPoints = (
...
-    getStrokeHatVector = (v: Point) => {
-      const scalar = s / Math.hypot(v.x, v.y);
-      return new Point(
-        v.x * scalar * strokeUniformScalar.x,
-        v.y * scalar * strokeUniformScalar.y
-      );
+    scaleHatVector = (hatVector: Point, scalar: number) => {
+      return hatVector.multiply(strokeUniformScalar).scalarMultiply(scalar);
+    },
+    applySkew = (v: Point) => {
```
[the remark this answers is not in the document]
luizzappa: All make use, except for the case of `round without skew`

```
  /**
   * ROUND (without skew)
   * Calculation: the projections are the two vectors parallel to X and Y axes
   * 
   * @see https://github.com/fabricjs/fabric.js/pull/8083
   */
  const projectionsRoundNoSkew = (
    A: Point,
    B: Point,
    C: Point,
    bisector: ReturnType<typeof getBisector>
  ): returnedProjections => {
    // correctSide is used to only consider projecting for the outer side 
    const bisectorVector = bisector.vector,
      correctSide = new Point(
        Math.abs(Math.atan2(bisectorVector.y, bisectorVector.x)) >= PiBy2 ? 1 : -1, 
        Math.abs(Math.atan2(bisectorVector.x, bisectorVector.y)) >= PiBy2 ? 1 : -1
      ),
      radiusOnAxisX = new Point(s * strokeUniformScalar.x, 0).multiply(correctSide), 
      radiusOnAxisY = new Point(0, s * strokeUniformScalar.y).multiply(correctSide),
      proj1 = applySkew(A.add(radiusOnAxisX)), 
      proj2 = applySkew(A.add(radiusOnAxisY));

    return [proj1, proj2].map(proj => ({
      "projectedPoint": proj,
      "originPoint": A,
      "bisector": bisector
      })
    );
  };
```

src/util/misc/projectStroke.ts
```diff
@@ -1,12 +1,20 @@
...
- * - `miter`: 2 points corresponding to the outer boundary and the inner boundary of stroke.
- * - `bevel`: 2 points corresponding to the bevel boundaries, tangent to the bisector.
- * - `round`: same as `bevel`
+ * Project stroke width on points returning projections for each point as follows:
+ * - `miter`: 1 point corresponding to the outer boundary and the inner boundary of stroke.
+ * - `bevel`: 2 points corresponding to the bevel possible boundaries, orthogonal to the stroke.
+ * - `round`: same as `bevel` when it has no skew, with skew are 4 points.
  * Used to calculate object's bounding box
+ * 
+ * @see https://github.com/fabricjs/fabric.js/pull/8083
```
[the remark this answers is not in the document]
luizzappa: Yes, when I migrate the documentation to this PR, I will create anchors for each case, and change the link in the code. That way, it will be easier to find the documentation.

src/util/misc/projectStroke.ts
luizzappa: I think it's a good one! I called it magnitude, if you prefer another name feel free to change it https://github.com/fabricjs/fabric.js/pull/8344/commits/3eec3d1c021fec636d5a39e6a560da4048c7c082

src/util/misc/projectStroke.ts
luizzappa: I think that in this specific case there is no risk, because there is a stroke on both sides. 

The problem happened in `miter` that one side had a stroke and the other didn't (and this projection here that shits everything). But it was solved considering the direction of the bisector.

ShaMan123: I am off for a few days
When I vome back I'll review everything and see if I have a case or not.

src/util/misc/projectStroke.ts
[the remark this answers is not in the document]
ShaMan123: Sorry
I meant does this apply to transfirming points in general

src/util/misc/projectStroke.ts
ShaMan123: Well you are... between the vector and the x hat vector

src/util/misc/projectStroke.ts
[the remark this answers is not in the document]
luizzappa: I did not understand your suggestion very well, would add this function in `util`? If that's it, it doesn't seem to me to be something used in other scenarios.

src/util/misc/projectStroke.ts
luizzappa: Ok, you're right.. It seems to me to be more intelligible, although I have to instantiate a class

```
Math.abs(calcAngleBetweenVectors(new fabric.Point(1, 0), bisectorVector))
```
..than this:

```
Math.abs(Math.atan2(bisectorVector.y, bisectorVector.x))
```

src/util/misc/projectStroke.ts
luizzappa: Okay, I'll update the documentation and then work on `skewX` and `skewY` applied at the same time.

luizzappa: Take a look at codesandbox ( https://codesandbox.io/s/project-stroke-points-with-skew-correction-yz7c77?file=/src/index.js ). No matter the stroke size, both sides can always influence the BBOX calculation:
openpath

The problem is in `miter`, when we project two sides, one of which has no stroke (observe the behavior of the red vector, if we took it into account, it would distort the BBOX calculation):

openpath miter

ShaMan123: I got confused.
I meant when `strokeLineCap === 'round'` but even in this case I don't see it is a possible case

ShaMan123 (changes requested):
- run `npm run prettier:write` and commit the changes
- checkout the change I did the the polgon unit tests I did in the prev PR to dix the tests https://github.com/fabricjs/fabric.js/pull/8083/files#diff-e37e2c1d592e1dddf9ca614c2e5cc9df854e6b4ee55f40eb9bbf3f2ae67498bc
- the main comment is to refactor the functions to be completely standalone. see my comments.
- the rest are es6/ts stuff I can polish

After that I will polish what remains and move towards approval

My last concern is that we unearthed a bug in the skewing controls. I am not sure if you noticed it or not. But it was there before this fix. But now it has become more noticeable and accute since (I am guessing) `_setPositionDimensions` is called all the time. I will look into meanwhile and update this comment with a video.

src/util/misc/projectStroke.ts
```diff
@@ -39,62 +54,287 @@ export const projectStrokeOnPoints = (
...
+  */
+
+  /**
+   * OPEN PATH START/END - Line cap: Butt
+   * If open path, no matter the type of the line join, the line is always the same
+   * Calculation: to find the projections, just find the points orthogonal to that stroke 
+   * 
+   * @see https://github.com/fabricjs/fabric.js/pull/8344#1-1-butt
+   */
+  const projectionsOpenPathButt = (
```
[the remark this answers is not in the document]
luizzappa: I was avoiding this because it's too much context to pass through the functions. I made the changes ([#034a90a](https://github.com/fabricjs/fabric.js/pull/8344/commits/034a90a4de872246ac283522d54bf4b6ea5a3a2b)), but it seems to me that it became less readable

Doing the transformations before passing doesn't seem appropriate to me, because the calculation logic will be split. For example, not all cases use the vectors `AB` and `AC` (see the case of the `projectionsRoundNoSkew` function).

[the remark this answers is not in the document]
ShaMan123: I will take a look
Can you run prettier so I can work on the code as well?

ShaMan123:
Meanwhile I have managed to fix skewing with controls.
I will PR soon so no worries about that.

src/util/misc/projectStroke.ts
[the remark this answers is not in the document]
luizzappa: @ShaMan123, just confirming, I commit only the prettier modified files that we're working on here (polyline.class.ts, projectStroke.ts, etc), right? Because it changed a bunch of files..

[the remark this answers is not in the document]
ShaMan123: Hmm.. prettier should have only changed the committed files.
The rest should be prettified already.
The files show up in github desktop but once you commit they are filtered out as empty.

luizzappa:
> Meanwhile I have managed to fix skewing with controls. I will PR soon so no worries about that.

I tried to reproduce it here, but I couldn't.. Anyway, one less problem

scale skew

The bug I had seen is when we edited the control points, but I already indicated it as a TODO.

ShaMan123:
What does the gif show?

luizzappa:
> What does the gif show?

Scaling an object with skewX = 45. Isn't that what was bugged? Now reading your message better, the bug was when the control changed the skew of the object?

src/util/misc/projectStroke.ts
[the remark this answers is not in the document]
luizzappa: https://github.com/fabricjs/fabric.js/pull/8344/commits/1636dd98c20d972f0484425710afcec8d39e3fd5

ShaMan123:
Yes, skew an object with the controls. You will see that the objects flickers between the initial and current skew values.
It originated in the control logic and `_setPositionDimensions` did in fact made it worse.

luizzappa:
> * add changelog log

Direct me here... What do I do? Do I add this to the `changelog.md` file under `## [next]`?

`fix(polyline): stroke bounding rect calculation [#8344]`

Or is it something more detailed, like:
```
$ git log --pretty="- %s (%h)"
```

luizzappa:
I ran prettier (https://github.com/fabricjs/fabric.js/commit/1636dd98c20d972f0484425710afcec8d39e3fd5) and committed. I'm done for today, tomorrow I'm back.

ShaMan123:
`fix(polyline): stroke bounding box calculation [#8344]`

ShaMan123:
Naming:
We need to decide what is best `hat` or `unit` and make everything the same.
- `getHatVector`
- `getOrthogonalUnitVector`
- `scaleHatVector`

ShaMan123:
I agree that the code is now unreadable and messy.
I suggest 2 options:
- create a class `ProjectionsContext` that will have all options passed to the main function and expose all needed methods such as `applySkew(v: Point)` (skew values referenced inside the method using `this`) etc. ad pass an instance of this class around.
- create a function `consumeOptions(options)` and pass `options` around

I think a class is cleaner.
I will give it a go.

ShaMan123:
I have refactored the PR into classes.
Take a look https://github.com/luizzappa/fabric.js-bbox/compare/poly-stroke-bounding-rect-ts...fabricjs:fabric.js:projections-context

ShaMan123:
After that we need to tackle the tests
I see there's a consistent diff of 1 px in outcomes (consistent is good)

ShaMan123:
some test is stuck... I will investigate

src/shapes/polyline.class.ts
```diff
@@ -149,28 +149,18 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
...
-          postScalingStrokeValue = strokeWidth;
-        } else {
-          preScalingStrokeValue = strokeWidth;
-          postScalingStrokeValue = 0;
-        }
-        const dimX = this.width + preScalingStrokeValue,
-          dimY = this.height + preScalingStrokeValue,
-          finalDimensions = new Point(dimX * this.scaleX, dimY * this.scaleY);
-        return finalDimensions.scalarAdd(postScalingStrokeValue);
+      _getTransformedDimensions(options) {
```
luizzappa: Nice! I overcomplicated..

src/util/misc/vectors.ts
```diff
@@ -75,8 +73,6 @@ export const getBisector = (A: Point, B: Point, C: Point) => {
  */
 export const getOrthogonalUnitVector = (
   v: Point,
-  counterClockwise: boolean = true
```
luizzappa: Just out of curiosity, does the typing cause any problems in this case?

luizzappa:
> Naming: We need to decide what is best `hat` or `unit` and make everything the same.
> 
> * `getHatVector`
> * `getOrthogonalUnitVector`
> * `scaleHatVector`
> * `unitVectorX`

We can standardize as Unit Vector. It seems to be more intelligible in several languages ​​(at least in portuguese, hat vector is not used much..)

luizzappa:
> I have refactored the PR into classes. Take a look [diff](https://github.com/luizzappa/fabric.js-bbox/compare/poly-stroke-bounding-rect-ts...fabricjs:fabric.js:projections-context)
> 
> Changes:
> 
> * ~returns only points (I don't mind reverting if there is a need, I just did it since I was in a deep refactor anyways)~ [d2dc716](https://github.com/fabricjs/fabric.js/commit/d2dc71602ac144e5909e4c66f718b45366e8ecb7)
> * class `StrokeProjections` in charge of line join (closed path)
> * class `OpenStrokeProjections` in charge of line cap (open path)
> * class `StrokeProjectionsBase` consists of all shared utils

Wow, it turned out great!!! Very easy to read and understand what is going on

> * we can rename to reflect strokeLineJoin/Cap instead of closed/open projections

yes, it makes sense

src/util/misc/vectors.ts
ShaMan123: On the contrary
It marked it as redundant

ShaMan123:
I have renamed the classes.
What do you say `getOrthonormalVector` instead of `getOrthogonalUnitVector`?
Normal is confusing. In this context it is "normalized" to 1 but in other is the normal, the orthogonal vector to the plane.
We leave it as is.

ShaMan123:
OK
Your turn!
> If you approve I would like you to merge it on your behalf so it has your name on it. You can do it by squash merging, undoing and then committing and pushing or something.

You can apply a diff as well BTW.
It is important to keep the contribution on your name!

After you merge it I want to address the following as part of tackling tests:
- 1 point edge case: should we return a round projection in this case?
- 0 size edge case: size returned is -1 because of stroke width equaling 1. I thought of doing a Math.max in `_setPositionDimensions`

src/shapes/polyline.class.ts
```diff
@@ -79,36 +80,41 @@ import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
...
+      _setPositionDimensions: function ({ left, top, fromSVG } = {}) {
+        const bbox = this._calcDimensions(),
+          strokeCorrection = new Point()
+            .scalarAdd(this.strokeWidth)
+            .divide(
+              this.strokeUniform
+                ? new Point(this.scaleX, this.scaleY)
+                : new Point(1, 1)
+            );
+        this.width = bbox.width - strokeCorrection.x;
```
ShaMan123, line 92: Here the 0 size = -1 edge case I thought of wrapping with Math.max
I wonder if it's related somehow to the 1px diff in tests

src/util/misc/projectStroke/index.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export const projectStrokeOnPoints = (
+  points: IPoint[],
+  options: TProjectStrokeOnPointsOptions,
+  openPath = false
+): TProjection[] => {
+  const projections: TProjection[] = [];
+
+  if (points.length <= 1) {
+    return projections;
```
ShaMan123, line 21: Here is the 1 point edge case that we need to consider returning a round projection

ShaMan123 (commented):
Great!
We had a git concurrency that ended up favorable. I committed the last rename and a moment after you squashed.

ShaMan123:
I see you decided for `getOrthonoraml`

ShaMan123:
I am done pushing for now
We can have a chat regarding tests and what remains.
Interested?

luizzappa:
Yes, we can talk tomorrow. It's late here already, needing to sleep haha ​​tomorrow is a holiday, I'd have the day to work on that.

ShaMan123:
great

luizzappa:
I had an insight in the case of `skewX` and `skewY` applied at the same time when stroke line join is `round`, now it's working [#2ff9ee2](https://github.com/fabricjs/fabric.js/pull/8344/commits/2ff9ee26dbee6e041b1122ce7216191f39dcf316)

luizzappa:
@ShaMan123  Let's focus on tests.

I remembered that I had encountered a strange behavior when running the tests, I detailed it better in the PR that we were dealing with the tests:
https://github.com/fabricjs/fabric.js/pull/8087#issuecomment-1229516514

src/util/misc/projectStroke/StrokeLineCapProjections.ts
```diff
@@ -52,7 +51,7 @@ export class StrokeLineCapProjections extends StrokeProjectionsBase {
    * @see https://github.com/fabricjs/fabric.js/pull/8344#1-1-butt
    */
   private projectButt() {
-    return [this.projectOrthogonally(this.A, this.T)];
+    return [this.projectOrthogonally(this.A, this.T, this.strokeProjectionMagnitude), this.projectOrthogonally(this.A, this.T, -this.strokeProjectionMagnitude)];
```
ShaMan123, line 54: I disagree.
The projection is always to the direction opposite to the vector AT

ShaMan123 (commented):
Yes!

ShaMan123:
Regarding tests.
Let's do the following:
1. make existing tests pass in the browser
2. make existing tests pass in node
3. merge this PR
4. open a testing PR
5. add visual tests
6. make them pass and investigate the spill over (could be that the goldens were getting recreated twice? Anyways the test suite was patched a lot since #8087)

Thoughts?

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -185,7 +184,7 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
       furthestX = new Point(
         newX,
         Math.sqrt(
-          circleRadius.y ** 2 - ((newX * circleRadius.y) / circleRadius.x) ** 2
+          newY ** 2 - ((newX * circleRadius.y) / circleRadius.x) ** 2
```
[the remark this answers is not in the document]
ShaMan123: Well I am a bit lazy.
I will look at the equations and derivative again.

ShaMan123:
Do you want to chat and brainstorm about tests?

luizzappa:
> Do you want to chat and brainstorm about tests?

I think it's a good one, what do you suggest?

I saw that the tests have changed, I no longer need to clone `fabricjs.com`? I'm trying with the command `npm run test:visual-browser` but I get an error I'm trying to figure out what it is

ShaMan123:
Let's chat.
My email is public in my profile.
Send me an email and we can do a google meet

luizzappa:
I fixed the custom controls to work with `skew`. However, it revealed a bug when `skewX` and `skewY` are applied at the same time.
skewX skewY same time

I was trying to avoid it, but there's no way, the solution will be by the equations:

[eq. 1] $X_f = X + Y * Tan(skewX)$

Limited by the equation (the circle of the `round` line join):

[eq.2] $(Y + X * Tan(skewY))^2 = r^2 - X^2$
So, $Y = \pm (\sqrt{r^2 - X^2} - X * Tan(skewY))$

Substituting in [eq.1]:
[eq.3] $X_f = X \pm (\sqrt{r^2 - X^2} - X * Tan(skewY))$

The solutions are the points where the derivative of [eq.3] is equal to zero. The first time I tried to solve it was not very trivial..

I'll focus on the `line-caps` cases, I think it will be easier. Then I come back to this case.

luizzappa:
@ShaMan123, if you need help getting the test to work, post an example of SVG that is failing. I'll import it on the canvas and do a more detailed debug of what is happening

ShaMan123:
I forgot about the `-l` testing option.
It launches the tests in the browser. 
So if you run `npm test -- -c chrome -l` you will be able to see visual diffs. You can use `--dev` as well for tests to rerun on changes.

luizzappa:
Nice! I will try tomorrow

src/util/misc/projectStroke/StrokeLineCapProjections.ts
ShaMan123: ?

ShaMan123:
I have another issue to discuss. `_getTransformedDimensions`.
I would argue that we should pass `strokeWidth: 0` as we do with `skewX/Y` since we account for stroke width when projecting.
I have tried this and found out that this is problematic. 
The logic assumes `A` is in the right position but the truth is it isn't. `strokeWidth` shifts the bbox since it is applied (unfortunately) from the top left corner.
I guess if we shift `A` this will be solved. 
I prefer this since it is less complicated since it doesn't relay on `_getTransformedDimensions` to patch the bbox.
This might or might not be related to the inaccuracy you noticed.

ShaMan123:
Since it seems like a nightmare we can consider doing it in the follow up PR

ShaMan123:
Perhaps it is a design issue.
I would expect that when I pass points to a polygon it will render exactly these points unshifted regardless of stroke width.
I will look into it closer.
Many bugs are surfaced and solved in this PR

ShaMan123 (commented):
The last commits revert `PathUtils.ts`, return all context from `_calcDimensions` and use `setDimensions` instead of `_setPositionDimensions` or `calcBBox`

ShaMan123:
After some thought I suggest we finalize this PR, perhaps checkout `Path` and restore on it `_setPositionDimensions` so it is completely untouched by this PR.

luizzappa:
> I have another issue to discuss. `_getTransformedDimensions`. I would argue that we should pass `strokeWidth: 0` as we do with `skewX/Y` since we account for stroke width when projecting. I have tried this and found out that this is problematic. The logic assumes `A` is in the right position but the truth is it isn't. `strokeWidth` shifts the bbox since it is applied (unfortunately) from the top left corner. I guess if we shift `A` this will be solved. I prefer this since it is less complicated logically as it doesn't relay on `_getTransformedDimensions` to patch the bbox. This might or might not be related to the inaccuracy you noticed.
> 
> **This will make `strokeCorrection` obsolete**

If I'm not mistaken, this "separation" of the `strokeWidth` (removes its effect on `_calcBBox()`, formerly `_setPositionDimensions()`, and adds on `_getPositionDimensions()`) makes it possible that when the `strokeUniform = false`, the resizing of the object occurs without the need to recalculate the projections, it's just adjustment in `scale` and in `strokeWidth` inside `_getPositionDimensions()`). 

Basically we would have to extend recalculation of projections here to all cases:

https://github.com/fabricjs/fabric.js/blob/5f57838b8ceada13f332ba23b582302bb4082fea/src/shapes/polyline.class.ts#L172-L179

I agree with you that it is clearer, but at a price of performance drop in non-uniform stroke.

luizzappa:
> After some thought I suggest we finalize this PR, perhaps checkout `Path` and restore on it `_setPositionDimensions` so it is completely untouched by this PR.

Restore  `_setPositionDimensions()` in polyline, too?

ShaMan123:
No, not for polyline.
And hopefully not for path if we manage to make all tests pass.
Did you run tests?
I almost fixed path. Still position is a bit off. Besides that seems OK.
Would love your input.
The difficulty is that some tests pass as false positives but should fail (it's the passing percentage that is inaccurate in many cases)


image



> If I'm not mistaken, this "separation" of the `strokeWidth` (removes its effect on `_calcDimensions()`, formerly `_setPositionDimensions()`, and adds on `_getPositionDimensions()`) makes it possible that when the `strokeUniform = false`, the resizing of the object occurs without the need to recalculate the projections, it's just adjustment in `scale` and in `strokeWidth` inside `_getPositionDimensions()`).

I am not sure I understand why scaling will require recalculation in this case. We will still use `_getTransformedDimensions` but without strokeWidth values.
Anyways **not** in this PR.

src/shapes/polyline.class.ts
ShaMan123: solved!

src/util/misc/projectStroke/index.ts
ShaMan123: solved

src/util/misc/projectStroke/StrokeLineCapProjections.ts
ShaMan123: @luizzappa ?

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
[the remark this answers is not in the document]
ShaMan123: I would like an explanation regarding the radius ratios `circleRadius.y / circleRadius.x`

src/util/misc/projectStroke/StrokeLineCapProjections.ts
```diff
@@ -0,0 +1,70 @@
...
+   * OPEN PATH START/END - Line cap: Butt
+   * If open path, no matter the type of the line join, the line is always the same
+   * Calculation: to find the projections, just find the points orthogonal to that stroke
+   *
+   * @see https://github.com/fabricjs/fabric.js/pull/8344#1-1-butt
+   */
+  private projectButt() {
+    return [
+      this.projectOrthogonally(this.A, this.T, this.strokeProjectionMagnitude),
+      this.projectOrthogonally(this.A, this.T, -this.strokeProjectionMagnitude),
```
ShaMan123, line 56: remove this IMO

ShaMan123 (approved):
> assumes `A` is in the right position but the truth is it isn't. `strokeWidth` shifts the bbox since it is applied (unfortunately) from the top left corner. I guess if we shift `A` this will be solved. I prefer this since it is less complicated logically as it doesn't relay on `_getTransformedDimensions` to patch the bbox. This might or might not be related to the inaccuracy you noticed.
> 
> **This will make `strokeCorrection` obsolete**

d6b07f4 - this fix
0157f22 - fix `exactBoundingBox = false`: bbox calculation should revert to the legacy (wrong) calculation.


Round + skew is still inaccurate.
We can move forward IMO.

ShaMan123:
Now the demo needs fixing of `setPositionByOrigin`
The point is off perhaps, or the origin.

luizzappa:
> > If I'm not mistaken, this "separation" of the `strokeWidth` (removes its effect on `_calcDimensions()`, formerly `_setPositionDimensions()`, and adds on `_getPositionDimensions()`) makes it possible that when the `strokeUniform = false`, the resizing of the object occurs without the need to recalculate the projections, it's just adjustment in `scale` and in `strokeWidth` inside `_getPositionDimensions()`).
> 
> I am not sure I understand why scaling will require recalculation in this case. We will still use `_getTransformedDimensions` but without strokeWidth values. Anyways **not** in this PR.

I was wrong, forget what I said.

luizzappa:
> No, not for polyline. And hopefully not for path if we manage to make all tests pass. Did you run tests? I almost fixed path. Still position is a bit off. Besides that seems OK. Would love your input. The difficulty is that some tests pass as false positives but should fail (it's the passing percentage that is inaccurate in many cases)

I'm taking a look now

luizzappa:
> Now the demo needs fixing of `setPositionByOrigin` The point is off perhaps, or the origin. I tend to say the origin.

I fixed the codesandbox, it was the `getObjectSizeWithStroke()`

luizzappa:
Looking at git actions, it looks like you were able to fix, as it's passing all the visual/unit tests.

> The difficulty is that some tests pass as false positives but should fail (it's the passing percentage that is inaccurate in many cases)

And how can you identify a false positive? To double check, I took the example of your print, 
and I imported it directly into the canvas, visually it seems to be correct:

[Generated golden]
svg_stroke_6

[canvas --> [https://codesandbox.io/s/goldens-test-2f3785](https://codesandbox.io/s/goldens-test-2f3785) ]
image


I've tried running all the tests over here, in `firefox` it works, but in `chrome` one of them fails by timeout. Is there any setting to increase this limit?

image

```
browser log: |
            {"type":"error","text":"Error: Browser timeout exceeded: 10s"}
            {"type":"error","text":"Error while executing test: Image resize filter test: Image resize with scaleY != scaleX"}
            {"type":"error","text":"[1014/002106.488:WARNING:registry_dict.cc(140)] Failed to convert string to dictionary\r\n\r\nDevTools listening on ws://127.0.0.1:9222/devtools/browser/1490c2bb-8bf5-4b09-a464-69f7da219401\r\n[1014/002111.794:INFO:CONSOLE(173)] \"Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently\", source: http://localhost:8081/227760705910000/test/lib/visualTestLoop.js (173)\r\n[1014/002111.922:INFO:CONSOLE(173)] \"Canvas2D: Multiple readback operations using getImageData are faster with the willReadFrequently attribute set to true. See: https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-will-read-frequently\", source: http://localhost:8081/227760705910000/test/lib/visualTestLoop.js (173)\r\n[1014/002111.947:INFO:CONSOLE(173)] 
(...)
```

ShaMan123:
I have increased the timeout. Should have been done regardless.
Tests pass. I have checked them.

ShaMan123:
Skewing + editing points is broken

ShaMan123:
Exposed `strokeOffset` so a dev can position the object after stroke

ShaMan123:
Recap regarding position.
I was wrong. The points position are kept. The bbox extends to cover stroke projections. This is the expected behavior. 
If a dev needs to adjust the position the exposed `strokeOffset` will provide the needed context to do that.
I am really happy with this!

luizzappa:
> Skewing + editing points is broken in the demo

fixed

luizzappa:
> I have increased the timeout. Should have been done regardless. Tests pass. I have checked them.

That wasn't enough. This setup did the trick [(source)](https://github.com/testem/testem/issues/1021):

`browser_start_timeout: 60,`

inside:
https://github.com/fabricjs/fabric.js/blob/ab02461ca78682cd6bed0840c5979fede267213a/test/testem.config.js#L6-L10



I don't think my machine is that bad (i3 16b of ram).. 

Anyway this test case failed, the tests here are kind of intermittent. Sometimes it fails, sometimes it passes:

image

image

luizzappa:
Regarding the test timeout, it seems to me that it is something related to the antivirus, its CPU consumption increases a lot when the tests start the browser process.

ShaMan123:
Some tests are unstable 
On my machine when I run chrome gradient fails sometimes.
There differences between browsers.


> That wasn't enough. This setup did the trick https://github.com/testem/testem/issues/1021:

> `browser_start_timeout: 60,`

Push this fix. I say these errors on CI and didn't know where to trace them.

src/util/misc/projectStroke/StrokeLineCapProjections.ts
ShaMan123: I got it wrong

ShaMan123 (commented):
Ready to merge on my end.
Do you want to fix anything else before I merge?
Line cap?

ShaMan123:
@asturur now is a good time for your feedback

ShaMan123:
from the [spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-linecap-dev)
image

ShaMan123:
Added line cap.
Please confirm (round especially)

ShaMan123 (commented):
Ready

ShaMan123 (approved):
Ready including line cap

src/util/misc/projectStroke/index.ts
ShaMan123: needs validation

ShaMan123 (approved):
Added JSDOC

ShaMan123:
this PR is epic

asturur:
Jumping in for a read. I won't have any kind of breaking comments at this stage

src/point.class.ts
```diff
@@ -351,6 +352,14 @@ export class Point {
     return new Point(this.x, this.y);
   }
 
+  /**
+   * return the magnitude of a vector
+   * @return {number}
+   */
+  magnitude(): number {
```
asturur, line 359: is it the distance? can we add in the JSDOC ` return the magnitued ( or distance from origin ) of a point` i m not sure we should refer a vector and a point like if they are the same thing

src/shapes/polyline.class.ts
```diff
@@ -79,56 +93,111 @@ import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
...
+            key === 'skewX' ||
+            key === 'skewY' ||
+            key === 'strokeMiterLimit' ||
+            key === 'strokeLineCap' ||
+            key === 'strokeLineJoin' ||
+            key === 'strokeLineWidth' ||
+            key === 'strokeUniform' ||
+            key === 'points')
+        ) {
+          this.setDimensions();
```
asturur, line 198: this wasn't automatic before even with the simple calculation, because the simple logic was good ( well except wasn't exact ) for every case.
So all good to me, but adding `points` here adds a new automatic feature and is partial.

I can set the whole array of points and i can get a bbox update, while i can't mutate a point and do the same.
Not sure this checking for points is a good idea outside skewX and skewY.
a end user can change skewX and skewY alone and so yes, maybe the bbox should update in that case, but the other changes are a developer choice, with JS code, and in that js code there should be a setDimensions too.

src/shapes/polyline.class.ts
```diff
@@ -235,19 +304,25 @@ import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
...
         );
-      parsedAttributes.fromSVG = true;
-      callback(new fabric[_class](points, extend(parsedAttributes, options)));
+      callback(
+        new fabric[_class](points, {
+          ...parsedAttributes,
+          ...options,
+          // we pass undefined to instruct the constructor to position the object using the bbox
+          left: undefined,
+          top: undefined,
```
asturur, line 322: left and top can just be in parsedAttributes. can we just do:
```js
const { left, top, allOtherAttributes } = parseAttributes(
          element,
          fabric[_class].ATTRIBUTE_NAMES
        );
```

and avoid setting back to undefined?

Also why this change? i ll read the description, if you know is not explained there can you elaborate?

src/util/misc/boundingBoxFromPoints.ts
```diff
@@ -1,21 +1,30 @@
...
-export const makeBoundingBoxFromPoints = (points: Point[]) => {
+export const makeBoundingBoxFromPoints = (points: IPoint[]) => {
+  if (points.length === 0) {
+    return {
+      left: 0,
+      top: 0,
+      width: 0,
+      height: 0,
+    };
+  }
```
asturur, line 18: when this actually happens?
i would prefer a default value for points rather than a specific case in the code.

```suggestion
const defaultPoints = [{ x: 0, y: 0 }];
export const makeBoundingBoxFromPoints = (points: IPoint[] = defaultPoints) => {
```

can't TS help here saying an empty array isn't good?

asturur: yeah my suggestion is garbage, because i wasn't thinking empty array doesn't trigger the default.

what about this stack overflow finding?
```
type NonEmptyArray<T> = T[] & { 0: T };
````

File: src/util/misc/boundingBoxFromPoints.ts [hunk 26]
```diff
@@ -1,21 +1,30 @@
...
+
   const { min, max } = points.reduce(
     ({ min, max }, curr) => {
       return {
         min: min.min(curr),
         max: max.max(curr),
       };
     },
-    { min: points[0], max: points[0] }
+    { min: new Point(points[0]), max: new Point(points[0]) }
```
asturur, line 27: was this the line that was bothering the compiler and forcing us use the early return?

src/util/misc/hypot.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @param {...number} values
+ * @return {number}
+ */
+export const hypot = (...values: number[]): number => {
+  let sumOfSquares = 0;
+  for (let i = 0; i < values.length; i++) {
+    sumOfSquares += values[i] * values[i];
+  }
+  return Math.sqrt(sumOfSquares);
+};
```
asturur, line 16: do we ever need the  3 coordinate cases in fabricJS?
if no, can we please reduce this to `return Math.sqrt(x * x + y * y)` ?

src/point.class.ts
ShaMan123: > i m not sure we should refer a vector and a point like if they are the same thing

I agree.
Maybe I refactor Point and extract Vector in a follow up

src/shapes/polyline.class.ts
ShaMan123: We discussed it as well, the mutation.
Let's revert points.
`setDimensions` is exposed for the dev

src/shapes/polyline.class.ts
ShaMan123: good point

src/util/misc/boundingBoxFromPoints.ts
ShaMan123: I was throwing if array was empty or if points were just objects

src/util/misc/hypot.ts
ShaMan123: This is here because of the chrome impl of Math.hypot that causes needless headache.
So yes it is needed

src/util/misc/boundingBoxFromPoints.ts
ShaMan123: Not sure we should go this way because I think (not sure) that TS will bug us in case of an array that you filtered or sliced or whatever since it won't know that surely it has a first element. You will need to cast it and that's annoying.
But this is not a TS thing only. Should cover all cases.

src/point.class.ts
asturur: no just call it a point, and jsut write somewhere that this represent the point distance from the origin, OR the magnitude of a vector that has origin on 0,0 and direction expressed by point p.
We don't need to refactor, just a different wording.

src/shapes/polyline.class.ts
asturur: either you revert everything and leave skewX and skewY, or you leave as it is.
I do really prefer only skewX and skewY that are in user hands.

The rest is not forbidden, but is a different feature, and can be introduced with some sort of planning.
Text has something similar right?

propertiesThatChangesWidthHeightInText = [...array of stuff]

same we do with properties that invalidate caching

I'm not saying is wrong, but i would like to introduce it on its own, with a dedicated moment of planning

src/point.class.ts
ShaMan123: What about moving it as a util yo vectors?
I prefer that

asturur:
So first of @luizzappa thanks for this long and continuous effort to bring this to the finish line.

Is all good for me, some overdoing i commented around but not blockers ( just think if it makes sense to simplify, to me it does ).
I don't have time and strenght those days to look at calculations i and i checked just the general changes to the classes.
Is optional calculation and is good.

I also like the idea of dividing the projectors in classes, so is easier to digest them.

A suggestion maybe you have looked into.
I do see you want to take in account skewX and skewY and that to me seems dangerous/unreliable.

A polygon can be at skew 0 anywhere, then just rotated a little bit, then put in a group that gets scaled X, and so the polygon get induced skewing and all our dreams of precisions are lost.
Precision is not THAT important.
We want to have a bounding box that is not clearly broken, and we can stop there, both for easyness of calculation and move on anyway.
tens of lines of code for nitty picky edge cases are not welcome.

On this topic, since i can't get through all the history now, i have some questions/comments

- did you consider apply the current transformation to each point and then evaluate the bounding box after that? in that case skewX and skewY are reduced to 0 always ( take all points, multiply them for the full transform matrix from group down, and apply bbox calculation there ). Would be it simpler in some way? Would it be equivalent at all?

- this code can be safely applied to paths too since we do have the code that makes curves segments with the purpose of measuring their distance. yes is an approximation but should detect spikes anyway

src/util/misc/hypot.ts
asturur: it is needed if you use it outside the bidimensional use case.
Otherwise the reduced 2d formula is equivalent since we aren't a polyfill library and we don't offere a precise math.hypot for nth dimension space.

My question indeed was, do we ever need the 3 or plus dimensions in hypotenuse?

test/unit/polygon.js
```diff
@@ -78,13 +78,67 @@
...
+
+    polygon.set('strokeLineJoin', 'round');
+    const round = polygon._getNonTransformedDimensions();
+    assert.equal(Math.round(round.x), 70, 'round x');
+    assert.equal(Math.round(round.y), 150, 'round y');
+    assert.deepEqual(polygon._getTransformedDimensions(), round, 'dims should match');
   });
 
+  QUnit.todo('polygon with exactBoundingBox true and skew', function (assert) {
+    var polygon = new fabric.Polygon([{ x: 10, y: 10 }, { x: 20, y: 10 }, { x: 20, y: 100 }], {
```
asturur, line 109: do we want to make this polygon larger than 3 points so that we can be sure all code is run properly? it would be nice to have a concave polygon with a spike that from the inside pierce the boundingbox so that we are sure we are covering strange cases.

asturur: this apply to the above polygin too.

luizzappa:
> Ready including line cap

Good job! I opened a review, but it's not showing up here..

I'll copy it here:

https://github.com/fabricjs/fabric.js/blob/f4de0291c53441c4087749ff3fcf6274473bdc78/src/util/misc/projectStroke/StrokeLineCapProjections.ts#L78-L94

We don't need those 4 projections:

image

Just the ones that point to the side of the square is enough.

line cap square

Remove the first two:

```diff
- this.A.add(orthogonalProjection),
- this.A.subtract(orthogonalProjection),
```

The rest is correct, for me it's good to go

luizzappa:
> I also like the idea of dividing the projectors in classes, so is easier to digest them.

The refactoring of @ShaMan123 turned out great!

> A polygon can be at skew 0 anywhere, then just rotated a little bit, then put in a group that gets scaled X, and so the polygon get induced skewing and all our dreams of precisions are lost.

I'm not very familiar with how fabricjs groups work, but if you "individually" calculate the projections for each object within a group I can't see this as a problem.




> ^^^ no this would work visually, but wouldn't wor to find the correct transformation before.
> Or maybe i can transform all points, **then i can take the for resulting corners**, and transform them back with the inverse?
> i m curious to know if that absorbs skewX/skewY and give equivalent results.

I don't know if I understood very well. In all cases just applying skew to the projections works, the current problem is only when `strokeLineJoin` is `round` and `skewX` and `skewY` are applied at the same time. The geometric figure that was previously known (a semi-circle), becomes just a curve segment (calculating the corner of this segment which is complicated. I thought it was an ellipse, but it is not.. [solving the equation](https://github.com/fabricjs/fabric.js/pull/8344/#issuecomment-1276978199) that I commented here I believe we found these corners)


image

Anyway, I believe we can continue in this PR without it. I'll think a little more about this problem, and if I find a solution, I'll open a new PR.

ShaMan123:
1. Precision:
> Precision is not THAT important.

I **absolutely** disagree. The whole point of a bounding box is precision, the whole point of this PR and similar efforts.
Product-wise is seems very unprofessional as viewed by a user (myself included).

2. `hypot`: 
I removed it, though I disagree as well. 
If hypot on chrome is wrong and fabric suffers from it why not expose hypot for devs?
Since I managed to get the code working without it I'll back off and let lint pick it up in the future.

4. Tests: we will follow up with visual tests on all cases in a separate PR. Too many files here already.
5. skewing and `qrDecompose`:
@asturur  your comment regarding this has made things a bit more clear to me. I now mistrust qrDecompose. I will raise this in the relevant threads.

src/util/misc/hypot.ts
ShaMan123: removed

src/shapes/polyline.class.ts
ShaMan123: > either you revert everything and leave skewX and skewY, or you leave as it is.

Not good enough accounting only for skew. I will introduce an array of props.

ShaMan123: 301276fe

src/shapes/polyline.class.ts
ShaMan123: actually not.
since object has default values of `{ top: 0, left: 0}` we must pass undefined

ShaMan123: This is not a real change.
`_setPositionDimensions` used this logic but was written differently depending on `fromSVG` instead.

ShaMan123: Once `exactBoundingBox=true` by default `fromSVG` can be removed as well. That's why I changed it.

ShaMan123:
We shall test this work in nested groups as well 
We can use the great demo @luizzappa built as a visual test, making projection very clear to the eye.

ShaMan123 (approved):
@luizzappa @asturur see the final commits

ShaMan123 (dismissed):
https://codesandbox.io/s/project-stroke-points-with-skew-correction-forked-xe09oo

can't wait for codesandbox action to work! Then no more of this copy paste bullshit

ShaMan123:
@luizzappa indicate all is good so I can merge with peace

ShaMan123:
@asturur surfacing this comment so you won't get surprised.
We found a bug in skewing controls. This PR aggravates it. Following up with a fix (perhaps w/o `qrDecompose`)


> Meanwhile I have managed to fix skewing with controls. I will PR soon so no worries about that. [master...fix-skewing-controls](https://github.com/fabricjs/fabric.js/compare/master...fix-skewing-controls)

ShaMan123:
@asturur any objections we make polygon controls part of core using the demo @luizzappa created?

luizzappa:
Everything seems to be correct. Thanks for all the support @ShaMan123 !

Regarding the tests, is it better to open a new PR (or are we still in https://github.com/fabricjs/fabric.js/pull/8087)?

ShaMan123:
My pleasure.
A new PR and a dedicated test file

src/shapes/polyline.class.ts
asturur: i see the planning part wasn't ok.

src/shapes/polyline.class.ts
asturur: yes this is hacky this is the exact part i wanted to give a second look at.

src/util/misc/hypot.ts
asturur: I want to use this example to point out a comunication issue
I ask you if between the calculation of projections, the 3+ dimension of hypot are used, if not to use the 2d version that is shorter.
You tell me that chrome has a suboptimal implementation, but that is written exactly over that file.
So i ask you gain the same thing, and you answer me with `removed`.
What is the process that brough you to the decision to remove the util? what are we using in place of the util now?

ShaMan123: The 2d version impl of Point.distanceFrom as we should have done from the beginning.

src/shapes/polyline.class.ts
```diff
@@ -41,11 +44,29 @@ import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
        * @deprecated
        * @type Boolean
        * @default false
+       * @todo set default to true and remove flag and related logic
```
asturur, line 47: we could in theory move the default to true, but not remove the flag totally unless we can benchmark that for large paths/polyline this code has zero performance impact

src/util/misc/hypot.ts
ShaMan123: I looked at it again. Gave another thought. Decided to hear your point out and look again with fresh eyes. That is the process. Then I remrmbered distanceFrom.

src/shapes/polyline.class.ts
ShaMan123: Not in this pr. It will be a nightmare adjusting tests

src/shapes/polyline.class.ts
ShaMan123: Suggestions?

asturur: i m not entirely sure. let me craft some weird SVG on purpose.
i do think that if you import from SVG we have to do something different and solve it in the svg import function.

src/util/misc/hypot.ts
ShaMan123: To be clear I marked it as removed since I wrote about it in the review

src/shapes/polyline.class.ts
ShaMan123: Agreed

ShaMan123: I have an idea
Is this the only use case to set position in the constructor?
If so we move that logic to from svg aftet initialization

src/util/misc/hypot.ts
ShaMan123: My bad for sure.

> You tell me that chrome has a suboptimal implementation, but that is written exactly over that file.

A lot of files and PRs so I don't remember it all.

src/shapes/polyline.class.ts
ShaMan123: Could we use `pathOffset` for that?
```js
setPositionByOrigin(poly.pathOffset, 'center', 'center')
```

ShaMan123: Do you mind we do that in the follow up?

ShaMan123: Then we add dedicated tests for svg cases as well

asturur:
> 1. Precision:
> 
> > Precision is not THAT important.
> 
> I **absolutely** disagree. The whole point of a bounding box is precision, the whole point of this PR and similar efforts. Product-wise is seems very unprofessional as viewed by a user (myself included).
> 

It is ok to disagree. Solutions and costs needs to be balanced.
If the bouding box is a bit thight on roundy corner on edgy use cases, i m super happy to guess a constant factor and just make the bounding box a bit larger to accomodate that.
When choosing between  a bunch of code and an accomodation, ultimate precision isn't that much important.

> 2. `hypot`:
>    I removed it, though I disagree as well.
>    If hypot on chrome is wrong and fabric suffers from it why not expose hypot for devs?
>    Since I managed to get the code working without it I'll back off and let lint pick it up in the future.

I just asked you to expose a leaner version of it since here there are no cases of 3+ dimensional distances. Right?
Add code when you need  it, and not because someone could one day come here and need it.


Said so the bounding box calculation in this way breaks ( obviously ) the skewing bounding box from controls that at least in the sandbox jump left and right often.
That is to take in account as a possible defect of this feature and we can't do much about it, and that is fine, not everyine is skewing all the time.

For the example of the skew i made before, it should also be taken in account that for roundy corner, and overall different solution could be to trace the path that the strokes create, approximate (basically is an exact match) the roundy part with a beizer curve, and the split it in segments with the measurement code, and then apply the transformations to those points, that would give you more points in total, but would solve the issue under skew or at least it would approximate good enough the skewed rounded corner.

It is already possible to transform a path with a canvas transform and obtain a new path that absorbs the transformation and move all its points and control points to look like identical to the transformed path ( at leat from my tests, i can't back this up with a theorem )

src/shapes/polyline.class.ts
asturur: no i m saying in the future maybe we can switch it to true, but the flag will probably stay.

ShaMan123:
> Said so the bounding box calculation in this way breaks ( obviously ) the skewing bounding box from controls that at least in the sandbox jump left and right often.

https://github.com/fabricjs/fabric.js/pull/8344#issuecomment-1279799561

> Add code when you need it, and not because someone could one day come here and need it.

I was being greedy. 

I suggested other options to calculate the bbox. We could explore them later on.

src/shapes/polyline.class.ts
ShaMan123: > we could in theory move the default to true, but not remove the flag totally unless we can benchmark that for large paths/polyline this code has zero performance impact

This is part of my difficulty in communication.
I read this and understand you want it changed since you didn't explicitly write later, future whatever.

src/shapes/polyline.class.ts
ShaMan123: I looked at this again I saw my comments are wrong as well.
No need to pass undefined as you said since the code checks if passed options have left, top **not** default value.

Too many open PRs so they leak in my mind to each other. I did something in a PR and then I work on 3 more and I start assuming wrong things.

src/shapes/polyline.class.ts
```diff
@@ -79,56 +108,104 @@ import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
...
-       * @return {Object} object.left X coordinate of the polygon leftmost point
-       * @return {Object} object.top Y coordinate of the polygon topmost point
-       * @return {Object} object.width distance between X coordinates of the polygon leftmost and rightmost point
-       * @return {Object} object.height distance between Y coordinates of the polygon topmost and bottommost point
+       * @returns {Point} top left position of the bounding box, useful for complementary positioning
+       */
+      setDimensions: function () {
+        const { left, top, width, height, pathOffset, strokeOffset } =
+          this._calcDimensions();
+        this.set({ width, height, pathOffset, strokeOffset });
```
asturur, line 159: i kind of get that `strokeOffset` is the value that tells us in which direction the bounding box move most of its new space, becuase that was always one of the things that this kind of new calculation was adding, the center of the shape with and without stroke is no more the same and this strokeOffset kind of represent that.
But where are we using it? I can't see use it in the PR nor in the codebase.

ShaMan123: Trash my comments entirely in favor of 
f158267

ShaMan123 (approved):

asturur:
> Regarding polygon controls. The community will appreciate that as well. There are many issues open asking to fix that. I heard chat of making it part of core. @asturur ? v6 has changed some internals that broke it.

We should definitely add polygon/path controls after 6.0 is released.
Not bound to the object automatically, but definitely available and supported

src/shapes/polyline.class.ts
asturur: github mixes comments sometimes ( it does it to me ).
Indeed this answer is for the other thread.
strokeOffset still a mistery? or your answer was related and i don't see how?

asturur:
i need to craft a polygon with a translate transform, that i think gets absorbed with top and left, let me do that.

src/shapes/polyline.class.ts
ShaMan123: Related to the other thread indeed.
Can't find it anymore. Too much chat.

Don't understand your questions

asturur:
Added the test in master, if it pass you can merge.
I ll probably come back on that automatic update on _set

src/shapes/polyline.class.ts
asturur: i ll try easier, at the cost of looking unpolite.

Show me where strokeOffset is used.
i see it calculated, not used.

asturur (approved):
This PR is good.
There are some things i don't like on the generic changes in polygon, and i will come back to them:
- new special _set: for the  polygon and the bounding box
- setDimension is new but has 2 task, setDimensions and return me the top,left and that is weird. Would be probably less weird if it would return all it calculated, and then we extract what we need from it.
- boundingBox early return should go away if is only a compiler issue and not a real code issue that fabricJS could trigger ( seems also double protected in the polyline class too.

src/shapes/polyline.class.ts
ShaMan123: As I wrote in the thread it is meant for the dev.
Think of using originX/Y in a case of a large miter for example. The poly might shift a lot because of this miter. So a dev might want to adjust the position so the origin is in respect to the points instead of the entire stoked miter

asturur: ok so i guessed right what is it, and is unused.
I'll think of it.

PR #8346, opened 2022-10-03
ci(): changelog action + merge directive

ShaMan123:
The changelog action is failing due to an unknown reason on #8135 , see [failing job](https://github.com/fabricjs/fabric.js/actions/runs/3173069251/jobs/5168265853) and [CHANGELOG](https://github.com/fabricjs/fabric.js/pull/8135/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed)

This PR removes the action and adds a simple diff command to check if the file has changed instead of the action

In addition added `.gitattribute` instructing git to do a merge union in case of conflicts in CHANGLOG.


.github/workflows/build.yml
```diff
@@ -56,15 +56,15 @@ jobs:
   changelog:
     if: ${{ always() }}
```
ShaMan123, line 57: not sure why this is needed

ShaMan123 (commented):
READY

ShaMan123:
@asturur what if we disable the lint action for now?
So we can see the green tick mark of all passing and then reinstate it in the end of the migration

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# https://stackoverflow.com/a/13264117/9068029
+CHANGELOG.md merge=union
```
ShaMan123, line 2: This is a cool thing instructing git to do a merge union in case of conflicts

ShaMan123 (commented):
ready

.gitattributes
ShaMan123: tested it in this PR when I ran update from master and indeed it fast forwarded and merged the file as expected

asturur:
Why i don't see action starting here?

asturur (approved):

PR #8365, opened 2022-10-14
include textBackgroundColor check in hasStyleChanged

jakesjews:
Prevent an undefined exception in stylesToArray when converting style objects which only contain a textBackgroundColor property

### DIFF
--- a/test/unit/text.js
+++ b/test/unit/text.js
@@ -394,6 +394,13 @@
     assert.deepEqual(obj.styles, [], 'empty style object has been removed');
   });
 
+  QUnit.test('text toObject can handle style objects with only a textBackgroundColor property', function(assert) {
+    var text = new fabric.Text('xxx');
+    text.styles = { 0: { 0: { textBackgroundColor: 'blue' } } };
+    var obj = text.toObject();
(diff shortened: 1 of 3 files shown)


test/unit/text.js
```diff
@@ -394,6 +394,13 @@
     assert.deepEqual(obj.styles, [], 'empty style object has been removed');
   });
 
+  QUnit.test('text toObject can handle style objects with only a textBackgroundColor property', function(assert) {
+    var text = new fabric.Text('xxx');
+    text.styles = { 0: { 0: { textBackgroundColor: 'blue' } } };
+    var obj = text.toObject();
+    assert.deepEqual(obj.styles, [{ start: 0, end: 1, style: { textBackgroundColor: 'blue' }}], 'empty style object has been removed');
```
melchiar, line 401: Thanks for the PR!

Just a couple things. This message should be updated though to something that fits the test. Also, PRs now require adding an entry to the changelog (https://github.com/fabricjs/fabric.js/blob/master/CHANGELOG.md). Everything else looks good to me!

jakesjews: whoops! totally missed that I'll get a fix up for that

melchiar (approved):
Looks good!

PR #8372, opened 2022-10-15
feat(): SideEffect

ShaMan123:
I will use this PR to propose a change and discuss it since it is easier to write code and comment in PRs.

## Gist
- class `SideEffect`
- register/unregister a side effect on object/canvas/whatever level
- flush side effects in the `_set` method and in other relevant places
- should side effects run in initialization?

## Motivation
- history (undo/redo) discussed by the team - the dev could register as many side effects they want and trigger custom logic
- replacing stateful #7920 
- the need to run side effects and block them at runtime
- a cleaner way to write side effects - move things like `dimensionsAffectingProps` under a class that contains the logic as well, all under one roof. Easier to read, develop and debug.
- something similar to events - you can register/unregister a side effect, control it, create more. Eliminates the need to subclass, Easier to control behavior and easier to understand I hope.
- https://github.com/fabricjs/fabric.js/pull/8344#pullrequestreview-1143233737 and others regarding what side effects should or shouldn't run by default.
- use instead of events in cases in which we don't want `off` to deregister the event and kill the side effect.
- perf: using `set({ ... })` invokes side effects for each key. That is suboptimal and can be addressed using side effects.
- since using `set({ ... })` invokes side effects for each key I can think of an edge case in which passing the same object with keys ordered differently will result in a different effect

## Comments
I think this could be part of v6. No rush. As I wrote, this is a discussion that might turn into a PR.

## Usage
```ts

class Polyline {
  constructor(...) {
     this.registerSideEffect(
          new SideEffect(
            'StrokeBBoxSideEffect', 
            ['strokeWidth', 'points', ...], 
           () => this.setDimensions()
      ))
  }

  flushSideEffects(key, value) {
    this.sideEffects.forEach(effect => effect.invoke(key, value))
  }
}

const lazyPolyline = new Polyline(...);
lazyPolyline.set('strokeWidth', 10) // sets dimensions
lazyPolyline.getSideEffect('StrokeBBoxSideEffect').disable()
lazyPolyline.set('strokeWidth', 20) // does nothing
lazyPolyline.unregisterSideEffect('StrokeBBoxSideEffect')
lazyPolyline.set('strokeWidth', 30) // does nothing for good

const shouldRunDimensionsEffect = !lazyPolyline.getSideEffect('StrokeBBoxSideEffect').isEqual({ strokewidth: 10, strokeUniform: false, ... })
```

### DIFF
--- a/src/SideEffect.ts
+++ b/src/SideEffect.ts
@@ -1,0 +1,10 @@
+export abstract class SideEffect<T, K extends keyof T> {
+  abstract name: string;
+  abstract keys: K[] | '*';
+  abstract onChange(key: K, value: T[K], prevValue: T[K]): void;
+  invoke(key: K, value: T[K], prevValue: T[K]) {
+    (this.keys === '*' || this.keys.includes(key)) &&
+      value !== prevValue &&
+      this.onChange(key, value, prevValue);
+  }
+}
--- a/src/SideEffect.ts
+++ b/src/SideEffect.ts
@@ -1,9 +1,13 @@
 export abstract class SideEffect<T, K extends keyof T> {
   abstract name: string;
   abstract keys: K[] | '*';
+  enable = true;
+
   abstract onChange(key: K, value: T[K], prevValue: T[K]): void;
+
   invoke(key: K, value: T[K], prevValue: T[K]) {
-    (this.keys === '*' || this.keys.includes(key)) &&
+    this.enable &&
+      (this.keys === '*' || this.keys.includes(key)) &&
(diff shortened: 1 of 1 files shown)


ShaMan123:
I propose that `Object.set` returns an object/class instance:
```ts

class EffectsFlusher {
  constructor(effects: SideEffect[])

  // invokes side effects with options passed from set
  flush(options) {
     // dismisses side effects that should run only once
  }

  // did the set operation change the instance
  changed(): boolean

}

```

PR #8374, opened 2022-10-16
fix(#8344): stroke projection

luizzappa:
Added tests for https://github.com/fabricjs/fabric.js/pull/8344

Tests revealed bugs and edge cases that this PR addresses.

The goldens make reading the diff impossible since github boils. View the diff here instead https://github.com/fabricjs/fabric.js/compare/8374-view-diff


luizzappa:
The tests are generating around 3,000 images. It makes sense? Or is it better to define some sets of points beforehand, and just apply `scale`, `skew` and `miterLimit`?

luizzappa:
It seems to me that the goldens were generated by node (which has a problem with miter-limit)

image

ShaMan123:
If we can test all cases with very little tests that is the best

luizzappa:
I reduced the amount of tests to be more targeted. I think it's better this way.

Goldens look like they are being generated by node instead of browser, due to miter-limit issue:

acuteAngle-polygon-20skewX-0skewY

luizzappa:
Tests reveal a bug when the `strokeLineJoin` is `bevel` or `miter` and has a straight angle

image


The problem seems to me is that in these cases we only project "outwards", this works for almost all points arrangements, but in the straight angle a straight line is formed, so we need projections to both sides.

image

If we check if the angle bisector is equal to 2PI, and in those cases project to both sides, we fix that. (not tested yet)

ShaMan123:
> If we check if the bisector angle is equal to PI, and in those cases project to both sides, we fix that. (not tested yet)

Maybe we use line cap projections in the case?

ShaMan123:
> Goldens look like they are being generated by node instead of browser, due to miter-limit issue:

What cmd are you using?

ShaMan123:
image

make sure you run with `-- -c chrome`

Resize the goldens. White space is very bad since it is counted as pixels, making the diff% inaccurate.

ShaMan123:
I see that is stroke line cap tests the stroke line join is wrong, I am correct?
Does it come from a big in tests?

luizzappa:
> Maybe we use line cap projections in the case?

Yes, conceptually it makes even more sense.

These changes here fix this bug:

https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeLineJoinProjections.ts#L213-L222

```diff
      case 'miter':
-        return this.projectMiter();
+       return this.bisectorAngle === 0 
+          ? new StrokeLineCapProjections(this.A, this.B, this.options).projectButt()
+          : this.projectMiter();
```

```diff
      case 'bevel':
-        return this.projectMiter();
+       return this.bisectorAngle === 0 
+          ? new StrokeLineCapProjections(this.A, this.B, this.options).projectButt()
+          : this.projectBevel();
```

luizzappa:
> I see that is stroke line cap tests the stroke line join is wrong, I am correct? Does it come from a bug in tests?

yes, i made a mistake in the code

ShaMan123:
I need to understand the cases for the straight line. 
It seems to me that this bug can occur only on the edges (first/last point) because a point in the middle that forms a straight line with its surrounding points shouldn't project anything. 
Thoughts?

ShaMan123:
So if I am correct it is an edge case of a polygon with 2 points?

ShaMan123:
I have been thinking to expose a method on polyline, and now this makes even more sense:
```js
isOpen() {
   return (true/false for polyline/polygon) && this.points.length > 2

}
```

Then polygon become a very small subclass, only changing `isOpen`

luizzappa:
> I need to understand the cases for the straight line. It seems to me that this bug can occur only on the edges (first/last point) because a point in the middle that forms a straight line with its surrounding points shouldn't project anything. Thoughts?

You're right

> So if I am correct it is an edge case of a polygon with 2 points?

Not necessarily, there may be a straight angle with more points, for example:

image

points = [{x: 6, y: 6}, 
{x: 1, y: 1}, 
{x: 6, y: 6}, 
{x: 36, y: 36}, 
{x: -42.711504510665065, y: 173.59483211415397}, 
{x: -63.264587151680985, y: 180.70936072065956}, 
{x: -84.77119408675864, y: 187.3414936778875}, 
{x: 143, y: 171}, 
{x: 10.289783513792408, y: 10.149937092784299}]

ShaMan123:
Good point
I suggest change `projectStroke/index`

ShaMan123:
What happens when the `bisectorAngle % Math.PI` is very close to 0?

ShaMan123:
> I suggest changing `projectStroke/index`

Trash that

ShaMan123:
So I think the problem is with `getAcuteAngleFactor`. Should it return 0 in this case?

ShaMan123:
Thinking here...
If the `bisectorAngle % Math.PI` === 0 then the angle between the bisector and the side is `Math.PI / 2`.
This is the case in which we should project to both sides.
So changing `getAcuteAngleFactor` to return `-Math.sign(Math.acos(angle))` seems to do the job.
Then we need to modify the code that uses `getAcuteAngleFactor`
Thoughts?

ShaMan123:
This seems correct for `projectRoundNoSkew` as well

luizzappa:
> If the `bisectorAngle % Math.PI` === 0 then the angle between the bisector and the side is `Math.PI / 2`.
> This is the case in which we should project to both sides.

In fact, we must project to both sides when `bisectorAngle` is equal to zero. See the image below, the dashed vectors are the `bisectors`, and the filled vectors are the `projections`. The middle point whose `bisectorAngle` is equal to `Math.PI` doesn't come into play. What matters are the two points on the sides that have the `bisectorAngle` equal to zero.

image

> So changing `getAcuteAngleFactor` to return `-Math.sign(Math.acos(angle))` seems to do the job

It's a more elegant way of doing what already happens today:
https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeProjectionsBase.ts#L24

> This seems correct for `projectRoundNoSkew` as well

As in the `round` case we use projections in the direction of the X and Y axis, it does not suffer from this problem.

straightangle line joins

luizzappa:
When the `bisectorAngle` is equal to zero we'll have to return two projections, but I see two ways it can be done:

**First way**:  inside `projectMiter` and `projectBevel`:

https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeLineJoinProjections.ts#L118

```diff
- return [this.applySkew(this.A.add(miterVector))];
+ return this.bisectorAngle === 0
+   ? [this.applySkew(this.A.add(miterVector)), this.applySkew(this.A.subtract(miterVector))]
+   : [this.applySkew(this.A.add(miterVector))]
```

Something similar here:

https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeLineJoinProjections.ts#L87

**Second way**:
 
> These changes here fix this bug:
> 
> https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeLineJoinProjections.ts#L213-L222
> 
> ```diff
>       case 'miter':
> -        return this.projectMiter();
> +       return this.bisectorAngle === 0 
> +          ? new StrokeLineCapProjections(this.A, this.B, this.options).projectButt()
> +          : this.projectMiter();
> ```
> 
> ```diff
>       case 'bevel':
> -        return this.projectMiter();
> +       return this.bisectorAngle === 0 
> +          ? new StrokeLineCapProjections(this.A, this.B, this.options).projectButt()
> +          : this.projectBevel();
> ```

luizzappa:
Initially I was preferring the second way. However, thinking about the first way we encapsulate all the miter/bevel projection calculation logic in a single function, in the second way it is separated in two places.

luizzappa:
Or a **third way**:

add this clause inside the `projectBevel`, `projectMiter` (it's better than the second way, because it encapsulates everything in a single function)

```
if (this.bisectorAngle === 0) return new StrokeLineCapProjections(this.A, this.B, this.options).projectButt();
```

https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeLineJoinProjections.ts#L97


https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/StrokeLineJoinProjections.ts#L86

ShaMan123:
I would like it to be:
```ts
  projectBevel() {
    const projections: Point[] = [];
    [this.B, this.C].forEach((to) => {
      projections.push(
        this.projectOrthogonally(this.A, to, this.strokeProjectionMagnitude)
      );
      // in case the angle is 0 we project to both sides
      this.bisectorAngle === 0 &&
        projections.push(
          this.projectOrthogonally(this.A, to, -this.strokeProjectionMagnitude)
        );
    });
    return projections;
  }
```

To my understanding there is no need to apply on miter since when the angle is 0 the limit is infinity.

luizzappa:
As `AB` and `AC` are on the same line when `bisectorAngle` is equal to `zero`, we can consider only one of them.

```
  projectBevel() {
    const projections: Point[] = [],
              sides = this.bisectorAngle === 0 ? [this.B] : [this.B, this.C];
    sides.forEach((to) => {
      projections.push(
        this.projectOrthogonally(this.A, to, this.strokeProjectionMagnitude)
      );
      // in case the angle is 0 we project to both sides
      this.bisectorAngle === 0 &&
        projections.push(
          this.projectOrthogonally(this.A, to, -this.strokeProjectionMagnitude)
        );
    });
    return projections;
  }
```

ShaMan123:
I thought of that but then said to myself what if B === A. Then the bisector will be 0.
So I decided to leave as is but rethinking it, though AB will not project anything would ABC project what is needed?
Could you add a test for that case?
I will push another test for nested poly under group.

ShaMan123:
> But if we fix handling the zero vector it should work.

9c00f88

ShaMan123:
added a test for group e20b2ec
take a look
something is wrong. might be that poly should account for the total skewing value not only its own

ShaMan123:
updated the demo on #8344

ShaMan123:
[Group demo](https://codesandbox.io/s/project-stroke-points-with-skew-correction-forked-bueyte?file=/src/index.js)
Needs work

https://user-images.githubusercontent.com/34343793/196602669-202cc7cb-d17c-4250-b80e-bd31d527c3da.mp4

ShaMan123:
updated the demo
now respects group, but still more adjustments to do

ShaMan123:
This is another one
It is hard to keep track of changes.
Maybe we should work on a branch/PR so we can discuss diffs
https://codesandbox.io/s/project-stroke-points-with-skew-correction-forked-qpk67d?file=/src/index.js:2626-2771

ShaMan123:
I think we need to pass the total skewing of an object or better off we use the raw transform
```js
qrDecompose(target.calcTransformMatrix())
```
The problem with the decomposition is `skewY === 0`

luizzappa:
> I thought of that but then said to myself what if B === A. Then the bisector will be 0. So I decided to leave as is but rethinking it, though AB will not project anything would ABC project what is needed? Since it will still be affected by the 0 vector I don't think so. But if we fix handling the zero vector it should work. Could you add a test for that case? I will push another test for nested poly under group.

It seems to me that nothing is rendered when two points are equal.

Nothing rendered with:
```
[
    { x: 100, y: 100 },
    { x: 100, y: 100 },
    { x: 200, y: 100 },
    { x: 210, y: 110 }
  ]
```

It's only when I make the points slightly different that something is rendered:
```
  [
    { x: 100, y: 100 },
    { x: 100.1, y: 100 },
    { x: 200, y: 100 },
    { x: 210, y: 110 }
  ]
```
image

luizzappa:
> This is another one, tried triggering `setDimensions` from with group but it didn't seem to do anything. It is hard to keep track of changes. Maybe we should work on a branch/PR so we can discuss diffs https://codesandbox.io/s/project-stroke-points-with-skew-correction-forked-qpk67d?file=/src/index.js:2626-2771

Let's work on a new PR

luizzappa:
> I think we need to pass the total skewing of an object or better off we use the raw transform
> 
> ```js
> qrDecompose(target.calcTransformMatrix())
> 
> // or
> 
> target.calcTransformMatrix()
> ```
> 
> The problem with the decomposition is `skewY === 0`
> 
> For example projecting along the axis must be wrong if an object is contained in a transformed plane

Hmmm `calcTransformMatrix()` does not seem to me to bring the transformation of the plane.

image

Wouldn't I have to bring the skew in the object's transformation matrix? 
Or am I confusing something?

ShaMan123:
> Let's work on a new PR

OK I suggest you work locally on a branch that merged `sandbox-ci` and then checkout the changes to the PR branch.
That way you could start a dev server with the sandbox live and running.
That is what I do.

luizzappa:
> I thought of that but then said to myself what if B === A. Then the bisector will be 0.
> So I decided to leave as is but rethinking it, though AB will not project anything would ABC project what is needed?
> Since it will still be affected by the 0 vector I don't think so. But if we fix handling the zero vector it should work.
> Could you add a test for that case?
> I will push another test for nested poly under group.



> > It seems to me that nothing is rendered when two points are equal.
> 
> Usually means there is NaN pollution.

You were right. When two vertices are equal, the `calcOrhogonalProjection()` returns NaN.

```
    calcOrthogonalProjection(from, to, magnitude = this.strokeProjectionMagnitude) {
        const vector = this.createSideVector(from, to);
        const orthogonalProjection = getOrthonormalVector(vector);
        const correctSide = StrokeProjectionsBase.getAcuteAngleFactor(orthogonalProjection, this.bisectorVector);

        return this.scaleUnitVector(orthogonalProjection, magnitude * correctSide);
    }
```
As `from` === `to`, `vector` is equal to `{x: 0, y: 0}`, so `getOrthonormalVector()` returns `NaN` (and it really makes sense, it's not possible to compute the orthogonal to a point, there are infinite solutions)

I added this situation in the test case https://github.com/fabricjs/fabric.js/pull/8374/commits/604a6bccee1f92a8b09381ceab9b7455b1290320

How can we handle this? I thought of something like this but it's buggy with `miter`
```
    projectBevel() {
        return [this.B, this.C].map((to) => 
            this.A.x === to.x && this.A.y === to.y 
            ? this.A 
            : this.projectOrthogonally(this.A, to)
        );
    }
```

For example:
```
points = [
    { x: 100, y: 100 },
    { x: 100, y: 120 },
    { x: 100, y: 120 },
    { x: 210, y: 110 }
  ]
```
The stroke formed by the equal points (marked in red), is considered as a `bevel` if we compare the points three by three. But the canvas plots as if it were `miter`:

image

It suspect that the canvas abstracts the equal points and plots the strokes as if it were:
```
points = [
    { x: 100, y: 100 },
    { x: 100, y: 120 },
    { x: 210, y: 110 }
  ]
```

If I'm right, we can get around this right at the beginning of the `projectStrokeOnPoints()`, if two adjacent points are equal, we consider it as a single point. I will try tomorrow
https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/util/misc/projectStroke/index.ts#L13-L23

ShaMan123:
- zero vector:
NaN originates from `getUnitVector => magnitude`. The zero vector has magnitude of 0 (da!). The thing is that `0/0` (`x/magnitude`) equals NaN. 
However what should be the return value `getUnitVector(zeroVector)`? should it return the zero vector or NaN vector as it does now?

- skipping equal points sound good to me

src/util/misc/projectStroke/index.ts
```diff
@@ -18,47 +18,55 @@ export const projectStrokeOnPoints = (
...
-          ).project()
-        );
-      } else {
-        projections.push(
-          ...new StrokeLineJoinProjections(A, B, C, options).project()
-        );
-      }
-    });
+  if (reduced.length === 1) {
+    openPath = true;
```
ShaMan123, line 36: seems that pathOffset is miscalculated in this case

ShaMan123 (commented):
Please review
- 1 point case
- reduced points

luizzappa:
> Please review
> 
> * 1 point case

I think that when we have only one point, nothing is rendered... And it doesn't seem to be a problem with NaN, as was the case with two equal points. I did a test drawing directly on the canvas:

```
ctx = canvas.getContext("2d");
ctx.beginPath();
ctx.moveTo(100, 100);
ctx.strokeStyle = "red";
ctx.lineWidth = 10;
ctx.lineTo(100, 100);
ctx.stroke();
```

Nothing is rendered, at least 1 pixel is required for something to be rendered: `ctx.lineTo(100+1, 100+1);`


Specifically about fabricjs, when there is only one point, the `lineTo` is in the same position as the `moveTo`, so nothing is rendered following the above mentioned canvas behavior:

https://github.com/fabricjs/fabric.js/blob/e0eed4581914912aa058d7bf5a7b590d7cbb6f4d/src/shapes/polyline.class.ts#L256-L274

luizzappa:
> * However what should be the return value `getUnitVector(zeroVector)`? IMO it return the zero vector [730394e](https://github.com/fabricjs/fabric.js/commit/730394e9cb5bbed23aa0be0ef19fa329312cc33c)

I have doubts if it is better to handle this case directly in this function. Could it not end up generating a silent error for other uses? For example, here we discovered this problem because it returned a NaN and was not rendering anything. If the unit vector of a zero vector is being computed, there is some problem at the origin that needs to be addressed that may not have the same treatment that we are giving here..


I would treat this problem only in this context, maybe a method in `StrokeProjectionsBase`:
` _getUnitVector = (v: Point): Point => v.eq(new Point()) ? v : getUnitVector(v)`

What do you think?

luizzappa:
Actually I think its' better to treat this here:

```
  calcOrthogonalProjection(
    from: Point,
    to: Point,
    magnitude: number = this.strokeProjectionMagnitude
  ) {
    const vector = this.createSideVector(from, to);
    const orthogonalProjection = getOrthonormalVector(vector);
    const correctSide = StrokeLineJoinProjections.getOrthogonalRotationFactor(
      orthogonalProjection,
      this.bisector
    );
    return this.scaleUnitVector(orthogonalProjection, magnitude * correctSide);
  }
```

```diff
    const vector = this.createSideVector(from, to);
+   if (vector.eq(zeroVector)) return zeroVector;
```

src/util/misc/projectStroke/StrokeLineCapProjections.ts
```diff
@@ -76,20 +76,31 @@ export class StrokeLineCapProjections extends StrokeProjectionsBase {
...
-      projectedA.subtract(orthogonalProjection),
-    ].map((p) => this.applySkew(p));
+    const projections: Point[] = [];
+
+    if (this.A.eq(this.T)) {
+      const projection = new Point(1, 1).scalarMultiply(
+        this.strokeProjectionMagnitude
+      );
+      projections.push(this.A.add(projection), this.A.subtract(projection));
+    } else {
```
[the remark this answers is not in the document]
ShaMan123: sounds correct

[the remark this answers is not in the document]
ShaMan123: removed

src/util/internals/findRight.ts
```diff
@@ -0,0 +1,19 @@
...
+  return -1;
+};
+
+export const findRight = <T>(
+  array: T[],
+  predicate: (value: T, index: number, array: T[]) => boolean
+) => {
+  const index = findIndexRight(array, predicate);
+  return index > -1 ? array[index] : undefined;
+};
```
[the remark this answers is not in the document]
ShaMan123: done

ShaMan123:
```diff
    const vector = this.createSideVector(from, to);
+   if (vector.eq(zeroVector)) return zeroVector;
```

I am OK with this but I still think `getUnitVector(zeroVector) === zeroVector` is better than a NaN point. I terms of debugging they are quite similar, NaN is more destructive, I agree.

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -97,7 +97,18 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
...
    */
   projectBevel() {
-    return [this.B, this.C].map((to) => this.projectOrthogonally(this.A, to));
+    const projections: Point[] = [];
+    [this.B, this.C].map((to) => {
+      projections.push(
+          this.projectOrthogonally(this.A, to)
+      );
+      // if `alpha` equals 0, we need to project for both sides.
+      this.alpha === 0 &&
```
ShaMan123, line 106: > Is this part still needed?

Returning a favor
Since we reduce points it is correct to go with your suggestion and run this logic outside the loop

ShaMan123:
Regarding one point rendering:

https://codesandbox.io/s/fabric-vanillajs-sandbox-forked-x9oroo?file=/src/index.ts

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
luizzappa: makes sense

luizzappa:
> ```diff
>     const vector = this.createSideVector(from, to);
> +   if (vector.eq(zeroVector)) return zeroVector;
> ```
> 
> I am OK with this but I still think `getUnitVector(zeroVector) === zeroVector` is better than a NaN point. I terms of debugging they are quite similar, NaN is more destructive, I agree.

I don't have a strong opinion on this. If you think it's better, keep it that way.

luizzappa:
> Regarding one point rendering:
> 
> https://codesandbox.io/s/fabric-vanillajs-sandbox-forked-x9oroo?file=/src/index.ts
> 
> from the [spec](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-linecap-dev) (found it now when it is useless) image
> 
> _Originally posted by @ShaMan123 in [#8344 (comment)](https://github.com/fabricjs/fabric.js/pull/8344#issuecomment-1279727195)_
> 
> I can't find what the spec says.


OK, now I got it. The problem then is only when lineCap = butt that nothing is rendered. 
I'm going to refactor the code to allow passing options of what we want to test for each caseToTest, so in this specific case we can test lineCap for both polyline and polygon

luizzappa:
I separated the singlePoint from what I called generalCasesToTest (formerly casesToTest) https://github.com/fabricjs/fabric.js/pull/8374/commits/59c7ac1b3717566016741b94f1c37370d6eb8dd5

luizzappa:
image

luizzappa:
I thought reduced would solve the single point problem, but I was wrong.
image

I returned with your commit that dealt with this for square's lineCap case: https://github.com/fabricjs/fabric.js/pull/8374/commits/abb7a2eadd9cb02ea7cb2c78530ae9a5355331d8

I added the round lineCap case:
https://github.com/fabricjs/fabric.js/pull/8374/commits/93c0b887814ee6d6d7e0b9a39d677a47b04abd37

luizzappa:
I encountered a problem in calculating the `strokeMiterLimit` calculation when `strokeUniform = true`. The calculation just with the angle (`hypotUnitScalar`) was giving divergent values, I went back to the calculation with `miterVectorLength/(strokeWidth/2)`

```diff
-      ? hypotUnitScalar
+      ? magnitude(
+           this.scaleUnitVector(this.bisector, this.options.strokeMiterLimit)
+        )
```

It seems to me that it is now correct. However, if you use a chromium browser, it seems to me that there is a bug in the browser itself of calculating the `strokeMiterLimit` for large values..

For example, look at this example I made: https://codesandbox.io/s/project-stroke-points-with-skew-correction-forked-b69sln?file=/src/index.js

Open in **Chrome** Version 106.0.5249.119 (Official version) 64 bits (Windows 10):

image

Open in **Firefox** 102.3.0esr (64-bit) (Windows 10):

image

Mathematically the `strokeMiterLimit` was not reached, so a miterVector is projected, however in chrome it transforms (incorrectly) to bevel. In the case of firefox it correctly maintains the miter.

ShaMan123:
I have seen this as well
I am guessing that since we discovered chrome implements hypot in an abnormal way this is the result. Internally, they must use hypot when drawing to canvas.
Not our concern.

ShaMan123:
Well actually this means that if we use `Math.hypot` this should solve the issue

luizzappa:
> Well actually this means that if we use `Math.hypot` this should solve the issue

I tested here, it had no effect..

luizzappa:
I encountered a problem when the points are too close together. Every vertex must have a projection orthogonal to it too (even for `bevel` and `round` strokeLineJoin). The figures look very weird, but it's behavior..

image

For the `round` case, I found another problem when the points are too close. As I project to both sides, projecting to the opposite side ends up messing up the calculation. Basically I have to keep only the projections to the outside.

image

image

luizzappa:
I have already fixed locally, I will organize and commit.

luizzappa:
Inside `vectors.ts`, I created the `isBetweenVectors` function to return if a vector is between two other vectors. It would be nice to add unit test to it, can I include it in `test/unit/util.js` or do I insert it somewhere else?

ShaMan123:
That is fine, place it next to other vector tests.
The test suite is a mess. We will get to that after the major migration is done.

luizzappa:
> I encountered a problem when the points are too close together. Every vertex must have a projection orthogonal to it too (even for `bevel` and `round` strokeLineJoin). The figures look very weird, but it's behavior.. https://codesandbox.io/s/bug-orthogonal-p0cnlx?file=/src/index.js
> 
> image
> 
> For the `round` case, I found another problem when the points are too close. As I project to both sides, projecting to the opposite side ends up messing up the calculation. Basically I have to keep only the projections to the outside.
> 
> image
> 
> image

fixed at https://github.com/fabricjs/fabric.js/pull/8374/commits/415f9192a37be15ae36c4c9a91cd6dd7a45f3498

https://codesandbox.io/s/fix-bug-orthogonal-8c8lm5?file=/src/fabricjs.js

image

image

ShaMan123:
Good job!!

ShaMan123:
Did you add these cases to the visual tests?

ShaMan123:
https://github.com/fabricjs/fabric.js/commit/74eaaa0b40278faad6dcd7b4dde3203ca4c1f505#r87923533

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -98,19 +108,16 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
    */
   projectBevel() {
     const projections: Point[] = [];
-    [this.B, this.C].map((to) => {
+    // if `alpha` equals 0 or 2*PI, the projections are the same for `B` and `C`
+    (this.alpha === 0 || this.alpha === twoMathPi
```
ShaMan123, line 112: `this.alpha % twoMathPi === 0`?

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -215,22 +228,53 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
...
+    // when `alpha` equals 0 or 2*PI, we have a straight line, so the way to find the start/end is different.
+    const isStraightLine = this.alpha === 0 || this.alpha === twoMathPi,
+      // change the origin of the projections to point A
+      // so that the cross product calculation is correct
+      newOrigin = this.applySkew(this.A),
+      proj0 = projections[isStraightLine ? 0 : 2].subtract(newOrigin),
+      proj1 = projections[isStraightLine ? 1 : 0].subtract(newOrigin),
+      // when `isStraightLine` === true, we compare with the vector opposite AB, otherwise we compare with the bisector.
+      comparsionVector = isStraightLine
+        ? this.applySkew(this.AB.scalarMultiply(-1))
```
ShaMan123, line 263: Is this safe in case AB = 0?

src/shapes/polyline.class.ts
```diff
@@ -144,8 +144,10 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
           top: bbox.top - legacyCorrection,
           pathOffset: new Point(pathOffsetX, pathOffsetY),
           strokeOffset: new Point(bboxNoStroke.left, bboxNoStroke.top).subtract(
-            bbox.left,
-            bbox.top
+            new Point(bbox.left, bbox.top)
```
ShaMan123, line 174: caught error

src/shapes/polyline.class.ts
```diff
@@ -144,8 +144,10 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
...
+          ),
+          strokeDiff: new Point(bbox.width, bbox.height).subtract(
```
ShaMan123, line 176: this will be used in skewing I control I believe

src/shapes/polyline.class.ts
```diff
@@ -185,16 +212,25 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
...
-              strokeWidth: 0,
-              // disable skewing bbox calculations
-              skewX: 0,
-              skewY: 0,
-            })
-          : this.callSuper('_getTransformedDimensions', options);
+        if (this.exactBoundingBox) {
+          // calculate size
+          const { width, height } =
+            options && !(options.width && options.height)
```
ShaMan123, line 218: `_getTransformedDimensions` is used by the skewing control logic so it needs to calculate size properly
However passing width/height is not supported unless we apply scaling after projecting the points. What do you think?

ShaMan123 (commented):
- Updated from master
- fix `_getTransformedDimensions` except when passing width/height

src/shapes/polyline.class.ts
ShaMan123: we should add a unit test for this

src/shapes/polyline.class.ts
ShaMan123: we should unit tests for `_getTransformedDimensions`

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
luizzappa: I think the only case that AB = 0 is when we have 1 point (which is handled in strokeLineCap). All others, reducedPoints will eliminate. am I wrong?

ShaMan123: Correct!

src/shapes/polyline.class.ts
ShaMan123: It is actually silly
No one should pass width height to a function that calculates them
I got confused

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
luizzappa: nice!

src/shapes/polyline.class.ts
luizzappa: added here https://github.com/fabricjs/fabric.js/pull/8374/commits/b895215350215d0904a0febf9c248bfea8af0833, but i don't know if it's what you have in mind

test/unit/polygon.js
```diff
@@ -254,4 +254,45 @@
...
+    assert.ok(pathOffset instanceof fabric.Point);
+    assert.ok(strokeOffset instanceof fabric.Point);
+
+    // Values
+    assert.equal(left, 10.485714075442775);
+    assert.equal(top, 14.784917784669414);
+    assert.equal(width, 27.707709196083425);
+    assert.equal(height, 21.750672506349947);
+    assert.deepEqual(pathOffset, new fabric.Point(14.999999999999998, 17.000000000000004));
+    assert.deepEqual(strokeOffset, new fabric.Point(-0.48571407544277534, -2.784917784669414));
```
ShaMan123, line 295: This doesn't make sense, does it? I would expect it to be values > 0

The test is perfect

ShaMan123 (commented):
Regarding testem....
What can I say?
You can dig in their bugs and try to find the problem...
Might be some chrome missing launch arg.
I have no idea.
What happens if you run with `--dev` flag?

ShaMan123:
Actually I remember that this is happens when an error is thrown but testem doesn't surface it, rather it tries to handle or something and report it.
I thing there's an option to override it.
Try running it in node context to eliminate that is the problem because node at least reports the correct error.
What a broken testing experience

ShaMan123:
Running in node reveals the test never ends
It is endless

ShaMan123:
How many cases are there?
I have reached 2000 assertions before killing the process

luizzappa:
> What happens if you run with `--dev` flag?

Same thing..

> How many cases are there? I have reached 2000 assertions before killing the process

There are 2.268 test cases

> Running in node reveals the test never ends It is endless So there is a bug somewhere in the test file

On node and firefox it goes to the end, only on Chrome it gives timeout.

luizzappa:
Using --dev flag revealed a few more things.. There seems to be a memory leak in chrome:

`Erro code: SBOX_FATAL_MEMORY_EXCEEDED`

Tracking the memory allocation, it seems to me that the rendered canvas for each test case is not freed from memory, as time goes by you can see that the shallow size of the CanvasRenderingContext and HTMLCanvasElement only increase in size. As we have more than 2000 tests, there comes a time when the memory overflows (here it always overflows on test number 1270 - lambdaLetter of type polyline with strokeLineCap=round, strokeUniform=false, skewX=0, skewY=0)

memory leak

Could you try to run all cases in chrome? It could be something on my machine

test/unit/polygon.js
luizzappa: We're calculating the difference when strokeWidth=0 (and no scale, no skew) versus when we could have strokeWidth and scale/skew. In theory it sounds like it should be positive. 

This test has a peculiarity that it is 2 points, and when strokeWidth = 0 I don't really know what happens. I have to think about it.

luizzappa:
Just to note here as a TODO in PR to handle groups. It has a very weird bug, when we select objects, and then deselect them, the BBOX is totally messed up. Is it something related to qrDecompose? It looks like it loses skewY, and and compensates with angle and skewX.

https://codesandbox.io/s/bug-group-ketuln


bug group

test/unit/polygon.js
luizzappa: It can happen that there is strokeOffset < 0 when skew is applied. The logic goes something like this:

image

A real case: https://codesandbox.io/s/strokeoffset-mqqxdq

**p2 -> skewY = 0**

`p2._calcDimensions()`

{ strokeOffset: 
  type: "point",
  x: 6.035533905932738,
  y: 2.5
}

**p3 -> skewY = 45**

`p3._calcDimensions()`

{ strokeOffset: 
  type: "point",
  x: 6.035533905932738,
  y: -1.4644660940672622
}

ShaMan123:
- It makes sense that memory overflows now that I think of it. We keep 3 canvases for each test.... We could try to adjust this.... To triage disable this function https://github.com/fabricjs/fabric.js/blob/master/test/lib/visualCallbackQunit.js#L35
I have a simple idea that wasn't done before since this behavior goes back before the test suite to the website. We could use the goldens server and write all outputs to disk to `cli_output`. It's not a bad idea.
- qrDecompose is to blame for group bbox discrepancy AFAIK, you are correct. Though I have an idea for this as well. Since skewY is not preserved in decomposition why don't we post multiply the transform with the inverse shearingY matrix, decompose and then use the skewY value with the decomposition?
- Should offsetY be 20-16?
image

Regardless,  I can get how offset with skewing produces negative offset but I don't get why the

test/unit/polygon.js
ShaMan123: I got confused here as well
Thought it was `strokeDiff` that has negative values.
Sorry for the ruckus

ShaMan123: We should add that to the test as well

luizzappa: add here https://github.com/fabricjs/fabric.js/pull/8374/commits/9079fffc54fef8d4066151916f8258f63931afcc

luizzappa:
> * It makes sense that memory overflows now that I think of it. We keep 3 canvases for each test.... We could try to adjust this.... To triage disable this function https://github.com/fabricjs/fabric.js/blob/master/test/lib/visualCallbackQunit.js#L35
>   I have a simple idea that wasn't done before since this behavior goes back before the test suite to the website. We could use the goldens server and write all outputs to disk to `cli_output`. It's not a bad idea.

I'll take a look

> * qrDecompose is to blame for group bbox discrepancy AFAIK, you are correct. Though I have an idea for this as well. Since skewY is not preserved in decomposition why don't we post multiply the transform with the inverse shearingY matrix, decompose and then use the skewY value with the decomposition?

I still have a very superficial view of it, I need to familiarize myself more. I think working in PR to fix the groups will help me. 

> * Shouldn't offsetY be 20-16?
>   image


The way it was implemented is what I mentioned:

`strokeOffset.y = Y point (without strokeWidth, scale, skew) - Y point (with strokeWidth, scale, skew) = 10 (first image) - 16`

image

https://github.com/fabricjs/fabric.js/blob/52c5611098dfcb647958db88d64b0a3a53ca1de0/src/shapes/polyline.class.ts#L130-L131

https://github.com/fabricjs/fabric.js/blob/52c5611098dfcb647958db88d64b0a3a53ca1de0/src/shapes/polyline.class.ts#L146-L148

`bboxNoStroke` it's the bbox with no border, no scale and no skew. These are the original points (`this.points`).

If the intention when exposing this context is the `strokeOffset` of the current figure (which would be 20-16), you would have to multiply `this.points` by the transformation matrix of the object before calling makeBoudingBoxFromPoints for `bboxNoStroke`

ShaMan123:
> If the intention when exposing this context is the strokeOffset of the current figure (which would be 20-16), you would have to multiply this.points by the transformation matrix of the object before calling makeBoudingBoxFromPoints for bboxNoStroke

Thinking of the use case, I think that is the intension. Since the use case would be the adjust position by the offset value.
Thoughts?
Good catch!

Regarding working on group. What do you have in mind?
Look at #8337

ShaMan123:
@luizzappa try https://github.com/fabricjs/fabric.js/pull/8311 to see if tests don't fail
I changed it so it doesn't store canvases in memory any longer

ShaMan123:
If that also doesn't work we should give up trying to fix it and use #8402 
That will output all failed visuals to a downloadable link

luizzappa:
> Thinking of the use case, I think that is the intention. Since the use case would be to adjust position by the offset value.

https://github.com/fabricjs/fabric.js/pull/8374/commits/e6d780e0c4f1549b81e24cb10243b1e302fc5cbc

> Regarding working on group. What do you have in mind?
> Look at #8337

Finishing this PR I start working on groups. I believe there are 3 main tasks:
- `setDimensions` not working on groups
- The problem of selecting/grouping objects messes up the BBOX (which seems to be `qrDecompose`)
- Tests revealed a bug in the BBOX even with a single object being inside a group (I still haven't stopped to investigate). Example:

orthogonalProjection4-uniform--25skewX-35skewY-grouped

Anything else that should be addressed?

luizzappa:
> @luizzappa try #8311 to see if tests don't fail I changed it so it doesn't store canvases in memory any longer

Great, I'll check it out tomorrow!

src/shapes/polyline.class.ts
```diff
@@ -154,28 +152,34 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
...
-          offsetX - offsetY * Math.tan(degreesToRadians(this.skewX));
-        const pathOffsetY =
-          offsetY - pathOffsetX * Math.tan(degreesToRadians(this.skewY));
-        // TODO: remove next line
-        const legacyCorrection =
-          !this.fromSVG && !this.exactBoundingBox ? this.strokeWidth / 2 : 0;
+        const bbox = makeBoundingBoxFromPoints(points),
+          // Remove scale effect, since it's applied after
+          matrix = calcDimensionsMatrix(
+            Object.assign(options, { scaleX: 1, scaleY: 1 })
```
ShaMan123, line 158: Is this intentional?
It mutates options
```suggestion
            { ...options, scaleX: 1, scaleY: 1 }
```

ShaMan123:
Since I wrote Group I can point out where to look and what could be wrong so I can save time.
Group is a layout mechanism so my guess is that during the layout pass (`_applyStrategyResult`) setDimensions needs to be called.

ShaMan123:
https://github.com/fabricjs/fabric.js/blob/d19b7fd58f121149ea66b05f86e0c39687f61602/src/shapes/group.class.ts#L888

ShaMan123:
OK wait
Another question
The projections are relative t the canvas plane, right?
That is why inside group it is wrong.
This isn't good...
Let's triage first.
If we decompose the full transform matrix and pass that to project points that should do.

src/shapes/polyline.class.ts
luizzappa: no, I made a mistake. Thanks!

luizzappa:
> @luizzappa try #8311 to see if tests don't fail
> I changed it so it doesn't store canvases in memory any longer

It worked! 2,268 tests completed without running out of memory

luizzappa:
> The projections are relative to the canvas plane, right?

yes


I didn't quite understand what happens when we add an object to a group. For example, this polyline

```
var p2 = new fabric.Polygon(
  [
    {x:79.63237629347668, y:-41.260831452943876},
    {x:84.9774418788902, y:-48.28812548381641},
    {x:85.52113122175491, y:-50.718528761996765}
  ],
  {
    scaleX: 2,
    scaleY: 3,
    skewX: 30,
    skewY: 45,
    exactBoundingBox: true
  }
);
```

Inside `_calcDimensions` has the `options`:
```
_calDimensions options = { 
scaleX: 1,
scaleY: 1,
skewX: 30,
skewY: 45
}
```

But the simple fact of including in a group:

`g = new fabric.group([p2])`

`options` inside `_calcDimensions` becomes:

luizzappa:
Disregard my previous answer. I deleted

> The projections are relative to the canvas plane, right?

yes

ShaMan123:
Once an object enter Group it is transformed relative to Group center in Group plane coords
Meaning that the object in not aware of the transforms made to group.
It can be by using calcTransformMatrix,

ShaMan123:
updated my comment

luizzappa:
I think I understand the `qrDecompose` "problem" a little better. The video commented here (https://github.com/fabricjs/fabric.js/pull/8337) is very enlightening. 

Correct me if I'm wrong..
When we rotate an object (with the `angle` property), the transformation matrix scale/skew "slots" are used to reflect this.

`Transform matrix = [ scaleX, skewY, skewX, scaleY, offsetX, offsetY ]`

 `qrDecompose` allows me to go back to the original `scaleX`, `scaleY`, `skewX` and `angle` values. The "problem" is that we always lose reference to what the original `skewY` was.

So if `angle = 0`, I can ignore qrDecompose, and directly use the values ​​of the transformation matrix slots ( `[ scaleX, skewY, skewX, scaleY, offsetX, offsetY ]` ). That way you wouldn't lose the original `skewY`.

However, if the `angle` is non-zero, we totally lose the original reference (whether it was a rotation or a "pure skewY") and `qrDecompose` will always consider it as a rotation.

luizzappa:
I'm thinking here.. If we calculated the inverse of the rotation matrix (inverse of `calcRotateMatrix`) and multiplied by the transformation matrix, we would eliminate the angle effect, and then we could extract the `scale`/`skew` properties directly from the transformation matrix (just like when `angle` = 0)

luizzappa:
> Just to note here as a TODO in PR to handle groups. It has a very weird bug, when we select objects, and then deselect them, the BBOX is totally messed up. Is it something related to qrDecompose? It looks like it loses skewY, and and compensates with angle and skewX.
> 
> https://codesandbox.io/s/bug-group-ketuln
> 
> bug group [ bug group ](https://user-images.githubusercontent.com/65685842/198643882-0f2877b9-2bd0-4f48-aaf0-db82dea1cd30.gif) [ ](https://user-images.githubusercontent.com/65685842/198643882-0f2877b9-2bd0-4f48-aaf0-db82dea1cd30.gif)


(and yes, the problem I mentioned in the group's BBOX is when there is `skewX` and `skewY` applied. qrDecompose loses the skewY reference and considers it as a rotation, it messes everything up)

ShaMan123:
Once #8417 is merged I will update this branch
And then we finalize this PR and merge
I am thinking we should utilize stroke projection for Path as well.

luizzappa:
> I am thinking we should utilize stroke projection for Path as well.

That's make sense

As the fix of the `group` will not be fast, I removed it from the tests for now. https://github.com/fabricjs/fabric.js/pull/8374/commits/f8c012e295b3cb3124a4f023b36f74e06ec886c7 

Maybe we have to rethink these tests. In some shapes the golden will be a way for firefox, chrome and node because of implementations/bugs in the engines themselves.. This will make the test fail

Another thing, there are some images that are blurred:

**uniform-twoPointsClose-5miterLimit-25skewX-35skewY**
uniform-twoPointsClose-5miterLimit-25skewX-35skewY

Thought this would solve: `fabric.Object.prototype.noScaleCache = false;` , but not.. Any ideia what this could be?

ShaMan123:
I can't find that test.
All uniform tests I looked at are off. Could it be that scaling the viewport messes up things?
I suggest adding the control logic you wrote so we can see the projection vectors

ShaMan123:
`noScaleCache` affects caching only while scaling with the controls

test/visual/stroke_projection.js
```diff
@@ -300,7 +300,7 @@ QUnit.module('stroke projection', (hooks) => {
             },
             golden: `stroke-projection/${
               isPolygon ? 'strokeLineJoin' : 'strokeLineCap'
-            }/${testCase.name}/${strokeLineCap}${
+            }/${testCase.name}-${strokeLineCap}${
```
ShaMan123, line 303: I patched this since tests were failing.
I think it might be best to open a folder for each test.
#8311 fixes the option for creating directories on the fly or should at least

luizzappa:
Visual tests revealed a bug when there is only 1 point and `strokeUniform` is true. Scaling has no effect in this situation, so we have to remove its effect on the projection. Fixed at https://github.com/fabricjs/fabric.js/pull/8374/commits/8264ce973723888d776888e52c8faf0e144bb8a5

luizzappa:
> (probably blurring as well, try `objectCaching: false`)?

That solved it. The object had this property, but the blue rectangle behind it did not...

luizzappa:
Visual tests revealed a bug with `strokeLineJoin = round`, `strokeUniform = true`, and with `skew`.  Fixed at: https://github.com/fabricjs/fabric.js/pull/8374/commits/6cad304b2fd0725ecd9ac87c776c6db96b072f62

**orthogonalProjection1-uniform--0skewX-30skewY-**
orthogonalProjection1-uniform--0skewX-30skewY-

Initially the calculation, when `strokeUniform = true`, removed the scale effect on the circle radius. But in reality we must remove the scale effect directly in the calculation of the farthest points. For some cases this calculation difference was barely visible, so it looked like it was correct, but in some edge cases the error appeared (like the one in the picture above).

After fix:

orthogonalProjection1-uniform--0skewX-30skewY

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -199,32 +199,37 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
...
+          scaleY /
+          Math.sqrt(1 / scaleY ** 2 + (1 / scaleX ** 2) * shearing.y ** 2)
+        : circleRadius / Math.sqrt(1 + shearing.y ** 2),
       furthestY = new Point(
-        Math.sqrt(
-          circleRadius.x ** 2 - ((newY * circleRadius.x) / circleRadius.y) ** 2
-        ),
+        // Safe guard due to floating point precision. In some situations the square root
+        // was returning NaN because of a negative number close to zero.
+        Math.sqrt(circleRadius ** 2 - newY ** 2) || 0,
```
ShaMan123, line 222: is this safeguard good enough?
`Math.max` perhaps?

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -199,32 +199,37 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
...
         newY
       ),
-      newX =
-        circleRadius.x / Math.sqrt(1 + Math.tan(degreesToRadians(skewX)) ** 2),
+      newX = strokeUniform
+        ? circleRadius /
+          scaleX /
+          Math.sqrt(1 / scaleX ** 2 + (1 / scaleY ** 2) * shearing.x ** 2)
```
ShaMan123, line 228: Not straight forward,
Maybe add an explanation?

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
```diff
@@ -199,32 +199,37 @@ export class StrokeLineJoinProjections extends StrokeProjectionsBase {
...
-    const circleRadius = new Point()
-        .scalarAdd(this.strokeProjectionMagnitude)
-        .multiply(this.strokeUniformScalar),
-      newY =
-        circleRadius.y / Math.sqrt(1 + Math.tan(degreesToRadians(skewY)) ** 2),
+    const circleRadius = this.strokeProjectionMagnitude,
+      newY = strokeUniform
+        ? circleRadius /
+          scaleY /
+          Math.sqrt(1 / scaleY ** 2 + (1 / scaleX ** 2) * shearing.y ** 2)
```
ShaMan123, line 217: same

ShaMan123 (commented):
Are we at the finish line?
Since meanwhile we migrated Polys I will integrate that into this PR.

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
luizzappa: I think so. `Math.max(NaN, 0)` and `Math.Min(NaN, 0)` return `NaN` in both cases.

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
luizzappa: I found it more straightforward to add in the documentation. I just did: https://github.com/fabricjs/fabric.js/pull/8344#2-3-2-round-skew

image

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
ShaMan123: I meant inside the sqrt

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
ShaMan123: Excellent

src/util/misc/projectStroke/StrokeLineJoinProjections.ts
luizzappa: Do you think it's safer? I have no idea..

ShaMan123: If nan is caused due to sqrt of a negative number then sure, it will solve iy entirely

ShaMan123: Anyways there is a util ifNaN

luizzappa:
There was a bug in the case of `strokeLineCap = square` when there was a `scale` effect and uniform stroke (basically, when creating the vector, forgot to consider removing the scale effect). I fixed here: https://github.com/fabricjs/fabric.js/pull/8374/commits/309357b058f742c6ca34eb9aa1b080e1e91a1bbe

Here it looks small, but if we increase the difference between `scaleX` and `scaleY` it becomes more visible.

**acuteAngle-uniform--0skewX-0skewY-**
Before:

acuteAngle-uniform--0skewX-0skewY-

After:

acuteAngle-uniform--0skewX-0skewY-

ShaMan123:
Excellent 
That was what I was seeing.
The next thing I what to discuss is _getTransformedDimensions.
Should it accept width/height in options. Group uses it internally, so that majes me think for the sake of conforming to that. But it really seems silly.

luizzappa:
> Are we at the finish line?
> Since meanwhile we migrated Polys I will integrate that into this PR.

I have individually checked all goldens and they are correct. Tests pass in firefox and chrome. Only on node, there are still some tests not passing.

I believe it is a problem with the implementation of node. But the differences are not being saved in the cli_ouput folder for me to confirm that this is it. I'm using this here: https://github.com/fabricjs/fabric.js/pull/8311

image

luizzappa:
> Excellent That was what I was seeing. The next thing I what to discuss is _getTransformedDimensions. Should it accept width/height in options? Group uses it internally, so that makes me think for the sake of conforming to that, should we add some logic to polys? But it really seems silly.


You're thinking about it here:

```
      _getTransformedDimensions: function (options) {
        if (this.exactBoundingBox) {
          // calculate size
          const { width, height } =
            options && !(options.width && options.height)
              ? this._calcDimensions(options)
              : { width: this.width, height: this.height };
          return this.callSuper('_getTransformedDimensions', {
            width,
            height,
            ...(options || {}),
            // disable stroke bbox calculations
            strokeWidth: 0,
            // disable skewing bbox calculations
            skewX: 0,
            skewY: 0,
          });
        } else {
          return this.callSuper('_getTransformedDimensions', options);
        }
      },
```

If I'm not mistaken, the problem I've seen is that `_getTransformedDimensions` is invoked multiple times in the canvas rendering process, this would cause the projections to be calculated countless times unnecessarily.

I think that makes it more performative:

https://github.com/fabricjs/fabric.js/blob/1d58727a2deb7b55e4c2ec9872dd81c70449375a/src/shapes/polyline.class.ts#L183-L194


This brings me back to another question...

The function `_getNonTransformedDimensions` should return width/height without any transformation (scale and skew), right?

The way we currently implement `setDimensions` changes the width/height, then  `_getNonTransformedDimensions` no longer returns untransformed dimensions. 

https://github.com/fabricjs/fabric.js/blob/1d58727a2deb7b55e4c2ec9872dd81c70449375a/src/shapes/polyline.class.ts#L171-L175


Am I right?

luizzappa:
https://github.com/fabricjs/fabric.js/pull/8374/commits/a3fdbb1cedf6722b287c3d48776bf52b3ad97cfe fixed the round case with `skewX` and `skewY` at the same time. 

https://github.com/fabricjs/fabric.js/pull/8374/commits/bcef2a5fab20163c5bc93fa1a946d3d319b971f6 added test cases for that

Documentation: https://github.com/fabricjs/fabric.js/pull/8344#2-3-2-round-skew

luizzappa:
Well I think we implement a solution that correctly calculates BBOX for all possible cases. I have individually checked all goldens and they are correct.

What's pending in this PR:

- Test in node is failing. I imagine it's because of that implementation problem with the miter limit. But the diff images are not being saved to cli_output so I can't tell what it really is.

- `_getTransformedDimensions()` --> I don't think it's a good idea to invoke dimensioning in this function.. It can cause a pretty big performance problem as any interaction with the canvas will call this function for all canvas objects unnecessarily.

https://github.com/fabricjs/fabric.js/blob/bcef2a5fab20163c5bc93fa1a946d3d319b971f6/src/shapes/polyline.class.ts#L216-L223

- `_getNonTransformedDimensions()` --> I think we need to make a change here as we change the width and height of the object by recalculating the dimensions.

https://github.com/fabricjs/fabric.js/blob/bcef2a5fab20163c5bc93fa1a946d3d319b971f6/src/shapes/polyline.class.ts#L204-L208

Something like:

```
      _getNonTransformedDimensions: function () {
        if (this.exactBoundingBox) {
                const { width, height } = this._calcDimensions({ scaleX: 0, scaleY: 0, skewX:0, skewY:0})
                return new Point(this.width, this.height)
        }
        return this.callSuper('_getNonTransformedDimensions');
      },
```

ShaMan123:
- we can disable node tests
```ts
tests.push({
  ...,
disabled: fabric.isLikelyNode 
})
```
- `_getNonTransformedDimensions` is used only to render an object's background. We will probably remove it since it is confusing.
- I agree that there is no need to recalculate stroke projection for both `_getNonTransformedDimensions` and `_getTransformedDimensions`, however still I am unsure about some of it. I will think about it.
- This leads me to a question posed by @asturur: Is it possible, now that the projection is stable to extract skewing from the calculation?

ShaMan123:
We have 1 failing test (might be just a change in accuracy)

And controls:

https://user-images.githubusercontent.com/34343793/202153234-22d66f54-a8c4-4906-81c6-4eaa5d3cf226.mp4

ShaMan123:
The problem is that now `_getTransformedDimensions` is wrong since passed options are disregarded

ShaMan123:
I would argue that since `_getNonTransformedDimensions` is used to obtain the size of the shape in the transformed plane we just need to reverse scaling and skewing on the output point

ShaMan123:
Could you checkout unrelated files?

src/shapes/polyline.class.ts
```diff
@@ -202,17 +216,30 @@ export class Polyline extends FabricObject {
...
+        size = new Point(options.width ?? width, options.height ?? height);
+      } else {
+        size = new Point(
+          options.width ?? this.width,
+          options.height ?? this.height
+        );
+      }
+      return size.multiply(
+        new Point(options.scaleX || this.scaleX, options.scaleY || this.scaleY)
+      );
```
luizzappa, line 240: Is this right? If the `strokeUniform = true` we shouldn't we add `strokeWidth`?

luizzappa: Is it not safer to call `_getTransformedDimensions` with the options? 
That way, any change in logic there, we would take advantage of here too.

luizzappa:
> * This leads me to a question posed in the past by @asturur: Is it possible, now that the projection is stable to extract skewing from the calculation?

What do you mean by extracting skew from the calculation? Remove the skew effect as we do with scale in the case of strokeUniform?

I didn't quite understand, if you can provide more context or an example..

luizzappa:
> I would argue that since `_getNonTransformedDimensions` is used to obtain the size of the shape in the transformed plane we just need to reverse scaling and skewing on the output point

Something link: `return new Point(this.width/this.scaleX, this.height/this.scaleY)` ? IMO isn't enough, for example the `scale` changes the `miter length` when `strokeUniform = true`, so the `width`/`height` is not directly proportional to the scale.

> _getNonTransformedDimensions is used only to render an object's background/eraser. We will probably remove it since it is confusing. But this flags a divergence from the rest of shapes in behavior. We need to fix that.

Making a `console.log` inside `_getNonTransformedDimensions` I couldn't see it being invoked in the canvas rendering. Because of this, it seems to me that calling `projectStrokeOnPoints` inside it is not as problematic as calling inside `_getTransformedDimensions` (which is invoked every repaint)

I would go this way:
```
      _getNonTransformedDimensions: function () {
        if (this.exactBoundingBox) {
                const { width, height } = this._calcDimensions({ scaleX: 1, scaleY: 1, skewX: 0, skewY: 0})
                return new Point(width, height)
        }
        return this.callSuper('_getNonTransformedDimensions');
      },
```

src/shapes/polyline.class.ts
```diff
@@ -255,8 +255,7 @@ export class Polyline extends FabricObject {
       changed &&
       (((key === 'scaleX' || key === 'scaleY') &&
         this.strokeUniform &&
-        this.strokeBBoxAffectingProperties.includes('strokeUniform') &&
-        this.strokeLineJoin !== 'round') ||
```
luizzappa, line 208: I noticed that the changes made to the projection of the `round` case made it also necessary to recalculate the projection when `strokeUniform = true`

luizzappa: fixed at https://github.com/fabricjs/fabric.js/pull/8374/commits/0810190790ff202b8ad228f6e200068550b11e42

ShaMan123:
> I don't understand why it wasn't reversed. Can I manually delete?

Yes

I wonder if the revert commit will override the merge commit when we try to merge that PR, I think not since we are squashing. No worries, I'll take care of that....

If you set `backgroundColor` on a poly `_getNonTransformedDimensions` will be called in each rendering cycle.

> What do you mean by extracting skew from the calculation? Remove the skew effect as we do with scale in the case of strokeUniform?

Yes, that's the idea I guess. But it doesn't seem feasible.

Adding more context such as `resizingOffset` is a good option IMO.

luizzappa:
> If you set `backgroundColor` on a poly `_getNonTransformedDimensions` will be called in each rendering cycle.

Hmm wasn't aware of that.. But I think it needs fixing when we apply skew. For example, skewX = 45:

image

This should ignore the skew and consider the dimensions of the BBOX..

luizzappa:
> > What do you mean by extracting skew from the calculation? Remove the skew effect as we do with scale in the case of strokeUniform?
> 
> Yes, that's the idea I guess. But it doesn't seem feasible.

The `skew` is directly related to each of the control point positions, so it is not possible to find a **single value** to remove the `skew` from the entire object. What you could do is a function that iterates over each of the control points, removes the `skew`, and then applies it again when necessary (no need to recalculate the projections, just apply the new `skew` value). I'm pretty sure that's possible

luizzappa:
> Adding more context such as `resizingOffset` is a good option IMO.

I started to implement this logic, but other questions came up that we need to define..

For example, if I create a polygon with `width` and `height` defined, what should happen?

I think the best solution would be to proportionally change the position of the control points to match the defined width/height. 

For example, 
```
new fabric.Polygon(
    [{x: 0, y:10}, {x: 20, y:30} ],
    width: 40,
    height: 50
)
```

Width with just the control points: `20 - 0 = 20`
Height with just the control points: `30 - 10 = 20`

```
delta width = 40/20 = 2
delta height = 50/20 = 2.5
```

The points should undergo the following transformation:
```
[
 {  x: 0 * 2 = 0, 
    y: 10 * 2.5 = 25 }, 
 {  x: 20 * 2 = 40, 
    y: 30 * 2.5 = 75 } 
]
```

So: 

final width = `40 - 0 = 40`
final height = `75 - 25 = 50`

Sound reasonable?

ShaMan123:
I like both of your suggestions
- if it is possible to extract skewing and reapply it in a post calculation sounds good... though I recall you said skewing affects the arrangement of points and the miter limit so I don't understand how that can be achieved.
- scaling the polygon to a given width and height sound correct. We don't to alter the points, only the scale, right?
- the thing with `getNonTransformedDimensions` is that it runs during the rendering process **after** the canvas context has been transformed.

ShaMan123:
In general maybe we should wrap this up and merge it.
It is annoying working with the diff, waiting for all the goldens to load.
Thoughts?

src/shapes/polyline.class.ts
ShaMan123: I get your point and normally would agree but feels like the divergence of the calculation is so big we should detach from super

src/shapes/polyline.class.ts
```diff
@@ -158,9 +193,9 @@ export class Polyline extends FabricObject {
   }
 
   setBoundingBox(adjustPosition?: boolean) {
-    const { left, top, width, height, pathOffset, strokeOffset } =
+    const { left, top, width, height, pathOffset, strokeDiff } =
       this._calcDimensions();
-    this.set({ width, height, pathOffset, strokeOffset });
+    this.set({ width, height, pathOffset, strokeDiff });
```
ShaMan123, line 198: add strokeOffset

src/shapes/polyline.class.ts
```diff
@@ -170,27 +205,42 @@ export class Polyline extends FabricObject {
...
+      ? // TODO: fix this
+        new Point(this.width, this.height)
       : super._getNonTransformedDimensions();
   }
 
   /**
    * @override stroke and skewing are taken into account when projecting stroke on points,
-   * therefore we don't want the default calculation to account for skewing as well
+   * therefore we don't want the default calculation to account for skewing as well.
+   * Though it is possible to pass `width` and `height` in `options`, doing so is very strange, use with discretion.
```
ShaMan123, line 216: Once your suggestion regarding width/height will be fulfilled this line should be removed

ShaMan123 (dismissed):
As before, **GREAT** work! Everything is precise.
I have checked all goldens as well, so I can confirm this PR is ready.


To hide goldens from file diff view (so it is possible to view the code diff) run the following from the dev console:

```js
document.body.querySelectorAll('div[data-tagsearch-path]').forEach(div => div.getAttribute('data-tagsearch-path').startsWith('test/visual/golden/stroke-projection') && (div.style.display='none'))
```

I have cleaned up whatever artifacts remained

src/shapes/polyline.class.ts
ShaMan123: 422be8f

ShaMan123 (dismissed):
@asturur as far as I am concerned this PR is ready
Any objections?

ShaMan123:
Notes for follow up:
`getNonTransformedDimensions` should be removed. It is confusing in name and in usage. It is used in 2 cases in fabric:
- `_renderBackground` which we have discovered that is wrong with skewing, therefore it seems best to fix that function and remove `getNonTransformedDimensions` as a dependecy
- Eraser uses it to resize itself according to its parent... https://github.com/fabricjs/fabric.js/blob/71d4d3bb69875a4d260a38228280f5ad42a65dd0/src/mixins/eraser_brush.mixin.ts#L60
Same logic should apply to it. Though `_renderBackground` can be refactored to draw before the context is transformed, the eraser cannot.

asturur:
There are a bunch of things that i don't get about this PR, mayne @ShaMan123 we can try monday to talk about it?
One of them is the necessity to pass all that options to getTransformedDimensions.

getNonTransformedDimensions is a relic of the time when there was just a simple relationship between the standard width / height and the final size.

Now width and height start to become a bit weird too.
I do think that a focused thought of how width / height are used in fabricJS today and what is their meaning for all the shapes could help decide what we should do with them.
Everytime we started from width/height and we calculate the size of the bounding box, while what we want 95% is the size of the bounding box, and for polyline/polygon/paths/triangle width and height are a side effect and not anymore the cause of the size of the shape.

I can't review the PR now because i have difficulty catching up with the stroke uniform code, but it would have been better if you just added test as the title suggested, left the bugs, get the tests in ( 90% of the files ? ) and then get a PR that would have introduced fixes and swap the tests.

ShaMan123:
Always for chatting.
Good points

luizzappa:
> * if it is possible to extract skewing and reapply it in a post calculation that sounds good... though I recall you said skewing affects the arrangement of points and the miter limit so I don't understand how that can be achieved. Since it is a perf optimization (unless it fixes something related to `getNonTransformedDimensions`) we can get back to it in a follow up.

I may have been wrong, but I think it's just scaling that changes the point arrangement (and consequently the logic of calculating the projections). The skew is always applied at the end of the projection, so we could remove it. I need to double check.

> * scaling the polygon to a given width and height sound correct. We don't need to alter the points, only the scale, right?

Initially I thought so too, but if we pass `scale` in the parameters (in addition to `height` and `width`), how would we proceed?

```
new fabric.Polygon(
    [{x: 0, y:10}, {x: 20, y:30} ],
    width: 40,
    height: 50,
    scaleX: 3,
    scaleY: 10
)
```

> * the thing with `getNonTransformedDimensions` is that it runs during the rendering process **after** the canvas context has been transformed. Could it be that it is incorrect for other shapes as well?

It's incorrect too, and so the BBOX...

image

I can correct the BBOX of the circle if I you guide me, the calculation should be the same as we did for the strokeLineJoin round.. Would I just extend _getTransformedDimensions?

luizzappa:
I created a first version ( https://github.com/fabricjs/fabric.js/pull/8452 ) with the property `resizeOffset` keeping `width` and `height` unchanged. But it's still polluted with commits and goldens from this PR here, when it's merged I'll adjust it.

ShaMan123:
@luizzappa how are you? Hope all is good.
This PR is next in line.
I wanted to ask if you are interested in PRing poly controls since you fixed them and they work in all cases.
As we discussed, we want them as part of core exposed to anyone that wants them via a simple import.

src/shapes/polyline.class.ts
```diff
@@ -7,11 +7,25 @@ import { IPoint, Point } from '../point.class';
...
 import { classRegistry } from '../util/class_registry';
 import { makeBoundingBoxFromPoints } from '../util/misc/boundingBoxFromPoints';
+import { calcDimensionsMatrix, transformPoint } from '../util/misc/matrix';
 import { projectStrokeOnPoints } from '../util/misc/projectStroke';
+import { TProjectStrokeOnPointsOptions } from '../util/misc/projectStroke/types';
 import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
 import { toFixed } from '../util/misc/toFixed';
 import { FabricObject, fabricObjectDefaultValues } from './Object/FabricObject';
 
+const strokeProjectionOptions: (keyof TProjectStrokeOnPointsOptions)[] = [
```
ShaMan123, line 17: I forgot what this is.
Shouldn't it be covered by `strokeBBoxAffectingProperties`?
The only diff is the scale keys...
I need help to recall the difference if any

ShaMan123 (commented):
updated from master
@asturur  the goldens make reading the diff impossible since github boils I have created a dummy branch without them just for this purpose https://github.com/fabricjs/fabric.js/compare/8374-view-diff

src/shapes/polyline.class.ts
ShaMan123: Maybe not...
Is not a big deal anyways

ShaMan123 (dismissed):

src/shapes/polyline.class.ts
luizzappa: Are you suggesting to use `strokeBBoxAffectingProperties` instead of `strokeProjectionOptions` in `_getTransformedDimensions` function?

```diff
     if ( 
       Object.keys(options).some((key) => 
+       (
+               strokeBBoxAffectingProperties.includes( 
+                      key as keyof TProjectStrokeOnPointsOptions 
+               ) 
+              || this.strokeUniform
+       )
       ) 
     ) {

     if ( 
       Object.keys(options).some((key) => 
-         strokeProjectionOptions.includes( 
-          key as keyof TProjectStrokeOnPointsOptions 
         ) 
       ) 
     ) {
```

Seems to make sense to me and avoids recalculating the projections when `strokeUniform = false` with scaling.

luizzappa:
> @luizzappa how are you? Hope all is good.
> This PR is next in line.
> I wanted to ask if you are interested in PRing poly controls since you fixed them and they work in all cases.
> As we discussed, we want them as part of core exposed to anyone that wants them via a simple import.

Hey @ShaMan123 !

I can work on it. It is still not very clear to me what we are going to expose. Just expose the `anchorWrapper`, `positionHandler` and `actionHandler` functions, and does the developer handle the rest? The developer would do something like this:

```
import { polyPositionHandler, polyActionHandler, polyAnchorWrapper } from 'fabric'

const poly = canvas.getActiveObject();
const lastControl = poly.points.length - 1;

poly.controls = poly.points.reduce(function (acc, point, index) {
      acc["p" + index] = new fabric.Control({
        positionHandler: polyPositionHandler,
        actionHandler: polyAnchorWrapper(
          index > 0 ? index - 1 : lastControl,
          polyActionHandler
        ),
        actionName: "modifyPolygon",
        pointIndex: index
      });
      return acc;
 }, {});

canvas.requestRenderAll();
```

src/shapes/polyline.class.ts
ShaMan123: Perhaps
I just saw it when I reviewed to merge I did and didn't understand why there are 2 similar objects that serve for a very similar purpose. So I thought I messed up the merge commit...

ShaMan123:
First of all we expose a PolyControl class ready with all the handlers.
Then, we can expose a method such as you suggested above `createPolyControls` or something

src/shapes/polyline.class.ts
luizzappa: https://github.com/fabricjs/fabric.js/pull/8374/commits/494a10ee2f8c2278ae9a55b20bf50cf6ee25b064

ShaMan123:
@asturur this is waiting too long

src/shapes/polyline.class.ts
```diff
@@ -14,18 +14,6 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
 import { toFixed } from '../util/misc/toFixed';
 import { FabricObject, fabricObjectDefaultValues } from './Object/FabricObject';
 
-const strokeProjectionOptions: (keyof TProjectStrokeOnPointsOptions)[] = [
-  'scaleX',
-  'scaleY',
```
ShaMan123: @luizzappa did I merge this correctly

ShaMan123 (approved):
Dear @luizzappa !
I hope I find you in a good and happy state.
My apologies for not seeing this PR through.
It is very good work and a shame it has been stale for so long.
I have done my best to fast forward it to the current master but I might have done something wrong. I suspect so. But I doubled checked with a view diffs (https://github.com/fabricjs/fabric.js/compare/8374-view-diff...8374-polyline-diff) to make sure but still, your review is priceless.
Once you give a green light I merge it straight away.

src/shapes/polyline.class.ts
ShaMan123: This is the commit you removed it so I think it is good to go
https://github.com/fabricjs/fabric.js/commit/494a10ee2f8c2278ae9a55b20bf50cf6ee25b064

ShaMan123: I also tested visuals against 8d5f319acc9362860c495e8ec50b4c850b04e111 so I am happy

ShaMan123:
I approved, if you have rights to squash and merge you are welcome to do so.

ShaMan123 (dismissed):

ShaMan123 (approved):

asturur:
@luizzappa  thanks for bringing this forward.
The reason why was waiting so long is because i m trying to finish the TS transition of the library and write documentation.
I would eventually not merge anything till that is done but @ShaMan123 can't really resist.

The issue with merging things is that we add new things to the library that later with a revise of documentation could entirely stop to make sense and then we would be in the weird position to make changes to code just merged.

I really would love to not add more code while we are in this beta status for so long.

ShaMan123:
It is a bug fix I need
As many of my PRs, these are things I need

ShaMan123:
I brought up from the dead

ShaMan123:
I sort of accept the fact that my prs wait until they are stale, not happy about it.
But I do not think it supports people contributing to the project, the opposite actually

asturur:
I know i m not happy everything to be stale either.
I accepted TS and Classes as a way to improve the library, even if i don't fully believe in them, and i m committed to finish it now. As i said 1000 times, adding more stuff while doing it is counter poductive.

PR #8380, opened 2022-10-17
fix(): skewing controls

ShaMan123:
## Motivation

Skewing is buggy in some cases .
#8344 has introduced changes that break the skewing controls.

## Description

Existing logic is brilliant! It used the distance between the pointer and the origin point to determine the size of the target's axis (twice the diff). Then it extracted skewing as follows:

$$
\displaylines{
x' = x + y \cdot shear_x \\
shear_x = \tan(skew_x) \\
skew_x = \arctan(\frac{x'-x}{y})
}
$$

Apparently the problem lied in the calculation of width (x) when calculating the skewed height (y) when skewing y

Problems with the previous impl:
- skewing in a group - `qrDecompose` seems to mess everything up (I think)
- skewing an object with skewing set on the counter axis - ~~fixed in https://github.com/fabricjs/fabric.js/compare/fix-skewing-controls2~~ (thanks to @luizzappa pointing out that skewX depends on skewY but not the other way round), still buggy inside group
- It assumes the change to size is linear to the change in pointer, which was correct until we fixed stroke bbox calculations. ~~I can't think of a decent way around this unless we don't calculate the projections while skewing.~~ This led me to seek a solution that compares against the last skewing state.

I have tried to come up with a different approach, relying purely on linear algebra.
If fabric move away from decomposition then we can look into that (see Gist).
This PR is intended eventually to work only with matrices, but as long as fabric relays on decomposition skewY will not work.
I think if we can live without decomposing we should (canvas API doesn't decompose), but that's a tale for another time.
Another thought I had is that in the case of skewing Y we know the value of skewX so that might be enough to alter the decomposition to return a skewY value, @asturur ?

## Changes

- Fixed the second skewing interaction - skewing is now maintained.
- Fixed cross skewing

### Calculating Shearing
We are looking for the shearing values of the transform matrix defined by the pointer offset.

$$
\displaylines{
\Delta shear=
\begin{bmatrix}
1 & x \\ 
y & 1
\end{bmatrix}, 
size=
\begin{bmatrix}
w \\
h
\end{bmatrix} \\
\Delta pointer=
\begin{bmatrix}
p_x \\ 
p_y 
\end{bmatrix}=
\Delta size = \Delta shear \cdot size=
\begin{bmatrix}
1 & x \\
y & 1
\end{bmatrix} 
\cdot
\begin{bmatrix}
w \\
h
\end{bmatrix}=
\begin{bmatrix} 
h \cdot x \\
 w \cdot y
\end{bmatrix} \\
\begin{cases}
p_x=h \cdot x \\
p_y= w \cdot y
\end{cases} \Rightarrow
\begin{cases}
x=\frac{p_x}{h} \\
y=\frac{p_y}{w}
\end{cases} \Rightarrow
\Delta shear=
\begin{bmatrix}
1 & \frac{p_x}{h} \\ 
\frac{p_y}{w} & 1
\end{bmatrix}
}
$$

### Calculating the new transform
The following solves for `skewX`, ~~same process is done for `skewY`~~ (need to account for cross axis skewing)
This logic is not used in the PR but is here in case we decide to set the transform matrix directly instead of decomposed values.
1. First we need to calculate the transform without the existing shearing:

$$
\displaylines{
t = \begin{bmatrix}
a' & 0 \\ 
b' & d'
\end{bmatrix} \\
\begin{bmatrix}
a & c \\ 
b & d
\end{bmatrix} =
\begin{bmatrix}
1 & s \\ 
0 & 1
\end{bmatrix} 
\cdot
t =
\begin{bmatrix}
1 & s \\ 
0 & 1
\end{bmatrix} 
\cdot
\begin{bmatrix}
a' & 0 \\ 
b' & d'
\end{bmatrix} =
\begin{bmatrix}
a' + s \cdot b' & s \cdot d' \\ 
b' & d'
\end{bmatrix} \\
\begin{cases}
a = a' + s \cdot b' \\
b = b' \\
c = s \cdot d' \\
d = d'
\end{cases} \Rightarrow
\begin{cases}
s = \frac{c}{d} \\
t = \begin{bmatrix}
a - s \cdot b & 0 \\ 
b & d
\end{bmatrix} 
\end{cases} \Rightarrow
t = \begin{bmatrix}
a - \frac{c}{d} \cdot b & 0 \\ 
b & d
\end{bmatrix} 
}  
$$

2. Apply the new shearing

$$
\displaylines{
\Delta shear_x=
\begin{bmatrix}
1 & \frac{p_x}{h} \\ 
0 & 1
\end{bmatrix} \\
t'_x = \Delta shear_x \cdot t=
\begin{bmatrix}
1 & \frac{p_x}{h} \\ 
0 & 1
\end{bmatrix}
\cdot
\begin{bmatrix}
a - \frac{c}{d} \cdot b & 0 \\ 
b & d
\end{bmatrix}  =
\begin{bmatrix}
a + (\frac{p_x}{h} - \frac{c}{d}) \cdot b & \frac{p_x}{h} \cdot d \\ 
b & d
\end{bmatrix}
}
$$

Doing the same for `skewY` results in:
**WRONG** - we must figure out how to extract scaleX/skewX

$$
\displaylines{
s = \frac{b}{a} \\
\Delta shear_y=
\begin{bmatrix}
1 & 0 \\ 
\frac{p_y}{w} & 1
\end{bmatrix} \\
t = 
\begin{bmatrix}
a & c \\ 
0 & d - \frac{b}{a} \cdot c
\end{bmatrix} \\
t'_y= \Delta shear_y \cdot t =
\begin{bmatrix}
a & c \\ 
\frac{p_y}{w} \cdot a & d + (\frac{p_y}{w} - \frac{b}{a}) \cdot c
\end{bmatrix}
}
$$

3. Another approach:

$$
\displaylines{
T = T_{translate} \cdot T_{rotate} \cdot T_{scale} \cdot T_{skewX} \cdot T_{skewY} \\
T_{no skew} = T_{translate} \cdot T_{rotate} \cdot T_{scale} = T  \cdot T_{skewY} ^{-1} \cdot T_{skewX} ^{-1} \\
T_{apply Skew}(skewX', skewY') = T_{no skew} \cdot T_{skewX'}  \cdot T_{skewY'} \\
}
$$

**I have managed to get skewX to work with both approaches.** SkewY however...

## TO DO
- [x] verify skewing one axis with existing skew value on the other
- [x] origin switching: change or account for size when skewing against direction to fit the mouse more accurately
- [ ] skewing inside group

## In Action
tracking mouse
multiple interactions
skewY


asturur:
### let's address the `i don't trust qrDecompose` first:
The fact you don't trust qrDecompose is because you don't accept that that has a narrow scope.
Given a series of transformations done in fabricjs order, or in an order that is equivalent, you obtain a matrix. qrDecompose is going to give you a set of **EQUIVALENT** parameters that represent that transformations if you apply them again in fabricJS order.
It can't tell you more because everything as been mashed together.
if i tell you `6`, you can answer me with 5 + 1, 3 * 2, 6, 4 + 2. Are all valid answers and i can't pretend you guess the original one.
That is what qrDecompose is doing and is doing it good.

qrDecompose is a pillar of fabricJS geometry and a constrain, and we can't avoid using it, it would come at the cost of a lot of calculations when skew is involved.


### Let's give the proper introduction to those bugs

Skew controls, compared to other controls, have an harsh logic.
They have visual bugs when both skewX and skewY are mixed together. 
Skewing on both axis is an edge case and our brain difficultly wraps around it.

Which is this always buggy behaviour?
Let's start from the non buggy one.
Take a rectangle and skew it along one axe only, X or Y.

image

If you press down shift and start skewing, you will see that with one skew only, the mouse always follow the control and all is good, you don't ecounter surprises.

When both skew are used, you will see that the skew controls start to move at a different speed compared to the mouse cursor. this becase the calculation that are there can't reverse engineer this question:
Given an object with
- angle A,
- scale 4,2
- skewX : 33
- skewY: 12
taken the mouse at position x,y on the control for skewY
moving the mouse to x1 = x +3 and y1+ y+1
what value i have to give to skewY in order for the control to be the closest to the mouse cursor but  still on the line determined by the side of the bounding box where the skew control is located?

I didn't answer that question, i didn't even try after a couple of attempts and we should be limited to pay a little price in code complexity for that to work.

The issue with skew controls is the predictability of the next value of skew to assign when both skew are involved

### how does inkscape handle it? 
Inkscape use qrDecompose more.
If i remember correctly after skewing they reset the object transform with a sort of an equivalent trasnform and kill one of the skews. In practice i think they they use the equivalent of qrDecompose **MORE**.
Can we see how other interactive tool solve this issue with double skewing before trying to fix the code?

src/controls/controls.actions.ts
```diff
@@ -351,51 +351,43 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
-    } else {
-      newSkew = radiansToDegrees(
+      totalSkewSize = diff.y * 2,
+      currentSkew = target.skewY;
+    const skewSign =
+      (transform.originX === LEFT && transform.originY === BOTTOM) ||
+      (transform.originX === RIGHT && transform.originY === TOP) ||
+      targetHasOneFlip(target)
+        ? -1
+        : 1;
```
asturur, line 377: not sure if this is intentional.
SkewSign is supposed to change sign every time one of those conditions are true.
So this is not equivalent at all, i know is a WIP, but in case you missed that this is a completely change of direction in behavior, i ll tell you or you could waste ours on it

ShaMan123:
Great context.
Thank you! Much appreciated.
What I saw after reading the skew controls is that as you mentioned it doesn't start from the current skew state but starts from 0.
That seems a bug to me. That is why skew jumps back/flickers once you do a mouse down or use setDimensions

> a binary search in future skew values that will make the mouse and the control conincident

> a binary search in future skew values that will make the mouse and the control conincident

Can't guess what these mean.

> how does inkscape handle it?

Could you add visuals?

src/controls/controls.actions.ts
ShaMan123: thanks
didn't really understand this logic yet

asturur:
Inkscape normalize the bounding box on rotation and not in skew.
They get double skewing right, meaning no JUMPS, but they also have the control desynced with the mouse.
Probably jumps are removed if we remove the dependency from the bounding box size.
We do guess for the current transformation how much skew should be added for a mouse movement and we add eithout trying to match any position.

That would probably remove weird jumps entirely

ShaMan123:
I have done a major refactoring that will probably enable not using `qrDecompose` for skewing (in case one day `qrDecompose` will fade out and matrix will b the only thing defining object transform).
Fixed the second+ skewing interaction.
I will add an explanation soon.

Please check it for accuracy.
I will need help to finalize skewY

ShaMan123 (commented):
- **Updated description**

- origin switching is weird IMO now that successive interactions are supported
origin switch

- skewY after skewX is broken
skewY after skewX


Please review

src/shapes/object.class.ts
```diff
@@ -1687,52 +1688,54 @@ import { enlivenObjectEnlivables } from '../util/misc/objectEnlive';
...
+        let center = this._findCenterFromElement();
+        const t = this.transformMatrix;
+        if (t) {
+          const { scaleX, scaleY, angle, skewX } = qrDecompose(t);
+          this.flipX = false;
+          this.flipY = false;
+          this.set('scaleX', scaleX);
+          this.set('scaleY', scaleY);
+          this.angle = angle;
+          this.skewX = skewX;
```
ShaMan123, line 1714: this is a case that you warned from
I tried calling `set` instead of manual assignment but since polys run a side effect this breaks.
I can set `initialized=false` so it won't run.

ShaMan123 (commented):
fixed tests

ShaMan123:
> The fact you don't trust qrDecompose is because you don't accept that that has a narrow scope.

To me that is dangerous since I can't use it everywhere and for everything so I keep my guards up. 
IMO (could be a matter of taste) remembering such details is an overkill and in general feels fragile.
I avoid impl that don't cover all usable cases. And if by chance I must I rename them as DANGEROUS_XXX

ShaMan123:
IMO origin logic should be dumped and handled by the dev

ShaMan123:
Or we try to fix it - switching an origin during a skew interaction fc70ee9

luizzappa:
At first glance I would say that determining `skew` based on pointer variation would be very unstable. Because the new `width` or `height` after applying the `skew` is totally dependent on the position of the polygon points.

For example, the two triangles below have the same initial width, but if we apply the same `skewX` to both, the first one will have a wider `width` because one of the vertices has a much larger Y.

$X_f = X_i + Y * Tan(skewX)$

image

But **it surprised me**, I tested it here and the accuracy is very good. I tried to force some extreme cases, and it looked pretty decent.

Maybe it improves the accuracy a bit more if you apply the inverse of the scale to the `offset`, since 1 pixel moved on the mouse is not the same on the canvas when `scale` is different from 1.

ShaMan123:
## Under group

seems that we need to account for the container's skewing we calculating offset

https://user-images.githubusercontent.com/34343793/196619662-db960e1d-cea3-494c-a376-a58ab7decc5f.mp4


https://user-images.githubusercontent.com/34343793/196619812-33441d5f-fb78-4185-bb17-77b70a975226.mp4


https://user-images.githubusercontent.com/34343793/196619938-a3361f42-b420-44b3-aaf4-95f7cd0eda5b.mp4

https://user-images.githubusercontent.com/34343793/196620236-bad51b56-c807-48c0-b0ca-6441c456a67b.mp4


https://user-images.githubusercontent.com/34343793/196620691-0ffe99e3-b72b-4108-a274-2f39681bc0f9.mp4

https://user-images.githubusercontent.com/34343793/196621079-9a31599a-38f9-4780-bca2-f4b19d8f2920.mp4

ShaMan123 (commented):
I have tested against master.
Skewing under group seems the same there.
It wasn't fixed as part of the group rewrite.
I am quite convinced it is possible. 
But I think it deserves its own PR since this PR has matured and is becoming large.

src/controls/controls.actions.ts
```diff
@@ -268,134 +269,80 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+            }
+          : {
+              otherAxis: 'y',
+              otherScaleKey: 'scaleY',
+              counterOrigin: target.resolveOriginY(originY),
+              key: 'skewX',
+              valueBefore: skewX,
+            },
+      dim = target._getTransformedDimensions(),
+      offset = pointer.subtract(new Point(ex, ey))[axis],
```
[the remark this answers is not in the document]
ShaMan123: I will try

src/controls/controls.actions.ts
```diff
@@ -268,134 +269,80 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+    const changed = valueBefore !== target[key];
+
+    if (changed) {
+      const otherSkew = new Point(target.skewX, target.skewY)[otherAxis];
+      const compensationFactor =
+        otherSkew !== 0
+          ? dim[otherAxis] / target._getTransformedDimensions()[otherAxis]
+          : 1;
+      compensationFactor !== 1 &&
+        target.set(otherScaleKey, compensationFactor * scale);
```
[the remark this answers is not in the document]
ShaMan123: Credit belongs to @asturur 
I simply refactor it.
Tried to wrap my ahead around.
Would love additional context

src/controls/controls.actions.ts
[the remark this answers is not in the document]
ShaMan123: Tried it and doesn't seem to do any good.
Seems accurate to the point except when crossing over to negative values
Correct me if I am wrong

src/controls/controls.actions.ts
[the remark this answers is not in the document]
luizzappa: Every `skew` does the following transformation:

`skewX` ==> $X_f = X_i+ Y * Tan(skewX)$
`skewY` ==> $Y_f = Y_i+ X * Tan(skewY)$

The two values ​​of $X_f$ and $Y_f$ are now dependent on the other axis. In order not to have a circular reference from one skew impacting the calculation of the other, I need an order in which `skew` is applied, which can be one of two:


first apply `skewY` and then `skewX` ==> $X_f = X_i+ Y_f * Tan(skewX)$

or

first apply `skewX` and then `skewY` ==> $Y_f = Y_i+ X_f * Tan(skewX)$

The order chosen in canvas is the first one, so any change in `skewY` affects the value of $X_f$ (hence the `width`) when there is `skewX` applied. What this `compensationFactor` does is compensate for the increase in `width` to keep it the same size.

Due to this order of application, we only need to apply this `compensationFactor` when `skewX !==0` and the transformation occurs on the `Y axis`.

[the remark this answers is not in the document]
ShaMan123: Marvelous, thanks!

src/controls/controls.actions.ts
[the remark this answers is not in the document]
luizzappa: Although it seems that the problem is when it crosses to negative values, but in reality it is because of the arrangement of the polygon points.

That's the problem I mentioned [here.](https://github.com/fabricjs/fabric.js/pull/8380#issuecomment-1283383318) Depending on the arrangement of the polygon points, a change in `skew` will not necessarily change the `width` of the bbox (or it will not vary the `width` in the same proportion as the mouse movement)

I made an example with two cases that will be clearer: https://codesandbox.io/s/skew-control-0frlui?file=/src/index.js 

Try applying `skewX` to the rectangle. See that it works correctly, with good accuracy even crossing to negative values ​​to some extent.

control skew square

Then try to apply `skewX` on the triangle. The mouse movement does not follow the control point. Because the triangle only starts to change its width when the top vertex reaches one of the edges.

control skew triangle

So depending on the arrangement of the points the mouse movement will follow the control point well or not. (I don't see this as a big problem)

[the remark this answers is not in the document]
luizzappa: If you think that for UX it is really important for the mouse to be fully synchronized with the control point you can compensate for the difference with the scale. For example, you moved the mouse to the right 10 pixels, but the polygon after applying the new skewX grew only 5 pixels wide, you apply a new scaleX factor to reach 10 pixels and stay in sync with the mouse.

ShaMan123 (commented):
Fixed origin anchoring!
Instead of determining the anchor origin using the offset, uses the skew value.

### Before


https://user-images.githubusercontent.com/34343793/196859188-ea2be1eb-843a-4b4e-b8f5-225fbdfa785c.mp4




### After


https://user-images.githubusercontent.com/34343793/196859080-2731a5a7-0a82-4e98-b688-cb19c2a3f350.mp4

src/controls/controls.actions.ts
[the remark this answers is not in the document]
ShaMan123: This is why the calculated matrix in the description works for `skewX` but not for `skewY`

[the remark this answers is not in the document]
ShaMan123: Let me see if I understand:

$$
\displaylines{
shear_x=
\begin{bmatrix}
1 & x \\ 
0 & 1
\end{bmatrix}, 
shear_y=
\begin{bmatrix}
1 & 0 \\ 
y & 1
\end{bmatrix} \\
shearing = shear_x \cdot shear_y=
\begin{bmatrix}
xy & x \\ 
y & 1
\end{bmatrix}
}
$$

src/shapes/object.class.ts
asturur: intialized false has to go away, do not cement it.

src/util/misc/objectTransforms.ts
```diff
@@ -60,13 +60,13 @@ export const applyTransformToObject = (
...
     center = new Point(translateX, translateY);
-  object.flipX = false;
-  object.flipY = false;
-  Object.assign(object, otherOptions);
-  object.set({ scaleX, scaleY });
+  object.set({
+    flipX: false,
+    flipY: false,
+    ...otherOptions,
+  });
```
asturur, line 69: thos are made explicitly to run set only on necessary side effects, that should be as little as possible.
The only we need is scale, that reset flip or enabe flip.
This shouldn't be the norm, so please don't do it

src/util/misc/objectTransforms.ts
```diff
@@ -76,12 +76,14 @@ export const applyTransformToObject = (
...
-  target.flipX = false;
-  target.flipY = false;
+  target.set({
+    flipX: false,
+    flipY: false,
+    scaleX: 1,
+    scaleY: 1,
+    skewX: 0,
+    skewY: 0,
+  });
```
asturur, line 86: let's leave this as Object.assign(target, {});
today set has the particular meaning of `side effect needed` i do not want to offuscate which part of the code now are using it because necessary

asturur:
@ShaMan123 i couldn't read all yet.
I see visual test changes and that could mean broken svg import, because you didn't change the svgs in this pr.

Also ( i will read the PR comments) but if this is missing please add on the gifs an indication of the transform that is applied to the object before you try skewing ( is there a skew of the opposite side yet )?

P.S.
Canvas api does not decompose because it does not need add, it just stack transformations on top, it multiplies all the time, and it doesn't care that you want to know how much is the total scale. or the toltal skew.

Did you look at inkscape codebase by chance?

asturur:
Also this comment is also because i didn't read the pr.

I do ask that if you try to fix the control skewing, you do it OUTSIDE the special case of polygon with acute angles and exact bounding boxes and large stroke width.
You do it for the simple rect. When that case is safe and sound, we look into the other.

This time dividing the issue in small understandable changes is mandatory.
no exceptions please.

src/util/misc/objectTransforms.ts
ShaMan123: done

src/controls/controls.actions.ts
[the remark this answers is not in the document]
ShaMan123: edited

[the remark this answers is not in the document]
luizzappa: That's right. In the calculations in the PR description seem to me to have an error in the development of this calculation here:

$\Delta shear . size$

ShaMan123 (commented):
I have updated the description.
Please read it carefully.

I don't know what I think about this PR.
It fixes some things but cross skewing isn't as good as https://github.com/fabricjs/fabric.js/compare/fix-skewing-controls2

All in all I have no idea how to move forward since:
- it seems `qrDecompose` makes it really difficult to handle skewing inside groups. I suggested we look into extracting skewY from the decomposition when skewing Y since we know the skewX value before hand.
- couldn't get the correct matrix operation to extract skewY

ShaMan123:
I am giving up on this attempt.
Made some progress but still it is not good enough, especially when cross skewing.
https://github.com/fabricjs/fabric.js/compare/fix-skewing-controls2 is better since it simply refactors the code, preserve the original logic and fixes continuous interactions.
If we do `exactBoundbox = false` during skewing calculation that fix the problem with stroke projection

ShaMan123:
I take that back.
Looked at it again.
Once skewing y when target is skewX != 0 is fixed this is good to go.

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,108 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+      skewingBefore = new Point(target.skewX, target.skewY)[axis],
+      shearingStart = Math.tan(degreesToRadians(skewingStart)),
+      counterOriginFactor = new Point(
+        target.resolveOriginX(originX),
+        target.resolveOriginY(originY)
+      )[counterAxis],
+      offsetFactor = -Math.sign(counterOriginFactor) * flip * 2;
+
+    const shearing =
+      (offset * offsetFactor) / Math.max(dimBefore[counterAxis], 1) +
```
ShaMan123, line 294: This is the line that needs fixing
if `axis === y` it seems `dimBefore[counterAxis]` should be something else.

@luizzappa would love your help with this

ShaMan123 (commented):
This PR has a future

src/controls/controls.actions.ts
ShaMan123: test it by skewing X to a big value and then start skewing Y
You will see the the skew change is not li

ShaMan123: fixed 89d2c0b

ShaMan123:
I have an idea how to fix skewing with exactBoundingBox now that this is fixed
In a dedicated PR

ShaMan123:
## Skewing in Group

$$
\displaylines{
T = T_{translate} \cdot T_{rotate} \cdot T_{scale} \cdot T_{skewX} \cdot T_{skewY}  \\
T_{final} = T_{group} \cdot T \\
T_{no skewY} = T_{final}  \cdot T_{skewY} ^{-1} \\
}
$$

Then is do:
`const decomposed = { ...qrDecompose(`$T_{no skewY}$`), skewY }`

ShaMan123:
still need to account for
- scaling and then skewing X
- rotating pointer accuracy

src/canvas.class.ts
```diff
@@ -806,7 +806,8 @@ import { FabricObject } from './shapes/object.class';
             lastX: pointer.x,
             lastY: pointer.y,
             theta: degreesToRadians(target.angle),
-            width: target.width * target.scaleX,
+            width: target.width,
+            height: target.height,
```
ShaMan123, line 810: nothing seems to be using these 2
but it makes sense to store them without scaling

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+        target.resolveOriginX(originX),
+        target.resolveOriginY(originY)
+      )[counterAxis],
+      offsetFactor = -Math.sign(counterOriginFactor) * flip * 2,
+      // let a, b be the size of target
+      // let a' be the value of a after applying skewing
+      // then:
+      // a' = a + b * skewA => skewA = (a' - a) / b
+      // the value b is tricky since skewY is applied before skewX
+      b =
```
ShaMan123, line 298: This is the fix

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+    const changed = skewingBefore !== target[skewKey];
+
+    if (changed && axis === 'y') {
+      // we don't want skewing to affect scaleX so we factor it by the inverse skewing diff to make it seem unchanged to the viewer
+      const { skewX, scaleX } = target,
+        dimBefore = target._getTransformedDimensions({ skewY: skewingBefore }),
+        dimAfter = target._getTransformedDimensions(),
+        compensationFactor = skewX !== 0 ? dimBefore.x / dimAfter.x : 1;
+      compensationFactor !== 1 &&
+        target.set('scaleX', compensationFactor * scaleX);
```
ShaMan123, line 322: refactored from `compensateScaleForSkew`

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+      // if skewing = 0 use the offset to determine anchoring
+      origin =
+        ((target[skewKey] === 0 &&
+          getLocalPoint(transform, CENTER, CENTER, x, y)[axis] > 0) ||
+        target[skewKey] > 0
+          ? -1
+          : 1) *
+          -Math.sign(counterOriginFactor) *
+          0.5 +
+        0.5;
```
ShaMan123, line 363: this condition is the same a previous separate conditions smashed together

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
-    var hasSkewed = currentSkew !== newSkew;
-    if (hasSkewed) {
-      var dimBeforeSkewing = target._getTransformedDimensions().x;
-      target.set('skewY', newSkew);
-      compensateScaleForSkew(target, 'skewX', 'scaleX', 'x', dimBeforeSkewing);
-    }
-    return hasSkewed;
+    const counterOriginFactor = new Point(
+        target.resolveOriginX(transform.originX),
+        target.resolveOriginY(transform.originY)
```
ShaMan123, line 351: once #8390 is in will need to refactor this bit

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+      // let a' be the value of a after applying skewing
+      // then:
+      // a' = a + b * skewA => skewA = (a' - a) / b
+      // the value b is tricky since skewY is applied before skewX
+      b =
+        axis === 'y'
+          ? target._getTransformedDimensions({
+              scaleX: 1,
+              scaleY: 1,
+              // since skewY is applied before skewX b (=width) is not affected by skewX in this case
```
ShaMan123, line 303: is this straight forward enough?
can you understand it?

ShaMan123 (commented):
FIXED!!

src/controls/controls.actions.ts
ShaMan123: credits to @luizzappa

src/controls/controls.actions.ts
asturur: so if this is the fix, what are all the other diffs about?

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+      offset = pointer
+        .subtract(new Point(ex, ey))
+        .divide(new Point(target.scaleX, target.scaleY))[axis],
+      flip = new Point(flipX ? -1 : 1, flipY ? -1 : 1)[axis],
+      skewingStart = new Point(skewX, skewY)[axis],
+      skewingBefore = new Point(target.skewX, target.skewY)[axis],
+      shearingStart = Math.tan(degreesToRadians(skewingStart)),
+      counterOriginFactor = new Point(
+        target.resolveOriginX(originX),
+        target.resolveOriginY(originY)
```
asturur, line 290: my PR is changing resolveOriginX and resolveOriginY away from methods ( since they don't use this ) and there is just an exported function ( that could be an util ) that is called resolveOrigin and takes indifferently originX or originY

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
-      target.set('skewX', newSkew);
-      compensateScaleForSkew(target, 'skewY', 'scaleY', 'y', dimBeforeSkewing);
+  function skewHandler(axis: TAxis, eventData, transform, x, y) {
+    const { target } = transform,
+      {
+        counterAxis,
+        origin: originKey,
+        lockSkewing: lockSkewingKey,
+        skew: skewKey,
+      } = AXIS_KEYS[axis];
```
asturur, line 344: Is this const used locally here only? if yes instead of renaming the keys, could we define the object with the correct key names right away and keep it in this file here?
Seems like not a generic const

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+    if (target[lockSkewingKey]) {
+      return false;
```
asturur, line 346: we should probably move all of those in its own wrapper if possible, something like `lockedBy('prop', handler)` just an idea for a task. We have scale, skew, rotate that all have the same concept

src/controls/controls.actions.ts
asturur: also this seems silly to resolve both origin when we know we are going to get one only. it depends on counterAxis we should be able to figure it out.

src/canvas.class.ts
asturur: probably are super old

src/controls/controls.actions.ts
asturur: like just adding another key to that object and avoiding creating the point

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
-    if (target[oppositeSkew] !== 0) {
-      var newDim = target._getTransformedDimensions()[axis];
-      var newValue = (reference / newDim) * target[scaleToCompensate];
-      target.set(scaleToCompensate, newValue);
-    }
+    const { counterAxis, skew: skewKey } = AXIS_KEYS[axis],
+      offset = pointer
+        .subtract(new Point(ex, ey))
+        .divide(new Point(target.scaleX, target.scaleY))[axis],
+      flip = new Point(flipX ? -1 : 1, flipY ? -1 : 1)[axis],
```
asturur, line 284: before the condition was `has a flip only`, while now we look at the flip independently.
Are we sure is equivalent?

src/controls/controls.actions.ts
```diff
@@ -268,138 +270,116 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+      skewingStart = new Point(skewX, skewY)[axis],
+      skewingBefore = new Point(target.skewX, target.skewY)[axis],
```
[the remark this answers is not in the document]
asturur: or look at least i would create inside the handler itself a fat arrow
```
const chooseByAxisGen (axis) => (...twoValues) => new Point(...twoValues)[axis];
const chooseBetween = chooseByAxisGen(axis);
// and then i would just use 
skewingStart = chooseBetween(skewX, skewY),
```
for me at leat is way more clear what is going on rather than have a bunch of point there in the code that are no points but more buckets of positioned values

ShaMan123:
I agree with all.
And will work it in. 
The line I marked as the fix is one of 3. I matked since that made the difference between abadoning the PR or not. It was the missing piece.
The fixes:
1. Calculation has been changed to a calculating the diff against the previous value instead of against no skew dims.
2. Account for skew start
3. SkewY not dependant on skewX

src/controls/controls.actions.ts
ShaMan123: In general controls need es6 + TS work
That can be added there

src/controls/controls.actions.ts
ShaMan123: Decide what to merge first and I will fix in the other

ShaMan123 (commented):
The only thing remaining is originX/Y changes
Tell me if it is good enough

ShaMan123 (commented):
READY after comments and extensive skewing with:
- scaling
- flipping
- rotating
- initial skewing
- double skewing
- vpt
- successive skewing

src/controls/controls.actions.ts
ShaMan123: I tested it 
seems fine
The logic seems more robust so 1ba2724 seems not to have any affect but I put it there for sanity

ShaMan123:
seems 91c5d72 fixed https://github.com/fabricjs/fabric.js/pull/8380#issuecomment-1291574280

asturur:
Ok so i tried it, and the flip behaviour is changed.
Is not the end of the world but i m trying to fix it, with few result.
If i manage i ll add a PR over thi PR or i ll let it go

You can compare a single flip target with master to see that some direction changes when skewing and it just feels innatural.

ShaMan123:
So funny
I just found the bug and pushed and then received your comment
Try now!

ShaMan123:
> You can compare a single flip target with master to see that some direction changes when skewing and it just feels innatural.

If this continues please share a screen cast

src/controls/controls.actions.ts
```diff
@@ -267,243 +269,180 @@ import { renderCircleControl, renderSquareControl } from './controls.render';
...
+        target[skewKey] > 0
+          ? 1
+          : -1) * skewingSide,
+      // anchor to the opposite side of the skewing direction
+      // normalize value from [-1, 1] to origin value [0, 1]
+      origin = -skewingDirection * 0.5 + 0.5;
+
+    const finalHandler = wrapWithFireEvent(
+      'skewing',
+      wrapWithFixedAnchor((eventData, transform, x, y) =>
```
ShaMan123, line 405: Do you think this affects perf?

asturur:
Going to retry now, i was literally adding -1 and 1 randomly around with the flipX !== flipY logic

ShaMan123 (commented):
It is fixed now
But I have seen another related bug that exists on master as well
I think it happens only when skewing Y and when `skewY * skewX < 0`, then target is kind of flipped

https://user-images.githubusercontent.com/34343793/198539158-b250e6ba-22ae-4701-82c3-f7601330fb56.mp4

asturur:
yes now it behaves more consistently.
Any other eventual change in behaviour is irrelevant for me.

Good pr!

asturur (approved):

PR #8388, opened 2022-10-21
feat(circle): new attr(counterClockwise) of circle

737261520:
new attr(counterClockwise) of circle

### DIFF
--- a/src/shapes/circle.class.ts
+++ b/src/shapes/circle.class.ts
@@ -54,6 +54,14 @@ const Circle = fabric.util.createClass(
      */
     endAngle: 360,
 
+    /**
+     * Specify whether the drawing should be counterclockwise or clockwise
+     * probably will change to degrees in next major version
+     * @type Boolean
+     * @default false
+     */
+    counterClockwise: false,
+
     cacheProperties: FabricObject.prototype.cacheProperties.concat(
       'radius',
       'startAngle',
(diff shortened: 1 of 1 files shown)


src/shapes/circle.class.ts
```diff
@@ -54,6 +54,14 @@ const Circle = fabric.util.createClass(
      */
     endAngle: 360,
 
+    /**
+     * Specify whether the drawing should be counterclockwise or clockwise
+     * probably will change to degrees in next major version
```
ShaMan123, line 59:
```suggestion
     * probably will change to degrees in next major version
```

ShaMan123 (changes requested):
Thanks for PRing
Looks good: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/arc#syntax

- [ ] We should add a unit test and visual test, read https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#-testing, I can help with that
- [ ] add a log to the changelog
- [ ] add to `toObject` `intialize` methods

asturur:
Hi @737261520 can i ask you how this is used?
Is this for the startAngle/endAngle? or for some path winding logic?

737261520:
> Hi @737261520 can i ask you how this is used? Is this for the startAngle/endAngle? or for some path winding logic?

look this: https://developer.mozilla.org/zh-CN/docs/Web/API/CanvasRenderingContext2D/arc

PR #8390, opened 2022-10-21
chore(TS): Convert object geometry in its own class

asturur:
## Motivation
Moving forward with the TS migration

## Description

- basic properties and methods of object like top,left,width,height... and all the other that infuence position are moved to the ObjectOrigin class.
- basic properties and methods that influence dimensions of bounding box like skew, scale and strokeUniform and few other things, have been moved to ObjectGeometry class
- FabricObject extends ObjectGeometry, that extends ObjectOrigin, trying to keep code divided by concern and TS happy.
- Basic typing across all code touched, and mandatory es6 conversion ( const, let, fat arrows, some default args, some spread operators)

Really nothing else
Diffs are unreadable, i m sorry bewteen spacing and syntax changes git can't handle it.

## Changes

All added lines are JSDOCS that were missing

BREAKING:
- some chainability removed: setCoords, scale, scaleToWidth, scaleToHeight
- adjustPosition is gone ( what it was even for? )


asturur:
i think i m at 50% of this file and till now i didn't find any issue or bad thing

src/shapes/group.class.ts
```diff
@@ -2,6 +2,10 @@
 import { Point } from '../point.class';
 import { FabricObject } from './object.class';
 
+export class Group extends FabricObject {
+
+}
+
```
asturur, line 8: just to make typings happy for now

asturur:
Ok i have 29 Unit test to address by the code is done, apart the part that needs to be changed to get back the UTs in shape.

asturur:
22 UTs to go

ShaMan123 (commented):
- fixed typos
- added comments
- fixed buildStats diff sign
- removed duplicate TBoundingBox => TBBox
- added comment regarding `getScaledWidth/Height`
I think we should deprecate them and expose `getSize` that will return the size in canvas plane

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
-      },
-
-      /**
-       * Returns the coordinates of the object based on center coordinates
-       * @param {Point} point The point which corresponds to the originX and originY params
-       * @return {Point}
-       */
-      // getOriginPoint: function(center) {
-      //   return this.translateToOriginPoint(center, this.originX, this.originY);
-      // },
```
asturur, line 175: deleted forever

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
-        if (typeof to === 'string') {
-          offsetTo = originXOffset[to];
-        } else {
-          offsetTo = to - 0.5;
-        }
-        this.left += xFull * (offsetTo - offsetFrom);
-        this.top += yFull * (offsetTo - offsetFrom);
-        this.setCoords();
-        this.originX = to;
-      },
```
asturur, line 291: deleted forever

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
+      );
+
+      this.left = originPoint.x;
+      this.top = originPoint.y;
+
+      this.originX = this._originalOriginX;
+      this.originY = this._originalOriginY;
+      this._originalOriginX = undefined;
+      this._originalOriginY = undefined;
+    }
```
asturur, line 391: would love to refactor those to be just inside the object.rotate method, and remove those as standalone methods. Not today

asturur:
@ShaMan123  ok this is it.
The breaking changes are listed, see if there is something that really stands out ( in theory we have no code changes here ), then i want to move on to more conversion

src/mixins/object_geometry.mixin.ts
```diff
@@ -1,870 +1,898 @@
...
+        translateX: center.x,
+        translateY: center.y,
+        scaleX: this.scaleX,
+        scaleY: this.scaleY,
+        skewX: this.skewX,
+        skewY: this.skewY,
+        flipX: this.flipX,
+        flipY: this.flipY,
+      },
+      value = composeMatrix(options);
```
ShaMan123, line 868: Here is where I think of trying the skewY trick

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
+   * @default 0
+   */
+  angle: TDegree;
+
+  /**
+   * Width of a stroke used to render this object
+   * @type Number
+   * @default 1
+   */
+  strokeWidth: number;
```
ShaMan123, line 115: I know this class requires strokeWidth and strokeUniform but I think they belong in an abstract super class
Thoughts?
But we can move ahead and think about this in a follow up

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
+   * Calculate object bounding box dimensions from its properties scale, skew.
+   * @param {Object} [options]
+   * @param {Number} [options.scaleX]
+   * @param {Number} [options.scaleY]
+   * @param {Number} [options.skewX]
+   * @param {Number} [options.skewY]
+   * @private
+   * @returns {Point} dimensions
+   */
+  _getTransformedDimensions(options: any = {}): Point {
```
ShaMan123, line 150: I want to rename this important method to `getDimensions` or `calcDimensions`
I think it should deprecate `getScaledWidth/height`

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
+    const dimOptions = {
+      scaleX: this.scaleX,
+      scaleY: this.scaleY,
+      skewX: this.skewX,
+      skewY: this.skewY,
+      width: this.width,
+      height: this.height,
```
ShaMan123, line 157: I put width and height here a while back
But now (after thinking of a fix to skewing controls + stroke projection) I understand it is wrong.
I makes no sense that a consumer of this method will provide the size.
Does it? The whole point of the method is to return the size of transform.
Passing width/height lets you "project" the bbox with a transform....
Thoughts?

src/mixins/object_origin.mixin.ts
```diff
@@ -1,345 +1,404 @@
...
+    toOriginX: TOriginX,
+    toOriginY: TOriginY
+  ): Point {
+    let x = point.x,
+      y = point.y;
+    const offsetX = resolveOrigin(toOriginX) - resolveOrigin(fromOriginX),
+      offsetY = resolveOrigin(toOriginY) - resolveOrigin(fromOriginY);
+
+    if (offsetX || offsetY) {
+      const dim = this._getTransformedDimensions();
```
ShaMan123, line 208: This part I didn't touch in the group PRs but it requires thought
This method is restricted to the containing plane of the object.
Should it be? Is it confusing?

ShaMan123: Of course changing this method means chaning all the rest.

src/mixins/object_origin.mixin.ts
ShaMan123: Agreed
I would add `normalizePoint` to the list.
I think it is used only by controls.

src/mixins/object_geometry.mixin.ts
```diff
@@ -1,870 +1,898 @@
...
-        }
-
-        return finalDimensions.scalarAdd(postScalingStrokeValue);
-      },
+  /**
+   * Calculate object dimensions from its properties
+   * @private
+   * @returns {Point} dimensions
+   */
+  _getNonTransformedDimensions(): Point {
```
ShaMan123, line 881: We should deprecate this method as well.
It is used only by `Object._renderBackground`

ShaMan123 (commented):
Looks great
From all I commented I would rename `_getTransformedDimensions` ASAP
I guess you would prefer a dedicated PR.
Great job!

ShaMan123 (approved):

src/mixins/object_geometry.mixin.ts
ShaMan123: And is weird in some context

src/mixins/object_origin.mixin.ts
asturur: maybe yes i don't want to over engineer, no one is going to look at those classes and they are going to look at Object only or Object + interaction.

I want to put a target on eoy for finishing this thing and all the changes that do not touch the public api can be postponed for me.

src/mixins/object_origin.mixin.ts
asturur: let's have a renaming task for all of those, but let's talk about it in a video call so that we can actually have a plan, and then we just go and rename. Renaming needs to be done before release of course.

src/mixins/object_geometry.mixin.ts
asturur: yes getNonTransformedDimensions is not used anymore as before. is no more the base for getTransformed.

asturur:
@ShaMan123 can you open 2 tasks with your comments? is not that i m lazy i m really short in time and i don't want to loose them.
One for what we want to rename, and one for what we want to merge where use and remove.

PR #8391, opened 2022-10-23
chore-ts(): parser typing and es5-ify

Lazauya:
Initial first-stab typing and es5-ificiation for everything related to `parseSVGDocument`. There SHOULD be NO functionality change to my knowledge, only rewrites of old code styles. Stuff like changing `for(i = 0; i < els.length; i++`) to `for(const el of els)`. There's a A LOT that needs discussion here.

Sorry about the number of files changed. It looks like a lot of stuff, but most of the files are really small (less than 20 lines and made shorter with es5).

The `ElementsParser` in src/parser/elements_parser.ts has been changed to a es5 class. There's also a lot of stuff in that file that makes it difficult to type. See comments about it for further info.


src/parser/elements_parser.ts
```diff
@@ -1,92 +1,136 @@
...
 
-  proto.createObject = function (el, index) {
+  createObject(el: Element, index: number) {
     const klass = this.findTag(el);
-    if (klass && klass.fromElement) {
+    // TODO: firgure out a better way. maybe modify sig of FabricObject?
+    if (klass && (klass as any).fromElement) {
       try {
-        klass.fromElement(el, this.createCallback(index, el), this.options);
+        (klass as any).fromElement(el, this.createCallback(index, el), this.options);
```
Lazauya, line 89: fromElement exists on SOME fabric objects (like the gradient). Maybe we can add a default inherited method, IRDK.

src/parser/parseSVGDocument.ts
```diff
@@ -41,24 +58,20 @@ export function parseSVGDocument(doc, callback, reviver, parsingOptions) {
...
     // we're likely in node, where "o3-xml" library fails to gEBTN("*")
     // https://github.com/ajaxorg/node-o3-xml/issues/21
-    descendants = doc.selectNodes('//*[name(.)!="svg"]');
-    const arr = [];
-    for (i = 0, len = descendants.length; i < len; i++) {
-      arr[i] = descendants[i];
-    }
-    descendants = arr;
+    // TODO: aparently selectNodes is a .NET thing. see about replacing?
+    descendants = (doc as any).selectNodes('//*[name(.)!="svg"]').slice();
```
Lazauya, line 74: this seems to be some kind of compatibility thing with .NET (internet explorer, I think). Not sure what to do about this. `selectNodes` doesn't exist in the standard lib.

ShaMan123:
@Lazauya I know @asturur wants to rewrite the parser.
He got pissed at me that I migrated it since he doesn't want anyone to waste time.
Let's wait for him to comment.
I do think that it will take a **long** time until anyone gets to rewriting it (> 1 year IMO) so it might be worthwhile.

If you want to continue the work ping me or @asturur so that we point out what to do.
Maybe we should post that on the TS thread once Object has been fully migrated.

Meanwhile:
- [ ] `src/util/path.ts` - convert to a directory, es6, types.
- [ ] filter
- [ ] controls

Let's wait for @asturur to comment on these as well since I think he might want filter or controls to himself

src/parser/elements_parser.ts
ShaMan123: For these kinds of things best practice is to declare an interface and force classes to implement it with `implements`.
The stricter approach is declaring an abstract base class but that is most likely incompatible for fabric since Object extends a base class that is unrelated.

src/parser/constants.ts
```diff
@@ -23,28 +26,50 @@ export const reFontDeclaration = new RegExp(
...
+  path = 'path',
+  circle = 'circle',
+  polygon = 'polygon',
+  polyline = 'polyline',
+  ellipse = 'ellipse',
+  rect = 'rect',
+  line = 'line',
+  image = 'image',
+  text = 'text',
+}
```
asturur, line 39: how enums move from TS to JS? i know const enum gets removed at compilation time, but those?

asturur: If those Enums aren't really useful for the code flow, i prefer the static arrays

src/parser/parseSVGDocument.ts
asturur: is under the `isLikelyNode` so is either a xlmDom parser thing

src/parser/parseUseDirectives.ts
```diff
@@ -1,79 +1,81 @@
...
       ) {
         continue;
       }
 
-      if (attr.nodeName === 'transform') {
+      if (attr?.nodeName === 'transform') {
         currentTrans = attr.nodeValue + ' ' + currentTrans;
-      } else {
-        el2.setAttribute(attr.nodeName, attr.nodeValue);
+      } else if (attr) {
```
asturur, line 62: exactly, we can't wait line 62 to check attr, if we are worried that is false. either we do on top, or we trust is an attribute

src/parser/parseUseDirectives.ts
```diff
@@ -1,79 +1,81 @@
...
+    if (el.parentNode) {
+      parentNode = el.parentNode;
+      parentNode.replaceChild(el2, el);
     }
+    // TODO: ask about this
+    // please tell me if deleting/changing this will affect functionality
+    // // some browsers do not shorten nodeList after replaceChild (IE8)
+    // if (nodeList.length === oldLength) {
+    //   nodeIdx++;
+    // }
```
asturur, line 79: we can trash patches for buggy old browsers.
IE8 is not supported since long, i m not sure how JSDOM behaves, but we should have visual tests or tests for this case that would fail under node

test/unit/object.js
```diff
@@ -19,7 +19,7 @@
   });
 
   QUnit.test('constructor & properties', function(assert) {
-    assert.ok(typeof fabric.Object === 'function');
+    assert.ok(FabricObject === 'function');
```
asturur, line 22: this shouldn't be necessary for now

asturur:
The parser we have is bad, apart the hard code, because it works from inside outside.
It gets a list of valid svg tags that can be mapped to objects, and then it works its way out to the svg tag to understand what attributes get inherited from parents.
This completely breaks clipPaths on group and requires external code to keep group parsed.

Ideally we want something in reverse, that start from the svg tag and works its way in inspecting what it finds, if it finds a group tag with a clipPath, it can't pass the clipPath to the children, but will create a group that does that.

This require a full rewrite and can be done without deleting the old one.
It can be a nice effort if someone has time to spare, to create a new parallel svg parser that can slowly grow to replace the current one.
It requires an action plan but i m sure it would end up with way less code and easier to follow.

I imagine a recursive function that inspect a node, knows which attributes stay on the inspected node and which attributes get down to the children, and move down in the tree.

asturur:
also @Lazauya we don't need to type everything internally. Yes is good for us, but sometimes not. I don't want to write more complicated code because so that can be typed and TS can agree that is safe. That code will run at runtime and whatever weird data the element carry over it in the attributes won't care about our typiing and just freak out.

ShaMan123:
We can use AST to parse and walk the tree.
That alone should reduce code by half.

kristbp2:
> also @Lazauya we don't need to type everything internally. Yes is good for us, but sometimes not. I don't want to write more complicated code because so that can be typed and TS can agree that is safe. That code will run at runtime and whatever weird data the element carry over it in the attributes won't care about our typiing and just freak out.

I agree, the ideal would be to have simple code that uses types, but that can be hard.

> We can use AST to parse and walk the tree.
> That alone should reduce code by half.

I think this would solve @asturur's gripe about the types being overly complicated if we just start fresh.

I have a lot of experience with this kind of parsing, and I'd like to work on it if people are cool with that (as in, no one thinks I'd just be wasting my time). I think there are some minor things to work out before that though still, like a mixin for `fromElement` or whatever.

Lazauya:
> also @Lazauya we don't need to type everything internally. Yes is good for us, but sometimes not. I don't want to write more complicated code because so that can be typed and TS can agree that is safe. That code will run at runtime and whatever weird data the element carry over it in the attributes won't care about our typiing and just freak out.

I agree, the ideal would be to have simple code that uses types, but that can be hard.

> We can use AST to parse and walk the tree.
> That alone should reduce code by half.

I think this would solve @asturur's gripe about the types being overly complicated if we just start fresh.

I have a lot of experience with this kind of parsing, and I'd like to work on it if people are cool with that (as in, no one thinks I'd just be wasting my time). I think there are some minor things to work out before that though still, like a mixin for `fromElement` or whatever.

src/parser/applyViewboxTransform.ts
```diff
@@ -148,6 +173,7 @@ export function applyViewboxTransform(element) {
     el = element.ownerDocument.createElementNS(svgNS, 'g');
     // element.firstChild != null
     while (element.firstChild) {
```
Lazauya, line 175: This feels like a bug because we never change `element`. Is that done implicitly somehow? This confuses me.

PR #8394, opened 2022-10-24
ci(): TS test suite

ShaMan123:
## Motivation

#8392 

## Description

Added TS testing ability


ShaMan123 (commented):
READY!

TS tests fail as they should
I think we ought to add a `module` field to `package.json` and change the index file and exports to fix that

asturur:
I tried to resurrect and test this PR but now rollup doesn't allow to handle the build output destination as you wanted.

asturur:
if i clear the dist folder and i let it run in the dist folder, then it can run. It can't run in cli_output

PR #8395, opened 2022-10-25
ci(): bundle size report

ShaMan123:
## Motivation

@asturur !
We want to have a report showing the bundle size change in a PR

## Description

Findings from https://github.com/rollup/awesome:
- https://github.com/ritz078/rollup-plugin-filesize - prints bundle sizes only
- https://github.com/TrySound/rollup-plugin-size-snapshot - bundle sizes only but more detailed (accounts for treeshaking and more) and better in general
- https://github.com/btd/rollup-plugin-visualizer - visualizing module sizes of the project (doesn't seem helpful)
- https://github.com/doesdev/rollup-plugin-analyzer - per size analysis and for the whole bundle **BUT** read this https://github.com/doesdev/rollup-plugin-analyzer#why-is-the-reported-size-not-the-same-as-the-file-on-disk

I am going with `rollup-plugin-size-snapshot` but for the PR it doesn't matter so choose the one(s) you like
Maybe it is good to use `rollup-plugin-analyzer` only on the files changed in the PR using the `filter` option (I will try that as well)


asturur:
i see the changes to build for the comment.
Do we need a comment stripper before some step?

ShaMan123:
Maybe...
Or does terser have an option for that.
Pretty sure it does

ShaMan123:
https://github.com/terser/terser#format-options

image

ShaMan123:
Ahh I got it.
`rollup-plugin-analyzer` is misleading us. The comment is striped off or else we would have seen a diff to the dist file

asturur:
merging as it is, i want to see it on real PRs, this is 100% fine to iterate over

PR #8400, opened 2022-10-29
chore(TS): controls

ShaMan123:
## Description

Migrating controls:
- modularize
- type
- consumed `normalizePoint` into `getLocalPoint`

## Proposed Changes for Follow Up
- [ ] As discussed previously I want to make the Control class bound to an Object. This will make existing Control signatures redundant since `fabricObject` will be a class property. This means controls will no longer be shared, another over discussed topic I advocate. In purpose I didn't do it in this PR (as @asturur requested not to over refactor) so we can migrate most of controls with no major disagreements. Something worth looking at is that there are 2 signatures `XXX(eventData, control, fabricObject)` and `XXX(eventData, fabricObject, control)`. That is weird. But if we agree on bound controls both will become `XXX(eventData)`.  If not we should change them for consistency.
- [ ] Refactor wrappers to behave like middleware? mutating context and passing it on to the next action. This will remove any need to create wrappers in run time, seems to be easier to understand and reduces code.
```ts
const skewX = (...params) => [
   isLocked, // flag context somehow to disable next calls
   saveOrigin, // stores origin on context,
   prepareSkewingTransform, // mutates context
   skewObject, // uses the mutated context and set the result value
   restoreOrigin, // uses the mutated context
   fireEvent // uses the action result
].reduce((context, action) => action(context), prepareContext(...params))
```


src/controls/rotate.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  if (target.lockRotation) {
+    return false;
+  }
+
+  const lastAngle = Math.atan2(ey - pivotPoint.y, ex - pivotPoint.x),
+    curAngle = Math.atan2(y - pivotPoint.y, x - pivotPoint.x);
+  let angle = radiansToDegrees(curAngle - lastAngle + theta);
+
+  if (target.snapAngle > 0) {
```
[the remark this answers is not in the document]
ShaMan123: Blame points to #3383

https://github.com/fabricjs/fabric.js/commit/10703055c0895b2bf0745c07da6024308322fda7#commitcomment-88268859

src/controls/control.class.ts
```diff
@@ -1,372 +1,395 @@
...
+    ctx: CanvasRenderingContext2D,
+    left: number,
+    top: number,
+    styleOverride: ControlRenderingStyleOverride | undefined,
+    fabricObject: FabricObject
+  ) {
+    styleOverride = styleOverride || {};
+    switch (styleOverride.cornerStyle || fabricObject.cornerStyle) {
+      case 'circle':
+        renderCircleControl.call(
```
ShaMan123, line 373: I want to make the controls renders bound to the class as well one day

src/controls/controls.render.ts
```diff
@@ -30,15 +52,15 @@ export function renderCircleControl(
...
+  let myLeft = left,
     myTop = top,
     size;
   ctx.save();
-  ctx.fillStyle = styleOverride.cornerColor || fabricObject.cornerColor;
+  ctx.fillStyle = styleOverride.cornerColor || fabricObject.cornerColor || '';
   ctx.strokeStyle =
-    styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor;
-  // as soon as fabric react v5, remove ie11, use proper ellipse code.
+    styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor || '';
```
ShaMan123, line 62: This is a change I need to verify is not breaking
Done only to silence TS

src/shapes/object.class.ts
```diff
@@ -16,9 +16,7 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
 import { createCanvasElement } from '../util/misc/dom';
 import { ObjectGeometry } from '../mixins/object_geometry.mixin';
 import { qrDecompose, transformPoint } from '../util/misc/matrix';
-
-type StaticCanvas = any;
-type Canvas = any;
+import { Canvas, Shadow, StaticCanvas } from '../__types__';
 
 const ALIASING_LIMIT = 2;
```
ShaMan123, line 21: I want to move this to config or constants
Where does it belong?

ShaMan123 (commented):
I still have some things I need to fix and to work on:
- [x] tests and build
- [ ] skew: port #8398 
- [ ] extract lockXXX to a wrapper


But I would like your input so please review and comment

ShaMan123:
Updated description. Please read and comment

src/controls/rotate.ts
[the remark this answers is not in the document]
ShaMan123: Looking at #3383 it seems it was intended to be on FabricObject and that makes sense.
If so, should we add these values to json import/export?

ShaMan123:
I don't know why build is passing
Tests fail with fabric is not defined, build defines fabric$1

src/canvas.class.ts
```diff
@@ -738,21 +741,6 @@ import { FabricObject } from './shapes/object.class';
...
-       * @param {Event} e Event object
-       * @param {fabric.Object} [target] inserted back to help overriding. Unused
-       */
-      _getActionFromCorner: function (alreadySelected, corner, e, target) {
-        if (!corner || !alreadySelected) {
-          return 'drag';
-        }
-        var control = target.controls[corner];
-        return control.getActionName(e, control, target);
-      },
```
asturur, line 754: I don't remember the reason of this function.
It says something about overriding, and if is an utile imported directly can't be overridden.
I ll dig in the past to understand why this should or could be overridden.

src/controls/control.class.ts
```diff
@@ -1,372 +1,392 @@
...
-     * if the method `cursorStyleHandler` is provided, this property is ignored.
-     * @type {String}
-     * @default 'crosshair'
-     */
-    cursorStyle: 'crosshair',
+  constructor(options: Partial<Control>) {
+    for (const key in options) {
+      // @ts-expect-error TS fails to understand the assignment
+      this[key] = options[key];
+    }
```
asturur, line 147: but why can't we have `Object.assign(this, options)` ? why do we need a loop?

src/controls/control.class.ts
```diff
@@ -1,372 +1,394 @@
...
+  /**
+   * The control handler for mouse down, provide one to handle mouse down on control
+   * @param {Event} eventData the native mouse event
+   * @param {Object} transformData properties of the current transform
+   * @param {Number} x x position of the cursor
+   * @param {Number} y y position of the cursor
+   * @return {Boolean} true if the action/event modified the object
+   */
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  mouseDownHandler: TransformAction = () => {};
```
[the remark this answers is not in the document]
asturur: can't assign fat arrows in this way, they end up being public fields instead of part of the prototype.

src/controls/control.class.ts
```diff
@@ -1,372 +1,392 @@
...
+  getVisibility(fabricObject: FabricObject, controlKey: string) {
+    // @ts-expect-error TODO remove directive once fixed
+    const objectVisibility = fabricObject._controlsVisibility;
+    if (
+      objectVisibility &&
+      typeof objectVisibility[controlKey] !== 'undefined'
+    ) {
+      return objectVisibility[controlKey];
+    }
+    return this.visible;
```
asturur, line 274: i think this is one of the few cases that is really equivalent to
```js
return objectVisibility?.[controlKey] ?? this.visible
```

src/controls/control.class.ts
```diff
@@ -1,372 +1,392 @@
...
+  positionHandler(
+    dim: Point,
+    finalMatrix: TMat2D,
+    fabricObject: FabricObject,
+    currentControl: Control
+  ) {
+    return new Point(this.x, this.y)
+      .multiply(dim)
+      .add(new Point(this.offsetX, this.offsetY))
+      .transform(finalMatrix);
```
asturur, line 295: the old code was better.
This creates 5 points... the old one was really smpler 

```js
return transformPoint(
        {
          x: this.x * dim.x + this.offsetX,
          y: this.y * dim.y + this.offsetY,
        },
        finalMatrix
      );
```

src/controls/control.class.ts
asturur: The render function isn't bound just because i wanted to handle the old cornerstyle and keep 2 functions separated. `render` could have been just another option to add to the controls deprecating the `cornerStyle ` option in the  object.

It would be more logic to have a predefined style for controls without any type/corner/color and all the rest is done on the control itself.

src/controls/control.class.ts
ShaMan123: we can do that

src/controls/control.class.ts
ShaMan123: I wonder if it makes any difference...
I'll change

src/controls/controls.render.ts
```diff
@@ -30,15 +47,15 @@ export function renderCircleControl(
...
+      (styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor);
+  let myLeft = left,
     myTop = top,
     size;
   ctx.save();
   ctx.fillStyle = styleOverride.cornerColor || fabricObject.cornerColor;
   ctx.strokeStyle =
     styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor;
-  // as soon as fabric react v5, remove ie11, use proper ellipse code.
+  // TODO: as soon as fabric react v5, remove ie11, use proper ellipse code.
```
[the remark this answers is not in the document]
asturur: we can use ctx.drawEllipse instead of drawArc probably. We can do that.

src/controls/controls.render.ts
```diff
@@ -69,18 +91,20 @@ export function renderCircleControl(
...
+ * @param {Object} styleOverride override for FabricObject controls style
+ * @param {FabricObject} fabricObject the fabric object for which we are rendering controls
  */
 export function renderSquareControl(
-  ctx,
-  left,
-  top,
-  styleOverride,
-  fabricObject
+  this: Control,
```
asturur, line 98: i think this `this` isn't needed. is it? we are calling this with `call`

src/canvas.class.ts
ShaMan123: makes sense
so you can override to which control you are passing the event..?

src/controls/default_controls.ts
```diff
@@ -1,117 +1,122 @@
...
+} from './actions';
+import { Control } from './control.class';
 
-  objectControls.ml = new fabric.Control({
+export const defaultControls = {
+  ml: new Control({
     x: -0.5,
     y: 0,
-    cursorStyleHandler: scaleSkewStyleHandler,
+    cursorStyleHandler: scaleSkewCursorStyleHandler,
```
asturur, line 22: We can't rename those. are still part of the public api. Is not that the name is so bad that needs to be changed.

src/controls/control.class.ts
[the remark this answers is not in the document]
ShaMan123: so I'll make it optional

src/controls/default_controls.ts
```diff
@@ -1,117 +1,122 @@
...
-})(typeof exports !== 'undefined' ? exports : window);
+// this is breaking the prototype inheritance, no time / ideas to fix it.
+// is important to document that if you want to have all objects to have a
+// specific custom control, you have to add it to Object prototype and to Textbox
+// prototype. The controls are shared as references. So changes to control `tr`
+// can still apply to all objects if needed.
+fabric.Textbox.prototype.controls = {
+  ...(fabric.Textbox.prototype.controls || {}),
+  ...textboxDefaultControls,
+};
```
asturur, line 122: This isn't anymore under `if (fabric.Textbox)` meaning that non importing the textbox will create an error.
Let's keep it under if until we can move this in the Textbox file directly.

src/controls/control.class.ts
ShaMan123: `objectVisibility?.[controlKey] ` 
this is new syntax for me

src/controls/default_controls.ts
ShaMan123: Ahh I messed up exports?

src/controls/rotate.ts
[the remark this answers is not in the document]
asturur: in json we don't import export behaviour data.
snapAngle was supposed to be on object, i remember that ( i think was contributed by stefan a friend of mine ), it was on object because we used to do fabric.Object.prototype.snapAngle = and set it to al objects.
Is also true that having it per object offers more flexibility ( imagine you want to rotate rects only by 90 degree, or you can have weird applications i don't want to dictate what are you doing with snapping )
So yes we could add it to the Interactivity mixin list of properties

src/controls/rotate.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  // normalize angle to positive value
+  if (angle < 0) {
+    angle = 360 + angle;
+  }
+  angle %= 360;
+
+  const hasRotated = target.angle !== angle;
+  // TODO: why aren't we using set?
+  target.angle = angle;
```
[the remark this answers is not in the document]
asturur: why do we need set? is this an old comment or a new one?

src/controls/controls.render.ts
ShaMan123: It is for TS to cast the this context of the method
Doesn't affect anything else AFAIK
I think these methods belong in the control class

src/controls/util.ts
```diff
@@ -0,0 +1,134 @@
...
+            'center',
+            'center',
+            originX,
+            originY
+          )
+        : new Point(target.left, target.top),
+    p2 = target.angle
+      ? point.rotate(-degreesToRadians(target.angle), center)
+      : point,
+    localPoint = p2.subtract(p);
```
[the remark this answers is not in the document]
asturur: can we leave it as an non exported function inside this util file?
could happen that is needed for something else later is packaged as a unit

src/controls/util.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+  if (localPoint.x <= -padding) {
+    localPoint.x += padding;
+  }
+  if (localPoint.y >= padding) {
+    localPoint.y -= padding;
+  }
+  if (localPoint.y <= padding) {
+    localPoint.y += padding;
+  }
```
asturur, line 129: i hate the complications from padding

src/controls/rotate.ts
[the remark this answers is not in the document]
ShaMan123: I wrote it

src/mixins/object_geometry.mixin.ts
```diff
@@ -674,10 +675,7 @@ export class ObjectGeometry extends ObjectOrigin {
    * @return {TMat2D}
    */
   getViewportTransform(): TMat2D {
-    if (this.canvas && this.canvas.viewportTransform) {
-      return this.canvas.viewportTransform;
-    }
-    return iMatrix.concat() as TMat2D;
+    return this.canvas?.viewportTransform || (iMatrix.concat() as TMat2D);
```
asturur, line 678: i m not sure we are transpiling those yet.
But the issue is that if undefined is not involved those becomes longer.
`if typeof canvas !== 'undefined'` rather than a simple `if this.canvas`

src/controls/control.class.ts
ShaMan123: Feels like you didn't answer why not bound but why separated.
I suggest adding them as private methods as is.

asturur:
Seems all good to me apart:
- the extra this argument in the renderCircle and renderSquare control
- the renaming of the cursor style functions
- i have to look into that getActionFromCorner

src/controls/control.class.ts
ShaMan123: seems to me they fall in the exact pattern of a class member

src/mixins/object_geometry.mixin.ts
ShaMan123: please elaborate
I lack the knowledge

src/controls/rotate.ts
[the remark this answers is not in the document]
ShaMan123: Maybe you should do that in #8401 so the conflict does rise?

src/controls/rotate.ts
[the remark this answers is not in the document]
ShaMan123: For the sake of consistency... The rest of the actions use set and allow the dev to do more customization in the `_set` method.

src/controls/controls.actions.ts
```diff
@@ -1,783 +0,0 @@
...
-    moveY && target.set('top', newTop);
-    if (moveX || moveY) {
-      fireEvent('moving', commonEventInfo(eventData, transform, x, y));
-    }
-    return moveX || moveY;
-  }
-
-  controls.scaleCursorStyleHandler = scaleCursorStyleHandler;
-  controls.skewCursorStyleHandler = skewCursorStyleHandler;
-  controls.scaleSkewCursorStyleHandler = scaleSkewCursorStyleHandler;
```
ShaMan123, line 748: I haven't messed up exports

src/controls/default_controls.ts
ShaMan123: https://github.com/fabricjs/fabric.js/pull/8400/files/d435c6b4a0dc1682aa74ee5c97a2710cb0f86e41#r1008768595

asturur:
Regarding the tasks you proposed , piping the function is a well known pattern as good as another, there is no reason to change it, it allows for reusability, keeps functions with a single concerns, is newish ( added in 4.0 ) i don't see the reason to change it, is just added chaos.

Regarding sharing or no, that is really a no.
You can still make NON shared controls if you want, creating them new for each object. but if you embed that in the code in the library you can't do the contrary anymore. Why add this limitation?

I m totaly happy with a control set that is able to render itslef moving from object to object rather than many copies of it one per object.

src/controls/default_controls.ts
ShaMan123: They were renamed in the file as local vars
I simply undid that

asturur: ohhh ok i didn't see that.

ShaMan123:
> Regarding sharing or no, that is really a no.

We will remain disagreed on this topic
The why is because it is very unintuitive. The same as the discussion we had on default values. Exactly IMO. A dev wanting a change to propagate should loop over existing objects. Elsewise it is a reactive mobx kind of thing.
I view controls as part of Object, you view it as a separate layer.
A dev mutating that object will never understand that he changed the entire app.

src/mixins/object_geometry.mixin.ts
ShaMan123: What is the diff between this syntax and `obj?.[key]` ?

src/controls/rotate.ts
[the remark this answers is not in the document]
ShaMan123: 08a4bab27

src/controls/rotate.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  if (target.lockRotation) {
+    return false;
+  }
+
+  const lastAngle = Math.atan2(ey - pivotPoint.y, ex - pivotPoint.x),
+    curAngle = Math.atan2(y - pivotPoint.y, x - pivotPoint.x);
+  let angle = radiansToDegrees(curAngle - lastAngle + theta);
+
+  if (target.snapAngle && target.snapAngle > 0) {
```
ShaMan123, line 58: 214bc5202

ShaMan123 (commented):
Added `isLocked` util, not very useful but at least it makes it simple to find all the locking logic
Extracting to a wrapper will cause too much unnecessary code changes

ShaMan123:
> > Regarding sharing or no, that is really a no.
> 
> We will remain disagreed on this topic The why is because it is very unintuitive. The same as the discussion we had on default values. Exactly IMO. A dev wanting a change to propagate should loop over existing objects. Elsewise it is a reactive mobx kind of thing. I view controls as part of Object, you view it as a separate layer as far as I can tell. A dev mutating that object will never understand that he changed the entire app.

@asturur I have a compromise.
If we expose a class `ControlSet` that is basically a set of controls.
The class will hold the original control set ref as follows:
```ts
class ControlSet {
  defaultControls: Record<string, Control>;
  controls: Record<string, Control>;

  constructor(controls: Record<string, Control> = {}) {
    this.defaultControls = controls;
  }
  getControl(key: string) {
    return this.controls[key] || this.defaultControls[key];
  }
  setControl(key: string, control: Control) {
    this.controls[key] = control;
  }
}
```

This is a win win solution. 
For your side no one will touch this ControlSet class, but continue as previously by mutating the ref that was passed to the ControlSet (and if necessary could pass a new ControlSet to the object). 
For my side the dev can use `setControl` and is safeguarded from unwanted sharing. No need for a gotcha/documentation/issues.
We keep what you want and make it intuitive.
I hope I convinced you.
Awaiting your response.

It also allows use to move control logic from interactivity to the ControlSet since it will be bound to the Object.
Things like control visibility etc.

ShaMan123:
> And I can fix the derivation of textbox controls with this approach

A win-win-win you-me-textbox

```ts
class ControlSet {
  defaultControls: Record<string, Control> | ControlSet;
  controls: Record<string, Control>;

  constructor(controls: Record<string, Control> | ControlSet = {}) {
    this.defaultControls = controls;
  }
  getControl(key: string): Control {
    return (
      this.controls[key] ||
      (this.defaultControls instanceof ControlSet
        ? this.defaultControls.getControl(key)
        : this.defaultControls[key])
    );
  }
  setControl(key: string, control: Control) {
    this.controls[key] = control;
  }
}

const objectControls = new ControlSet(defaultControls);
const textboxControls = new ControlSet(objectControls);
textboxControls.setControl('mr', new Control({}));
objectControls.setControl('a', new Control({}));
textboxControls.getControl('a');
```

ShaMan123:
It also allows use to move control logic from interactivity to the ControlSet since it will be bound to the Object.
Things like control visibility etc.

```ts
class BoundObjectControls {
  constructor(controlSet: ControlSet, object: FabricObject) {
    this.controls = controlSet;
    this.object = object;
  }
  getVisibility(...) {
    
  }
  setVisibility(...) {
    
  }
}

```

ShaMan123:
we can use setter/getter to make it non breaking if that is crucial

```ts
class ControlSet {
  defaultControls: Record<string, Control> | ControlSet;
  controls: Record<string, Control>;

  constructor(controls: Record<string, Control> | ControlSet = {}) {
    this.defaultControls = controls;
  }
  get(key: string): Control {
    return (
      this.controls[key] ||
      (this.defaultControls instanceof ControlSet
        ? this.defaultControls.getControl(key)
        : this.defaultControls[key])
    );
  }
  set(key: string, control: Control) {
    this.controls[key] = control;
  }
}
```

ShaMan123 (commented):
Thinking of defaultControls
I think they should be assigned in the interactivity mixin.
And that textbox should subclass the mixin.
And that we use applyMixins for it instead of subclassing

Not sure though
Thoughts?

asturur:
@ShaMan123 i want to do the TS migration, i don't want to do refactors. There is a lot of work to do and the more we do because we think is necessary, the later we can go back to features and website and docs

I disagree on TS and classes the more i write them. peace.

If we want to finish the ts migration we stick to that, we take notes on what we would like to see changed, we organize them and for now we move on head down on the goal

Even the isLocked wrapper wasn't an idea i would have chased now.

ShaMan123:
I understand.
These are notes for us to discuss.
Ideas.
Do you prefer I put them elsewhere? Do you prefer not hearing them?
I don't understand

ShaMan123:
So if I understood you, my comments are throwing you off.
So if I put everything in a discussion(s) that what be better?

asturur:
> So if I understood you, my comments are throwing you off. So if I put everything in a discussion(s) that what be better?

How can i know that you put them in the PR because you want to discuss them somewhere else another day?
If i see them in the PR i think you want to address them now.

src/mixins/object_geometry.mixin.ts
asturur: none, but where we use obj?.[key] we were already looking for undefined ( ex: the overrideStyle object ). Here we want just to check if canvas is falsy in any way.

```
obj.first?.second

const temp = obj.first;
const nestedProp =
  temp === null || temp === undefined ? undefined : temp.second;
```

If we land on supported browsers that do not support the `?.` all those equality check get added in the codebase. I just don't like to think all that stuff gets added in.

Not a big issue anyway

src/controls/controls.render.ts
asturur: so is not a real argument? i can call the function like `renderSquareControl(ctx, left... )` ?

asturur: I see when built it gets removed.
Weird that being a type annotation looks like a real argument.
How do someone know just reading at the code?

asturur:
let's skip getActionFromCorner check up and let's merge this.
I checked and i don't have much memories apart that it was a slimdown of the previous function that was there since long.
The idea was to move Drag in the realm of controls too, so this is probably a safe change

asturur (approved):

src/controls/controls.render.ts
ShaMan123: Weird indeed

PR #8401, opened 2022-10-29
Chore(TS) Object interactivity

asturur:
## Motivation
Continuing conversiont to TS.

## Description

Continue the TS migration, change to imports.
For now the mixin is not optional.

Ideally we can split in this way:

importing static canvas adds the stacking mixin
importing canvas adds staticCanvas and the interactivity mixin + controls.
I m not yet sure how the optional import is supposed to work

## Changes
- converted mixin to Class
- divided setCoord in the interactive vs non interactive part
- moved oCoords to ObjectInteractivity
- moved calcOCoords to ObjectInteractivity
- moved _corner to ObjectInteractivity

BREAKING:
- removed more chainability
- changed default value of _corner from 0 to '' ( to avoid the weird typing )
- removed the skipCornerCoord from setCoords, that was a silly performance optimization 
-


src/util/applymixins.ts
```diff
@@ -0,0 +1,12 @@
...
+    Object.getOwnPropertyNames(baseCtor.prototype).forEach((name) => {
+      Object.defineProperty(
+        derivedCtor.prototype,
+        name,
+        Object.getOwnPropertyDescriptor(baseCtor.prototype, name) ||
+          Object.create(null)
+      );
+    });
+  });
+}
```
asturur, line 12: unused. i commited because i wanted to use it, we can keep it and then trash it if is not useful

src/shapes/object.class.ts
```diff
@@ -2133,8 +2128,3 @@ const fabricObjectDefaultValues: TClassProperties<FabricObject> = {
 };
 
 Object.assign(FabricObject.prototype, fabricObjectDefaultValues);
-
-(function (global) {
-  const fabric = global.fabric;
-  fabric.Object = FabricObject;
-})(typeof exports !== 'undefined' ? exports : window);
```
asturur, line 2140: moved this to InteractiveFabricObject

src/shapes/object.class.ts
```diff
@@ -625,6 +618,8 @@ export class FabricObject extends ObjectGeometry {
    */
   static __uid = 0;
 
+  callSuper?: TCallSuper;
+
```
asturur, line 622: this is a compatibility layer for classes that now import extended classes that are not created with fabric.createClass

src/shapes/object.class.ts
```diff
@@ -122,15 +125,15 @@ export class FabricObject extends ObjectGeometry {
...
+  cornerColor: string;
 
   /**
    * Color of controlling corners of an object (when it's active and transparentCorners false)
    * @since 1.6.2
    * @type String
    * @default null
    */
-  cornerStrokeColor: string | null;
+  cornerStrokeColor: string;
```
asturur, line 136: empty string has the same effect of null, so we can reduce to string and keep it compatible with the embedded types of ctx.fillStyle and ctx.strokeStyle

src/shapes/fabricObject.class.js
```diff
@@ -0,0 +1,12 @@
...
+} from "../mixins/object_interactivity.mixin";
+
+// TODO somehow we have to make a tree-shakeable import
+
+export { InteractiveFabricObject as FabricObject }
+
+(function (global) {
+  const fabric = global.fabric;
+  fabric.Object = InteractiveFabricObject;
+})(typeof exports !== 'undefined' ? exports : window);
```
asturur, line 12: if we don't find a better way, this together with the dead code remover plugin for rollup can do the optional imports.

asturur:
@ShaMan123 @melchiar this is ready to go

src/mixins/object_interactivity.mixin.ts
```diff
@@ -1,394 +1,497 @@
...
+  /**
+   * Sets corner and controls position coordinates based on current angle, width and height, left and top.
+   * oCoords are used to find the corners
+   * aCoords are used to quickly find an object on the canvas
+   * lineCoords are used to quickly find object during pointer events.
+   * See {@link https://github.com/fabricjs/fabric.js/wiki/When-to-call-setCoords} and {@link http://fabricjs.com/fabric-gotchas}
+   * @return {void}
+   */
+  setCoords(): void {
+    if (this.callSuper) {
```
ShaMan123, line 174: please explain why this is needed
transient until everything is migrated?
I don't understand why needed since everything funnels through FabricObject

src/mixins/object_interactivity.mixin.ts
```diff
@@ -1,394 +1,497 @@
...
+    ctx.strokeRect(-size.x / 2, -size.y / 2, size.x, size.y);
+  }
+
+  /**
+   * @private
+   * @param {CanvasRenderingContext2D} ctx Context to draw on
+   * @param {Point} size
+   * @param {Object} styleOverride object to override the object style
+   */
+  _drawBorders(ctx: CanvasRenderingContext2D, size: Point, styleOverride: Record<string, any> = {}): void {
```
ShaMan123, line 267: private?

src/shapes/fabricObject.class.js
ShaMan123: I am currently for the several entry points approach
```ts
import { Canvas, Text } from 'fabric/node';
import { Canvas, Text } from 'fabric/static'; // or other name
import { Canvas, Text } from 'fabric'; // interactive by default
```

src/shapes/object.class.ts
ShaMan123: good
I came across that in #8400  as well

src/mixins/object_interactivity.mixin.ts
```diff
@@ -1,394 +1,497 @@
...
+  /**
+   * keeps the value of the last hovered corner during mouse move.
+   * 0 is no corner, or 'mt', 'ml', 'mtr' etc..
+   * It should be private, but there is no harm in using it as
+   * a read-only property.
+   * this isn't cleaned automatically. Non selected objects may have wrong values
+   * @type number|string|any
+   * @default 0
+   */
+  __corner: number | string;
```
ShaMan123, line 38: why not make all 0 assignments null or undefined?

ShaMan123 (commented):
Did a superficial review sine the diff is not readable
Looks good
A few minor comments/questions

ShaMan123:
I see you missed prettier and changelog, they are failing

src/shapes/fabricObject.class.js
asturur: well the point is that Rect and Circle have to extend either from Object or Interactive Object, so that is why i get confused on how i change that throug imports.
Yes we can have different entry points that create 3 bundles and people that use the bundle entirely can do that, or they can just import the single TS files as they see fit. But at that point, you have to provide 2 version of Rect/Circle/Text...?

Till i m not done converting i ll ignore it

src/mixins/object_interactivity.mixin.ts
asturur: fabricCreate.class is assembling a new class from InteractiveObject and creating a function with prototype.
At that point `super.setCoords()` does not work anymore because the method gets called in a context that is not an es6 class. checking if callSuper exists is done to understand if we are using a createClass construct.

src/mixins/object_interactivity.mixin.ts
asturur: maybe? not sure, at some point we have to decide what becomes private for the sake of reducing exposure Not sure if someone would want to call _drawBorders directly

src/mixins/object_interactivity.mixin.ts
asturur: yes i thought the same, we could move it to 0 as default rather than null. I don't think anyone is gonna cry for that.

src/shapes/fabricObject.class.js
ShaMan123: A headache indeed
So Mixins, all of them

src/mixins/object_interactivity.mixin.ts
```diff
@@ -186,7 +204,9 @@ export class InteractiveFabricObject extends FabricObject {
    * with the control, the control's key and the object that is calling the iterator
    * @param {Function} fn function to iterate over the controls over
    */
-  forEachControl(fn: (arg0: any, arg1: string, arg2: InteractiveFabricObject) => any) {
+  forEachControl(
+    fn: (arg0: any, arg1: string, arg2: InteractiveFabricObject) => any
```
ShaMan123, line 208:
```suggestion
    fn: (control: any, key: string, fabricObject: InteractiveFabricObject) => any
```

src/mixins/object_interactivity.mixin.ts
ShaMan123: I am confused
I thought removing 0 in favor of null/undefined

asturur: sorry nor zero nor null. just empty string or optional with undefined.

i din't like null it often makes no sense to me.

src/shapes/fabricObject.class.js
asturur: Yes but it stays complicated mixin or class is still a bit complicated.

When we have a full ts folder we can worry about thay

src/mixins/object_interactivity.mixin.ts
ShaMan123: I am asking since I saw the code still has `__corner = 0`

asturur: either empty string or undefined i suppose. With optional string ( and so undefined ) looking more clean

PR #8405, opened 2022-10-31
chore(TS): more FabricObject types

ShaMan123:
## Description
More types
Left Gradient/Pattern types alone since there are open PRs for them, I don't want to create conflicts
Exposed `getCanvasRetinaScaling` on ObjectGeometry for ease of use


src/mixins/object_geometry.mixin.ts
```diff
@@ -602,6 +602,10 @@ export class ObjectGeometry extends ObjectOrigin {
     return this.scale(value / this.height / boundingRectFactor);
   }
 
+  getRetinaScaling() {
+    return this.canvas?.getRetinaScaling() || 1;
```
ShaMan123, line 606: exposed this instead of having to check if canvas exists
I wonder if we should change this to
`return this.canvas?.getRetinaScaling() || config.devicePixelRatio;`

Should the method be protected?

ShaMan123 (commented):
READY

src/shapes/object.class.ts
```diff
@@ -1402,13 +1408,13 @@ export class FabricObject extends ObjectGeometry {
...
-  _setLineDash(ctx, dashArray) {
+  _setLineDash(ctx: CanvasRenderingContext2D, dashArray?: number[] | null) {
     if (!dashArray || dashArray.length === 0) {
       return;
     }
-    // Spec requires the concatenation of two copies the dash list when the number of elements is odd
+    // Spec requires the concatenation of two copies of the dash array when the number of elements is odd
     if (1 & dashArray.length) {
-      dashArray.push.apply(dashArray, dashArray);
+      dashArray.push(...dashArray);
```
ShaMan123, line 1417: Should we mutate the arg?
Or should we instead
```ts
ctx.setLineDash(1 & dashArray.length ? [...dashArray, ...dashArray]: dashArray);
```

src/pattern.class.ts
```diff
@@ -144,7 +144,7 @@ export class Pattern {
    * @param {Array} [propertiesToInclude] Any properties that you might want to additionally include in the output
    * @return {object} Object representation of a pattern instance
    */
-  toObject(propertiesToInclude: (keyof this)[]) {
+  toObject(propertiesToInclude?: (keyof this)[]) {
```
asturur, line 147: pretty much sure will have to go back to Record<string, any> for this, but for now let's try

src/constants.ts
```diff
@@ -8,6 +8,7 @@ export const twoMathPi = Math.PI * 2;
 export const PiBy180 = Math.PI / 180;
 export const iMatrix = Object.freeze([1, 0, 0, 1, 0, 0]) as TMat2D;
 export const DEFAULT_SVG_FONT_SIZE = 16;
+export const ALIASING_LIMIT = 2;
```
asturur, line 11: added here but removed nowhere?

src/mixins/object_geometry.mixin.ts
```diff
@@ -602,6 +602,10 @@ export class ObjectGeometry extends ObjectOrigin {
...
+  }
```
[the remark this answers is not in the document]
asturur: (but at least should be called `getCanvasRetinaScaling`)

asturur (dismissed):

src/mixins/object_geometry.mixin.ts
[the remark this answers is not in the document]
ShaMan123: Naming: good point
In half of the cases the check in needed so it seems.
The other we can assert Or use `!` or what you suggested (need to think how)

src/constants.ts
ShaMan123: moved from Object

src/mixins/object_geometry.mixin.ts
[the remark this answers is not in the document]
ShaMan123: So I will rename and we merge?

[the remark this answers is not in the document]
ShaMan123: Well OK maybe if we do like I did casting `this: this & Required<this, 'canvas'>`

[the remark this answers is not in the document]
ShaMan123: renamed

src/shapes/object.class.ts
ShaMan123: from the tests I see this is intentional

ShaMan123 (commented):
ready

src/shapes/object.class.ts
asturur: Yeah in theory we shouldn't and we should modify it every single time.
But also eventually we can do what we want with property assigned to an instance?
The intention was to not redo it every time

asturur (approved):

PR #8406, opened 2022-10-31
chore(TS): Circle

ShaMan123:
## Description

Migrate shapes using the transformer

## Changes

- `SHARED_ATTRIBUTES` moved to parser
- `setRadius` not chainable

## Gist

View the diff without whitespace!

### DIFF
--- a/src/shapes/circle.class.ts
+++ b/src/shapes/circle.class.ts
@@ -6,12 +6,6 @@ import { degreesToRadians } from '../util/misc/radiansDegreesConversion';
 import { sin } from '../util/misc/sin';
 import { FabricObject } from './fabricObject.class';
 
-/**
- * Circle class
- * @class Circle
- * @extends fabric.Object
- * @see {@link Circle#initialize} for constructor definition
- */
 export class Circle extends FabricObject {
   /**
    * Radius of this circle
@@ -40,7 +34,6 @@ export class Circle extends FabricObject {
    * @private
    * @param {String} key
    * @param {*} value
-   * @return {Circle} thisArg
    */
   _set(key: string, value: any) {
     super._set(key, value);
--- a/src/shapes/circle.class.ts
+++ b/src/shapes/circle.class.ts
@@ -134,7 +134,7 @@ export class Circle extends FabricObject {
         '<path d="M ' + startX + ' ' + startY,
         ' A ' + radius + ' ' + radius,
         ' 0 ',
-        +largeFlag + ' 1',
+        `${largeFlag} 1`,
         ` ${endX} ${endY}`,
         '" ',
         'COMMON_PARTS',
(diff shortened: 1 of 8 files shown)


src/shapes/circle.class.ts
```diff
@@ -1,232 +1,223 @@
...
+      const { radius } = this;
+      const start = degreesToRadians(this.startAngle),
+        end = degreesToRadians(this.endAngle),
+        startX = cos(start) * radius,
+        startY = sin(start) * radius,
+        endX = cos(end) * radius,
+        endY = sin(end) * radius,
+        largeFlag = angle > 180 ? '1' : '0';
+      return [
+        '<path d="M ' + startX + ' ' + startY,
```
asturur, line 134:
```suggestion
        `<path d="M ${startX} ${startY}`,
```

asturur: those we can change. Probably there will be a nicer way to get rid of those arrays in future too

src/shapes/circle.class.ts
```diff
@@ -1,232 +1,223 @@
...
+   * @returns {Promise<Circle>}
+   */
+  static fromObject(object: object): Promise<Circle> {
+    return FabricObject._fromObject(Circle, object);
+  }
+}
+
+export const circleDefaultValues: Partial<TClassProperties<Circle>> = {
+  type: 'circle',
+  radius: 0,
```
asturur, line 206: should we bump radius to 100 as default? creating circles that are unusable by default seems just wrong

src/shapes/circle.class.ts
```diff
@@ -1,232 +1,223 @@
...
+      const start = degreesToRadians(this.startAngle),
+        end = degreesToRadians(this.endAngle),
+        startX = cos(start) * radius,
+        startY = sin(start) * radius,
+        endX = cos(end) * radius,
+        endY = sin(end) * radius,
+        largeFlag = angle > 180 ? '1' : '0';
+      return [
+        '<path d="M ' + startX + ' ' + startY,
+        ' A ' + radius + ' ' + radius,
```
asturur, line 135:
```suggestion
        ` A ${radius} ${radius}`,
```

src/shapes/circle.class.ts
```diff
@@ -1,232 +1,223 @@
...
+        ' 0 ',
+        +largeFlag + ' 1',
```
asturur, line 137:
```suggestion
        `${largeFlag} 1`,
```

src/shapes/circle.class.ts
```diff
@@ -1,232 +1,223 @@
...
+        ' ' + endX + ' ' + endY,
```
asturur, line 138:
```suggestion
        ` ${endX} ${endY}`,
```

src/shapes/circle.class.ts
ShaMan123: agreed

asturur:
not bumping radius now to avoid breaking UTs, we can do next time. if it gets green i merge this

asturur (approved):

PR #8409, opened 2022-10-31
ci(): fix transformer

ShaMan123:
## Description

Bunch of fixes to accommodate the agreed migration pattern

## Gist

```bash
npm run cli transform -- -o -ts 
```


ShaMan123 (commented):
Just the transformer + dev deps

ShaMan123 (commented):
I have transformed almost all files with it
So it seems stable apart from missing parens in some callSuper replacements
I suggest using it
It makes it fun simple quick and easy

READY

ShaMan123:
The transformer has ran its course

PR #8412, opened 2022-10-31
chore(TS): Path

ShaMan123:


ShaMan123 (commented):
ready

asturur (approved):
removed the changes in setDimensions/setBoundingBox just to keep this as pure migraiton.

ShaMan123:
So I should PR the reverted commit since now Path and polys have diverged in behavior.

PR #8414, opened 2022-10-31
chore(TS): object mixins

ShaMan123:
## Description

This PR migrates all remaining object mixins.
I had to split files that defined both canvas and object mixins. The canvas mixins were left untouched.
I tried a different approach in this PR, using `applyMixins`.
My conclusion is that we should avoid it. It is fragile and TS doesn't stomach it well.
I prefer the approach of collection mixin or chaining classes.
For dynamic protos like with SVG `applyMixins` is fine.

## Gist

As long as a mixin doesn't have a constructor or a super class applyMixins works as expected
I had to disable constructor assigning since it overrides the prev.
But I guess there is a way to merge them, I can't think of a mixin which needs it


asturur:
i have a bunch of questions about the overlap of ancestry and stacking.
Why does stacking needs to extends ancestry?

Does ancestry have any usefulness outside the interactivity with activeSelection and group in general?

ShaMan123:
> Why does stacking needs to extends ancestry?

That was a bad attempt I forgot to revert

src/shapes/fabricObject.class.ts
```diff
@@ -0,0 +1,28 @@
...
+import { FabricObjectAncestryMixin } from '../mixins/object_ancestry.mixin';
+import { FabricObjectObjectAnimationMixin } from '../mixins/object_animation.mixin';
+import { InteractiveFabricObject } from '../mixins/object_interactivity.mixin';
+import { FabricObjectObjectStackingMixin } from '../mixins/object_stacking.mixin';
+import { FabricObjectObjectStraighteningMixin } from '../mixins/object_straightening.mixin';
+import { applyMixins } from '../util/applyMixins';
+
+// TODO somehow we have to make a tree-shakeable import
+
+applyMixins(InteractiveFabricObject, [
```
ShaMan123, line 12: This is not bad 
Though we can chain them with extends... like geometry and interactivity
Or we can make all use this pattern.

src/shapes/fabricObject.class.ts
```diff
@@ -0,0 +1,28 @@
...
+  FabricObjectObjectStraighteningMixin,
+  FabricObjectSVGExportMixin,
+  FabricObjectObjectAnimationMixin,
+]);
+
+// export interface InteractiveFabricObject
+//   extends FabricObjectAncestryMixin,
+//     FabricObjectObjectStackingMixin,
+//     FabricObjectObjectStraighteningMixin,
+//     FabricObjectSVGExportMixin,FabricObjectObjectAnimationMixin {}
```
ShaMan123, line 24: The link of the TS mixins states to do this but it won't let me since (I think) it needs to be in the same file declaring the class

src/util/applyMixins.ts
```diff
@@ -0,0 +1,22 @@
...
+/***
+ * https://www.typescriptlang.org/docs/handbook/mixins.html#alternative-pattern
+ */
+export function applyMixins<T extends TClass, S extends TClass>(
+  derivedCtor: T,
+  constructors: S[]
+) {
+  constructors.forEach((baseCtor) => {
+    Object.getOwnPropertyNames(baseCtor.prototype).forEach((name) => {
+      name !== 'constructor' &&
```
ShaMan123, line 12: changed!
No constructor overrides!

ShaMan123 (commented):
ready

src/shapes/fabricObject.class.ts
ShaMan123: fixed

ShaMan123 (commented):
Conflicts have been addressed

> Does ancestry have any usefulness outside the interactivity with activeSelection and group in general?

I think so. It is useful in some cases. And since Group is now a powerful object I guess more use cases will merge and needs to probe the object tree.

---

IMO the stacking/ancestry mixins should be part of the prototype chain of FabricObject, perhaps the rest as well. SVG can be left as a dynamic mixin.
But the decision will not be breaking so we can move forward and consolidate it afterwards.
Thoughts?

src/mixins/object_animation.mixin.ts
```diff
@@ -0,0 +1,184 @@
...
+   * @param {Object} [options] Options object
+   */
+  _animate<T>(key: string, to: T, options: TAnimationOptions = {}) {
+    const path = key.split('.');
+    const propIsColor = this.colorProperties.includes(path[path.length - 1]);
+    const currentValue = path.reduce((deep: any, key) => deep[key], this);
+
+    to = to.toString();
+    if (!propIsColor) {
+      if (~to.indexOf('=')) {
```
ShaMan123, line 88: What is this?
I want to simplify it but I have no idea what it does

src/shapes/fabricObject.class.ts
```diff
@@ -1,13 +1,20 @@
...
+import { ObjectEvents } from '../EventTypeDefs';
+import { FabricObjectSVGExportMixin } from '../mixins/object.svg_export';
 import { InteractiveFabricObject } from '../mixins/object_interactivity.mixin';
+import { applyMixins } from '../util/applyMixins';
 
 // TODO somehow we have to make a tree-shakeable import
 
-export { InteractiveFabricObject as FabricObject };
+export class FabricObject extends applyMixins(InteractiveFabricObject, [
+  FabricObjectSVGExportMixin,
```
ShaMan123, line 10: Should we put the svg mixin in the proto chain as well?

src/mixins/object_animation.mixin.ts
```diff
@@ -0,0 +1,184 @@
...
+    );
+  }
+
+  /**
+   * @private
+   * @param {String} key Property to animate
+   * @param {String} to Value to animate to
+   * @param {Object} [options] Options object
+   */
+  _animate<T>(key: string, to: T, options: TAnimationOptions = {}) {
```
ShaMan123, line 81: This method is so ugly
Once #8297 is merged I will rework this mixin to use an array animation perhaps... with color somehow

index.js
```diff
@@ -15,14 +15,13 @@ import './src/canvas.class'; // optional interaction
...
-import './src/mixins/object_stacking.mixin';
-import './src/mixins/object.svg_export';
-import './src/mixins/stateful.mixin';
-import './src/mixins/animation.mixin'; // optional animation
+import './src/mixins/canvas_serialization.mixin'; // optional serialization
+import './src/mixins/canvas_gestures.mixin'; // optional gestures
+import './src/mixins/canvas_animation.mixin'; // optional animation
+import './src/mixins/canvas_straightening.mixin'; // optional animation
+import './src/shapes/fabricObject.class';
+import './src/mixins/object_stacking.mixin'; // removed in #8461
```
ShaMan123, line 23: transient #8461

ShaMan123 (commented):
combined animation mixin with straightening since both animate object.
Must say that the animation mixin is a vile piece of work. 
`src/mixins/object_stacking.mixin.ts` will be removed in #8461 , kept only for tests

src/mixins/object_animation.mixin.ts
ShaMan123: Can/should be done in a follow up

asturur: this is the ancestor of `to.includes('=')` the tilde is a binary not operator, so bascially since -1
https://wsvincent.com/javascript-tilde/

how old was this line?

asturur (approved):

asturur:
@ShaMan123 i m merging because nothing struck me here.
I know i have no clear how we are going to deal with mixins in general, probably we should take the time to brainstorm some idea.
Conditionally mix permutations of mixins doesn't seem an option here.

src/mixins/object_animation.mixin.ts
ShaMan123: https://github.com/fabricjs/fabric.js/commit/f270ca0259a67c1b41af4b40b3925d3931ed3bac#diff-58618d9e7746cd656ccf18e71d215f47c64423197bf3f0c262b941035811c8e5

More than 10 years

PR #8415, opened 2022-11-01
Create functions to generate standard control set for whoever wants new control set and new controls per object

asturur:
For Immutability super fans this function can help to have quickly a new control set and new controls when creating new instances

## Gist

closes #7914


src/controls/default_controls.ts
```diff
@@ -14,7 +14,8 @@ import {
 } from './actions';
 import { Control } from './control.class';
 
-export const defaultControls = {
+// use this function if you want to generate new controls for every instance
+export const createObjectDefaultControls = () => ({
```
ShaMan123, line 18: naming createDefaultControls?

ShaMan123 (commented):
Yes!
A great step forward for supplying some kind of alternative.
We need to expose these exports, the index file needs to be modified to export this file

asturur:
> Yes! A great step forward for supplying some kind of alternative. We need to expose these exports, the index file needs to be modified to export this file

As soon as we are done with TS + imports, we can make a bundle and the index.js file goes away.
At that point we find a way to import / export everything we need to expose

PR #8416, opened 2022-11-01
ci(): remove unwanted file report stats

ShaMan123:
## Description

We don't want the useless file reporting

That plugin was silly even if we wanted/needed a per file report since it is inaccurate (not checking size of the actual bundle of build, just guessing)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,8 +36,7 @@ jobs:
           script: |
             const fs = require('fs');
             return JSON.stringify({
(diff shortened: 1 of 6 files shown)


ShaMan123 (commented):
OK
Removed unwanted

We can attach the table to the lcov report comment...

asturur:
Yes we want to know how big is fabricJS or more builds of fabricJS that can be proposed as standard packages ( not interactive for example ).
We jsut want to keep an eye on the global bundle mainly.

PR #8417, opened 2022-11-01
chore(TS): Polyline/Polygon

ShaMan123:
## Changes

- exposed protected method `Polyline#isOpen(): true` as in `isOpenPath` that Polygon overrides to `false`, removing duplicate logic
- renamed: fromElementGenerator => polyFromElement
- removed `Point#type`, an annoying thing that makes tests fail when converting IPoint to Point, useless and dangerous if spreading a point with an object since `{ ...point, ...objectWithType }` is not deepEqual `{ ...objectWithType, ...point }`

### DIFF
--- a/src/shapes/polyline.class.ts
+++ b/src/shapes/polyline.class.ts
@@ -38,6 +38,8 @@ export class Polyline extends FabricObject {
    * @todo check if you really need to recalculate for all cases
    */
   strokeBBoxAffectingProperties: (keyof this)[];
+  fromSVG: boolean;
+  pathOffset: Point;
 
   /**
    * Constructor
@@ -58,7 +60,7 @@ export class Polyline extends FabricObject {
    *   top: 100
    * });
    */
-  constructor(points: Point[] = [], options: object = {}): Polyline {
+  constructor(points: Point[] = [], options: any = {}) {
     super({ points, ...options });
     this.initialized = true;
     const bboxTL = this.setDimensions();
@@ -116,8 +118,7 @@ export class Polyline extends FabricObject {
       top: bbox.top - legacyCorrection,
       pathOffset: new Point(pathOffsetX, pathOffsetY),
       strokeOffset: new Point(bboxNoStroke.left, bboxNoStroke.top).subtract(
-        bbox.left,
-        bbox.top
+        new Point(bbox.left, bbox.top)
       ),
     };
   }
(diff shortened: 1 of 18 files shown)


ShaMan123 (commented):
READY

src/shapes/polyline.class.ts
```diff
@@ -1,353 +1,354 @@
...
-       * Calculate the exact bounding box taking in account strokeWidth on acute angles
-       * this will be turned to true by default on fabric 6.0
-       * maybe will be left in as an optimization since calculations may be slow
-       * @deprecated
-       * @type Boolean
-       * @default false
-       * @todo set default to true and remove flag and related logic
-       */
-      exactBoundingBox: false,
+  private initialized: true | undefined;
```
ShaMan123, line 65: I typed it this way since we setting it to false by default is pointless because the value will be assigned after super is called and by then set has been called as well. The whole point of the flag is to prevent that call.
So it is undefined while calling super and true once super is ready.

src/shapes/polyline.class.ts
```diff
@@ -1,353 +1,354 @@
...
       extraParam: 'points',
     });
-  };
-})(typeof exports !== 'undefined' ? exports : window);
+  }
+}
+
+export const polylineDefaultValues: Partial<TClassProperties<Polyline>> = {
+  type: 'polyline',
+  points: [],
```
ShaMan123, line 336: not null anymore, maybe should be

ShaMan123: or no default value actually

ShaMan123: f6eb764

src/shapes/polyline.class.ts
```diff
@@ -1,353 +1,349 @@
...
   /**
-   * List of attribute names to account for when parsing SVG element (used by {@link fabric.Polyline.fromElement})
+   * List of attribute names to account for when parsing SVG element (used by {@link Polyline.fromElement})
    * @static
-   * @memberOf fabric.Polyline
+   * @memberOf Polyline
    * @see: http://www.w3.org/TR/SVG/shapes.html#PolylineElement
    */
-  fabric.Polyline.ATTRIBUTE_NAMES = fabric.SHARED_ATTRIBUTES.concat();
+  static ATTRIBUTE_NAMES = [...SHARED_ATTRIBUTES];
```
ShaMan123, line 296:
```suggestion

```

src/shapes/polyline.class.ts
asturur: yeah we have to rework this. there is something similar also in text and just can't be. There must be a way to make the code straight.

asturur (approved):

PR #8420, opened 2022-11-03
chore(TS): Texts 1st patch

ShaMan123:
## Changes

Moved selectionStart/End logic outside of TextStyleMixin to IText where it belongs

## Gist

I migrated IText, Textbox as well since it made sense
Next PR I will handle IText mixins


test/unit/text.js
```diff
@@ -1,6 +1,13 @@
 (function() {
 
-  QUnit.module('fabric.Text');
+  QUnit.module('fabric.Text', {
+    before() {
+      fabric.config.configure({ NUM_FRACTION_DIGITS: 2 });
```
ShaMan123, line 5: I have no idea why this is needed
w/o tests fail
I will inspect

src/shapes/text.class.ts
```diff
@@ -1,1821 +1,1802 @@
...
+  missingNewlineOffset() {
+    return 1;
+  }
+
+  /**
+   * Returns 2d representation (lineIndex and charIndex) of cursor
+   * @param {Number} selectionStart
+   * @param {Boolean} [skipWrapping] consider the location for unwrapped lines. useful to manage styles.
+   */
+  get2DCursorLocation(selectionStart: number, skipWrapping?: boolean) {
```
ShaMan123, line 495: migrated from style mixin

ShaMan123 (commented):
This is it
I would like to spilt up Text **in a follow up** into:
- renderer
- measurer

asturur:
argh 3000 lines. let's start.

test/unit/text.js
asturur: all the tests are usually done with NUM_FRACTION_DIGITS that was the old defaults to avoid large exports and approximation error between js implementations

ShaMan123: I pointed this out since it was I change I made to tests for them to pass, before they were passing without `NUM_FRACTION_DIGITS` assignment, not sure if it was pollution or what

src/shapes/text.class.ts
```diff
@@ -1,1821 +1,1793 @@
...
+
+  protected __skipDimension: boolean;
+  protected textLines: string[];
+  protected _textLines: string[][];
+  protected _unwrappedTextLines: string[][];
+  protected _text: string[];
+  protected cursorWidth: number;
+  protected __lineHeights: number[];
+  protected __lineWidths: number[];
+  protected _forceClearCache: boolean;
```
asturur, line 348: some of those are useful to read in certain cases. I'm not sure we should flag them as protected as part of the migration

ShaMan123: I can remove. I have no opinion on this

asturur: i m about to commit some comments in the source code and a forgot var/const let's not overlap. i m about to merge this

asturur (approved):

PR #8421, opened 2022-11-03
chore(TS): IText behavior mixins

ShaMan123:
## Description

Continues #8420 
This PR migrates IText mixins into classes that are chained to create the final IText class.
I have treated itext_behavior as the base class, holding more general methods, while itext_key_behavior and itext_click_behavior  hold specific methods. itext_key_behavior is also in charge of the hidden textarea since it is part of the key behavior scope.
I moved some methods around to accommodate my approach but this is far from perfect.
I would rename the classes as well.
I have used abstract methods to make it easier to migrate the mixins. A follow up should do a heavier refactor, moving more methods and making the mixins more defined in scope.

## Gist

Proto chain as follows:
1. itext_behavior extends Text
2. itext_key_behavior 
3. itext_click_behavior 
4. ITEXT


ShaMan123:
Build stats is wrong

> **Build Stats**
> 
> file / KB (diff)	bundled	reduced*	minified	gzipped
> fabric.min.js	1078.104 (**+63.882**)		331.222 (**+30.636**)	93.176 (**+7.604**)
> src/mixins/itext_behavior.mixin.ts	43.055 (**+43.055**)	44.184 (**+44.184**)		
> src/mixins/itext_click_behavior.mixin.ts	9.402 (**+9.402**)	9.602 (**+9.602**)		
> src/mixins/itext_key_behavior.mixin.ts	19.997 (**+19.997**)	19.050 (**+19.050**)		
> src/mixins/object_interactivity.mixin.ts	20.364 (**+0.001**)	18.729 (**0.000**)		
> src/mixins/observable.mixin.ts	4.599 (**+0.001**)	3.374 (**0.000**)		
> src/mixins/text_style.mixin.ts	9.339 (**+9.339**)	9.283 (**+9.283**)		
> src/shapes/itext.class.ts	19.490 (**-0.741**)	18.364 (**-2.820**)		
> src/shapes/text.class.ts	54.621 (**-5.601**)	50.654 (**-12.394**)		
> src/shapes/textbox.class.ts	14.160 (**-1.123**)	14.104 (**-3.030**)		
> *inaccurate, see [link](https://github.com/doesdev/rollup-plugin-analyzer#why-is-the-reported-size-not-the-same-as-the-file-on-disk)

asturur:
Tried my best for conflicts, let's see.
Decription sucks.

asturur:
28kb what? is that broken again?

ShaMan123:
Updated description

ShaMan123:
b94dff4 fixed the wrong stats

ShaMan123:
2kb less 🥳

asturur:
So we know that text mixins are messy. I don't think right now renaming classes/mixins/files is a good idea.
Important is that we type what is typeable.
There is code that can probalby improved in the text interaction area, let's collect ideas in a doc with examples and let's move one. is on par or better than before and that is what it count since this is a tech migration

ShaMan123 (commented):
Agreed

asturur (approved):

asturur:
@ShaMan123 i ll start Static canvas today, and i look forward merging Image and observable types when we can.
I also know you probably want to look at group?

ShaMan123:
Yep. Started group. Will try to fix the failing test in Image.
What about the animation pr?

Issue #7501, opened 2021-11-18
Skewing bug when destroying group or removing elements from group


aastefian:
## Version
4.6.0 - but happens since 1.6.0

## Test Case
http://jsfiddle.net/aastefian/6dv5b1pq/80/

## Information about environment
Browser: Version 95.0.4638.69 (Official Build) (64-bit)

## Steps to reproduce
Create a shape element (maybe all elements have the bug but didnt test yet), add skewY to it (lets say 20), create a group with that element and then call `destroy()` on the group or `removeWIthUpdate()` on every element in the group.

## Expected Behavior
skewY should stay 20. Angle and skewX should not be modified. We depent on this because we have a web editor in which all attributes are very visible and it is very bad UX to see all values shifting around when ungrouping.

## Actual Behavior
The element's state gets modified and skewY is set to 0 and skewX and angle are given values which visually makes the shape look same(almost identical) but the attributes are set incorrectly (skewY is reset to 0, skewX and angle get a value).

This bug seems to have been introduced when going from version 1.5.0 to 1.6.0 with the introduction of `qrDecompose ` function which seems to decompose the transform matrix into different transformations but always extracts skewY as 0. this might be a down side using QR decomposition which only extracts angle, scaleX/Y skewX but **not** skewY. I am thinking maybe a combination of LU decomposition and QR decomposition might give us all transforms (but this is just a theory, not sure if its mathematically possible even). Also, not sure if the 2x3 matrix actually stores all transformations including skewY or not.

I already worked on a small fix but I didn't figured it out quite yet, i will let you know if i can propose a PR.

Check: https://github.com/fabricjs/fabric.js/compare/v1.5.0...v1.6.0# on line 523 of misc.js, this is where the function was added.

asturur:
Hi @aastefian this is by design, i'm sorry.
If you manage to fix it, great.
When objects are grouped, all the properties of the object and parents get saved in a single transform matrix of 6 numbers.
From those 6 numbers we need to pull out:

- top
- left
- angle
- skewX
- skewY
- scaleX
- scaleY

7 informations.

7 variables, less equations, we need to fix one of those to a constant.
The only working math formula i found requires skewY to be set at 0.

aastefian:
I see. But when you apply skewing on a group do you apply on the matrix both skewX and skewY or not?

I want to understand if the information is even there in the first place and it's only a matter of exctracting it with the right math.

asturur:
[...]

Issue #4392, opened 2017-10-19
If grouping object with flipX, that object obtains rotation


LA-arealmedia:
## Version
1.7.15

## Test Case
https://jsfiddle.net/LA2937/fga9j48h/

## Steps to reproduce
1. Make a group selection of the triangle and rectangle;
2. Deselect the group;
3. Select the triangle and you'll see that now have rotation (angle = 180).

## Expected Behavior
The triangle should not have rotation after step 1 and 2.

## Actual Behavior
The triangle is obtaining rotation (angle = 180).

asturur:
The ungrouping logic apply a decomposition of the current situation.
The math formula works like that, i struggled to write it and i do not think i can write one better.

For some reason a flipX is equivalent to a flipY + rotation.
Can probably safeguarded but the risk of breaking stuff is much.

LA-arealmedia:
I verified that some math formula transformation is made when grouping the objects - I understand that. Although, I believe that must be made a reversion of the changes to flip and rotation, after an ungrouping action. What you think?

asturur:
that works for very simple cases but not the complicate one. that is the
problem.

2017-10-19 13:58 GMT+02:00 LA-arealmedia <notifications@github.com>:

LA-arealmedia:
But this seems to be a very easy thing to do. The matrix is making the needed angle transformation to turn off the flipX (when creating a group). The only thing that I think is missing is to restore the previous configurations of the object (angle and flip).

melvynhills:
I could fix this behaviour for my needs with these changes (not thoroughly tested yet).

As soon as flip values are changed, subtracting 180° from the angle and set back the original flip values seems to do the trick.

```javascript
fabric.Group.prototype.realizeTransform = function(object) {
  var matrix = object.calcTransformMatrix(),
      options = fabric.util.qrDecompose(matrix),
      center = new fabric.Point(options.translateX, options.translateY),
  // custom code start ------------------------------------------------------------------
      originalFlipX = object.flipX,
      originalFlipY = object.flipY;
  // custom code end ------------------------------------------------------------------

  object.flipX = false;
  object.flipY = false;
  object.set('scaleX', options.scaleX);
  object.set('scaleY', options.scaleY);
  object.skewX = options.skewX;
  object.skewY = options.skewY;

  // custom code start ------------------------------------------------------------------
  if (originalFlipX !== object.flipX || originalFlipY !== object.flipY) {
    object.flipX = originalFlipX;
[...]

Issue #3795, opened 2017-03-22
Flip Object


mariusturcu1:
## Version
1.7.6

## Test Case
https://ufile.io/206db

If you flip a object in x coordonate, after that select 2 or more objects, deselect, seems like flipX is 0 and flipY becomes 1... but the render is still good, only the controls are not good
Same problem using images or custom shapes

asturur:
that is normal. When ungrouping fabric takes one of the possible math solution to your problem.

Probably a flipx can be represented with a flipy + a 180 rotation.

Nothing we are gonna change since is an edge case and transformations are equivalent.

melvynhills:
I could fix this behaviour for my needs with these changes (not thoroughly tested yet).

As soon as flip values are changed, subtracting 180° from the angle and set back the original flip values seems to do the trick.

```javascript
fabric.Group.prototype.realizeTransform = function(object) {
  var matrix = object.calcTransformMatrix(),
      options = fabric.util.qrDecompose(matrix),
      center = new fabric.Point(options.translateX, options.translateY),
  // custom code start ------------------------------------------------------------------
      originalFlipX = object.flipX,
      originalFlipY = object.flipY;
  // custom code end ------------------------------------------------------------------

  object.flipX = false;
  object.flipY = false;
  object.set('scaleX', options.scaleX);
  object.set('scaleY', options.scaleY);
  object.skewX = options.skewX;
  object.skewY = options.skewY;

  // custom code start ------------------------------------------------------------------
  if (originalFlipX !== object.flipX || originalFlipY !== object.flipY) {
    object.flipX = originalFlipX;
    object.flipY = originalFlipY;
    options.angle -= 180;
  }
  // custom code end ------------------------------------------------------------------

  object.angle = options.angle;
  object.setPositionByOrigin(center, 'center', 'center');

  return object;
};

```

PR #8425, opened 2022-11-05
Decode transform matrix

### LINKED ISSUE #5079: qrDecompose.angle does not match obj.angle with flipX applied
## Version
2.3.3

## Test Case
Go to http://fabricjs.com/matrix-transformation and run this script in console:

```js
var obj = __canvas.item(0);
function testObject() {
  var transformValues = fabric.util.qrDecompose(obj.calcTransformMatrix());
  var match = obj.angle === transformValues.angle;
  console.log(match ? 'yes' : 'NO', obj.angle, transformValues);
}
testObject(); // obj.angle and matrix.angle match
obj.flipX = true;
__canvas.requestRenderAll();
setTimeout(testObject, 100); // obj.angle correctly zero. matrix.angle is now 180
```

## Steps to reproduce

Run the script above or:

1. add an object to canvas.  obj.angle will be zero.  calcTransformMatrix + qrDecompose angle will be zero
1. obj.flipX = true; requestRenderAll()
1. obj.angle remains zero.  qrDecompose now returns 180

## Expected Behavior

`fabric.util.qrDecompose(obj.calcTransformMatrix()).angle` and `obj.angle` should both equal zero, or qrDecompose shouldn't return an angle property if it can't return an accurate value.

## Actual Behavior

After the flip is applied, the result from qrDecompose is 180. 

If qrDecompose can't be used this way, what is the recommended way to get the ungrouped object properties from a selection's object:modified event?

### LINKED ISSUE #7620: Inconsistent (de)serialization state with use `calcTransformMatrix` and `qrDecompose`
Hi. I need to clone objects and i should serialize the objects for my purposes.
When i try to copy the object which is flipped, i got unexpected transformation (see below).
How we can fix it?

## Version
4.6.0

## Test Case
https://jsfiddle.net/k3wof27e/

Nodejs or browser? Browser
Which browsers? All

## Steps to reproduce
- Open https://jsfiddle.net/k3wof27e/
- Flip rect manually
- Press button to clone object

## Expected Behavior
Cloned object is have same rotation state and position as original object

## Actual Behavior
Cloned object is not flipped. It's looks like it's flipped, because transform controls are flipped, but also object rotated to 180deg.

luizzappa:
## Motivation

As commented on https://github.com/fabricjs/fabric.js/pull/8337, regardless of whether we use qr decomposition, or any other strategy, we will be in a situation with more variables than equations available, which makes it impossible to extract all options from transform matrix unless some assumptions are made. In the case of `qrDecompose`, it is considered `skewY` = 0, and as a result the other options are found. This generates decompositions that do not correspond to reality.

However, knowing the rotation angle, we are able to find the other options correctly. This PR is based on that.

closes #7620, closes #5079

## Description

Created the `decodeTransformMatrix`  function that receives the `transform matrix` and the `angle`, and returns the options correctly. `Angle` is an optional parameter, if not given, `qrDecompose` is used (it will be the "best guess" for the lack of information provided).

**The logic behind**:

An object's transformation matrices are applied in that order until we reach the final transform matrix:

image

And:
- a = scaleX
- b = skewY
- c = skewX
- d = scaleY
- e = offsetX
- f = offsetY 

The translation does not interfere with the other transformations, because of this, we can disregard it and work with 2x2 matrices:

image

Since we have the rotation, we can remove this effect from the transformation matrix by multiplying by the inverse of the rotation matrix and obtaining an identity matrix:
image

With this, we get the following final transformation matrix (without translation and rotation):
image

So, if we have a transformation matrix **A** (without translations and rotation), we can find the values ​​of a (scaleX), b (skewY), c (skewX) and d (scaleY).
image

Solving this system, we find:

image

## Changes

In this PR it's just an inclusion in `fabric.util`. But this will enable fix issues due to the imprecision of `qrDecompose`. 

For example, an object with `skewY`, when entering and leaving a group, has `rotation` instead of `skewY` because of `qrDecompose`.
bug group

It also causes problems in object rendering... The three objects below are the same and have skewY. When we put 2 of them in a group, see that the geometric shape is changed:
bug group2

Other people have run into these problems: https://github.com/fabricjs/fabric.js/issues/4392, https://github.com/fabricjs/fabric.js/issues/7501, https://github.com/fabricjs/fabric.js/issues/3795, https://github.com/fabricjs/fabric.js/issues/3526, https://github.com/fabricjs/fabric.js/issues/3482, https://github.com/fabricjs/fabric.js/issues/3052

## In Action

See the unit test


test/unit/util.js
```diff
@@ -985,6 +1011,55 @@
...
+    for (let angle = 0; angle <= 360; angle += 30) {
+      [[1, 1], [1, 4], [5, 1], [3, 8]].forEach(([scaleX, scaleY]) => {
+        for (let skewX = -80; skewX < 90; skewX += 20) {
+          for (let skewY = -80; skewY < 90; skewY += 20) {
+            [
+              [false, false], 
+              [true, false], 
+              [false, true], 
+              [true, true]
+            ].forEach(([flipX, flipY]) => {
```
luizzappa, line 1026: This nested loop works on node but gives weird error in firefox and chrome...

Firefox:
```
Stdout: 
             [GFX1-]: RenderCompositorSWGL failed mapping default framebuffer, no dt
```

It only works when I run fewer test cases. It seems to be some memory overflow.

I didn't find much about this error, just this here: https://bugzilla.mozilla.org/show_bug.cgi?id=1716303. It seems to me to be a log spamming bug. Maybe this causes memory overflow(?)

luizzappa: I fixed it here https://github.com/fabricjs/fabric.js/pull/8425/commits/6353f9d27e0b6425a29664a6a575758f7ab01d82 by separating it into several test cases. It was kind of ugly, but it was the way I managed to make it run in firefox and chrome without error.

ShaMan123:
I gave a look and tried out a few things.
My assumptions:
1. angle is linear, meaning to get the total rotation of a nested object I add the angles of all the ancestors to the object. I use this to get the angle when using your decomposition.
2. When decomposing while SVG importing using qrDecompose is fine since it doesn't matter.

Outcome looks promising 643b1ee38, feel free to cherry-pick or merge [`decodeTMat-impl`](https://github.com/fabricjs/fabric.js/compare/decodeTMat-impl).

PlaneChange is something I want to discuss. Arguments are `(object: FabricObject, from: T2DMatrix, to: T2DMatrix)`. What do we do in this case? I will make it clearer. The signature of PlaneChange shouldn't really be `(object: FabricObject, from: T2DMatrix, to: T2DMatrix)`. I did that because of a bug in clipPath addressed #8298 (It was unaware of its ancestors). It should be `(object: FabricObject, from: FabricObject, to: FabricObject)`. Then we can get the angle of each object and calculate the rotation.
How does that sound?
I think `addTransformToObject` and `removeTransformFfromObject` can be addressed in a same way, refactored to use plane change.



---

<details><summary>controls12</summary>

Changes made to the green rect, looks correct

```json
{
  "version": "4.6.0",
  "objects": [
    {
      "type": "rect",
      "version": "4.6.0",
      "left": 38,
      "top": 201,
      "width": 150,
      "height": 150,
      "fill": "red",
      "skewX": 0.15,
      "skewY": 36
    },
    {
      "type": "rect",
      "version": "4.6.0",
      "left": 20,
      "top": 2,
      "width": 150,
      "height": 150,
      "fill": "#020aed",
      "scaleX": 1.24,
      "scaleY": 0.81,
      "angle": 35.95,
      "skewX": 25.46
    },
    {
      "type": "group",
      "version": "4.6.0",
      "left": 60.65,
      "top": 28,
      "width": 320.4,
      "height": 335.5,
      "objects": [
        {
          "type": "rect",
          "version": "4.6.0",
          "left": -29.85,
          "top": -167.75,
          "width": 150,
          "height": 150,
          "fill": "green",
          "angle": 30,
          "skewX": 14.71,
          "skewY": 36
        },
        {
          "type": "rect",
          "version": "4.6.0",
          "left": -29.85,
          "top": -167.75,
          "width": 150,
          "height": 150,
          "fill": "yellow",
          "angle": 45,
          "skewX": 14.71
        }
      ]
    }
  ]
}


```

</details>


<details><summary>controls13</summary>

Changes made to:
- green rect - looks correct
- yellow rect - looks correct
- purple rect - **not certain**, controls are position on the smaller axis of the rect

```json
{
  "version": "5.2.0",
  "objects": [
    {
      "type": "rect",
      "version": "5.2.0",
      "left": 38,
      "top": 201,
      "width": 150,
      "height": 150,
      "fill": "red",
      "skewX": 0.15,
      "skewY": 36
    },
    {
      "type": "rect",
      "version": "5.2.0",
      "left": 20,
      "top": 2,
      "width": 150,
      "height": 150,
      "fill": "#020aed",
      "scaleX": 1.24,
      "scaleY": 0.81,
      "angle": 35.95,
      "skewX": 25.46
    },
    {
      "type": "group",
      "version": "5.2.0",
      "left": 152.65,
      "top": 21,
      "width": 320.4,
      "height": 335.5,
      "scaleX": 0.75,
      "skewY": 24.57,
      "objects": [
        {
          "type": "rect",
          "version": "5.2.0",
          "left": -29.85,
          "top": -167.75,
          "width": 150,
          "height": 150,
          "fill": "green",
          "angle": 30,
          "skewX": 14.71,
          "skewY": 36,
          "flipX": true,
          "flipY": true
        },
        {
          "type": "rect",
          "version": "5.2.0",
          "left": -29.85,
          "top": -167.75,
          "width": 150,
          "height": 150,
          "fill": "yellow",
          "angle": 45,
          "skewX": 14.71
        }
      ]
    },
    {
      "type": "group",
      "version": "5.2.0",
      "left": 329.65,
      "top": 65,
      "width": 320.4,
      "height": 335.5,
      "scaleX": 1.29,
      "scaleY": 1.29,
      "flipX": true,
      "objects": [
        {
          "type": "rect",
          "version": "5.2.0",
          "left": -29.85,
          "top": -167.75,
          "width": 150,
          "height": 150,
          "fill": "purple",
          "angle": 30,
          "skewX": 14.71,
          "skewY": 36
        },
        {
          "type": "rect",
          "version": "5.2.0",
          "left": -29.85,
          "top": -167.75,
          "width": 150,
          "height": 150,
          "fill": "pink",
          "angle": 45,
          "skewX": 14.71
        }
      ]
    }
  ]
}

```

</details>

test/unit/util.js
ShaMan123: I am experiencing this on my machine running in firefox
Can you post the command you are running?

ShaMan123: I see the logs on CI....
I have no idea why and what

luizzappa: npm run test -- -c chrome

luizzappa: I did it in a way that looks a little better. I'll commit

luizzappa: https://github.com/fabricjs/fabric.js/pull/8425/commits/9d8921a9113d8700d1776036743d58f03422cc89 
I adapted from your commit. The problem seems to be test cases that are too large.. Locally ran without problems

test/unit/util.js
```diff
@@ -1058,11 +1058,11 @@
     });
   };
 
-  QUnit.test('decodeTransformMatrix with angle', function (assert) {
-    for (let angle = 0; angle <= 360; angle += 30) {
+  for (let angle = 0; angle <= 360; angle += 30) {
+    QUnit.test(`decodeTransformMatrix with ${angle}`, function (assert) {
```
ShaMan123, line 1062: This is SO weird!

ShaMan123: How did you even understand that this is what makes it work????

ShaMan123: Did you try renaming assert.close? Might it be a name collision?

luizzappa:
> Outcome looks promising [643b1ee](https://github.com/fabricjs/fabric.js/commit/643b1ee381317526a4be65325a93a394081bb802), feel free to merge [`decodeTMat-impl`](https://github.com/fabricjs/fabric.js/compare/decodeTMat-impl).

Looks good!

The problem I had commented on in groups:

BEFORE
> The three objects below are the same and have skewY. When we put 2 of them in a group, see that the geometric shape is changed
> group


AFTER 
group parcial fix

ShaMan123:
Well done.
Can you add a visual test for this exactly?

luizzappa:
> 1. angle is linear, meaning to get the total rotation of a nested object I add the angles of all the ancestors to the object. I use this to get the angle when using your decomposition.

Hmm I'm not sure about that, I think it's not enough. Although the total rotation is the sum of the angles, when decomposing the transformation matrix we must proceed in stages, because the inverse of the rotation matrix has an order to apply, I'm going to set up a unit test to verify this.

> 2\. using qrDecompose while importing for SVG  is fine since it doesn't matter.

I don't have much knowledge to comment here.. Does importing a SVG group with an element with skewY not a problem?

test/unit/util.js
luizzappa: > Did you try renaming assert.close? Might it be a name collision?

I tried too, it didn't work

> How did you even understand that this is what makes it work????

It seems to me that too many asserts in a single test cause this problem in firefox/chrome. A simple test like this gives the same problem.

```
  QUnit.test.only('many assertions in a single test', function (assert) {
    for (let i = 0; i <= 28000; i += 1) {
      assert.equal(i, i, 'msg')
    }
  });
```

ShaMan123:
I am guessing the blame for the error is QUnit. 

> I don't have much knowledge to comment here.. Does importing a SVG group with an element with skewY not a problem?

@asturur will probably have a solid opinion about SVG. 
My approach: it worked until now, it has no bbox rendering so it is fine. However your question is valid, it is best to add a visual test (import svg file)

Tell me if you need any help finalizing this PR. It is good!
Any objections merging the branch I forked here?

ShaMan123:
> Hmm I'm not sure about that, I think it's not enough. Although the total rotation is the sum of the angles, when decomposing the transformation matrix we must proceed in stages, because the inverse of the rotation matrix has an order to apply, I'm going to set up a unit test to verify this.

Could you provide more context?

ShaMan123:
This is how I plan to refactor `sendObjectToPlane`

```ts

export const sendObjectToPlane = (
  object: FabricObject,
  to?: FabricObject
): TMat2D => {
  const from = object.group;
  const t = calcPlaneChangeMatrix(
    from?.calcTransformMatrix(),
    to?.calcTransformMatrix()
  );
  const angle =
    object.angle - (from?.getTotalAngle() || 0) + (to?.getTotalAngle() || 0);
  applyTransformToObject(
    object,
    multiplyTransformMatrices(t, object.calcOwnMatrix()),
    angle
  );
  return { transform: t, angle };
};


```

ShaMan123:
Giving a bit more thought to plane change...
After the fix the clip path is merged I am not entirely sure `sendObjectToPlane` will be needed. If we expose a method that is in charge of setting clip path then it can address the plane change internally. So I think it could be removed.

luizzappa:
> > > My assumptions:
> > > 1. angle is linear, meaning to get the total rotation of a nested object I add the angles of all the ancestors to the object. I use this to get the angle when using your decomposition.

> > Hmm I'm not sure about that, I think it's not enough. Although the total rotation is the sum of the angles, when decomposing the transformation matrix we must proceed in stages, because the inverse of the rotation matrix has an order to apply, I'm going to set up a unit test to verify this.
> 
> Could you provide more context?

Take a look here: https://github.com/fabricjs/fabric.js/pull/8425/commits/5c027c7f2ce464632641a3be737b49346584f824

https://github.com/fabricjs/fabric.js/blob/5c027c7f2ce464632641a3be737b49346584f824/test/unit/util.js#L1067

https://github.com/fabricjs/fabric.js/blob/5c027c7f2ce464632641a3be737b49346584f824/test/unit/util.js#L1134

These two tests create a nested group. 
The `scale` values ​​should be the multiplication of the scale along the chain:

https://github.com/fabricjs/fabric.js/blob/5c027c7f2ce464632641a3be737b49346584f824/test/unit/util.js#L1104-L1106

But adding only the angles (first test) the values ​​are not correct.

We must go through the chain and remove the rotation from the transformation matrix, and the resulting matrix must be decomposed (now with an angle equal to zero, as I removed its effect on the transformation matrix). This is what I do in the second test. PS.: ignore translateX and translateY, they are giving incorrect values. I need to better understand how fabricjs handles them within a group.

luizzappa:
> PlaneChange is something I want to discuss. Arguments are `(object: FabricObject, from: T2DMatrix, to: T2DMatrix)`. What do we do in this case? I will make it clearer. The signature of PlaneChange shouldn't really be `(object: FabricObject, from: T2DMatrix, to: T2DMatrix)`. I did that because of a bug in clipPath addressed in #8298 (It was unaware of its ancestors). It should be `(object: FabricObject, from: FabricObject, to: FabricObject)`. Then we can get the angle of each object and calculate the rotation. How does that sound? I think `addTransformToObject` and `removeTransformFfromObject` can be addressed in a same way, refactored to use plane change.

I merged with https://github.com/fabricjs/fabric.js/compare/decodeTMat-impl. Let me get better acquainted with the implementation and I'll give my opinion. It has some inner workings that I was unaware of, this branch will help me understand better.

ShaMan123:
Great, I have the context I need.
So now I don't understand how decoding is possible.
Since the rotation matrix is applied to all the matrices after it, if we extract the angle and reapply it at the top of the matrix chain it will act over the matrices before it.
Furthermore, I don't understand how scaling can be produced by multiplying the scale values if skewing exists (Adding skewing values to the tests makes everything fail).

ShaMan123:
This makes me about the entire concept, rather than on the solution.
Why does fabric need to decompose a matrix?
In #8380 I tried w/o decomposing and came close.
Besides that?
I have asked this before.
Canvas API doesn't offer such a functionality, and it seems we have found out why.
It does allow acting over the transform matrix.
Apart for controls I am not sure fabric uses decomposed values.

luizzappa:
> Great, I have the context I need.
> So now I don't understand how decoding is possible.

Indeed, decomposition is possible when we are right on the canvas plane. As this test proves:

https://github.com/fabricjs/fabric.js/blob/5c027c7f2ce464632641a3be737b49346584f824/test/unit/util.js#L1062

But when we go through nested group transformations, I had thought it was possible, but it doesn't seem to be... Tracing options only through the transformation matrix doesn't have enough information (actually it makes sense, as there are several different variables in each plane..)

luizzappa:
A stupid question, but it might enlighten me. I still don't understand why we need to decompose a transform matrix if we store the options (skew, scale, etc.) in the objects/groups?

For example I have

```
{
      "type": "group",
      "angle": 10,   
      "scaleX": 2,
      "scaleY": 3, 
      "skewX": 10,
      "skewY": 16   
      "objects": [
        {
          "type": "rect",
          "angle": 30,
          "scaleX": 2,
          "scaleY": 4, 
          "skewX": 20,
          "skewY": 36
        }
}
```

Why couldn't I do:
```
finalScaleX = group.scaleX * rect.scaleX
finalScaleY = group.scaleY * rect.scaleY
finalAngle = group.angle + rect.angle
finalSkewX = tan(group.skewX) + tan(rect.skewX)
finalSkewX = tan(group.skewY) + tan(rect.skewY)
```

ShaMan123:
Good question. I don't know the answer. Searching all qrDecompose usages doesn't show something out of the ordinary.

ShaMan123:
So you suggest altering `multiplyTransformMatrices`?

luizzappa:
> So you suggest altering `multiplyTransformMatrices`?

No no.. I wouldn't mess with the transformation matrix and its operations.

I think I better understand what's going on.. There's **no problem with matrix decomposition when given an angle**, the problem is that I had a misconception when it's inside a group. 

Going back to the example of the group with a rectangle inside:

```
{
      "type": "group",
      "angle": 10,   
      "scaleX": 2,
      "scaleY": 3, 
      "skewX": 10,
      "skewY": 16   
      "objects": [
        {
          "type": "rect",
          "angle": 30,
          "scaleX": 2,
          "scaleY": 4, 
          "skewX": 20,
          "skewY": 36
        }
}
```

What happens is that when an object is inside a group, there is **an order in which the transformations are applied**. First the group receives scale/skew, and the object will apply a new scale/skewy on top. 

So this will never be true: (since the rotation, skew, scale of the group generated a new space, and the rectangle transformations will start from that new space.)

```
finalScaleX = group.scaleX * rect.scaleX
finalScaleY = group.scaleY * rect.scaleY
finalAngle = group.angle + rect.angle
finalSkewX = tan(group.skewX) + tan(rect.skewX)
finalSkewX = tan(group.skewY) + tan(rect.skewY)
```

This here is conceptually incorrect:

https://github.com/fabricjs/fabric.js/blob/5c027c7f2ce464632641a3be737b49346584f824/test/unit/util.js#L1134

luizzappa:
For example:

`Group` with `rotation = 30 degrees and scaleX, scaleY, skewX, skewY = 1` will generate a new space that  that will represent the rotation into `scale` and `skew`.

$$ matrixRotation * matrixScale * matrixSkewX * matrixSkewY = $$

$$ \begin{pmatrix}
cos 30 & - sin 30\\
sin 30 & cos 30
\end{pmatrix}
\begin{pmatrix}
1 & 0\\
0 & 1
\end{pmatrix}
\begin{pmatrix}
1 & 0\\
0 & 1
\end{pmatrix}
\begin{pmatrix}
1 & 0\\
0 & 1
\end{pmatrix} =
$$

$$ \begin{pmatrix}
cos 30 & - sin 30\\
sin 30 & cos 30
\end{pmatrix}
$$ 

So: 
$scaleX = cos 30$
$skewY = sin 30$
$skewX = - sin 30$
$scaleY = cos 30$

So if I have a rectangle inside this group with just `scaleX = 2`, in fact it will also have `skew` and `scaleY` because of the rotation of the group that generated a new space.


$$ rectangle transformation = $$

$$ groupTransformation * rectangleScale = $$

$$ 
\begin{pmatrix}
cos 30 & - sin 30\\
sin 30 & cos 30
\end{pmatrix}
\begin{pmatrix} 
2 & 0\\ 
0 & 1 
\end{pmatrix} =
$$

$$
\begin{pmatrix}
2 cos 30 & - sin 30\\ 
2 sin 30 & cos 30 
\end{pmatrix}
 $$

So for the rectangle:
$scaleX = 2 cos 30$
$skewY = 2 sin 30$
$skewX = - sin 30$
$scaleY = cos 30$

Incorrectly I was thinking that the final `scaleX` of the rectangle would be $finalScaleX = group.scaleX * rectangle.scaleX = 1 * 2 = 2$


Given that, in relation to the assumption:

> 1. angle is linear, meaning to get the total rotation of a nested object I add the angles of all the ancestors to the object. I use this to get the angle when using your decomposition.

What we actually have to do in the case of nested groups is take the transformation matrix of the object and decompose by its angle only. We don't need to go up the chain and add up..  
It makes sense?

asturur:
Hi @luizzappa  first of all thank you for confirming that there are more variables than equations and so qrDecompose is doing what it can do and can't do more than that.

This PR as usally as a lot of comments, i will read it but i have to give it a lower priority.
I won't merge it before 6.0 since is an additive ( improving ) change and the focus now is to bring what we have on TS.

Please don't be put off by that and keep up the good work.

ShaMan123:
closing
hopefully #8767 will provide a better solution

PR #8431, opened 2022-11-09
Observable types

ShaMan123:
## Motivation

Enforce strict typing for events handlers in the Observables.

## Description

~~This PR showcases the pros of TS.~~
~~With it using the observable is so easy and fun.~~
~~Found a few discrepancies in the code and fixed them.~~

This PR add a type definition file that describes in minute detail every event fired by fabricJS.
Events are divided by kind, transformation events, selection events, canvas/group events ( add and remove from the object tree), and interaction events.

Every class then takes a dynamic type of event type that is used to bring down to Observable class the type of event that will fire.

Doing so when using the on function for example, Typescript will be able to suggest what events are available and what options are available in the callback.

There are no functional changes on top of that, apart some event property change that has been found clashing with types and so fixed.
The changed events are:
- after:render, added the ctx property to the callback
- selection:cleared, fixed the last instance that was using target instead of deselected
- before:selection:cleared, changed from target to deselected ( those were all `target` but it seems to me that `deselected` makes more sense )

## Changes

- fix some events that were fired with missing/wrong data 

## In Action

ezgif com-gif-maker (4)

### DIFF
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -1,8 +1,8 @@
-import { Point } from '../point.class';
-import { FabricObject } from '../shapes/fabricObject.class';
-import { Group, LayoutEventData } from '../shapes/group.class';
-import { TEvent, TransformEvent, TPointerEvent } from '../typedefs';
-import { Canvas } from '../__types__';
+import { Point } from './point.class';
+import { FabricObject } from './shapes/fabricObject.class';
+import { Group, LayoutEventData } from './shapes/group.class';
+import { TEvent, TransformEvent, TPointerEvent } from './typedefs';
+import { Canvas } from './__types__';
 
 export type TModificationEvents =
   | 'moving'
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -156,11 +156,13 @@ export type TObjectEvents = ObjectPointerEvents &
     removed: { target: Group | Canvas };
 
     // IText
+    // todo - move to file
     'selection:changed': never;
     changed: never;
     tripleclick: XTransformEvent;
 
     // Group
+    // todo - move to file
     layout: LayoutEventData;
 
     // erasing
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -1,8 +1,70 @@
+import type { Control } from './controls/control.class';
 import type { Point } from './point.class';
 import type { FabricObject } from './shapes/fabricObject.class';
 import type { Group } from './shapes/group.class';
-import type { TEvent, TPointerEvent, TransformEvent } from './typedefs';
+import type { TOriginX, TOriginY, TRadian } from './typedefs';
+import type { saveObjectTransform } from './util/misc/objectTransforms';
 import type { Canvas } from './__types__';
+
+export type ModifierKey = 'altKey' | 'shiftKey' | 'ctrlKey';
+
+export type TPointerEvent = MouseEvent | TouchEvent;
+
+export type TransformAction<T extends Transform = Transform, R = void> = (
+  eventData: TPointerEvent,
+  transform: T,
+  x: number,
+  y: number
+) => R;
+
+export type TransformActionHandler<T extends Transform = Transform> =
+  TransformAction<T, boolean>;
+
+export type ControlCallback<R = void> = (
+  eventData: TPointerEvent,
+  control: Control,
+  fabricObject: FabricObject
+) => R;
+
+export type ControlCursorCallback = ControlCallback<string>;
+/**
+ * relative to target's containing coordinate plane
+ * both agree on every point
+ */
+
+export type Transform = {
+  target: FabricObject;
+  action: string;
+  actionHandler: TransformActionHandler;
+  corner: string;
+  scaleX: number;
+  scaleY: number;
+  skewX: number;
+  skewY: number;
+  offsetX: number;
+  offsetY: number;
+  originX: TOriginX;
+  originY: TOriginY;
+  ex: number;
+  ey: number;
+  lastX: number;
+  lastY: number;
+  theta: TRadian;
+  width: number;
+  height: number;
+  shiftKey: boolean;
+  altKey: boolean;
+  original: ReturnType<typeof saveObjectTransform>;
+};
+
+export type TEvent<E extends Event = TPointerEvent> = {
+  e: E;
+};
+
+export type TransformEvent<E extends Event = TPointerEvent> = TEvent<E> & {
+  transform: Transform;
+  pointer: Point;
+};
 
 export type TModificationEvents =
   | 'moving'
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -5,6 +5,7 @@ import type { Group } from './shapes/group.class';
 import type { TOriginX, TOriginY, TRadian } from './typedefs';
 import type { saveObjectTransform } from './util/misc/objectTransforms';
 import type { Canvas } from './__types__';
+import type { IText } from './shapes/itext.class';
 
 export type ModifierKey = 'altKey' | 'shiftKey' | 'ctrlKey';
 
@@ -213,8 +214,10 @@ export type CanvasEvents = CanvasPointerEvents &
         };
 
     // IText
-    'text:selection:changed': { target: FabricObject };
-    'text:changed': { target: FabricObject };
+    'text:selection:changed': { target: IText };
+    'text:changed': { target: IText };
+    'text:editing:entered': { target: IText };
+    'text:editing:exited': { target: IText };
 
     // misc
     'contextmenu:before': SimpleEventHandler<Event>;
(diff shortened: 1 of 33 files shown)


src/EventTypeDefs.ts
```diff
@@ -0,0 +1,170 @@
...
+    // selection
+    selected: never;
+    deselected: never;
+
+    // tree
+    added: { target: Group | Canvas };
+    removed: { target: Group | Canvas };
+
+    // IText
+    // todo - move to file
```
ShaMan123: depends on #8420

src/mixins/object_interactivity.mixin.ts
```diff
@@ -19,7 +20,9 @@ type TOCoord = IPoint & {
 
 type TControlSet = Record<string, Control>;
 
-export class InteractiveFabricObject extends FabricObject {
+export class InteractiveFabricObject<
+  T extends TObjectEvents = TObjectEvents
+> extends FabricObject<T> {
```
ShaMan123, line 25: made it generic for future use, when #8420 is merged I will move the text events to there so TS won't show they are possible to listen to on other objects (same for group)

src/shapes/group.class.ts
```diff
@@ -3,6 +3,36 @@ import { Point } from '../point.class';
...
+  | 'layout_change'
+  | 'imperative';
+
+export type LayoutContext = {
+  type: LayoutContextType;
+  /**
+   * array of objects starting from the object that triggered the call to the current one
+   */
+  path?: Group[];
+  [key: string]: any;
```
ShaMan123, line 21: todo in follow up

src/typedefs.ts
```diff
@@ -131,7 +131,7 @@ export type TEvent<E extends Event = TPointerEvent> = {
   e: E;
 };
 
-export type TransformEvent = TEvent & {
+export type TransformEvent<E extends Event = TPointerEvent> = TEvent<E> & {
```
ShaMan123, line 134: I think all the types related to events should be extracted from this file
Didn't want to cause too many merge conflicts in parallel PRs

src/shapes/itext.class.ts
```diff
@@ -1,6 +1,13 @@
-//@ts-nocheck
+// @ts-nocheck
+import { TObjectEvents, XTransformEvent } from '../EventTypeDefs';
 import { FabricObject } from './fabricObject.class';
 
+export type ITextEvents = TObjectEvents & {
```
ShaMan123, line 5: todo in #8420

ShaMan123 (commented):
`src/canvas.class.ts` contains the fixes done in the PR

ShaMan123 (commented):
extracted all event types from `src/typedefs.ts` to `src/EventTypeDefs.ts`

src/EventTypeDefs.ts
```diff
@@ -0,0 +1,222 @@
...
+export type TEvent<E extends Event = TPointerEvent> = {
+  e: E;
+};
+
+export type BasicTransformEvent<E extends Event = TPointerEvent> = TEvent<E> & {
+  transform: Transform;
+  pointer: Point;
+};
+
+export type TModificationEvents =
```
ShaMan123, line 69: Up to here are types from typedefs

ShaMan123 (commented):
READY to MERGE

asturur:
@ShaMan123 i changed the description of the PR with the one i would have loved to find when i approached the PR.
That would have save me the first 15 minutes of `shit shit shit shit what is all this shit`.
Scrolling lines of combined types doesn't magically tell me your intentions on how you wanted to type those things and why, on a pr of 26 files my process is:

- first check everything quickly
- then close away the files that have no meaningful changes
- take notes on the other so i avoid to leave comments that maybe are clarified by some other file later
- then when i have a grasp of what is going on reset my expectations and actually comment on what is really just important/meaningful

this process requires time, an introduct really helps

src/canvas.class.ts
```diff
@@ -441,10 +441,12 @@ import { saveObjectTransform } from './util/misc/objectTransforms';
         this._objectsToRender = undefined;
         // removing active object should fire "selection:cleared" events
         if (obj === this._activeObject) {
-          this.fire('before:selection:cleared', { target: obj });
+          this.fire('before:selection:cleared', { deselected: [obj] });
```
asturur, line 444: So here i m thorn between deselected and maybeDeselected or something like that.
Typescript allow for self descovery of data structure, that means that we can care a bit less about consistency ( all events have target! ) and have names that make sense.
In this case deselected is past, while the before:selection:cleared can be interrupted and is not deselected yet.

Ideas?

src/EventTypeDefs.ts
```diff
@@ -0,0 +1,222 @@
...
+        };
+
+    // IText
+    'text:selection:changed': { target: FabricObject };
+    'text:changed': { target: FabricObject };
+
+    // misc
+    'contextmenu:before': SimpleEventHandler<Event>;
+    contextmenu: SimpleEventHandler<Event>;
+  };
```
asturur, line 222: this is useful, yes, but needs to be a suggestion and not a lockdown.
The developer needs to be able to do

`rect.fire('whateveriwant', ({ ...myStuff }) => { })`

I didn't try the branch, is this possible or locked?

asturur:
I left the only important comment i have, i would like to merge this work, meanwhile i ll search some other PR to merge, if i don't get answers ( and that is totally fine ) if tomorrow i have time i ll try to figure out the answer and in case fix it.

src/canvas.class.ts
ShaMan123: I get your point. A good point indeed

ShaMan123: MaybeDeselcted is bad naming. Target or desecltionTarget are better

src/EventTypeDefs.ts
ShaMan123: TS will complain.
I can look into it

ShaMan123: I think that if we accept string in addition to the event names TS won't show autocomplete.

ShaMan123:
Great description, point taken

ShaMan123:
I am taking a look now

ShaMan123:
Seems I managed to enable passing other events that are not part of the spec, with an additional overload to `on|once` signature.
However it seems that using the methods with an obj doesn't autocomplete the keys, but does autocomplete the callback. 

```ts
import { FabricObject } from "./fabricObject.class";
import { Group } from "./group.class";

new FabricObject({}).on('resizing',opt=>opt.)
new FabricObject({}).on('',opt=>opt.)
new FabricObject({}).on('resizing1',(opt)=>opt.)
new Group({}).once('layout',opt=>opt.)
new Group({}).on('dgdfg', (opt: { a: boolean }) =>opt.)
new Group({}).on({
    'mousedown:before':opt=>opt.transform,
    'mousedown':opt=>opt.button,
    'added': opt => opt.target,
    
    
})
```

ShaMan123:
Added IText event spec.
Good enough to merge.

asturur:
Gave a stab at solving conflicts

ShaMan123:
Joining now.
Group needs some care as well

ShaMan123:
This PR is an example why #8394 is important
Now after fixing conflicts I would have wanted a test to tell that types work (autocomplete) as expected

asturur:
> This PR is an example why #8394 is important Now after fixing conflicts I would have wanted a test to tell that types work (autocomplete) as expected

I understand that.
But i m thorn between having something that is another thing to test rather than a support.
We are going to add type tests, i m really looking at things to complete the migration before adding more stuff on top.

To be honest i don't even expect this kind of complexity in types, i don't think TS should be a replacemente for documentation.

ShaMan123:
> We are going to add type tests, i m really looking at things to complete the migration before adding more stuff on top.

This is vague

asturur:
> > We are going to add type tests, i m really looking at things to complete the migration before adding more stuff on top.
> 
> This is vague

Meaning that i m looking at things in order of priority to close the TS migration.

asturur (dismissed):

src/EventTypeDefs.ts
```diff
@@ -0,0 +1,228 @@
...
+
+export type ObjectPointerEvents = TPointerEvents<'mouse'>;
+export type CanvasPointerEvents = TPointerEvents<'mouse:'>;
+
+export type ObjectEvents = ObjectPointerEvents &
+  DnDEvents &
+  ObjectModifiedEvents & {
+    // selection
+    selected: never;
+    deselected: never;
```
asturur, line 175: to me looks like that selected and deselected do get `target` and optionally 'e' ( if they happen by mouse )

ShaMan123: I recall that the object is fired without context
But need to double check

ShaMan123:
I think we should merge and come back to it if necessary

src/EventTypeDefs.ts
ShaMan123: As opposed to the events fired on canvas

asturur: I verified and added e + target.

asturur (approved):

PR #8433, opened 2022-11-09
chore(TS): Collection

ShaMan123:
## Changes

- **BREAKING** `insertAt` signature - `(objects, index)` => `(index, ...objects)`. Done to comply with the rest of the methods and resemblance to `Array.splice`
- **BREAKING** removed `fabric.Collection`, but I can put it back

## Gist

Moved all methods using `super` call to a temporary super class of Group/Canvas.
Once both are migrated the super class won't be necessary


src/mixins/collection.mixin.ts
```diff
@@ -81,87 +72,86 @@
...
+    forEachObject<T>(
+      callback: (
+        this: T | undefined,
+        object: FabricObject,
+        index: number,
+        array: FabricObject[]
+      ) => any,
+      context?: T
+    ) {
+      this.getObjects().forEach(callback.bind(context));
```
asturur, line 85: note for andrea, try to remove the context and the bind.
Does it make sense with fat arrows?
We either don't use it, or use with prebound functions.

asturur:
Fabric.Collection wasn't really a public interface, whoever was extending from that for custom use cases can still use the createCollectionMixin.

Changes are good, i want review tomorrow morning slowly and merge this

src/mixins/collection.mixin.ts
ShaMan123: I think we can drop context

ShaMan123: I think I tried it and tested but decided not to commit

asturur: removed now

src/static_canvas.class.ts
```diff
@@ -35,7 +34,48 @@ import { CommonMethods } from './mixins/shared_methods.mixin';
    */
   // eslint-disable-next-line max-len
   fabric.StaticCanvas = fabric.util.createClass(
-    fabric.Collection,
+    class extends createCollectionMixin(CommonMethods) {
+      add(...objects: FabricObject[]) {
+        super.add(...objects);
+        objects.length > 0 && this.renderOnAddRemove && this.requestRenderAll();
```
asturur, line 40: One day i would like to go over all of those `this.renderOnAddRemove` and decide if we really need them.
Those are good for demos because you have to write less, but sometimes boring during normal application development becuase sometimes you are doing something complex and you never want to render automatically.

asturur (approved):

src/static_canvas.class.ts
ShaMan123: +1
Perhaps this is part of the scope of #8372

ShaMan123:
Fabulous!
Moving to Group/ActiveSelection

PR #8438, opened 2022-11-14
chore(TS): finalize path migration

ShaMan123:
## Description
 finalizes #8412 by reverting 027f20eca2e08df90e9113f49d27aa400a280b1b
see https://github.com/fabricjs/fabric.js/pull/8412#pullrequestreview-1178245147

It is a fix to a bad solution I did in #8083


asturur:
I don't get what is doing this PR apart just reverting the piece i intentioanlly removed.
Why do we want to change this? is not meaningful right now.
Can you try to explain me why is it?

ShaMan123:
setDimensions was changed to align with Polyline in previous PRs. You mentioned that it is weird that the function returns a point. 
I agreed and refactored Polyline.setDimensions in #8417 . This PR aligns Path to the same behavior.

asturur:
I can merge this and finish on the other side so the attribution of this cleanup and test is not lost.

asturur (approved):

PR #8443, opened 2022-11-17
chore(TS): Image

ShaMan123:

### DIFF
--- a/src/shapes/image.class.ts
+++ b/src/shapes/image.class.ts
@@ -406,7 +406,7 @@ export class Image extends FabricObject {
    * @param {LoadImageOptions} [options] Options object
    * @see https://developer.mozilla.org/en-US/docs/HTML/CORS_settings_attributes
    */
-  setSrc(src: string, { crossOrigin, signal }: LoadImageOptions) {
+  setSrc(src: string, { crossOrigin, signal }: LoadImageOptions = {}) {
     return loadImage(src, { crossOrigin, signal }).then((img) => {
       typeof crossOrigin !== 'undefined' && this.set({ crossOrigin });
       this.setElement(img);
--- a/src/shapes/image.class.ts
+++ b/src/shapes/image.class.ts
@@ -21,11 +21,7 @@ export type ImageSource =
   | HTMLCanvasElement;
 
 /**
- * Image class
- * @class Image
- * @extends FabricObject
  * @tutorial {@link http://fabricjs.com/fabric-intro-part-1#images}
- * @see {@link Image#initialize} for constructor definition
  */
 export class Image extends FabricObject {
   /**
@@ -152,12 +148,13 @@ export class Image extends FabricObject {
    * @return {Image} thisArg
    * @chainable
    */
-  setElement(element: ImageSource, options: object): Image {
+  setElement(element: ImageSource, options: object = {}): Image {
     this.removeTexture(this.cacheKey);
     this.removeTexture(`${this.cacheKey}_filtered`);
     this._element = element;
     this._originalElement = element;
-    this._initConfig(options);
+    this.set(options);
+    this._setWidthHeight(options);
     element.classList.add(Image.CSS_CANVAS);
     if (this.filters.length !== 0) {
       this.applyFilters();
@@ -174,6 +171,7 @@ export class Image extends FabricObject {
 
   /**
    * Delete a single texture if in webgl mode
+   * @todo make static and use instanceof WebGLBackend
    */
   removeTexture(key: string) {
     const backend = fabric.filterBackend;
@@ -193,6 +191,7 @@ export class Image extends FabricObject {
     ['_originalElement', '_element', '_filteredEl', '_cacheCanvas'].forEach(
       (element) => {
         cleanUpJsdomNode(this[element as keyof this]);
+        // @ts-expect-error disposing
         this[element] = undefined;
       }
     );
@@ -618,16 +617,6 @@ export class Image extends FabricObject {
       fabric.document.getElementById(element) || element,
       options
     );
-  }
-
-  /**
-   * @private
-   * @param {Object} [options] Options object
-   */
-  _initConfig(options: object) {
-    options || (options = {});
-    this.setOptions(options);
-    this._setWidthHeight(options);
   }
 
   /**
--- a/src/shapes/image.class.ts
+++ b/src/shapes/image.class.ts
@@ -759,8 +759,8 @@ export class Image extends FabricObject {
       resizeFilter && enlivenObjects([resizeFilter], filterOptions),
       enlivenObjectEnlivables(object, options),
     ]).then(
-      ([img, filters = [], [resizeFilter], hydratedProps]) =>
-        new Image(img, {
+      ([el, filters = [], [resizeFilter] = [], hydratedProps = {}]) =>
+        new Image(el, {
           ...object,
           crossOrigin,
           filters,
--- a/src/shapes/image.class.ts
+++ b/src/shapes/image.class.ts
@@ -478,6 +478,7 @@ export class Image extends FabricObject {
       this._filteredEl = null;
       this._filterScalingX = 1;
       this._filterScalingY = 1;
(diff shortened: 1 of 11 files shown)


src/shapes/image.class.ts
```diff
@@ -1,868 +1,832 @@
...
+    this.filters = [];
+    this.cacheKey = `texture${FabricObject.__uid++}`;
+    this._initElement(element, options);
+  }
+
+  /**
+   * Returns image element which this instance if based on
+   */
+  getElement() {
+    return this._element;
```
ShaMan123, line 143: changed

src/shapes/image.class.ts
```diff
@@ -1,868 +1,832 @@
...
+      return this.src || '';
+    }
+  }
+
+  /**
+   * Alias for getSrc
+   * @param filtered
+   * @deprecated
+   */
+  getSvgSrc(filtered?: boolean) {
```
ShaMan123, line 394: can be removed

src/shapes/image.class.ts
```diff
@@ -1,868 +1,832 @@
...
+    this._filterScalingY = canvasEl.height / this._originalElement.height;
+  }
+
+  /**
+   * Applies filters assigned to this image (from "filters" array) or from filter param
+   * @method applyFilters
+   * @param {Array} filters to be applied
+   * @param {Boolean} forResizing specify if the filter operation is a resize operation
+   */
+  applyFilters(filters = this.filters || []) {
```
ShaMan123, line 468: failing tests for some reason

src/shapes/image.class.ts
```diff
@@ -1,868 +1,832 @@
...
-          fabric.document.getElementById(element) || element,
-          options
-        );
-      },
-
-      /**
-       * @private
-       * @param {Object} [options] Options object
-       */
-      _initConfig: function (options) {
```
ShaMan123, line 667: removed

src/shapes/image.class.ts
asturur: this is fine, returning the empty object wasn't really smart.

asturur:
the merge conflicts aren't great here. Must be my filter change.
Can you prioritize fixing this merge conflict before doing other fabricJS work?

ShaMan123:
Tomorrow I will finish this up

src/shapes/image.class.ts
```diff
@@ -1,869 +1,810 @@
...
+    if (!filter || (scaleX > minimumScale && scaleY > minimumScale)) {
+      this._element = elementToFilter;
+      this._filterScalingX = 1;
+      this._filterScalingY = 1;
+      this._lastScaleX = scaleX;
+      this._lastScaleY = scaleY;
+      return;
+    }
+    if (!fabric.filterBackend) {
+      fabric.filterBackend = fabric.initFilterBackend();
```
ShaMan123, line 442: conflict

src/shapes/image.class.ts
```diff
@@ -1,869 +1,810 @@
...
+      this._element = this._filteredEl;
+      this._filteredEl
+        .getContext('2d')
+        .clearRect(0, 0, sourceWidth, sourceHeight);
+      // we also need to resize again at next renderAll, so remove saved _lastScaleX/Y
+      this._lastScaleX = 1;
+      this._lastScaleY = 1;
+    }
+    if (!fabric.filterBackend) {
+      fabric.filterBackend = fabric.initFilterBackend();
```
ShaMan123, line 508: same

src/shapes/image.class.ts
```diff
@@ -1,869 +1,810 @@
...
+  }
+
+  /**
+   * Paint the cached copy of the object on the target context.
+   * it will set the imageSmoothing for the draw operation
+   * @param {CanvasRenderingContext2D} ctx Context to render on
+   */
+  drawCacheOnCanvas(ctx: CanvasRenderingContext2D) {
+    ctx.imageSmoothingEnabled = this.imageSmoothing;
+    super.drawCacheOnCanvas(ctx);
```
ShaMan123, line 548: could the error be related to this line?

ShaMan123 (commented):
triaging

src/shapes/image.class.ts
```diff
@@ -1,869 +1,810 @@
...
+    }
+  }
+
+  /**
+   * Delete a single texture if in webgl mode
+   * @todo make static and use instanceof WebGLBackend
+   */
+  removeTexture(key: string) {
+    const backend = fabric.filterBackend;
+    if (backend && backend.evictCachesForKey) {
```
ShaMan123, line 182: change!

src/shapes/image.class.ts
```diff
@@ -1,869 +1,801 @@
...
-
-        // needs to clear out or WEBGL will not resize correctly
-        this.removeTexture(this.cacheKey + '_filtered');
-
-        if (filters.length === 0) {
-          this._element = this._originalElement;
-          this._filteredEl = null;
-          this._filterScalingX = 1;
-          this._filterScalingY = 1;
-          return this;
```
ShaMan123, line 504: found it!

ShaMan123 (commented):
Finally after redoing everything I found the very very minor error in coversion.
It was hidden in `fromObject`, didn't pass `src`

src/shapes/image.class.ts
```diff
@@ -1,869 +1,797 @@
...
+
+export type ImageSource =
+  | HTMLImageElement
+  | HTMLVideoElement
+  | HTMLCanvasElement;
+
+/**
+ * @tutorial {@link http://fabricjs.com/fabric-intro-part-1#images}
+ */
+export class Image extends FabricObject {
```
asturur, line 28: this name is a pain as it is Object now. doing new Image() can't be really done.

src/shapes/image.class.ts
```diff
@@ -1,869 +1,797 @@
...
+        offset = rHeight - pHeight / scaleY;
+        if (pAR.alignY === 'Mid') {
+          cropY = offset / 2;
         }
-        var scaleX = this._filterScalingX,
-          scaleY = this._filterScalingY,
-          w = this.width,
-          h = this.height,
-          min = Math.min,
-          max = Math.max,
```
asturur, line 604: this was purely done to lend an hand to the minifier.
And i occasionally do, but i also think probably terser is smart enough to do it itself?

src/shapes/image.class.ts
```diff
@@ -1,869 +1,797 @@
...
-       * @param {Object} [options] Options object
-       * @return {fabric.Image} thisArg
-       * @chainable
-       */
-      setElement: function (element, options) {
-        this.removeTexture(this.cacheKey);
-        this.removeTexture(this.cacheKey + '_filtered');
-        this._element = element;
-        this._originalElement = element;
-        this._initConfig(options);
```
asturur, line 176: mmm this is a weird change.
If i do setElement today, i can pass to options random things and it will be applied by setOptions.
Now we locked to width/height.
To be fair we should leave the setOptions call in.
Should we move set(options) inside here? so we maintain the old logic, or we should remove size options.

asturur: I think simplification is better.
setElement should set the element.
We can comment that setElement will reset width/height but not cropX and cropY ( that is weird by itself )

asturur (approved):

src/shapes/image.class.ts
ShaMan123: I understand 
Too messy IMO

src/shapes/image.class.ts
ShaMan123: I agree, wasn't sure what to do cause if we name it FabricImage then maybe every thing should become FabricXXX

src/shapes/image.class.ts
ShaMan123: I did cause of simplification, too many side effects are confusing. The dev can call set whenever. `setElement` should focus on its task. It makes sense that setting dimensions is part of its scope

PR #8449, opened 2022-11-19
ci(): Use simpler disk stats for the bundler

asturur:
Reading file size from fsStatSync for now that seems more accurate for the bytes.
Created a new function to don't remove the old one.


asturur:
@ShaMan123 when we have time/will we can build something nicer, for now this is completely fine.

ShaMan123:
I Agree. Is good enough.
We should merge #8402 
It fixes a few things

PR #8452, opened 2022-11-20
refactor(polyline): introduces resizingOffset

luizzappa:
## Motivation

The calculation of the poly's BBOX alters the original `height` and `width` values ​​of the object's, causing a different behavior from other fabricjs objects and making the return of the `_getNonTrasformedDimensions` function incorrect.

## Description

It introduces `resizingOffset` property that stores how much we must add to the `width` and `height` of the object to find the correct value for the BBOX and does not change the originals values of `width` and `height`.

## Changes

In addition of the `resizingOffset` property, extend `transformMatrixKey` given that the properties `strokeUniform`, `strokeLineJoin`, `strokeLineCap`, `strokeMiterLimit` change the `width`/`height` of the BBOX, but the original matrix transform cache does not take into account these changes.

This was causing a bug in the controls when changing some property that the cache is not aware of.

image

As the `strokeMiterLimit` may or may not change the final `width`/`height`, I thought it would be better to store the `resizingOffset` property in the cache key instead of each of the properties individually (`storkeLineJoin, StrokeLineCap, strokeUniform, strokeMiterLimit`)

```
  transformMatrixKey(...args: any): string {
    const oldKey = super.transformMatrixKey(...args),
      newPart = this.resizingOffset || '';
    return oldKey + newPart;
  }
```


ShaMan123:
[diff](https://github.com/luizzappa/fabric.js-bbox/compare/test-stroke-bbox...fix-TransformedDimensions)

ShaMan123:
If I understand correctly what @asturur means we are going to take a closer look at width/height in general. So this effort might be redundant. Hopefully we will discuss it in the upcoming days.

ShaMan123:
we will look at this in the context of #8767

ShaMan123:
@luizzappa would want to try to do polys in #8767 ?

luizzappa:
> @luizzappa would want to try to do polys in #8767 ?

@ShaMan123, working on fabric.js internals is always rewarding (and most of the time fun)   :)

The only problem is that I don't have a lot of time, so I need a direction of what has to be done and preferably smaller blocks. If the task fits these requirements, you can count on me. Trying to understand everything that is happening in that PR would take some time that I don't have at the moment.

ShaMan123:
Great
I will ping you when I think it is time
And explain it all

ShaMan123:
closing in favor of #8374

PR #8457, opened 2022-11-22
fix(IText): key maps

ShaMan123:
## Motivation
https://github.com/fabricjs/fabric.js/discussions/8419#discussioncomment-4201286

## Description

Wrongly migrated properties


ShaMan123:
> **Build Stats**
> 
> fabric	1061.814 (**+1.733**)	313.087 (**+0.515**)

I find it hard to believe the stats are correct

ShaMan123:
ready

ShaMan123 (commented):
added a test

asturur:
@ShaMan123 i refactored it to be its own const, little bit typed and kept on the default values, otherwise was copied every instance and not modificable as a whole.

asturur:
> > **Build Stats**
> > file / KB (diff)	bundled	minified
> > fabric	1061.814 (**+1.733**)	313.087 (**+0.515**)
> 
> I find it hard to believe the stats are correct

I think the stats are correct, before as a type it would disappear, while now it stays in the code, so that's normal i think.

ShaMan123:
OK
#7865 will change this again I believe, but doesn't have to.
Let's merge

PR #8462, opened 2022-11-23
chore(TS): Shadow

ShaMan123:
## Description

Migrate Shadow and remove stub type

### DIFF
--- a/src/shadow.class.ts
+++ b/src/shadow.class.ts
@@ -178,10 +178,12 @@ export class Shadow {
       return data;
     }
 
-    const proto = Shadow.prototype;
+    const defaults = Shadow.prototype;
     const out: Record<string, unknown> = {};
     for (const key in data) {
-      if (data[key as keyof typeof data] !== proto[key as keyof typeof data]) {
+      if (
+        data[key as keyof typeof data] !== defaults[key as keyof typeof data]
+      ) {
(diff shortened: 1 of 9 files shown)


src/shadow.class.ts
```diff
@@ -1,240 +1,217 @@
...
    * Regex matching shadow offsetX, offsetY and blur (ex: "2px 2px 10px rgba(0,0,0,0.2)", "rgb(0,255,0) 2px 2px")
-   * @static
-   * @field
-   * @memberOf fabric.Shadow
    */
   // eslint-disable-next-line max-len
-  fabric.Shadow.reOffsetsAndBlur =
+  static reOffsetsAndBlur =
     /(?:\s|^)(-?\d+(?:\.\d*)?(?:px)?(?:\s?|$))?(-?\d+(?:\.\d*)?(?:px)?(?:\s?|$))?(\d+(?:\.\d*)?(?:px)?)?(?:\s?|$)(?:$|\s)/;
-})(typeof exports !== 'undefined' ? exports : window);
```
asturur, line 240: i hate this regex and i wonder if it should exists

asturur (approved):

asturur (approved):

src/shadow.class.ts
ShaMan123: It can be simplified for sure

PR #8466, opened 2022-11-23
fix(Itext): show incorrect pointer position after scale changed

thuytv-gl:
After @ShaMan123 migrated the Pointer.transform util, point already included the scales so no need to multiply object's width/height with scaleX/scaleY

https://user-images.githubusercontent.com/107306319/203584307-4eafca7f-3a47-4800-8691-63c77d487646.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(Itext): show incorrect pointer position after scale changed
 - chore(TS): migrate text classes/mixins [#8408](https://github.com/fabricjs/fabric.js/pull/8408)
 - chore(TS): migrate Collection [#8433](https://github.com/fabricjs/fabric.js/pull/8433)
 - ci(): Simplify filestats even more [#8449](https://github.com/fabricjs/fabric.js/pull/8449)
--- a/src/mixins/itext_click_behavior.mixin.ts
+++ b/src/mixins/itext_click_behavior.mixin.ts
@@ -249,7 +249,7 @@ export abstract class ITextClickBehaviorMixin extends ITextKeyBehaviorMixin {
       lineIndex = 0;
     for (let i = 0, len = this._textLines.length; i < len; i++) {
       if (height <= mouseOffset.y) {
-        height += this.getHeightOfLine(i) * this.scaleY;
+        height += this.getHeightOfLine(i);
         lineIndex = i;
         if (i > 0) {
           charIndex +=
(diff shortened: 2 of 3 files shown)


ShaMan123:
Nice. I will test

ShaMan123 (approved):
Good catch!

ShaMan123:
@thuytv-gl Thanks! More contributions are welcome!

thuytv-gl:
@ShaMan123 thanks, I'm working on vertical text for CJK characters, picked it up from an old repository that I pushed in the past for my job.
The repo is here https://github.com/thuytv-scuti/fabric-CJK-vertical, do you have any tips for me.
Thanks!

ShaMan123:
I know for a fact we want vertical text to be part of core (@asturur talked to me about that). If you are willing to PR I can help.
My domain in text is RTL since I speak Hebrew.
Vertical text is something I/we don't know enough of, in terms of UX/UI and basically everything. So help will be appreciated.
I wonder if we could abstract the methods to allow both vertical and horizontal writing. If not we can add dedicated methods.
I believe #7864 will give you some insight of the internals you need to modify.
I would like to have a better notion of logic, maybe plan it so we have a delegate perhaps that is in charge of resolving the next char etc. given the direction. Then we can expose the methods we need on that delegate and make things simpler.
Are you interested?

ShaMan123:
#7796 is another PR that tends more to the delegate approach (BidiResolver)

ShaMan123:
And of course, my tip would be fix it at core level, or else you will always have some kind of discrepancy.
But I am bias cause I want it part of core.

ShaMan123:
And if you decide for maybe write a description of expected behavior + some videos.

asturur:
Every change that bring usability improvements to people of different cultures is welcome.

asturur:
Also @thuytv-gl if you would like to spend some words on how vertical text is supposed to work i m curious.
Does it have composition? how does it looks like?
Do you have any native application that does it correctly and that you can record a movie of?

thuytv-gl:
Hi @asturur,
I'm a Vietnamese and we use latin alphabet, but I have some knowledge about vertical writing.
The input will not be so differences like:
- Write from right to left, top top bottom
- Chinese use full sized Han
- Japanese has Han and 2 more alphabet system
  - Hiragana - full size + half size
  - Katakana - full size + half size
  - Numbers - full size + half size
 - Brackets and latins will be flip
 - Input Han with composition
 
(Need a native speaker to confirm!)
### Here is a demo

https://user-images.githubusercontent.com/107306319/204084238-8af0cd98-8aea-45e5-8169-165297a9c7dd.mp4

ShaMan123:
It looks really good.
My input would be that I would like you to work on a branch that is forked with the fixes I made to RTL so we don't have a lot of conflicts. I haven't migrated it yet to v6. But on the other hand I don't want it to hold you back.
Let's see what @asturur has to say.

PR #8455, opened 2022-11-21
chore(TS): Group

ShaMan123:
## Description

Migrate Group & ActiveSelection

## Changes

- removed redundant deep clone of option in `fromObject` 4e8db1adf
- removed `Rect` `TObject` stub types

## Gist

Will have conflicts with #8431 - should resolve with `union` strategy, sort of. I can do it

PR #8468, opened 2022-11-25
fix(PatternBrush): `getPatternSrc`, rm `getPatternSrcFunction`

ShaMan123:
## Motivation

https://github.com/fabricjs/fabric.js/pull/8455#discussion_r1032098929

## Description

`getPatternSrcFunction` was weird, mutating the function itself by converting it to string. Bad practice and no need for it.
It seemed that in v4/v5 Pattern stopped accepting strings to execute as functions, breaking PatternBrush. It was forgotten since (I am guessing) it had no tests.
So I have removed the method to align with Pattern and added tests that cover these cases.
Refactored `PatternBrush.getPatternSrc` signature to accept color as an arg. Am not sure it is needed to say the truth. We can use `this.color` because the function runs in context always, seems so.

## Changes

Non breaking since `this.color` will still work as expected in `getPatternSrc`

## Gist

Added visual tests.
Found  a leak in test config.


test/visual/freedraw.js
```diff
@@ -2056,9 +2015,43 @@
...
+    else {
+      fabric.config.configure({ browserShadowBlurConstant: BROWSER_SHADOW_BLUR[getBrowserAgent()] });
+    }
+    fabric.config.configure({
+      enableGLFiltering: false
+    });
+    fabric.Object.prototype.objectCaching = true;
+  });
+  hooks.after(() => {
+    // fabric.config.restoreDefaults();
```
ShaMan123, line 2048: This is really bad.
Other suites depend on some configuration made in this test that leaks.

ShaMan123 (commented):
READY and TESTED

asturur:
The patternBrush is more likely a demo / blueprint rather than something useful.
I don't think the color parameter is needed, all other brushes work using the color property on the instance

asturur:
on top of that getPatternSrcFunction imho was a possible security flaw.
Imaging adding to this.color a kind of function that then gets executed with the same rights as any javascript loaded on the website.
Definetely a security issue, reason why Pattern does not have that function anymore

asturur (approved):

PR #8470, opened 2022-11-26
feat(Textbox): min/max width

### LINKED ISSUE #2376: Entering text in a Textbox object expands it when there aren't newlines

### LINKED ISSUE #5114: Support for word-breaking on Textbox

### LINKED ISSUE #5911: How can we set MaxWidth in fabric.js canvas Textbox ?

### LINKED ISSUE #7888: How to limit the maxWidth of textbox？

ShaMan123:
## Motivation

https://github.com/fabricjs/fabric.js/issues/7888
https://github.com/fabricjs/fabric.js/discussions/7891
Wants to limit text input

https://github.com/fabricjs/fabric.js/issues/5911
Wants to avoid the textbox going out of canvas when inputing text

https://github.com/fabricjs/fabric.js/issues/5114
Wants to wrap text when reach a certain width becase of long words

It seems that maxWidth should indeed break the words.
so while starting to implement it is fine, i don't think it closes the related issues.

_Originally posted by @asturur in https://github.com/fabricjs/fabric.js/issues/7981#issuecomment-1153081614_
      
closes #7888 
closes #7891
closes #5911
related #1184
closes #2376 
closes #5114
closes #1162

## Description

port #7981
This PR supports `minWidth` `maxWidth`, breaking lines to fit the size. Splitting words (`word-break`) is also possible when using with `splitByGrapheme` (use `minWidth` as well or set it to true after initialization). 

## Changes

I have refactored logic not to care if  `maxWidth < minWidth`. Values are not changed by fabric and everything works!
`maxWidth` will be overriden by `minWidth` and by `_actualMaxWidth` (min values + largest word) in case of conflict.

### splitByGrapheme + maxWidth

https://user-images.githubusercontent.com/34343793/173128644-76ff3024-bdd8-4708-862a-4caf544de091.mp4

### maxWidth

https://user-images.githubusercontent.com/34343793/173128681-de8311c4-5c18-45bc-bf55-37db4f23d32b.mp4

### onInput + maxWidth
https://user-images.githubusercontent.com/34343793/173128702-0191fda3-e4c3-4844-8ea9-5154473bb2d9.mp4


ShaMan123 (commented):
updated from master

ShaMan123 (commented):
updated from master

Matergi:
A solution could be to create a new Fabric class that extends the Fabric.Textbox class. Here's an example:

```js
const LimitedTextbox = fabric.util.createClass(fabric.Textbox, {
  onInput: function (e: any) {
    const oldText = `${this.text}`.replaceAll('\n', '');

    /**
     * Sometimes, when you insert a line break, it doesn't appear to Fabric as an 'insertLineBreak' but simply sets the date to null.
     * Therefore, it's important to check that it's not deleting any characters, and if the date is null, it means that a new line is being inserted
     */
    if (e.inputType === 'insertLineBreak' || (e.inputType !== 'deleteContentBackward' && !e.data)) {
      this._textBeforeEdit = oldText;
      this.hiddenTextarea.value = oldText;
      this.text = oldText;
      return;
    } else if (e.inputType !== 'deleteContentBackward') {
      if (this.width > this.maxWidth) {
        this._textBeforeEdit = oldText;
        this.hiddenTextarea.value = oldText;
        this.text = oldText;
        return;
      }

      if (this._textLines.length > this.maxLines) {
        this._textBeforeEdit = oldText;
        this.hiddenTextarea.value = oldText;
        this.text = oldText;
        return;
      }
    } else {
      /**
       * It's needed because otherwise, if I write many characters, he automatically adjusts the width even if the visible text doesn't change.
       * So, every time I delete something, I reset the width.
       * I do it during deletion; otherwise, if I do it in the maxWidth check, I never block it, and it would keep writing
       */
      this.width = this.maxWidth;
    }

    this.callSuper('onInput', e);
  },
});

...

textbox = new LimitedTextbox('custom text', {
  maxLines: 1,
  maxWidth: 100,
});
```

ShaMan123:
I will probably abandon this PR for a better design.
we exposed methods that can be overridden to achieve text layout.
In the future I hope we provide a robust API that has layout management

ShaMan123:
Stale, can be achieved by overriding `_wrapText`, `getGraphemeDataForRender`

PR #8430, opened 2022-11-07
feat(): DrawShape, DrawOval, DrawPoly

ShaMan123:
## Motivation

We discussed adding a poly brush to core.
I need it for a project so why not. Here it is.

closes #6319

## Description

Interaction managers that are in charge of drawing a shape.
I have sticked to brush impl though I don't agree with the event naming and adding the shape to canvas by default. Thought that was best until we refactor all brushes (#8476 ).

## Changes

`_onDoubleClick` calls brush method
Furture PR should make all pointer events funnel through the active brush.

## Gist

~~First merge #8417~~
closes #8428 

### DrawShapeBase
base class to shape creation interaction

### DrawShape
Simple and good pointer interaction creating a shape

### DrawOval
same, extends `DrawShape`

### DrawPoly
- mouseDown/Up adds a point, dblclick ends interaction

What do you think about naming?

vpt isn't there yet, see videos below, I think this is a bug in canvas vpt when skewing, unrelated to the PR

## In Action

https://user-images.githubusercontent.com/34343793/200395859-b680b393-169c-40cc-9f38-996afb8bdb64.mp4

**centered + shiftKey**

https://user-images.githubusercontent.com/34343793/200395563-aed69f8d-8030-4104-ade9-71ac95bc79de.mp4

**Unrelated and unexpected caching effect at the end**

https://user-images.githubusercontent.com/34343793/200350928-43b8dc16-d7a0-4b26-a467-8e0bba44e99f.mp4

https://user-images.githubusercontent.com/34343793/200351457-66af6bfb-6f18-4969-bf61-5beeef5ed5ba.mp4

**vpt**

https://user-images.githubusercontent.com/34343793/200396413-482911b8-0028-4168-bc97-375adaac5af6.mp4

**With vpt + objectCaching**
BBox is off!

https://user-images.githubusercontent.com/34343793/200351370-253f1ca3-edc3-4299-ab0e-f89ecbc6164b.mp4

PR #7880, opened 2022-04-17
feat(V6): brush layering POC

ShaMan123:
https://user-images.githubusercontent.com/34343793/163721860-fbb5989e-568d-455d-8c67-01a629931960.mp4

This is a nice POC that shows how to use Layer in an advanced way as described in https://github.com/fabricjs/fabric.js/pull/7860#issuecomment-1094571580
Notice that the brush is drawn below the red static rect as well as below the overlayColor. It's because the Layer is placed there.

[Edit fabric_app (forked)](https://codesandbox.io/s/fabric-app-forked-4lrm23?fontsize=14&hidenavigation=1&module=%2Fsrc%2FApp.tsx&theme=dark)

View the [diff](https://github.com/fabricjs/fabric.js/compare/v6-layer...v6-brush-layering)

PR #8476, opened 2022-11-28
feat(Brush API): `Observable`, `clipPath`

### LINKED ISSUE #5431: clipPath for canvas ignores free drawing

### LINKED ISSUE #6955: Adding Freehand with ClipPath to Object Group

ShaMan123:
## Motivation

closes #7175 
closes #7134 
closes #6955 
closes #5431

## Description

This PR is a port of #7175.
It sets sail to the new Brush API, which IMO should not be called brush anymore.
As part of the new API, brush now extends `Observable` and is subscribed to all canvas events by default.
I believe this change is so good that it will unlock lots of options including moving interaction code (e.g. group selector) to a dedicated brush.
I also think `freeDrawingBrush` should be renamed and become an array of interactions (this will enable supporting multi touch interactions).
To allow all this I have exposed a long desired feature, `FabricEvent#preventDefault`. Brushes listen to the before event (of down, move, up) and prevent default in case they operate and then canvas knows not to run the default action. This is great since there are interactions that should not prevent default behavior such as visualizing pointers etc.
In order to make brush flexible in terms of how it interacts with canvas events I have exposed the `subscribe` method that wires the brush logic to the canvas events so subclassing this method is very powerful.

## Changes

**This is a big PR but not a huge one. There are a lot of goldens changed.**

API changes:

- Brush subclasses Observable
- Observable now fires a `FabricEvent` instance to support `preventDefault` and `stopPropagation`
- Brush has a `subscribe` method that connects between lifecycle methods to canvas events. This way a brush can start its lifecycle when contextmenu event is fired instead of when mousedown is fired w/o changing logic accept the subscribe method
- Created `SimpleBrush`: a subclass of `BaseBrush` that is wired to interact with a simple `down-move-up` lifecycle
- Exposed a common `render` + `_render` method on `BaseBrush` which conforms to Object#render and is in charge of:
   1. Preparing the context (transformation etc.)
   2. calling `_render` which is implemented by each brush
   3. render clipPath
   4. cleanup
- enable/disable brush
- brush doesn't add result to canvas anymore and fires `interaction:completed` instead

Minor Changes:

- Added a `clipPath` property to `BaseBrush` and exposed clip path methods
- Added logic to apply the clip path to the resulting object created by the brush
- Moved common logic to base brush
- private flag `_isCurrentlyDrawing` is now a method `isCurrentlyDrawing` (doesn't have to be private)
- canvas fires `resize` from `setDimensions`, brush listens to it to call `_setBrushStyles` instead of canvas doing it
- canvas down, move, up default behavior has been extracted to dedicated methods for ease coding
- renamed `_onResize` => `_onWindowResize` (confusing)
- moved `_setDimensionsImpl` to `Canvas` instaed of `SelectableCanvas`
- removed `button` from `TPointerEventInfo` as we discussed
- deprecated `Canvas#isFreeDrawing` in favor of a fine grained brush option `enabled`
- Moved freedraw goldens to a dedicated folder and added tests that cover this PR

BREAKING:
  - brush doesn't add result to canvas anymore
  - `path:created`, `before:path:created` events are deprecated, use `interaction:completed` instead
  - `_render` method is now protected, use `render` instead
  - renamed `PencilBrush#convertPointsToSVGPath` => `getPathFromPoints`
  - `Canvas#freeDrawingCursor` => `Brush#cursor`
  - `Canvas#isFreeDrawing` => `brush.enable/disable`

## Gist

- [ ] first merge #8430 https://github.com/fabricjs/fabric.js/compare/poly-brush2...clip-brush
- [ ] TODO: PencilBrush doesn't respect `limitToCanvasSize`

## In Action (from #7175)

~~https://codesandbox.io/s/upbeat-microservice-mkbzs?file=/src/App.tsx~~
[STALE sandbox](https://codesandbox.io/s/fabric-app-forked-9hcztm?fontsize=14&hidenavigation=1&module=%2Fsrc%2FBrushClipPath.tsx&theme=dark)

[Edit fabric_app (forked)](https://codesandbox.io/s/fabric-app-forked-n57e00?fontsize=14&hidenavigation=1&module=%2Fsrc%2FBrushClipPath.tsx&theme=dark)

Pay no attention to eraser bug. It's fixed by the group PRs.
ezgif com-gif-maker (3)

**With Eraser patched up**
ezgif com-gif-maker (4)

image

**`absolutePositioned`** + viewportTransform
ezgif com-gif-maker (6)


src/brushes/base_brush.class.ts
```diff
@@ -161,6 +169,88 @@ export abstract class BaseBrush {
...
+  }
+
+  /**
+   * @private
+   * @param {CanvasRenderingContext2D} ctx
+   * @param {FabricObject} clipPath
+   */
+  drawClipPathOnCache(ctx: CanvasRenderingContext2D, clipPath: FabricObject) {
+    // TODO: no proto calls
+    FabricObject.prototype.drawClipPathOnCache.call(this, ctx, clipPath);
```
ShaMan123, line 188: I don't like the proto calls but for now it is what it is

ShaMan123 (commented):
clip path is ready!
I have adapted the Eraser code but it is still in a broken state. I will handle it in a follow up. I need it as well so I will push the new API forward with Eraser in mind.
Next step I believe will be extending Observable.

ShaMan123:
## Brush API
- blocking flag: if false canvas will handle the event regardless of brush
- extends Observable: **all** canvas events (including non pointer events) will be retargeted to brush, brush will listen to relevant events and execute logic accordingly. This way the dev can use on/off to fine tune behavior.
- Simultaneous multiple active brushes: 
   - assign a canvas element/context to for each brush to draw on using #7860. This will enable many things e.g. tracking users pointers while drawing. 
   - This will also make it possible to draw stuff in respect to object stacking and will make bg/overlay/topContext redundant.
   - And allow to have multiple active brushes
- Move active selection to a brush including groupSelector?
- Handle animations in a dedicated brush?
- lifecycle (not very opinionated):
  - start
  - step: render a step rather than the entire state
  - end

@asturur thoughts?

ShaMan123:
I am thinking it might be a subclass of Layer

ShaMan123:
Extending Group is a good idea. Even only for exposing one pattern less and sticking to good old Object lifecycle

ShaMan123:
So my question would be:
Can we make Object an imperative renderer? Can we use the cache canvas as the context the "brush" will draw over?

ShaMan123:
@asturur I have encountered a bug in brush design I would like to discuss in order to find the best solution.
Brush respects vpt in a very specific way. Its position shouldn't be affected by vpt but its strokeWidth should?
However if vpt changes while drawing... not good. Position is changed.
I have a POC working with `absolutePointer` but I am not sure I will manage to respect `strokeWidth`. Generally speaking interactions are absolute and are not subject to vpt changes. Am I right to assume that?
Should we break brush and make it not subject to vpt and perhaps expose a transform matrix option, same as object?
Thoughts?
This is holding me back.

ShaMan123 (commented):
updated from master

ShaMan123:
This PR is GREAT!

asturur:
@ShaMan123 i think there are some things that i agree with in this PR.
Do you mind if i start slicing them and importing in smaller units?

ShaMan123:
Go ahead
You should first merge #8380, or at least work with the diff against it if you don't want to work on that now.

ShaMan123:
This is was good refactor and provided a promising direction
However it has become stale
If in need I will probably rewrite what I need so that:

- Brush is an event target
- All data is in the viewport
- It is bound to its own layer #7880

PR #8479, opened 2022-11-29
ci(): PR tagging + prerelease on merge

ShaMan123:
## Motivation

As discussed we want CI to release an rc version on PR merge.
This will make my life easier since I can work on the release candidate instead of continuously maintaining a fork and having to remember what changes were breaking and so on. 
It opens fabric for early adopters.
And it will make bug triaging a lot easier since we will have a version for each PR as well as make migrating between versions easier (the dev could migrate one prerelease at a time in case of breaking changes so they aren't overwhelmed).

## Description

This PR adds an action with 2 jobs:
1. bump-version: runs on PRs and increments package version according to the release label on the PR (changing the label is possible)
2. release-on-push: once the PR is merged to master a release will be created and published. 

## Gist

By default a PR will be incremented by a prerelease.
The `release:XXX` labels control the incrementing and `norelease` blocks it.  
There are labels (`CI/CD`, `docs`) that will block bumping the version only if there isn't a release label, meaning they block default incrementing.
Then once the PR is merged a release will be created or updated **unless** the PR has a blocking label (`norelease`) or if the commit includes the key word `norelease`.

It seems the release action doesn't run after the release is created via the push action so I call it instead.


.github/workflows/version.yml
```diff
@@ -0,0 +1,104 @@
...
+            const LABEL_PREFIX = 'release:';
+            const releaseLabels = context.payload.pull_request.labels
+              .map((label) => label.name)
+              .filter((label) => label.startsWith(LABEL_PREFIX));
+            if (releaseLabels.length > 1) {
+              throw new Error(`Found more than one release label: ${releaseLabels.join(', ')}`);
+            }
+            // increment version
+            const nextVersion = shouldInc ?
+              semver.inc(version, releaseLabels[0]?.replace(LABEL_PREFIX, '') || 'prerelease', 'rc'):
```
ShaMan123, line 66: rc can be changed to whatever, beta etc.

ShaMan123 (commented):
This PR is ready. Needs to be tested after merging to see what happens in 2 cases:
- version changed in PR
- version didn't change

Still not able to figure out why labelling a PR doesn't trigger the workflow.

ShaMan123 (commented):
I dumped the action in favor of using the rest API directly. Much better.
Tested all cases, it's good to go.
By default a PR will be incremented by a prerelease.
The `release:XXX` labels control the incrementing and `norelease` blocks it.
Then once the PR is merged a release will be created or updated.

ShaMan123:
There seems to be an issue when the bot commits tests stop, maybe to prevent an infinite loop?

ShaMan123 (commented):
0

ShaMan123 (commented):
trigger?

ShaMan123 (commented):
trigger?

ShaMan123 (commented):
trigger!

ShaMan123 (commented):
Labelling doesn't trigger but reviewing does

ShaMan123 (commented):
bump

ShaMan123 (commented):
Did another iteration.
Release labels now take precedence over other labels, e.g. `CI/CD` is marked as a label that should bump the version. As long as there isn't a release label the version will not bump. 

Release action results: https://github.com/fabricjs/fabric.js/releases/tag/v6.0.0-rc.0

Now the PR is perfect

ShaMan123 (commented):
bump test

ShaMan123 (commented):
bump fin

asturur:
temporarly closing @ShaMan123 this code is actually pushing code to npm from here. please don't push more commits for now.

ShaMan123:
It isn't this PR
I drafted a manual release.

.github/workflows/version.yml
```diff
@@ -0,0 +1,100 @@
...
+            }, null, 2));
+            execSync('npx prettier --write package.json');
+
+            if (nextVersion !== headVersion) {
+              console.log(`Bumping version from ${version} to ${nextVersion}, current head version is ${headVersion}`);
+              return nextVersion;
+            } else {
+              console.log(`Version is up to date (${headVersion})`);
+              return '';
+            }
```
asturur, line 92: in my opinion we should use the command  npm version patch/minor/major if we want to bump the version instead of writing our own script

ShaMan123: I wasn't aware that was an option. I will look into it

asturur: But i m not sold on incrementing the version automatically yet. I think goin in the release section and push the button to release a new version is quick enough and still allow visibility on what are you publishin

ShaMan123: The question is what will happen if 2 PR incremented to the same tag, we merge one update from master...
Will it go mad?

ShaMan123: The impl I choose was to bump the version as part of the PR, avoiding more commits in the main history log

ShaMan123: I tried using `npm version`, it is the same as using semver with these exceptions:
- npm tags the commit
- npm needs the tree to be clean from changes, meaning that to use it in PRs we must revert the version of package.json, commit and then run `npm version`. That is an overkill.

ShaMan123: > But i m not sold on incrementing the version automatically yet. I think goin in the release section and push the button to release a new version is quick enough and still allow visibility on what are you publishing

The more CI the less headache of maintance and not only, it makes fabric more accessible to devs. Think of early adopters, think of an rc for each PR. In terms of getting devs to contribute that is a very good way IMO since a dev working on the rc can find their way around changes of a single PR and understand how to contribute a fix, and they will have the motivation to do so since they are rc devs. 
In general in will make fabric's development lifecycle a lot shorter and quicker.

ShaMan123: https://stackoverflow.com/questions/12278660/adding-tags-to-a-pull-request kills tagging in PRs for good

ShaMan123: It can be done in a separate commit after the PR is merged but I was under the impression we wanted to avoid that

ShaMan123:
Labeling works!

ShaMan123 (commented):
@asturur this PR is good to go on my end after checking stuff in https://github.com/fabricjs/tagging-ci

What needs to be checked after merging is that the release flow executes correctly

.github/workflows/release.yml
```diff
@@ -0,0 +1,60 @@
...
+              return;
+            }
+            try {
+              const { data: { id } } = await github.rest.repos.getReleaseByTag({
+                owner,
+                repo,
+                tag,
+              });
+              console.log(`Updating release ${tag}`);
+              res = await github.rest.repos.updateRelease({
```
ShaMan123, line 42: This is a part we might want to change.
We can simply return instead of updating the release

ShaMan123:
I tried merging on the test repo to trigger the `release` event with no success

ShaMan123:
Fixed and tested (dry run in the test repo) the publishing actions

ShaMan123:
Ohh I hate it when everything should work and then this random test dies

ShaMan123 (commented):
all the flow had been tested in https://github.com/fabricjs/tagging-ci

ShaMan123 (commented):
updated desc

ShaMan123:
extracted to a standalone

PR #8485, opened 2022-12-03
chore(TS): convert StaticCanvas

asturur:

### DIFF
--- a/src/canvas.class.ts
+++ b/src/canvas.class.ts
@@ -1418,6 +1418,11 @@ import { saveObjectTransform } from './util/misc/objectTransforms';
         // this.discardActiveGroup();
         this.discardActiveObject();
         this.clearContext(this.contextTop);
+        if (this._hasITextHandlers) {
+          this.off('mouse:up', this._mouseUpITextHandler);
+          this._iTextInstances = null;
+          this._hasITextHandlers = false;
+        }
         return this.callSuper('clear');
       },
 
--- a/src/static_canvas.class.ts
+++ b/src/static_canvas.class.ts
@@ -1,4 +1,3 @@
-//@ts-nocheck
 import { config } from './config';
 import { VERSION } from './constants';
 import { createCollectionMixin } from './mixins/collection.mixin';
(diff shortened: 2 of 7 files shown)


src/canvas.class.ts
```diff
@@ -1418,6 +1418,11 @@ import { saveObjectTransform } from './util/misc/objectTransforms';
         // this.discardActiveGroup();
         this.discardActiveObject();
         this.clearContext(this.contextTop);
+        if (this._hasITextHandlers) {
+          this.off('mouse:up', this._mouseUpITextHandler);
+          this._iTextInstances = null;
+          this._hasITextHandlers = false;
+        }
```
asturur, line 1425: StaticCanvas shouldn't have Itext handlers, if someone come with a real use case we can revert this back

asturur:
@ShaMan123  ready to go for me.
See if you see something meaningful that needs to be changed

src/util/misc/dom.ts
```diff
@@ -50,3 +50,9 @@ export const toDataURL = (
   format: ImageFormat,
   quality: number
 ) => canvasEl.toDataURL(`image/${format}`, quality);
+
+export const isHTMLCanvas = (
+  canvas: HTMLCanvasElement | string
+): canvas is HTMLCanvasElement => {
+  return !!canvas && (canvas as HTMLCanvasElement).getContext !== undefined;
```
ShaMan123, line 57: I wonder if instanceof works on node as well nowadays

src/util/types.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { Pattern } from '../pattern.class';
+
+export const isFiller = (filler: TFiller | string): filler is TFiller => {
+  return !!filler && (filler as TFiller).toLive !== undefined;
+};
+
+export const isPattern = (filler: TFiller): filler is Pattern => {
+  return (
+    !!filler &&
+    (filler as Pattern).offsetX !== undefined &&
```
ShaMan123, line 15: no need for this logic, why not instanceof?

src/static_canvas.class.ts
```diff
@@ -9,1855 +8,1834 @@ import { requestAnimFrame } from './util/animate';
...
+  /**
+   * Describe canvas element extension over design
+   * properties are tl,tr,bl,br.
+   * if canvas is not zoomed/panned those points are the four corner of canvas
+   * if canvas is viewportTransformed you those points indicate the extension
+   * of canvas element in plain untrasformed coordinates
+   * The coordinates get updated with @method calcViewportBoundaries.
+   * @memberOf fabric.StaticCanvas.prototype
+   */
+  vptCoords: TCornerPoint;
```
ShaMan123, line 203: `TCornerPoint`: not a great name

src/static_canvas.class.ts
```diff
@@ -9,1855 +8,1834 @@ import { requestAnimFrame } from './util/animate';
...
+        cssValue += 'px';
+        this.hasLostContext = true;
+      }
+
+      if (!backstoreOnly) {
+        this._setCssDimension(prop as keyof TSize, cssValue);
+      }
+    });
+
+    if (this._isCurrentlyDrawing) {
```
ShaMan123, line 526: this part shouldn't be in StaticCanvas as well, it is canvas domain

src/static_canvas.class.ts
```diff
@@ -9,1855 +8,1834 @@ import { requestAnimFrame } from './util/animate';
...
+    }
+  }
+
+  /**
+   * Paint the cached clipPath on the lowerCanvasEl
+   * @param {CanvasRenderingContext2D} ctx Context to render on
+   */
+  drawClipPathOnCanvas(
+    ctx: CanvasRenderingContext2D,
+    clipPath: TCachedFabricObject
```
ShaMan123, line 860: nice and simple!

src/static_canvas.class.ts
```diff
@@ -9,1855 +8,1834 @@ import { requestAnimFrame } from './util/animate';
...
+      .map((prop) => {
+        const fill = this[prop + 'Color'];
+        if (isFiller(fill)) {
+          const shouldTransform = this[prop + 'Vpt'],
+            vpt = this.viewportTransform,
+            object = {
+              width: this.width / (shouldTransform ? vpt[0] : 1),
+              height: this.height / (shouldTransform ? vpt[3] : 1),
+            };
+          return fill.toSVG(object as Rect, {
```
ShaMan123, line 1374: I don't like `object as Rect`
I prefer fixing the types in the Filler

src/static_canvas.class.ts
```diff
@@ -9,1855 +8,1834 @@ import { requestAnimFrame } from './util/animate';
...
+   * @TODO this seems to handle patterns but fail at gradients.
+   * @private
+   */
+  _setSVGBgOverlayColor(markup: string[], property: 'background' | 'overlay') {
+    const filler = this[property + 'Color'];
+    if (!filler) {
+      return;
+    }
+    if (isFiller(filler)) {
+      // @ts-ignore TS is so stubbordn that i can't even check if a property exists.
```
ShaMan123, line 1484: This is really too much, I agree

src/util/types.ts
ShaMan123: I saw the comments

ShaMan123 (approved):
`src/util/types.ts`: 
I saw your comment regarding using instanceof. My suggestion is simple. We create a base class that is exposed only internally for instanceof checks.
So we do something like:
```
export class TextBase extends FabricObject {

}

export class Text extends TextBase {
   ...logic
}

```

```ts
import { TextBase, type Text } from '..';

function isText(what: FabricObject): what is Text {
   return what instanceof TextBase
}
```

A bit clumsy IMO. Is it that bad including it?

src/static_canvas.class.ts
asturur: yeah i agree.

src/static_canvas.class.ts
asturur: yeah so object needs to be a fabricObject, but it can be FabricObject | StaticCanvas to be honest, it  just needs to have TSize or extends the interface of a TSize

asturur:
oh ok, i understand what you mean with the empty classes.
A empty shell class that is the base class ONLY for a specific class, in its own file so it doesn't crete any code import.
I can try that when i convert Canvas

asturur:
> `src/util/types.ts`: I saw your comment regarding using instanceof. My suggestion is simple. We create a base class that is exposed only internally for instanceof checks. So we do something like:
> 
> ```
> export class TextBase extends FabricObject {
> 
> }
> 
> export class Text extends TextBase {
>    ...logic
> }
> ```
> 
> ```ts
> import { TextBase, type Text } from '..';
> 
> function isText(what: FabricObject): what is Text {
>    return what instanceof TextBase
> }
> ```
> 
> A bit clumsy IMO. Is it that bad including it (importing the real class)?

Well you can't code split anymore.
If you work with canvas and rect, you have to import patterns or text just for the purpose of checking the objects you have aren't Text.
 At leat this is my impression.

asturur:
@ShaMan123 Did you reorder imports with some logic?
If prettier doesn't do that don't do it, is that some plugin you have or was willingful?

src/static_canvas.class.ts
ShaMan123: So let's change the signature at `fill#toSVG`

ShaMan123:
Is there a way for prettier to do that?

ShaMan123:
A quick search:
https://github.com/trivago/prettier-plugin-sort-imports
https://medium.com/dooboolab/using-eslint-prettier-and-sort-imports-vscode-extensions-for-formatting-open-source-project-16edf317129d => https://marketplace.visualstudio.com/items?itemName=amatiasq.sort-imports

but vscode does that on my machine... not sure if it is core or a plugin

asturur:
is the reorder part that isn't great.
I don't want meaningless commits in the history, when you need to debug something, reading that list trying to figure out if there is a change or a reorder only.
Or if we rename a file then we will have a bit at time changes of ordering all over the place.

ShaMan123:
> is the reorder part that isn't great.

If we make it part of config that won't happen since it will be reordered before committing by vscode/prettier

ShaMan123:
We should move control rendering logic to canvas as well.
Things like https://github.com/fabricjs/fabric.js/blob/c0b7c1caeb7714a2b7ce4051e5f91c26e6ed3871/src/static_canvas.class.ts#L844

ShaMan123:
I don't understand how the static canvas file passes TS, many fields aren't typed. Did you do some magic config I missed?

ShaMan123:
I see, it is because of CollectionMixin types, I will ty to fix

PR #8489, opened 2022-12-04
chore(TS): polish text

ShaMan123:
## Changes

- extract the measuring context to top level
- types


src/shapes/text.class.ts
```diff
@@ -608,7 +592,7 @@ export class Text<
     charStyle: any,
     forMeasuring?: boolean
   ) {
-    ctx.textBaseline = 'alphabetical';
+    ctx.textBaseline = 'alphabetic';
```
ShaMan123, line 595: thank you TS

ShaMan123, line 595: Thanks to TS.
I am pretty sure this fixes an open bug

ShaMan123 (commented):
Text is very messy
This PR is another step forward 
extracted `getMeasuringContext` to top level

src/shapes/text.class.ts
```diff
@@ -381,25 +391,8 @@ export class Text<
...
-   * @private
-   * @param {String} text Text string
-   * @param {Object} [options] Options object
-   */
-  getMeasuringContext() {
-    if (!fabric._measuringContext) {
-      fabric._measuringContext =
-        (this.canvas && this.canvas.contextCache) ||
-        createCanvasElement().getContext('2d');
+      this.pathSegmentsInfo = getPathSegmentsInfo(path.path);
```
asturur, line 394: i really prefer this to stay in path. It makes easier to swap paths and also the path can use those segment infos for other stuff in the future, since this is info of the path itself.

src/shapes/text.class.ts
asturur: i have seen this too in the past.
I think the baseline work correctly because `alphabetic` is the default and wrong values are ignored.
At least i never noticed anything wrong with baseline.

src/shapes/text.class.ts
```diff
@@ -1141,16 +1131,7 @@ export class Text<
...
-        method,
-        ctx,
-        lineIndex,
-        0,
-        line.join(''),
-        left,
-        top,
-        lineHeight
-      );
+      this._renderChar(method, ctx, lineIndex, 0, line.join(''), left, top);
```
asturur, line 1134: there is some weird reason why lineHeight was there, i ll check it and report back

src/util/path.ts
```diff
@@ -2,13 +2,34 @@
...
+  length: number;
+};
+
+type CurveInfo = PathSegmentInfoCommon & {
+  iterator: (pct: number) => Point;
+  angleFinder: (pct: number) => number;
+  length: number;
+};
+
+export type PathSegmentInfo = {
```
asturur, line 26: we are not doing types with T anymore.
I think is good because it helps understand when is a type and when is class, on top of that i press `T` and i see 99% only types in the autocomplete.

At some point we ll have to sort this out

asturur:
@ShaMan123 i ll try to do some little work now and come back to this PR, maybe i can fix the pathInfo myself and merge.

src/shapes/text.class.ts
ShaMan123: I wasn't sure, saw it wasn't in the signature any longer

src/shapes/text.class.ts
ShaMan123: legit
so we type it on path? or we type text with 
```ts
path: Path & { info: WhatEver }
```

src/util/path.ts
ShaMan123: Correct
I am not for and not entirely against.
The only true thing I dislike are name starting with T, then it is a bit weird: `TTransformEvent`

src/shapes/text.class.ts
ShaMan123: We can change the signature to a destructed param, then adding/removing a key is not biggy in terms of subclassing/overriding

src/util/path.ts
ShaMan123: Then names like `TEvent` might be confusing since we might have wanted to abbreviate `TransformEvent` into `TEvent`

src/shapes/text.class.ts
asturur: i typed in on path, if the build pass i ll merge this PR

asturur (approved):

PR #8490, opened 2022-12-04
fix(scaleObject): keep last sign when over origin

luizzappa:
## Motivation

When we cross the axis, `newPoint.x` or `newPoint.y` = 0, this makes `Math.sign` = 0 and bugs the `flip` property. 

Behavior in a master build: https://codesandbox.io/s/bug-flip-fpzk7b

(We cross the origin, `flipX` = true, but when we cross back it stays true instead of false)

bug 2 flip

## Changes

If `newPoint.x` or `newPoint.y` = 0 we use the `transform.signX` or `transform.signY`  (from what I understand this is the last sign), otherwise it arbitrarily assigns 1 (lack of information to know where it's coming from).

## In Action

With the fix: https://codesandbox.io/s/fix-flip-9m0jsh

fix flip

### DIFF
--- a/src/controls/scale.ts
+++ b/src/controls/scale.ts
@@ -145,14 +145,8 @@ function scaleObject(
     // by center and scaling using one middle control ( default: mr, mt, ml, mb), the mouse movement can easily
     // cross many time the origin point and flip the object. so we need a way to filter out the noise.
     // This ternary here should be ok to filter out X scaling when we want Y only and vice versa.
-    signX =
-      by !== 'y'
-        ? Math.sign(newPoint.x === 0 ? transform.signX || 1 : newPoint.x)
-        : 1;
-    signY =
-      by !== 'x'
-        ? Math.sign(newPoint.y === 0 ? transform.signY || 1 : newPoint.y)
-        : 1;
+    signX = by !== 'y' ? Math.sign(newPoint.x || transform.signX || 1) : 1;
+    signY = by !== 'x' ? Math.sign(newPoint.y || transform.signY || 1) : 1;
     if (!transform.signX) {
       transform.signX = signX;
     }
--- a/src/controls/scale.ts
+++ b/src/controls/scale.ts
@@ -145,8 +145,14 @@ function scaleObject(
     // by center and scaling using one middle control ( default: mr, mt, ml, mb), the mouse movement can easily
     // cross many time the origin point and flip the object. so we need a way to filter out the noise.
     // This ternary here should be ok to filter out X scaling when we want Y only and vice versa.
-    signX = by !== 'y' ? Math.sign(newPoint.x) : 1;
-    signY = by !== 'x' ? Math.sign(newPoint.y) : 1;
+    signX =
+      by !== 'y'
+        ? Math.sign(newPoint.x === 0 ? transform.signX || 1 : newPoint.x)
+        : 1;
+    signY =
+      by !== 'x'
+        ? Math.sign(newPoint.y === 0 ? transform.signY || 1 : newPoint.y)
+        : 1;
     if (!transform.signX) {
       transform.signX = signX;
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(scaleObject): handle when scale is 0 to not bug flip [#8490](https://github.com/fabricjs/fabric.js/pull/8490)
 - chore(): refactor `Object.__uid++` => `uid()` [#8482](https://github.com/fabricjs/fabric.js/pull/8482)
 - chore(TS): migrate object mixins to TS [#8414](https://github.com/fabricjs/fabric.js/pull/8414)
 - chore(TS): migrate filters [#8474](https://github.com/fabricjs/fabric.js/pull/8474)
--- a/test/unit/controls_handlers.js
+++ b/test/unit/controls_handlers.js
@@ -263,25 +263,59 @@
       );
       wrapped(eventData, transform, x, y);
     });
-    ['X', 'Y'].forEach(axis => {
-      const signKey = `sign${axis}`;
-      const scaleKey = `scale${axis}`;
-      const isX = signKey === 'signX';
-      QUnit.test(`scaling ${axis} keeps the same sign when scale = 0`, function (assert) {
-        transform = prepareTransform(transform.target, isX ? 'mr' : 'mb');
-        const fn = fabric.controlsUtils[`scaling${axis}`];
-        const point = new fabric.Point(
-          isX ? 0 : 10,
-          isX ? 10 : 0
-        );
-        transform[signKey] = 1;
-        fn(eventData, transform, point.x, point.y);
+    ['ml', 'mt', 'mr', 'mb'].forEach(controlKey => {
+      const axis = {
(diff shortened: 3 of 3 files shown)


src/controls/scale.ts
```diff
@@ -145,8 +145,14 @@ function scaleObject(
     // by center and scaling using one middle control ( default: mr, mt, ml, mb), the mouse movement can easily
     // cross many time the origin point and flip the object. so we need a way to filter out the noise.
     // This ternary here should be ok to filter out X scaling when we want Y only and vice versa.
-    signX = by !== 'y' ? Math.sign(newPoint.x) : 1;
-    signY = by !== 'x' ? Math.sign(newPoint.y) : 1;
+    signX =
+      by !== 'y'
+        ? Math.sign(newPoint.x === 0 ? transform.signX || 1 : newPoint.x)
```
ShaMan123, line 150:
```suggestion
        ? Math.sign(newPoint.x || transform.signX || 1)
```

ShaMan123 (changes requested):
@luizzappa how are you?
Let me see if I understand correctly. This is an edge case happening only when the scale value is 0 after scaling has been performed. Meaning, once the interaction continues and the scale value is not 0 then the the flip value behaves as expected. Am I right?
So basically this is for smoothness of the flipping value?

Could you add a unit test that mocks this under `test/unit/controls_handlers.js`? And a changelog.

luizzappa:
Hey @ShaMan123 !

> This is an edge case happening only when the scale value is 0 after scaling has been performed

Yeap

> Meaning, once the interaction continues and the scale value is not 0 then the the flip value behaves as expected. Am I right?
> So basically this is for smoothness of the flipping value?

No, once this case occurs it messes up the entire logic of the flip (look at the gif of my first example).

It's kind of hard to express in words why the bug happens because it's a joint behavior of `scaleObject`and `_set` after crossing the orign once. But I'll try...

What "triggers" the `flip` property change starts in this this block here from the `scaleObject` function. When the old sign is different from the current one, it inverts the scale sign (`scale *= -1`):

https://github.com/fabricjs/fabric.js/blob/04971af6d54e24ea1bdcd0b09b7641b984bd5606/src/controls/scale.ts#L185-L189

With this sign inversion, there in the `_set` function it checks if scale < 0, it changes the flip property. (it's kind of a trick, because regardless of whether the change is from `flipX = true to false` or `flipX=false to true`, scale<0 is always set just to perform the flipX inversion within this condition)

https://github.com/fabricjs/fabric.js/blob/04971af6d54e24ea1bdcd0b09b7641b984bd5606/src/shapes/object.class.ts#L1009-L1012

**And why when `newPoint = 0` mess everything up?**

Suppose I already crossed the origin (`flipX=true`):
`newPoint.x = -5` and scaling up until we hit `newPoint.x = 0`

Upon reaching this case, `signX = Math.sign(0) = 0` -> this change `transform.signX = 0` 

I continue scaling, crossing the origin, so:
`newPoint.x = 0.01`

As `transform.signX = 0` (it was set in my last function call when `newPoint.x = 0`), we got into this condition which makes the last signal equal to the current one (set `transform.signX = signX = 1`)

https://github.com/fabricjs/fabric.js/blob/04971af6d54e24ea1bdcd0b09b7641b984bd5606/src/controls/scale.ts#L150-L152

As now the two signs are equal, I no longer enter in the condition that will trigger the `flipX` property inversion:

https://github.com/fabricjs/fabric.js/blob/04971af6d54e24ea1bdcd0b09b7641b984bd5606/src/controls/scale.ts#L185-L189

That is, I do not invert the `scaleX` sign, consequently in the `_set` function I no longer fall under the condition of `value < 0` to change the property `flipX = true`, it remains `false` even after we cross the origin and continue.

ShaMan123:
I looked at the example again, now I see what you mean. Sorry for not paying enough attention.
However I am unable to reproduce on v5.2.4
https://codesandbox.io/s/bug-flip-forked-erc05h?file=/index.html

luizzappa:
> I looked at the example again, now I see what you mean. Sorry for not paying enough attention. However I am unable to reproduce on v5.2.4 https://codesandbox.io/s/bug-flip-forked-erc05h?file=/index.html

Yes, I noticed that too. It only happens in version 6. At version 5, for some reason, `newPoint.x` and `newPoint.y` never take the value 0 (it's always something close, 0.5 or -0.5). Therefore, the bug does not happen.

Tracing the source of this I found it here:

`getLocalPoint` returns slightly different values ​​from one version to another with the same X input:

https://github.com/fabricjs/fabric.js/blob/04971af6d54e24ea1bdcd0b09b7641b984bd5606/src/controls/scale.ts#L136-L142

Diving deeper in `getLocalPoint` function, the difference comes from here:

Version 6:

https://github.com/fabricjs/fabric.js/blob/04971af6d54e24ea1bdcd0b09b7641b984bd5606/src/controls/util.ts#L138

image


Version 5:

https://github.com/fabricjs/fabric.js/blob/b030f44cf59151cc6ebf27c3fdcd5793e268477b/src/controls.actions.js#L234

image

Some implementation difference changed the values, and in version 5 makes `newPoint` never be 0.

Rather than investigating what caused this, I found it safer to treat the `scaleObject` function to deal when `newPointer = 0`.

ShaMan123:
v6 `normalizePoint`
https://github.com/fabricjs/fabric.js/blob/ae6c245ee9052119765fb28e08ffea9ce94c3eae/src/controls/util.ts#L96:L117

v5 `getLocalPoint`
https://github.com/fabricjs/fabric.js/blob/ab377538da61b343b520019bd0c804f67db9e4e5/src/mixins/object_origin.mixin.js#L138:L154

To me they seem identical

ShaMan123:
I want to understand where it comes from, since this bug seems to be only where it surfaced.

luizzappa:
> I want to understand where it comes from, since this bug seems to be only where it surfaced.

I agree with you... 

It's kind of hard to isolate what's going on because there are so many moving parts.

I looked deeper. Despite the difference in values ​​coming from the `getLocalPoint` function, it seems to me that the problem is not in this change from `toLocalPoint` to `normalizePoint`. But the way `left` and `top` are calculated.

> However I am unable to reproduce on v5.2.4
> https://codesandbox.io/s/bug-flip-forked-erc05h?file=/index.html

Ok, version 5 is buggy too... By changing `scaleX = 0.5` makes this bug sometimes appears as well. https://codesandbox.io/s/bug-flip-forked-i08snt

bug flip v5

The original example with version 5 was "lucky" mathematically not finding some value that generates `newPoint = 0`.

ShaMan123 (dismissed):
I refactored the tests
My only last question would be, should the default signX/Y values be 0?
In the one hand it makes sense, on the other the flip code will probably meet an edge case

Tests pass and fail on master which means this issue is fixed

ShaMan123 (approved):
updated from master
@asturur any objections?

asturur:
no no objections here, i wonder what really changed.

asturur:
The calculation seems identical

ShaMan123:
I missed it at first as well.
It is an edge case that occurs when crossing the origin axis while scaling.
Now instead of zeroing out the sign value and consequently f***ing flip values and retains the sign.

asturur (approved):

PR #8497, opened 2022-12-06
fix(TS): EventSpec recognition

ShaMan123:
## Description

This PR fixes the recognition of the EventSpec generic type that was lost by Group.
Emphasizes the for #8394 
Disclaimer: TS has taken a toll from me in this inane effort. Nothing good to say about it today.


src/static_canvas.class.ts
```diff
@@ -257,6 +261,28 @@ export class StaticCanvas extends createCollectionMixin(
...
+
+  // TODO: move to canvas
+  interactive: boolean;
+  upperCanvasEl: HTMLCanvasElement;
+  contextTop: CanvasRenderingContext2D;
+  wrapperEl: HTMLDivElement;
+  cacheCanvasEl: HTMLCanvasElement;
+  protected _isCurrentlyDrawing: boolean;
+  freeDrawingBrush: BaseBrush;
+  _activeObject: FabricObject;
```
ShaMan123, line 275: All of there belong to canvas along with `drawControls` and `_applyCanvasStyle`

ShaMan123 (commented):
I think we should consider making `FabricObjectSVGExportMixin` a mixer function like collection and drop `applyMixins`

I need this merged

asturur:
Well i think there is a balance between everyhing magically typed and type safety.
Right now you lean too much on everything typed and that is tiring

src/static_canvas.class.ts
asturur: So how this passed before withtout the ts-nocheck on top?
How did i get a clea build without those types?

asturur (approved):

ShaMan123:
It was because of the return type of collection mixin that was any

PR #8502, opened 2022-12-10
BREAKING: Change from optional deps to peer deps

asturur:
## Motivation

This is mostly for discussion and see what happens to CI.
If we move canvas and JSDOM from optional to peer dependencies they won't be installed by default compared to now that they could fail to install and just give a warn.

Developers that have a working nodejs fabric setup ( probably a minority ) can just install the dependencies themselves.

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "prettier:check": "prettier --check .",
     "prettier:write": "prettier --write ."
   },
-  "optionalDependencies": {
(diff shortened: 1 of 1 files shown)


ShaMan123:
This makes sense
Should we use peerDependenciesMeta?

ShaMan123:
What happened here? Why did you close?

PR #8518, opened 2022-12-18
patch(TS): canvas patch init logic

ShaMan123:

### DIFF
--- a/src/static_canvas.class.ts
+++ b/src/static_canvas.class.ts
@@ -322,33 +322,19 @@ export class StaticCanvas<
 
   constructor(el: string | HTMLCanvasElement, options = {}) {
     super();
-    this._init(el, options);
-  }
-
-  /**
-   * @private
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
tests pass

src/static_canvas.class.ts
```diff
@@ -322,33 +322,19 @@ export class StaticCanvas<
...
-  /**
-   * @private
-   * @param {HTMLCanvasElement | String} el <canvas> element to initialize instance on
-   * @param {Object} [options] Options object
-   */
-  _init(el: string | HTMLCanvasElement, options = {}) {
     this.renderAndResetBound = this.renderAndReset.bind(this);
     this.requestRenderAllBound = this.requestRenderAll.bind(this);
-    this._initStatic(el, options);
+    this._objects = [];
```
asturur, line 327: I would move this in the collection mixin anyway.

PR #8520, opened 2022-12-18
patch(TS): canvas patch init 2

ShaMan123:
## Description

Refactor canvas init logic, more work can be done in this scope

## Changes

- canvas init
- dumped event binding tests since we bind in constructor from now on
- remove `requestRenderAllBound`, `renderAndResetBound`
- ~~fix `_initRetinaScaling` regression from #8510, BREAKING since now should be called only if `isRetinaScaling` (doesn't check in the method as previously did)~~


src/canvas.class.ts
```diff
@@ -501,20 +501,16 @@ export class Canvas<
    */
   constructor(el: string | HTMLCanvasElement, options = {}) {
     super(el, options);
+    // @ts-ignore
+    this._initEventListeners();
+    this._createCacheCanvas();
```
ShaMan123, line 506: should this belong to `initElements`? Or does it relay on retina scaling being initialized?

src/static_canvas.class.ts
```diff
@@ -278,6 +275,22 @@ export class StaticCanvas<
     kill: (reason?: any) => void;
   };
 
+  constructor(el: string | HTMLCanvasElement, options = {}) {
+    super();
+    this.renderAndReset = this.renderAndReset.bind(this);
+    this.requestRenderAll = this.requestRenderAll.bind(this);
```
ShaMan123, line 281: renamed these

ShaMan123 (commented):
READY

ShaMan123 (commented):
updated from master

ShaMan123:
now things are broken

ShaMan123:
@asturur this is high priority 
I have no idea the full extent of the regression but to fix it I had to init size on upper canvas, something I didn't locate be #8510 
I then tried to use `setDimensions` instead of `initSize` with no luck.
please review

src/canvas/canvas.class.ts
```diff
@@ -497,34 +497,16 @@ export class SelectableCanvas<
...
     this._createCacheCanvas();
   }
 
-  /**
-   * @private
-   */
-  _initRetinaScaling() {
+  protected _initRetinaScaling() {
     super._initRetinaScaling();
     this.__initRetinaScaling(this.upperCanvasEl, this.contextTop);
```
ShaMan123, line 511: the regression was here
I changed this method to run w/o a check if is retina scaling and call it only if so

src/canvas/canvas.class.ts
```diff
@@ -1252,30 +1234,15 @@ export class SelectableCanvas<
...
   }
 
-  /**
-   * @private
-   */
-  _createCacheCanvas() {
+  protected _createCacheCanvas() {
     this.cacheCanvasEl = this._createCanvasElement();
-    this.cacheCanvasEl.setAttribute('width', `${this.width}`);
-    this.cacheCanvasEl.setAttribute('height', `${this.height}`);
```
ShaMan123, line 1266: size will be set by `setDimensions` on init

src/canvas/canvas_events.ts
```diff
@@ -215,47 +224,6 @@ export class Canvas extends SelectableCanvas {
     );
   }
 
-  /**
-   * @private
-   */
-  private _bindEvents() {
```
ShaMan123, line 221: no need to have this now that it is impossible to init canvas twice

src/canvas/static_canvas.class.ts
```diff
@@ -271,10 +272,7 @@ export class StaticCanvas<
    * in order to return it to original state on dispose
    * @type string
    */
-  _originalCanvasStyle?: string;
-
-  renderAndResetBound: () => void;
-  requestRenderAllBound: () => void;
```
ShaMan123, line 277: dropped

ShaMan123 (commented):
ready

src/canvas/static_canvas.class.ts
```diff
@@ -283,7 +281,24 @@ export class StaticCanvas<
 
   constructor(el: string | HTMLCanvasElement, options = {}) {
     super();
-    this._init(el, options);
+    this.renderAndReset = this.renderAndReset.bind(this);
+    this.requestRenderAll = this.requestRenderAll.bind(this);
```
ShaMan123, line 285: these were renamed,
see description

ShaMan123 (commented):
mergable

src/canvas/static_canvas.class.ts
asturur: Yes yes i have read everything i was in the middle of things and then i had to go to bed.
Probably going to let them both names.

asturur:
Bind on constructor is a great normal thing we should have done as soon as we started moving canvas, i really didn't thin of it.
I don't want to have requestRenderAll and renderAndReset naturally bound for those reasons:
- are not meant to be callbacks or event handlers
- an unexperienced dev would see them working and not understand those need to be bound
- you could override them or swap them on the prototype and then assign back to the instance and loose the binding
So i delete the convenience bound functions ( with the bound suffix ) but i m not binding this.
If there is a complain that the old one were useful we can add them back

The reorganization is great, i don't like the idea of adding a parameter to setDimensions to have it skip rendering just because we need it for initialization, since we managed to remove eventsBound: boolean, let's not add a similar one.
I separated setDimenions in setDimensions + protected setDimensionsImplementation, and teh constructor can call the latter.

ShaMan123 (commented):
I absolutely agree

ShaMan123:
@asturur why e1ffe0cac797d20c8c76ac09bc8d77d104d55fb2?

asturur:
it is how it was before and i wouldn't mind if it doesn't save half kb on the minified build.

asturur (approved):

PR #8522, opened 2022-12-19
chore(): use context in static constructors

ShaMan123:
## Motivation

Using `this` in static constructors makes it very easy to subclass because the subclass won't have to impl its own static constructors. And it relieves fabric from redundancies.
The dev will need to register the class not matter what.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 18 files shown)


ShaMan123 (commented):
Looking at `fromElement` in light of this PR I don't understand why it doesn't utilize `fromObject`.
Meaning, each class should parse attributes and pass them to `fromObject` and that is it.

asturur:
This is a good change. Some cruft removed

asturur (approved):

PR #8526, opened 2022-12-20
feat(): Add Textbox overflowBreakWord option

wxul:
## Motivation

Textbox option `splitByGrapheme` does not meet my needs

## Description

The Textbox adds an `overflowBreakWord` option to achieve a result similar to the css property `overflow-wrap: break-word`. Unlike the default, the width can be adjusted, and unlike `splitByGrapheme: true`, the word will be kept as complete as possible.

[MDN: overflow-wrap](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap)

#### Textbox default

https://user-images.githubusercontent.com/6262240/208657448-7d1afca7-65db-4515-a6f5-8748598a0e81.mp4

#### Textbox with `splitByGrapheme: true`

https://user-images.githubusercontent.com/6262240/208657487-fa1bc625-7fed-4758-b36e-084bdbeb6771.mp4

#### Textbox with `overflowBreakWord: true`

https://user-images.githubusercontent.com/6262240/208657522-11d627df-d207-4a94-abff-fac6526c223c.mp4

### DIFF
--- a/src/shapes/textbox.class.ts
+++ b/src/shapes/textbox.class.ts
@@ -45,6 +45,14 @@ export class Textbox extends IText {
   splitByGrapheme: boolean;
 
   /**
+   * Implementation of css property overflow-wrap.
+   * unlike splitByGrapheme, words will have reasonable newlines
+   * Will block splitByGrapheme option!
+   * @type Boolean
+   */
+  overflowBreakWord: boolean;
+
+  /**
    * Unlike superclass's version of this function, Textbox does not update
    * its width.
    * @private
@@ -246,8 +254,11 @@ export class Textbox extends IText {
   _wrapText(lines: Array<any>, desiredWidth: number): Array<any> {
     const wrapped = [];
     this.isWrapping = true;
+    const wrapFunction = this.overflowBreakWord
+      ? this._wrapLineOfWordBreak
+      : this._wrapLine;
     for (let i = 0; i < lines.length; i++) {
-      wrapped.push(...this._wrapLine(lines[i], i, desiredWidth));
+      wrapped.push(...wrapFunction.call(this, lines[i], i, desiredWidth));
     }
     this.isWrapping = false;
     return wrapped;
@@ -291,6 +302,83 @@ export class Textbox extends IText {
    */
   wordSplit(value: string): string[] {
     return value.split(this._wordJoiners);
+  }
+
+  /**
+   * Wraps a line of text using the width of the Textbox and a context. For overflowBreakWord option
+   * @param {Array} line The grapheme array that represent the line
+   * @param {Number} lineIndex
+   * @param {Number} desiredWidth width you want to wrap the line to
+   * @param {Number} reservedSpace space to remove from wrapping for custom functionalities
+   * @returns {Array} Array of line(s) into which the given text is wrapped
+   * to.
+   */
+  _wrapLineOfWordBreak(
+    _line,
+    lineIndex: number,
+    desiredWidth: number,
+    reservedSpace = 0
+  ): Array<any> {
+    desiredWidth -= reservedSpace;
+    const graphemeLines = [];
+    let text = _line || '',
+      i = 0,
+      j = 0,
+      offset = 0,
+      temp = '',
+      width = 0,
+      prevGrapheme = '',
+      largestLetterWidth = 0;
+
+    if (!text.length) {
+      graphemeLines.push([]);
+    }
+    while (text.length > 0) {
+      i = 0;
+      // Get the maximum string at a fixed width
+      width = 0;
+      prevGrapheme = '';
+      for (let k = 0, len = text.length; k < len; k++) {
+        const box = this._getGraphemeBox(
+          text[k],
+          lineIndex,
+          offset + k,
+          prevGrapheme,
+          true
+        );
+        width += box.kernedWidth;
+        largestLetterWidth = Math.max(largestLetterWidth, box.kernedWidth);
+        if (width > Math.max(largestLetterWidth, desiredWidth)) {
+          i = k;
+          break;
+        }
+        prevGrapheme = text[k];
+      }
+
+      // Prevent some edge case, may lead to an infinite loop
+      if (i === 0) {
+        graphemeLines.push(this.graphemeSplit(text));
+        break;
+      }
+
+      temp = text.substring(0, i);
+
+      if (text.length !== i) {
+        // Get last space to split words
+        const lastSpace = temp.lastIndexOf(' ');
+        j = lastSpace >= 0 ? lastSpace + 1 : 0;
+      }
+      const currentLine = temp.substring(0, j || temp.length);
+      offset += currentLine.length;
+
+      graphemeLines.push(this.graphemeSplit(currentLine));
+      text = text.substring(currentLine.length, text.length);
+      j = 0;
+    }
+    if (largestLetterWidth > this.dynamicMinWidth) {
+      this.dynamicMinWidth = largestLetterWidth;
+    }
+    return graphemeLines;
   }
 
   /**
@@ -405,7 +493,7 @@ export class Textbox extends IText {
    * @return Number
    */
   missingNewlineOffset(lineIndex) {
-    if (this.splitByGrapheme) {
+    if (this.splitByGrapheme || this.overflowBreakWord) {
       return this.isEndOfWrapping(lineIndex) ? 1 : 0;
     }
     return 1;
@@ -456,7 +544,9 @@ export class Textbox extends IText {
    */
   toObject(propertiesToInclude: Array<any>): object {
     return super.toObject(
-      ['minWidth', 'splitByGrapheme'].concat(propertiesToInclude)
+      ['minWidth', 'splitByGrapheme', 'overflowBreakWord'].concat(
+        propertiesToInclude
+      )
     );
   }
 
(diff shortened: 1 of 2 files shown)


ShaMan123:
Hi @wxul 
This PR looks really good.
We will review it in the near future.
It is a requested feature, nice!

src/shapes/textbox.class.ts
```diff
@@ -44,6 +44,14 @@ export class Textbox extends IText {
    */
   splitByGrapheme: boolean;
 
+  /**
+   * Implementation of css property overflow-wrap.
+   * unlike splitByGrapheme, words will have reasonable newlines
+   * Will block splitByGrapheme option!
+   * @type Boolean
+   */
+  overflowBreakWord: boolean;
```
ShaMan123, line 53: IMO we should introduce a `breakLineStrategy` prop or something similar instead of a flag and perhaps make `splitByGrapheme` an option

src/shapes/textbox.class.ts
```diff
@@ -293,6 +304,83 @@ export class Textbox extends IText {
...
+          text[k],
+          lineIndex,
+          offset + k,
+          prevGrapheme,
+          true
+        );
+        width += box.kernedWidth;
+        largestLetterWidth = Math.max(largestLetterWidth, box.kernedWidth);
+        if (width > Math.max(largestLetterWidth, desiredWidth)) {
+          i = k;
```
ShaMan123, line 352: Doesn't this make the last letter spill over the bbox?
Should it be `i = Math.max(k - 1, 0)` ?

src/shapes/textbox.class.ts
```diff
@@ -293,6 +304,83 @@ export class Textbox extends IText {
...
+        graphemeLines.push(this.graphemeSplit(text));
+        break;
+      }
+
+      temp = text.substring(0, i);
+
+      if (text.length !== i) {
+        // Get last space to split words
+        const lastSpace = temp.lastIndexOf(' ');
+        j = lastSpace >= 0 ? lastSpace + 1 : 0;
```
ShaMan123, line 369: nice!
this is the actual word break logic if I understand correctly

src/shapes/textbox.class.ts
```diff
@@ -293,6 +304,83 @@ export class Textbox extends IText {
...
+      if (i === 0) {
+        graphemeLines.push(this.graphemeSplit(text));
+        break;
+      }
+
+      temp = text.substring(0, i);
+
+      if (text.length !== i) {
+        // Get last space to split words
+        const lastSpace = temp.lastIndexOf(' ');
```
ShaMan123, line 368: we should use a regex (`this._wordJoiners`) here in case of tabs and stuff I think.
Thoughts?

src/shapes/textbox.class.ts
```diff
@@ -293,6 +304,83 @@ export class Textbox extends IText {
...
+  /**
+   * Wraps a line of text using the width of the Textbox and a context. For overflowBreakWord option
+   * @param {Array} line The grapheme array that represent the line
+   * @param {Number} lineIndex
+   * @param {Number} desiredWidth width you want to wrap the line to
+   * @param {Number} reservedSpace space to remove from wrapping for custom functionalities
+   * @returns {Array} Array of line(s) into which the given text is wrapped
+   * to.
+   */
+  _wrapLineOfWordBreak(
```
ShaMan123, line 316: I think it is best to add in the comment an explanation of how the method works
Something like:
This method splits text into lines by looking for the last space in a measured piece of text that fits the bounds. 
If a space is found, the line breaks before the last word in the sequence.
If not, it splits the last word and breaks the line.

ShaMan123 (commented):
I will test it and comment in case I have more findings.
Good job!
Could you add a line to the CHANGELOG?

src/shapes/textbox.class.ts
wxul: I agree, but how to deal with the previous versions of `splitByGrapheme` ?

src/shapes/textbox.class.ts
ShaMan123: correct

src/shapes/textbox.class.ts
ShaMan123: We can mark it as `deprecated` and introduce its predecessor and have logic supporting falling back to `splitByGrapheme`.
Or we break.
@asturur ?

src/shapes/textbox.class.ts
wxul: Very helpful !

src/shapes/textbox.class.ts
wxul: Since `lastIndexOf` only supports string, I changed the implementation

ShaMan123: looks fine
There are other options, not sure if it makes any difference, things such as `String#matchAll` iterator or a more complex regex:
last occurrence of `\s`: https://regex101.com/r/DVP2Ix/1
all up to last occurrence of `\s`: https://regex101.com/r/3cwB7r/1

ShaMan123: We could then use `this._wordJoiners.source` to compose a new one.
But again, no idea if this impacts perf, and if it does it is probably extremely minor

ShaMan123 (changes requested):
I have deprecated `splitByGrapheme` in favor of `textOverflow`. Is the naming good enough?

I have tested live and have a few concerns:
- should initial size be equal to the largest word if `width` is not set? This question might be a non issue due to #8470 . Current behavior is like splitByGrapheme, width is set to the largest char.
- lines with whitespace only should be dropped

Text value: `fabric.js sandbox sfdg sdfg sdfg sdfgsssa dsfgert ergertq wtrh srth`

https://user-images.githubusercontent.com/34343793/208711888-93ff028b-8f5a-44fd-8a52-1f8d7b021c78.mp4


- After this is merged we need to visit #8470 and see how it works and add a test perhaps.

src/shapes/textbox.class.ts
wxul: > looks fine There are other options, not sure if it makes any difference, things such as `String#matchAll` iterator or a more complex regex: last occurrence of `\s`: https://regex101.com/r/DVP2Ix/1 all up to last occurrence of `\s`: https://regex101.com/r/3cwB7r/1

I wrote some simple tests, using the regular in the example takes far more time than a simple loop

ShaMan123: Nice

ShaMan123:
After looking into cursor position I see there is an unrelated bug/wrong behavior of the cursor with line breaking.
It happens with `textOverflow: 'anywhere'` as well and predates this PR.
Looking at HTMLTextarea I see that when a line breaks all preceding white space is appended to that line without affecting the bounding box, while the next line starts with a non white space char. The cursor runs over the white space length - 1, then is drops to the next line start while remaining bound to the bounding box (while hitting the arrow the cursor remains at the edge of the textarea until white space has finished).
Cursor location/position correlates to text length in fabric. I think in order to align with HTMLTextarea we should address that. And/or replace white space with `''` for cursor logic.
However, that should/could be in a separate PR.
I will try this out and see if it is too complicated.
Thoughts?

src/shapes/textbox.class.ts
```diff
@@ -101,7 +110,7 @@ export class Textbox extends IText {
         charCount++;
         realLineCount++;
       } else if (
-        !this.splitByGrapheme &&
+        this.resolveTextOverflowStrategy() !== 'anywhere' &&
```
ShaMan123, line 113: We need to test this
Styles in general

src/shapes/textbox.class.ts
```diff
@@ -93,7 +110,7 @@ export class Textbox extends IText {
         charCount++;
         realLineCount++;
       } else if (
-        !this.splitByGrapheme &&
+        !this.resolveTextOverflowStrategy() &&
```
ShaMan123, line 113: guessing this...
no tests and not enough context to understand this.
Why doesn't it use the split lines?

ShaMan123 (commented):
@asturur @melchiar I need your help here to determine 2 issues:
1. CharSpacing + line breaking not on the first word still keeps some spacing 

https://user-images.githubusercontent.com/34343793/208917314-677fbe44-10ef-458f-8ae0-43b688ad0dea.mp4

2. Style map generation, see comment

I am done here.
@wxul Thanks!

---

TODO: cursor is off with text overflow. It should break before the last space of a line (if exists) to the beginning of the next. This doesn't happen. Should be addressed in a follow.

src/shapes/textbox.class.ts
ShaMan123: introduced `resolveTextOverflowStrategy`

asturur:
Hi @wxul  this PR looks good to me, i m focused those days in finishing the TS migration and i would like to look at this shortly after.
Please don't feel ignored, we will get here.

asturur:
(this feature was requested in the past too, we had an implementation but then we decided to do no exagerate with options)

ShaMan123 (approved):
updated from master

melchiar:
Code looks good to me, though I'm wondering if maybe the css property `hyphens: auto` would be a better behavior to replicate than `overflow-wrap: break-word`, since inserting a hyphen between a word break is more common typographically.

https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens

umtdemr:
I noticed an issue with this PR. If we keep pressing space, the lines are getting broken.

ShaMan123:
@umtdemr can you provide a video?

umtdemr:
https://user-images.githubusercontent.com/61540809/221537573-0469bd3c-e703-47a8-81ed-b33f980f44f2.mov

ShaMan123:
nice!
we should adapt the code to break after a space and keep space length -1 in the new line

PR #8527, opened 2022-12-20
feat(): ContextualPoint

ShaMan123:
## Motivation

We discussed this here https://github.com/fabricjs/fabric.js/pull/8510#discussion_r1051592521

The motivation is to make geometry easier, without having to think all the time about planes and matrices

## Description

If we decide to move forward with this planeChange can be reduced.

we can make it part of Point and not a subclass.

### DIFF
--- a/src/ContextualPoint.ts
+++ b/src/ContextualPoint.ts
@@ -1,0 +1,101 @@
+import { iMatrix } from './constants';
+import { Point } from './point.class';
+import { FabricObject } from './shapes/Object/Object';
+import { StaticCanvas } from './static_canvas.class';
+import { TMat2D } from './typedefs';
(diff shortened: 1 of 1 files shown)


jiayihu:
Good design, but make it also non-assignable to a Point of different plane and make it easy to understand with a tag property, like "canvas-element-plane" because a matrix is hard to read. E.g.

```
const pointInCanvasElementPlane = new fabric.ContextualPoint(0, 0, "canvas-element-plane")
const pointInParentPlane = new fabric.ContextualPoint(0, 0, "parent-plane")

const getRelativeX = (point: fabric.ContextualPoint<"parent-plane">): number => point.x

getRelativeX(pointInCanvasElementPlane) // Error
getRelativeX(pointInParentPlane) // Okay
```

ShaMan123:
I am abandoning this

PR #8534, opened 2022-12-22
fix(): Draggable Text Migration Regression

ShaMan123:
## Description

Draggable text was broken in #8013 (db8739a8f72ea77645424bbf6ab678411477be2f) and #8421 (due to `ts-nocheck`)

## Gist

I have noticed that undoing a text drop isn't supported natively and I wonder why. Would love your feedback.
The textarea value is changed programmatically on drop. Is there a way to do that and have the change enter the native history stack?

## In Action

My screen recorder doesn't record the drag image... so it is pointless


src/mixins/itext_behavior.mixin.ts
```diff
@@ -547,24 +548,24 @@ export abstract class ITextBehaviorMixin<
...
     const retinaScaling = this.getCanvasRetinaScaling();
     const bbox = this.getBoundingRect(true);
     const correction = pos.subtract(new Point(bbox.left, bbox.top));
-    const offset = correction.add(diff).scalarMultiply(retinaScaling);
+    const vpt = this.canvas.viewportTransform;
+    const offset = correction.add(diff).transform(vpt, true);
     //  prepare instance for drag image snapshot by making all non selected text invisible
     const bgc = this.backgroundColor;
-    const styles = object.clone(this.styles, true);
+    const styles = clone(this.styles, true);
```
ShaMan123, line 568: bug from #8421

src/mixins/itext_behavior.mixin.ts
```diff
@@ -679,7 +686,8 @@ export abstract class ITextBehaviorMixin<
    * @param {object} options
    * @param {DragEvent} options.e
    */
-  dragOverHandler({ e }: TEvent<DragEvent>) {
+  dragOverHandler(ev: TEvent<DragEvent>) {
```
ShaMan123, line 689: same

src/mixins/itext_behavior.mixin.ts
```diff
@@ -762,21 +770,19 @@ export abstract class ITextBehaviorMixin<
    * in order to change the drop value or to customize styling respectively, by listening to the `drop:before` event
    * https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations#performing_a_drop
    * @private
-   * @param {object} options
-   * @param {DragEvent} options.e
    */
-  dropHandler({ e }: TEvent<DragEvent>) {
+  dropHandler(ev: TEvent<DragEvent>) {
```
ShaMan123, line 774: same

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -106,6 +106,7 @@ export abstract class ITextClickBehaviorMixin<
     if (
       !this.canvas ||
       !this.editable ||
+      this.__isDragging ||
```
ShaMan123, line 109: bug from #8013

src/mixins/itext_behavior.mixin.ts
```diff
@@ -573,25 +574,31 @@ export abstract class ITextBehaviorMixin<
...
+      const size = new Point(dragImage.width, dragImage.height)
+        .scalarMultiply(retinaScaling)
+        .transform(vpt, true);
+      dragImageCanvas.width = size.x;
+      dragImageCanvas.height = size.y;
+      const ctx = dragImageCanvas.getContext('2d');
       ctx.scale(1 / retinaScaling, 1 / retinaScaling);
+      const [a, b, c, d] = vpt;
+      const origin = new Point().transform(vpt, true);
+      ctx.transform(a, b, c, d, -origin.x, -origin.y);
```
ShaMan123, line 593: fix drag image vpt

ShaMan123 (commented):
READY!
9ec644015 can be reverted if you think `isIdentityMatrix` should remain internal

ShaMan123 (commented):
added test
now it is possible to see there is a blur in the drag image.
no idea how to handle it

asturur:
I didn't see this PR here.
What was the actual bug and how did we make it?
I see there was a bad if with a missing condition in a conflict resolution, but what did it cause?

src/mixins/itext_behavior.mixin.ts
```diff
@@ -531,67 +532,70 @@ export abstract class ITextBehaviorMixin<
...
+      const dragImageCanvas = createCanvasElement();
+      const size = new Point(dragImage.width, dragImage.height)
+        .scalarDivide(retinaScaling)
+        .transform(vpt, true);
+      dragImageCanvas.width = size.x;
+      dragImageCanvas.height = size.y;
+      const ctx = dragImageCanvas.getContext('2d');
       ctx.scale(1 / retinaScaling, 1 / retinaScaling);
+      const [a, b, c, d] = vpt;
+      const origin = new Point().transform(vpt, true);
```
asturur, line 589: isn't this exactly vpt[4] and vpt[5]? what is this line doing?

src/mixins/itext_behavior.mixin.ts
ShaMan123: this is a fix to a bug unrelated to the migration

ShaMan123:
I commented all fixes and changes and referenced where they originated from.

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: merge conflict

src/mixins/itext_behavior.mixin.ts
ShaMan123: happened due to use of the transformer and no test to discover it

src/mixins/itext_behavior.mixin.ts
ShaMan123: happened because no test

src/mixins/itext_behavior.mixin.ts
ShaMan123: happened because no test

src/mixins/itext_behavior.mixin.ts
ShaMan123: you might be right.
I will check.

ShaMan123: the logic behind it is the same as send 0,0 to the transformed plane

ShaMan123:
@asturur I have another fix I am working on

src/mixins/itext_behavior.mixin.ts
```diff
@@ -531,67 +532,69 @@ export abstract class ITextBehaviorMixin<
   /**
    * Override to customize the drag image
    * https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setDragImage
-   * @param {DragEvent} e
-   * @param {object} data
-   * @param {number} data.selectionStart
-   * @param {number} data.selectionEnd
-   * @param {string} data.text
-   * @param {string} data.value selected text
```
asturur, line 539: was this wrong? which is the correct documentation?

ShaMan123: it is wrong, only selectionStart/End remains

ShaMan123: scratch that.
those are the only used by the method.

src/canvas/canvas_events.ts
```diff
@@ -766,25 +769,22 @@ export class Canvas extends SelectableCanvas {
    * @param {Event} e Event object fired on mouseup
    */
   __onMouseUp(e: TPointerEvent) {
-    const transform = this._currentTransform,
-      groupSelector = this._groupSelector,
-      isClick =
-        !groupSelector || (groupSelector.left === 0 && groupSelector.top === 0);
```
asturur, line 772: i m ok changing the way we detect if we are in a click or drag state for the mouse on canvas, but it should continue to be a const isClick as before. There is no reason to generate a change in the history for the other lines.

src/canvas/canvas_events.ts
```diff
@@ -856,14 +856,20 @@ export class Canvas extends SelectableCanvas {
...
     if (shouldRender) {
       this.requestRenderAll();
-    } else if (!isClick) {
+    } else if (
+      !this._isClick &&
+      !(
+        isInteractiveTextObject(this._activeObject) &&
+        this._activeObject.isEditing
+      )
+    ) {
```
asturur, line 872: this wasn't necessary before, what kind of bug is causing?
Is because the renderTop is not necessary or because the renderTop was causing a visual artifact?

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -191,7 +192,10 @@ export abstract class ITextClickBehaviorMixin<
         return;
       }
     }
-
+    if (this.__isDragging && options.isClick) {
+      // false positive drag event, is actually a click
+      this.setCursorByClick(options.e);
+    }
```
asturur, line 198: since `this.setCursorByClick` wasn't here before, who was taking care of it?

asturur:
> READY! [9ec6440](https://github.com/fabricjs/fabric.js/commit/9ec644015bb2128c5d397a9b9c29bd28ffba48e5) can be reverted if you think `isIdentityMatrix` should remain internal

isIdentityMatrix is definetely internal, but we don't have a way to test internal utils right now, so we can't make that distinction yet.

asturur:
@ShaMan123 do we have a way to turn the drag feature off? this was a request i had initially and i don't think we got there ever.
Drag text support can get in your way pretty much and it should be possible to turn it off.
Otherwise hastly reselecting text that is already selected will cause a drag operation and that may be a forced ux that we are  forcing on apps.

src/canvas/canvas_events.ts
ShaMan123: 8b7971f

src/canvas/canvas_events.ts
ShaMan123: The bug was in `isClick`, it was wrong so logic was executing correctly...
What happens is top context renders and clears text selection.
I thought instead of this check to render text selection for `renderTop`, making it more robust
Thoughts?

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: I believe this is why you altered the logic in the original PR stating you changed something (forgot what)
Until now I missed what it did and why you (probably) did it.
Trying to click a selection does the following:
- mouse down before: set `__isDragging` to true
- mouse down: if `__isDragging === false` set cursor by click => **skipped**
- drag events fire or **not**, we are interested in the case they do not.
- mouse up: now we know if the interaction was a click or not so we can set cursor by click if `__isDragging` to compensate for what didn't happen in mouse down

ShaMan123:
> Otherwise hastly reselecting

I am not sure this can happen.

But regardless there is a way.
Return false from `onDragStart` will do the job.
Fine tuning drag and drop by controlling `canDrop` return value.

```ts
IText.prototype.onDragStart = () => false

```

Is that enough? Or you want a flag?

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: > I believe this is why you altered the logic in the original PR stating you changed something (forgot what)
Until now I missed what it did and why you (probably) did it.

Sorry I meant to link to it but now I think is simply unrelated

ShaMan123: I would add tests but it is quite a nightmare I believe since I need a sequence of drag events...
Am I being lazy? Might be.

ShaMan123: Maybe give me a few more hours and I try to write an extensive module of visual drag effects for the following cases:
- drag start
- drag over self
- drag over other target
- drag enter/leave canvas
- click

Doesn't sound awful actually

asturur:
is enough to start, we don't need to add more and more, if that is work it can be documented.
If we need a flag we can use it there inside that function.

ShaMan123:
> Return false from onDragStart will do the job.

tested this now. Something is not good enough. will push a fix

src/mixins/itext_behavior.mixin.ts
```diff
@@ -246,7 +247,7 @@ export abstract class ITextBehaviorMixin<
   /**
    * Aborts cursor animation, clears all timeouts and clear textarea context if necessary
    */
-  abortCursorAnimation() {
+  abortCursorAnimation(skipClearing = false) {
```
ShaMan123, line 250: verify this change is needed after #8543

ShaMan123 (commented):
fixed overlapping drag/drop target rendering effects and cursor behavior
after #8583 #8543 are merged I will adapt this PR and think about testing

asturur:
> I am not sure this can happen.

i mostly sure it happens it happens to me all the time and is the reason when we introduced this i siad it should be optional.
_searchPossibleTargets needs to be refactored, out of this PR, to avoid using this.targets and just return targets.
We need to be able to search targets without mutating global state.

src/mixins/itext_behavior.mixin.ts
ShaMan123: no needed anymore since #8583 fixed a lot of edge cases

ShaMan123 (commented):
Well this is good to go
All known issues seem to be solved
I would like to add tests...

ShaMan123 (commented):
Added tests

ShaMan123 (commented):
fixed another small bug, focusing hiddenTextarea after drag end since it got blurred during the drag

@asturur I have done a follow up to this PR #8598 , please move to that once this is merged

ShaMan123:
OR close this PR and merge that directly
I prefer changes to be made there instead of porting.

asturur:
For the blur in the dragged image i can fix it, i think the principle is the same as retina scaling

asturur:
Doing this today and the next, now i have guests i ll be off some hours

src/shapes/itext.class.ts
```diff
@@ -283,6 +283,18 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
     this.renderCursorOrSelection();
   }
 
+  /**
+   * @override block cursor/selection logic while rendering the exported canvas
+   * @todo this workaround should be replaced with a more robust solution
```
asturur, line 288: i made years ago a renderExport function for those cases.
I would run the usual render code but without side effects and without runing the current caches and with less blur.
I should officialize it, it covers also those cases.

asturur: This is fine anyway

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -192,6 +195,8 @@ export abstract class ITextClickBehaviorMixin<
       return;
     }
 
+    shouldSetCursor && this.setCursorByClick(options.e);
```
asturur, line 198: What is unclear to me is why this shouldn't have triggered on mouse down.
Not a big deal, this is not relevant to merging.
Can you try to explain anyway this is here? Setting cursor was usually mouse down tas

asturur: I see the mouseDown has an extra `this.__isDragging ||` how do we trigger dragging before mouse down event?

asturur:
> I have noticed that undoing a text drop isn't supported natively and I wonder why. Would love your feedback. The textarea value is changed programmatically on drop. Is there a way to do that and have the change enter the native history stack?

I do think yes.
Probably we should attempt at forwarding the drop event with the same data to the native textarea.
So we enter edit, we capture the event on canvas, but we send it to the textarea, we check if it generates an input event to automatically adjust text.

But please let this idea at rest now we have other stuff to do

asturur (dismissed):

src/shapes/itext.class.ts
ShaMan123: I am for. An object should have a method for that.
BTW #8298 offers a solution as well.

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: `_mouseDownHandlerBefore` sets `this.__isDragging` to keep selection, it is clearer in #8598 https://github.com/fabricjs/fabric.js/blob/27dcd9e011f7f4fd235e8a7490c2e1775f68fd40/src/mixins/DraggableTextDelegate.ts#L12-L21 )

ShaMan123:
> So we enter edit, we capture the event on canvas, but we send it to the textarea, we check if it generates an input event to automatically adjust text.
> 
> But please let this idea at rest now we have other stuff to do

I don't think the DOM will let use retarget an event.
Letting it rest. History isn't hard to impl by yourself, dragging is. So I am OK with it.
In my projects I will do so.

asturur:
I m sure you can clone events and forward them where you want, i did so in the past to have a layer that was capturing some events but setting others to a sibling below

asturur (approved):

src/mixins/itext_click_behavior.mixin.ts
asturur: ok. so mosedown:before shoud allow for the developer to interact before fabric does anything.
That was the point of that event.
Let's try to not break that contract.

ShaMan123: So you suggest moving logic to mouse down?
Sounds correct

ShaMan123: 2aadf5d53

PR #8537, opened 2022-12-27
refactor(): `getRegularPolygonPath` => `Polygon.getRegularPolygonPoints`, `Polygon.createRegularPolygon`

ShaMan123:
## Motivation

I will need this bit soon and it seems wasted as a path util and more relevant as part of polygon.

## Description

I have refactored `getRegularPolygonPath` into `Polygon.getRegularPolygonPoints`, returning points instead of path commands. Anyone needing path commands can easily iterate over the points and create it.

### DIFF
--- a/src/shapes/polygon.class.ts
+++ b/src/shapes/polygon.class.ts
@@ -1,10 +1,46 @@
 import { fabric } from '../../HEADER';
+import { halfPI } from '../constants';
+import { Point } from '../point.class';
 import { TClassProperties } from '../typedefs';
+import { cos } from '../util/misc/cos';
+import { sin } from '../util/misc/sin';
 import { Polyline, polylineDefaultValues } from './polyline.class';
 
 export class Polygon extends Polyline {
   protected isOpen() {
     return false;
+  }
+
+  static getRegularPolygonPoints({
+    numVertexes,
+    radius,
+  }: {
+    numVertexes: number;
+    radius: number;
+  }) {
+    const interiorAngle = (Math.PI * 2) / numVertexes;
+    // rotate the polygon by 1/2 the interior angle so that the polygon always has a flat side on the bottom
+    const rotationAdjustment =
+      -halfPI + (numVertexes % 2 === 0 ? interiorAngle / 2 : 0);
+    return new Array(numVertexes).fill(0).map((_, i) => {
+      const rad = i * interiorAngle + rotationAdjustment;
+      return new Point(cos(rad), sin(rad)).scalarMultiply(radius);
+    });
+  }
+
+  static createRegularPolygon({
+    numVertexes,
+    radius,
+    ...options
+  }: {
+    numVertexes: number;
+    radius: number;
+  } & Record<string, unknown>) {
+    return new this(this.getRegularPolygonPoints({ numVertexes, radius }), {
+      left: 0,
+      top: 0,
+      ...options,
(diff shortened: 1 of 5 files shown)


ShaMan123 (commented):
READY

src/shapes/polygon.class.ts
```diff
@@ -1,11 +1,47 @@
...
   }
+
+  static getRegularPolygonPoints({
+    numVertexes,
+    radius,
+  }: {
+    numVertexes: number;
+    radius: number;
+  }) {
+    const interiorAngle = (Math.PI * 2) / numVertexes;
```
asturur, line 21: we have a const for `Math.PI * 2`

src/shapes/polygon.class.ts
```diff
@@ -1,12 +1,48 @@
...
+  }: {
+    numVertexes: number;
+    radius: number;
+  } & Record<string, unknown>) {
+    return new this(this.getRegularPolygonPoints({ numVertexes, radius }), {
+      left: 0,
+      top: 0,
+      ...options,
+    });
+  }
```
asturur, line 45: this extra static method seems like not so much useful.
`new Polygon(Polygon.getRegularPolygonPoints({ numVertexes, radius }), options);`

is the extra wrapper worth it? I understand left and top equal 0 are part of it, but where the polygon is positioned seems like not something we really care about. I would argue most of people would say that center of the canvas is a better default.
On top of that if i use originX/Y center, defaulting the top left to 0,0 does not really help

asturur: i would say of not having this extra static method

ShaMan123: It is worth it.
Setting left, top to 0 fixes the position for all cases of origin. Without it it is fixed at origin center, no matter what origin you pass to it since `getRegularPolygonPoints` is calculated from origin point 0,0 center.
https://github.com/fabricjs/fabric.js/blob/b6f076cc0c817cd7f70abc7028c9d04295d121a8/src/shapes/polyline.class.ts#L73

ShaMan123: > but where the polygon is positioned seems like not something we really care about

For consistency

test/unit/polygon.js
```diff
@@ -255,4 +255,45 @@
...
+    ];
+
+    assert.deepEqual(roundDecimals(penta), roundDecimals(expectedPenta), 'regualr pentagon should match');
+    assert.deepEqual(roundDecimals(hexa), roundDecimals(expectedHexa), 'regualr hexagon should match');
+    assert.deepEqual(
+      fabric.Polygon.createRegularPolygon({ numVertexes: 5, radius: 50 }).toObject(),
+      new fabric.Polygon(expectedPenta, { left: 0, top: 0 }).toObject(),
+      'same polygon'
+    );
+    assert.deepEqual(
```
ShaMan123, line 293: this test will fail with left,top = 0

asturur:
I have been thinking this and the reason why i didn't merge it right away is because i really like as a util.
Why does it have to be attached to the polygon class?

Can't just be a `createRegularPolygon` and return the polygon itself ?

ShaMan123:
Can be a util
It is a question of scope, where would the dev look for it.
Since it is strictly a polygon I thought to move it there but I don't really mind.

PR #8542, opened 2022-12-29
chore(TS) Remove small mixins that are too small to be worthy of existance

asturur:
## Motivation
Continue migration of TS

## Changes
- BREAKING: remove canvas.straightenObject
- BREAKING: cloneWithoutData clones only the size
- BREAKING: canvas.clone now correctly return either a canvas or staticCanvas
- 3 mixin file less.


ShaMan123 (commented):
IMO canvas straightening belongs to animation

asturur:
all canvas animations are possibly dumped, are 4 premade animations that who needs them need probably more than 4 and can just make them

asturur:
removed `straightenObject` from canvas. Is not using any canvas property, is another:

objet.method()  => method(object)

not worth anyone time

asturur:
I think all the canvas animation can become functions that you can import to which you give a canvas and an object and you obtain an animaton.
Does not need to be on the instance, seems like something that was started and never really developed, and they look like more like 4 example animations to me
delete, bring to center ( x or y ), straighten

src/canvas/static_canvas.class.ts
```diff
@@ -1296,9 +1298,9 @@ export class StaticCanvas<
   createSVGRefElementsMarkup(): string {
     return ['background', 'overlay']
```
ShaMan123, line 1299:
```suggestion
    return (['background', 'overlay'] as const)
```

will type the array and remove the need for the extra as 'overlayColor' | `backgroundColor`

ShaMan123 (commented):
I am for removing the animation mixin for good

src/canvas/static_canvas.class.ts
ShaMan123: 6b10575

src/canvas/static_canvas.class.ts
```diff
@@ -1643,6 +1645,120 @@ export class StaticCanvas<
...
+   * create quick image copies of a canvas without passing for the dataUrl string
+   * @param {Number} [multiplier] a zoom factor.
+   * @param {Object} [options] Cropping informations
+   * @param {Number} [options.left] Cropping left offset.
+   * @param {Number} [options.top] Cropping top offset.
+   * @param {Number} [options.width] Cropping width.
+   * @param {Number} [options.height] Cropping height.
+   * @param {(object: fabric.Object) => boolean} [options.filter] Function to filter objects.
+   */
+  toCanvasElement(
```
ShaMan123, line 1717: why don't we create a new StaticCanvas instance instead of all this mess? I don't believe it will cost in perf and we will benefit in clarity

ShaMan123 (dismissed):
see my comments

src/canvas/static_canvas.class.ts
ShaMan123: Well actually I tried it but it means that all objects need to be iterated over to set canvas on them so retina will be correct.
So no

ShaMan123:
```ts
  toCanvasElement(
    multiplier = 1,
    { width, height, left, top, filter } = {} as TToCanvasElementOptions
  ): HTMLCanvasElement {
    const scaledWidth = (width || this.width) * multiplier,
      scaledHeight = (height || this.height) * multiplier,
      zoom = this.getZoom(),
      newZoom = zoom * multiplier,
      vp = this.viewportTransform,
      canvasEl = createCanvasElement(),
      objectsToRender = filter ? this._objects.filter(filter) : this._objects;
    canvasEl.width = scaledWidth;
    canvasEl.height = scaledHeight;
    const canvas = new StaticCanvas(canvasEl, {
      width: scaledWidth,
      height: scaledHeight,
      enableRetinaScaling: false,
      viewportTransform: [
        newZoom,
        0,
        0,
        newZoom,
        (vp[4] - (left || 0)) * multiplier,
        (vp[5] - (top || 0)) * multiplier,
      ] as TMat2D,
      backgroundColor: this.backgroundColor,
      backgroundImage: this.backgroundImage,
      backgroundVpt: this.backgroundVpt,
      overlayColor: this.overlayColor,
      overlayImage: this.overlayImage,
      overlayVpt: this.overlayVpt,
      clipPath: this.clipPath,
    });
    canvas.calcViewportBoundaries();
    objectsToRender.forEach((obj) => obj._set('canvas', canvas));
    canvas.renderCanvas(canvasEl.getContext('2d')!, objectsToRender);
    objectsToRender.forEach((obj) => obj._set('canvas', canvas));
    return canvasEl;
  }
```

asturur: yes this is a long standing issue, there is not easy way to export a render to a different canvas without swapping the canvas values or the object canvas reference. Something that can be solved with a bit of thinking, but also not a large issue as long as this method works

src/canvas/static_canvas.class.ts
```diff
@@ -1643,6 +1657,217 @@ export class StaticCanvas<
...
+   * Clones canvas instance without cloning existing data.
+   * This essentially copies canvas dimensions since loadFromJSON does not affect canvas size.
+   * @returns {StaticCanvas}
+   */
+  cloneWithoutData(): StaticCanvas {
+    const el = createCanvasElement();
+    el.width = this.width;
+    el.height = this.height;
+    // this seems wrong. either Canvas or StaticCanvas
+    return new StaticCanvas(el);
```
ShaMan123, line 1754: you can do
```ts
new this.constructor(el)
```

Then it takes the instance class

asturur: tried already , is not happy about it. Neither with new this(el) nor with new this.contrsuctor(el).

PR #8543, opened 2022-12-29
refactor(): canvas text editing manager

ShaMan123:
## Motivation

Another step making text classes readable, simpler, **Scoped** which they are absolutely not

## Description

extracted the logic that handles canvas itext instances from itext_behavior to a dedicated manager

## Changes

- register itext from `_set`. This will handle cases such as nested itexts.
- delegate mouse move invocation to the manager


src/mixins/itext_behavior.mixin.ts
```diff
@@ -1061,8 +992,8 @@ export abstract class ITextBehaviorMixin<
     }
     this.fire('editing:exited');
     isTextChanged && this.fire('modified');
+    this._canvasMoveDisposer && this._canvasMoveDisposer();
     if (this.canvas) {
-      this.canvas.off('mouse:move', this.mouseMoveHandler);
```
ShaMan123, line 1065: changed this bit to use a disposer `_canvasMoveDisposer` in order to safeguard from a case where while editing text it is removed from canvas. I will add a test for it.

src/shapes/text.class.ts
```diff
@@ -94,7 +94,7 @@ export class Text<
    * @type Array
    * @private
    */
-  _dimensionAffectingProps: (keyof this)[];
+  _dimensionAffectingProps: string[];
```
ShaMan123, line 97: TS went mad because of this line

src/canvas/CanvasTextEditingManager.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  informMouseUp() {
+    this.targets.forEach((obj) => {
+      obj.__isMousedown = false;
+    });
+  }
+
+  add(target: IText | Textbox) {
+    if (!this.disposer) {
+      const disposer = this.canvas.on('mouse:up', () => {
```
ShaMan123, line 32: I will try to remove the need for the subscription

ShaMan123 (commented):
Progress

src/mixins/itext_behavior.mixin.ts
ShaMan123: added

src/canvas/CanvasTextEditingManager.ts
ShaMan123: done

ShaMan123 (commented):
seems ready
tested live as well

ShaMan123 (commented):
updated from master

ShaMan123 (commented):
extensive test

asturur:
Need to read the diffs better, but this seems ok to me.

src/shapes/itext.class.ts
```diff
@@ -166,10 +167,14 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
...
     if (this.isEditing && this._savedProps && key in this._savedProps) {
       this._savedProps[key] = value;
-    } else {
-      super._set(key, value);
+      return this;
+    }
+    if (key === 'canvas') {
+      this.canvas instanceof Canvas &&
+        this.canvas.textEditingManager.remove(this);
+      value instanceof Canvas && value.textEditingManager.add(this);
```
asturur, line 175: Is his the correct place to catch this?
Can't we leverage the collection onObjectAdded/onObjectRemoved to handle this?

src/mixins/itext_behavior.mixin.ts
```diff
@@ -461,37 +401,21 @@ export abstract class ITextBehaviorMixin<
...
-  initMouseMoveHandler() {
-    this.canvas.on('mouse:move', this.mouseMoveHandler);
-  }
-
   /**
    * @private
    */
-  mouseMoveHandler(options) {
+  protected mouseMoveHandler(options) {
     if (!this.__isMousedown || !this.isEditing) {
```
asturur, line 419: i don't want to change more on this PR, but looks like that there is an opportunity for simplfication if we find a way to handle this either on the canvas or adding and removing this event handler every mousedown/mouseup rather than on editing entered/exit.

Defintely lower priority, but i point it out because following what was .__isMouseDown and why do we need to set it off on all instances every mouse up is something you forget every time and you have to rediscover

src/mixins/itext_behavior.mixin.ts
```diff
@@ -445,7 +385,7 @@ export abstract class ITextBehaviorMixin<
     }
     if (this.canvas) {
       this.canvas.calcOffset();
-      this.exitEditingOnOthers(this.canvas);
+      this.canvas.exitTextEditing();
```
asturur, line 388: do we want to have this alias here, or since the textEditingManager is a present in every class we just call the method directly and we are explicit we are using that manager?

src/shapes/itext.class.ts
ShaMan123: it is not good enough taking into account nested targets

src/mixins/itext_behavior.mixin.ts
ShaMan123: Not sure I understand

src/canvas/TextEditingManager.ts
```diff
@@ -0,0 +1,40 @@
...
+      if (obj.isEditing) {
+        obj.exitEditing();
+      }
+    });
+  }
+
+  /**
+   * called from canvas mouse up
+   */
+  informMouseUp() {
```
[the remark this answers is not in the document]
ShaMan123: refactored

src/mixins/itext_behavior.mixin.ts
ShaMan123: seems to have worked...

src/mixins/itext_behavior.mixin.ts
```diff
@@ -521,7 +434,6 @@ export abstract class ITextBehaviorMixin<
       this.selectionStart !== currentStart ||
       this.selectionEnd !== currentEnd
     ) {
-      this.restartCursorIfNeeded();
```
ShaMan123, line 524: important fix
no reason to do this here, in case of a cursor the animation will be called from mouse up

src/mixins/itext_behavior.mixin.ts
```diff
@@ -461,46 +399,21 @@ export abstract class ITextBehaviorMixin<
...
-  /**
-   * @private
+   * called by {@link canvas#textEditingManager}
    */
-  mouseMoveHandler(options) {
-    if (!this.__isMousedown || !this.isEditing) {
-      return;
-    }
-
+  updateSelection(e: TPointerEvent) {
```
ShaMan123, line 411: this method will be called only if a registered instance is editing so no need to recheck

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -181,15 +186,11 @@ export abstract class ITextClickBehaviorMixin<
...
-        return;
-      }
+    if (activeObject !== this) {
+      // avoid running this logic when there is an active object
+      // this because is possible with shift click and fast clicks,
+      // to rapidly deselect and reselect this object and trigger an enterEdit
+      return;
     }
 
     if (this.__lastSelected && !this.__corner) {
```
ShaMan123, line 196: this bit is really harsh, hard to understand logic

ShaMan123 (commented):
I am not sure why the mouse move handler (now renamed to `updateSelection`) is not called from he object's mousemove event.

- It seems `selected` is handled by mouse up so no need to false it from the manager
- rm `__isMousedown`: 
   - used for updating selection **only** when mouse move after mouse down (now handled by the manager)
   - used to render the cursor with global alpha = 1, fixed to not run cursor animation in such a case


This PR is becoming less scoped and is about to clash with #8583 #8534 

A few fixes so tests pass and I am done hopefully

ShaMan123 (commented):
READY

src/shapes/itext.class.ts
asturur: What does it mean nested targets?

src/mixins/itext_key_behavior.mixin.ts
```diff
@@ -111,7 +111,7 @@ export abstract class ITextKeyBehaviorMixin<
    * Override this method to customize cursor behavior on textbox blur
    */
   blur() {
-    this.abortCursorAnimation();
+    !this.selected && this.abortCursorAnimation();
```
asturur, line 114: not sure what `this.selected` is anymore and if is consistent across all fabric.
I would rather not use it.

asturur: let's not make this change. if the user select a ui textarea the default is to not show 2 cursor animations

src/mixins/itext_behavior.mixin.ts
```diff
@@ -1,15 +1,12 @@
 // @ts-nocheck
 
 import { getEnv } from '../env';
-import { ObjectEvents, TEvent } from '../EventTypeDefs';
+import { ObjectEvents, TEvent, TPointerEvent } from '../EventTypeDefs';
 import { Point } from '../point.class';
 import { Text } from '../shapes/text.class';
-import { TPointerEvent } from '../typedefs';
```
asturur, line 7: do we have 2 definitions or was this a broken import?

src/mixins/itext_behavior.mixin.ts
```diff
@@ -92,11 +88,9 @@ export abstract class ITextBehaviorMixin<
    * Initializes all the interactive behavior of IText
    */
   initBehavior() {
-    this.initAddedHandler();
-    this.initRemovedHandler();
     this.initCursorSelectionHandlers();
     this.initDoubleClickSimulation();
-    this.mouseMoveHandler = this.mouseMoveHandler.bind(this);
+    this.updateSelection = this.updateSelection.bind(this);
```
asturur, line 93: this method needs to be called something that is both mouse move and handler, so that we avoid it using differently later.
Like updateSelectionOnMouseMove or something like that.

asturur:
I think this all in all is better than what we had before, is also a change that will open up to regressions for sure.
If you can do that rename and remove the `!this.selected` then we can merge

src/mixins/itext_behavior.mixin.ts
ShaMan123: Ne already

asturur:
Oh also please tell me more about those nexted Itexts
Are you talking about adding a group that has an itext inside?

src/mixins/itext_behavior.mixin.ts
ShaMan123: broken import due to `@ts-nocheck`

src/mixins/itext_key_behavior.mixin.ts
ShaMan123: exactly
I was aiming for that and I am to blame for it, see #7784
The thing is it causes a flicker on mouse down since the hidden textarea blurs and immediately focuses.

ShaMan123: But maybe best to have this is a separate PR and test it with all cases...

ShaMan123: `selected` is an internal thing in text, maybe should be removed

src/shapes/itext.class.ts
ShaMan123: yes, text under groups

src/mixins/itext_behavior.mixin.ts
ShaMan123: a237a9ac0

ShaMan123 (commented):
applied changes from review

src/mixins/itext_behavior.mixin.ts
```diff
@@ -193,7 +132,7 @@ export abstract class ITextBehaviorMixin<
       },
     };
 
-    obj.animate('_currentCursorOpacity', targetOpacity, {
+    obj._animate('_currentCursorOpacity', targetOpacity, {
```
ShaMan123, line 135: Just to be clear this fixes the cursor regression,
I don't mind PRing this line, just had to fix it to see cursor works with this PR changes

ShaMan123: #8589

asturur (approved):

asturur:
This is good work imho, just not convinced about that _set usage.
Would really leave set to small things.
I m sure later we can find a way to use onObject added for that.
I don't see an issue with onObjectAdded traversing the objects.
Something is setting canvas prop on every object in a group when we move a group from a canvas to another, so this mechanism already exists somewhere

ShaMan123:
Very good point
We can move away from set

PR #8547, opened 2022-12-30
refactor(): BREAKING Animation removing byValue, removing fxStraigthen.

asturur:
## Motivation

I would like the animation code and interface to be simpler.
So i m proposiing some simplificaiton, it makes also documenting animation easier.

## Description

Removal some flexibility in the api in exchange of some code semplification

i also create a folder `parkinglot` where i put the straighten code and i think i m putting the canvas effects too.
Those needs to be changed in examples or something that the developer can grab, but i don't think they should be part of af the class itself. We have 4 effects, while the possibilities are so many that i think providing 4 of them and nothing else just makes noise.

Other things i would like to solve:
- keypath in animation. I don't have a solution right now but ideally i would like to list the cases in which a keypat is needed ( gradients i m sure and little else probably )
- duality between byValue and endValue, i would like to keep one only.

## Changes

- Object.animate takes only an object with one or more properties
- animate does not take += or -= anymore, we can write an addition or subtraction when we do the animation end value.
- removed barrel imports
- Object.animate does not undefine anymore the onChange and onComplete of multiple keys. Reasons are that with requestRenderAll that is not needed anymore, and since you can cancel a single animation, if you cancel the one that carry the actual callbacks the others are not working anymore. We need to be clear on the fact that you shoudn't do expensive things on onChange and that you should use requestRenderAll rather than renderAll.

### naming changes 
```ts
export type TOnAnimationChangeCallback<T, R = void> = (
  value: T,
  valueRatio: number,
  durationRatio: number
) => R;
```
into
```ts
export type TOnAnimationChangeCallback<T, R = void> = (
  value: T,
  valueProgress: number,
  timeProgress: number
) => R;
```
Reason: ratio is generic, can be 3, 4, -3 0.5. We have values here that are always between 0 and 1 and represent the progress between start and end. I think this naming would be more clear.


asturur:
@ShaMan123  @Lazauya Let's talk about this small changes here. Also do you have any opinion around endValue and byValue?

src/shapes/Object/AnimatableObject.ts
```diff
@@ -96,37 +56,21 @@ export abstract class AnimatableObject<
    */
   _animate<T extends number | TColorArg>(
     key: string,
-    to: T,
+    endValue: T,
     options: Partial<TAnimationOptions<T>> = {}
   ) {
     const path = key.split('.');
     const propIsColor = this.colorProperties.includes(path[path.length - 1]);
-    const currentValue = path.reduce((deep: any, key) => deep[key], this);
```
ShaMan123, line 104: This line handled the key path dot notation using `key.split('.')`

ShaMan123 (commented):
- Is parkinglot an known thing? I love the name but I am guessing you will rename to something generic like `examples`. 
- I would change the content of the parkinglot to simple methods, not mixins, e.g. `removeObjectWithFading`
- As suggested earlier I would expose an animation controller class that runs numerous animations in parallel (maybe also in sequence) and execute callbacks with an array of values for each animation in the controller. That will make `Animatable#_animate` dead. It was a mess, now is better but still is pretty shit. I would remove it entirely. The animation API is good, flexible and straight forward. Does this abstraction provide any value?
- regarding `renderAll` vs `requestRenderAll`, I thought I should always call `renderAll` from within an animation callback since it executes from within a requested frame. If I call `requestRenderAll` it will request another frame so my animation effects will always be a frame late. Please explain and correct me if I am wrong.
- `byValue` `endValue`, AFAIAC it was solved in #8297 . And TS will warn the dev when passing both of them, or should at least.
- naming changes - I am for. `valueProgress` is good and I would go for `durationProgress` cause that is more accurate vs. `timeProgress` cause time is confusing. Duration is the time since we started so no ambiguity.

asturur:
> * Is parkinglot an known thing? I love the name but I am guessing you will rename to something generic like `examples`.
> * I would change the content of the parkinglot to simple methods, not mixins, e.g. `removeObjectWithFading`

Parking lot is just a middleground between deleting the code and fix it now. Are just files i don't think we should deal with now. Are not even supposed to work, just memory. I tried to keep them in functional state tho.

> * As suggested earlier I would expose an animation controller class that runs numerous animations in parallel (maybe also in sequence) and execute callbacks with an array of values for each animation in the controller. That will make `Animatable#_animate` dead. It was a mess, now is better but still is pretty shit. I would remove it entirely. The animation API is good, flexible and straight forward. Does this abstraction provide any value?
> * regarding `renderAll` vs `requestRenderAll`, I thought I should always call `renderAll` from within an animation callback since it executes from within a requested frame. If I call `requestRenderAll` it will request another frame so my animation effects will always be a frame late. Please explain and correct me if I am wrong.

The issue with renderAll is that is gonna render N times you call it. requestRenderAll just the last one is going to trigger.
So for now, if you animate 3 properties ( left, top, angle ) you should use requestRenderAll.

> * `byValue` `endValue`, AFAIAC it was solved in #8297 . And TS will warn the dev when passing both of them, or should at least.

I want one to disappear. not be warned about. What is the deal with having both?
Seems to me that byValue is less understandable than endValue.
Is natural to say i want to animate left from 30 to 60 or to 60 only.
is easy to say, i want to move by 30 right, so i will animate left and i to `this.left + 30`
byValue fall short with colors. i can think of an animation from red to blue, but is a bit stranger to think with byValue, saying that you are animating by value with `[-255, 0, 255]` in order to cancel red channel and increase the blue.

> * naming changes - I am for. `valueProgress` is good and I would go for `durationProgress` cause that is more accurate vs. `timeProgress` cause time is confusing. Duration is the time since we started so no ambiguity.

Yes for me this is the same, i just dislike ratio.

src/shapes/Object/AnimatableObject.ts
asturur: i just moved it down, we need to execute it only if startValue is not passed.

asturur:
Ok renames pushed up.
We can get back to those changes first week of new year, but likely i would like byValue to go.

ShaMan123:
I didn't want to have both by/end value. It was something I wasn't sure you would accept breaking.
Kill it.

ShaMan123:
IMO we could move the key path logic to `_set` or `set` instead of part of animation logic

src/util/animation/types.ts
```diff
@@ -5,13 +5,13 @@ export type AnimationState = 'pending' | 'running' | 'completed' | 'aborted';
 /**
  * Callback called every frame
  * @param {number | number[]} value current value of the animation.
- * @param valueRatio ∈ [0, 1], current value / end value.
- * @param durationRatio ∈ [0, 1], time passed / duration.
+ * @param {number} valueProgress ∈ [0, 1], current value / end value.
```
ShaMan123, line 8: > `current value / end value.`

seems off.

asturur:
> IMO we could move the key path logic to `_set` or `set` instead of part of animation logic

That is more a lodash thing, and dangerous becase then people find a way to exploit it with `parent.__proto__._proto__.__proto__` and do some strange code pollution things.

keypath logic can go away as soon as we can animate a gradient or a pattern or a point of polygon.
I m sure it can be done already, is probably that this object.animate is supposed to be a shortcut to use util.animate and i adding some constrains. But if we have the animate function clearly used on obejcts, we can remove object.animate entirely.
As an added bonus developers can use any animation engine to change the values on objects.

asturur:
@ShaMan123 to be clear, i m ok breaking things because we remove code and functionalities, i m not ok breaking things now to add more functionalities or more options.

ShaMan123 (commented):
I don't get why the base class accepts byValue while subclasses accept endValue.
Can't we remove byValue entirely and calc it just for easing?
Updating the ts tests is in order as well.

asturur:
> I don't get why the base class accepts byValue while subclasses accept endValue. Can't we remove byValue entirely and calc it just for easing? Updating the ts tests is in order as well.

Oh well it was the easy way to get there.
The subclass is not exposed to the final dev, and calculating it for the easing means calculating it every tick.
So all in all i thought that this change was sufficient and not wrong.

Is there any particular reason why it sounds bad to you?

asturur:
Where are those ts tests?

Lazauya:
I wanted to push some doc and naming fixes (durationProgress -> timeProgress mostly). I tried creating a PR from my own fork branch but it didn't work. Am I being stupid?

test/ts/animation.ts
```diff
@@ -9,11 +9,12 @@ animate({
...
 });
 animate({
+  // @ts-expect-error `byValue` is not part of options
   byValue: 2,
 });
-// @ts-expect-error only one of (`byValue` | `endValue`) is allowed
 animate({
   endValue: 3,
+  // @ts-expect-error `byValue` is not part of options
   byValue: 2,
```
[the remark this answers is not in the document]
ShaMan123: Fixed by using signature overload 85bd8eef2

ShaMan123 (commented):
The only question remaining is should we merge between animate and animateColor.
 I say we should.

ShaMan123:
> Is there any particular reason why it sounds bad to you?

It is weird that we calculate `byValue` according to `endValue` in the subclass and then calculate `endValue` according to `byValue` in the base class but I can live with it.
For sure it makes a lot more sense accepting `endValue` rather than `byValue` in the exposed method.

src/util/animation/ColorAnimation.ts
```diff
@@ -1,26 +1,36 @@
...
+};
 
 const wrapColorCallback = <R>(
   callback?: TOnAnimationChangeCallback<string, R>
 ) =>
   callback &&
-  ((rgba: TRGBAColorSource, valueRatio: number, durationRatio: number) =>
-    callback(new Color(rgba).toRgba(), valueRatio, durationRatio));
+  ((rgba: TRGBAColorSource, valueProgress: number, durationProgress: number) =>
+    callback(new Color(rgba).toRgba(), valueProgress, durationProgress));
```
ShaMan123, line 27: I also think that the the color animation callbacks should pass a `Color` instance rather that a color string
```suggestion
    callback(new Color(rgba), valueProgress, durationProgress));
```

ShaMan123:
> The only question remaining is should we merge between animate and animateColor?

If we change color animation options to accept only a color instance (which makes sense and is not that breaking) we could merge it into `animate` w/o a problem by checking `instance of Color` and extend the `isArrayAnimation` functionality

ShaMan123:
Looking at the code more closely I think `byValue` doesn't belong to the base class...
It is only defined on it but not used by it.
We can define it on the subclasses for their own use.
Thoughts?

src/util/animation/ColorAnimation.ts
ShaMan123: > > The only question remaining is should we merge between animate and animateColor?
> 
> If we change color animation options to accept only a color instance (which makes sense and is not that breaking) we could merge it into `animate` w/o a problem by checking `instance of Color` and extend the `isArrayAnimation` functionality

https://github.com/fabricjs/fabric.js/compare/small-animation-preferences...animate-color

ShaMan123:
> > The only question remaining is should we merge between animate and animateColor?
> 
> If we change color animation options to accept only a color instance (which makes sense and is not that breaking) we could merge it into `animate` w/o a problem by checking `instance of Color` and extend the `isArrayAnimation` functionality

https://github.com/fabricjs/fabric.js/compare/small-animation-preferences...animate-color

asturur:
Let's merge this first part, thanks for the fixes and types improvement and thanks for trying @Lazuya
Also @Lazauya i don't see why opening a PR from your branch to this branch shouldn't work.
Also @ShaMan123 if you want to commit freely to a pr please squash the commits. Is bad to open N commits and check one by one what did change. Don't change the course of a PR, and if you want to commit squash them so i can at least disagree/agree quickly.

asturur:
What should i do? can i merge or we need to do something else?

ShaMan123:
merge
I can change stuff to accommodate color animation in a follow up

ShaMan123:
just realized this PR broke text cursor animation
will PR a fix right away

PR #8549, opened 2022-12-30
build() Remove header.js, remove fabric global object, use exports from TS to build the lib

asturur:
## Motivation

Try to package a build with proper import export withour relying on the fabric global object

## Description

Slowly trying to get this to work, i hoped it was straight forward but it didn't work out of the box

### Main changes
- removed header and fabric global references from the source code.
- In order to have document, window, isLikelyNode working a stopgap object returned from getEnv() has been created
- removed index.js in favor of a standard import structure ( for now no exports )

### Other changes
- converted stateful mixin to have this working
- removed setupState from the stateful mixin
- util addListener and removeListener are gone


tsconfig.json
```diff
@@ -25,7 +25,7 @@
 
     /* Modules */
     "module": "es2022" /* Specify what module code is generated. */,
-    // "rootDir": "./",                                  /* Specify the root folder within your source files. */
+    "rootDir": "./",                                  /* Specify the root folder within your source files. */
```
asturur, line 28: this was the fixing change, could be that it would have worked with the old dependencies anyway

asturur:
Tomorrow i will try to make this work.
I will make a list of everything that is current fabricJS object, what is going to be in the export, what is left out and why.

We need to solve global space pollution for consts that is broken since we switched to rollup

asturur:
type generation does not work, this ecosystem is shit.

ShaMan123:
Why do you dislike the index files?
I don't get it.
It makes sense to have an index file for controls, brushes etc.
imports are broken

ShaMan123:
Seems to work now

ShaMan123:
IMO the HEADER fucks it all up

asturur:
The issue with index files is that sometimes they pull in everything even when you don't need it, is an extra file to maintain we don't need.

asturur:
Yes the HEADER needs to be removed yet.
But i was expecting to get some stuff running and defined already.
Not sure why it didn't catch during build

asturur:
The situation is exactly as before, type definitions aren't generated for this export structure.

asturur:
got rid of header with a getEnv function that returns document window and the other things that we can get at runtime only.

ShaMan123:
I got types during build

asturur:
Ok i have something weird going on, i ll think about it again later or another day.
Now the prototype inheritance is completely fucked up for somer reason and what we set on the prototype is undefined while running the constructor.

I should try to commit the removal of header in a separate PR that does not touch anything else.

asturur:
So is the export statement that breaks our current rollup + typescript setup.
In order to get it to work i need to update to update to rollup 3.
Updating rollup to 3 breaks the code and the classes.

I don't want to understand why now, so i avoided using the export statement, and of course now i can't get any type emitted.

But i can get rid of header and the internal fabric object entirely, and that is good enough for a bit of progress.

src/canvas/static_canvas.class.ts
```diff
@@ -430,18 +430,16 @@ export class StaticCanvas<
 
     this.width = this.width || lowerCanvasEl.width || 0;
     this.height = this.height || lowerCanvasEl.height || 0;
+    this.viewportTransform = [...this.viewportTransform];
```
ShaMan123, line 433: good catch!

src/shapes/Object/Object.ts
```diff
@@ -1703,10 +1706,10 @@ export class FabricObject<
    * @param {Boolean} [options.withoutShadow] Remove current object shadow. Introduced in 2.4.2
    * @return {Image} Object cloned as image.
    */
-  cloneAsImage(options: any) {
-    const canvasEl = this.toCanvasElement(options);
-    // TODO: how to import Image w/o an import cycle?
-    return new fabric.Image(canvasEl);
+  cloneAsImage(options: any): Image | null {
```
ShaMan123, line 1709: I can't accept this.
Let's remove it entirely.
So much ugliness for a method of 2 lines

src/shapes/image.class.ts
```diff
@@ -795,7 +792,31 @@ export const imageDefaultValues: Partial<TClassProperties<Image>> = {
...
+   * @param {Boolean} [options.withoutShadow] Remove current object shadow. Introduced in 2.4.2
+   * @return {Image} Object cloned as image.
+   */
+  cloneAsImage(options: any): Image {
+    const canvasEl = this.toCanvasElement(options);
+    // TODO: how to import Image w/o an import cycle?
+    return new Image(canvasEl);
+  }
+});
+
```
ShaMan123, line 820: 🤮

src/shapes/text.class.ts
```diff
@@ -1609,10 +1609,8 @@ export class Text<
         ? style.fontFamily
         : `"${style.fontFamily}"`;
     return [
-      // node-canvas needs "weight style", while browsers need "style weight"
-      // verify if this can be fixed in JSDOM
-      fabric.isLikelyNode ? style.fontWeight : style.fontStyle,
-      fabric.isLikelyNode ? style.fontStyle : style.fontWeight,
+      style.fontStyle,
+      style.fontWeight,
```
ShaMan123, line 1613: ?

src/util/misc/dom.ts
```diff
@@ -1,18 +1,18 @@
...
-  fabric.document.createElement('canvas');
+  getEnv().document.createElement('canvas');
 
 /**
  * Creates image element (works on client and node)
  * @return {HTMLImageElement} HTML image element
  */
 export const createImage = (): HTMLImageElement =>
-  fabric.document.createElement('img');
+  getEnv().document.createElement('img');
```
ShaMan123, line 15: I would move these method to env as well

src/util/misc/findScaleTo.ts
```diff
@@ -1,4 +1,4 @@
-interface IWithDimensions {
+export interface IWithDimensions {
```
ShaMan123, line 1: can we use `TSize` instead? Or are the comments important?

ShaMan123 (changes requested):
see comments

src/shapes/image.class.ts
ShaMan123: If you must keep this make it a util

src/controls/default_controls.ts
```diff
@@ -122,14 +127,14 @@ FabricObject.prototype.controls = {
...
   // is important to document that if you want to have all objects to have a
   // specific custom control, you have to add it to Object prototype and to Textbox
   // prototype. The controls are shared as references. So changes to control `tr`
   // can still apply to all objects if needed.
-  fabric.Textbox.prototype.controls = {
-    ...(fabric.Textbox.prototype.controls || {}),
+  Textbox.prototype.controls = {
+    ...(Textbox.prototype.controls || {}),
     ...textboxDefaultControls,
   };
```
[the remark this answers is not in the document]
asturur: yes we should do for both case, so we don't have to import this file but Object and Textbox will take care of it.

src/env.ts
```diff
@@ -0,0 +1,82 @@
...
+  'ontouchstart' in fabricDocument ||
+  (fabricWindow &&
+    fabricWindow.navigator &&
+    fabricWindow.navigator.maxTouchPoints > 0);
+  config.configure({
+    devicePixelRatio: fabricWindow.devicePixelRatio || 1,
+  });
+}
+
+setupEnv();
```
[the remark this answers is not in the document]
asturur: The issue with lazy initializing this is that it also takes care of a config value.
At that point you don't know when that config value get requested

src/shapes/Object/Object.ts
asturur: for now it stays like this, we can reiterate.
I don't think cloneAsImage is that useful.
I also don't think many utils are that useful, but removing them now or stop exposing them now creates issues with tests
I m putting back things right now more than removing them, i was hoping i could just move around imports rather than change all of this

src/util/misc/findScaleTo.ts
asturur: this was a ts compiler issue, i just added export to get it passing.
Not sure if TSize allow for passing objects that have more than width and height, but i can try it

fabric.js
```diff
@@ -0,0 +1,414 @@
...
+import { HueRotation } from './src/filters/hue_rotation.class';
+import { Invert } from './src/filters/invert_filter.class';
+import { Noise } from './src/filters/noise_filter.class';
+import { Pixelate } from './src/filters/pixelate_filter.class';
+import { RemoveColor } from './src/filters/removecolor_filter.class';
+import { Resize } from './src/filters/resize_filter.class';
+import { Saturation } from './src/filters/saturate_filter.class';
+import { Vibrance } from './src/filters/vibrance_filter.class';
+
+const filters = {
```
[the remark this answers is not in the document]
asturur: yes the fabricJS bundle is not tree shakable indeed.
I m just replicating the old bundle.
There are other obstacles to tree shaking like some side effects we run ( the prototype changes or creating and assigning controls )
When we get exports to work we can experiment with treeshaking but i don't have a setup to do it right now.

This is like the old index.js, but brining the fabricjs object build outside the codebase and single files.

fabric.js
```diff
@@ -0,0 +1,414 @@
...
+}
+
+if (typeof exports !== 'undefined') {
+  exports.fabric = fabric;
+} else if (typeof define === 'function' && define.amd) {
+  /* _AMD_START_ */
+  define([], function () {
+    return fabric;
+  });
+}
```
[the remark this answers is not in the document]
asturur: this should be `export { fabric }` buf if i do it, is all broken.
rollup stop working, and if i update rollup to 3 it works, but then the code does not work anymore.
(if you look in the commit list you find various iterations of all of this )
Better here for now than in the codebase itself.

src/canvas/static_canvas.class.ts
asturur: i catched just because this is something that breaks with exports.
the values we put on the prototype become misteriously inaccessible, so viewportTransform is undefiend rather than the default and i have no idea why.
I m not sure if is a side effect of Object.freeze or else.
But yes this was a bug.

src/canvas/static_canvas.class.ts
```diff
@@ -1774,7 +1772,7 @@ Object.assign(StaticCanvas.prototype, {
   clipPath: undefined,
 });
 
-if (fabric.isLikelyNode) {
+if (getEnv().isLikelyNode) {
```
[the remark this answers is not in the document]
asturur: getEnv is a stop gap thing

ShaMan123:
@asturur I am sorry if I made changes against your latest commits.
Seems that we must upgrade rollup deps for the tool to work.
Then as you said, it uncovers that proto assignments are not working. I have to say that makes sense. I didn't under how assigning values to the prototype propagated to the instance. Somehow it worked up to now.
We only need to expose a static `getDefaults` or what ever and call it from within the constructor and it seems we are ready to roll

asturur:
@ShaMan123 can you please revert your latest commits?
I m handling this and i ll bring to a poitn in which at least we don't have header and a list of imports with no structure.
Then we can reiterate right after.
I decided on purpose to avoid lib updates and exports on this PR since then nothing was working and too many things changing all at once wasn't easy do debug.

asturur:
> @asturur I am sorry if I made changes against your latest commits. Seems that we must upgrade rollup deps for the tool to work. Then as you said, it uncovers that proto assignments are not working. I have to say that makes sense. I didn't understand how assigning values to the prototype propagated to the instance. Somehow it worked up to now. We only need to expose a static `getDefaults` or what ever and call it from within the constructor and it seems we are ready to roll

We can do that when we get rid of the defaults on the prototype as we discussed in the past.
We don't do it right now, this PR already ballooned up beyond what we like and i m trying to reduce it down to less changes not more.

asturur:
I also don't like the idea of having frozen prototypes. so i want to understand why it breaks, if is a transpiling option or what

ShaMan123:
This could be due to bumping typescript....
If so, I don't see a way around that

ShaMan123:
I will try to fork this and apply getDefaults and see what happens.
ThoughtS?

ShaMan123:
reverted deps
build is broken

asturur:
> I will try to fork this and apply getDefaults and see what happens. ThoughtS?

This is not forkable right now is a mess.
Wait for it to be merged

ShaMan123:
I almost lost my wits and sanity the past hour.
It is that vile bug that we encountered a while back with class fields and `set`.
Called `set` again and tests started working.
I absolutely hate build tools

ShaMan123:
Some shit is happening in the constructor. Is that what you meant when stating that the prototype chain broke somehow?
I set defaults in Object, but then Circle has no default values set even though it calls super. WTF?

ShaMan123:
I am out

ShaMan123:
`get-defaults` is the branch I was experimenting on

https://github.com/fabricjs/fabric.js/compare/bundle...get-defaults

asturur:
> Some shit is happening in the constructor. Is that what you meant when stating that the prototype chain broke somehow? I set defaults in Object, but then Circle has no default values set even though it calls super. WTF?

Yes.
So that is why i m doing a step back and just removing header and not touching anything else.
Defintely something weird is going on, and i prefer a small progress and investigation later.

Most of our defaults can probably go on the public property of class, there are really a few that i care to be tweakable and shared between classes and we can probably find ad hoc solution for those.

I will make a list, maybe we can store them in config

src/shapes/text.class.ts
asturur: this is not true anymore this was fixed long time ago in node-canvas

asturur: https://github.com/Automattic/node-canvas/issues/1620

asturur:
Need to stop now but at this point next move is to merge https://github.com/fabricjs/fabric.js/pull/8487/files and then continue this one

asturur:
So one big difference seems that the official ts plugin does not support browserlist. so it is not transpiling anything and we are emitting radically different code.
On the other hand currently in master seems to me we are transpiling also spred operators on objects.
I m a bit lost, i will let go for one day or two and the go to some deep compare.


Those are the diffs between master built and rollup 3 + plugins built.
[textfile.txt](https://github.com/fabricjs/fabric.js/files/10329414/textfile.txt)

ShaMan123:
Did you try to configure the browserlist option in the ts plugin? Same as I referenced the tsconfig file there is an optikn to point to the briwserlist file

asturur:
no i didn't try, it balooened out of scope of this PR for me. I have opened a new one for the dependency updates only and you can freely experiment there if you manage to get the latest rollup + latest official plugins to work correctly.
But just a pure dependency bump.

I'll just finish header removal and take a day off from fabricJS

asturur:
Ok i still need to make a real compare between this export and old export and see what got lost, but this is ready to go.
All changes that were part of the previous tenative of using exports or rationalizing things have been removed.
There is only header and `fabric.` removal and the stopgap getEnv.

getEnv or config are somehow related.
We can argue document and window are part of the env, but so is the retina scaling value.
FilterBackend is probably part of the env or of the config, we could make cases that make both seem appropriate.

@ShaMan123 give it another check.


I didn't remove any utils, not even the internal one. 
With the current test framework removing a util from the export means not testing it.

ShaMan123:
> We can argue document and window are part of the env, but so is the retina scaling value.

`devicePixelRatio` is definitely part of env. I put it in config cause there was no better option at the time. But it is not configurable, it is a device preset, so it belongs to env.

FilterBackend should be configurable. But maybe it should be standalone and have a setter of its own. Not sure.

index.js
```diff
@@ -1,57 +1,439 @@
...
+if (typeof exports !== 'undefined') {
+  exports.fabric = fabric;
+} else if (typeof define === 'function' && define.amd) {
+  /* _AMD_START_ */
+  define([], function () {
+    return fabric;
+  });
+} else if (typeof window !== 'undefined') {
+  window.fabric = fabric;
+}
```
ShaMan123, line 439: what about this?
next PR?
this is handled by the umd output in rollup

src/env.ts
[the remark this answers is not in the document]
ShaMan123: Good point.
So we are talking about drp right?
If we move it to env will that be enough to do a lazy init?

src/filters/FilterBackend.ts
```diff
@@ -8,6 +7,11 @@ export type FilterBackend = WebGLFilterBackend | Canvas2dFilterBackend;
 
 let filterBackend: FilterBackend;
 
+/**
+ * Verifies if it is possible to initialize webgl or fallback on a canvas2d filtering backend
+ * @param [strict] pass `true` to create the backend if it wasn't created yet (default behavior),
```
ShaMan123, line 12: no strict param here

src/mixins/stateful.mixin.ts
```diff
@@ -1,115 +1,97 @@
...
-        if (!this[destination]) {
-          return this.setupState(options);
-        }
-        saveProps(this, destination, this[propertySet]);
-        if (options && options.stateProperties) {
-          saveProps(this, destination, options.stateProperties);
-        }
-        return this;
-      },
+export class statefulMixin {
```
ShaMan123, line 54: rename to `StatefulMixin`

src/shapes/Object/StackedObject.ts
```diff
@@ -49,7 +48,8 @@ export class StackedObject<
     while (parent) {
       if (target === parent) {
         return true;
-      } else if (parent instanceof fabric.StaticCanvas) {
+        //@TODO ideally find a different way to identify a canvas without importing it
```
ShaMan123, line 51: I don't think this is a big deal.
Canvas is imported in every app so why does it matter?

ShaMan123:
```suggestion

```

ShaMan123 (commented):
IMO env should be a class with an exported instance, same as config for the following reasons:
- test stubbing/mocking is easy (no need for `setEnvForTests`, an object is better to export in this case than multiple exports)
- setting up custom env can done easily by the dev (subclassing) if we expose some setters
- internal state (is initialized etc.). We will probably need that when tackling SSR and node env.
- code styling


Shouldn't the index file be `.ts`? And shouldn't it reside under `src`?

All in all looks fine

ShaMan123:
> also 16k of unminified code disapparead means i forgot something.
> 
> 400 lines of code are missing. and 4k bytes.

All the index files

index.js
asturur: We have to fix exports.
Then this will go away automagically i hope.
Next PR or next 2, or next 3, i m not sure how hard it is.
That issue we had with prototype breaking and i m pointing the finger to different transpiling could be a pandora box of issues

src/shapes/Object/StackedObject.ts
asturur: It does not matter 100%.
But for example i have an use case in which i want to just produce text  pngs and i can do it without a canvas instance.
It would be great if fabric was that flexible.

> IMO env should be a class with an exported instance, same as config for the following reasons:
> 
> * test stubbing/mocking is easy (no need for `setEnvForTests`, an object is better to export in this case than multiple exports)
the reason why i added multiple exports is because fabric.getDocumnet is nicere than fabric.getEnv().document.
I think that getEnv() is a stop gap solution and until we make bundling work we shouldn't think we either need it or we need a different one.

i think the order of priority should be:
bundling > removing ts-nochecks > node vs browser imports > those details here.

asturur:
> > also 16k of unminified code disapparead means i forgot something.
> > 400 lines of code are missing. and 4k bytes.
> 
> All the index files
> 
> `index.js` could be slimmer if we use the removed index files which would have levelled the count

The removed index files didn't get in the final build ever, unless you would use them as objects ( for example easing did, but wasn't 400 lines ).
Something else is missing and will pop out as something untested.
Canvas.fxRemove and fxStraighten is one piece, but is not all.

Issue #8359, opened 2022-10-11
Polyline/polygon editing breaks when flipped


graphics-et-al:
## Version

5.2.4 (also affects earlier versions)

## Test Case

-  [https://codepen.io/graphics-et-al/pen/JjvwNgE](https://codepen.io/graphics-et-al/pen/JjvwNgE)

## Information about environment

Browser: Chrome, Firefox

## Steps To Reproduce

1. Use the code at [http://fabricjs.com/custom-controls-polygon](http://fabricjs.com/custom-controls-polygon) or the fabric-extensions library at [https://github.com/yassilah/fabric-extensions ](https://github.com/yassilah/fabric-extensions )to make an editable polygon or poyline

2. Set either flipX or flipY to 'true' to flip the object
3. Enable polyline editing, drag a point

## Expected Behavior
Dragging point to modify the editable polygon/line

## Actual Behavior
Madness- object randomly repositions in odd ways

ShaMan123:
Will be addressed by complementary work of #8344

ShaMan123:
I guess you never tried skewing.

ShaMan123:
I guess we will make polygon controls part of the source code

ShaMan123:
duplicate #6886

graphics-et-al:
Thanks @ShaMan123 ! Look forward to the results.

If you're interested: I have developed editable curves: see [https://forum.dev.graphicsetal.com/fabricjs/curve.html](https://forum.dev.graphicsetal.com/fabricjs/curve.html)

The limitations in this bug still apply :(

Source is open, have a look :)

ShaMan123:
> If you're interested: I have developed editable curves: see https://forum.dev.graphicsetal.com/fabricjs/curve.html

Great stuff.
Consider contributing to upstream. I think it can be part of it.

PR #8556, opened 2023-01-01
Poly custom control

luizzappa:
## Description

Custom poly controls as part of fabric.js core. 

Fixes:
- Poly with `skew`
- Poly with `flip`
- Poly with `strokeUniform`

Closes:
https://github.com/fabricjs/fabric.js/issues/8359

## Gist

```
  const poly = canvas.getActiveObject();
  poly.controls = fabric.controlsUtils.PolyControl.createPolyControls(
    poly.points.length, 
    { cursorStyle: 'crosshair' } //pass custom options
  );
  canvas.requestRenderAll();
```

## In Action

~~https://codesandbox.io/s/custom-control-poly-95y2im~~

https://codesandbox.io/s/custom-control-poly-fix-flip-95y2im


luizzappa:
@ShaMan123 , I was in doubt of how to expose the `PolyControl` class. I put it inside `fabric.controlsUtils`

asturur:
What is the reason to have them as a class rather than a bunch of functions to construct a normal control with? Do they need to be different? if yes, why?

luizzappa:
@asturur, I used class because I needed a new property (`pointIndex`) on the control to be able to relate a control to the polygon point (so that we can change the position of the point according to the mouse position).

But I think if I used factories and closure for that it would work too and we could just go with functions instead of class (I don't know if it has an impact on performance):

```
function factoryPolyPositionHandler(pointIndex) {
  return function polyPositionHandler(dim, finalMatrix, fabricObject) {
           //etc
  }  
}

function factoryPolyActionHandler(pointIndex) {
  return function polyActionHandler(eventData, transform,  x, y) {
           //etc
  }  
}
```

If it makes more sense, I can refactor.

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+    return unskewedPoint;
+  }
+
+  // This function locate the controls.
+  // It'll be used both for drawing and for interaction
+  positionHandler(dim: Point, finalMatrix: TMat2D, polyObject: Polyline) {
+    const x = polyObject.points[this.pointIndex].x - polyObject.pathOffset.x,
+      y = polyObject.points[this.pointIndex].y - polyObject.pathOffset.y;
+    return transformPoint(
+      { x: x, y: y },
```
ShaMan123, line 39:
```suggestion
    return new Point(x, y) .transform(
```

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+      multiplyTransformMatrices(
+        polyObject.canvas?.viewportTransform ?? ([1, 0, 0, 1, 0, 0] as TMat2D),
```
ShaMan123, line 41:
```suggestion
        polyObject.canvas?.viewportTransform ?? iMatrix,
```

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+        polyObject.calcTransformMatrix()
+      )
+    );
+  }
+
+  // This function define what the control does.
```
ShaMan123, line 47:
```suggestion
  // This function defines what the control does.
```

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+  // It'll be called on every mouse move after a control has been clicked and is being dragged.
+  // The function receive as argument the mouse event, the current trasnform object
```
ShaMan123, line 49:
```suggestion
  // The function receives as argument the mouse event, the current transform object
```

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+    const poly = transform.target as Polyline,
+      currentControl = poly.controls[poly.__corner] as PolyControl,
+      mouseLocalPosition = normalizePoint(
+        poly,
+        new Point(x, y),
+        'center',
+        'center'
+      ),
+      polygonBaseSize = new Point(poly.width, poly.height),
+      size = poly._getTransformedDimensions({}),
```
ShaMan123, line 67:
```suggestion
      size = poly._getTransformedDimensions(),
```

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+  // Keep the polygon in the same position when we change its `width`/`height`/`top`/`left`
+  static anchorWrapper(anchorIndex: number, fn: TransformActionHandler) {
+    return function (
+      eventData: TPointerEvent,
+      transform: Transform,
+      x: number,
+      y: number
+    ) {
+      const poly = transform.target as Polyline,
+        absolutePoint = transformPoint(
```
ShaMan123, line 97: drop all `transformPoint` in favor of `Point#transform`

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+          new Point(
+            poly.points[anchorIndex].x - poly.pathOffset.x,
+            poly.points[anchorIndex].y - poly.pathOffset.y
+          ),
+          poly.calcTransformMatrix()
+        ),
+        actionPerformed = fn(eventData, transform, x, y),
+        polygonBaseSize = poly._getNonTransformedDimensions(),
```
ShaMan123, line 105: inconsistent with line 66

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+  // transform.target is a reference to the current object being transformed,
+  static polyActionHandler(
+    eventData: TPointerEvent,
+    transform: Transform,
+    x: number,
+    y: number
+  ) {
+    const poly = transform.target as Polyline,
+      currentControl = poly.controls[poly.__corner] as PolyControl,
+      mouseLocalPosition = normalizePoint(
```
ShaMan123, line 60: I think is best to use `getLocalPointer`

ShaMan123 (changes requested):
nice and clean.
I would rename the file to `PolyControl.ts`. We will eventually rename all I presume.
Use this syntax for method comments:
```ts
/**
 * comment
 */
```

We discussed making controls as classes...
I will voice a few more points.
Since poly controls are very defined I don't see why not.
It is not like others that you mix and match.
I prefer creating a base class with all the wrappers and having the class lifecycle invoke them (since there is a generic lifecycle to all controls).
Then it is a question of subclassing specific parts instead of constructing and nesting wrappers.
Everything is moving to classes, why not controls as well?

ShaMan123:
flip needs to be fixed

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+        absolutePoint,
+        newPosition.x + 0.5,
+        newPosition.y + 0.5
+      );
+      return actionPerformed;
+    };
+  }
+
+  static createPolyControls(
+    controlPoints: Array<Point>,
```
ShaMan123, line 129: truth be told, the points aren't being used so the number of vertexes is enough

src/controls/polycontrol.class.ts
```diff
@@ -0,0 +1,149 @@
...
+    options?: Partial<Control>
+  ) {
+    const lastControl = controlPoints.length - 1,
+      controls = controlPoints.reduce((acc, point, index) => {
+        acc['p' + index] = new PolyControl(
+          {
+            actionName: 'modifyPolygon',
+            actionHandler: PolyControl.anchorWrapper(
+              index > 0 ? index - 1 : lastControl,
```
ShaMan123, line 138: this can use modulu in such a case
`(index - 1) % numVertexes

src/controls/polycontrol.class.ts
luizzappa: Typescript complains because options is not an optional parameter in `polyline#_getTransformedDimensoins()`

https://github.com/fabricjs/fabric.js/blob/720a984e0cb5e08a56574a87e09ce6de30b3b485/src/shapes/polyline.class.ts#L153

Maybe it's better to leave it optional in this PR: https://github.com/fabricjs/fabric.js/pull/8374

luizzappa:
> flip needs to be fixed

https://github.com/fabricjs/fabric.js/pull/8556/commits/f1ab52c03ff00b5ce56bd18d57492fc1c373f0fa

luizzappa:
> And we should add tests.

I'll work on it

> I would expose this as part of `src/controls/default_controls.ts` as same as the rest in that file.

it's not clear to me how to expose inside default_controls.ts.. Do we create an object in the fabric namespace?

luizzappa:
Are you guys updating fabric.js demos ( http://fabricjs.com/custom-controls-polygon )? Although it is now quite simple, perhaps it is worth it in terms of visibility.

ShaMan123:
The website is considered deprecated and even dead.
We are going to rewrite it from scratch after v6 stabilizes.
Welcome to join to any effort you'd like.

src/controls/polycontrol.class.ts
ShaMan123: yes

src/controls/PolyControl.ts
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * This function locates the controls.
+   * It'll be used both for drawing and for interaction.
+   */
+  positionHandler(dim: Point, finalMatrix: TMat2D, polyObject: Polyline) {
+    const x = polyObject.points[this.pointIndex].x - polyObject.pathOffset.x,
+      y = polyObject.points[this.pointIndex].y - polyObject.pathOffset.y;
+    return new Point(x, y).transform(
+      multiplyTransformMatrices(
+        polyObject.canvas?.viewportTransform ?? iMatrix,
```
ShaMan123, line 48: canvas is always defined in when running a control action
so I guess `polyObject.canvas!.viewportTransform` is enough.
but perhaps a better fix would be to assert the type in the control class

src/controls/PolyControl.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  static createPolyControls(numOfControls: number, options?: Partial<Control>) {
+    const lastControl = numOfControls - 1,
+      controls = {} as Record<string, PolyControl>;
+    for (let idx = 0; idx < numOfControls; idx++) {
+      controls[`p${idx}`] = new PolyControl(
+        {
+          actionHandler: PolyControl.anchorWrapper(
+            idx > 0 ? idx - 1 : lastControl,
```
ShaMan123, line 143: I gave this a bit of thought.
All the other vertexes remain in place so it doesn't really matter which one we anchor to, right?
If so we don't really need to pass this information and make `polyActionHandler` decide to which point to anchor by itself.
Furthermore, it doesn't depend on the number of vertexes...

src/controls/PolyControl.ts
ShaMan123: take a look d9195a406

ShaMan123: https://codesandbox.io/s/custom-control-poly-fix-flip-forked-3k408w?file=/src/index.js

ShaMan123 (commented):
OK
This is good.

src/controls/PolyControl.ts
ShaMan123: take a look d9195a406
https://codesandbox.io/s/custom-control-poly-fix-flip-forked-3k408w?file=/src/index.js

ShaMan123 (approved):

asturur:
@luizzappa i understand the class seems cleaner, but the current interface has a single class and adds action handler, style handler, and what not.
The original demo indeed was working with that interface ( ugly and fastly written, that is true ).
We are not going to change this interface for 6.0, and if you need an extra property on the point when initializing it, you can still add it when you do `new Control({...})` since it will assign whatever you pass to this.

So if we want to have the polygon points embedded in the npm package with tests rather than having to copy it from the demo, we need to do it with the current interface.

asturur:
The website is for the version 5, the current version, on which no one is working right now.
Updating a demo to look less broken is nice, but that would require fix the demo for the v5 version of the polygon.

src/controls/PolyControl.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  static getSize(poly: Polyline) {
+    return new Point(poly.width, poly.height);
+  }
+
+  /**
+   * This function locates the controls.
+   * It'll be used both for drawing and for interaction.
+   */
+  positionHandler(dim: Point, finalMatrix: TMat2D, polyObject: Polyline) {
```
asturur, line 43: is there a reason why finalMatrix is not good enough?

asturur (changes requested):
I want to be clear using the request for changes that if we want to have the demo example officialized in the repo, we need to make it boil down to the action handler and the special anchor point wrapper.

Regarding skewed polygon support, that seems too much to me.
There is no reason to skew a polygon since you can skew the points and have a normal polygon after you normalize the skew in the points coordinats.

Is better to have simpler code rather trying to handle edge cases that could be neutralized in a different way. Polygon editing is not a fabricJS feature so we don't have to support all the use cases of the polygons.
Otherwise we have to get into things like:
- gradient fill, how does change the offsets of the colors when i change the points
- pattern fill, same as above
- editing of polygons inside containers with random transforms applied

ShaMan123:
If no classes then we could pass `pointIndex` as state, the same as we do with scale/skew control logic. We pass data as part of the transform object or as an additional arg. It doesn't have to be on the instance. That is the pattern.


> Is better to have simpler code rather trying to handle edge cases that could be neutralized in a different way. Polygon editing is not a fabricJS feature so we don't have to support all the use cases of the polygons. Otherwise we have to get into things like:
> 
> * gradient fill, how does change the offsets of the colors when i change the points
> * pattern fill, same as above
> * editing of polygons inside containers with random transforms applied

I disagree.
If we expose something it should work out of the box. If not, it should be an unofficial example.
A lot of what I have/had against fabric is that exactly. The expectation IMO is what I have stated.

Regarding gradient and pattern there could come a feature PR that exposes wrappers to handle that, same as `wrapWithFixedAnchor`. 
Regarding editing a polygon under group that **must be fixed**.

src/controls/PolyControl.ts
ShaMan123: I agree, finalMatrix should be used since it accounts for group

ShaMan123:
fixing editing under group might be blocked by #8374

asturur:
The point of the control api was exactly to give a tool to developers to build what they need for custom interactions.
Not a tool to make us build and maintain everything that could be needed by any app.
Is just a matter of scope.

I agree that the transform data should carry every bit that can be useful to have context during the handler, i m not sure we can capture the index there, but again if you create a control and pass in the constructor the pointIndex you will have it available.

src/controls/PolyControl.ts
luizzappa: well noted, nice!

src/controls/PolyControl.ts
luizzappa: I didn't understand very well what `finalMatrix` represents (only considers the `angle`of the object(?) ), but it is not the multiplication of the transformation matrix of the viewport and the object (what we need to correctly render the position of the control points)

Source of `finalMatrix`:

https://github.com/fabricjs/fabric.js/blob/b6f076cc0c817cd7f70abc7028c9d04295d121a8/src/shapes/Object/InteractiveObject.ts#L170-L197

luizzappa:
> Is better to have simpler code rather trying to handle edge cases that could be neutralized in a different way. Polygon editing is not a fabricJS feature so we don't have to support all the use cases of the polygons. Otherwise we have to get into things like:
> 
> * gradient fill, how does change the offsets of the colors when i change the points
> * pattern fill, same as above
> * editing of polygons inside containers with random transforms applied

In my opinion, if this is the driver, it is better to leave it as an example of use than in the core of the library. At least in the example there is a more detailed guide that allows the developer to understand and adapt to the needs of his application.

> I agree that the transform data should carry every bit that can be useful to have context during the handler, i m not sure we can capture the index there, but again if you create a control and pass in the constructor the pointIndex you will have it available.

Typescript complains that there is no `pointIndex` property in the Control interface.

I ended up passing `pointIndex` inside `transform`.

> Regarding editing a polygon under group that **must be fixed**.

It was supposed to work within a group as well, since `calcTransformMatrix()` calculates the final matrix of all planes when we are within groups. But it's still buggy, the problem seems to me to be with `qrDecompose` that generates a transformation matrix different from the one we calculated, it messes up all the points.

ShaMan123:
updated from master
I would like the file to be renamed to `poly` now that it does not contain a poly control.

src/controls/PolyControl.ts
ShaMan123: It transforms to the center of the target.
`dim` is the size in canvas plane.
But you are right, using it doesn't make sense in this context

ShaMan123:
a6fc8dd fixes group
skew logic should be restored if not since it is a fix (which is more than enough... it is absurd IMO needing to defend this) then because it is a very complex fix and important knowledge virtually no one could reach (fact is no one did for years and issues were submitted).

ShaMan123:
And not fixing skew, why fix flip or group?

luizzappa:
> [a6fc8dd](https://github.com/fabricjs/fabric.js/commit/a6fc8ddaec1c6eb09bf927b3d0caf666ce4e3164) fixes logic under group

nice! 

Is there any method to recalculate the BBOX of a group? We need to invoke it here:

https://github.com/fabricjs/fabric.js/blob/a6fc8ddaec1c6eb09bf927b3d0caf666ce4e3164/src/controls/polyControl.ts#L64

It seems to me that this behavior is because the BBOX is not updated:


bug group2

ShaMan123:
Yes there is but this should be handled by group...
I don't understand why it isn't.
The modified event should fire and inform group the perform layout.

https://github.com/fabricjs/fabric.js/blob/818a134cb750049d0e1a8013d47a8e90c2eb7d99/src/shapes/group.class.ts#L348

https://github.com/fabricjs/fabric.js/blob/818a134cb750049d0e1a8013d47a8e90c2eb7d99/src/shapes/group.class.ts#L315-L318


The imperative method is `triggerLayout` but this logic should not invoke it.
Does the modify event trigger at the end of the interaction?

ShaMan123:
Works as expected.
Group and active objects relationship isn't yet defined properly.
It might not render the active objects etc.

https://user-images.githubusercontent.com/34343793/211140341-0f46c0bb-95cd-4bbe-b36b-74265e4b4d26.mp4

So for now you should cancel caching while modifying

luizzappa:
> So for now you should cancel caching while modifying

Ok, disabling `objectCaching` worked, but how did you recalculate the BBOX of the group. Here it continues with the initial BBOX:

bug group3

https://codesandbox.io/s/custom-control-poly-fix-group-2bv3hi?file=/src/index.js

off topic:
This BBOX border of any polygon inside a group is incorrect.. I'll adjust it in another PR.

image

ShaMan123:
It doesn't work simply because group has to be initialized as follows:
```ts
new fabric.Group(objects, { subTargetCheck: true, interactive: true })
```
I am surprised it worked at all. Ohh it is beause you set the active object. If you comment that line you won't be able to touch the polygon nested in the group.

Regarding the bbox. Didn't we discuss it in the tests PR? Wasn't it solved? I forgot.

luizzappa:
> Regarding the bbox. Didn't we discuss it in the tests PR? Wasn't it solved? I forgot.

No.. Actually that PR is blocking this fix. I won't include it there so as not to pollute it even more.

ShaMan123:
It should be simple...
Telling group to layout after setDimensions.
Same as what happens in Object#_set

asturur:
>  Everything is moving to classes, why not controls as well?

I get asked the same question over and over again. We can't change all. Migrations have costs.

Can this feature works with the old interface?
Is that really bad that can't be done or is so bad code that is not worth it?

Yes everything is moving to classes and is being painful, it is taking forever, a lot of old code is swapped with new code that is supposed to do the same things but the code has to be different. Is a zero sum game that is not fun to play.

asturur:
ok, had to force fabric object on the window for some reason on codesandbox.

It works great, it sticks to the current pattern, i m merging it.

asturur (approved):

ShaMan123:
@asturur please reconsider skewing!!2@!!

biggestdickus:
Maybe this might give you some ideas:
https://jsfiddle.net/RobertWStewart/pyqjrfa4/
2D polygon flip, rotate, resize, whatever by just redrawing the polygon.

PR #8557, opened 2023-01-01
Updated rollup dependencies

asturur:
## Motivation

This update seems to have no side effects on the build
Just a reference branch to experiment with using exports


package.json
```diff
@@ -103,9 +103,7 @@
     "prettier": "2.7.1",
     "ps-list": "^8.1.0",
     "qunit": "^2.17.2",
-    "rollup": "^2.75.6",
-    "rollup-plugin-size-snapshot": "^0.12.0",
-    "rollup-plugin-terser": "^7.0.2",
+    "rollup": "^3.9.0",
     "rollup-plugin-ts": "^3.0.2",
```
ShaMan123, line 107: what about the ts plugin?

ShaMan123: I saw you added and then removed it...
You must update typescript as well I believe

ShaMan123: Ohh I see that is the cause of the major fuck up in classes

asturur:
Should i continue on this PR or move to more TS finishing?

ShaMan123:
whatever you think
we should manage to work with rollup, we are currently hacking it

asturur:
I think i understood the code change issue.
When we do in typescripy

```
Class A {
  propertyB: number;
}
```

without giving it a full value, either that gets removed from the traspiler, or it gets initialized to void.

Initializing to void override whatever is on the prototype.

Using `declare` solve the issue, but the older plugin somehow knew how to handle it on its own probably because it deferred to TS for JS emitting.

I think the safe approach, while boring, is to add `declare` in front of those, because that is what we are actually doing, declaring that the value will be there.

asturur:
On top of that i don't like our declarations are emitted right now.
typescript isn't bundling the code of course, so is not emittin a single declaration file but many of them ( one per file ) and i don't see anything understanding that once the file is bundled.

.browserslistrc
```diff
@@ -1,3 +1,2 @@
 # Browsers that we support
-
-chrome >= 58
+chrome >= 100
```
asturur, line 2: just trying around, not supposed to change this file

index.ts
```diff
@@ -427,13 +427,6 @@ const fabric = {
...
-  /* _AMD_START_ */
-  define([], function () {
-    return fabric;
-  });
-} else if (typeof window !== 'undefined') {
-  window.fabric = fabric;
-}
+export default fabric;
+
+export { fabric };
```
asturur, line 432: i m really not sure all this mess is worth for this change tho.

package.json
```diff
@@ -76,8 +76,15 @@
...
-    "@rollup/plugin-json": "^4.1.0",
+    "@babel/plugin-transform-typescript": "^7.20.7",
+    "@babel/preset-env": "^7.20.2",
+    "@babel/preset-typescript": "^7.18.6",
+    "@rollup/plugin-babel": "^6.0.3",
+    "@rollup/plugin-json": "^6.0.0",
+    "@rollup/plugin-terser": "^0.2.1",
+    "@rollup/plugin-typescript": "^10.0.1",
+    "@swc/cli": "^0.1.59",
+    "@swc/core": "^1.3.24",
```
asturur, line 87: we are not keeping all of those, searching for a simple subset

ShaMan123:
A different approach:
Since all browsers support the module type script tag and allow imports why should we mess around eith building.
Just TS and be done

asturur:
Can you link me to the module script tag specs? i don't understand which feature you refer to.

asturur:
For later reference for me:
https://medium.com/@martin_hotell/typescript-library-tips-rollup-your-types-995153cc81c7

ShaMan123:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

asturur:
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules

Oh ok the MJS browser modules.
I m **NOT** gonna pioneer that thing.
is unclear how those files get transferred? are those multiple network connections?
is unclear how do you get the files at all with NPM et similars.
is unclear if devs need to repack them anyway because they have to bundle
What happens if the network of dependencies becomes a huge mess?

I m also not ready to support it. If someone makes me a question i have to go and study.

I want to do canvas and drawing not pioneer js stuff.
Babel and TS are already a stretch for me

The old string concatenation build was awesome now that i look back.
Fast. modular. worked. lasted 12 years.

Issue #3386, opened 2016-10-29
IText character background producing unwanted lines when created with small font size


FlemmingH:
Hi
If i create an IText object with a small fontsize and semitransparent background for each character and then scale it manually up, the width of each character-background overlaps and with the added opacity a line is getting vissible! I suspect this to be an error in fabric.js !?

I started this question on [Stackoverflow](http://stackoverflow.com/questions/40214168/fabricjs-itext-overlapping-character-background-color-making-lines-when-semitra). 
Here's a [fiidle](https://jsfiddle.net/FlemmingH/uq85Lf3j/). Try scale it up manually.

Issue #3378, opened 2016-10-27
Select and delete text in Textbox object does not clear the selection


prokilogrammer:
## Version
1.6.5

## Test Case
http://jsfiddle.net/1azfo9t6/

## Steps to reproduce
1. Create a canvas with `selection = false`
2. Initialize a Textbox object
3. Write text that wraps to multiple lines
4. Select all the text inside the box
5. Press "backspace" or "delete" to delete all the text

## Expected Behavior
All the text gets deleted AND selection also gets cleared

## Actual Behavior
All the text gets deleted BUT selection does NOT get cleared

asturur:
oh ok. i see the bug :)

prokilogrammer:
Do you have a fix in mind? I tried taking a stab at it but didn't get too far. Looks like oldWidth and oldHeight of IText are being set to the shrunken box dimensions after removing characters - https://github.com/kangax/fabric.js/blob/master/src/shapes/itext.class.js#L314-L321. I tried setting oldWidth/oldHeight in `removeChars()` (https://github.com/kangax/fabric.js/blob/master/src/mixins/itext_key_behavior.mixin.js#L606) but no effect.

I would be happy to send a PR if you guide me in the right direction.

asturur:
Yes the fix is clear the right part of the screen. i m heavely working on text those days for other reason, if i step into it, i will drop a notice here.

PR #8560, opened 2023-01-02
fix(IText): **NO** clear context top during rendering

ShaMan123:
## Motivation

There is an edge case where while using context top (free drawing) and a render occurs text instances render and clear their bbox from context top.
They shouldn't
Take a look at the video ending in a mouse up.

https://user-images.githubusercontent.com/34343793/210237649-65b58d80-6bb1-40fe-a562-82315656e0e1.mp4

## Description

`clearContextTop` suffers from a built in problem.
It clears the context with the known width/height values.
Efforts were made to overcome that (quite a lot it seems) and most of cases (distinct case no. 1, see below) were indeed fixed but not all (e.g. transforming). It did cause a need to track every case and clear the context before the change propagates to the instance, which is folly. More interaction, more cases, a mess.

There are 2 distinct cases in which we should consider how to act:
1. changing selection only (which is not part of the rendering cycle due to perf, imperative in sight of context state)
2. changing visuals (part of rendering, declarative in sight of context state)

Seems that efforts were focused on clearing context for the first case but the second case was neglected (rendering, transforming and many others). It can be understood since the first case was/is the main dev effort for making itext work. The second case is an integration thing, app level.

## Changes

removed clearing context top:
- vpt change
- size change: `initDimensions`

## Gist

The render method calls `renderCursorOrSelection` that runs only if in editing mode.
If executed it marks top context as dirty for the canvas rendering cycle.

Looking at what brush does and applying it to itext:
If the entire canvas is rendered it is 100% proof since the top context will be cleared by the canvas and then selection will be drawn by the instance that is being edited.
If we mess with the context w/o rendering canvas to handle selection as we do with key/input events then we must clear context top as part of logic and we must take into account any changes to size of instance and we should mark top context as dirty as well.
So the rule of thumb should be if we mess with selection alone then we call `clearContextTop`.

`initDimensions` is a good example for the confusion around this. The context was cleared before changes to size were committed to avoid leaving traces, but even though it was not 100% proof. Since canvas must be rendered in order for the new size to be visible it is a pure no. 2 distinct case so we can safely hook to the rendering cycle.


src/shapes/itext.class.ts
```diff
@@ -273,7 +273,7 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
    * @param {CanvasRenderingContext2D} ctx Context to render on
    */
   render(ctx: CanvasRenderingContext2D) {
-    this.clearContextTop();
+    // this.isEditing && this.clearContextTop();
```
ShaMan123, line 276: Thoguhts?

asturur: Let's use blame and lets' see when it was introduced and in which context.

asturur: image

6 years ago....

asturur: Linked to those 2 issues:
https://github.com/fabricjs/fabric.js/issues/3378
and

In the PR it was linked to this clearContextTop
https://github.com/fabricjs/fabric.js/pull/3388/files#diff-b8bf422caf940e8585d2c4dac83f8ef76ceb2a1c0ffbf986202492dad12a7cfaR335-R348
That was already safeguarded for isEditing and this.active ( that does not exist anymore )

asturur: in this PR here 
https://github.com/fabricjs/fabric.js/pull/7802/files#diff-b8bf422caf940e8585d2c4dac83f8ef76ceb2a1c0ffbf986202492dad12a7cfaL281-L291
you removed clearContextTop from iText and you added it to generic object, but you lost the isEditing boolean.
So i would argue that the isEditing boolean should come back. no?

ShaMan123: I am to blame indeed.
Why did I move it there in the first place? Does it make sense??
#3386 seems unrelated.
#3378 is, however the bug wasn't solved for all cases. I will add tests proving that.
I think I understand my confusion now. 
https://github.com/fabricjs/fabric.js/blame/d6f9c8f98813717372c1510193edae123dbfba11/src/shapes/itext.class.js#L303-L320

src/shapes/itext.class.ts
```diff
@@ -220,8 +220,10 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
    * @private
    */
   initDimensions() {
-    this.isEditing && this.initDelayedCursor();
-    this.clearContextTop();
+    if (this.isEditing) {
+      this.initDelayedCursor();
+      this.clearContextTop();
```
ShaMan123, line 225: here

src/shapes/itext.class.ts
```diff
@@ -406,6 +398,16 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
     return boundaries;
   }
 
+  /**
+   * Renders cursor on context Top, outside the animation cycle, on request
+   * Used for the drag/drop effect.
+   * If contextTop is not available, do nothing.
+   */
+  renderCursorAt(selectionStart: number) {
```
ShaMan123, line 406: moved to be part of scope

src/shapes/textbox.class.ts
```diff
@@ -51,8 +51,10 @@ export class Textbox extends IText {
     if (!this.initialized) {
       return;
     }
-    this.isEditing && this.initDelayedCursor();
-    this.clearContextTop();
+    if (this.isEditing) {
+      this.initDelayedCursor();
+      this.clearContextTop();
```
ShaMan123, line 56: and here

src/shapes/itext.class.ts
ShaMan123, line 406: moved to its siblings in scope

src/canvas/canvas.class.ts
```diff
@@ -1631,17 +1631,6 @@ export class SelectableCanvas<
...
     instance.set(originalProperties);
   }
-
-  setViewportTransform(vpt: TMat2D) {
-    if (
-      this.renderOnAddRemove &&
-      isInteractiveTextObject(this._activeObject) &&
-      this._activeObject.isEditing
-    ) {
-      this._activeObject.clearContextTop();
```
ShaMan123, line 1641: will be handled by `contextTopDirty` from now on

src/shapes/itext.class.ts
asturur: i think it was moved to support dragging and in the process we did not notice the isEditing.
Is enough to add it back i suppose and add a test for isEditing.

src/shapes/itext.class.ts
asturur: the only change that did not require a comment 🤣🤣🤣🤣

asturur:
it become a large change to understand at a glance.
Can you describe it at your best with words?

src/shapes/itext.class.ts
```diff
@@ -272,22 +271,13 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
...
     this.renderCursorOrSelection();
   }
 
-  /**
-   * @private
-   * @param {CanvasRenderingContext2D} ctx Context to render on
-   */
-  _render(ctx: CanvasRenderingContext2D) {
-    super._render(ctx);
-  }
```
asturur, line 289: this was probably here because when it was this.callSuper('_render') it was quite complicated to jump a class that di not have a method and move the ladder by 2 steps

src/shapes/itext.class.ts
ShaMan123: 🤣

ShaMan123 (commented):
Explaining what is going on:
`clearContextTop` suffers from a built in problem.
It clears the context with the known width/height values.
Efforts were made to overcome that (quite a lot it seems) and most of cases were indeed fixed but not all (e.g. transforming). It did cause a need to track every case and clear the context before the change propagates to the instance, which is folly.

The are 2 distinct cases in which we should consider how to act:
1. changing selection only (which is not part of the rendering cycle due to perf, imperative)
2. changing visuals (part of rendering, declarative)

Seems that efforts were focused on clearing context for the first case but the second case was neglected (rendering, transforming and many others).

Is that a good enough explanation?

The only immune way to cases is to mark the context as dirty and let the rendering cycle handle it. Same a brush does.
`initDimensions` is a good example. The context was cleared before changes to size were committed to avoid leaving traces, but even though I have found it not to work I think.

The thing is that imperative rendering is confusing. Some parts must be handled by the instance for perf reasons, same as brush.
But then we must have a mechanism that handles rendering.

ShaMan123:
Another thing that came to me is that `Brush#needFullRender` should have kicked in and rendered the entire path while drawing. Looking into the code, top context is rendered before main context and since text render method clear top context this is what happened

> There is an edge case where while using context top (free drawing) and a render occurs text instances render and clear their bbox from context top. They shouldn't Take a look at the video ending in a mouse up.
> 
>  fabric.js.sandbox.-.Google.Chrome.2023-01-02.15-26-25_Trim.mp4

ShaMan123:
updated my explanation a bit

src/shapes/itext.class.ts
```diff
@@ -306,19 +296,10 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
     } else {
       this.renderSelection(ctx, boundaries);
     }
+    this.canvas!.contextTopDirty = true;
```
asturur, line 299: Ok the difference here is that swapping this with the one from our render function here https://github.com/fabricjs/fabric.js/pull/8560/files#diff-876b5f345ecc11628b120c60dfe0b3fde14630cc3940551bcb4ed23f41c580bcL275 we are swapping a partial contextTop clear with a full context clear.

That is fine if it is needed and solves bugs.
Is that the case?

src/mixins/itext_behavior.mixin.ts
```diff
@@ -258,9 +258,7 @@ export abstract class ITextBehaviorMixin<
     this._currentCursorOpacity = 1;
 
     //  make sure we clear context even if instance is not editing
-    if (shouldClear) {
-      this.clearContextTop();
-    }
+    shouldClear && this.clearContextTop();
```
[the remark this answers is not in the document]
ShaMan123: in case it doesn't, u mean?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(IText): **NO** clear context top during rendering #8560
```
asturur, line 5: let's be more clear on what happened here.
let's try to summarize what we changed with what because we are still clearing, just in a different moment of renderAll

ShaMan123: done

src/mixins/itext_behavior.mixin.ts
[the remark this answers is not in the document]
ShaMan123: how? everything is unreadable

src/shapes/itext.class.ts
ShaMan123: The render method calls `renderCursorOrSelection` that runs only if in editing mode.
If the entire canvas is rendered it is 100% proof since the top context will be cleared by the canvas and then selection will be drawn by the instance that is being edited.
If we mess with the context outside w/o rendering canvas as we do with key/input events then we must clear context top as part of logic and we must take into account any changes to size of instance

ShaMan123: I will add this to the decription

src/mixins/itext_behavior.mixin.ts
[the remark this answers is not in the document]
asturur: i think the minifier swap `if (condition) { single line }` for `condition && singleline` automatically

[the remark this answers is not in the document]
asturur: not an issue of this pr but i don't think it make sense to be terse manualy now

ShaMan123:
@asturur I have updated description and I proud of it now.
Tests reveal the fix isn't complete

ShaMan123:
@asturur I am baffled
The goldens I've committed are correct.
But... freak show time.
When testing in my machine so weird shit is going on.
Look
image


I was thinking it is retina scaling, but I am not sure where the fuck it is

ShaMan123:
So I am fed up with this shit and in reality I fixed it 🤬😡😖😱.
Please help 😩😔😵

ShaMan123:
Found the regression causing this. see #8520

ShaMan123:
after merging #8520 tests pass!

asturur:
Sorry i m a bit busy today but i m free tomorrow.
I ll try to see if i understand what is going on.

asturur:
Merged the retina fix.

asturur:
Merging now, great to have added tests.

I had to revert the pure stylistic change i couldn't help myself.

asturur (approved):

PR #8565, opened 2023-01-04
fix(): `_initRetinaScaling`  initializaing the scaling regardless of settings in Canvas.

ShaMan123:
## Description

fixes the regression from #8510 
I decided to open this PR to have a pure scope for this fix cause I guess that is what you want, instead of fixing as part of #8520 

Decide if you want to close this and merge #8520 or merge this and then that. Anyways if you do the second when dealing with conflict #8520 should win.

## Changes

now `_initRetinaScaling` should be called only if `isRetinaScaling` (doesn't check in the method as previously did)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): `_initRetinaScaling` regression
 - fix(): regression of canvas migration with pointer and sendPointToPlane [#8563](https://github.com/fabricjs/fabric.js/pull/8563)
 - chore(TS): Use exports from files to build fabricJS, get rid of HEADER.js [#8549](https://github.com/fabricjs/fabric.js/pull/8549)
 - chore(): rm `fabric.filterBackend` => `getFilterBackend` [#8487](https://github.com/fabricjs/fabric.js/pull/8487)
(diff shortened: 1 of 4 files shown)


asturur:
How the regression was created do we know? the fix you propose seems unrelated to the changes that happend in that pr.
What is the regression actually?

ShaMan123:
If you look at the override of `_initRetinaScaling` you can see it doesn't check if `_isRetinaScaling` though it should, as does the super method in static canvas.
https://github.com/fabricjs/fabric.js/pull/8510/files#r1062365124
https://github.com/fabricjs/fabric.js/pull/8510/files#r1062366689

ShaMan123:
the most simple fix should be 
```
this._isRetinaScaling() && this.__initRetinaScaling(this.upperCanvasEl, this.contextTop); 
```

But it makes more sense to check outside of a method named initRetinaScaling

asturur:
ok now i understand.
That message should have be on the description.

Regarding what makes more sense is usually the smaller change visually.
Imagine if you added the check inside the subclass in the canvas file, that would have explained the issue too.
While with those changes the question was, why are we moving a check from inside out and how that fixes anything.

asturur (dismissed):

asturur (approved):

PR #8573, opened 2023-01-06
BREAKING: Remove statefulCache and stateful

asturur:
## Motivation
While we are not ready to let the stateful mixin go, we agree on removing the stateful support as a feature.
This PR removes Stateful support in order to simplify fabricJS code handling of changes.
On top of that we should optimize always for the common use case.
If you want stateful changes, there are less expensive way to have them compared to brute force saving and comparing of object trees.

## Changes
Object.stateful is no more an option
Object.statefulCache is no more an option

Those were defaulted as fault since long, so if you didn't set them on true on purpose, this shouldn't affect you.


src/shapes/Object/Object.ts
```diff
@@ -2069,35 +2051,19 @@ export const fabricObjectDefaultValues = {
...
     'globalCompositeOperation',
     'shadow',
     'visible',
-    'backgroundColor',
     'skewX',
     'skewY',
-    'fillRule',
-    'paintFirst',
-    'clipPath',
-    'strokeUniform',
```
asturur, line 2100: removed the overlab between stateful properties and cache properties.
Stateful properties can't be deleted since they are still needed to set group as dirty.

ShaMan123 (commented):
nice!
We can remove `layout` from stateProperties of group since group layout bubbles to its parents

ShaMan123:
just realized this PR broke text cursor animation
will PR a fix right away

PR #8574, opened 2023-01-07
chore(TS): Add declare and remove useless methods from Object.

asturur:
## Motivation

Finishing up basic object migration.

## Description
- change the public properties to `declare` so that any transpiler understand those should be removed.
- ~~added the default values in the constructor, while still making possible to change them globally before initialization of objects~~ failed / scope blown up
- ~~cacheProperties and stateProperties needs to stay on prototype, having them as static is not really possible since it would require duplicating them for each class and would add another constrain when subclassing. Apparently static properties aren't usable with inheritance~~ failed / scope blown up
- Fixed as much as possible TS complains, a couple have been ignored
- Object._removeTransformMatrix moved as utility since is used only from the svg parser

## Changes

- removed Object.centerObject and similar. The methods were aliases for canvas counterparts with the added variable that they would be no-op if there wasn't a canvas. Seems jus confusing. Developers should call those methods from the canvas instance instead of relying on the fact that the canvas is maybe there maybe not and the code won't do nothing but won't fail


asturur:
Messed up a bunch of things with cacheProperties i need to fix yet.
adding all the declare everywhere wil let us exports or class syntax normally.

asturur:
This is going to be another painful PR on top of certifying that class are a mistakes and are less powerful and useful than functions. Know i m sure.

asturur:
@ShaMan123 this thing is becoming more complicated than planned.

Using public properties or defaults in the constructor is triggering the set calls. set calls executes side effects in different order depending on the order of the keys of the options and results changes.
We could update tests but it won't be stable anyway since order of iteration on keys is not guaranteed.

I already tried to assign defaults outside of set ( as we were doing before because they were on the prototype ) but that can't be done because they can't be assigned on `this` before the constructor run, and the constructor of the super classes expect them to be set already.

We could stop executing side effects on the constructor entirely but that would mean we would need to find what breaks and fix it manually.

asturur:
Reducing the scope of the pr to unlock exports

ShaMan123:
I have to say this PR seems a step back rather than forward.
Something to discuss.
All this declare mess, did you try to alter `tsconfig#strictPropertyInitialization` beforehand? Is it related? Well, I guess it is easy to revert, a simple search & replace. But a lot of work for a transient step.
I still don't understand where you set the default values in the constructor. Are we still hacking it via the prototype? If so I guess that is what is causing all this mess. Am I right?
Regarding `set`, it is something we discussed and a fault in design as I see it. I read that class properties are initialized in the constructor with `Object#defineProperty` and after that via `Object#set`. That is the spec. And it makes sense. No side effects from the constructor block.
We should align to that. #8372 suggest a way to manage effects but I think we need a robust and simple way such as setter. I know you dislike it, I do too. You suggested proxy. I read about that. I have no idea how to incorporate with the class `new` command (can only think of static initializers which is a turn off).
We are at the finish line which is always the hardest part in projects, right?
We made amazing progress. Let's decide how to manage this. It is a crucial decision.
All in all my point is that we decided to align to modern js but this PR is doing the opposite. 
What do you think?

ShaMan123:
take a look at this POC
https://github.com/fabricjs/fabric.js/compare/init-define-props-poc
Circle and ellipse tests pass.

ShaMan123:
I prefer a chat if you are willing
Let's solve this and plan to end of the migration, tackling remaining concerns.

asturur:
Some premise, adding items to the prototype isn't `hacking`.

declare is needed because otherwise those are public properties, so if you don't assign a value, you need to declare them.
The fact that the current TS plugin is fixing them ( but the new one and babel aren't ) is a hint of something wrong.

Babel transpiler either errors because there is a declare or they are actual undefined properties, but existing.

image

Looks like the above code is valid, so when the code gets transpiled and the type definition removed, those are undefined props, without declare ( and a removal from the transpiler ) the code is not correct.

Putting the defaults in the contructor as i did in the previous commits doesn't work for many classes that have `set` initializators that move up and down between the sub and super classes.

Circle and ellipses works for me too, group and active selection do not.

The reason why i m not handling it in this PR anymore is because every class will need ad hoc changes to work with setting defaults in the constructor, test changes too, so we will need more PRs.

So i m descoping this PR to adding declare only, and limiting the diff to that as much as possible, so at least we can try again to enable exports from the main index.

https://github.com/fabricjs/fabric.js/pull/8574/commits/19aa171c2821c30fb64194997a65da18e6585066 this is the commit where i moved all defaults in constructors and then i strated roll back.

ShaMan123:
> Some premise, adding items to the prototype isn't hacking.

I thought class properties do not belong to the prototype, only to the instance. Methods belong to the prototype.

asturur:
I also separated cache properties and state properties from default values because those will either need to be in some config or stay in the prototype. Their usage is unfit for static, static does not inherit with the subclasses apparently.

ShaMan123:
static field do inherit


```ts
class A {
  static method() {
    return 1;
  }
  static arr = [1,2,3]
  static type = 'A'
}

class B extends A {
  static {
    this.arr.push(4)
    this.type = 'B'
  }
  static method() {
    return super.method() + 2;
  }
}

console.log(B.method(), B.arr, B.type)
```



```ts
"use strict";
var _a;
class A {
    static method() {
        return 1;
    }
}
A.arr = [1, 2, 3];
A.type = 'A';
class B extends A {
    static method() {
        return super.method() + 2;
    }
}
_a = B;
(() => {
    _a.arr.push(4);
    _a.type = 'B';
})();
console.log(B.method(), B.arr, B.type);

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G8BQ1oQC5nQS2NAtgKboAWA9gCYAUAlIiqtAE7ECuTAdtAIwDcDAXwYYsuaGCZNoAXmgBtbgBoATIoDMAXWGYcedAE8ADoRnQA5LDNIhSUJBgAhaIQAe6Qhwox4yVCN30jNCk2BAAdBJMYYasECRUACw0DKgh4QbGpmYOVqhCfjpiRKSUtIGMLOjsXBCsxlHF5NR0ANTQyvx51ki2ZBwQZCCEYSBkAOZUDmGNpTSK0FOR81MZhDRAA)

asturur:
> > Some premise, adding items to the prototype isn't hacking.
> 
> I thought class properties do not belong to the prototype, only to the instance. Methods belong to the prototype.

Sure for pure state properties that sounds a thumb rule to which you can everyone agree with.
That does not work if you don't want to define the same properties along many classes or refactor entirely your code to have it work with this rule.

ShaMan123:
I need an example.
Static init seems to bridge that gap, does it not?

ShaMan123:
BTW in the example the static `arr` property is shared between classes. Pushing 4 to `arr` in B mutated it for A as well

asturur:
They inerhit just on static things.
You can't put cacheProperties on static in a class, and then grab them in a subclass in a convenient way outside of static.

ShaMan123:
We discussed this.
Once a class is initialized state should not be modified by the prototype. Don't you agree?
Is this what we are discussing here?

ShaMan123:
> They inerhit just on static things. You can't put cacheProperties on static in a class, and then grab them in a subclass in a convenient way outside of static.

`cacheProperties` is used in one place in fabric (Object) and `staticProperties` in Group once. The rest are definitions by subclasses. So I don't see it as a problem. Besides we can define a method.
```ts
getCacheProps() {
    return this.constructor.cacheProperties
}
```

ShaMan123:
take a look 50a5242d2
POC success for group and AS

ShaMan123:
I also think type should be static or go away entirely

asturur:
> Once a class is initialized state should not be modified by the prototype. Don't you agree? Is this what we are discussing here?

No, you know that i don't agree on that.
Changing the prototype is fine as an option if you want to do that and if you know what are you doing.
We shouldn't encorauge anyone to do it tho, nor is necessary to block that.
We decided to block the functionality because in team discussion that was seen as a drive for happyness, so we are trying to get there.

Having the cache properties and defaults on the static or in config, is indeed a way to get people to still modify class configuration without the need to go in subclassing just to change for example the default value of the corner shape or the border thickness.

Also what are we discussing here?

The PR is descoped to just declare, and we are discussing why my attempt to put all properties on the instance, all at once, failed.

Speaking of javascript instead:
```
getCacheProps() {
    return this.constructor.cacheProperties
}
```

The method does not change the issue that you have to define them in every class even if is not needed. while in the prototype they are just available to everyone.

When running getCacheProps from Textbox this.constructor refer to Textbox and doesn't goes up to Itext or Text looking for cacheprops.

The prototype is more powerful that a distinction of state and methods made for the sake of making it, and the fact that you can define a getter on the prototype and having it return a value defined once in a place, without having it per instance, is the proof that that rule is flimsy and more red tape that really useful.

asturur:
> I also think `type` should be static or go away entirely

Type is not much more useful now, yes, but we are not deleting it  since deleting it gives no benefit and there may be apps relying on that and we are generating a lot of headaches already.

> take a look [50a5242](https://github.com/fabricjs/fabric.js/commit/50a5242d2fedee5952bcf2c7ac590a0557d092eb)
> POC success for group and AS

Unfortunately That POC works because cacheProperties and stateProperties are still on the Object prototype up in the chain.
Is when i converted everything that pieces start to fall apart.

I just said the matter is getting complicated and that i had to descope and we need more PRs to get there.

ShaMan123:
> Unfortunately That POC works because cacheProperties and stateProperties are still on the Object prototype up in the chain.
Is when i converted everything that pieces start to fall apart.

extracted cacheProperties and stateProperties to static f3d3c7857

ShaMan123:
So since the POC seems successful my question would be do you agree migrating all shapes to that pattern before continuing exports and bundle?

ShaMan123:
> > Once a class is initialized state should not be modified by the prototype. Don't you agree? Is this what we are discussing here?
> 
> No, you know that i don't agree on that. Changing the prototype is fine as an option if you want to do that and if you know what are you doing. We shouldn't encorauge anyone to do it tho, nor is necessary to block that. We decided to block the functionality because in team discussion that was seen as a drive for happyness, so we are trying to get there.
> 
> Having the cache properties and defaults on the static or in config, is indeed a way to get people to still modify class configuration without the need to go in subclassing just to change for example the default value of the corner shape or the border thickness.
> 
> Also what are we discussing here?
> 
> The PR is descoped to just declare, and we are discussing why my attempt to put all properties on the instance, all at once, failed.
> 
> Speaking of javascript instead:
> 
> ```
> getCacheProps() {
>     return this.constructor.cacheProperties
> }
> ```
> 
> The method does not change the issue that you have to define them in every class even if is not needed. while in the prototype they are just available to everyone.
> 
> When running getCacheProps from Textbox this.constructor refer to Textbox and doesn't goes up to Itext or Text looking for cacheprops.
> 
> The prototype is more powerful that a distinction of state and methods made for the sake of making it, and the fact that you can define a getter on the prototype and having it return a value defined once in a place, without having it per instance, is the proof that that rule is flimsy and more red tape that really useful.

I don't understand anymore

ShaMan123:
> The method does not change the issue that you have to define them in every class even if is not needed. while in the prototype they are just available to everyone.

This is incorrect. I showed that in the comments above. There is static inheritance.

ShaMan123:
You told me it upsets you seeing time and effort wasted. That is my point here.

Static inheritance:


```ts
class A {
  static method() {
    return 1;
  }
  static arr = [1,2,3]
  static type = 'A'
}

class B extends A {
  static {
    this.arr.push(4)
    this.type = 'B'
  }
  static method() {
    return super.method() + 2;
  }
}

class C extends B {

}

console.log('a', A.method(), A.arr, A.type);
console.log('b', B.method(), B.arr, B.type);
console.log('c', C.method(), C.arr, C.type);
```



```ts
"use strict";
var _a;
class A {
    static method() {
        return 1;
    }
}
A.arr = [1, 2, 3];
A.type = 'A';
class B extends A {
    static method() {
        return super.method() + 2;
    }
}
_a = B;
(() => {
    _a.arr.push(4);
    _a.type = 'B';
})();
class C extends B {
}
console.log('a', A.method(), A.arr, A.type);
console.log('b', B.method(), B.arr, B.type);
console.log('c', C.method(), C.arr, C.type);

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G8BQ1oQC5nQS2NAtgKboAWA9gCYAUAlIiqtAE7ECuTAdtAIwDcDAXwYYsuaGCZNoAXmgBtbgBoATIoDMAXWGYcedAE8ADoRnQA5LDNIhSUJBgAhaIQAe6Qhwox4yVCN30jNCk2BAAdBJMYYasECRUACw0DKgh4QbGpmYOVqhCfjpiRKSUtIGMLOjsXBCsxlHF5NR0ANTQyvx51ki24FDQAMLObh5e0E7I3bZkHBBkIIRhIGQA5lRmYGaKcGGNpTTbsBGSh2EZhDT8wDNzC0ur6wBGW+O7xE202w7HTF9nRhcrjd5otlmszMAXgM3iVmttoZF4f9jJcgA)

ShaMan123:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

ShaMan123:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static#calling_static_members_from_a_class_constructor_and_other_methods

asturur:
Let's see if i can recap and understand what is unclear.

This pr was aiming at putting all the values in the constructor for the classes, as we decided we would do.
That has been proven not doable unless a per class small change is done in the initialization sequence, hence, i took a step back and i m just doing the changes to enable the use of the export keyword. Doing both together was a step too big for a single pr.

I defined clearly in my message that in order to get there we need to remove side effects and the use of our set in the constructor, but that requires a per class change and checkup, so every class will have its own pr.
This concept has been also confirmed by you in the message below mine with `no side effect on the constructor block`.

I also said that cache properties and state properties are nor state nor static, are probably more a configuration businnes and for how static and normal methods work together i don't see them happy together.

If i try to mix static and non static methods i get errors as i expect to.

https://www.typescriptlang.org/play?ssl=37&ssc=15&pln=37&pc=19#code/MYGwhgzhAECC0G8BQ1oQC5nQS2NAtgKboAWA9gCYAUAlIiqtAE7ECuTAdtAIwDcDAXwYRiVANaEAngBpoANzAhWhOskbRsAM2hVS2CADpgZDhiatjTA2CZXsHUKwqEI4qTVUN1x02RCEDEDIAcyoAcjBgMTCafnUhVAS0TBw8GyZoAF5oAG1uaQAmaQBmAF1hFNxodEkAB0Is6DDYMKQhJFBIGAAhaEIAD3RCDgoYeDVkrCqJ1D1DdINa1ggSKgAWGi9qkn0DGvrGsO7WxIqptNtGnIMb2GtbWTCJSTDH54Kw0rjJ1IJicmonnULHQ7C4EFY9SsRFIlFo0AA1NACt8hO1OlBoABhPqDYajaC9ZBtJAdEwQPwBIKhCKvOAGGEA2iyO7pFl7Ooqfg+Cn+QIhcIAIzp3QZ-zhNFkorZhI59ViZN8fOp4WAdKxYthgNkGplGv2XNJPPQ0Dw2Q4hAA7tjaNzySbBY0LdburajQYROhws86Rs7UqqQLBXKVEA

Then i added comment here and there to answer the comment around hackiness and steps back.
I m doing just step forwards to the goal of having typescripts and the default values defined per instance rather than on the prototype and limiting prototype usage to the minimum possible.

asturur:
> I also said that cache properties and state properties are nor state nor static, are probably more a configuration businnes and for how static and normal methods work together i don't see them happy together.

The circl/ellips poc is mostly ok if it wasn't that defaults aren't configurable.
defaults needs to sit in an object that the user can access, and can change before starting the app.
The object then gets merged in the this at construction time.
So something like i was trying here before i deleted it ( you can still see it in the commit i linked above ).
We probably need a getDefualts method that can mix them all together as you shown, and we need to either use that in the constructor or read from the same source object

test/unit/path.js
```diff
@@ -127,11 +127,6 @@
     updatePath(path, REFERENCE_PATH_OBJECT.path, true);
     assert.deepEqual(path.toObject(), REFERENCE_PATH_OBJECT);
     updatePath(path, REFERENCE_PATH_OBJECT.path, false);
-    var left = path.left;
-    var top = path.top;
-    path.center();
-    assert.equal(left, path.left);
-    assert.equal(top, path.top);
```
asturur, line 134: the path didn't have a canvas, so this call to path.center() was doing nothing.

ShaMan123:
Thanks for elaborating.
I modified the example you linked. It had a typo, that is why there were js errors.
Regarding 19aa171c2821c30fb64194997a65da18e6585066 I would argue that declare ruined your effort but as you said. Next effort.


```ts
class A {
  static method() {
    return 1;
  }
  set(key, value) {
    // TS stop complaining
    if (this.constructor.arr.includes(key)) {
      console.log('ack', key, value);
    }
  }
  static arr = [1,2,3]
  static type = 'A'
}

class B extends A {
  static {
    this.arr.push(4)
    this.type = 'B'
  }
  static arr = [...A.arr, 'key', 'key2'];
  static method() {
    return super.method() + 2;
  }
}

class C extends B {

}

console.log('a', A.method(), A.arr, A.type);
console.log('b', B.method(), B.arr, B.type);
console.log('c', C.method(), C.arr, C.type);

const c = new C();
const b = new B();

C.method = () => {
  console.log('modified static method');
  return 5;
}

c.set('key', 4);
console.log(b.constructor.type)

C.arr = [...C.arr, 'foo'];
c.set('foo', 'bar')
console.log(c.constructor.method())
```



```ts
"use strict";
var _a;
class A {
    static method() {
        return 1;
    }
    set(key, value) {
        // TS stop complaining
        if (this.constructor.arr.includes(key)) {
            console.log('ack', key, value);
        }
    }
}
A.arr = [1, 2, 3];
A.type = 'A';
class B extends A {
    static method() {
        return super.method() + 2;
    }
}
_a = B;
(() => {
    _a.arr.push(4);
    _a.type = 'B';
})();
B.arr = [...A.arr, 'key', 'key2'];
class C extends B {
}
console.log('a', A.method(), A.arr, A.type);
console.log('b', B.method(), B.arr, B.type);
console.log('c', C.method(), C.arr, C.type);
const c = new C();
const b = new B();
C.method = () => {
    console.log('modified static method');
    return 5;
};
c.set('key', 4);
console.log(b.constructor.type);
C.arr = [...C.arr, 'foo'];
c.set('foo', 'bar');
console.log(c.constructor.method());

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G8BQ1oQC5nQS2NAtgKboAWA9gCYAUAlIiqtAE7ECuTAdtAIwDcDAXwYRiVANaEAngBpoANzAhWhOskbQA9BugAVAMpp0ZAA7RgZfMfDYONgOYNU2AGbQqpbBAB05jhiaswEZMXmBMITagrBSEEOJSNKqOjL4QZCCEXiBkdlQA5GDAYnmyEjLyiso0-OpCqHWGWLjQYUzQALzQANrc0gBM0gDMALrCmDh46JLGhB3QebB5SEJIoJAwAELQhAAe6IQcFDDwao0T9Ooe3q1exqwQJFQALDTJV15TM3N5G0v1Y008K05l0vGDYKFwrI8mUSvMyn08sMamdmkRSJRaBdGCx0OwuBBWDMQujyNQ6ABqaB9FFCFZrKDQADC2z2ByO0C2yGWSFWZD86Uy2VyBThENJmJosghrWlH2mKn4qUFWRy+QARnCNl4JeTZNrZZz5TNqnyBRlVSLgHCmTriGTaLJbYbbZ9FbzUugzHMOIQAO7M2hK-kYaDqn3+zlB3m23VzLHtAB82OVFuF+XwlBc2EIFFReF1eVNqFx+OgAFZ+PSvCJ0PlYbIXsHzUK1eqfCH0AEgmQQm7XkhneEQWCvEOmNDnGQyEilTXRHkpzPoeqwkWzWk02rgB2-F3AsE7RjyTQgA)

test/unit/path.js
ShaMan123: great seeing junk removed
we cleaned up a lot
and more will come for sure

src/shapes/Object/InteractiveObject.ts
```diff
@@ -111,6 +121,28 @@ export class InteractiveFabricObject<
...
+   */
+  _updateCacheCanvas() {
+    const targetCanvas = this.canvas;
+    if (this.noScaleCache && targetCanvas && targetCanvas._currentTransform) {
+      const target = targetCanvas._currentTransform.target,
+        action = targetCanvas._currentTransform.action;
+      if (
+        this === (target as InteractiveFabricObject) &&
+        action.slice &&
+        action.slice(0, 5) === 'scale'
```
ShaMan123, line 138: startsWith?
```suggestion
        action.startsWith('scale')
```

src/shapes/Object/InteractiveObject.ts
```diff
@@ -111,6 +121,28 @@ export class InteractiveFabricObject<
...
+      ) {
+        return false;
+      }
+    }
+    return super._updateCacheCanvas();
+  }
```
ShaMan123, line 144: I love this!
It is well scoped.

asturur:
I m not sure declare are doing anything to the code rather than claryfing for tools that those are types and not actual code.
Do you think are doing something else?

src/shapes/Object/Object.ts
```diff
@@ -820,10 +801,12 @@ export class FabricObject<
...
-   * Sets object's properties from options
+   * Sets object's properties from options, for class constructor only.
+   * Needs to be overridden for different defaults.
+   * @protected
    * @param {Object} [options] Options object
    */
-  setOptions(options: Record<string, any> = {}) {
+  protected setOptions(options: Record<string, any> = {}) {
     this._setOptions(options);
   }
```
ShaMan123, line 811: do we really need this alias?
It is called only once in fabric from this contrusctor

src/shapes/Object/Object.ts
```diff
@@ -1759,6 +1710,8 @@ export class FabricObject<
       'center'
     );
     const originalCanvas = this.canvas;
+    // static canvas and canvas have both an array of InteractiveObjects
+    // @ts-ignore this needs to be fixed somehow, or ignored globally
```
ShaMan123, line 1714: good point

src/shapes/ellipse.class.ts
```diff
@@ -27,8 +27,6 @@ export class Ellipse extends FabricObject {
    */
   constructor(options: Record<string, unknown>) {
     super(options);
-    this.set('rx', (options && options.rx) || 0);
-    this.set('ry', (options && options.ry) || 0);
```
ShaMan123, line 31: I am surprised this doesn't break width and height values

src/shapes/line.class.ts
```diff
@@ -103,7 +101,11 @@ export class Line extends FabricObject {
     // make sure setting "fill" changes color of a line
     // (by copying fillStyle to strokeStyle, since line is stroked, not filled)
     const origStrokeStyle = ctx.strokeStyle;
-    ctx.strokeStyle = this.stroke || ctx.fillStyle;
+    if (isFiller(this.stroke)) {
+      ctx.strokeStyle = this.stroke.toLive(ctx);
+    } else {
+      ctx.strokeStyle = this.stroke ?? ctx.fillStyle;
+    }
```
ShaMan123, line 108: this code looks like a util
I think it is duplicated a few times

src/shapes/polygon.class.ts
```diff
@@ -1,3 +1,4 @@
+import { IPoint } from '../point.class';
```
ShaMan123, line 1: ?

src/shapes/textbox.class.ts
```diff
@@ -457,11 +457,16 @@ export class Textbox extends IText {
...
+// @TODO: Many things here are configuration related and shouldn't be on the class nor prototype
+// regexes, list of properties that are not suppose to change by instances, magic consts.
+// this will be a separated effort
 export const textboxDefaultValues: Partial<TClassProperties<Textbox>> = {
   type: 'textbox',
   minWidth: 20,
   dynamicMinWidth: 2,
+  // with the new controls this noScaleCache here shouldn't be needed anymore
   lockScalingFlip: true,
+  // with the new controls this noScaleCache here shouldn't be needed anymore
```
ShaMan123, line 469: care to elaborate?

src/util/transform_matrix_removal.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+};
+
+/**
+ * This function is an helper for svg import. it removes the transform matrix
+ * and set to object properties that fabricjs can handle
+ * @private
+ * @param {Object} preserveAspectRatioOptions
+ */
+export const _removeTransformMatrix = (
```
ShaMan123, line 39: naming?

src/util/types.ts
```diff
@@ -10,11 +10,20 @@ import type { Text } from '../shapes/text.class';
...
+export const isFiller = (
+  filler: TFiller | string | null
+): filler is TFiller => {
   return !!filler && (filler as TFiller).toLive !== undefined;
 };
 
+export const isSerializableFiller = (
+  filler: TFiller | string | null
+): filler is TFiller => {
+  return !!filler && (filler as TFiller).toObject !== undefined;
```
ShaMan123, line 24: not a big deal but if i am here:

```suggestion
  return !!filler && typeof (filler as TFiller).toObject === 'function';
```

ShaMan123:
I don't know. It is so unclear what happens under the hood. What is taken care of rollup, the plugins and native classes.
Looking at the build file:

image

ShaMan123 (commented):
looks fine
a few small comments

src/shapes/Object/Object.ts
ShaMan123: and `_setOptions` is an alias for `set` so that can go away as well.
too many methods that do the same shit

src/shapes/Object/Object.ts
ShaMan123: I think static canvas should have the non interactive object as a type while canvas should have the interactive object
Wondering how complicated and dumb that might turn out

src/shapes/Object/Object.ts
asturur: not this should go because no strange stuf in the constructor. It will go when we move the defaults

src/shapes/ellipse.class.ts
asturur: because constructor({...}) is doing that up in the chain with setOptions.
The issue could be if you pass with after rx,ry. But those are exactly the kind of things that needs to addressed when we setup the defaults.
I would also not bother, i prefer to say you have to set both of them, and zero is not a good value if not for both. I wouldn't try autofixing for the developer that have to understand this once.

Same issue with polygon bounding box, line points, radius and witdh/height. there are so many things we try to take care of we probably shouldn't.

If it wasn't late now i woudl say that width and height should be internal prop for geometry and the one we set for rect and text should be some sort of public width/height that nothing have to do with the geometry calculcations.

src/shapes/polygon.class.ts
asturur: probably a leftover of a previous thing.

src/shapes/textbox.class.ts
asturur: I think i was thinking at the width behaviour and that scaling does not affect textbox because we usually change the width. but i think i forgot of both axis scaling

src/util/transform_matrix_removal.ts
asturur: don't know pick one. I think exportin only one is good enough

src/shapes/ellipse.class.ts
ShaMan123: > If it wasn't late now i woudl say that width and height should be internal prop for geometry and the one we set for rect and text should be some sort of public width/height that nothing have to do with the geometry calculcations.

Sounds good

> Same issue with polygon bounding box, line points, radius and witdh/height. there are so many things we try to take care of we probably shouldn't.

we should draft topics for a team meeting

src/shapes/polygon.class.ts
asturur: Oh right, when it comes to assigning defaults this will have a constructor too.

src/shapes/ellipse.class.ts
asturur: yes this whole how to use set/ set and _set, side effects, will be a recurring theme

ShaMan123:
@asturur don't have an option in VSCODE that it formats on save running prettier?

ShaMan123:
look https://github.com/fabricjs/fabric.js/compare/vscode-config-patch

asturur:
> @asturur don't have an option in VSCODE that it formats on save running prettier?

Works sporadically for me, both here and at work prettier does not kick in for me.

asturur:
I have mixed the code with the dependency update branch with export and type declaration and i have a local branch with only 101 failing test from 174 i started after merging.
That means the `declare` are avoiding side effects we didn't plan for.
As soon as i m down to 0 fails it means that this work, i ll put all the declare here and then we can merge it

asturur:
ok down to 9 failures but i have to stop now

asturur:
ok so those additional declare changes ( no test changes, no code changed ) should give us support for export and types emitting, using babel OR the other rollup plugin ( That does not support browserlist )

asturur:
let's give priority to this today if possible, i want to move asap to discussing browerlist file and what we support and transpile on which tool.
Practicing with the new import strategy and see if tree shaking is possible is another big chunk of work that i would like to be available for whoever wants to try it

ShaMan123:
how can i help?

src/shapes/textbox.class.ts
```diff
@@ -31,7 +31,7 @@ export class Textbox extends IText {
    * Cached array of text wrapping.
    * @type Array
    */
-  declare __cachedLines: Array<any> | null = null;
+  declare __cachedLines: Array<any> | null;
```
ShaMan123, line 34: this is a dead prop

ShaMan123 (approved):
I would expect TS to expose a flag in the config controlling this instead of this shitty declare business

asturur:
> how can i help?

Usually just approve if the PR is ok, or block if there is something you really don't like

ShaMan123:
Just saw this shit
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#the-usedefineforclassfields-flag-and-the-declare-property-modifier
https://www.typescriptlang.org/tsconfig#useDefineForClassFields

Is this what was creating this shit strorm??

ShaMan123:
and https://github.com/microsoft/TypeScript/pull/33509

asturur:
Yes i didn't read all the links deeply, but exactly the part of initializing as undefined.
The older ts-plugin was shielding us from that.

We will have to assign our own values with define properties too since we are doing it manually if we want to do that on spec.
Babel does that, and we can take their small initialization function.

ShaMan123:
So a babel plugin in will save us from declare?

PR #8585, opened 2023-01-10
chore(TS) use export to get type definitions

asturur:
## Motivation

Get out of our transition type build definitely.

## Description

I had to switch to babel after i was burned by the ts-plugin experience.
Babel output more errors and is more strict and more configurable.
In the configuration of rollup there are also some new bits about export d.ts files that will help us emit type declaration that mirror the interface without exporting internal types.

The build directory now has a src folder that contains type definition for each class and utils.

Just a note on class properties:

```js
    constructor() {
      _defineProperty(this, "browserShadowBlurConstant", 1);
      _defineProperty(this, "DPI", 96);
      _defineProperty(this, "devicePixelRatio", 1);
      _defineProperty(this, "perfLimitSizeTotal", 2097152);
      _defineProperty(this, "maxCacheSideLimit", 4096);
      _defineProperty(this, "minCacheSideLimit", 256);
      _defineProperty(this, "disableStyleCopyPaste", false);
      _defineProperty(this, "enableGLFiltering", true);
      _defineProperty(this, "textureSize", 4096);
      _defineProperty(this, "forceGLPutImageData", false);
      _defineProperty(this, "cachesBoundsOfCurve", true);
      _defineProperty(this, "fontPaths", {});
      _defineProperty(this, "NUM_FRACTION_DIGITS", 4);
    }
```

If we pick up a transpile target that does not support public properties ( likely to be true ) this is how they should look like and also this is how we should define them when creating them from default values.
The _defineProperty function is very small and we should use the same, but i don't think we can force babel then to recognize our.
We could either keep them different or move the public property to manually defined values in the constructor for the sake of consistency.
I don't have a strong opinion as long as defaults are configurable.

Speakin of supported JS and code

`canvas?.viewportTransform` becomes `canvas === null || canvas === void 0 ? void 0 : canvas.viewportTransform`

And i hate it.

I highly prefer `canvas && canvas.viewportTransform`

## Changes
- BREAKING: `import fabric from 'fabric'` rather than `import { fabric } from 'fabric'` ( not a choice appartenly )
- updated all dependencies to modern rollup
- added babel to transpile
- added babel preset/env and babel preset/typescript for necessary plugins and browserlist support
- kept configuration as neutral as possible

## Follow ups

Looking in the built file for `, void 0);` will identify missed declared properties ( still many ) and they need to be cleaned up


asturur:
@ShaMan123 remade this PR from scratch since the other branch was tainted from experiments

asturur:
Oh yes of course now we need something that set fabric into the window to properly work when added as a script i assume

asturur:
Well great :) it works either on firefox or on browser

src/shapes/textbox.class.ts
```diff
@@ -27,12 +27,6 @@ export class Textbox extends IText {
    */
   declare dynamicMinWidth: number;
 
-  /**
-   * Cached array of text wrapping.
-   * @type Array
-   */
-  declare __cachedLines: Array<any> | null;
```
asturur, line 34: pointed out this was unused.

asturur:
compared to master we are saving around 2000 bytes here, some babel magic probably.
```
-rw-r--r--  1 abogazzi  staff  306289 Jan 10 11:48 dist/fabric.min.js
vs
-rw-r--r--  1 abogazzi  staff  303949 Jan 10 11:44 dist/fabric.min.js
```

asturur:
@ShaMan123 that's it.

If we want export and type declaration this is the PR

We need a follow up PR that removes more undefind class properties and then we should start moving the defaults in the class.

rollup.config.mjs
```diff
@@ -0,0 +1,47 @@
...
+// https://rollupjs.org/guide/en/#configuration-files
+export default [
+  {
+    input: process.env.BUILD_INPUT?.split(splitter) || ['./index.ts'],
+    output: [
+      {
+        file: process.env.BUILD_OUTPUT || './dist/fabric.js',
+        name: 'fabric',
+        format: 'umd',
+        sourcemap: true,
```
ShaMan123, line 18: I read that maybe we should add
exports: 'named'

ShaMan123 (commented):
Shouldn't we add more fields to the package for exports and types?

asturur:
> Shouldn't we add more fields to the package for exports and types?

Yes we don't need to do all now. But yes we need to tell package.json where types are at least

asturur:
export named doesn't make much sense, but if is possible we could do it.
i think import fabric from 'fabric' makes more sense if you an do `import { Canvas } from 'fabric'`;
But is all really a test here, let's find out trying. IF we can revert the breaking change later, great.

ShaMan123:
> i think import fabric from 'fabric' makes more sense 

I agree, or `import * as fabric from 'fabric'`

asturur:
Ok merging and reiterating soon with some code shaking test

PR #8589, opened 2023-01-11
fix(IText): cursor animation regression

ShaMan123:
introduced by #8547 breaking `Object#animate` signature


ShaMan123:
closed via 5cd68fea0

PR #8593, opened 2023-01-12
chores(TS) Remove more empty declarations

asturur:
## Motivation

Consistency in property declaration in classes.
Avoid transpilers to pre-define and override definitions with undefined of certain props.

The issue to recap was:
```js
class A {
  constructor(options) {
    ...setOptions
  }
}

class B extends A {
 
  myProperty: boolean;
  constructor(options) {
    super(...optionsB ...optionsA);
    // at this point myProperty has been set by the constructor of A
    // this below line is killing the work of constructor A
    defineProp(myProperty, undefined)
  }
}
```

This may be because our setOptions or other constructor practices are bad patterns, but we are not drilling into that now, this is a simpe way to be safe.
On top of that babel is doing this transpilation, so actually adding a property without declare or without a default value is a TS error that we are avoiding because we a flag in the compiler turned down.
So maybe turning that flag up is another thing we want to do.

## Description

Make all property initialization the same.
If a property needs to be defined in the constructor at runtime, we will need to add declare in front of it, regardless that tests aren't failing.
At least we are consistent and we don't have unexpected behaviours only for some properties.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
               size: { fabric: { minified: fs.statSync('dist/fabric.min.js').size, bundled: fs.statSync('dist/fabric.js').size } }
             });
       - name: checkout src files
-        run: git checkout origin/master -- src index.js
+        run: git checkout origin/master -- src index.ts
       - name: upstream build stats
         run: npm run build -- -s
       - name: persist
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## [next]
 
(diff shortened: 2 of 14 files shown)


asturur:
Moving forward with this, there isn't really nothing changed nor is optional right now.
We could probably do a second round for service classes ( stroke projection and others ) to have default values that make sense instead of undefined.
That will make types easier and less declare around.
But those defaults are anyway useless because are ment to be overwritten right away ( like an empty array of points for the base brush ) so i m not sure where is the advantage there.

Maybe there is some extra optimization the browsers guts can do when they know a property will likely exists rather than appear from nothing.

Happy to dig more there but absolutely not now

PR #8595, opened 2023-01-13
Tests for poly custom control [#8556]

luizzappa:
## Description

Add tests to poly custom control https://github.com/fabricjs/fabric.js/pull/8556


luizzappa:
github actions unit tests are failing because I introduced `assert.close` (asserting that a number is approximately equal to an expected number, within a given tolerance):

https://github.com/fabricjs/fabric.js/blob/8b66c94c8b14eca2c40093e15d74f59716d75b98/test/node_test_setup.js#L172-L184

asturur:
oh thank you i wanted to follow up with this, but i m using all the time i have to finish ts and library build/package

test/node_test_setup.js
```diff
@@ -157,3 +157,28 @@ QUnit.assert.strictEqual = function (actual, expected, message) {
...
+
+  message = message || (actual + " should be within " + maxDifference + " (inclusive) of " + expected + (result ? "" : ". Actual: " + actualDiff));
+
+  this.pushResult({
+    result,
+    actual: actualDiff,
+    expected: maxDifference,
+    message
+  })
+};
```
asturur, line 184: this is useful

asturur:
Something seems slightly off for browsers. Do you know what is it or do you need help?

luizzappa:
Okay, I inserted the `assert.close` method in the wrong file, so it wasn't accessible in firefox and chrome contexts. Fix https://github.com/fabricjs/fabric.js/pull/8595/commits/f774766a145c14cf309d63602a9c3a146f700bf6 

It's good to go.

PR #8598, opened 2023-01-14
refactor(IText): extract draggable text logic to a delegate

ShaMan123:
## Motivation

making IText logic more scoped and readable
Another step forward typing text

## Description

After most of DnD bugs have been fixed in #8534 I have extracted all logic to an external delegate

## Changes

The following have been extracted to `DraggableTextDelegate`:
- `__isDragging` => rename to `__mouseDownInPlace` and exposed as `DraggableTextDelegate#isActive` and as `IText#shouldStartDragging` for canvas
- __dragStartFired
- __isDraggingOver
- __dragStartSelection
- __dragImageDisposer
- setDragImage
- onDragStart (alias exposed on iText as well)
- canDrop (alias exposed on iText as well)
- dragEnterHandler
- dragOverHandler
- dragLeaveHandler
- dragEndHandler
- dropHandler

Changes to draggable text logic:
- types and organizing
- the call to start the drag lifecycle has been moved to mousedown instead of mousedown:before allowing devs to control the lifecycle 2aadf5d
- `canDrop` 8bcc8a61d36a57859980a4d161396222cb51a919 6fcadb86452589ea3d3a42fd5b5a062475c3362f:
   - `!e.defaultPrevented` is now part of logic (all calls to `canDrop` were prefixed with it)
   - `targetCanDrop` is called intead of `canDrop` so overriding `IText#canDrop` will have an affect

Additional:
- moved `__lastSelected` to click behavior where it belongs (`selected` should be moved there as well)
- consolidated all init handler methods of click mixin into `initBehavior`

## Gist

First merge #8534 => https://github.com/fabricjs/fabric.js/compare/fix-text-dnd...text-dnd-followup


ShaMan123:
This is ready I think.
Wanted to point out that the coverage report is rubbish since it computes coverage for node and we have a lot of tests running only in the browser. Since QUnit is shit we can't do that as of now w/o an ugly workaround.

ShaMan123 (commented):
Changed some minor things the few last commits:
highlighted changes in comments.


and added some more tests

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -1,31 +1,66 @@
...
-
     this.on('mousedown', this.onMouseDown);
+
+    // TODO: replace this with a standard assignment when shitty `clone` is removed
+    Object.defineProperty(this, 'draggableTextDelegate', {
+      value: new DraggableTextDelegate(this),
+      configurable: false,
+      enumerable: false,
+      writable: true,
+    });
```
ShaMan123, line 42: this assignment

asturur:
Moving to this now.

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -1,31 +1,66 @@
...
+    // Initializes "dbclick" event handler
+    this.__lastClickTime = +new Date();
     // for triple click
     this.__lastLastClickTime = +new Date();
-
     this.__lastPointer = {};
-
     this.on('mousedown', this.onMouseDown);
+
+    // TODO: replace this with a standard assignment when shitty `clone` is removed
```
asturur, line 36: i'll truncate this comment before merging up to the appropriate part.

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -1,31 +1,66 @@
...
+   */
+  onDragStart(e: DragEvent) {
+    return this.draggableTextDelegate.onDragStart(e);
+  }
+
+  /**
+   * @public override this method to control whether instance should/shouldn't become a drop target
+   */
+  canDrop(e: DragEvent) {
+    return this.draggableTextDelegate.canDrop(e);
```
asturur, line 63: why we would suggest to overrid the class and not create a custom draggable test delegate?

src/util/types.ts
```diff
@@ -71,6 +71,8 @@ export const isFabricObjectWithDragSupport = (
   return (
     !!fabricObject &&
     typeof (fabricObject as FabricObjectWithDragSupport).onDragStart ===
+      'function' &&
+    typeof (fabricObject as FabricObjectWithDragSupport).shouldStartDragging ===
```
asturur, line 75: would be there a case in which onDragStart exists but not shouldStartDragging?

asturur:
This PR is good as it is, but the draggableTextDelegate ( i this a good name ? ) should be somehow optional or we should have an empty one that allow for disabling the feature and not carrying te code around if not wanted.

Probably a matter of an extra class extension and everyone can import text from the preferred export.

asturur (approved):

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: both can be done...
I was thinking of your guideline of being able to control dnd with ease, so I felt it is best to have methods on the instance so the dev doesn't need to understand the delegate

src/util/types.ts
ShaMan123: no

PR #8600, opened 2023-01-15
BREAKING: `extend`, `clone` => internal `cloneDeep`

### LINKED ISSUE #8599: task(): remove extend, clone
clone is a shitty piece of logic.
In case of cyclic objects it will break
We must remove it for good.
No real reason to use it any longer
I suggest exposing an internal `cloneDeep`:
```
export const cloneDeep = <T>(object: T): T =>
  JSON.parse(JSON.stringify(object));
```

This is good since if cloning a FabricObject it will use the `toObject` method. And for all others is good enough.

ShaMan123:
## Motivation

closes #8599 

## Description

As described in the issue:
> clone is a ~~shitty~~ outdated and unmaintanable piece of logic. In case of cyclic objects it will break We must remove it for good. No real reason to use it any longer

extend is an artifact of the past, no need with spread arrays.

## Changes

replace all `clone(obj, true)` with `cloneDeep` that uses json stringify (Not a huge fan but seems ok, especially since cloning FabricObject instance this way uses `toObject` which is great) and remove all `extend`, `clone` calls in favor of object spreads.

I would argue we can remove the need of `cloneDeep` entirely.
It is used in:
- stateful: that thing is dying
- `Object#fromObject`: need to inspect if it is needed or a redundant safeguard because the constructor is in charge of safeguarding against mutation of props
- text style: easy to clone that since it has a known structure

BREAKING:
`clone` and `extend` are used in all examples unfortunately so the community must have adopted them.
Devs using `extend` on classes should mutate the prototype directly (or with `defineProperty`) or subclass.
Using `clone` or `extend` to assign to an object was always a bad idea. Use lodash or whatever.


src/mixins/canvas_animation.mixin.ts
```diff
@@ -1,8 +1,6 @@
 // @ts-nocheck
 
-import { getEnv } from '../env';
-
-fabric.util.object.extend(
+Object.assign(
```
ShaMan123: I wonder if this works, it should I guess
Aren't tests skipped?

ShaMan123 (commented):
I suggest viewing the diff w/o whitespace due to `src/canvas/canvas_gestures.mixin.ts`

ready

asturur:
in my opinion this pr should look like this: https://github.com/fabricjs/fabric.js/pull/8613

And we can address the other refactors of gestures and types separately.

asturur (approved):

PR #8603, opened 2023-01-16
chore(TS): types and tweaks for intersection file

asturur:
## Motivation
Finalize the migration for this file

## Changes
- ~~changed Intersection in static methods with this~~ took back this,
- renamed contains in includes and made it protected. ( contains has a geometry meaning of shape contains point that overlaps a bit )
- added types where missing

### DIFF
--- a/src/intersection.class.ts
+++ b/src/intersection.class.ts
@@ -1,3 +1,4 @@
+import { result } from 'lodash';
 import { Point } from './point.class';
 
 /* Adaptation of work of Kevin Lindsey (kevin@kevlindev.com) */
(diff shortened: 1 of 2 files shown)


ShaMan123:
I thought I did this already...
Then I looked and saw I forgot to publish it and PR.

src/intersection.class.ts
```diff
@@ -66,24 +65,35 @@ export class Intersection {
...
+  ): Intersection {
+    const b2xb1x = b2.x - b1.x,
+      a1yb1y = a1.y - b1.y,
+      b2yb1y = b2.y - b1.y,
+      a1xb1x = a1.x - b1.x,
+      a2ya1y = a2.y - a1.y,
+      a2xa1x = a2.x - a1.x,
+      uaT = b2xb1x * a1yb1y - b2yb1y * a1xb1x,
+      ubT = a2xa1x * a1yb1y - a2ya1y * a1xb1x,
+      uB = b2yb1y * a2xa1x - b2xb1x * a2ya1y;
```
ShaMan123, line 84: I am guessing all of this can be reduced by using `dotProduct` `crossProduct`
but it is a headache

Don't you prefect subtracting points and working with vectors?

asturur: I can do that as a fun thing when the rest is done, and only if i save some bytes, otherwise i see it as non useful

asturur:
@ShaMan123 anything holding this PR?

ShaMan123:
Go ahead

PR #8604, opened 2023-01-16
ci(lint): allow non null assertion

ShaMan123:
## Motivation

non null assertion is convenient

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,6 +15,7 @@ module.exports = {
   rules: {
     '@typescript-eslint/ban-types': 1,
     '@typescript-eslint/ban-ts-comment': 1,
+    '@typescript-eslint/no-non-null-assertion': 'off',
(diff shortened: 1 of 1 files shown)


asturur:
Can we leave them as yellowish? When we have all lint fixed we can figure out how much warnings spam we have, at that point we can judge better between off and yellow warning

ShaMan123:
Sure
is a way to start a discussion and save stuff that otherwise are lost in my machine

asturur:
In general errors needs to be super visible and warning visible.
So the number of warnings we get will influence what we leave as warning and what is silently accepted as normal.

I do think we should spend always a minute to see if there is an alternative to the non null assertion, otherwise we wouldn't have come here to typescript.

I m ok with this as non warning, if there are more important warnings to see, but i would like to get it there first, to a passing lint check

PR #8605, opened 2023-01-16
chore(TS): Remove ts-nocheck from pattern and add types

asturur:
## Motivation
Finish TS migration for this file

## Changes
Removed the setOptions method for Pattern.

### DIFF
--- a/src/pattern.class.ts
+++ b/src/pattern.class.ts
@@ -1,4 +1,3 @@
-//@ts-nocheck
 import { config } from './config';
 import { TCrossOrigin, TMat2D, TSize } from './typedefs';
 import { ifNaN } from './util/internals';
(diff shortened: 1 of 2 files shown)


asturur:
Actions are not starting for me today, i wonder what is going on.

asturur:
image

src/pattern.class.ts
```diff
@@ -141,7 +137,7 @@ export class Pattern {
           this.source.naturalWidth === 0 ||
           this.source.naturalHeight === 0))
     ) {
-      return '';
+      return null;
```
ShaMan123, line 140: doesn't this cause a problem in canvas?
I understand why you prefer this
I would look into usage and do `|| ''`

ShaMan123 (approved):

src/pattern.class.ts
asturur: i tried, if you set up a non valid color for canvas, you still get black.
so null or empty string it doesn't change.
empty string is as non valid as null for a fillStyle/strokeStyle.

createPattern api from context can also return null if pattern creation fails.

PR #8606, opened 2023-01-17
chore(TS): remove Static canvas ts-nocheck

asturur:
## Motivation

Continue TS migration

## Changes
no changes as far i can see.

### DIFF
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -1,6 +1,7 @@
 import type { Control } from './controls/control.class';
 import type { Point } from './point.class';
 import type { FabricObject } from './shapes/Object/FabricObject';
+import type { FabricObject as StaticFabricObject } from './shapes/Object/Object';
 import type { Group } from './shapes/group.class';
 import type { TOriginX, TOriginY, TRadian } from './typedefs';
 import type { saveObjectTransform } from './util/misc/objectTransforms';
@@ -192,8 +193,8 @@ type CanvasSelectionEvents = {
 };
 
 export type CollectionEvents = {
-  'object:added': { target: FabricObject };
-  'object:removed': { target: FabricObject };
+  'object:added': { target: StaticFabricObject };
+  'object:removed': { target: StaticFabricObject };
 };
 
 type BeforeSuffix<T extends string> = `${T}:before`;
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -2,12 +2,16 @@ import type { Control } from './controls/control.class';
 import type { Point } from './point.class';
 import type { FabricObject } from './shapes/Object/FabricObject';
 import type { FabricObject as StaticFabricObject } from './shapes/Object/Object';
+import type { FabricObjectSVGExportMixin } from './mixins/object.svg_export';
 import type { Group } from './shapes/group.class';
 import type { TOriginX, TOriginY, TRadian } from './typedefs';
 import type { saveObjectTransform } from './util/misc/objectTransforms';
 import type { Canvas } from './canvas/canvas_events';
 import type { IText } from './shapes/itext.class';
 import type { StaticCanvas } from './canvas/static_canvas.class';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/no-unused-vars
+export interface BaseFabricObject extends StaticFabricObject, FabricObjectSVGExportMixin {}
 
 export type ModifierKey = keyof Pick<
   MouseEvent | PointerEvent | TouchEvent,
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -11,7 +11,9 @@ import type { IText } from './shapes/itext.class';
 import type { StaticCanvas } from './canvas/static_canvas.class';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/no-unused-vars
-export interface BaseFabricObject extends StaticFabricObject, FabricObjectSVGExportMixin {}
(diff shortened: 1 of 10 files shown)


src/EventTypeDefs.ts
```diff
@@ -192,8 +193,8 @@ type CanvasSelectionEvents = {
 };
 
 export type CollectionEvents = {
-  'object:added': { target: FabricObject };
-  'object:removed': { target: FabricObject };
+  'object:added': { target: StaticFabricObject };
+  'object:removed': { target: StaticFabricObject };
```
ShaMan123, line 197: we need to define this for Canvas/Group with FabricObject

ShaMan123 (commented):
We need to figure out what to do with Group in this context.
Maybe make it generic
```ts
class Group<Interactive extends boolean = true> extends ... {
   declare _objects: Interactive extends true ? FabricObject: BaseFabricObject
}

```


What about renaming the non interactive FabricObject for good?

src/EventTypeDefs.ts
asturur: yes still a wip not sure what is the best solution here.
I tried to make collection dynamic but i failed, the point is that i should be able to say, something that extends the base object, while is not easy. I tried with interfaces too

asturur: Ok this makes canvas and static canvas pass the checks, maybe for group we need to refactor again, as soon as we get there we find out

src/canvas/static_canvas.class.ts
```diff
@@ -1500,10 +1493,9 @@ export class StaticCanvas<
    * @param {string[]} [properties] Array of properties to include in the cloned canvas and children
    * @returns {Promise<Canvas | StaticCanvas>}
    */
-  clone(properties: string[]): Promise<this> {
+  clone(properties: string[]): Promise<StaticCanvas> {
```
asturur, line 1496: Should i do a declare with this one on the EventedCanvas? Promise<this> doesn't work.

src/canvas/static_canvas.class.ts
```diff
@@ -1602,21 +1594,13 @@ export class StaticCanvas<
...
       originalRetina = this.enableRetinaScaling,
       canvasEl = createCanvasElement(),
-      // @ts-ignore
-      originalContextTop = this.contextTop,
-      objectsToRender = filter ? this._objects.filter(filter) : this._objects;
+      objectsToRender = filter ? this._objects.filter((obj) => filter(obj)) : this._objects;
     canvasEl.width = scaledWidth;
     canvasEl.height = scaledHeight;
-    // @ts-ignore
-    this.contextTop = null;
```
asturur, line 1616: nulling the contextTop doesn't seem to have anymore any effect.

src/gradient/gradient.class.ts
```diff
@@ -94,7 +94,11 @@ export class Gradient<
    */
   declare excludeFromExport?: boolean;
 
-  private declare id: string | number;
+  /**
+   * ID used for SVG export functionalities
+   * @type number | string
+   */
+  readonly declare id: string | number;
```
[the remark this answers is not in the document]
ShaMan123: hopefully one day that will be part of the svg scope

src/canvas/static_canvas.class.ts
ShaMan123: legit

src/canvas/static_canvas.class.ts
ShaMan123: can you explain why it isn't working with this?

src/canvas/static_canvas.class.ts
```diff
@@ -722,6 +716,10 @@ export class StaticCanvas<
     }
   }
 
+  drawControls(ctx: CanvasRenderingContext2D) {
+    // Static canvas has no controls
+  }
```
ShaMan123, line 721: shouldn't we alter `renderCanvas`?
Or is too much?

src/canvas/static_canvas.class.ts
ShaMan123: I don't really understand when or why it did have an effect

ShaMan123 (commented):
I suggest modifying `renderCanvas` so that Canvas can override the part that is needed for renderControls and remove it form static canvas for good

ShaMan123:
Did you try making Collection accept another generic that will be the object type?
Group and canvas can do the same. I think it is good.

ShaMan123:
And if we do so we can pass the object type to the event spec

src/canvas/static_canvas.class.ts
asturur: There is a typescript error that complains that this isn't what he wants, i don't have it, but i can retrigger it and cpy paste it next time i work on this pr

src/canvas/static_canvas.class.ts
asturur: I wanted to do that, but since it does it in 2 parts is not as simple as running an extra effect before or after the super call.

asturur:
> Did you try making Collection accept another generic that will be the object type? Group and canvas can do the same. I think it is good.
> 
> ```ts
> export function createCollectionMixin<TBase extends Constructor, TObject extends BaseFabricObject>(Base: TBase) {
>   class Collection extends Base {
>     /**
>      * @type {TObject[]}
>      * @TODO needs to end up in the constructor too
>      */
>     _objects: TObject[] = [];
> ```

That was the first thing i did, but then i couldn't use it.
After doing exactly what you did here in the example, i couldn't call

`... extends createCollectionMixin<TClass, TOtherClass>(Class)` typescript didn't like it

asturur:
Followup task https://github.com/fabricjs/fabric.js/issues/8614

PR #8609, opened 2023-01-18
chore(TS) Remove ts-nocheck from filters ( resize filter will have its own pr )

asturur:
## Motivation

Continue TS migration

## Description
Remove ts-nocheck ad do type adjustments

### DIFF
--- a/src/filters/blur_filter.class.ts
+++ b/src/filters/blur_filter.class.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import type { TClassProperties } from '../typedefs';
 import { createCanvasElement } from '../util/misc/dom';
 import { BaseFilter } from './base_filter.class';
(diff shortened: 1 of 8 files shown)


src/filters/blur_filter.class.ts
```diff
@@ -65,14 +64,14 @@ export class Blur extends BaseFilter {
...
+    const canvas1 = resources.blurLayer1!;
+    const canvas2 = resources.blurLayer2!;
     if (canvas1.width !== width || canvas1.height !== height) {
       canvas2.width = canvas1.width = width;
       canvas2.height = canvas1.height = height;
     }
-    const ctx1 = canvas1.getContext('2d'),
-      ctx2 = canvas2.getContext('2d'),
+    const ctx1 = canvas1.getContext('2d')!,
+      ctx2 = canvas2.getContext('2d')!,
```
asturur, line 74: we probably need a type that is HTMLCanvasElementThatWorks that never return null on getContext('2d').
We just need to agree that if your machine has no memory to create canvases anymore fabricJS is not going to work anyway.

ShaMan123: We can modify the global type of the window/dom

ShaMan123: Or better off,
we use `AssertKeys` in `createCanvasElement`

PR #8610, opened 2023-01-19
chore(TS): IText types

ShaMan123:
## Motivation

part of removing ts-nocheck effort

## Description

remove ts-nocheck for  itext_behavior, itext_click_behavior, IText 

## Changes

smashed click behavior init methods into initBehavior
BREAKING: changed `getLocalPointer` signature

## Gist

first merge #8598


ShaMan123 (commented):
done

src/shapes/itext.class.ts
```diff
@@ -170,6 +168,7 @@ export class IText extends ITextClickBehaviorMixin<ITextEvents> {
    */
   _set(key: string, value: any) {
     if (this.isEditing && this._savedProps && key in this._savedProps) {
+      // @ts-expect-error irritating TS
       this._savedProps[key] = value;
```
asturur, line 172: to add context around this, i think is because we mutate more state other than `isEditing` when going editing mode.
I ll open a task to fix it seems it seems just wrong as a pattern.

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -1,14 +1,20 @@
...
 import { stopEvent } from '../util/dom_event';
 import { invertTransform, transformPoint } from '../util/misc/matrix';
 import { DraggableTextDelegate } from './DraggableTextDelegate';
+import { ITextEvents } from './itext_behavior.mixin';
 import { ITextKeyBehaviorMixin } from './itext_key_behavior.mixin';
 
-export abstract class ITextClickBehaviorMixin<EventSpec extends ObjectEvents>
+function isRightClick(button?: number) {
+  return button === RIGHT_CLICK;
+}
```
asturur, line 13: in canvas file we define this function
```
function checkClick(e: TPointerEvent, value: number) {
  return !!(e as MouseEvent).button && (e as MouseEvent).button === value - 1;
}
```
So on top of reusing it, i wonder why there we use `value - 1` and here were are using value directly.

src/mixins/itext_click_behavior.mixin.ts
```diff
@@ -128,8 +123,8 @@ export abstract class ITextClickBehaviorMixin<EventSpec extends ObjectEvents>
    * initializing a mousedDown on a text area will cancel fabricjs knowledge of
    * current compositionMode. It will be set to false.
    */
-  _mouseDownHandler({ e }: TransformEvent) {
-    if (!this.canvas || !this.editable || (e.button && e.button !== 1)) {
+  _mouseDownHandler({ e, button }: TPointerEventInfo) {
+    if (!this.canvas || !this.editable || isRightClick(button)) {
```
asturur, line 127: i see 2 things here ( and on `_mouseDownHandlerBefore` ):
- we move to evaluate the button option rather than the event.button.
- we changed from a `not a left click` to a `is a right click` that leaves the middle click and other possible clicks in an new state ( i would definitely avoid that ).

On top of that it looks like we expose on the event this button value that is not the real button value from the event but more a fabricjs custom value. I wonder if we should remove that value completely and let people look at the event itself

asturur: ok seems like we used to look at button because `buttons` wasn't on IE. But that is a topic or another day.
I think the only important thing here is to change this isRightClick with a `!isLeftClick` as it was before so we don't change behaviour

src/mixins/itext_click_behavior.mixin.ts
asturur: Answer i found later: because here we are not inspecting e.button directly.

asturur:
Let me know if you have time to address that rightclick vs not left click, so that then i can merge.
If you don't have time for it i can do it tomorrow.

src/shapes/itext.class.ts
ShaMan123: I agree

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: it was previously e.button.
`checkClick` is used on the native event before the synthetic event is fired by fabric. The synthetic event has the `button` attr set and is `e.button+1`. It is far from ideal that that values aren't identical. But that is what it is.
I thought we should use the button attr of the synthetic event or use `checkClick` and made it clear we are checking 
 to avoid a right click.

src/mixins/itext_click_behavior.mixin.ts
ShaMan123: I didn't mean to change behavior. It was unclear to me

ShaMan123: `(e.button && e.button !== 1)` means to me:
value is not 0 (not left click) and not middle click => is right click
Why middle click is alright I didn't understand.

ShaMan123: I think it is best you do it

ShaMan123:
BTW I wanted to move all IText classes under a folder same as Object, makes sense IMO
Text the same.
In general are we going to rename files before beta? I think we should.

src/mixins/itext_click_behavior.mixin.ts
asturur: I just don't want to chane the behavioru in a typing PR.
I think the reason of being `not a left click` is because by default fabric wasn't doing anything else with other clicks ( other than allowing for event firing ). So as long as it matters to us, right, middle and the other 2 clicks ( called 4 and 5 that are in some custom mouse ) we just don't do anythin.

asturur:
Renaming files or moving them is not a huge issue for me at this point in changes

src/mixins/itext_click_behavior.mixin.ts
asturur: Ok somethin weird is going on here.
The old code was introduced 6 years ago to guard against NON LEFT CLICKS.
That was the intention.
The code seems wrong in that regard since the value of button for left clicks is `0` and so checking for `!==1` was plain wrong.
On top of that UTs have mocked events so not really reliable.
To avoid mixing types with changes, i just kept the exact code that was there before, but what this code is doing should be investigated.

asturur: here the context:
https://github.com/fabricjs/fabric.js/pull/3888/files#diff-ca85fc40679e7fc083b650fbb7e7851f09d84a83da4401ad2ec326f57e31febaR86

asturur (dismissed):

asturur (approved):

PR #8611, opened 2023-01-19
Update settings.json

ShaMan123:
## Description

let vscode fix import order on save

## Gist

There is a prettier plugin as well:
https://www.npmjs.com/package/prettier-plugin-organize-imports:

> This is the same as using the "Organize Imports" action in VS Code.

> This plugin inherits, extends and overrides the built-in Prettier parsers for babel, babel-ts, typescript and vue, i. e., it's incompatible with other plugins that do the same... so only the last loaded plugin that exports one of those parsers will function.

So no to that

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,6 +12,6 @@
   "editor.formatOnSave": true,
   "editor.formatOnPaste": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": true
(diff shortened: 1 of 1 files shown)


asturur:
i do not want to organize imports.
It should be off, changing import order is just distracting in PR reviews.

ShaMan123:
On the other hand. Once we do it that is it.
Exactly the as same prettier

PR #8613, opened 2023-01-19
chore(): #8600 with less changes

asturur:

### DIFF
--- a/src/mixins/stateful.mixin.ts
+++ b/src/mixins/stateful.mixin.ts
@@ -1,6 +1,6 @@
 // @ts-nocheck
 import type { FabricObject } from '../shapes/Object/Object';
-import { extend } from '../util/lang_object';
+import { clone } from '../util/lang_object';
 
 const originalSet = 'stateProperties';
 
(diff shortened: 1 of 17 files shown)


ShaMan123 (commented):
I view it as follows:
You removed array utils from fabric. Why?
Because fabric isn't lodash.
I remove object utils from fabric because fabric isn't lodash.
Then you disagree and I am confused.
I understand you want to limit breaking changes but why keep 2 crippled unused methods? 
clone is used only with the deep flag
extend cannot use deep, meaning anyways this is "BREAKING". Just like the other PR. And both PRs save the dev from infinite recursion. 
 
I think this PR needs defending, not the other.

asturur:
Is not about defending i want to limit changes in general to the codebase and instead of 300 comments i wanted to show you how the diffs should look like.
We can also avoid exporting clone, but renaming clone to cloneDeep is just unnecessary and we should always avoid doing that ( also because is used in 3 places ).

Removing clone from the export before having a way to test internal functions means also not being able to test it, that was another part of the equation for which i didn't want to remove it from exports right away.

asturur:
(This wasn't meant to be merged at all, it was the outcome of our conversation when i told you i ll show you with an example )

PR #8620, opened 2023-01-24
chore(): Arrange mixins + text files

ShaMan123:
## Description

move text, itext mixins under dedicated `Text`, `IText` folders and rename files
extract `Collection` `Observable` and `CommonMethods` to `src`, they are not mixins so they don't belong under the mixins folder.
parked `canvas_animation.mixin` in the parkinglot


ShaMan123 (commented):
This is a no brainer.
Just moved files around and renamed.
~~I would move `src/mixins/canvas_animation.mixin.ts` to the parkinglot as part of this effort.~~ **DONE**. Then under mixins we have stateful left waiting to be deprecated and eraser which has been refactored to be part of brushes.

ShaMan123 (commented):
done

asturur (approved):

PR #8622, opened 2023-01-24
TS(): Testing fabric 6 package configuration and code splitting. Fix fabric object access

asturur:
## Motivation

Try to package the library with modern standards
Some notes i took from here about main/module/browser
https://esbuild.github.io/api/#main-fields

IN theory from a react or bundled project we do:

```js
import * as fabric from 'fabric'
```

or

```js
import { Canvas, Textbox } from 'fabric'
```

Without any of those beeing tree-shakable for now.

And from browsers or cdn imports we refer to
```
'fabric/dist/fabricjs.umd.min.js'
```

And also i still have no idea if the main es build is supposed to be minified or not.

## Changes

- main fabric export is now the UMD, while the `module` directive is an ES build named fabric.js.
- removed access to the global fabric. constant in a mixin that was a leftover of the ENV migration
- added copy-paste temporary data to getEnv(), not a great place but for now it works as a generic container
- added an extra build script ( unused for now ) to experiment with single file transpiling and bundling


rollup.config.mjs
```diff
@@ -17,6 +17,12 @@ export default [
         format: 'umd',
         sourcemap: true,
       },
+      {
+        file: process.env.BUILD_OUTPUT || './dist/fabric.es.js',
+        name: 'fabric',
+        format: 'es',
+        sourcemap: true,
+      },
```
asturur, line 25: this change makes it work with the 3 setup i tested so far, but is all rollup based, i want to check webpack too and and the standard next.js and SWC setup.

I m not saying i think we should have 2 builds ( or 3 ) but right now i m collecting data on how to ship something that works out of the box and is not tied to a gigantic fabric monolith.

asturur:
I made some experiments do not take the diffs as definitives yet, looking at options

.browserslistrc
```diff
@@ -1,3 +1,5 @@
 # Browsers that we support
 
-chrome >= 58
+chrome >= 64
+safari >= 11
+firefox >= 58
```
asturur, line 5: those for examples are browsers released Jan 2018.
But to be honest i would like to decide by canvas features.
Like which is the minimum set of browsers with advanced text metrics?
For the beta is not relevant we can readjust

.babelrcAlt
```diff
@@ -0,0 +1,19 @@
...
+      }
+    ],
+    [
+      "@babel/typescript",
+      {
+        "allowDeclareFields": true
+      }
+    ]
+  ]
+}
```
asturur, line 19: this would be an alternative config file, very loose, that would behave as tsc, just stripping types.
This could be useful for developers that want to import files directly and then want to bundle and transpile on their own.

asturur:
Ok this seems a decente approach, apart i need to fix that we can't add random things to the fabric object, because now it is an export and you can't add things to it, is read only.

asturur:
Ok after a bit of pain i m here:
image

This works, isn't exactly what we want now but is a path forward.
The import code looks like this:

```js
import { useEffect } from 'react'
import { Canvas } from 'fabric/dist/src/canvas/Canvas';
import { Textbox } from 'fabric/dist/src/shapes/Textbox';
import './App.css'

function App() {

  useEffect(() => {
    const c = document.getElementById('c');
    if (c!.hasAttribute('data-fabric')) {
      return;
    }
    const fabricCanvas = new Canvas('c');
    const text = new Textbox('Hello 6.0', {});
    fabricCanvas.add(text);
  }, []);

  return (
    <div className="App">
      <canvas id="c" width="1280" height="720" ></canvas>
    </div>
  )
}

export default App
```

And the import logic works, while the app doesn't because we are full of side effects yet.
side effects will break tree shaking both for the single imports and for dead code removal from the main bundle.

We have 2 builds, one for ES and one from UMD, type works and then we can also transpile the SRC folder as it is for single file imports, but that shouldn't be considered a feature yet, but more something to experiment with.

Is still strange that in order to import Text and Canvas i have to pull in both Rect and Group but eventually we still have some import to tweak here and there.

asturur:
Also at some point we shoud check esbuild seems like it does what rollup does but in a tenth of the time.

src/constants.ts
```diff
@@ -1,7 +1,8 @@
 import { TMat2D } from './typedefs';
+// use this syntax so babel plugin see this import here
+import { version } from '../package.json';
 
-// TODO: consider using https://github.com/swiing/rollup-plugin-import-assertions so we can import json in node and have rollup build pass
-export { version as VERSION } from '../package.json';
+export const VERSION = version;
```
asturur, line 5: a note about import assertion.
Looks like that the whole JSON as to be imported as a default import, or at least this is what was happening while i added the babel plugin to support import assertion.

src/env.ts
```diff
@@ -2,13 +2,19 @@
...
+};
+
 type TFabricEnv = {
   document: Document;
   window: Window;
   isTouchSupported: boolean;
   isLikelyNode: boolean;
   nodeCanvas: Canvas;
   jsdomImplForWrapper: any;
+  copyPasteData: TCopyPasteData;
```
ShaMan123, line 17: this should die in a PR I need to port

src/shapes/Object/Object.ts
```diff
@@ -1640,7 +1640,8 @@ export class FabricObject<
   cloneAsImage(options: any): Image {
     const canvasEl = this.toCanvasElement(options);
     // TODO: how to import Image w/o an import cycle?
-    return new Image(canvasEl);
+    const ImageClass = classRegistry.getClass('image');
```
ShaMan123, line 1643: nice!

src/constants.ts
ShaMan123: weird that the syntax must be like this. but babel...

ShaMan123 (commented):
First of all, good job! This part is the hardest and most frustrating IMO.
I don't have too much to say accept that it is disappointing that there is so much trial and error in this field, I would expect rollup to explain all this in a few boilerplates. But here we are.

I don't understand, if I want to use fabric in a tree shakable manner I can't use the rollup outputs? Sounds ridiculous, why would anyone use rollup then? I can't believe it is true.

My main point to change in this PR are the index files, hopefully that will fix this issue.
We must not use `const`, only import/export. That is why we should have more index files for utils, filters etc.
I am quite sure this is what is holding back tree shaking.
I also think it must be possible doing:
```ts
import {Canvas, Textbox} from 'fabric'
```
And have tree shaking work w/o importing files directly. That is the point of tree shaking, elsewise you are manually reducing the tree by manual imports. So that can't be considered as tree shaking. No need to support that IMO once tree shaking works.
I am also against exporting a default monolith, I thought you are against as well. Is this just a step forward? If a dev wants to use the fabric namespace they should do:
```ts
import * as fabric from 'fabric'
```
What is your opinion?

src/constants.ts
asturur: is just the plugin i found that is not great.
If i use the good one, to support assertion, i get all the package.json in the codebase.
While this is a plugin that swaps the statement with a const, but has been poorly written

src/env.ts
asturur: if different copy paste code make this unnecessary, great.

ShaMan123: yes, using the native event `transferData` is the path forward

asturur:
> First of all, good job! This part is the hardest and most frustrating IMO. I don't have too much to say accept that it is disappointing that there is so much trial and error in this field, I would expect rollup to explain all this in a few boilerplates. But here we are.
> 
> I don't understand, if I want to use fabric in a tree shakable manner I can't use the rollup outputs? Sounds ridiculous, why would anyone use rollup then? I can't believe it is true.
> 
> My main point to change in this PR are the index files, hopefully that will fix this issue. We must not use `const`, only import/export. That is why we should have more index files for utils, filters etc. I am quite sure this is what is holding back tree shaking. I also think it must be possible doing:
> 
> ```ts
> import {Canvas, Textbox} from 'fabric'
> ```
> 
> And have tree shaking work w/o importing files directly. That is the point of tree shaking, elsewise you are manually reducing the tree by manual imports. So that can't be considered as tree shaking. No need to support that IMO once tree shaking works. I am also against exporting a default monolith, I thought you are against as well. Is this just a step forward? If a dev wants to use the fabric namespace they should do:
> 
> ```ts
> import * as fabric from 'fabric'
> ```
> 
> What is your opinion?

There are 2 way to obtain partial imports.
One is the bundler actually going for the file and understanding what code is used and what not and remove it from the bundle or from the single JS files you import ( bundles or not ).
That could work both from fabric or from the single files,  but actually can't because of:
- prototype changes
- assigning controls how we do it
- class Registry
When you want to use this way, you mark the dependency as `side effect free` and you let the bundler run human written code that will remove the extra code. FabricJS is not side effect free yet.

The other way is to provide disjointed files and you import what you need, and each file follow its own dependency graph. On top of that the approach above can be still be applied.

I do agree that we should not create consts in the export, that breaks tree shaking, and that can be fixed, but since all the rest is not ready now, i wouldn't bother.

The official method of import stays `import { a, b } from 'fabric'`  for now and the one with the mixed files is just a way to have a ready way to experiment with it but we should consider it not supported, i wouldn't document it either.

The index filex are necessary but i don't want  them to be used in the code files because then we have to guards for us using the `* as filters` and then using it as an object. So i would totally use them to build the package but i would avoid to refer them in the middle of the code if possible for preferences

It could also be that some weird export sytnax works too `export { a, b } as filters` but never tried

ShaMan123:
Tackling these:
- prototype changes - a quick look shows we can eliminate the few usages by extracting logic to a util (`_setLineDash` and a few things in Group that I did just because importing was impossible at the time)
- assigning controls - we can do that in the constructor, it makes sense
```ts
this.controls = objectDefaultControls
```
- class Registry - I don't understand what is blocking here...

Regarding the index files, ok.
I still lack the understanding of what is missing. I am trying to run the code with examples that worked. Some don't agree to with master and with this branch. I will try more.

ShaMan123:
Running next project know fails
Before I managed to get it working with a dynamic import w/o SSR.
I don't understand what is happening now but it tries to evaluate fabric before I import it...

asturur:
Let's add the fabricjs-webpack changes in place for the basice react setup, and you can push there a folder with the next example? so we can test it.

ShaMan123:
possible, it is the next app for #8135 
https://github.com/fabricjs/fabric.js/tree/sandbox-ci/.codesandbox/templates/next

Must use the cli or link fabric manually
If you want I can push it there

ShaMan123:
> The class registry gets run when you import a file for that class.
That means that it force you to keep all the code that is in the Canvas file for example outside the. class itself, if the bundler goes for unused code removal will also remove teh call to the registry because it has no idea what are you doing with that, if it belong to the Canvas call or to the below or above class.

Didn't understand. Sorry.
Unused classes shouldn't get registered, that is what we want, right? Do you think the opposite is happening? That since the class is registered it gets bundled no matter what and won't be shaken off?


> Assigning controls in the constructor is a no go because then you don't have anymore a control set but hundreds of them

If you assign the same `defaultObjectControls` object it is still a single shared object as you wish it to be.
True, it will require a loop to assign a different control set to objects. But no more.
We have been in this discussion so I won't poke your eye. Sorry if I did.

ShaMan123:
> Fixing the default values as first will clear some lines of code that are outside classes and will let us see better what is left.

Ahh so you think this is holding back tree shaking because of the assignment?

ShaMan123:
We could use a devops/CI person... lending a hand

asturur:
so the issue with `unused code` is that if this is our file

```js
Class Canvas {

}

classRegistry.setClass('canvas', Canvas);
```

and then in the application

```js
import { Canvas } from '../canvas/Class';
```

The `unused code removal` will won't know what to do with  the line `classRegistry.setClass('canvas', Canvas);`.
It will skip it and break fabric behaviour.
It either skips all the function calls not imported breaking the app, or import them all not saving any code.
It can't pick up smartly.

asturur:
Sorry i m not being super clear with explanations.
Everything that is inside a file, called as a function instead of being an export, is possibly concern with tree shaking.

ShaMan123:
@asturur take a look at 7d22d71
It is squashed so you an revert it.
Should fix most issues with SSR along with improved rollup config.
Thoughts?

ShaMan123:
I accidently committed dist files. Sorry.
Changes made:
`getEnv` can't be called when importing because then SSR isn't possible with decent importing.
- node entry point - didn't remove imports such as Canvas, Textbox. That can be done in a follow up
- AnimationFrameProvider - lazy init for SSR
- StaticCanvas - moved node methods to node index, fix `cloneWithoutData` return value to match node StaticCanvas
- object default values: workaround for SSR

ShaMan123:
try the following on branch `v6-sandbox` and on branch `v6-sandbox-test1`
`npm run sandbox start -- -t next`
`npm run sandbox start -- -t node`

`v6-sandbox`: all works accept next api that should work on server side and use fabric in node context

ShaMan123:
I am very excited to share that SSR works out of the box with #8632 tested in `v6-sandbox`

ShaMan123:
And this PR now works with next as well with SSR config tested in `v6-sandbox-test1`

ShaMan123:
https://stackoverflow.com/questions/55339256/tree-shaking-with-rollup

ShaMan123:
I have made progress with tree shaking and found the following:
- registering classes can't be done as we do if we want tree shaking (I think that is what you meant in a previous comment)
- we need to solve how static class properties are build with babel or something since in the build these props are assigned to the class with a `_defineProperty` call that too blocks tree shaking (which is absolutely strange since the ecosystem uses classes and tree shaking and static props)

Take a look at 800624448, it is a test for tree shaking. Tree shaking works for brushes for example. You can see that I didn't import brushes and the entire brushes code was romoved.

asturur:
https://github.com/fabricjs/fabricjs-webpack/tree/master/next-app

added the broken next-app here.
I see that the breakage is the bundler trying to do stuff with node deps.
So since we know that needs to be solved, we can merge this first part and continue with the node entrypoint

PR #8629, opened 2023-01-31
fix(): Env patch

ShaMan123:
## Motivation

Make env setup lazy so it runs when classes are initialized instead of when fabric is imported

## Description

As far as I can tell 
https://github.com/fabricjs/fabric.js/blob/028fd89bbcaea2a6193abf805637082bddd3eb30/src/shapes/Object/defaultValues.ts#L100
is the last `getEnv` call that runs from import and once we refactor default values to a static method that should go away

## Changes

- StaticCanvas node streaming methods are now part of the class no matter what. Not sure I like it but I don't think is that bad for now. Once we have a bundle for node we can refactor
- `setupEnv` is called lazily
- DOMParser is not needed on the global scope

### DIFF
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -42,6 +42,7 @@ import { pick } from '../util/misc/pick';
 import { matrixToSVG } from '../util/misc/svgParsing';
 import { toFixed } from '../util/misc/toFixed';
 import { isCollection, isFiller, isPattern, isTextObject } from '../util/types';
+import type { JpegConfig, PngConfig } from 'canvas';
 
 const CANVAS_INIT_ERROR = 'Could not initialize `canvas` element';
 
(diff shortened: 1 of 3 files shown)


src/canvas/StaticCanvas.ts
```diff
@@ -1693,6 +1694,24 @@ export class StaticCanvas<
...
+  }
+
+  /**
+   * node only method
+   * @throws when called from a browser environment
+   */
+  createJPEGStream(opts?: JpegConfig) {
+    const impl = getNodeCanvas(this.lowerCanvasEl);
+    return impl && impl.createJPEGStream(opts);
+  }
```
ShaMan123, line 1714: look here: https://github.com/Automattic/node-canvas/blob/fdf709a7b08abae33a93c510b96f71df6c13c7b0/types/index.d.ts#L89-L112

We should think of adding `toBuffer` and other methods or might be better to expose a single `getNodeCanvas` method or even better to alter the `getElement` method to return the node canvas instance if running in node. Thoughts?

ShaMan123:
```ts
getElement() {
    getEnv().isLikelyNode
      ? getNodeCanvas(this.lowerCanvasEl)
      : this.lowerCanvasEl;
  }
```

or even as a different method

ShaMan123: another PR for sure

ShaMan123:
will be closed by 7d22d71f2

PR #8632, opened 2023-01-31
ci(): Node build

ShaMan123:
## Motivation

#8208  + supporting SSR out of the bbox

## Description

continues #8622 with additional config changes and fixes to support SSR and remove `require` calls and awkwardness from fabric env setup.

After this is merged we can change to node index file.

## Changes

view the diff from #8622  https://github.com/fabricjs/fabric.js/compare/v6-package-testing...node-build

- split env file to browser, node and to an index file that is consumed by fabric
- node entry point - didn't remove imports such as Canvas, Textbox. That can be done in a follow up
- AnimationFrameProvider - lazy init for SSR, not calling `getEnv` during import
- StaticCanvas - moved node methods to node index, fix `cloneWithoutData` return value to match node StaticCanvas (or any subclass)
- object default values: workaround for SSR, can be done differently. I want to rename `isLikelyNode` because we are now certain it is node. Haven't done it yet so the PR isn't polluted. Give me a green light and I will (or in a follow up)
- ~~created index files for exports, put them under the folders and renamed to index. If you don't want the name to be index to avoid importing from it (though I don't really understand why) revert 8829519 and naming will be restored.~~ reverted e4fab60

## Gist

In order to support SSR `getEnv` can't be called when importing fabric because then env is setup incorrectly because at the time of importing window/document are undefined and only when canvas is initialized then env should be setup.

## In Action

Test this config by doing the following:
checkout `v6-sandbox`
run `npm run sandbox start -- -t next`
You will see in the browser tab 2 canvases, one with next SSR dynamic import and one plain and simple, both working.
In the node tab hit the download buttons to see the node build in action

OR run in gitpod
 [Gitpod Ready-to-Code](https://gitpod.io/#https://github.com/fabricjs/fabric.js/pull/8632)


ShaMan123 (commented):
ready

ShaMan123:
I need somehow to tell rollup to create the type index with a different name. Tried `outFile` in tsconfig without any luck.
Also it seems we need to do something like
```ts
export declare module 'fabric/node' {

}
```

asturur:
@ShaMan123  can we reduce this PR to the node entry points and leave code splitting optimizations for another PR?
I would like to see only the additional changes to rollup to get a node build, and the necessary changes to the file to obtain the node build.

Then we can move to the other stuff like codesplitting and tree shaking, and similars that could also be that we don't manage to support right away and i don't want to dig another rabbit hole of to follow.

index.ts
```diff
@@ -1 +1 @@
-export * as default from './fabric';
+export * from './fabric';
```
asturur, line 1: if we want the umd module to keep working my understanding is that this needs to be an export default.
Does this change fix something broken?

src/canvas/SelectableCanvas.ts
```diff
@@ -1503,6 +1503,7 @@ export class SelectableCanvas<
     wrapperEl.removeChild(lowerCanvasEl);
     this.contextCache = null;
     this.contextTop = null;
+    // TODO: interactive canvas should NOT be used in node, therefore there is no reason to cleanup node canvas
```
asturur, line 1506: well for some weird reason people could want to draw a picture of something with controls.
I wouldn't argue over people intentions/needs. If it works it works, if it can't work it can't work.

src/env/node.ts
```diff
@@ -0,0 +1,52 @@
...
+    window: fabricWindow,
+    isTouchSupported,
+    isLikelyNode: true,
+    nodeCanvas,
+    jsdomImplForWrapper,
+    copyPasteData,
+  };
+};
+
+setEnv(getEnv());
```
asturur, line 52: is there a reason why node and browser are not mirrored?
this does setEnv while browser doesn't.

src/canvas/SelectableCanvas.ts
ShaMan123: A weird concept indeed
So you are for... OK
I thought we wanted to reduce the bundle but to say the truth I don't care.
One day we can think about it. So I should add add streaming to canvas as well.

src/env/node.ts
ShaMan123: Yes there is, to support SSR we can't setup env during import so it must be set lazily.
Of course there is another approach of not setting it at all and letting the dev do that.

src/util/animation/AnimationFrameProvider.ts
```diff
@@ -1,23 +1,28 @@
...
+      };
+  }
   return _requestAnimFrame.call(getEnv().window, callback);
 }
 
 export function cancelAnimFrame(handle: number): void {
+  if (!_cancelAnimFrame) {
+    _cancelAnimFrame =
+      getEnv().window.cancelAnimationFrame || getEnv().window.clearTimeout;
+  }
```
asturur, line 26: in another PR i want to come back over this file.
I want to know with current browser support and current JSDOM who does or does not support request animation frame and if we still need polyfills.

asturur:
The PR looks good to me, the right amount of changes.
I would like to better understand choices on extensions and support.
We can't get away without a umd build, and if we keep one it should have a default export and a js extension for clarity sake an expectations.

I m not sure i m correct on this, i could do some research, but if you have infos please dump them

index.ts
ShaMan123: There is no need for a default export. It is wrong.
Rollup handles the naming in the umd build according to the `name` field.
So this allows all build types including es, cjs

ShaMan123 (commented):
I still need to declare for TS `fabric/node` and `fabric/browser` modules. 
And I think I will rename `fabricNode.ts` to `index.node.ts`.
I will ping you when I am done

ShaMan123:
> I would like to better understand choices on extensions and support.
We can't get away without a umd build, and if we keep one it should have a default export and a js extension for clarity sake an expectations.

The umd build is here.
I used `mjs` `cjs` to align with the ecosystem. umd is commonjs and therefore I gave it the cjs extension. 
I exported paths such as `import * as fabric from 'fabric/node.cjs'` so devs can control which build to use with ease according to their build system needs etc.
IMO it is the most clear way, I don't think too many people know what umd is (I didn't until we reached that point).

src/shapes/Object/defaultValues.ts
```diff
@@ -97,7 +97,8 @@ export const fabricObjectDefaultValues = {
   lockSkewingY: false,
   lockScalingFlip: false,
   excludeFromExport: false,
-  objectCaching: !getEnv().isLikelyNode,
+  // TODO: restore once default values are refactored to a method
+  objectCaching: true, //  !getEnv().isLikelyNode
```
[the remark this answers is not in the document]
asturur: no let's keep this pr as it is, for once is small.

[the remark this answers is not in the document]
ShaMan123: >  for once is small.

I held back hard! 😅

asturur:
let me know when you are done so i can play with the exports and examples.

ShaMan123 (commented):
I renamed the output files to `index` and `index.node`

This is PR ready apart from TS not recognizing 
`import * as fabric from 'fabric/node'`

I tried many things and read a lot of stuff online and wasted 3-4 hours on it so I am giving up
Seems that it is another TS limitation
Tried declaring modules in d.ts file with no success...
Added `paths` in tsconfig after reading about it and seeing rxjs that has that config and imports work for it but couldn't get it to work for fabric (However I might have managed to get it working but symlinking is known to work is this cases and I didn't test it w/o symlinks)

Let's move on and get back to it.


You can play in `v6-sandbox` branch which is up to date with this PR

ShaMan123:
We need to consider removing index.ts so vscode and TS don't import that for types. It isn't needed anyways. It is simply an alias for fabric.ts

ShaMan123:
A workaround for type recognition of `fabric/node` is renaming `index.node` to `node`
I think I will do that

ShaMan123:
> A workaround for type recognition of `fabric/node` is renaming `index.node` to `node` I think I will do that

Well I don't know. I am leaving as is since it is clearly as TS bug.
The dev can configure paths in tsconfig as follows to fix it:
```json
paths: {
   'fabric/node': './node_modules/fabric/dist/index.node'
}
```

asturur:
looking in a couple of things for node, but in general this PR is good as it is.
I probably don't want to build node with rollup nor having a mjs and cjs version.
I think bundling pe se is a browser issue only but i will open a follow up pr with that

index.ts
asturur: i didn't try this yet, if it doesn't work as planned, we can still put back the default. but as it is, index.ts is a duplicate of fabric not usefult anymore.

asturur: Also this wasn't here for the naming, but becuase some tool explicitly required a default export.
The name is gone since we don't do anymore `import { fabric } from 'fabric'`

asturur: no sorry, name is because window.fabric. but is still unrelated to the default export.

src/env/node.ts
asturur: ok this is the issue in general with doing things at import time, it creates issues in general as a process.
Probably we should let the developer do that, but i m not ready to think about that.

ShaMan123:
Sharing some research on typing submodules:
https://github.com/Microsoft/TypeScript/issues/17945
https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html#library-file-layout

package.json
```diff
@@ -76,6 +76,7 @@
     "@rollup/plugin-terser": "^0.3.0",
     "@rollup/plugin-typescript": "^11.0.0",
     "@types/fs-extra": "^9.0.13",
+    "@types/jsdom": "^20.0.1",
```
asturur, line 79: either use jsdom types for 19 or upgrade jsdom to 20

asturur (changes requested):
i that is the only thing we need to change and we are good to go

src/env/index.ts
```diff
@@ -0,0 +1,19 @@
+import { TFabricEnv } from './types';
+import { getEnv as getBrowserEnv } from './browser';
+
+let env: TFabricEnv;
+
+export const setEnv = (value: TFabricEnv) => {
+  env = value;
+};
+
+export const getEnv = () => env || getBrowserEnv();
```
asturur, line 10: i m lost a bit on this thing here.
Why again do we need an index file and the setEnv part if we could use the same trick with `isInitialized`

Seems like we want fabric importing from here, but having the import statement over the env/node file create a value with setEnv that nullify the browser env. Is that correct?

asturur:
> This PR ןד ready apart from TS not recognizing

i can get the node types recognized if i used them from their folder right away.

ShaMan123:
[b326dd6](https://github.com/fabricjs/fabric.js/pull/8632/commits/b326dd6033aac1a285daed1ae090736aaff8a366) fixes `fabric/node` types

asturur:
regarding isLikelyNode, i can. count only 4 usage in the codebase.

Webglprobe, objectCaching, parseSVG, and cleanUpJsdomNode

Now my opinion is that cleanUpJsdomNode and webglProbe and objectCaching are env dependendant, and if we don't like the boolean, getEnv could return:
- EnvCleanUp ( that is a noop in the brower env )
- shouldCacheByDefault a boolean
- webglProbe would be a getEnv function that is different for browser and node.

parseSVG should be tested to see if this bug is still actual or isn't anymore.

Do you have a totally different idea?

ShaMan123:
I wanted to add some comments. I will in the follow up.

I agree with what you suggest. 
WebGL is browser only. So it shouldn't even exist in the node entry point.
We can mark files with a `.browser.ts` or `.node.ts` (shared files stay as is) extension and start having different imports.

Env cleanup is just a jsdom thing. I vote to remove jsdom and extend the node env to support what we really need. 
But if you disagree moving the cleanup to env is great.
Even more, we can look into what I did with Canvas, moving the dispose/destroy methods of Canvas/Image to node subclasses. I looked into it in this PR and decided not to go into it for scope reasons, but is doable and pretty simple. 

Renaming the boolean is good also, but I prefer removing it. For tests we can add a global flag in the config files.

PR #8652, opened 2023-02-04
patch(): node build follow up

ShaMan123:
## Motivation

Follow up #8632 

## Description

The dirty part of #8632

## Changes

- `cleanUpJsdomNode` => `getEnv().dispose`
- `getEnv().window` => `getWindow` (done for ease of coding, finding etc. and one day perhaps when we split browser logic we can use window directly since there really is no use for it in node)
- `getEnv.document` => `getDocument` (same)
- refactor WebGLProbe and expose NodeGLProbe (a stub) both under GLProbes, now part of env
- remove `isLikelyNode`: tests use `isNode()` defined in config files
- remove `setEnvForTests` => use `setEnv` directly


src/parser/parseSVGDocument.ts
```diff
@@ -41,26 +41,13 @@ export function parseSVGDocument(doc, callback, reviver, parsingOptions) {
...
-    // we're likely in node, where "o3-xml" library fails to gEBTN("*")
-    // https://github.com/ajaxorg/node-o3-xml/issues/21
-    descendants = doc.selectNodes('//*[name(.)!="svg"]');
-    const arr = [];
-    for (i = 0, len = descendants.length; i < len; i++) {
-      arr[i] = descendants[i];
-    }
-    descendants = arr;
-  }
-
```
ShaMan123, line 63: didn't tests this yet

src/util/animation/AnimationFrameProvider.ts
```diff
@@ -1,28 +1,9 @@
...
- * In order to get a precise start time, `requestAnimFrame` should be called as an entry into the method
- * @param {Function} callback Callback to invoke
- */
 export function requestAnimFrame(callback: FrameRequestCallback): number {
-  if (!_requestAnimFrame) {
-    _requestAnimFrame =
-      getEnv().window.requestAnimationFrame ||
-      function requestAnimationFramePolyfill(callback: FrameRequestCallback) {
-        return getEnv().window.setTimeout(callback, 1000 / 60);
-      };
```
ShaMan123, line 17: I am actually for having the polyfill and eventually dumping jsdom

ShaMan123 (commented):
- `cleanUpJsdomNode` => `getEnv().dispose`
- `getEnv().window` => `getWindow`
- `getEnv.document` => `getDocument`
- refactor WebGLProbe and expose NodeGLProbe (a stub) on env
- remove `isLikelyNode`: tests use `isNode()`
- remove `setEnvForTests` => use `setEnv` directly

src/env/node.ts
```diff
@@ -20,31 +20,37 @@ const { window: virtualWindow } = new JSDOM(
       FetchExternalResources: ['img'],
```
ShaMan123, line 20: I don't think this is needed, it seems to be a relic of the past

ShaMan123 (commented):
Good thing I held back and waited with all this 
A lot of ink for very little, basically a no brainer

src/env/node.ts
asturur: it could still be needed to actually load base64 images.
Since canva is not mandatory for jsdom, loading images is completely optional.
That is my understanding, if you want them to actually be loaded, you have to ask for it

src/util/animation/AnimationFrameProvider.ts
asturur: dumping JSDOM for now gives us zero benefits, if we find a different backend renderer that does not need it great.
Fabric once had different code for node and web and i don't want to go back there.

src/env/index.ts
```diff
@@ -12,8 +21,3 @@ export const getEnv = () => env || getBrowserEnv();
 export const getDocument = (): Document => getEnv().document;
 
 export const getWindow = (): Window => getEnv().window;
-
-export const setEnvForTests = (window: Window) => {
-  env.document = window.document;
-  env.window = window;
-};
```
asturur, line 19: this was specific for tests, for now setEnv is a necessity of delaying an import / definition issue, and i don't think it is a feature yet. I would like this to stay setEnvForTests ( and even that was some kind of issue of testing and not a real need ) and setEnv to not be exposed

src/env/index.ts
```diff
@@ -1,3 +1,12 @@
+/**
+ * This file is consumed by fabric.
+ * The `./node` and `./browser` files define the env variable that is used by this module.
+ * The `./node` module sets the env at import time.
+ * The `./browser` module is defined to be the default env and doesn't set the env at all.
+ * This is done in order to support SSR.
```
asturur, line 6: instead of referring to SSR that is a specific case and nomenclature of React, let's generally say `isomorphic code` that is how someone defined it broadly. 
https://en.wikipedia.org/wiki/Isomorphic_JavaScript

src/env/node.ts
```diff
@@ -1,17 +1,17 @@
...
 import { TCopyPasteData, TFabricEnv } from './types';
 
-const { implForWrapper: jsdomImplForWrapper } = utils1;
-const { Canvas: nodeCanvas } = utils2;
+const { implForWrapper: jsdomImplForWrapper } = utils;
 
 const copyPasteData: TCopyPasteData = {};
 
-const { window: virtualWindow } = new JSDOM(
+const { window: fabricWindow } = new JSDOM(
```
asturur, line 14: do we want to call this JSDOMWindow since this file is exactly node only?

src/filters/GLProbes/WebGLProbe.ts
```diff
@@ -37,26 +29,20 @@ class WebGLProbe {
 
   /**
    * query browser for WebGL
-   * @returns config object if true
    */
-  queryWebGL() {
-    if (getEnv().isLikelyNode) {
-      return;
-    }
+  queryGL() {
```
asturur, line 33: why renaming WebGL to GL? 
queryWebGL is meant to detect support for WEBGL, i think the old name is good.
GLPrecision change i think is ok, since the constant is stil GL_PRECISION_XXXXX in the WEBGL api.

src/parser/parseSVGDocument.ts
asturur: I think that if SVG import test pass, it is ok.
The issue was not finding `getElementsByTagName('*')`

asturur:
Seems all good to me, apart some naming issues.
For me setEnv vs setEnvForTests is imporant.
setEnv is not a feature for now, even Env was kind of an experiment itself for the lack of not knowing what to do with fabric object and HEADER.js file.
So let's not expose it outside the context of tests.

src/env/node.ts
ShaMan123: I think it is dead.
Not typed and I searched the repo and couldn't find it.

src/env/index.ts
ShaMan123: I was thinking we should expose `setEnv` for devs who work on a strange env, someone wrote about it in #8208

src/filters/GLProbes/WebGLProbe.ts
ShaMan123: because I saw there is support for GL on node canvas...
so a dev could somehow wire it up with fabric and use a NodeGLProbe

src/util/animation/AnimationFrameProvider.ts
ShaMan123: I am backing off this, but 

> Fabric once had different code for node and web and i don't want to go back there.

That is the exact point of Env, a minimal diff in a common exposed interface.
That is why I think one day JSDOM can fade away.
But is not important.

ShaMan123:
> For me setEnv vs setEnvForTests is imporant.
> setEnv is not a feature for now, even Env was kind of an experiment itself for the lack of not knowing what to do with fabric object and HEADER.js file.
> So let's not expose it outside the context of tests.

How?

src/filters/GLProbes/WebGLProbe.ts
ShaMan123: Anyways since we need a probe for node I had to rename the base class/folder

ShaMan123 (commented):
Remaining
- [x] GL naming
- [ ] setEnv for tests

apart that is good to go
You can change as you see fit and merge

src/env/index.ts
asturur: in case we will have specific pr with new feature for it after evaluating which are the side effects.

src/env/node.ts
asturur: yes is not more there indeed.
It is replaced by resources 'usable' we have below.

src/filters/GLProbes/WebGLProbe.ts
asturur: The npm module is called `gl` but is still the WebGL specs api.

asturur:
This PR introduces a circular dependency.
For rollup is just a warning, but we need to find a way to block build on those.
Unbundled code wouldn't work with a circular dependency

ShaMan123:
let me look

test/node_test_setup.js
```diff
@@ -82,7 +82,7 @@ var virtualWindow = new jsdom.JSDOM(
     resources: new CustomResourceLoader(),
```
ShaMan123, line 82: Do we really need this?
Is just a console log...
And then we don't need to reconfigure the env so we won't need to expose setEnvForTests as well

ShaMan123:
Regarding the import cycle
Truth is we can simply import the browser env directly from the WebGLProbe since it runs only on browser env

asturur:
yes is not an issue right now we just need a way to detect them properly

ShaMan123:
OR we can make initFilterBackend async and do a dynamic import of the GLProbe depending on getEnv.
This will allow us to remove the probe from the env

test/node_test_setup.js
asturur: we won't need to expose it in fabricJS, nor remove it, we will probably be able to import it directly from the build folder.

ShaMan123:
> Regarding the import cycle Truth is we can simply import the browser env directly from the WebGLProbe since it runs only on browser env

That is wrong! Since then we init the browser env at import time

ShaMan123:
We can safely use `document` though

asturur:
we will find a solution, the unbundle import is just a playground now, and is node only, so not an issue.

asturur (approved):

test/node_test_setup.js
ShaMan123: I mean we don't need this configuration

ShaMan123: The CustomResourceLoader justs logs to console a line
It is meaningless

ShaMan123: So we can use the node env from fabric w/o this headache

PR #8654, opened 2023-02-04
Passing`propertiesToInclude` to `toObject` to export custom fields fr…

zhangshine:
…om Gradient and Pattern

## Motivation
I want to export custom properties from Gradient and Pattern.

## Changes
Passing `propertiesToInclude` to `toObject`


ShaMan123:
AFAIK all Gradient/Pattern/Shadow props are exported in `toObject` so I don't see why this is needed.
Are you subclassing any?

zhangshine:
No, we are not subclassing any. We pass additional properties when constructoring a new object, and export these properties. Everything works as normal except Gradient and Pattern.

asturur:
Seems ok to me, we would need a small test and a backport to v6 eventually.
The issue with this approach is that unless for simple properties like `id` or `name` properties that are included in a gradient/pattern may not be included in an object and so causing a bunch of undefined keys on the object side or the Gradient/Pattern side.
With a single argument for the export there isn't much alternative.

asturur:
Please add a test and fix lint.

I also need to think if this is a breaking change.

ShaMan123:
Funny, it is a topic we touched.
The strict approach would be not allowing to assign arbitrary props on an instance and subclassing if in need.
I side this approach but I think this PR is a viable solution as long as we allow non strict assignment of props in the constructor.

zhangshine:
Done. Also removed some duplicated codes.

ShaMan123:
I think v6 should drop the entire concept of `propsToInclude` (and repeating myself, remove default values of course)
`toObject` should do what it does strictly and the dev can subclass, iterate and do what they like.

For v5 this makes sense

PR #8659, opened 2023-02-05
docs(): add a command that creates docs from TS annotations

asturur:
This comand is a substitute of the build_docs we had on fabricjs.com
It creates html pages similar to http://fabricjs.com/docs/

It looks liks this
image

And i didn't look at custom themes yet, we need to be able to build something that blends with the new website at some point


ShaMan123:
Is there a decent search?

ShaMan123:
doesn't docosourus et al do that built in?

ShaMan123:
I really love react-boostrap's website
Looking at the repo this came up 
https://github.com/react-bootstrap/react-bootstrap/blob/4bb44894cf10dec511931a243ccb47e742107fe8/package.json#L39

ShaMan123:
more reasearch
- https://www.docz.site/ - looks good since it builds from md files 
- https://tsx-docs.vercel.app/ - looks bad

I couldn't understand how next.js does their docs which are also great

ShaMan123:
I have no idea how it takes code and makes a website

asturur:
I just removed fabricjs test website to the main branch we had before, 3 hours ago.
I don't see docosaurus so much more flexible than gatsby.
I also don't see a search in this package, exactly as it wasn't in the old one ( there was a mix with the google api search keys that do not exist anymore since long ).
Maybe there is also a gatsby plugin that do this, meanwhile i found this.

asturur:
pulling this in, while is probalby not needed in the long run, is still useful to verify that when we move large chunks of documentation around ( like in the interface pr ) this is still able to build.
At some point it will be deleted is not a feature nor is supposed to be consumed by the public at this stage

PR #8661, opened 2023-02-05
chore(): index files

ShaMan123:
## Description

index files for exporting and tree shaking

## Changes

- index files
- default controls refactor: split into a file for exports and a file for side effects


ShaMan123:
I will remove the test to a dedicated PR

ShaMan123 (commented):
I would like to move classREgistry from util to src and out of the utils export scope since it doesn't belong there, it is not a util
And I want to rename controlsUtils to controls

Awaiting Green light

ShaMan123 (commented):
ready

fabric.ts
```diff
@@ -1,425 +1,57 @@
...
-  getWindow,
-  setEnvForTests,
-  parseStyleAttribute,
-  parsePointsAttribute,
-  parseFontDeclaration,
-  parseTransformAttribute,
-  getCSSRules,
-};
+} from './src/filters';
+export * as filters from './src/filters';
```
asturur, line 57: doesn't this re-export also the one we exported above?

fabric.ts
```diff
@@ -1,425 +1,57 @@
...
+export { IText } from './src/shapes/IText/IText';
+export { Textbox } from './src/shapes/Textbox';
+export { Group } from './src/shapes/Group';
+export { ActiveSelection } from './src/shapes/ActiveSelection';
+export { Image } from './src/shapes/Image';
+export { createCollectionMixin } from './src/Collection';
+
+export * as util from './src/util';
+
+export * from './src/parser';
```
asturur, line 45: Not being a named export this seems just hides visibility. Can we see the named exports here?

fabric.ts
```diff
@@ -1,425 +1,57 @@
...
+
+export { Control } from './src/controls';
+export * as controlsUtils from './src/controls';
```
asturur, line 48: as for filters, isn't this a doulbe export? controlsUtils.Control and Control?

asturur:
> I would like to move classREgistry from util to src and out of the utils export scope since it doesn't belong there, it is not a util And I want to rename controlsUtils to controls
> 
> Awaiting Green light

We can move classRegistry around, but control utils are controls utils, aren't controls per se.
Also

asturur:
it still creates those things here:

```js
var index = /*#__PURE__*/Object.freeze({
  __proto__: null,
  BaseFilter: BaseFilter,
  BlackWhite: BlackWhite,
  BlendColor: BlendColor,
  BlendImage: BlendImage,
  Blur: Blur,
  Brightness: Brightness,
  Brownie: Brownie,
  Canvas2dFilterBackend: Canvas2dFilterBackend,
  ColorMatrix: ColorMatrix,
  Composed: Composed,
  Contrast: Contrast,
  Convolute: Convolute,
  Gamma: Gamma,
  Grayscale: Grayscale,
  HueRotation: HueRotation,
  Invert: Invert,
  Kodachrome: Kodachrome,
  Noise: Noise,
  Pixelate: Pixelate,
  Polaroid: Polaroid,
  RemoveColor: RemoveColor,
  Resize: Resize,
  Saturation: Saturation,
  Sepia: Sepia,
  Technicolor: Technicolor,
  Vibrance: Vibrance,
  Vintage: Vintage,
  WebGLFilterBackend: WebGLFilterBackend,
  getFilterBackend: getFilterBackend,
  initFilterBackend: initFilterBackend
});
```
So i m not sure how this is different from what we have now.

fabric.ts
ShaMan123: yes you are right
should the above belong to filters and be exported as standalone? I think it should

fabric.ts
ShaMan123: I don't mind but is missing the whole point of moving it to a standalone index
I can return them here

fabric.ts
ShaMan123: > controlsUtils.Control and Control?

This is part of why I want to rename so then `controls.Control` `controls.wrapWithXXX`
A nice scoped namespace

ShaMan123:
For direct import it makes a difference for sure
FYI: `/*#__PURE__*/` this annotation is described in rollup tree shaking to instruct the bundler not to treat it as a side effect

fabric.ts
ShaMan123: I want them to be
`export * as parser from './src/parser';`

ShaMan123 (commented):
renamed `cotrols.render` => `controlRendering`
renamed and move class registry

PR #8662, opened 2023-02-05
chore(TS): Move cache properties as static

asturur:
## Motivation

Part of the effort of getting rid of prototype fiddling.

## Description

Just moving those as static arrays
No behaviour changes, what was possible before is possible now too.
You can't have anymore cache properties that are different between instances of the same class, but that was never meant to have any sense, it was just a side effecto of being on the prototype and so on the instance too.

Stateful tests are failing, i m opening a different pr that will remove stateful.


asturur:
i probably need to add cacheProperties to every class now. Didn't think of it.
I'll do when  https://github.com/fabricjs/fabric.js/pull/8663 is merged

asturur:
Can i go with this? or is there something we want to tweak?

ShaMan123 (commented):
What about stateProps?

asturur:
> What about stateProps?

another PR.

PR #8663, opened 2023-02-05
chore(TS): Remove stateful mixin

asturur:
## Motivation

The Stateful mixin had 2 defects.
1) it tried to detect change on objects on the fly ( complex and not good for performances )
2) It had some weird comparing code we don't want to maintain

The direction

## Description
- remove stateful mixin and its tests

## Changes
- renamed _dimensionAffectingProps to textLayoutProperties, static.
- changes to textLayoutProperties require the use of set, exactly as cache properties


src/shapes/IText/IText.ts
```diff
@@ -149,6 +150,8 @@ export class IText<
    */
   declare caching: boolean;
 
+  static textLayoutProperties = Text.textLayoutProperties;
```
ShaMan123, line 153: no need for this
static props are inherited

src/shapes/Text/Text.ts
```diff
@@ -79,6 +79,22 @@ const additionalProps = [
   'pathAlign',
 ] as const;
 
+const textLayoutProperties: string[] = [
```
ShaMan123, line 82: why not define it directly on the class?

ShaMan123 (commented):
Looks great
Can stateProps and cacheProps become static as  well??

src/shapes/Text/Text.ts
```diff
@@ -1690,11 +1703,11 @@ export class Text<
           this.setPathInfo();
         }
         needsDims =
-          needsDims || this._dimensionAffectingProps.indexOf(_key) !== -1;
+          needsDims || this.constructor.textLayoutProperties.includes(_key);
         isAddingPath = isAddingPath || _key === 'path';
       }
     } else {
-      needsDims = this._dimensionAffectingProps.indexOf(key) !== -1;
+      needsDims = this.constructor.textLayoutProperties.includes(key);
```
asturur, line 1710: This thing here remind me of te issue we have with side effects.
sometimes those fit in _set, but if those are slightly expensive we want to run them only once for set.
But this bring us to handle in set both single and multi key use case.
@ShaMan123  should check your side effects pr solves this issue with 10 lines of code or find a valid way to remove those with a little footprint.

src/shapes/IText/IText.ts
ShaMan123: I see #8662

src/shapes/Text/Text.ts
ShaMan123: don't think it is 10 lines but it solves this bad design

src/shapes/Text/Text.ts
asturur: i like the single line of the claration

src/shapes/IText/IText.ts
asturur: are they? my understand is that they are inherited only in static methods.
If is not needed even better.

ShaMan123:
Another thought
We might want to get rid of static fields because they compile to a side effect and a method doesn't or is it a matter of fine tuning rollup/browserlist?

src/shapes/IText/IText.ts
ShaMan123: image

```js
class A {
 static m=5
}
class B extends A {}
console.log(B.m)
```

asturur:
they compile to a sideEffect you mean they become
``` Object.prop = value ``` ?
I m going to check what the output is like.

Seems like classes aren't ready if i have to create a getter for a prop because they are handled differently.
It also leaves me with the issue of creating a setter too.

asturur:
yes so they become side effects.
When we get to code tree shaking we have to decide.
It could also be that we leave the lib unbundled for who wants to optimize the bundle and imports the single files.
The other will use a large premade bundle built with rollup.

I m kind of tired to spinning in circle to js/ts issues
This migration definitely took a toll on me.

asturur:
This is ready to go, not tackling side effects now, the scope is removing stateful.

ShaMan123 (approved):
Yes
The maintenance is a really unpleasant part.
I looked at the links of tree shaking in rollup
https://rollupjs.org/configuration-options/#treeshake-modulesideeffects

We can expose a rollup config for consumers or whatever
Never mind that lets move fabric forward
It has taken a toll on me too

Issue #7670, opened 2022-02-10
`V6`: 🎉 Group Rewrite


ShaMan123:
## V6
Hi everyone,
**fabric is moving forward!**
We have 2 main goals for v6:
1. js/ts migration which is covered here #7596 
2. Group rewrite 🎉

You can see #8316

## Group Rewrite
**This is a tracking issue for the group rewrite**, read below for more detail including the current state of things and gist.
We aim to fix and support long desired features including:
1. Nested selection - **DONE**
2. Resizing and customized layout - **DONE**
4. Transform and coordinate system (which is relative to the parent group) - **DONE**

### Motivating Issues/Discussions
#7473 https://github.com/fabricjs/fabric.js/pull/7316 https://github.com/fabricjs/fabric.js/issues/6776 https://github.com/fabricjs/fabric.js/discussions/7136 https://github.com/fabricjs/fabric.js/discussions/7299 https://github.com/fabricjs/fabric.js/issues/7130 https://github.com/fabricjs/fabric.js/issues/7142 #7449 and anything with a `group` tag

### How do you fit in?
1. You can become an alpha tester (we aim to publish an alpha tag to npm regularly). 
**IMPORTANT NOTICE**: all changes made to the v6 branch are experimental and can be removed or changed without notice and without proper support (including supported features from v5). fabric might change drastically between commits. Meaning if you want in - it's on you.

You can join the effort by [**contributing**](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)

**This description is updated continuously with relevant information**
You don't need to scroll and read endless comments (unless you are a maintainer :))

---

## Current State
Fixed logic across fabric for nested objects to be selectable.
Group has been completely rewritten and Layer has been introduced. As part of the rewrite group is now a layout engine, you can use it to customize layout of objects.
In general the rewrite allows any object to be selectable and interactive if it is referenced in the `_objects` array of it's parent and it references it's parent via the `group` property. If you build custom objects with nesting this is what you are looking for.

### Usage
It is advised to use `set` on group/objects from now on (until #7596 is done, then it might change due to setters)
Changing `subTargetCheck` or `interactive` enables/disables crucial functionality so for these props it is not an advice but a **MUST**.
```js
new fabric.Group(objects, {
  subTargetCheck: true,
  interactive: true   //  enables selecting subtargets
});
```

### Progress 
- [x] #7858 
- [x] #7859 
- [x] #7916 
- [x] #8000 
[...]

PR #8665, opened 2023-02-05
fix(): selection logic to support nested multiselection

### LINKED ISSUE #4529: Preserve order of selected objects in multiple selection

### LINKED ISSUE #6840: Object order changes when grouping items selected individually (shift-clicked) with preserveObjectStacking: true

ShaMan123:
## Motivation

- fix selection logic to support nested multiselection + object stacking control on ActiveSelection
- make selection logic more scoped and orderly (I have future thoughts making it an interaction brush so scoping logic is great), it can be extracted once this PR is merged easily to a standalone manager
- make canvas scoped so in the future stuff can be extracted to standalone modules and it can be reduced in size and complexity

ports and closes #7878
ports logic from #7882 not including rendering logic that will be done after this PR

closes #6840 
closes #4529 

### Selection 
This PR refactors what was known as canvas grouping logic (formerly the canvas grouping mixin).
We are talking about a bunch of methods that were in charge of selecting objects in response to user events.
Selection is distinct into 2.
What is known as **multiselection (shift+click)** and what is known as **selection/group selection (drag)** defined as `_groupSelector` (the blue rect that selects objects)
Logic was scattered and unscoped (methods shared logic,  e.g https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097929354) for some reason, what made this effort difficult, confusing and time consuming. This is why I squashed them as follows:
- `handleSelection` (renamed `_maybeGroupObjects`): now [called before `_shouldClearSelection`](https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097610998) , [not calling `setCursor`](https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097930423)
   - `_collectObject`: [moved `collectObjects` logic to collection](https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097933666) (moved some tests as well)
   - `_groupSelectedObjects`
   - the condition to run this logic block from mouse up
- `handleMultiSelection` (renamed `_handleGrouping`): use `searchPossibleTargets` instead of hacking `findTarget` 30ee763
   - `_shouldGroup`: The sensitive commit 9a0736a625df10f6377496c7f29240ab49eb30e6 moved the check `activeObject.__corner` to `_shouldGroup`, see [comment](https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097934238)
   - `_updateActiveSelection`: ~~[object stacking question to be resolved](https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1098187072)~~ https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097924017
   - `_createActiveSelection` https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097922604

I updated the logic to act with the constant active selection ref (see below) - https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097931465.
So basically **the result of this PR is 2 methods that are solely in charge of selection logic**.
I decided to rename the main methods because:
- naming was unclear (I still think it is, suggestions are welcome)
- since I squashed the condition to run the logic into the main method, whoever is using it directly (poor unfortunate fellow) might experience a break of the method being skipped because of the condition and not understanding why. However these are private methods so I regret to tell this dev, tough luck, or actually good fortune, since now I think it is much readable and therefore reliable.
- renaming make a breaking change so less breaking. It breaks the app but shows up everywhere and then the dev must read migration docs and they see what happened and how to move forward as opposed of having their app continue working but then bugs appear and edge cases etc.

### Active Selection Ref
Made canvas hold a constant ref to active selection. This is done to allow devs to listen to events and handle it as they wish, what was impossible until now (you could subscribe after an active selection was created). It also allows to subclass active selection and make canvas use it.
https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1097562518

- replaced most of `isActiveSelection` calls with a simple `obj === canvas._activeSelection`

### MultiSelection Stacking Order

- exposed `multiSelectStacking: 'canvas-stacking' | 'selection-order'` that controls how multiselected objects are added to the active selection. Logic is handled by `ActiveSelection#multiSelectAdd`
- fixed the bug from `_updateActiveSelection`
- refactored `findTarget(e, skipGroup) => FabricObject` => `findTarget(e) => FabricObject`: `skipGroup` is not needed anymore, it was buggy so now calling `searchPossibleTarget` directly and scoped internal logic

30ee763

### Other Changes

- fixed ancestry methods to comply with active selection in the tree
- simplified `_chooseObjectsToRender`
- renamed `_groupSelector` keys for clarity 58c050d45782bf140487e4dfa3ee962bef78786d
- added a [minimal rendering change to Group](https://github.com/fabricjs/fabric.js/pull/8665#pullrequestreview-1286130059) in order to test the PR
- removed dead prop `_previousPointer`
- disabled node assertion config fd145fe, it was passing false positives

## Gist

TO DO
- [x] https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1098187072 30ee763


ShaMan123 (commented):
good

src/canvas/Canvas.ts
```diff
@@ -1096,12 +1107,12 @@ export class Canvas extends SelectableCanvas {
     const pointer = this.getPointer(e, true);
     // save pointer for check in __onMouseUp event
     this._previousPointer = pointer;
-    const shouldRender = this._shouldRender(target),
-      shouldGroup = this._shouldGroup(e, target);
+    const shouldRender = this._shouldRender(target);
+    let shouldGroup = false;
```
asturur, line 1111: how do we know that this was always false in all possible combinations?
If this is a micro optimization ( don't run should group if i should clear selection ) i would not do it.
or i would carefully check the commit history before changing it.

src/canvas/Canvas.ts
```diff
@@ -1495,158 +1509,99 @@ export class Canvas extends SelectableCanvas {
...
-        (allowIntersect &&
-          currentObject.containsPoint(selectionX2Y2, undefined, true))
-      ) {
-        group.push(currentObject);
-        // only add one object if it's a click
-        if (isClick) {
-          break;
-        }
-      }
-    }
```
asturur, line 1631: comments like: `moved this in a collection mixin called `collectObject`` Help me save entire minutes.

src/Collection.ts
```diff
@@ -304,6 +305,36 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
 
       return newIdx;
     }
+
+    /**
```
asturur, line 309: Given a boundingbox, return all the objects of the collection that are contained in the box.
If includeInterseting is true, also the objects that are across the bounding box are selected.

asturur (changes requested):
revert the shouldGroup boolean detection order if it there isn't a specific case that is fixing or needs it.

src/canvas/Canvas.ts
ShaMan123: I didn't write the description yet
sorry

src/Collection.ts
ShaMan123: what about naming?

asturur: i Don't have an opinion here. collectObjects if there is an explanation written on top is good enough.

asturur: i thought it was ready. I will re-review.

asturur:
I want to be clear that many times i said that having the group:
- able to select a single object
- able to adjust the bouding box without resetting it
- able to be automatically updated on add and remove
Was a big and good enough change for 6.0 and we should have stopped there.

src/canvas/SelectableCanvas.ts
```diff
@@ -490,6 +494,12 @@ export class SelectableCanvas<
   protected declare _isCurrentlyDrawing: boolean;
   declare freeDrawingBrush?: BaseBrush;
   declare _activeObject?: FabricObject;
+  protected readonly _activeSelection: ActiveSelection;
```
asturur, line 497: is this necessary? or is just avoiding us to create an ActiveSelection when needed?
I would rather create on need rather than right away.
Is it a matter of code complexity or there is more?

ShaMan123: It is meant for controlling the stack order of active selection
https://github.com/fabricjs/fabric.js/issues/6840
https://github.com/fabricjs/fabric.js/issues/4529

ShaMan123:
> Was a big and good enough change for 6.0 and we should have stopped there.

Why? I want group to function as expected and I already achieved it. Now it needs porting and merging.
It is the finish line of Group

src/canvas/Canvas.ts
ShaMan123: I looked into this very closely.
Let me explain.

    shouldGroup = this._shouldGroup(e, target);
    if (this._shouldClearSelection(e, target)) {
      this.discardActiveObject(e);
    } else if (shouldGroup) {
      // in order for shouldGroup to be true, target needs to be true
      this._handleGrouping(e, target!);
      target = this._activeObject;
    }

`_shouldClearSelection` and `_shouldGroup` or opposites of each other

https://github.com/fabricjs/fabric.js/blob/84b7a357cef0c899ecd6b3fc0a74de2d57283d21/src/canvas/SelectableCanvas.ts#L725
https://github.com/fabricjs/fabric.js/blob/a39e01cf5ceae66809272051f77319345d8492d2/src/canvas/Canvas.ts#L1477

The whole point of this change is to squash them together. The logic is too complex and scattered around everywhere

ShaMan123: so when you agree I will squash them here

src/canvas/SelectableCanvas.ts
asturur: You linked to feature requests for this feature of which i m not against fixing.
I guess i m asking, why the _activeSelection needs to be always created and added to the canvas? it does not seem we are leveraging that anywhere, we could create it when is needed? or something is blocking us?

asturur:
> > Was a big and good enough change for 6.0 and we should have stopped there.
> 
> Why? I want group to function as expected and I already achieved it. Now it needs porting and merging. It is the finish line of Group

The fact that developers wants a feature and you are able to write it, it does not mean that it needs to get done.
Every addition makes the code more and more complex and new in behaviour, that is why i want to do things in stages.
Now let's move on but every time i say stop refactors/ additions and we continue merging new things.

src/canvas/SelectableCanvas.ts
asturur: We are not leveraging the fact that the activeSelection is prebuilt.
We save maybe a check to create it but then we add a check to empty it on selection cleared.
Where is the benefit?

asturur: I m also thinking, how do i keep the click order if i don't want the stack order? i can't see where preserve object stacking actually makes the decision?

ShaMan123: The benefits:
```ts
const a = canvas.getActiveSelection()
a.on('object:added', ({target}) =>a.sendObjectToBack(target))

```
But this might be achieved with selection events

The other benefit is the ability to set the instance, meaning the dev can subclass active selection and use it w/o turning core upside down

ShaMan123: It is not mandatory to initialize it in constructor. I can do it when needed

ShaMan123:
For me is part of the same effort #7670
I want group to work as I expect it w/o spending days on transformation logic and impossible interaction logic and hacking fabric to the point of me rewriting it.

src/canvas/SelectableCanvas.ts
ShaMan123: > I m also thinking, how do i keep the click order if i don't want the stack order? i can't see where preserve object stacking actually makes the decision?

can you explain more? I can understand this in a few ways

ShaMan123: Never did active selection have a way to control how objects are stacked in it

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
+            ? [activeObject, target]
+            : [target, activeObject])
+        );
+
+        this._hoveredTarget = this._activeSelection;
+        // ISSUE 4115: should we consider subTargets here?
+        // this._hoveredTargets = [];
+        // this._hoveredTargets = this.targets.concat();
+        this._setActiveObject(this._activeSelection, e);
+        this._fireSelectionEvents([activeObject], e);
```
ShaMan123, line 1545: squashed `_createActiveSelection`

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
+          const insertAt =
+            index === -1
+              ? //  `target` is in front of all other objects
+                activeObject.size()
+              : index;
+          activeObject.insertAt(insertAt, target);
+          this._hoveredTarget = activeObject;
+          this._hoveredTargets = [...this.targets];
+        }
+        this._fireSelectionEvents(prevActiveObjects, e);
```
ShaMan123, line 1530: squashed `_updateActiveSelection`

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
-      return;
-    }
-    if (groupingTarget === activeObject) {
-      // if it's a group, find target again, using activeGroup objects
-      groupingTarget = this.findTarget(e, true);
-      // if even object is not found or we are on activeObjectCorner, bail out
-      if (!groupingTarget || !groupingTarget.selectable) {
-        return;
-      }
-    }
```
ShaMan123, line 1513: this is why `activeObject.type === 'activeSelection'`
https://github.com/fabricjs/fabric.js/blob/a39e01cf5ceae66809272051f77319345d8492d2/src/canvas/Canvas.ts#L1481-L1483

ShaMan123: moved it under `_handleGroup` => `isActiveSelection` block

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
+    ) {
+      if (isActiveSelection(activeObject)) {
+        if (target === activeObject) {
+          // find target from active objects
+          target = this.findTarget(e, true);
+          // if nothing is found or we are on activeObject's corner, bail out
+          if (!target || !target.selectable) {
+            return false;
+          }
+        }
```
ShaMan123, line 1503: here it is
sub target check

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
-      // activeObject or if an activeSelection already exists
-      // TODO at time of writing why `activeObject.type === 'activeSelection'` matter is unclear.
-      // is a very old condition uncertain if still valid.
-      (activeObject !== target || activeObject.type === 'activeSelection') &&
-      //  make sure `activeObject` and `target` aren't ancestors of each other
-      !target.isDescendantOf(activeObject) &&
-      !activeObject.isDescendantOf(target) &&
+      // group target and active object only if they are different objects
+      // else we try to find a subtarget of `ActiveSelection`
+      (activeObject !== target || isAS) &&
```
ShaMan123, line 1484: and here is the condition

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
-      activeSelection.remove(target);
-      this._hoveredTarget = target;
-      this._hoveredTargets = this.targets.concat();
-      if (activeSelection.size() === 1) {
-        // activate last remaining object
-        this._setActiveObject(activeSelection.item(0) as FabricObject, e);
+      !target.onSelect({ e }) &&
+      // make sure we are not on top of a control
+      !activeObject.__corner
+    ) {
```
ShaMan123, line 1494: squashed `_shouldGroup` because of the side effect and weird condition for sub target check

src/canvas/Canvas.ts
```diff
@@ -1433,7 +1431,7 @@ export class Canvas extends SelectableCanvas {
         : null,
       // only show proper corner when group selection is not active
       corner =
-        (!activeSelection || !activeSelection.contains(target)) &&
+        (!activeSelection || target.group !== activeSelection) &&
```
ShaMan123, line 1434: perf

src/canvas/Canvas.ts
```diff
@@ -835,7 +826,17 @@ export class Canvas extends SelectableCanvas {
...
+        const objects = this._collectObjects(e);
+        // do not create group for 1 element only
+        if (objects.length === 1) {
+          this.setActiveObject(objects[0], e);
+        } else if (objects.length > 1) {
+          this._activeSelection.add(...objects);
+          this.setActiveObject(this._activeSelection, e);
+        }
+      }
+      this._groupSelector = null;
```
ShaMan123, line 839: squashed `_groupSelectedObjects`

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
-      !target.isDescendantOf(activeObject) &&
-      !activeObject.isDescendantOf(target) &&
+      // group target and active object only if they are different objects
+      // else we try to find a subtarget of `ActiveSelection`
+      (activeObject !== target || isAS) &&
+      //  make sure `activeObject` and `target` aren't ancestors of each other in case `activeObject` is not `ActiveSelection`
+      // if it is then we want to remove `target` from it
+      (isAS ||
+        (!target.isDescendantOf(activeObject) &&
+          !activeObject.isDescendantOf(target))) &&
```
ShaMan123, line 1489: fix

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
-    if (target.group === activeSelection) {
-      activeSelection.remove(target);
-      this._hoveredTarget = target;
-      this._hoveredTargets = this.targets.concat();
-      if (activeSelection.size() === 1) {
-        // activate last remaining object
-        this._setActiveObject(activeSelection.item(0) as FabricObject, e);
+      !target.onSelect({ e }) &&
+      // make sure we are not on top of a control
+      !activeObject.__corner
```
ShaMan123, line 1493: danger commit

src/canvas/Canvas.ts
```diff
@@ -1460,193 +1458,126 @@ export class Canvas extends SelectableCanvas {
...
-        }
-      }
-    }
-
-    if (group.length > 1) {
-      return group.filter((object) => !object.onSelect({ e }));
-    }
-
-    return group;
-  }
```
ShaMan123, line 1638: moved to collection

src/canvas/Canvas.ts
ShaMan123: This was tricky to validate.
The change seems to effect the following logic because `_shouldGroup` will return false if `activeObject.__corner` which is a change in behavior and since the code is messy it is hard to say if it is safe. However I will prove it doesn't change anything.

So we assume the following:
1. `activeObject.__corner` is defined
2. The original `_shouldGroup` returns true (if it returns `false` there is no change to behavior)

First of all `_handleGrouping` is skipped regardless of this change so we don't need to discuss that. => `activeObject` was not changed

(1) => `target === activeObject` => calculating `_findTargetCorner` is redundant

```diff
    let shouldGroup = false;
    if (this._shouldClearSelection(e, target)) {
      this.discardActiveObject(e);
-    } else if (this._shouldGroup(e, target)) { // true
+    } else if (this._shouldGroup(e, target)) { // false
+    // this block is skipped
-     shouldGroup = true; // remains false
     // may change active object
-      this._handleGrouping(e, target!); // skipped regardless
-      target = this._activeObject; // skipped `_handleGrouping` => no change to `this._activeObject` => `target === this._activeObject`
    }
+   // conclusion: `shouldGroup` is the only thing changed up to here
    // we start a group selector rectangle if
    // selection is enabled
    // and there is no target, or the following 3 condition both apply
    // target is not selectable ( otherwise we selected it )
    // target is not editing
    // target is not already selected ( otherwise we drage )
    if (
      this.selection &&
      (!target ||
        (!target.selectable &&
          // @ts-ignore
          !target.isEditing &&
          target !== this._activeObject))
    ) {
      const p = this.getPointer(e);
      this._groupSelector = {
        ex: p.x,
        ey: p.y,
        top: 0,
        left: 0,
      };
    }


+    if (target) { // true
+      const alreadySelected = target === this._activeObject;  // true
      if (target.selectable && target.activeOn === 'down') {
+        this.setActiveObject(target, e); // redundant since `target === activeObject`
      }
      const corner = target._findTargetCorner(
        this.getPointer(e, true),
        isTouchEvent(e)
+      ); // redundant since `corner === activeObject.__corner` because `target === activeObject`
+      if (target === this._activeObject && (corner || !shouldGroup)) { // corner is defined regardless
+     // conclusion: in this case `shouldGroup` is not used by the code
        this._setupCurrentTransform(e, target, alreadySelected);
        const control = target.controls[corner],
          pointer = this.getPointer(e),
          mouseDownHandler =
            control && control.getMouseDownHandler(e, target, control);
        if (mouseDownHandler) {
          mouseDownHandler(e, this._currentTransform!, pointer.x, pointer.y);
        }
      }
    }
-    const invalidate = shouldRender || shouldGroup; // was true
+    const invalidate = shouldRender || shouldGroup; // now it can be false, this correct since `shouldGroup` made no changes to state
```

https://github.com/fabricjs/fabric.js/blob/e502e1f74dc782c275875323066c7a06c76287be/src/canvas/Canvas.ts#L1098-L1150

ShaMan123 (commented):
one test to go
updated description

src/shapes/Group.ts
```diff
@@ -483,7 +483,18 @@ export class Group extends createCollectionMixin(FabricObject<GroupEvents>) {
...
+        this.canvas?.preserveObjectStacking &&
+        this._objects[i].group !== this
+      ) {
+        ctx.save();
+        ctx.transform(...invertTransform(this.calcTransformMatrix()));
+        this._objects[i].render(ctx);
+        ctx.restore();
+      } else if (this._objects[i].group === this) {
+        this._objects[i].render(ctx);
+      }
```
ShaMan123, line 497: this is here mainly for testing live
it is not a complete solution
that will be done in a follow
this open up a lot of edge cases

ShaMan123 (commented):
ready to merge

src/canvas/SelectableCanvas.ts
asturur: No it didn't it captured them in click order.
Now it seems that it capture in stack order.
But why one is wrong and one is good?

asturur:
> For me is part of the same effort #7670 I want group to work as I expect it w/o spending days on transformation logic and impossible interaction logic and hacking fabric to the point of me rewriting it.

Yes and i tried to slice the efforct in smaller pieces since start.
My fault i just need stricter guidelines for PRs and features.

asturur:
Before i review the PR can you describe the rendering followup?
i don't want to approve this and then find another PR that is needed to complete this one and we assume is necessary to move on.

src/canvas/SelectableCanvas.ts
ShaMan123: > No it didn't, it captured them in click order.

This is good. This is why we should hold a ref to active selection. So the dev can decide and run custom logic **OR** we could have an option for this on active selection: `selectionStrategy: 'click' | 'preserve-stacking'`.
At the time it made more sense to me to change it. It isn't better.
I would like to add this option.
What do you think?

for reference:
https://github.com/fabricjs/fabric.js/pull/8665/files#diff-47131ad6e0b321305307e39da0359c85513edd9b084e1fe1ef4941eab9e3f262L1536

https://github.com/fabricjs/fabric.js/pull/8665/files#diff-47131ad6e0b321305307e39da0359c85513edd9b084e1fe1ef4941eab9e3f262R1516-R1526

ShaMan123:
> Yes and i tried to slice the efforct in smaller pieces since start.
My fault i just need stricter guidelines for PRs and features.

And I try to do so. But sometimes it is hard and sometimes I get caught up with the fix and want to see it through and I prefer finishing it and then splitting it up because at the time of fixing I don't know what is needed and jumping between branches and updating them while maintaining proper history is time consuming.
I squashed logic from the 2 PRs because they overlapped and left rendering to a follow up since IMO that was the correct scope.
Do you consider this as a large PR? It changes a lot of **complex** logic in Canvas, it is difficult but scoped IMO at least.
I did my best.
Regarding rendering it is as we started discussing in the doc you made. There are issues of caching and rendering responsibility that we need to decide about.
I just pushed a [minimal rendering change to group](https://github.com/fabricjs/fabric.js/pull/8665/files#r1097994173) so it is visible this PR fixes what it claims to do.
For example what happens when running `findTarget` with `perPixelTargetFind` is true and a group with a nested selected object is tested? With this rendering change if you test a point over the selected object (and no other object is drawn there) it will return false but it should return true. This is part of the scope of your proposal of `renderExport`. Same goes to all rendering export methods.
Also if this group is caching it needs to be invalidated and currently it will not since the `group` ref of the object is set to the active selection.
All this was handled by #7882

src/canvas/Canvas.ts
ShaMan123: I am a bit confused because I thought it worked in #7882

src/canvas/Canvas.ts
```diff
@@ -1093,24 +1083,22 @@ export class Canvas extends SelectableCanvas {
...
-      // in order for shouldGroup to be true, target needs to be true
-      this._handleGrouping(e, target!);
+    let shouldRender = this._shouldRender(target);
+    let grouped = false;
+    if (this.handleMultiSelection(e, target)) {
+      // active object might have changed while grouping
       target = this._activeObject;
+      grouped = true;
+      shouldRender = true;
+    } else if (this._shouldClearSelection(e, target)) {
```
ShaMan123, line 1093: switched the order aaa5d1e84

ShaMan123 (commented):
b96d8cd37 - revertable, active selection ref
aaa5d1e84 - done this because it makes sense in light of scoping `_groupSelector` and extracting it from canvas.
58c050d45 - change `_groupSelector` key names, were very confusing
d0473016d - rename methods and squash back logic I extracted in a prev commit

This is now looking like something close to an API

src/canvas/SelectableCanvas.ts
ShaMan123: > is this necessary? or is just avoiding us to create an ActiveSelection when needed? I would rather create on need rather than right away. Is it a matter of code complexity or there is more?

> It is not mandatory to initialize it in constructor. I can do it when needed

I have changed my mind. It is necessary. How would you add an event listener before there is a ref?

src/canvas/Canvas.ts
ShaMan123: https://github.com/fabricjs/fabric.js/commit/aaa5d1e84681491cbd196177f388eacfb65b57a0

ShaMan123 (commented):
I believe this PR is ready for your review
Sorry for the back and forth and mess in reviews.
**I have rewritten the description, please read it through.** I dedicated time and effort to make it readable and helpful.

What remains?
- [ ] I can/should add a visual test of `handleMultiSelection` since it is a sensitive thing
- [ ] active selection stacking

I don't plan on committing any changes accept these remains.
Awaiting your review.

asturur:
yes the complexities of handling multi selection inside groups or different groups are high, and not going to build caching around that.

ShaMan123:
no need to
I #7882 I created a cache invalidation method for groups that have a selected object run by active selection
https://github.com/fabricjs/fabric.js/pull/7882/files#.

test/unit/canvas.js
```diff
@@ -1925,38 +1856,14 @@
...
-    assert.equal(aGroup._objects[3], circle1);
-    canvas.setActiveObject(aGroup)
-    canvas.renderAll();
-    // after rendering objects are ordered in canvas stack order
-    assert.equal(aGroup._objects[0], rect1);
-    assert.equal(aGroup._objects[1], rect2);
-    assert.equal(aGroup._objects[2], circle1);
-    assert.equal(aGroup._objects[3], circle2);
-  });
-
```
ShaMan123, line 1959: @asturur please explain this to me as part of the discussion of click order vs. stack order
I believe this is what convinced me it was a bug, that rendering changed the order of the objects in active selection.
Now with this PR active selection will respect both click order and stack order

ShaMan123: I would say before it didn't respect either

ShaMan123:
> No it didn't, it captured them in click order. Now it seems that it capture in stack order. But why one is wrong and one is good?

@asturur you were wrong, it never was click order. It was canvas stacking order always. Regardless I have promoted the fix to support both and then I added tests and saw this.

https://github.com/fabricjs/fabric.js/blob/ab6b4b481bda9d7c2ce61375b4f1cad5e1f556ed/test/unit/canvas.js#L377-L389

src/canvas/Canvas.ts
```diff
@@ -1457,196 +1445,144 @@ export class Canvas extends SelectableCanvas {
...
+    ) {
+      if (isAS) {
+        const prevActiveObjects =
+          activeSelection.getObjects() as FabricObject[];
+        if (target === activeSelection) {
+          // find target from active objects
+          target = this.searchPossibleTargets(
+            prevActiveObjects,
+            this.getPointer(e, true)
+          );
```
ShaMan123, line 1496: instead of `this.findTarget(e, true)`

src/canvas/SelectableCanvas.ts
```diff
@@ -949,17 +930,16 @@ export class SelectableCanvas<
...
+      aObjects.length >= 1 &&
       activeObject._findTargetCorner(pointer, isTouch)
     ) {
       return activeObject;
     }
     if (
       aObjects.length > 1 &&
-      isActiveSelection(activeObject) &&
-      !skipGroup &&
+      activeObject === this._activeSelection &&
```
ShaMan123, line 942: this condition is a bit silly because if aObjects > 1 we are dealing with the active selection but I think you put that for TS

ShaMan123 (commented):
Alright
Now it is done
You can resolve any conversation you like. I used it to navigate the code

Updated description. I had a fatal error, mixed up the names of the new methods and add more about the stack ordering solution

ShaMan123:
and of course something is wrong

ShaMan123 (commented):
Now with `findTarget` fixed and tested this is done
Was a hard one

asturur:
> > No it didn't, it captured them in click order. Now it seems that it capture in stack order. But why one is wrong and one is good?
> 
> _Originally posted by @asturur in [#8665 (comment)](https://github.com/fabricjs/fabric.js/pull/8665#discussion_r1098051972)_
> 
> @asturur you were wrong, it never was click order. It was canvas stacking order always. Regardless I have promoted the fix to support both and then I added tests and saw this.
> 
> https://github.com/fabricjs/fabric.js/blob/ab6b4b481bda9d7c2ce61375b4f1cad5e1f556ed/test/unit/canvas.js#L377-L389

The multiple selection is in click order.
After you select the first 2 obects, all the others were added in the active selectin inside the array as last one, and that is visible in the bug you linked.

On top of that i can't see anymore the activeSelection.toGroup method. what happened to it?

asturur:
ok toGroup and toActiveSelection have been killed in the group transition for some reason, i will add them back in a separate pr using whatever code we have now to do a similar thing

asturur:
@ShaMan123 cloned this branch locally, i will be offline 2 days reviewing the new selection process.

ShaMan123:
> have been killed in the group transition for some reason, i will add them back in a separate pr using whatever code we have now to do a similar thing

I documented this.
There is no reason for this any longer because there is no overhead logic performed.
```ts
new Group(canvas.getActiveObjects())
```

Why would we expose a method for this?

ShaMan123:
And to add to active selection
```
(canvas.getActiveObject() || canvas.getActiveSelection).add(...group.objects)
```

That is the whole point of a strict object tree

ShaMan123:
> The multiple selection is in click order.
After you select the first 2 obects, all the others were added in the active selectin inside the array as last one, and that is visible in the bug you linked.

For me this is inconsistent behavior. First it does stack order and then click order. That is why I considered it a bug when I saw it.

Another I thing I would like to understand from you is why does the multiselection logic runs on mouse down and not up

src/shapes/Object/InteractiveObject.ts
```diff
@@ -149,11 +149,15 @@ export class InteractiveFabricObject<
...
+   * Determines which corner is under the mouse cursor, represented by `pointer`.
+   * This function is return a corner only if the object is the active one.
+   * This is done to avoid selecting corner of non active object and activating transformations
+   * rather than drag action. The default behaviour of fabricJS is that if you want to trasform
+   * an object, first you select it to show the control set
    * @private
    * @param {Object} pointer The pointer indicating the mouse position
    * @param {boolean} forTouch indicates if we are looking for interaction area with a touch action
-   * @return {String|Boolean} corner code (tl, tr, bl, br, etc.), or false if nothing is found
+   * @return {String|Boolean} corner code (tl, tr, bl, br, etc.), or 0 if nothing is found.
```
asturur, line 160: imho this should be always a string, it makes much more sense. empty string if nothing is found.
Not touching it now

src/shapes/ActiveSelection.ts
```diff
@@ -23,6 +33,36 @@ export class ActiveSelection extends Group {
...
+        const index = this._objects.findIndex((obj) => obj.isInFrontOf(target));
+        const insertAt =
+          index === -1
+            ? //  `target` is in front of all other objects
+              this.size()
+            : index;
+        this.insertAt(insertAt, target);
+      });
+    }
+  }
```
asturur, line 64: why calling this `add` and not using the fact that this is a subclass of group, exactly to handle multi selections, and so its own add method is either super.add or this.insertAt?
I would gladly rename this in add. Maybe i ll merge this as it is, and i ll just open a PR with the things i wouldn't want to do

src/shapes/Group.ts
```diff
@@ -483,7 +483,18 @@ export class Group extends createCollectionMixin(FabricObject<GroupEvents>) {
...
-      this._objects[i].render(ctx);
+      // TODO: handle rendering edge case somehow
+      if (
+        this.canvas?.preserveObjectStacking &&
+        this._objects[i].group !== this
+      ) {
+        ctx.save();
+        ctx.transform(...invertTransform(this.calcTransformMatrix()));
+        this._objects[i].render(ctx);
+        ctx.restore();
```
asturur, line 494: this is the edge case handling, this is exactly how you do it.
The only improvement would be that instead of going back with an invert and a calc transform, you go straight to viewport transform using ctx.setTransform(this.canvas.viewportTransform).
I will add it in a follow up

src/shapes/Object/InteractiveObject.ts
ShaMan123: I touched it and reverted to keep scope. And there is that interaction PR that changes this.

src/shapes/ActiveSelection.ts
ShaMan123: thought of that as well.
wasn't sure what I/you think about that.
The reason I am not for is in case the dev wants to change the stack order, using `insertAt` is valuable so if we change it it will be annoying.

src/shapes/Group.ts
ShaMan123: I think this logic should be handled by active selection.
We should discuss this, there are a lot of implications.

src/shapes/ActiveSelection.ts
ShaMan123: but anyone can do `this._objects.insertAt` so I am ok with it

ShaMan123: do it

ShaMan123: so this will be add and insertAt will be left as is?

asturur: i was just saying to rename multiSelectAdd in add.
Because this class has to handle only multi selection logic and nothing else. so seems legit that its own add is special.

Are you thinking that people may want to call add of an active selection directly to do things in the active selection outside the canvas flow?

ShaMan123: Yes
Perhaps
What do you think?

src/canvas/SelectableCanvas.ts
asturur: `necessary` is out of topic here. Wasn't here, before and you could handle the selection just fine.
During the selection:created event you can attach events to the new selection.

But also i want to stress out that is unrelated. We are adding support for multi selection with objects located in different groups, and adding events on active selections, or offering a fixed reference to it, is a completely parallel issue that has been added here because it seemed a good idea while doing it.

There are reasons why in my opinion is not a good idea, one is that it make indivisible canvas and group and active selection, and that is worse than having to add events.
It is not wrong to have a fixed ref, but is not good to create it with canvas creation right away.
Having a ref allow us to not have imports to create the active selection, and that helps in dividing the code.

asturur:
@ShaMan123 as is i m merging this because there are many improvements here compared to the things i want to remove.
But i will need to come back with a PR ( like right away ) to remove/imporve the things i think aren't related to fixing the selection and are also new complexities i don't want to deal with.
- activeSelection.toGroup and group.toActiveSelection need to come back
- _activeSelection and getActiveSelection are a new behaviour i don't think is necessary and so i don't want to introduce
all the rest seems good to me

asturur (approved):

ShaMan123:
> activeSelection.toGroup and group.toActiveSelection need to come back

why?
```ts
new Group(canvas.getActiveObjects()) // toGroup 
(canvas.getActiveObject() || canvas.getActiveSelection()).add(...group.objects) // toActiveSelection 
```
https://github.com/fabricjs/fabric.js/pull/8665#issuecomment-1426662009
https://github.com/fabricjs/fabric.js/pull/8665#issuecomment-1426663719

asturur:
They were simply part of the api, not deprecated and useful, and have been removed in a larger PR without noticing

PR #8668, opened 2023-02-06
feat(): Control set

ShaMan123:
## Motivation

Long have we been discussing (and disputing about) shared controls.
The reason I was against it all along was because setting `rect.controls.br = new Control()` would change all fabric and it is absolutely an anti pattern. No one can expect that unless they are acquainted with fabric, with is a bad practice.
However you said there are benefits in a shared control set (and I found it hard to swallow because of this).
Another reason was Textbox controls. Saying fabric shares controls is not correct because of this edge case, so the benefits of the shared concept were missed out here.

With this clarity I now understand your point.
So if we handle what disturbs me I am for shared controls.
This PR does that exactly.
I wish to dedicate in to you @asturur as a gift from me.

## Description

fill out soon

### DIFF
--- a/fabric.ts
+++ b/fabric.ts
@@ -284,6 +284,7 @@ import {
 import { getLocalPoint } from './src/controls/util';
 import { wrapWithFireEvent } from './src/controls/wrapWithFireEvent';
 import { wrapWithFixedAnchor } from './src/controls/wrapWithFixedAnchor';
+import { createControlSet } from './src/controls/default_controls';
 
 /**
  * @todo remove as unused
@@ -310,6 +311,7 @@ const controlsUtils = {
   getLocalPoint,
   renderCircleControl,
   renderSquareControl,
+  createControlSet,
 };
 
 // EXPORTS
--- a/fabric.ts
+++ b/fabric.ts
@@ -363,8 +363,6 @@ import { parsePointsAttribute } from './src/parser/parsePointsAttribute';
 import { parseTransformAttribute } from './src/parser/parseTransformAttribute';
 import { getCSSRules } from './src/parser/getCSSRules';
 
-import './src/controls/default_controls';
-
 export {
   parseElements,
   parseAttributes,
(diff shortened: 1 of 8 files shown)


asturur:
we can look at this as soon as we are done with default values because controls need to come with defaultValues as any other thing that was in the prototype and now won't be anymore, otherwise the developer is getting always the default control set first and then has to change it instance by instance

ShaMan123:
This PR changes all of that
They are not part of the prototype anymore but they remain shared

src/controls/ControlSet.ts
```diff
@@ -0,0 +1,95 @@
...
+  target: T,
+  source?: S | HybridControls<S>
+) {
+  if (target.source) throw new Error('fabric.js: source is a reserved key');
+  return Object.defineProperties(target, {
+    source: {
+      value: source,
+      configurable: false,
+      enumerable: false,
+      writable: true,
```
ShaMan123, line 33: writable so source can be changed

ShaMan123 (commented):
This is ready!
And great
I love it
Shared and scoped controls on the same object, no need for anything, everyone happy out of the box

test/unit/textbox.js
```diff
@@ -178,8 +178,11 @@
       assert.deepEqual(obj.styles[2], textbox.styles[2], 'styles match at line 2');
       assert.deepEqual(obj.styles[2][0], textbox.styles[2][0], 'styles match at index 0');
       assert.deepEqual(obj.styles[2][1], textbox.styles[2][1], 'styles match at index 1');
-      fabric.Textbox.fromObject(obj).then(function(obj2) {
```
ShaMan123, line 181: this test was causing a lot of problems
it took me a lot of time to understand what is wrong
So because we spread the input arg in `Text.fromObject` passing instance to `fromObject` pollutes the created instance with class props

I don't think we should care about it, though it is an edge case. No sense in passing a live instance to `fromObject` but if you think differently then we can safeguard it by running `cloneDeep` before spreading

ShaMan123 (commented):
port after #8682 is merged:
- [x] `src/controls/default_controls.ts`
- [x] `src/controls/polyControl.ts`
- [x] `src/shapes/Object/InteractiveObject.ts`
- [ ] `src/shapes/Textbox.ts`
- [ ] `test/unit/canvas_events.js`
- [ ] `test/unit/object_interactivity.js`

PR #8672, opened 2023-02-07
fix(): set/discard active object return value

ShaMan123:
## Motivation

Working on #8665 made me need return values and then I encountered strange behavior

## Description

`setActiveObject` and `discardActiveObject` were chainable. We removed that but didn't return the needed values from `_setActiveObject` and `_discardActiveObject`.
Then I saw that `_discardActiveObject` return a value that is different than what is documented (true if deselected) and not aligned with `_setActiveObject` so I fixed it.
That surfaced another edge case when clearing canvas and having an active object that refuses to deselect clear can't discard it. So I fixed that as well and added tests.

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -1383,7 +1383,7 @@ export class SelectableCanvas<
    * Sets given object as the only active object on canvas
    * @param {FabricObject} object Object to set as an active one
    * @param {TPointerEvent} [e] Event (passed along when firing "object:selected")
-   * @return {Boolean} true if occurred
+   * @return {Boolean} true if the object has been selected
    */
   setActiveObject(
     object: FabricObject,
@@ -1397,16 +1397,14 @@ export class SelectableCanvas<
   }
 
   /**
-   * This is a private method for now.
    * This is supposed to be equivalent to setActiveObject but without firing
    * any event. There is commitment to have this stay this way.
    * This is the functional part of setActiveObject.
-   * @private
    * @param {Object} object to set as active
    * @param {Event} [e] Event (passed along when firing "object:selected")
-   * @return {Boolean} true if occurred
+   * @return {Boolean} true if the object has been selected
    */
-  protected _setActiveObject(
+  _setActiveObject(
     object: FabricObject,
     e?: TPointerEvent
   ): this is AssertKeys<this, '_activeObject'> {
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
Well, this is a scoped PR
Ready

src/canvas/SelectableCanvas.ts
```diff
@@ -1423,10 +1431,10 @@ export class SelectableCanvas<
    * This is the functional part of discardActiveObject.
    * @param {Event} [e] Event (passed along when firing "object:deselected")
    * @param {Object} object the next object to set as active, reason why we are discarding this
-   * @return {Boolean} true if the selection happened
+   * @return {Boolean} true if occurred
    * @private
    */
-  _discardActiveObject(e?: TPointerEvent, object?: FabricObject) {
+  protected _discardActiveObject(e?: TPointerEvent, object?: FabricObject) {
```
asturur, line 1437: this is not protected, we should also remove the `private` in the jsdoc. There are reasons why people want to use this and i use this too.

src/canvas/SelectableCanvas.ts
```diff
@@ -1438,8 +1446,9 @@ export class SelectableCanvas<
         this.endCurrentTransform(e);
       }
       this._activeObject = undefined;
+      return true;
     }
-    return true;
+    return false;
```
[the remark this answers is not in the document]
asturur: So this line was introduced 6 years ago with the creation of this method, and the PR wasn't saying anything around why was true or false. It does make sense that if no selection was changed, we say we also didn't discard anything.

src/canvas/SelectableCanvas.ts
```diff
@@ -1400,13 +1404,17 @@ export class SelectableCanvas<
    * @private
    * @param {Object} object to set as active
    * @param {Event} [e] Event (passed along when firing "object:selected")
-   * @return {Boolean} true if the selection happened
+   * @return {Boolean} true if occurred
```
asturur, line 1407: can we change those `occurred` with specific information has before `if a different object has been selected` and similarly for the other methods?

asturur:
@ShaMan123 i m asking to restore ( a better version of ) the previous JSDOC without usin `occurred` but being specific on when is true ( or false, the easier to describe clearly )
Also we shouldn't put protected on the _setActiveObject or _discardActiveObject, because they are useful.
I do use them often indeed.

Just let's not add protected, then i ll address the JSDOC on top of it being more specific on how or why you could call those directly.

asturur:
(i would change the JSDOC myself but i have no time right now)

ShaMan123 (commented):
done

asturur (approved):

Issue #8671, opened 2023-02-07
object dispose


ShaMan123:
Should Object#dispose call `off` to unsubscribe all events?
Should it unset the canvas ref?

I think so

PR #8673, opened 2023-02-07
fix(): Object dispose

ShaMan123:
## Motivation

#8671 
Another scoped PR from #8665 

## Description

The whole point of the dispose method AFAIK is to release memory quickly
So removing the canvas ref and all the events (which probably have a lot of side effects) is important

### DIFF
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -1830,11 +1830,9 @@ export class FabricObject<
    * override if necessary to dispose artifacts such as `clipPath`
    */
   dispose() {
-    // todo verify this.
-    // runningAnimations is always truthy
-    if (runningAnimations) {
-      runningAnimations.cancelByTarget(this);
-    }
+    runningAnimations.cancelByTarget(this);
+    this.off();
+    this._set('canvas', undefined);
   }
 
   /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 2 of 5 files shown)


src/shapes/Object/Object.ts
```diff
@@ -1830,11 +1830,9 @@ export class FabricObject<
...
    */
   dispose() {
-    // todo verify this.
-    // runningAnimations is always truthy
-    if (runningAnimations) {
-      runningAnimations.cancelByTarget(this);
-    }
+    runningAnimations.cancelByTarget(this);
+    this.off();
+    this._set('canvas', undefined);
```
ShaMan123, line 1835: using _set to propagate to all children

ShaMan123 (commented):
Thoughts?
This is as much a PR as a discussion
Ready to merge if you don't disagree

ShaMan123:
Came to me that disposing a canvas with a large tree with cost in perf since each object will be visited in the _set method

asturur:
I m not sure this PR is improving memory disposing, if we think it does we should be able to measure it in some way ( either from the heap allocation in the performance tab or in some other ways ).
In general this pr isn't dangerous or hurtful, and is ok to merge, but if we want to have PRs that improve something that is not a feature or a visbile bug as a general rule we should be able to point out what is changing

asturur (approved):

asturur:
is ok if dispose do something and traverse a tree, disposing isn't used in render cycles and is something that is done once in a while. What it takes it takes.

PR #8674, opened 2023-02-08
chore(TS): class interface for options/brevity

ShaMan123:
## Motivation

Using an interface with class is powerful and reduces the file size because docs can be moved to the interface and leave logic not polluted
The other motivation is that it provides a type for `options`. One would argue that this is one of the most important things for a TS consumer.

## Description

Extracted props to `ObjectProps` interface and removed comments from logic since VSCODE traces them correctly
image

## In Action

https://user-images.githubusercontent.com/34343793/217402648-ba9c96e1-23c8-4ce3-a9c2-c5d47b014ade.mp4


ShaMan123 (commented):
done

src/shapes/Circle.ts
```diff
@@ -38,6 +45,10 @@ export class Circle extends FabricObject {
     'endAngle',
   ];
 
+  constructor(options?: CircleProps) {
+    super(options);
+  }
```
asturur, line 50: If i m not mistaken as long as the  interface is fulfilled, we can add any property right?
foo, bar, baz are still valid. correct?

asturur:
What happens to transpiled unminified js? it looses all comments?
in that case isn't really good.

src/shapes/Circle.ts
ShaMan123: good question
the answer is no


```ts
interface A {
  foo: 'bar'
}

function actStrict(a:A){

}

function act(a:A & Record<string,unknown>){

}

actStrict({ foo:'bar', baz: 'pi' })
act({ foo:'bar', baz: 'pi' })
```



```ts
"use strict";
function actStrict(a) {
}
function act(a) {
}
actStrict({ foo: 'bar', baz: 'pi' });
act({ foo: 'bar', baz: 'pi' });

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgILIN4FgBQzkwD2hAXMgOQBGcU5uAvrrjAK4gJjCEjKJgDKYKMA4AKOCVQBKbDgZMcrdp268xE9ADJkAJQgJCUACYAeAM5DQAcwA0bANYhCAdxAA+GQsZycfQcLEMAmISKhpyG2RqAC8ycgAHYHJkeilcPlEgolIw2kiYuMTk1KA)

ShaMan123:
Comments are removed
image

Why are they needed in the bundle?

asturur:
i think my question is, who does import the js bundle and has no access to the packaged, does he find the JSDOCS? i think no because those get stripped away with types at this point.

ShaMan123:
I think yes because they have access to fabric.d.ts that is consumed by the code editor

ShaMan123:
Who uses it like that? Can you point it out? Is it a real case?

ShaMan123:
If you install via package manager you have `.d.ts`
Any other option is what exactly? Browser cdn? That isn't accessible anyways.

src/shapes/Circle.ts
asturur: I was about to merge this, but this blocks me.
I want to be able to do `Rect({ name: 'myname', id: '33' })` i think is a basic right was here before we introduced TS and shouldn't go away.

ShaMan123: Give me a few mins to take a look

asturur:
@ShaMan123 we should be able to add any property to a class that is not already defined with a specific type.
so `Rect({ top: 'a' })` should be an error, but `Rect({ name: 'hello' })` shouldn't be.
And so fromObject and toObject with `propertiesToInclude` should behave the same.

Can we just add `& Record<string, unknown>` to make it happen?

ShaMan123:
The follow up #8677  handles that exactly with success 
[I will revert the ](https://github.com/fabricjs/fabric.js/compare/ts-interface...types-to-object#diff-314ecc7e2f3a354fad7665ae60856ad451e28d5ca2af4aa7bfb889f691e4b8a1L48-L51)

ShaMan123:
Can we continue there?

asturur:
so should i close this and look at that?

ShaMan123:
merge this and move there or close this and we work there
did it for scoping

asturur:
ok will get to this today and the next, today

asturur:
Moving forward with this, the only thing i m not sold on is removing the jsdocs outside the code. The interface is obviously super useful for types and everything.
This will likely need to be done for rect and other parts i imagine but in that case let's create the interface first and move away the JSDodcs after.
If for some reasons we want jsdoc back in the codebase at least we can just revert the relevant commits

asturur (approved):

PR #8677, opened 2023-02-08
chore(TS): types for `toObject` and additionaProperties

ShaMan123:
## Motivation

#8674 opens up to possibility to type `toObject`

## Description

 TS limitations of static signatures prevents me from declaring an overload to Circle.fromObject
I decided to override though there is an option to declare the lass in a .d file and change the static signature there I think.
I researched it too long. 
Since we are talking about 50 classes or more it sounds better to have a .d file handle it.

## Changes

A lot of types
Moved declare fields around

## In Action

Try the following, test autocomplete in new, toObject, fromObject
```ts

import { Circle, Object as FabricObject } from "fabric";

class Circle2 extends Circle {
  declare foo: string;
}

const { foo, angle } = new FabricObject({ foo: "bar" }).toObject(["foo"]);
const { foo: bar, fill } = new FabricObject<{ foo: string }>().toObject([
  "foo",
]);

foo;
angle;
bar;
fill;

new Circle({}).fill;
new Circle({}).toObject().fill;

new Circle().toObject(["foo"]).foo;

new Circle({
  radius: 5,
  foo: "bar",
}).toObject(["foo"]).foo;

new Circle2().toObject(["foo"]).foo;

Circle.fromObject({
  a: "b",
  clipPath: {
    absolutePositioned: true,
    inverted: false,
    opacity: 1,
  },
  radius: 5,
});

```

https://user-images.githubusercontent.com/34343793/217529702-306e8772-79f0-4223-9030-cdaf58f68cd5.mp4


asturur:
As usual the important thing is that we can still export any property that has been added by the developer.
What is the status on this?

ShaMan123:
image

asturur:
@ShaMan123 i m not sure i can solve the conflicts correctly, the TS declarations confuses me at this point.
I have a couple of comments reading the code:
- we must be careful of not assuming that serialized props are the same for constructor. It seems to me that Shadow and Gradient could be points of confusions.
- if the standard fabric object can have a generic type with extra signautures, Circle<any, any, any>, i wish the extra props for the user would be the first one, so i can easily do Circle<TMyStuff> and letting the rest as the optionals.

asturur:
ok i tried with the conflicts. in case revert my commit

src/shapes/Circle.ts
```diff
@@ -205,6 +210,10 @@ export class Circle extends FabricObject implements CircleProps {
   }
 
   /* _FROM_SVG_END_ */
+
+  static fromObject(object: SerializedCircleProps): Promise<Circle> {
+    return super.fromObject(object) as Promise<Circle>;
+  }
```
asturur, line 216: something i noticed in the previous PR too, i would argue this can be solved with a declaration?
is weird that TS would force me to add code that does nothing to obey types.

src/shapes/Object/ObjectProps.ts
```diff
@@ -1,130 +1,8 @@
...
-
-  /**
-   * Padding between object and its controlling borders (in pixels)
-   * @type Number
-   * @default 0
-   */
-  padding: number;
-}
-
-export interface ObjectProps extends ObjectBaseProps {
```
asturur, line 127: i must have done something wrong because i don't see where this part went

src/shapes/Circle.ts
ShaMan123: correct
I was lazy

ShaMan123: This is another TS problem. Static class methods can't be generic using straight forward TS

asturur:
@ShaMan123 please i want to move on with this, so if you have spare time to work on fabric, instead of working on other prs prioritize this one. let's understand how this should be done and then eventually i can complete it for the other classes if you don't have time.

asturur:
for today i m off

src/shapes/Object/types/ControlProps.ts
```diff
@@ -0,0 +1,69 @@
...
+   * @deprecated
+   */
+  cornerStyle: 'rect' | 'circle';
+
+  /**
+   * Array specifying dash pattern of an object's control (hasBorder must be true)
+   * @since 1.6.2
+   * @type Array | null
+   */
+  cornerDashArray: number[] | null;
```
ShaMan123, line 54: I must say that corner is bad naming, should be control

ShaMan123 (commented):
`src/shapes/Object/ObjectProps.ts` turned into `src/shapes/Object/types`

recommended review order:

| Types | Classes |
| --- | --- | 
|  `src/shapes/Object/types/BaseProps.ts` | `src/shapes/Object/ObjectOrigin.ts`
|  `src/shapes/Object/types/SerializedObjectProps.ts`, `src/shapes/Object/types/ObjectProps.ts` | `src/shapes/Object/Object.ts`
| `src/shapes/Object/types/FabricObjectProps.ts` | `src/shapes/Object/InteractiveObject.ts`
| `src/shapes/Object/types/index.ts` | `src/shapes/Object/FabricObject.ts`
|  | `src/shapes/Circle.ts` 

Each type file depends on more types so review them as part of it

ShaMan123:
updated desc, ready

src/shapes/Object/Object.ts
```diff
@@ -490,7 +472,10 @@ export class FabricObject<EventSpec extends ObjectEvents = ObjectEvents>
    * @param {Array} [propertiesToInclude] Any properties that you might want to additionally include in the output
    * @return {Object} Object representation of an instance
    */
-  toObject(propertiesToInclude?: string[]): Record<string, any> {
+  toObject<
+    T extends Omit<Props & TClassProperties<this>, keyof SProps>,
+    K extends keyof T = never
+  >(propertiesToInclude?: K[]): { [R in K]: T[K] } & SProps {
```
asturur, line 478: This thing is unreadable and who invented typescript is responsible for my eyes bleeding

ShaMan123: It was not easy to write and definately not to test

asturur:
i had the laptop offline for all day because after a forced update it wasn't starting anymore ( i think the battery went off during update ).
The PR seems ok to me i will merge it and try to do another shape ( like RECT? ) to be sure i understand all the changes properly.

asturur (approved):

PR #8682, opened 2023-02-09
feat(): Hybrid Object

ShaMan123:
## Motivation

default values and shared controls discussions (a better version of closes #8668 )

Long have we been discussing (and disputing about) shared controls.
The reason I was against it all along was because setting `rect.controls.br = new Control()` would change all fabric and it is absolutely an anti pattern. No one can expect that unless they are acquainted with fabric, which is a bad practice. Thinking of an app that uses more than one canvas in which some line changes a control (meaning to change it for an object or for that canvas) affecting an unrelated canvas now having different controls.
However you said there are benefits in a shared control set (and I found it hard to swallow because of this).
Another reason was Textbox controls. Saying fabric shares controls is not correct because of this edge case, so the benefits of the shared concept were missed out here and yet another inconsistency and unexpected behavior was introduced.

To this rape discussion default values came in. You wanted the shared concept, I was against to keep object state uncompromised. 
Once I solved shared controls with a bit of help from SO and MDN proxy came to **provide a solution for classes as well** (see below).

With this clarity I now understand your point, I can see the benefits.
So if we handle what disturbs me I am for shared controls/values.

This PR does that exactly.
I wish to dedicate it to you @asturur as a gift from me.

## Description

We need an object (=`target`) that keeps the ref of a given source object and doesn't assign to it values or spread it.
We shall use it as a fallback object in case `target[key]` is not defined.

The concept of Hybrid is that once a key is touched (defined, set or deleted) it is marked and will be resolved after that to `target` always.
In other words, as long as the key isn't touched by target it is resolved to source.
This is a design must or else object behavior becomes undefined. e.g. deleting a key on the object should from then on return undefined even if there is a fallback source value. Anything else is twilight zone.
This is the scope of `createHybrid`.
`createHybrid` assigns the source ref to the target object and defines a proxy that treats source as fallback.
This way the ref is preserved so it is shared while the object (=`target`) is not shared so mutating it is scoped.
This approach is unlimited in levels of complexity, it can be nested.

Using proxy traps aligns us with class spec (assigning values in the constructors goes through the `defineProperty` trap and not the `set` trap) so by default, once we migrate `set`, no side effects will run during initialization.

I went ahead and added traps for side effects to replace `set` (as you suggested in the past) that call `transformValue` and `onChange` if needed. 
`transformValue` runs also from the `get` trap and can return a different value that will be used by the operation instead.
`onChange` is for side effects and can return false to revert the change that was made.

## Changes

- `createHybrid`
- `CommonMethods#restoreDefault` alias for TS
- `CommonMethods#restoreDefaults` alias for TS
- `CommonMethods#transformValue` get/set hook
- `CommonMethods#onChange` side effect hook that is invoked every time a change to a value occurs (see bubbling caveat)
- `Circle#set` - changed to prove the side effect logic

## Caveats
- ~~changing source doesn't call onChange on target, I think it should~~ must assign a hybrid as source for bubbling to occur

## Unlocked Possibilities
The set traps allows us to do more state syncs:
- `setCoords` - can virtually become an under the hood method that will be used rarely
- `calcTransformMatrix` - setting it in the trap and removing the need to have a cache key

## In Action

```ts
const defaultProps = createHybrid({ foo: 'bar' });
const circleDefaultProps = createHybrid({ radius: 10 }, defaultProps);

class Circle extends FabricObject {
   static getDefaultProps() {
       return circleDefaultProps;
   }
}

const defaultControls = createHybrid({ ml: new Control(), tl: new Control() });
const textboxControls = createHybrid({ ml: new ResizingControl() }, defaultControls);
defaultControls.tl = new Control();
defaultControls.bl = new Control();
textboxControls.tl === defaultControls.tl;
textboxControls.ml !== defaultControls.ml;

class Textbox extends IText {
   controls = createHybrid({}, textboxControls)
}
```

<details><summary><b>POC</b></summary>

```ts

const SOURCE_KEY = '__source__';

type Hybrid<T extends object, S extends object> = T & {
  [SOURCE_KEY]: S;
};

function createHybrid<T extends object, S extends object>(
  target: T,
  source?: S
) {
  return new Proxy(
    Object.defineProperties(target, {
      [SOURCE_KEY]: {
        value: source,
        configurable: false,
        enumerable: false,
        writable: true,
      },
    }),
    {
      get(target, p, receiver) {
        const source = Reflect.get(target, SOURCE_KEY);
        if (p === SOURCE_KEY) {
          return source;
        } else if (
          !Reflect.get(target, p) &&
          source &&
          Reflect.has(source, p)
        ) {
          return Reflect.get(source, p);
        } else {
          return Reflect.get(target, p);
        }
      },
      set(target, p, newValue, receiver) {
        const prevValue = Reflect.get(target, p, receiver);
        if (Reflect.set(target, p, newValue, receiver)) {
          p !== SOURCE_KEY &&
            prevValue !== newValue &&
            target instanceof A &&
            Object.getOwnPropertyDescriptor(target, p)?.enumerable &&
            target.runSideEffects(p, newValue, prevValue);
          return true;
        }
        return false;
      },
      ownKeys(target) {
        const ownKeys = Reflect.ownKeys(target);
        Reflect.ownKeys(Reflect.get(target, SOURCE_KEY) || {}).forEach(
          (key) => !ownKeys.includes(key) && ownKeys.push(key)
        );
        return ownKeys;
      },
      getOwnPropertyDescriptor(target, p) {
        const source = Reflect.get(target, SOURCE_KEY);
        return (
          Reflect.getOwnPropertyDescriptor(target, p) ||
          (source && Reflect.getOwnPropertyDescriptor(source, p))
        );
      },
      has(target, p) {
        const source = Reflect.get(target, SOURCE_KEY);
        return Reflect.has(target, p) || (!!source && Reflect.has(source, p));
      },
    }
  ) as Hybrid<T, S>;
}


// class logic

const defaultValues = { foo: 'bar' };

class BaseOptions {
  x = 'y';
  static getDefaults() {
    return defaultValues;
  }
  constructor(options = {}) {
    return createHybrid(
      Object.assign(this, options),
      BaseOptions.getDefaults()
    );
  }
  act() {
    return this.x;
  }
}

class A extends BaseOptions {
  constructor(options) {
    super(options);
  }
  act() {
    return this.x + 1;
  }
  runSideEffects(key:string,value,prevValue){
    console.log('side effect', {key, value, this: this[key], prevValue})
  }
}

const a = new A({ z: 'q' });
console.log('intial', a);
console.log('source', a[SOURCE_KEY]);

a.x = 5;
console.log('x assignment', a.x);
console.log('instanceof', a instanceof BaseOptions);
console.log('keys', Object.keys(a));

a.x = undefined;
a.x=undefined // should not trigger side effect
defaultValues.x = 200;
console.log(
  'x unassignmet returns default value run through subclass',
  a.act()
);
console.log('keys', Object.keys(a));

a[SOURCE_KEY] = undefined;
console.log('removed source', Object.keys(a));

Object.assign(a, {b:'x',foo:'fi'});


```

</details>

```ts
"use strict";
const SOURCE_KEY = '__source__';
function createHybrid(target, source) {
    return new Proxy(Object.defineProperties(target, {
        [SOURCE_KEY]: {
            value: source,
            configurable: false,
            enumerable: false,
            writable: true,
        },
    }), {
        get(target, p, receiver) {
            const source = Reflect.get(target, SOURCE_KEY);
            if (p === SOURCE_KEY) {
                return source;
            }
            else if (!Reflect.get(target, p) &&
                source &&
                Reflect.has(source, p)) {
                return Reflect.get(source, p);
            }
            else {
                return Reflect.get(target, p);
            }
        },
        set(target, p, newValue, receiver) {
            var _a;
            const prevValue = Reflect.get(target, p, receiver);
            if (Reflect.set(target, p, newValue, receiver)) {
                p !== SOURCE_KEY &&
                    prevValue !== newValue &&
                    target instanceof A &&
                    ((_a = Object.getOwnPropertyDescriptor(target, p)) === null || _a === void 0 ? void 0 : _a.enumerable) &&
                    target.runSideEffects(p, newValue, prevValue);
                return true;
            }
            return false;
        },
        ownKeys(target) {
            const ownKeys = Reflect.ownKeys(target);
            Reflect.ownKeys(Reflect.get(target, SOURCE_KEY) || {}).forEach((key) => !ownKeys.includes(key) && ownKeys.push(key));
            return ownKeys;
        },
        getOwnPropertyDescriptor(target, p) {
            const source = Reflect.get(target, SOURCE_KEY);
            return (Reflect.getOwnPropertyDescriptor(target, p) ||
                (source && Reflect.getOwnPropertyDescriptor(source, p)));
        },
        has(target, p) {
            const source = Reflect.get(target, SOURCE_KEY);
            return Reflect.has(target, p) || (!!source && Reflect.has(source, p));
        },
    });
}
// class logic
const defaultValues = { foo: 'bar' };
class BaseOptions {
    static getDefaults() {
        return defaultValues;
    }
    constructor(options = {}) {
        this.x = 'y';
        return createHybrid(Object.assign(this, options), BaseOptions.getDefaults());
    }
    act() {
        return this.x;
    }
}
class A extends BaseOptions {
    constructor(options) {
        super(options);
    }
    act() {
        return this.x + 1;
    }
    runSideEffects(key, value, prevValue) {
        console.log('side effect', { key, value, this: this[key], prevValue });
    }
}
const a = new A({ z: 'q' });
console.log('intial', a);
console.log('source', a[SOURCE_KEY]);
a.x = 5;
console.log('x assignment', a.x);
console.log('instanceof', a instanceof BaseOptions);
console.log('keys', Object.keys(a));
a.x = undefined;
a.x = undefined; // should not trigger side effect
defaultValues.x = 200;
console.log('x unassignmet returns default value run through subclass', a.act());
console.log('keys', Object.keys(a));
a[SOURCE_KEY] = undefined;
console.log('removed source', Object.keys(a));
Object.assign(a, { b: 'x', foo: 'fi' });

```

</details>

```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```

</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/LAKAxg9gdgzgLgAgMoHkCqAlAwgUQPoDSOAmggLwIDkeeMEArgE5gCmNlA3KKHAJ4AOLBAAleAI0YBLACYAeACoIWADzgso0mAghiAVizBwANMiWr1m7XoNwAfOQSKAZAgDeoBAgDaqTLkIkALoAXMhcIAC+4aAAZvRQhpLQCGCMLACGaqISMgpmahpaOvqGJkj5FkXWhrYAFB4IcOmMAOYscKHyRg10TKwA-KFIoACUbg1pcExQCFAsAO4IAAqMEMq89SCenijVcAB00iwxknMrEIKMcJIsMLVNre0m7lvbnj7o2PhExCHjr29PAA3dIAG3oLFCvWYLG6AMBkCgJxaTHSYlBkIQMTBMFhDUBSig9AAtixGGiMaFsaDcXCCZ55lImujMXBGBC6YCIpzPBERjz-gS2nB7s1hSZ+CY0qxJECyWMXvSUtB4AhoawHBhjhjDPthaLHsZkJ9-D8RuElZIYghavxyGQKL4vgFiAr8fTJtM1QwYRb6RElDShFabe6CQBCLUxHUHfUPcUIfhjJxOMOA9VCFNpt5RmP7AAW6TuGYlI2zCDd8IJnsYM1zNj17VqJcT5vLAZYQcFSoQNbr2obcbFT1bfoJETT3LTuJF8ZHktmCwAamCOb2DCxZfLuwTEar+GkgSvwUIKPXdUPDRKpRut4w21XtiHaueDjODQmF3N5se19LN3K96Vj2iYIOGDrGn43wkAgWaPgSB4sEeq5COBFDfr+mapvBgJzogpzwOkCQsBA1oAIKwdhIE7HsjZwCg8xQOclx8AAIrcqSSPwcAQIwH7ziM-T7OoJJkhSWHlm8eH7OyUBIDILA4DEMQ2HcX7LihEqHphD4gX2jTsiwY5cuW+nUrixm8gKngQIxBAsLwdx4cB9J7ogtlQPZjmagOuoeV5TnDnAun0q++z+Q5dxhZeCZOqaJBjAAPolbh8vsMS8Tg6RgPmmwgbUADWDljGQ9jhhFjn7KcYDgkcdxFbwyYuBVMD7Pw9AwLlDVljhIXVu0XotZZCBTo+woMUxqwsbw7EwJx3G8fxRpJjuCIqogGY+dGg5NnhZQmtBrrDZ4+l5T20XtBNzFkmxHFSAtfF7a2CDJZJNqbSmCAXfRjHXVcM13VxPF8S2SY9UqfXbKNBKFoFV7PYqrnrd6fSnl9vmxrtQX7VBLqQ28+lhbDS2li9KW1OG4YfS4RNFs2PqsKW+MjQKE6vGMRYiOIUhyF0yC2OEbPcCAAD0IspKCRZaKCEAtJIYDC25CBHNi9CgnAmFaBQrhYhAEChJQYjNJQI3ROAkswFoABCRYsCg3FJLA3bKA4lC8JwPRNNcYAIMK7Gq+rdwuSdA21srxzpGrGsoTAfps54bnsoYi0XNcKoOK4fKrfpqQZFk3MyGdby7CUBxS5ILRQPc+aSDAJip47MD8mmNu4vbaewHR-uR4HtTg9sIXxwg2UisH65TGHcA161yhx6AQvm1LCAUSoBSWK3dsO+niOJ-Qyd8Q3KpjzA9CXLUh+wIPDQj332ehzMU+1-sLsANQIAAjHPryyfJRxKSphh6oOWCPAKQUAWhGBBCeIwiFkInhGIjBOKoIAYn2DLFotRKAwAUkoZSNhKDPAaiYKBa5H4wFCGQrwDVAhaSQphPkDQ2YLyVukBw35l61B1gALwNgARxNnycIe4UEsDQbLTBpxrhggIcPB8wjUHoMwRmGR6QPi4x+IEB8oB0jPwcAAViEcghR4jKAu3LpXUkUA4AqOfnIoxojFGUAIk0YipEVEIGcURVgpEEAb3bo3OxsARFiIwZQBqMAZElwbOE2o6QRhaJADol2FB4gq1OCwaQ4QklkFSccdJ0gEBizVPmBgoIClQAgIgNkFc2iMDVDg44AC4CgBVj3aOJ4Z4OAAEwAAYemGKCcYjBDRTEIHiOYqApJEB9i0K0qOCASFCFko0fMqx6AtHzGqegYgapSwIdffYN8eqBLoEMzB4TIm0RiXEhJqi4qHUCA4XJJw5iZNAPIhxJi0jEggHKApyiTBRN1Nc+JZsgVl0thXKu6RnhiGCKYghGV9aUBOJQQRwsgA)


ShaMan123 (commented):
done

ShaMan123 (commented):
Wired classes with this amazing and elegant solution
Added a set trap for side effects

All tested

Stopping now to keep the scope in tact

ShaMan123:
I am reverting the `set` trap because we need to discuss it further.
It needs to be defined how the `get` trap will operate once a value is deleted or nullified.
```
delete obj.a
obj.a // currently if source has a defined this will be a value
```

I don't think this is what we want.
I think we want a `tocuhed` scheme. `get` should query `source` only if the key was never touched

ShaMan123 (commented):
I have committed a minimal POC for `set`, look at `Circle#radius`

ShaMan123 (commented):
Wow!

ShaMan123 (commented):
changing source now bubbles to targets! 
refactored file structure

This is perhaps the best code I wrote

ShaMan123:
@asturur I am done and won't touch this PR anymore

ShaMan123 (commented):
before update from master

asturur:
@ShaMan123 i think this is a level of complexity i don't want to go to.
It seems we want to replicate the behavior of the Object / Class with values on the prototype, but without having the values on the prototype.

I don't want to go on this road, i want to make things simpler.

asturur:
Closing for now and we will have separated controls for everyone.

ShaMan123:
We should use this for setter/getter at least

PR #8686, opened 2023-02-10
fix(test): Textbox fromObject test

ShaMan123:
## Motivation

This test is blocking me while working on features and makes me think my fixes are buggy 

You asked for scoped PRs so here it is.
Blocking control set fix.

## Description

this test was causing a lot of problems
it took me a lot of time to understand what was wrong
because we spread the input arg in `Text.fromObject` passing instance to `fromObject` pollutes the created instance with class props that are enumerable strangly. This might be a node/QUnit thing.

I don't think we should care about it, though it is an edge case. 
No sense in passing a live instance to `fromObject` but if you think differently then we can safeguard it by running `cloneDeep` before spreading it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
-- test(): Textbox `fromObject` test not to pass an instance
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
@melchiar can you shed some light on this test?

asturur (approved):

PR #8694, opened 2023-02-14
fix(test): remove bad node config

ShaMan123:
## Description

`deepEqual` reports a false positive. I know this because browser tests fail where node tests don't
Removing this config fixes it


asturur:
I don't remember why we added this in first place.

asturur (approved):

PR #8704, opened 2023-02-18
feat(): Control set

ShaMan123:
## Motivation

continues #8682 , merge that first

## Changes

controls are not attached to the prototype anymore and are hybrid.
each control is still shared, meaning that changing a property on the `br` control will change it everywhere, on all objects
```
rect.controls.br.x = 0.1
```

It is possible to extend the hybrid proxy to clone the control once it is changed. But the major issue of this PR is solving 
```
rect.controls.br = new Control()
```
so I think I will leave it as is.
A solution:
```
rect.controls.br = new Control({ ...rect.controls.br, x: 0.1 })
```

https://github.com/fabricjs/fabric.js/compare/hybrid...hybrid-controls

## In Action

https://user-images.githubusercontent.com/34343793/219964293-55e3d347-58ba-4630-a4ef-0159ca4d367d.mp4

https://user-images.githubusercontent.com/34343793/219965404-cb16d02d-5d25-46bd-bf6e-a913197549ac.mp4

https://user-images.githubusercontent.com/34343793/219967704-88e989ce-29fc-45a3-810b-2d70efddeb88.mp4


src/controls/polyControl.ts
```diff
@@ -131,5 +133,5 @@ export function createPolyControls(
       ...options,
     });
   }
-  return controls;
+  return createHybrid(controls, defaultObjectControls);
```
ShaMan123, line 136: I changed this so now poly controls have also default controls

ShaMan123: IMO @luizzappa 's fix for skewing should be reinstated

asturur:
closing as we are keeping what we have just duplicated every instance

PR #8706, opened 2023-02-18
WIP

asturur:

### DIFF
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -16,11 +16,13 @@ import { AssertKeys } from '../typedefs';
 import { isTouchEvent, stopEvent } from '../util/dom_event';
 import { sendPointToPlane } from '../util/misc/planeChange';
 import {
+  isActiveSelection,
   isFabricObjectWithDragSupport,
   isInteractiveTextObject,
 } from '../util/types';
 import { SelectableCanvas } from './SelectableCanvas';
 import { TextEditingManager } from './TextEditingManager';
+import { ActiveSelection } from '../shapes/ActiveSelection';
 
 const addEventOptions = { passive: false } as EventListenerOptions;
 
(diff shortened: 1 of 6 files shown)


ShaMan123 (commented):
The benefit of having a ref is that the dev can subclass ActiveSelection and set the ref to use the subclass. w/o the ref subclassing is not possible.
If the problem is initializing ActiveSelection that can easily be handled. We can declare it and set it when needed.
We could expose a `createActiveSelection` method instead. Or as you suggested we the dev can listen to events and create a new active selection from the `selection:created` event and set it to be the active object.

asturur:
As i mentioned pushing this feature in because we were fixing the nested multi selection, is not the way to go.
We can't say: i need to fix A because is broken and behaves bad, but meanwhile take also this new idea B.
It increase the scope of the TS migration, without ever reaching stability.

I want to revert it now, then i can open a PR with this reverted again to reintroduce it, for when we have time to thinkering about it, but i want to move on with the other tasks, without this in.

Is a matter of not shoveling everything in, and stopping refactors, as i m asking since november/december.

ShaMan123:
So can we have a chat about the scope of v6?

asturur:
For me was TS rewrite and Group rewrite

Then:
- interactive group
- polygons exact bounding box
- different layouts for groups
- brush api
... i lost track if we added something else new

are all changes we pushed in during the course of the events

So we have to finish this migration, move to a phase in which we can listen to effect of changes and publish tutorial and documentation.

ShaMan123:
There are PRs I need to port, text PRs (keyCode deprecation, rtl, etc.)
Some seem to me part of v6....
But we can draw the line and move on to v7.
I will port them so I can use them, doesn't mean you have to include them.

asturur:
keycode is fine, is not structural of how the dev use fabric, is part of modern js, but also can be done in 6.0.1? is that breaking anything?
rtl is a new feature will be 6.1

ShaMan123:
depends if I deprecate `keyCode` or if I remove it entirely, only to use `key` which is the modern js way to do it according to MDN.
You prefer deprecating and not removing until next major?

asturur:
I don't remember the PR to answer, but i m sure we can stop using keycode internally and leave event data intact, if is not possible or require a tons of code we will do it before releasing

asturur:
Well i can't revert this keeping a particular test running.
I will circle back, today i would like to progress on something on the default values or something else that is actionable.

PR #8707, opened 2023-02-19
ci(): Node test cleanup

ShaMan123:
## Motivation

a non brainer following #8694 

## Description

node test config cleanup
commits are meaningful and describe all that is needed

## Changes

removed `setEnvForTests` since there is no need to define `CustomResourceLoader`, logging is sufficient

image

## Gist

Gave a look at
- https://github.com/kulshekhar/ts-jest
- https://github.com/tom-sherman/blog/blob/main/posts/02-running-jest-tests-in-a-browser.md

seems promising

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- ci(test): cleanup node config (#8694 followup) [#8707](https://github.com/fabricjs/fabric.js/issues/8707)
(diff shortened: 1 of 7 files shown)


asturur:
The only reason we had that was to have a leaner error when running a planned error. It seems now the error logging is a bit shorter than it was in the past ( was a full page scroll of writing ).
Let's move forward without, it will be easy in case to add back just that.

asturur (approved):

PR #8708, opened 2023-02-19
refactor(IText): `copyPasteData` => clipboard events

ShaMan123:
## Motivation

Using clipboard events to support cross platform clipboard events for IText
ports and closes #7903 

**Another step towards modern js**

## Description

This PR removes all limitations from clipboard events and conforms with HTML standards.
Previously, for example you couldn't copy text between different pages with fabric based apps. Now it's supported out of the box.

## Changes

- Remove `copyPasteData` in favor of `ClipboardEvent#clipboardData`
- Split clipboard event logic from input logic by calling `preventDefault`


ShaMan123 (commented):
seems ready

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -295,32 +283,83 @@ export abstract class ITextKeyBehavior<
...
+    // prevent input event
+    e.preventDefault();
+    const clipboardData = e.clipboardData;
+    if (this.selectionStart === this.selectionEnd || !clipboardData) {
+      return false;
+    }
+    const value = this.getSelectedText();
+    clipboardData.setData('text/plain', value);
+    clipboardData.setData(
+      'application/fabric',
```
asturur, line 299: this should be application/fabric-textstyle in case we decide to copy something else in the future

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -78,7 +77,7 @@ export abstract class ITextKeyBehavior<
     this.hiddenTextarea.addEventListener('keyup', this.onKeyUp.bind(this));
     this.hiddenTextarea.addEventListener('input', this.onInput.bind(this));
     this.hiddenTextarea.addEventListener('copy', this.copy.bind(this));
-    this.hiddenTextarea.addEventListener('cut', this.copy.bind(this));
```
asturur, line 81: cut is fine as copy, since it has to delegate the removal of text to onInput to keep the logic centralized.

asturur (changes requested):
Do not split clipboard logic from input logic, since that is in that way by design.
Different input methods have different logic and we prefer to delegate to a single function the insertion and removal of text.
If there is some bug in the onInput derived from copy/paste with this new setup let's see if is fixable in onInput.

src/shapes/IText/ITextKeyBehavior.ts
ShaMan123: I think all should be uniform `application/fabric` 
I can add a `type: text` to it

src/shapes/IText/ITextKeyBehavior.ts
ShaMan123: I will centralize, but firing copy when cutting is not good

src/shapes/IText/ITextKeyBehavior.ts
ShaMan123: done

ShaMan123:
Seems the copy event must be prevented in order for the clipboard data to be written and persisted to the paste event.

ShaMan123:
> Using clipboard events to support cross platform clipboard events for IText

This isn't correct unfortunately. Styles are not respected. So I looked into it and I have a solution. I guess you would want to add it as is and circle back to support styling.

ShaMan123:
Closing this in favor of #8734 
Feel free to reopen and merge this.
#8734 is built on top of this.

PR #8709, opened 2023-02-19
dep(IText): `keyCode` and key maps => `handleSelectionChange`

ShaMan123:
## Motivation

`keyCode` is deprecated
ports and closes #7865 

## Description

**FIRST** merge #7903 

event maps are IMO a deprecated concept.  It is a strange impl. Overriding is easier on the codebase and for the dev I believe (e.g. execution order of key map when modifier keys are active - should shift/ctrl be first?).
In order not to break anything I have created `handleSelectionChange`. It is where all the logic lives so easy for overriding.

## Changes

https://github.com/fabricjs/fabric.js/compare/clipboard-events2...dep-key-code

- support passing functions to keyMap
- fix: modifier functions prior to non modifier in calling logic
- voided onKeyUp (didn't remove it for bidi support of crtl+shift to change text dir)

## Next

once this is merged I will port the remaining RTL PRs


asturur (changes requested):
PR #7865 , without the support of functions in keymap is all we need if we want to deprecate keyCode. Other changes stacked on top are not needed/ not wanted right now.
They add changes in favor of new way of doing things, but those new ways aren't better or worse, are just different, so the change itself is not worth because require re-learning without adding much.

ShaMan123:
It is better:
- Why should key up have only a map for ctrl? What about regular key up? It is a valid question regarding support of a feature. What would your answer be? I prefer to remove this feature entirely. It doesn't give us anything, does it?
- Handling execution order of modifier vs. non modifier logic requires overriding anyways
- Overriding and subclassing is the preferred pattern when changing functionality (do we agree on that now? I saw you wrote something about it not long ago)
- In terms of code maintenance it is much better because TS/VSCODE recognizes everything, doing `find all references` shows the usage of methods in key down. Before you would need to remember that key down/up use the methods in the maps... (In general the use of strings to call functions is bad practice)
- Doesn't render canvas unnecessarily - a method that needs rendering (`exitEditing`) should call it explicitly.
- It isn't a shared object, not part of default values, but part of the prototype as functionality should be

And It is not breaking
The only "real" change I made that a dev will experience was removing `ctrlKeysMapUp`, which I can restore.
Do you really think this isn't better?

https://github.com/fabricjs/fabric.js/blob/a56c4b2bc3b584e827a43dc0b2f81ff7b0653932/src/shapes/IText/ITextKeyBehavior.ts#L151-L181


---

I see it as part of modernizing the code base.
To me it is better, I put some effort to it, but whatever. Decide and let's move on.

asturur:
Yes please just deprecate keycode since you wanted to, don't do other changes to key handling because i don't want to change that part until all the other text things are settled ( RTL, vertical text ).
I don't want change code that works for modernity reason or metrics that can't be measured ( faster code, less code )
Every change that does not fix a bug or is not faster, has a re-learning cost that i don't want to pay.

PR #8711, opened 2023-02-19
fix(IText): layout regression

ShaMan123:
## Description

`_shouldClearDimensionCache` was changed in #8663, not taking into account state changes any longer including changes to `text` which should trigger a layout - `text` was changed but layout was skipped
This PR moderates those calls


src/shapes/Text/Text.ts
```diff
@@ -1438,23 +1438,12 @@ export class Text<
...
-   * @private
-   */
-  _shouldClearDimensionCache() {
-    const shouldClear = this._forceClearCache;
-    if (shouldClear) {
-      this.dirty = true;
-      this._forceClearCache = false;
-    }
-    return shouldClear;
-  }
```
ShaMan123, line 1456: dismantled this method.
`dirty` is now set in `initDimensions`
`_forceClearCache` is reset in `_clearCache`

makes a lot more sense

ShaMan123 (commented):
@asturur take a look

asturur:
So with the set part, we missed text changes that aren't coming from set.
I would rather prefer we keep things sa they are and we add the _forceClearCache boolean to true when we know there are those text changes.
In this way we fix the bug and we move forward and we don't mix the fix with another refactor

ShaMan123:
I want to share my reasoning.
`_shouldClearDimensionCache` is a state check, so as you said a long time ago, it shouldn't contain side effects but it does (discussed regarding `isCacheDirty`), it flags `dirty`.  Calling `initDimensions` doesn't flags `dirty`. If flagging `dirty` is needed (when rendering `isCacheDirty` => `_updateCahceCanvas` triggers a redraw because dimensions changed), it should be done from where the change happens. It should be scoped.
Furthermore, if `_forceClearCache` was true before calling `initDimensions` it would remain true even after dimension caches were cleared by `initDimensions`. Again, it should reside where caches are cleared.
So this is a bigger fix, a scoping fix. If you don't approve please share your thoughts.

asturur:
Going to solve conflicts and mere this tomorrow.
I ll make it more similar to the other one

asturur (approved):

PR #8712, opened 2023-02-19
fix(IText): drag image + retina

ShaMan123:
## Motivation

drag image was blurry

## Description

depends on #8711 to work properly but isn't blocked by it

## Changes
- fixed blur by changing width/height of HTMLImageElement as you suggested.
- fixed vpt blur: moved the vpt drawing logic to `createCanvasElement` to keep text sharp by avoiding drawing images of text.
- updated golden and made visual tests stricter
- positioned `fixed` so adding to body will not affect layout


ShaMan123 (commented):
done

src/shapes/IText/DraggableTextDelegate.ts
```diff
@@ -140,39 +140,45 @@ export class DraggableTextDelegate {
     };
     target.setSelectionStyles(styleOverride, 0, selectionStart);
     target.setSelectionStyles(styleOverride, selectionEnd, target.text.length);
+    target.dirty = true;
```
asturur, line 143: don't we need to use always `set` for dirty in case is inside an interactive group?

src/shapes/IText/DraggableTextDelegate.ts
```diff
@@ -140,39 +140,45 @@ export class DraggableTextDelegate {
...
     setStyle(dragImage, {
       position: 'absolute',
-      left: -dragImage.width + 'px',
+      left: `${-dragImage.width}px`,
       border: 'none',
+      transform: `scale(${1 / retinaScaling},${1 / retinaScaling})`,
     });
-    getDocument().body.appendChild(dragImage);
-    e.dataTransfer?.setDragImage(dragImage, offset.x, offset.y);
+    // wrap the image so css transform will be respected
```
asturur, line 173: di you try adding width/heigth to the style object divided by retina scaling, and see if the div is still necessary?

ShaMan123: Didn't try

src/shapes/IText/DraggableTextDelegate.ts
ShaMan123: well no,
since I need it only for the export I don't want the group to invalidate.
It is a render export matter, as we discussed. All because of this mix we must invalidate the cache so the export is fine...

src/shapes/IText/DraggableTextDelegate.ts
ShaMan123: done

ShaMan123 (commented):
changed width/height as you suggested.
moved the vpt logic to `createCanvasElement` to keep text sharp by avoiding drawing images of text.
updated golden and made visual tests stricter
done

asturur (approved):

PR #8714, opened 2023-02-20
BREAKING: remove object `type` in favor of `Class.name`

ShaMan123:
## Motivation

No values on prototype, continues the default value effort

## Description

removed `type` since it was a readonly static property

## Changes

- registered `FabricObject` in class registry. Only to base fabric object was registered

**BREAKING**: 
- removed `type` from all objects/filters. Use `instanceof` or `isType` instead
- `isType` will not work for old types, you must pass the class name, that is why `instanceof` is the recommended migration path


src/filters/ColorMatrixFilters.ts
```diff
@@ -8,7 +8,11 @@ export function createColorMatrixFilter(key: string, matrix: number[]) {
      * @param {String} type
      * @default
      */
-    type = key;
+    static readonly type = key;
+
+    getType() {
+      return super.getType();
+    }
```
ShaMan123, line 15: TS complains since protected methods can't be in a mixin so I made it public

ShaMan123 (commented):
ready
IMO most of filters `fromObject` can be move to the base class

asturur:
For 6.0 we will keep `type` as it is, a value like the others, we just deprecate it. It will go wherever all other defaults go.
Is not worth to make a special case for it

ShaMan123 (commented):
updated from master

src/gradient/Gradient.ts
```diff
@@ -200,7 +201,11 @@ export class Gradient<
       offsetX += object.width / 2;
       offsetY += object.height / 2;
     }
-    if (object.type === 'path' && this.gradientUnits !== 'percentage') {
+    if (
+      object instanceof BaseFabricObject &&
```
ShaMan123, line 205: maybe not good

ShaMan123 (commented):
ready

ShaMan123 (commented):
reverted changes made to filters

--- 

I want to remove `isType` as well
The reason is that logic like `isActiveSelection` or `isTextObject` (and more from `src/util/types.ts`) is not good enough anymore in case of an app subclassing these classes. It will fail the check, `instanceof` is the migration path.

ShaMan123:
@asturur I want your feed back on this when you are done with filters

ShaMan123:
The solution I can think about that doesn't involve `instanceof` (or its hacks I suggested by creating dummy superclasses just for the sake of the test) is using the classRegistry... which is not bad actually

asturur:
we are not removing type nor chaning its value.
We will have a plan to deprecate it that doens't break developers data.
Going to slim down the breaking changes here in some way. i ll assign to me.

ShaMan123:
> We will have a plan to deprecate it that doens't break developers data.

It doesn't.
It beaks `isType` and [b416b55](https://github.com/fabricjs/fabric.js/pull/8714/commits/b416b5520edb407a846f40295d1b465a4bbd5c3f)

ShaMan123:
I registered IText for example in `classRegistry` under 2 names, `IText` and `i-text` so it is backward compatible

ShaMan123:
exposing a getter as you did is a good option non the less

asturur:
yes is what i m doing, adding a getter for type and adding a test to verify old data works.
also isType needs to work for both old and new names for now

ShaMan123:
I added tests in the class registry tests, take a look there

asturur:
@ShaMan123 looking at:
https://github.com/fabricjs/fabric.js/pull/8714/commits/e1a67344a5a219590e43202ae07af486cd6762d6

I did some things:
- added  a get type / set type for compatibility with old code
- let go of the new name of Object that can be FabricObject
- add warnings about setting type, and should add a warning when developer access it
- keep isType compatibility with both old and new names ( we should put a warning that for now is case-insensitive )

Now before i change the tests from `Object` to `FabricObject` i want to hear your thoughts on the changes.
Then i need to write a compatibility test for old data

src/ClassRegistry.ts
```diff
@@ -34,26 +32,28 @@ export class ClassRegistry {
...
-      this[SVG].set(SVGTagName, classConstructor);
-    } else {
-      this[SVG].set(classConstructor.name, classConstructor);
-      // legacy
-      this[SVG].set(classConstructor.name.toLowerCase(), classConstructor);
-    }
+    this[SVG].set(
+      SVGTagName ?? classConstructor.name.toLowerCase(),
+      classConstructor
+    );
```
asturur, line 56: svg is always lowercase

src/Collection.ts
```diff
@@ -93,10 +93,14 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
...
+     * @param {...String} [types] When specified, only objects of these types are returned
      * @return {Array}
      */
-    getObjects() {
-      return [...this._objects];
+    getObjects(...types: string[]) {
+      if (types.length === 0) {
+        return [...this._objects];
+      }
+      return this._objects.filter((o) => types.includes(o.type));
```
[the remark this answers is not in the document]
ShaMan123: use `isType` instead

src/ClassRegistry.ts
ShaMan123: good point

src/Collection.ts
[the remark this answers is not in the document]
ShaMan123:
```suggestion
      return this._objects.filter((o) => o.isType(...types));
```

ShaMan123 (commented):
I suggest we define `type` as following:
```js
Object.defineProperty(o, 'type', {
      enumerable: false,
      // so we can define in subclass
      configurable: true,
      writable: false,
      value: 'i-text',
    });
```

this way trying to write over it will throw an error so all is good. The dev sees it and can easily migrate AND it is hidden from TS so TS will warn about it as well

asturur:
> I suggest we define `type` as following:
> 
> ```js
> Object.defineProperty(this, 'type', {
>       enumerable: false,
>       // so we can define in subclass
>       configurable: true,
>       writable: false,
>       value: 'i-text',
>     });
> ```
> 
> this way trying to write over it will throw an error so all is good. The dev sees it and can easily migrate AND it is hidden from TS so TS will warn about it as well

What is wrong with the setter? i could not add the setter and we would get the error, but the error is exactly what i don't want because today the property is writable
There is the warning on purpose.

ShaMan123:
I don't mind.
It is a difference in approach. I view breaking changes as fine as long as they are not silent, cause then they are deadly and consume days. If everything crashes and points me to the right line I am more than content.

asturur:
> I don't mind. It is a difference in approach. I view breaking changes as fine as long as they are not silent, cause then they are deadly and consume days. If everything crashes and points me to the right line I am more than content.

Breaking change we have a  tons, so for now we are not adding more.

src/Pattern.ts
```diff
@@ -54,7 +52,7 @@ export class Pattern {
   }
 
   set type(value) {
-    console.warn('Setting type has no effect', value)
+    console.warn('Setting type has no effect', value);
```
ShaMan123, line 55: As I said previously. I would prefer this to throw an error.
Think of an app that for some strange reason changes type. That app is now broken and the only indication is the console.

PR #8717, opened 2023-02-21
chore(TS): export typedefs

robothot:
## Motivation

export types in v6

## Description

solve https://github.com/fabricjs/fabric.js/discussions/8715

### DIFF
--- a/fabric.ts
+++ b/fabric.ts
@@ -50,3 +50,5 @@ export * as controlsUtils from './src/controls';
 import './src/controls/defaultControls';
 
 export * from './src/filters';
+
+export * from './src/typedefs';


wuxingping:
> ## Motivation
> export types in v6
> 
> ## Description
> solve #8715
> 
> ## Changes
> ## Gist
> ## In Action

okay,thank you,i think you should commit to merge v6(6.0.0-beta1) as well.i have been using v6.thank you so much.

ShaMan123:
Jumping in.
I have noticed the problem of types not exported.
I am not sure this fixes it properly.
We need to export all meaningful types.

robothot:
I see here that seems to have solved the problem https://github.com/fabricjs/fabric.js/pull/8725

PR #8719, opened 2023-02-21
TS() BREAKING: Moving default values away from prototype

asturur:
## Motivation

Not having default values on the protype was one of those things the team came in agreement on.
Modify the prototype somehow is considered not great.
Developers still need a way to modify the default behaviour of a class for small things like default fill, strokeWidth, or origin, or even just corner style.
So the properties that before were defined on the function prototype and for which there isn't a clear way to store on the class prototype ( if not assigning it to the prototype manually) are now stored in a static object called ownDefauts, handled by a getter static get defaultValues that goes recursively up to collect all values from the different classes.

The would still be able to change the defaults modify the content of the static ownDefaults and change all the class basic defualt values *before instances are initialized*

The caveat of this approach is that each instance has its own copy of the value now, so if you work with an high number of instances, those values are replicated in memory instead of being accessed on the prototype. This of course is a little amount of memory compared to the memory used by cache canvases in general.

This decision is made on the base that in theory class properties represent class state and not a blueprint, and so shouldn't be on the prototype. It seems red tape to me, since many state properties represent behaviour rather than state, and since a  method can be changed as a property, and since inheritance works perfectly fine.

TO BE TESTED YET: is also possible to offer a function, not exported in the main bundle but available in the codebase, that allow to restore the previous behaviour.
By emptying every ownDefaultObject and assigning the content to the prototype we go back to the previous behaviour, till that works can be kept as an option, but in the case of class implementation from JS departing even more from the old function prototype inheritance model that may be abandoned


asturur:
@ShaMan123 i think this is all i want to do for default values, with no extra layer of complexity or added code.
First step in this PR will be that i will apply to all classes to see that it actually works correctly with different level of nesting, then i ll try if is easy to change values ( didn't try yet ).

I don't think anything else is worth the complexity that brings

Controls will be looked at after this

src/shapes/Rect.ts
```diff
@@ -186,15 +200,7 @@ export class Rect extends FabricObject {
...
-export const rectDefaultValues: Partial<TClassProperties<Rect>> = {
-  type: 'rect',
-  rx: 0,
-  ry: 0,
-};
-
-Object.assign(Rect.prototype, {
-  ...rectDefaultValues,
-});
+Rect.prototype.type = 'rect';
```
asturur, line 203: those may also end up staying here for 6.0 since those aren't either fundamental or architecturally important.
Is just a string we need for legacy reason and that may be made obsolete by  Class.name at some point.

src/shapes/Object/Object.ts
```diff
@@ -895,20 +903,24 @@ export class FabricObject<
    * @param {Object} object
    */
   _removeDefaultValues(object: Record<string, any>) {
-    const prototype = classRegistry.getClass(object.type).prototype;
+    const topLevelClass = classRegistry.getClass(object.type);
```
ShaMan123, line 906:
```suggestion
    const topLevelClass = this.constructor;
```

src/shapes/Rect.ts
ShaMan123: This is why I made #8714, I saw this coming and we can already move to Class.name in that PR.

src/shapes/Object/Object.ts
```diff
@@ -610,12 +610,20 @@ export class FabricObject<
    */
   declare _transformDone?: boolean;
 
+  private static ownDefaults = fabricObjectDefaultValues;
+
+  static get defaultValues() :Record<string, any> {
```
ShaMan123, line 615: why not a method?
I prefer a method and I thought you were against getter/setter.
A method allows computation if needed

ShaMan123 (commented):
For the sake of discussion...
Why do we need default values?
- being able to tweak values w/o sublassing


what else?

src/shapes/Object/Object.ts
```diff
@@ -610,12 +610,20 @@ export class FabricObject<
...
+  }
+
   /**
    * Constructor
    * @param {Object} [options] Options object
    */
   constructor(options?: Partial<TClassProperties<FabricObject>>) {
     super();
+    // @ts-ignore constructor type
+    Object.assign(this, this.constructor.defaultValues);
```
ShaMan123, line 626:
```suggestion
    Object.assign(this, (this.constructor as typeof FabricObject).defaultValues);
```

src/shapes/Object/Object.ts
```diff
@@ -1021,7 +1033,8 @@ export class FabricObject<
       if (this.constructor.cacheProperties.includes(key)) {
         this.dirty = true;
         groupNeedsUpdate && this.group!.set('dirty', true);
-      } else if (groupNeedsUpdate && this.stateProperties.includes(key)) {
+        // @ts-ignore TS and constructor issue
+      } else if (groupNeedsUpdate && this.constructor.stateProperties.includes(key)) {
```
ShaMan123, line 1037:
```suggestion
      } else if (groupNeedsUpdate && (this.constructor as typeof FabricObject).stateProperties.includes(key)) {
```

ShaMan123 (commented):
constructor types

src/shapes/Object/Object.ts
ShaMan123: We shouldn't use `@ts-ignore`, use `@ts-expect-errror`

src/shapes/Object/Object.ts
asturur: it get confused i think when you try to access prototype from this.constructor, i need to see if it gives me back the `Function` prototype instead of the class one. i ll give it a shot

src/shapes/Object/Object.ts
asturur: this is a getter just because it started with a static default values, first then pivoted to a getter ( That allows computation too as far as i see ).
I m in general not against getters, i just don't ant to add them if i don't really need them.
But here to avoid using the prototype we need somehow to replicate the ladder of the prototypes anyway, and this is the simplest thing i could come up with.

asturur:
> For the sake of discussion... Why do we need default values?
> 
> * being able to tweak values w/o sublassing
> 
> what else?

Being to tweak the class
Being able to export without default values in the toObject
Being able to change one value in Object without having to subclass all the derivate classes because now they have to inherit from the new base class.

src/shapes/Object/Object.ts
ShaMan123: OK
`getDefaultValue() {}` can do the job 
I am for it, decide as you see fit

asturur:
@ShaMan123 i have still some broken tests i will fix asap, for now i just typed the new code. But this is more or less how the PR will be.
Filter do not need default values imho for a bunch of reasons:
- we can put the large strings/objetcs as static ( vertex/frament )
- all the other are exact settings of the filter, and not behavioural, every filter default is set as neutral, so you never use the default
- filter export is custom per filter and does not have the option to remove default values

asturur:
i hate that this add 1.2kb of code because all the new methods that needs to be repeated by instance.
I don't think there is a way around when inheritance is more than 2 level deep.

src/shapes/Object/Object.ts
asturur: yes a normal method i think is better.
.defaultValues or .getDefaultValues() is never used anyway from the developer if not for custom checks.
I'll change to a function.

asturur:
ok this is ready to go for me.
Filters in a followup because i need to type a bunch of code again.

src/shapes/Object/defaultValues.ts
```diff
@@ -31,10 +31,9 @@ export const cacheProperties = [
   'strokeMiterLimit',
   'backgroundColor',
   'clipPath',
-] as const;
+];
```
asturur, line 34: this is creating a lot of warning in the build, i removed it for now.
This is also not a const, developers are free to push here what they see may fit.
TS seems to get angry if then you say is a simple array of strings

index.node.ts
```diff
@@ -8,8 +8,7 @@ import {
 } from './fabric';
 import { FabricObject } from './src/shapes/Object/Object';
 
-// TODO: move back to default values when refactoring to method
-FabricObject.prototype.objectCaching = false;
+FabricObject.ownDefaults.objectCaching = false;
```
ShaMan123, line 11: This can reside in Object `getDefaultValues` by querying env...

src/canvas/StaticCanvas.ts
```diff
@@ -65,6 +65,25 @@ type TCanvasHydrationOption = {
   signal?: AbortSignal;
 };
 
+export const StaticCanvasDefaults = {
+  backgroundColor: '',
+  backgroundImage: null,
+  overlayColor: '',
+  overlayImage: null,
```
ShaMan123, line 72: are the blank values needed?

src/shapes/Ellipse.ts
```diff
@@ -22,6 +27,15 @@ export class Ellipse extends FabricObject {
 
   static cacheProperties = [...cacheProperties, 'rx', 'ry'];
 
+  static ownDefaults: Record<string, any> = ellipseDefaultValues;
+
+  getDefaultValues() {
+    return {
+      ...super.getDefaultValues(),
+      ...Ellipse.ownDefaults,
+    };
```
ShaMan123, line 36: if we are spreading and abandoning the shared object concept, is there a need for `ownDefaults`?

src/shapes/IText/IText.ts
```diff
@@ -18,6 +18,28 @@ type CursorBoundaries = {
...
+  editingBorderColor: 'rgba(102,153,255,0.25)',
+  cursorWidth: 2,
+  cursorColor: '',
+  cursorDelay: 1000,
+  cursorDuration: 600,
+  caching: true,
+  hiddenTextareaContainer: null,
+  _selectionDirection: null,
+  _reSpace: /\s|\n/,
+  inCompositionMode: false,
```
ShaMan123, line 36: these 3 are not really options.
`_selectionDirection` and `inCompositionMode` are state.
`_reSpace` is static?

src/shapes/Line.ts
```diff
@@ -13,31 +13,31 @@ const coordProps = ['x1', 'x2', 'y1', 'y2'];
 export class Line extends FabricObject {
```
ShaMan123, line 13: I am waiting for Line to die

src/shapes/Object/Object.ts
ShaMan123: what about this?

src/shapes/Object/defaultValues.ts
ShaMan123: I saw that too.
Another bad TS design

src/shapes/Polyline.ts
```diff
@@ -32,11 +36,28 @@ export class Polyline extends FabricObject {
...
+      ...super.getDefaultValues(),
+      ...Polyline.ownDefaults,
+    };
+  }
   /**
    * A list of properties that if changed trigger a recalculation of dimensions
    * @todo check if you really need to recalculate for all cases
    */
-  declare strokeBBoxAffectingProperties: (keyof this)[];
+  static strokeBBoxAffectingProperties: (keyof Polyline)[] = [
```
ShaMan123, line 51: maybe we rename like you did with text.
LayoutProperties?

src/shapes/Text/Text.ts
```diff
@@ -95,6 +95,67 @@ const textLayoutProperties: string[] = [
...
+    linethrough: -0.315,
+    overline: -0.88,
+  },
+  _fontSizeMult: 1.13,
+  charSpacing: 0,
+  styles: null,
+  deltaY: 0,
+  direction: 'ltr',
+  CACHE_FONT_SIZE: 400,
+  MIN_TEXT_WIDTH: 2,
```
ShaMan123, line 156: most of this is state or static or voidish initializers

src/shapes/Textbox.ts
```diff
@@ -3,6 +3,18 @@ import { TClassProperties } from '../typedefs';
 import { IText } from './IText/IText';
 import { classRegistry } from '../ClassRegistry';
 
+// @TODO: Many things here are configuration related and shouldn't be on the class nor prototype
+// regexes, list of properties that are not suppose to change by instances, magic consts.
+// this will be a separated effort
```
ShaMan123, line 8: saw this now

test/unit/activeselection.js
```diff
@@ -28,7 +28,7 @@
   QUnit.module('fabric.ActiveSelection', {
     afterEach: function() {
       canvas.clear();
-      canvas.backgroundColor = fabric.Canvas.prototype.backgroundColor;
+      canvas.backgroundColor = canvas.getDefaultValues().backgroundColor;
```
ShaMan123, line 31: this seems wrong. it is static.
```suggestion
      canvas.backgroundColor = canvas.constructor.getDefaultValues().backgroundColor;
```

Why not expose `restoreDefaultValues` or whatever?

ShaMan123 (commented):
I think `getDefaultValues` should be static. No reason for it to live on the instance.
Is there any reason. Default values is a static concept and after initialization no one will use it which is basically the definition/usage of static

index.node.ts
asturur: oh i can do that yes

src/canvas/StaticCanvas.ts
asturur: i think because are either string or gradients, and TS bother us then if are undefined.
I won't touch any default value in this PR, since this PR is just moving them.
I can experiment right away after

src/shapes/Ellipse.ts
asturur: yes, because is how you change the defaults.
Otherwise how do you change them?

Also in the description i was explaining that emptying the ownDefaults is a clean way to go back to defaults on the prototype if the developer prefer

src/shapes/IText/IText.ts
asturur: they are what they are before this pr, and this PR is just moving them away from the prototype.
I didn't even read the default values in this PR.
But yes those aren't defaults values and will be moved probably somewhere else

test/unit/activeselection.js
asturur: because we are not offering any functionality on default values, as we weren't before.
There was just a clean up of the export from defaults, and a chance to configure them.
And i don't think is useful to expand in that direction.

src/shapes/Object/Object.ts
asturur: > I think `getDefaultValues` should be static. No reason for it to live on the instance. Is there any reason. Default values is a static concept and after initialization no one will use it which is basically the definition/usage of static

i went a couple of time between static and not static and it doesn't make any difference. in general i use static functions when i know i will need them outside the instance ( like fromObject or fromElement ) but for this i don't see any need to be static.

asturur: @ShaMan123 for some weird reason github answered your question here.

src/shapes/Line.ts
asturur: you will have to wait a bit longer, i wanted to remove them too, but we are too far now with other changes.
Also this is just  extra code and removing it is not giving us any major benefit if not re-documentin and killing existing stack overflow answers.
But yes one day those are just polygons, and probably polygons are just paths

index.node.ts
asturur: i can't because we removed node from env.
This seems a good way to do it, in node, this default is configured differently.

asturur:
@ShaMan123 i added the rename and remove the use of getClass from removeDefaultValues ( i can use Object.getPrototypeOf in that case )

I think this is ready to go now, if you need some change try to mark them with request changes

src/shapes/Ellipse.ts
ShaMan123: `Object.getDefaultValues=()=> ({})`

ShaMan123:
> i went a couple of time between static and not static and it doesn't make any difference. in general i use static functions when i know i will need them outside the instance ( like fromObject or fromElement ) but for this i don't see any need to be static.

I agree. 
Default values are a class level thing. Or else they are not default. That is why IMO they are static

ShaMan123:
>  i can use Object.getPrototypeOf in that case 

I think is the same as this.constructor

ShaMan123:
Another point why static since it pollutes the instance with this niche method

asturur:
the static version is 118 bytes larger i can survive with it.
But i mstill thinking i m giving them away for free.

ShaMan123 (approved):

ShaMan123:
> But i mstill thinking i m giving them away for free.

I hope time will convince you if I didn't.

ShaMan123:
give me a green light and I will modify #8714 and remove types in favor of Class.name

asturur:
i m experimenting with filters right now on type, that is more straight forward, since type is identical to className there, and i ll see if i land on something i like there.
I'll finish this defaults/prototype/types/registry part since this is the part i didn't want to change, and if i can change it at least i ll land on something that i don't regret so much.

PR #8722, opened 2023-02-21
Fix() safeguard access to possibly undefined canvas

Ch1sKey:
## Motivation

Fix unguarded reference for `object.canvas`

## Description
`this.canvas.getRetinaScaling()` call can cause TypeError in current release in some cases. It happened with us few times.
Would be really helpful to get this fix as soon as possible. I saw it already fixed in @beta, but I don't see it coming at the nearest future.

## Changes

Enhance `.drawControls` method with guard for `.getRetinaScaling` call. Set default retinaScaling to 1.

## In Action

Smallest example to represent an issue. This is a bit far-fetched but represents edge cases which could happen in huge variety of race conditions.

Steps: Try to move object on canvas.
https://codesandbox.io/s/fabricjs-playground-forked-hpodkw

### DIFF
--- a/src/mixins/object_interactivity.mixin.js
+++ b/src/mixins/object_interactivity.mixin.js
@@ -212,7 +212,7 @@
       styleOverride = styleOverride || {};
       ctx.save();
       var retinaScaling = 1, matrix, p;
-      if(this.canvas) {
+      if (this.canvas) {
         retinaScaling = this.canvas.getRetinaScaling();
       }
       ctx.setTransform(retinaScaling, 0, 0, retinaScaling, 0, 0);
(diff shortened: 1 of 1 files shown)


ShaMan123:
note for v6: maybe we want to decide which properties (e.g. canvas) should not be public

asturur:
So i think the reproduction case is definetely a bad, meaning if you do that, you get the error.
I m ok safeguarding this, even if i would have preferred to see a real use case.

canvas shouldn't be private, because many times you may have access to an object and not its own canvas, and you use the reference to find the canvas.

The doc should really say that should be considered read only

ShaMan123:
> I m ok safeguarding this, even if i would have preferred to see a real use case.

Remember we solved this with the concurrency issue? dispose vs. destroy
So not to overkill efforts

asturur:
Yes i was just wondering if there was something else around

asturur (approved):

PR #8725, opened 2023-02-22
ci(): fix import autocomplete in dev mode

ShaMan123:
## Motivation

#8632 exposed `node` entry but it broke import autocomplete while developing fabric which is annoying.
This fixes it but makes `src` visible to devs.
We could have the publish script remove it...


asturur:
i think my node PR fixes this also, let's merge this and then let's pivot to not bundling node when we have time.
Bundling node isn't really necessary

asturur (approved):

PR #8734, opened 2023-02-23
feat(IText): Clipboard/Drag event cross platform style support

ShaMan123:
## Motivation

Continuing draggable text and supporting text styles

## Description

Adds cross platform support to the events so that styles are respected.
Tested with Google Docs, Microsoft Word (has some edge cases that can be handled in `CSSParsing`) and looks good.
Instead of setting `application/fabric` on the data transfer object it sets `text/html` which is a global standard.
Since it is a big change for a small feature (same as draggable text and text features in general) **it seems fit that it waits in line for whenever**.

## Changes

changes on top of #8708 
- `src/util/CSSParsing` - a mechanism for parsing css, needed for reading/writing the styles from/to html 
- `src/shapes/IText/DataTransferManager.ts` and `src/shapes/IText/ClipboardDataManager.ts` manage data transfer, `DraggableTextDelegate` now extends `DataTransferManager`

## In Action

Open an text editor
copy/paste drag/drop formatted text from/to fabric


src/config.ts
```diff
@@ -52,16 +52,6 @@ class BaseConfiguration {
...
 
-  /**
-   * When 'true', style information is not retained when copy/pasting text, making
-   * pasted text use destination style.
-   * Defaults to 'false'.
-   * @type Boolean
-   * @default
-   * @deprecated
-   */
-  disableStyleCopyPaste = false;
```
ShaMan123, line 63: If we want to preserve this functionality I think it is best to add an option to `IText` 
`patseOptions: source-formatting` | `destination-formatting` or something

ShaMan123 (commented):
@melchiar I think you will find this interesting and useful

ShaMan123 (commented):
changes on top of #8708 
- `src/util/CSSParsing` - a mechanism for parsing css, needed for reading/writing the styles from/to html 
- `src/shapes/IText/DataTransferManager.ts` and `src/shapes/IText/ClipboardDataManager.ts` manage data transfer

asturur:
So i think that in order for this to be merged, this should be an optional plugin.
There should just be a utility that convert text + style into HTML, and that can be imported ( not available in the bundle ) and can override  standard copy paste manager.
This is the exact thing that is fine as an utility, optionally used, unbundled, supported with tests but not available in the main bundle.
If you can do it like that you can merge it today

src/shapes/IText/DraggableTextDelegate.ts
```diff
@@ -169,30 +182,14 @@ export class DraggableTextDelegate {
...
-      const selection = (this.__dragStartSelection = {
-        selectionStart: target.selectionStart,
-        selectionEnd: target.selectionEnd,
-      });
-      const value = target._text
-        .slice(selection.selectionStart, selection.selectionEnd)
-        .join('');
-      const data = { text: target.text, value, ...selection };
-      e.dataTransfer.setData('text/plain', value);
-      e.dataTransfer.setData(
```
asturur, line 183: if you resort to having 
```
e.dataTransfer.setData(this.getFormattedClipboardData(value, styles))
```

you can just add a standard function that by default return `{ value, styles }` and the optinal one that returns what it has to return.

Who wants to use it will add it to the prototype in their own app repacing the old one, or will override the class if they prefer.

ShaMan123:
sounds great
will do

ShaMan123:
What about using dynamic imports instead?

ShaMan123 (commented):
chaos

ShaMan123 (commented):
refactored to support moduling/plugin concept

ShaMan123:
The html part can now be extract to an external plugin
What do you think?

ShaMan123:
I have lost myself in this effort and went astray
I will revisit and break to pieces
Wrong direction

PR #8735, opened 2023-02-23
fix(Intersection): bug causing selection edge case

ShaMan123:
## Motivation

I started experiencing a bug with selection
It collected objects if I clicked really fast on the canvas.

## Description

The bug lies in `isContainedInInterval`
I am to blame.
I wrote that assuming a state that apparently was incorrect.


ShaMan123 (commented):
ready

src/Intersection.ts
```diff
@@ -73,12 +76,12 @@ export class Intersection {
...
-    const b2xb1x = b2.x - b1.x,
-      a1yb1y = a1.y - b1.y,
+    const a2xa1x = a2.x - a1.x,
+      a2ya1y = a2.y - a1.y,
+      b2xb1x = b2.x - b1.x,
       b2yb1y = b2.y - b1.y,
       a1xb1x = a1.x - b1.x,
-      a2ya1y = a2.y - a1.y,
-      a2xa1x = a2.x - a1.x,
+      a1yb1y = a1.y - b1.y,
```
ShaMan123, line 84: changed order to be less confusing

ShaMan123: and found the algebra behind it and documented it

ShaMan123 (commented):
ready after cleanup and JSDOC

ShaMan123:
I want to tweak logic a bit more

asturur:
I want to read this slowly because i want to understand what the bug is and how gets fixed since the code looks fine, but reading the Math require me to dedicate a bit of time to that.
Of course if you could explain the math bug....

ShaMan123:
`isContainedInInterval` assumed that T, A, B are points on the same line.
That assumption was wrong. I thought so because I didn't fully understand the algebra in `intersectsLineLine` (not that know I do, but a lot more).
You can see the test I added that fails on master that proves the assumption was wrong.
```
T = (2, 2) 
A = (10, 3)
B = (30, 3)

isContainedInInterval(T, A, B)

const isContainedInInterval = (T: Point, A: Point, B: Point) => {
  const TA = new Point(T).subtract(A); === (-8, -1)
  const TB = new Point(T).subtract(B); === (28, 1)
  return (
    Math.sign(TA.x) !== Math.sign(TB.x) || Math.sign(TA.y) !== Math.sign(TB.y) // true but should be false
  );
};
```

The entire impl was 💩
Using vector angles is accurate but I wanted to stick to the same logic as the rest of the intersection tests.

ShaMan123 (commented):
Now it is ready for sure

ShaMan123 (commented):
finally renamed and support `infinite` to match the rest

asturur:
Is not that i have any issues with a new implementation, but at this point i m curious to understand what in 5.x this was doing and what was its scope.

src/Intersection.ts
```diff
@@ -54,8 +39,46 @@ export class Intersection {
...
+      );
+    } else {
+      const AB = createVector(A, B);
+      const AT = createVector(A, T);
+      const s = AT.divide(AB);
+      return infinite
+        ? Math.abs(s.x) === Math.abs(s.y)
+        : s.x === s.y && s.x >= 0 && s.x <= 1;
+    }
+  }
```
asturur, line 76:
```suggestion
  /**
   * check if point T is on the segment or line defined between A and B
   *
   * @param {Point} T the point we are checking for
   * @param {Point} A one extremity of the segment
   * @param {Point} B the other extremity of the segment
   * @param [extendToLine] if true checks if `T` is on the line defined by `A` and `B`
   * @returns true if `T` is contained
   */
  static isPointContained(T: Point, A: Point, B: Point, extendToLine = false) {
    if (A.eq(B)) {
      // Edge case: the segment is a point, we check for coincidence,
      // extendToLine has no meaning because there are infinite lines to consider
      return T.eq(A);
    } else if (A.x === B.x) {
      // Edge case: horizontal line.
      // we first check if T.x is on the same value, and then if T.y is contained between A.y and B.y
      return (
        T.x === A.x &&
        (extendToLine || (T.y >= Math.min(A.y, B.y) && T.y <= Math.max(A.y, B.y)))
      );
    } else if (A.y === B.y) {
      // Edge case: vertical line.
      // we first check if T.y is on the same value, and then if T.x is contained between A.x and B.x
      return (
        T.y === A.y &&
        (extendToLine || (T.x >= Math.min(A.x, B.x) && T.x <= Math.max(A.x, B.x)))
      );
    } else {
      // generic case: we normalize AT over AB.
      // then for the line case we just verify that AB has the same inclination than AT
      // for the segment case we need the same direction and the length of AT needs to be less than AB
      const AB = createVector(A, B);
      const AT = createVector(A, T);
      const s = AT.divide(AB);
      return infinite
        ? Math.abs(s.x) === Math.abs(s.y)
        : s.x === s.y && s.x >= 0 && s.x <= 1;
    }
  }
```

asturur: i think this is more clear for anyone including me, when in a week i ll forget about this code meaning

src/Point.ts
```diff
@@ -384,5 +385,3 @@ export class Point implements IPoint {
     );
   }
 }
-
-const originZero = new Point(0, 0);
```
asturur, line 388: this seems unused and with sligthly less minification potential, why the change?

ShaMan123: It was used before I refactored the logic
That I thought it makes sense and left it. I can dump it.

src/Intersection.ts
ShaMan123: it is
The only thing I dislike is the word normalize which is so generic it means nothing to me. I will add a bit to that

ShaMan123:
> Is not that i have any issues with a new implementation, but at this point i m curious to understand what in 5.x this was doing and what was its scope.

same here.

ShaMan123 (commented):
just the naming thing left, I don't mind to much

src/Intersection.ts
```diff
@@ -40,27 +40,50 @@ export class Intersection {
...
-   * check if `T` is contained in `[A, B]` by comparing the direction of the vectors from `T` to `A` and `B`
-   * @param T
-   * @param A
-   * @param B
+   * check if point T is on the segment or line defined between A and B
+   *
+   * @param {Point} T the point we are checking for
+   * @param {Point} A one extremity of the segment
+   * @param {Point} B the other extremity of the segment
+   * @param [extendToLine] if true checks if `T` is on the line defined by `A` and `B`
```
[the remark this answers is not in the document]
asturur: took me a while to understand that infinite wasn't Math.inifinite, and i was telling myself, infinite is always true.
So i tried with a different name.
Was this infinite here before we restructured the class from 5.x?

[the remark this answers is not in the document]
asturur: I see infinite i used in all the class and i think we should come up with a better name for it
considerSegmentAsLine, or something that. can be read and let not wonder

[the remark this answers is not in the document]
asturur: i ll put back infinite for consistency, and then we can rename all together

asturur (approved):

src/Intersection.ts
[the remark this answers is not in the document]
ShaMan123: > Was this infinite here before we restructured the class from 5.x?

no, I put it I think to centralize logic

PR #8738, opened 2023-02-25
CI(): replace browser test runner, `testem` => `karma`

ShaMan123:
## Motivation

We are not satisfied with the current test suite.
We discussed moving to jest.
So I did dome research and found that karma is what we should use to run jest in the browser, since shockingly (NOT!) jest is node only.
So karma is a prerequisite.
Anyways, it seems QUnit plays ok with karma so even if we postpone jest karma is a lot better than strange old testem.
It has almost everything out of the box including coverage and has an ecosystem as opposed to testem.

## Description

see [review](https://github.com/fabricjs/fabric.js/pull/8738#pullrequestreview-1314652738)


asturur:
6600 addition to package.log why i want to cry

package.json
```diff
@@ -73,7 +73,9 @@
     "@babel/preset-env": "^7.20.2",
     "@babel/preset-typescript": "^7.18.6",
     "@rollup/plugin-babel": "^6.0.3",
+    "@rollup/plugin-commonjs": "^24.0.1",
     "@rollup/plugin-json": "^6.0.0",
+    "@rollup/plugin-node-resolve": "^15.0.1",
```
asturur, line 78: do we need those 2?
i m planning to stop rolling up for node.

ShaMan123: maybe, if we want to use rollup to bundle test files one day, but for now I reverted.
The plugins for rollup and karma are shit

ShaMan123:
I think the only thing left is understanding why tests are hanging in CI

.github/workflows/tests.yml
```diff
@@ -16,17 +16,15 @@ jobs:
...
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm ci
       - run: npm run build -- -f
       - name: Run ${{ matrix.target }} ${{ matrix.suite }} headless test
-        uses: GabrielBB/xvfb-action@v1
-        with:
-          run: npm run test -- -c ${{ matrix.target }} -s ${{ matrix.suite }}
+        run: xvfb-run -a npm test -- -c ${{ matrix.target }} -s ${{ matrix.suite }}
```
ShaMan123, line 27: I think we need `xvfb-run` for firefox to have webGL
https://manpages.ubuntu.com/manpages/kinetic/en/man1/xvfb-run.1.html
https://bugzilla.mozilla.org/show_bug.cgi?id=1375585#c27

ShaMan123 (commented):
I am almost done
Just need to add description and comments. There is something to discuss about headless I have no idea about

test/karma.config.mjs
```diff
@@ -0,0 +1,167 @@
...
+// const plugins = [
+//   json(),
+//   ts(),
+//   babel({
+//     extensions: ['.ts', '.js'],
+//     babelHelpers: 'bundled',
+//   }),
+//   nodeResolve(),
+//   commonjs({ include: 'node_modules/**' })
+// ];
```
ShaMan123, line 19: once we merge this I will open a follow up that bundles tests files so we can start using esm and jest

test/karma.config.mjs
```diff
@@ -0,0 +1,167 @@
...
+          '--no-sandbox',
+          '--disable-gpu',
+          '--disable-translate',
+          '--disable-extensions',
+          '--remote-debugging-port=9222',
+        ],
+      },
+      FirefoxCI: {
+        base: 'Firefox'
+      }
```
ShaMan123, line 86: I tested with base `ChromeHeadless` and `FirefoxHeadless`
It seems there is no difference as long as CI uses `xvfb-run` but I might be wrong

test/karma.config.mjs
```diff
@@ -0,0 +1,167 @@
...
+    // list of files / patterns to load in the browser
+    files: [
+      { pattern: 'test/fixtures/**/*', included: false, served: true, watched: false, nocache: false },
+      { pattern: 'test/lib/*.js', included: true, served: true, watched: true, nocache: false },
+      { pattern: 'test/lib/tests.css', included: true, served: true, watched: true, nocache: false }, // qunit only
+      { pattern: 'test/visual/golden/**/*', included: false, served: true, watched: false, nocache: false },
+      { pattern: 'test/visual/assets/**/*', included: false, served: true, watched: false, nocache: false },
+
+      { pattern: 'dist/index.js', type: 'js', included: true, served: true, watched: true, nocache: true },
+      { pattern: 'dist/index.js.map', included: false, served: true, watched: false, nocache: true },
```
ShaMan123, line 110: follow up will also bundle src code directly so we could import modules directly even if not exposed/internal

test/karma.config.mjs
```diff
@@ -0,0 +1,167 @@
...
+
+      { pattern: 'test/karma.setup.js', type: 'js', included: true, served: true, watched: true, nocache: false },
+
+      // add test files last
+      ...(TEST_FILES ?
+        TEST_FILES.map(file => ({ pattern: file, type: 'js', included: true, served: true, watched: true, nocache: false })) :
+        [
+          process.env.TEST_SUITE === 'unit' && { pattern: 'test/unit/**/*.js', type: 'js', included: true, served: true, watched: true, nocache: false },
+          process.env.TEST_SUITE === 'visual' && { pattern: 'test/visual/**/*.js', type: 'js', included: true, served: true, watched: true, nocache: false },
+        ].filter((exists) => exists)
```
ShaMan123, line 120: I think there is no need to separate between suite any longer as well.
Browsers too since karma runs browsers in parallel
All in the follow up

test/karma.config.mjs
```diff
@@ -0,0 +1,167 @@
...
+     * https://github.com/karma-runner/karma/issues/2917#issuecomment-496473358
+     */
+    proxies: {
+      '/fixtures/': '/base/test/fixtures/',
+      '/golden_maker': '/base/test/lib/goldenMaker.html',
+      '/golden_maker.html': '/base/test/lib/goldenMaker.html',
+      '/golden/': '/base/test/visual/golden/',
+      '/assets/': '/base/test/visual/assets/',
+      '/goldens/': startGoldensServer().url,
+    },
```
ShaMan123, line 134: note the `base` prefix
An important karma thing that may drive someone mad

test/karma.config.mjs
```diff
@@ -0,0 +1,167 @@
...
+       */
+      qunit: {
+        showUI: true,
+        testTimeout: CI ? 15000 : 5000,
+        filter: process.env.QUNIT_FILTER || '',
+        reorder: false,
+        noglobals: true,
+        hidepassed: true,
+      }
+    }
```
ShaMan123, line 165: simple and powerful.
testem... you have a long way to go.

test/karma.setup.js
```diff
@@ -0,0 +1,58 @@
...
+
+// Add missing Jest functions
+// window.test = window.it;
+// window.test.each = (inputs) => (testName, test) =>
+//   inputs.forEach((args) => window.it(testName, () => test(...args)));
+// window.test.todo = function () {
+//   return undefined;
+// };
+// window.jest = jest;
+// window.expect = expect;
```
ShaMan123, line 14: this part is jest browser polyfills as suggested [here](https://github.com/tom-sherman/blog/blob/main/posts/02-running-jest-tests-in-a-browser.md), needs bundling so will wait

test/karma.setup.js
```diff
@@ -0,0 +1,58 @@
...
+
+
+window.isNode = () => false;
+
+window.TEST_CONFIG = window.__karma__.config;
```
ShaMan123, line 19: this is where the karma `client` config reaches

test/karma.setup.js
```diff
@@ -0,0 +1,59 @@
...
+
+
+// QUnit setup
+// karma-qunit treats todo as failure, see https://github.com/karma-runner/karma-qunit/issues/111
+QUnit.todo = QUnit.skip;
```
ShaMan123, line 24: another glitch to notice. todo is no longer really supported

test/unit/polygon.js
```diff
@@ -106,7 +106,7 @@
     assert.deepEqual(polygon._getTransformedDimensions(), round, 'dims should match');
   });
 
-  QUnit.todo('polygon with exactBoundingBox true and skew', function (assert) {
+  QUnit.test('polygon with exactBoundingBox true and skew', function (assert) {
```
ShaMan123, line 109: therefore, removed todo, and it should be tested. Was todo because it was in between PRs

test/visual/text_path.js
```diff
@@ -120,7 +120,7 @@
     test: 'various textpath aligned',
     code: textpath5,
     golden: 'textpath5.png',
-    percentage: 0.016,
+    percentage: 0.0175,
```
ShaMan123, line 123: make firefox pass

ShaMan123 (commented):
In general all config now resides in `test/karma.config.mjs`
`test/tests.mustache` => `test/karma.setup.js` (removed a lot of junk)
`src/Intersection.ts` to make prettier pass

READY!

test/karma.config.mjs
asturur: Why is that needed? today isn't.

test/karma.config.mjs
asturur: units vs visual is nice because it lets us separate concern on what are we testing.
Visual tests are more large scoped and probalby we need interaction tests too.
Different suites are good

asturur:
@ShaMan123 i don't want to bundle tests files nor mix the suites.
My plan for tests would be:
most utility tests move to node without brosers ( there is no need to test intersection code in browser and node just to make an example, or bbox calculation, nor class creation and properties )
Node tests would run on JS code output from babel and jest will just run over them.
No bundling. Jest is already slow, if we add layers on top of it we can forget snappy Unit tests.

All the visual and browser tests ( the one that need to interact with browser or for which browsers differences are significant ) will run in someway in browser ( karma ) but i want to avoid bundling them now.

So can you describe better what is needed in the followup and why?

test/karma.config.mjs
ShaMan123: it doesn't mean you won't be able to run a single suite standalone.
It means that if you want to run 2 tests from different suites it won't spawn 2 servers/browsers etc.
And it might have other implications with code coverage

ShaMan123:
I have in mind the following:
Starting working on fabric I run `npm run dev`. This starts watching everything, src and tests, using rollup and outputs to dist generated code already passed through nyc istanbul.
Then I run `npm test` select what test I want. It waits for the build to complete (just as it does now) and runs.
I can then write tests that import files directly for fabric. I write in esm, snapshots reside in separate file. Navigating tests becomes easy and coverage is supported for browsers.
Then I want to stub a method using sinon. I don't need to mess aroud with ugly test config. I install the dev package and import it. DONE. easy. fun. simple.
As it is now, you already do `npm test -- -w -f` before starting tests, so there is no difference in that but we gain everything else a normal, modern test framework has to offer us.

ShaMan123:
I want adding tests being so simple I will want to do it.
Today I force myself and battle tests to reproduce edge cases I manage to reproduce in the browser with ease

ShaMan123:
Anyways I can remove those comments.

ShaMan123:
bye bye
playwright won

PR #8742, opened 2023-02-26
chore(TS): BREAKING move also defaults for filters away from prototype

asturur:
## Motivation
Fortunately the defaults values for filters aren't important at all.
The only real change is the fragmentSource that doesn't exists anymore as a private string/object property but is always returned from the getFragmentSource function.

We still have a default value object otherwise we need to declare a constructor per instance.
Probably final typings will force us to add a constructor and in that case the concept of defaults for filters can be removed.
For now is at least terser because focus all the parameter logic in the base class.

The duality of BaseFilter<string> or BaseFilter<Record> is not needed anymore and has been removed ( was introduced for the migration )

Removed the duality between diff and difference for BlendColor. `diff` was introduced 9 years ago, but the correct name is `difference` that was paired with it later. Was never deprecated


asturur:
@ShaMan123 having the fragmentSource on the instance looks very strange
I was thinking of removing fragmentSource from the instance and just having getFramentSource method, that today is returning this.fragmentSource, doing a different job per class.
In this way the AbstractBaseFIlter<T> goes away and we have only BaseFilter.

Any concerns/ thoughts?

ShaMan123:
I am for

ShaMan123:
Is fragmentSource a static property? Isn't it per filter class?

asturur:
is it per class and there is no need to to have it as a property at all.
it can be returned fro getFragmentSource and getVertexSource

Adding `vertexSource` or `fragmentSource` as static or non static does not save us from having a method for the filters that have more than one possible fragment source.

at this point i keep just the method and i return the string from there

asturur:
the fact that all the code i added saved 1.3kb leaves me a bit confused.
I definitely added more than removed, with all the getFragmentSource and imports files

asturur:
Ok this is ready to go.
There are probably optimizations that can be done for main parameter ( but unless they save me another half kb i m not committing them )

asturur:
This is ready to go i think.
And i think it completes the part of the default values that is necessary for 6.0
If i can get away from `type` from objects as i did for filter i ll remove also that, otherwise not.

asturur:
oh no controls yet.
But controls can be solved cleanly with the same strategy.

ShaMan123:
I will review now.
I did type as you suggested #8714

src/filters/BlendColor.ts
```diff
@@ -1,8 +1,27 @@
 import { Color } from '../color/Color';
 import { TClassProperties } from '../typedefs';
-import { AbstractBaseFilter } from './BaseFilter';
+import { BaseFilter } from './BaseFilter';
 import { T2DPipelineState, TWebGLUniformLocationMap } from './typedefs';
 import { classRegistry } from '../ClassRegistry';
+import { blendColorFragmentSource } from './blendColor.shaders';
+
+type TBlendMode =
```
ShaMan123, line 8: I think we should export most types in general

src/filters/Blur.ts
```diff
@@ -8,6 +8,12 @@ import type {
 } from './typedefs';
 import { isWebGLPipelineState } from './typedefs';
 import { classRegistry } from '../ClassRegistry';
+import { fragmentSource } from './blur.shaders';
+
+export const blurDefaultValues: Partial<TClassProperties<Blur>> = {
+  blur: 0,
+  mainParameter: 'blur',
```
ShaMan123, line 15: shouldn't this be static as well?

src/filters/ColorMatrixFilters.ts
```diff
@@ -1,38 +1,27 @@
...
+    };
 
     static async fromObject(object: any) {
-      return new GeneratedColorMatrix(object);
+      return new this(object);
     }
   };
+  classRegistry.setClass(newClass, key);
+  return newClass;
 }
```
ShaMan123, line 25: I would make this not a mixin but a subclass
See https://github.com/fabricjs/fabric.js/blob/08e9431b6c19f763216f9061d7cec25774c58961/src/filters/ColorMatrixFilters.ts

src/filters/Resize.ts
```diff
@@ -1,12 +1,26 @@
...
+  resizeType: 'hermite',
+  scaleX: 1,
+  scaleY: 1,
+  lanczosLobes: 3,
+  fragmentSourceTOP: `
+    precision highp float;
+    uniform sampler2D uTexture;
+    uniform vec2 uDelta;
+    varying vec2 vTexCoord;
+  `,
```
ShaMan123, line 19: for consistency should be moved to a standalone file?

src/filters/Resize.ts
```diff
@@ -1,12 +1,26 @@
...
+};
+
+type TResizeType = 'bilinear' | 'hermite' | 'sliceHack' | 'lanczos';
```
ShaMan123, line 22: export

src/filters/BaseFilter.ts
```diff
@@ -12,33 +12,37 @@ import { GLPrecision } from './GLProbes/GLProbe';
...
-  declare vertexSource: string;
-
-  declare fragmentSource: T;
+  vertexSource = `
+    attribute vec2 aPosition;
+    varying vec2 vTexCoord;
+    void main() {
+      vTexCoord = aPosition;
+      gl_Position = vec4(aPosition * 2.0 - 1.0, 0.0, 1.0);
+    }`;
```
ShaMan123, line 45: this as well? Moving all this code out

src/filters/BlendImage.ts
```diff
@@ -1,14 +1,33 @@
...
+    varying vec2 vTexCoord;
+    varying vec2 vTexCoord2;
+    uniform mat3 uTransformMatrix;
+    void main() {
+      vTexCoord = aPosition;
+      vTexCoord2 = (uTransformMatrix * vec3(aPosition, 1.0)).xy;
+      gl_Position = vec4(aPosition * 2.0 - 1.0, 0.0, 1.0);
+    }
+    `,
+};
```
ShaMan123, line 30: same

ShaMan123 (changes requested):
This is a great cleanup! Files are so much readable and less intimidating (for me they were, WebGL still is)

- naming: can we keep camelCase or UpperFirst convesion? `colorMatrix.shaders` => `ColorMatrixShaders`, or perhaps create a folder `Shaders` and move all of them there. I like that you took it out and I like the idea of having them in a `Shaders` folder. They are consts so seems right to me.
- moving `fromObject` to base class with `new this` will save a lot of lines, there are only 2-4 filters that override it
- extract the rest of WebGL c code (is it c?) from filter files

src/filters/ColorMatrixFilters.ts
ShaMan123: mixins suck in TS

ShaMan123:
> * moving `fromObject` to base class with `new this` will save a lot of lines, there are only 2-4 filters that override it

when you do this make sure to spread the object and remove `type` from it so it won't override the getter (I am guessing that is happening currently)

src/filters/Blur.ts
asturur: (i already tried 3 times to remove it entirely or reduce its appareances)

src/filters/ColorMatrixFilters.ts
asturur: but this is not a mixin. this is a subclass generated with a function.

src/filters/Resize.ts
asturur: this resize one is too messy. but shouldn't be a property either, i will move this

asturur:
> > * moving `fromObject` to base class with `new this` will save a lot of lines, there are only 2-4 filters that override it
> 
> when you do this make sure to spread the object and remove `type` from it so it won't override the getter (I am guessing that is happening currently)

No is not happening because the constructor is avoing getting type in.
Right i forgot about new this(), and i played around with gettin the class from class registry and then i didn't like it.
i ll do this now, but not touching anything else for now.

asturur:
> * naming: can we keep camelCase or UpperFirst convesion? `colorMatrix.shaders` => `ColorMatrixShaders`, or perhaps create a folder `Shaders` and move all of them there. I like that you took it out and I like the idea of having them in a `Shaders` folder. They are consts so seems right to me.

i can move them in a folder yes, that will make the .shader. go away

asturur:
> * extract the rest of WebGL c code (is it c?) from filter files

is shader language, not sure the exact definition.
The syntax is C like but the rules are different.
Also the assumptions are different

src/filters/ColorMatrixFilters.ts
ShaMan123: I meant dynamic classes, sorry

asturur:
There are additional things that can be done here:
- resize filter can have the shader code extracted ( followup )
- some optimizations on how we do sendUniformData and getUniformLocations so that they can be defined as a map of a function rather than methods (helps abstracting writing a filter) ( follow up)
- the highp/medium/low precision code can be made common and injected in all shaders

PR #8744, opened 2023-02-27
fix(MultiSelection): add target from behind active selection

ShaMan123:
## Motivation

Adding a target from behind active selection was not possible

## Description
- if active selection is the target of the event we search for children to remove from selection or objects behind it to add to selection
- in case a group and all it's objects are selected we deselect group: fixes weird border box drawing


ShaMan123 (commented):
ready

src/shapes/ActiveSelection.ts
```diff
@@ -75,6 +75,13 @@ export class ActiveSelection extends Group {
       const parent = object.group;
       parent._exitGroup(object);
       object.__owningGroup = parent;
+      if (
+        this._objects.includes(parent) &&
+        parent._objects.every((object) => this._objects.includes(object))
+      ) {
+        // in case group and all it's objects are selected we deselect group
+        this.remove(parent);
+      }
```
asturur, line 84: this kind of interactivity with groups opens for a lot of edge cases.
I don't think automatically deselecting things outside user clicks is correct since we are assuming which is the correct UX here.
Why is deselecting the parent or the group or anything correct over the other option?
Does it create an interaction issue or just feel strange?

ShaMan123: with interactive groups it will be possible to deselect them in a multi selection interaction if only one object remains unselected in group since the object becomes the deselection target and the group will hang around, **SELECTED**

ShaMan123:
> hich PR broke it?
> I think it was behind a modified key in 5.x

I am not sure. Looking into it. But 5.x supports this

https://user-images.githubusercontent.com/34343793/221576763-ffe96acd-a212-42e7-a8cc-c0054084f866.mp4

src/shapes/ActiveSelection.ts
ShaMan123: I agree maybe it should be moved to multiSelectAdd

ShaMan123: I remember you said groups are simply containers.
What meaning does an empty selected container have in a selection context?

ShaMan123:
Found it.
#8665 I refactored `findTarget` not to accept `skipGroup` and moved the logic to where it is now, and I missed this case of needing to search all objects, not only selected objects

```diff

    findTarget: function (e, skipGroup) {
      if (this.skipTargetFind) {
        return;
      }

      var ignoreZoom = true,
          pointer = this.getPointer(e, ignoreZoom),
          activeObject = this._activeObject,
          aObjects = this.getActiveObjects(),
          activeTarget, activeTargetSubs,
          isTouch = isTouchEvent(e),
+          shouldLookForActive = false;

      // first check current group (if one exists)
      // active group does not check sub targets like normal groups.
      // if active group just exits.
      this.targets = [];
+ // skipGroup skips all logic that I commented out
      // if we hit the corner of an activeObject, let's return that.
      // if (shouldLookForActive && activeObject._findTargetCorner(pointer, isTouch)) {
      //   return activeObject;
      // }
      // if (aObjects.length > 1 && !skipGroup && activeObject === this._searchPossibleTargets([activeObject], pointer)) {
      //   return activeObject;
      // }
      // if (aObjects.length === 1 &&
      //   activeObject === this._searchPossibleTargets([activeObject], pointer)) {
      //   if (!this.preserveObjectStacking) {
      //     return activeObject;
      //   }
      //   else {
      //     activeTarget = activeObject;
      //     activeTargetSubs = this.targets;
      //     this.targets = [];
      //   }
      // }
      var target = this._searchPossibleTargets(this._objects, pointer);
      if (e[this.altSelectionKey] && target && activeTarget && target !== activeTarget) {
        target = activeTarget;
        this.targets = activeTargetSubs;
      }
      return target;
    },
```

src/shapes/ActiveSelection.ts
asturur: Groups were simple containers, yes, meaning the would just contain object and render them.
Now they are interactive, and the surface edge cases.
Most of them should be left to the developer, and we should patch the ones that can be caused by the user alone.
Probably is enough to avoid rendering the control boxes, i m not sure at this point since it doesn't make sense to select a group and the objects inside.

What i would answer quickly would be that if the group is selected, you can't select objects inside.
And probably is the solution i would adapt if it  simplify our life.

asturur:
> > hich PR broke it?
> > I think it was behind a modified key in 5.x
> 
> I am not sure. Looking into it. But 5.x supports this

Yes this was made possible in findTarget, i think at the end of the function where we would say that if there wasn't a sub target in the activeSelection, but there was eventually a target under it we would take that one

src/canvas/Canvas.ts
```diff
@@ -1489,9 +1489,14 @@ export class Canvas extends SelectableCanvas {
...
-          // find target from active objects
+          // find target from active objects to remove or from under active selection to add
           target = this.searchPossibleTargets(
-            prevActiveObjects,
+            [
+              ...this._objects.filter(
+                (object) => !prevActiveObjects.includes(object)
+              ),
+              ...prevActiveObjects,
+            ],
```
asturur, line 1499: I m trying to wrap my head around why we need to change the stack for searchPossibleTarget.
I understand you want the activeSelectionObject searched first, but i think that is job for `altSelectionKey`.
Isn'it?
Shouldn't we use the natural stack order here?

src/shapes/ActiveSelection.ts
ShaMan123: > What i would answer quickly would be that if the group is selected, you can't select objects inside.
> And probably is the solution i would adapt if it simplifies our life.

Not bad

ShaMan123: Your suggestion makes sense. If we agree on it I can try a patch

src/canvas/Canvas.ts
ShaMan123: > Yes this was made possible in findTarget, i think at the end of the function where we would say that if there wasn't a sub target in the activeSelection, but there was eventually a target under it we would take that one

I followed this logic and this comment

https://github.com/fabricjs/fabric.js/blob/26ed225b47288f775e598f719e218a660dd58dd0/src/canvas/SelectableCanvas.ts#L989

ShaMan123:
@asturur ?

ShaMan123:
@asturur what about this?

src/canvas/Canvas.ts
ShaMan123: I guess we can achieve it in a different way.
If `!target` => search the rest of the object on canvas to see if there is an object behind the selection that we can add

src/canvas/Canvas.ts
```diff
@@ -1481,11 +1481,13 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-            prevActiveObjects,
-            this.getPointer(e, true)
-          );
+          const pointer = this.getPointer(e, true);
+          target =
+            // first search active objects for a target to remove
+            this.searchPossibleTargets(prevActiveObjects, pointer) ||
+            //  if not found, search under active selection for a target to add
+            // `prevActiveObjects` will be searched but we already know they will not be found
+            this.searchPossibleTargets(this._objects, pointer);
```
ShaMan123, line 1490: I have made this simpler perhaps in 347ff7fec - revert it to see the prev revision
I belive this is more performant also

ShaMan123 (commented):
OK

src/canvas/Canvas.ts
ShaMan123: and closer to https://github.com/fabricjs/fabric.js/blob/master/src/canvas/SelectableCanvas.ts#L741C5-L743C15

asturur (approved):

asturur (approved):

PR #8745, opened 2023-02-27
refactor(): Control connection rendering

ShaMan123:
## Description

Moving the connecting line rendering to Control for 2 reasons:
- scoping logic to control
- one less thing to fix when trying to tackle the flipped group controls issue (which I wasted the entire afternoon on)

## Changes

BREAKING control rendering method signature


ShaMan123 (commented):
I am agnostic about this
It is better design

ShaMan123 (commented):
76691b0d0 proves qrDecompose is to blame for the group flip bug

asturur:
> [76691b0](https://github.com/fabricjs/fabric.js/commit/76691b0d080900264165e206cda86420ea9869f4) proves qrDecompose is to blame for the group flip bug

Yes we know that.
Because again, when you multiply transform matrix together you loose the history.
It is to blame for the lost of the flip information not for the control flip bug.

asturur:
This PR would be perfectly fine, if it wasn't breaking the control interface for the sake of breaking it

asturur (changes requested):
It make sense to render the connection for each control rather than having a loop of connections for each control separated from the contrl box.. The refactor of the method is effective ( in organization ) even without breaking the interface, so please restore left and top, even if the are less ergonomic than the point, is a psicological difference that makes no difference in practice, while breaking the interface does.

ShaMan123:
Yes, I don't mean to break for breaking. I was working on a few fixes in parallel so some things might have spilled over here

ShaMan123:
In order not to break I need to expose another method that calls `render` and `renderConnection` or else how would I have the info for rendering the connection line?

ShaMan123:
I can make renderConnection a standalone and disconnect it from `render` but that is ugly

asturur:
> In order not to break I need to expose another method that calls `render` and `renderConnection` or else how would I have the info for rendering the connection line?

i don't understand.
To me the only breaking change is swapping top,left with a point. And that is unnecessary.
What else is breaking? all the rest seem additive

src/shapes/Object/InteractiveObject.ts
```diff
@@ -217,11 +218,18 @@ export class InteractiveFabricObject<
...
+        this,
+        control
+      );
       // coords[key] are sometimes used as points. Those are points to which we add
       // the property corner and touchCorner from `_calcCornerCoords`.
       // don't remove this assign for an object spread.
       coords[key] = Object.assign(
         position,
+        { connection: connectionPosition },
         this._calcCornerCoords(control, position)
```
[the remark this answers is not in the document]
asturur: as long as `coords[key]` continue to have .x and .y as props, and tl, tr, bl, br available, you can add more if you want to have x and y also available under a `point` property. But the old ones should stay around.

src/controls/Control.ts
```diff
@@ -332,33 +342,62 @@ export class Control {
...
-    styleOverride: ControlRenderingStyleOverride | undefined,
+    position: TOCoord,
+    styleOverride: ControlRenderingStyleOverride = {},
     fabricObject: FabricObject
   ) {
-    styleOverride = styleOverride || {};
+    if (this.withConnection) {
+      this.renderConnection(
+        ctx,
+        position.connection,
```
asturur, line 389: oh i see this is the breaking issue.
You can just call this separately no?

src/shapes/Object/InteractiveObject.ts
```diff
@@ -491,8 +467,7 @@ export class InteractiveFabricObject<
     this.setCoords();
     this.forEachControl((control, key) => {
       if (control.getVisibility(this, key)) {
-        const p = this.oCoords[key];
-        control.render(ctx, p.x, p.y, options, this);
+        control.render(ctx, this.oCoords[key], options, this);
```
asturur, line 470: Can't you call `control.renderConnection` before `control.render` to avoid the breaking change?

ShaMan123: I thought of that...
I don't like it because then the rendering call is not centralized so overriding becomes annoying and hard to customize.
I suggest exposing a calling method, maybe named `renderControl` that will call `renderConnection` and `render` and eventually deprecate `render` since I have a few suggestions in terms of matrices to pass to this method, now that I reworked coords.
Thoughts?

src/controls/Control.ts
ShaMan123: correct

src/shapes/Object/InteractiveObject.ts
asturur: There is nothing to customize in renderConnection right now.
To me it seems totally fine to do those things slowly, this change isn't fixing anything, is moving the line responsibility from canvas to the control. And since we dont want to break this interface for now is basically the only thing we can do.

Regarding:
- i will have a method that call the old render + renderConnection
VS
- i will call render + renderConnection from here

little changes since the point is that we are not killing the old render, and so your freedom is somehow limited.
The only thing you would gain is that you can pass more things to a new render, and use them in renderConnection.

ShaMan123: great

src/shapes/Object/InteractiveObject.ts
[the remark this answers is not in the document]
ShaMan123: Doesn't really matter if Control#render receives what it needs?
I want to remove assigning the value

src/shapes/Object/InteractiveObject.ts
ShaMan123: I will use it for my fix of coords as well

ShaMan123 (commented):
I believe this is ready

ShaMan123:
done

ShaMan123 (commented):
squashed `_calcCornerCcoords` into `calcOCoords` 
broke oCoords - not a point anymore

ShaMan123:
merged into #8767

PR #8747, opened 2023-02-28
fix(Geometry): coords not respecting group

ShaMan123:
## Motivation

A leftover from the group rewrite #7670
Came to my attention again with #8723

## Description

`_getCoords` didn't respect group. `getCoords` did.
2 Methods touched the unsafe `_getCoords` apart from `getCoords`
- `containsPoint`
- `isContainedWithinObject`

At the time I didn't fully understand how `_checkTarget` works so I was afraid to break it. And left is as is. But a bug surfaced due to `isContainedWithinObject` inside group.

## Changes
`_getCoords`, `aCoords`, `lineCoords` all respect group
`containsPoint` usage is now with absolute true


ShaMan123 (commented):
seems correct, please double check me
it is complex

ShaMan123:
I made a real breakthrough still WIP on my local branch, I want to discuss it. I can perhaps ship something not breaking for v6. and the rest looks like v7

wuxingping:
how to install the version of patch-coords?

ShaMan123:
npm i fabricjs/fabric.js#branch_name

asturur:
@ShaMan123 there is so much history on aCoord, oCoords, lineCoords, and so many things that were implemented as partial changes, knowing that the system needed an overhaul.

Let's have a chat around those changes at the first occasion and let's go through all the reasons why coords are in this way.
The worst enemy is padding, that with selections inside groups is probably a mess ( i didn't check ).

ShaMan123:
I reviewed the entire logic and I have a proposal. This is not the entire story. I know understand how your proposal of bbox calculation fits in perfectly and might be none breaking and completely under the hood, plus better perf and accuracy

ShaMan123 (commented):
I am working on a real fix and not just a patch so this can be discarded or merged and then we pull our sleeves to fix coords from scratch.

ShaMan123:
handled by #8767

PR #8752, opened 2023-03-04
fix(): `endCurrentTransform` `mouseUp` overlap

ShaMan123:
## Motivation
https://github.com/fabricjs/fabric.js/discussions/8739#discussioncomment-5172270

## Description

This PR is another effort of the TS migration, scoping canvas classes.
I moved all object transform logic to Canvas, but I wish it to reside in a standalone class (A transform event).
This made it easy to fix the mentioned bug: removing the active object from within the modified event will trigger another `endCurrentTransform` and fire another modified event

## Changes

- moved logic that belong to Canvas transform logic from SelectableCanvas:
  - `_currentTransform`
  - `_shouldCenterTransform`
  - `_getOriginFromCorner`
  - `_setupCurrentTransform`
- squashed `_beforeTransform` into `_setupCurrentTransform`
- squashed `_finalizeCurrentTransform` into `endCurrentTransform`
- exposed `onObjectDiscarded` to run scoped discard logic
- `_findTargetCorner` now returns an empty string if nothing is found


ShaMan123 (commented):
I would like to extract all transform logic to a delegate

PR #8753, opened 2023-03-04
chore(TS) remove controls from prototype

asturur:
## Motivation

Controls are the last part of the data we have on the protoype.
This pr removes them and put them in the defaults.
Controls are still shared between instances because of javascript mutability of objects.


asturur:
@ShaMan123 probably we can go up to having a different control object per instance, since a way to rollback with few lines of code is still possible ( if the developer desire so ).

ShaMan123 (approved):
Great.
I didn't expect this pivot from you.

asturur:
> Great. I didn't expect this pivot from you.

I want simple code and options.
I don't want to maintain a project that is more complicated than an hobby, it adds too much friction on working on it when i should be doing fun things.

ShaMan123:
I want to add that logic that was adapted to shared controls (e.g. control visibility, passing down the object that uses the control etc.) might be thought of as obsolete.
I am leaving this comment to point out that in terms of scoping we can attach the object ref to a control and scope visibility to the control and not to the object. Same goes for coords.
Then doing something like `object.control.br.isVisible()`, `object.control.br.containsPoint()` is easy

ShaMan123:
And we can make the controls object a class holding more methods such as setting visibility

PR #8755, opened 2023-03-05
ci(): remove unused imports on save

ShaMan123:
## Motivation

You don't want organizing imports, I don't want manually removing unused imports

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,6 +12,7 @@
   "editor.formatOnSave": true,
   "editor.formatOnPaste": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": false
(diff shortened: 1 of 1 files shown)


asturur:
I see we miss the point again of my complain.
I don't want to see unnecessary code changes that i have to read before move on.
example:
image

Is not that i don't want organized imports.

I don't want files to changes just because a particular brand of editor has its own method of organizaing imports ( external deps first, vs type first, vs alphabetical, vs other ).

Since we can't have everyone using the same editor, and since who made prettier didn't consider that a priority, we don't have it.

asturur (approved):

ShaMan123:
The whole poiny of this pr is so i can easily respect your demand.

PR #8507, opened 2022-12-12
port Registry I/O

ShaMan123:
## Description

ports what remains from #8282 to #8500

## Changes

- renamed: `setClass` to `setJSONClass`, `getClass` to `getJSONClass`
- exposed `setClass`: register class to both JSON and SVG registeries
- changed signature of `getXXXClass` accepting a `strict` param indicating if it should throw or not
- use registry to resolve `enlivenObjectEnlivenables`
- changed `getJSONClass` signature to accept an object, needed to resolve Gradient in `enlivenObjectEnlivenables`
- refactored `enlivenObjectEnlivenables` to return an object **only** with enlived stuff, previously was returning everything
- changed Gradient to fit the registry
-  enliven Shadow in `enlivenObjectEnlivables` not in `set` (I don't think it is breaking but devs can add `type: shadow` to data in case they fall into an edge case)
- fix `fromObject` regression 3fd06607d

## Gist

closes #8282

PR #8756, opened 2023-03-05
chore(TS): more typings for toObject and classes

asturur:
## Motivation

This PR fixes all the typing warnings we accumulated in previous PRs and tries to extend the toObject and fromObject typiings to all classes.

The PR balooned out while trying to fix the typing that were spread across all files and classes.


asturur:
@ShaMan123 right now we got full of TS complains because of those new types.
I ll move forward converting everything to see if some auto resolve when the approach is consistent, but in case not, we need to find a solution of the new complains.
Maybe even be sparkle ts-expect-error(s) around

src/shapes/Image.ts
```diff
@@ -32,10 +49,30 @@ export const imageDefaultValues: Partial<TClassProperties<Image>> = {
...
 
+export interface SerializedImageProps
+  extends SerializedObjectProps {
+    src: string;
+    crossOrigin: string | null;
+    filters: any[];
+    resizeFilter?: any;
+    cropX: number;
+    cropY: number;
+  };
```
asturur, line 60: probably those will be similar in content, but different in types.
i think both crossOrigin and filters apply to both

asturur:
i ll try to do everythiing but not Text related classes in this PR.

ShaMan123:
I think we shouldn't add ts-expect-error yet, but only after we are done.

asturur:
is easy to remove, while is very hard right now to see what happens during the build, since the amount of warning is past the scrollable limit.
I won't commit ts-expect-errors, but if i need to, is simpler to put them while i work on the files

ShaMan123:
@asturur what if we pass a single interface/type to FabricObject instead of a few?
It will contain the props, serialized props and events keys.
It should be more flexible and should reduce the line count anf might be more readable.
Also if we need to change only a part it is definately easier than having it as a list of arguments we must specify.

src/Collection.ts
```diff
@@ -309,19 +310,20 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
     /**
      * Given a bounding box, return all the objects of the collection that are contained in the bounding box.
      * If `includeIntersecting` is true, return also the objects that intersect the bounding box as well.
+     * This is meant to work with selection. Is not a generic method.
```
ShaMan123, line 313: Why? I disagree. I was turned into a generic method

src/Collection.ts
```diff
@@ -309,19 +310,20 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
...
      * @returns array of objects contained in the bounding box, ordered from top to bottom stacking wise
      */
     collectObjects(
       { left, top, width, height }: TBBox,
       { includeIntersecting = true }: { includeIntersecting?: boolean } = {}
     ) {
-      const objects: BaseFabricObject[] = [],
+      const objects: InteractiveFabricObject[] = [],
```
ShaMan123, line 320: we should the object type as a generic to Collection

src/shapes/Group.ts
```diff
@@ -978,7 +980,7 @@ export class Group extends createCollectionMixin(
    */
   toObject(propertiesToInclude: (keyof this)[] = []) {
     return {
-      ...super.toObject([
+      ...this.toObject([
```
ShaMan123, line 983: ERROR

src/shapes/Path.ts
```diff
@@ -165,9 +174,12 @@ export class Path extends FabricObject {
...
    * @return {Object} object representation of an instance
    */
-  toObject(propertiesToInclude: (keyof this)[] = []) {
+  toObject<
+    T extends Omit<Props & TClassProperties<this>, keyof SProps>,
+    K extends keyof T = never
+  >(propertiesToInclude: K[] = []): { [R in K]: T[K] } & SProps {
     return {
-      ...super.toObject(propertiesToInclude),
+      ...this.toObject(propertiesToInclude),
```
ShaMan123, line 182: ERROR
circular

src/shapes/Polyline.ts
```diff
@@ -213,9 +228,12 @@ export class Polyline extends FabricObject {
...
    * @return {Object} Object representation of an instance
    */
-  toObject(propertiesToInclude?: string[]): object {
+  toObject<
+    T extends Omit<Props & TClassProperties<this>, keyof SProps>,
+    K extends keyof T = never
+  >(propertiesToInclude: K[] = []): { [R in K]: T[K] } & SProps {
     return {
-      ...super.toObject(propertiesToInclude),
+      ...this.toObject(propertiesToInclude),
```
ShaMan123, line 236: again

src/shapes/Rect.ts
```diff
@@ -122,8 +148,12 @@ export class Rect extends FabricObject {
...
    * @return {Object} object representation of an instance
    */
-  toObject(propertiesToInclude: string[] = []) {
-    return super.toObject(['rx', 'ry', ...propertiesToInclude]);
+  toObject<
+    T extends Omit<Props & TClassProperties<this>, keyof SProps>,
+    K extends keyof T = never
+  >(propertiesToInclude: K[] = []): { [R in K]: T[K] } & SProps {
+    // @ts-ignore toObject typing does not really work
+    return this.toObject([...RECT_PROPS, ...propertiesToInclude]);
```
ShaMan123, line 156: as well

src/shapes/Object/types/ObjectProps.ts
```diff
@@ -8,6 +10,7 @@ export interface ObjectProps extends SerializedObjectProps, ClipPathProps {
   fill: TFiller | string | null;
   stroke: TFiller | string | null;
   shadow: Shadow | null;
+  canvas?: StaticCanvas | Canvas;
```
ShaMan123, line 13:
```suggestion
  canvas?: StaticCanvas;
```

And in `src/shapes/Object/types/FabricObjectProps.ts`
```ts
  canvas?: Canvas;
```

ShaMan123 (commented):
`super.toObject` => `this.toObject` is a mistake I guess

src/Collection.ts
asturur: it isn't because it rely on something that is specific to the selection since it reads object.selectable.

asturur:
@ShaMan123 i m not done yet here, i just turned this PR into removing all the yellow warnings in the build process

asturur:
> @asturur what if we pass a single interface/type to FabricObject instead of a few? It will contain the props, serialized props and events keys instead of them being args. It should be more flexible and should reduce the line count and might be more readable. Also if we need to change only a part it is definately easier than having it as a list of arguments we must specify.

Now is like it is, i adapted all to this 3 arg pattern.
If you have a better idea you can redo it as long as typescript build without errors

src/shapes/Group.ts
asturur: Yes i have been trying to have this not complain and i tried different things, one was remove the toObject method from the base class

src/shapes/Image.ts
```diff
@@ -782,13 +780,14 @@ export class Image<
...
+  static fromObject<T extends TProps<SerializedImageProps>>(
+    { filters: f, resizeFilter: rf, src, crossOrigin, ...object }: T,
     options: { signal: AbortSignal }
-  ): Promise<Image> {
+  ) {
     return Promise.all([
       loadImage(src, { ...options, crossOrigin }),
       f && enlivenObjects(f, options),
+      // TODO: redundant - handled by enlivenObjectEnlivables
       rf && enlivenObjects([rf], options),
```
ShaMan123, line 791: we have been visiting this line a few times

src/shapes/Text/StyledText.ts
```diff
@@ -1,19 +1,28 @@
...
-> extends FabricObject<EventSpec> {
+  Props extends TProps<FabricObjectProps> = Partial<FabricObjectProps>,
+  SProps extends SerializedObjectProps = SerializedObjectProps,
+  EventSpec extends ObjectEvents = ObjectEvents
+> extends FabricObject<Props, SProps, EventSpec> {
   declare abstract styles: TextStyle;
   protected declare abstract _textLines: string[][];
   protected declare _forceClearCache: boolean;
-  protected declare abstract _styleProperties: string[];
+  static _styleProperties = styleProperties;
```
ShaMan123, line 25:
```suggestion
  static styleProperties = styleProperties;
```

src/shapes/Text/StyledText.ts
```diff
@@ -260,9 +269,10 @@ export abstract class StyledText<
    */
   getCompleteStyleDeclaration(lineIndex: number, charIndex: number) {
     const style = this._getStyleDeclaration(lineIndex, charIndex) || {},
-      styleObject: TextStyleDeclaration = {};
-    for (let i = 0; i < this._styleProperties.length; i++) {
-      const prop = this._styleProperties[i];
+      styleObject: TextStyleDeclaration = {},
+      styleProps = (this.constructor as typeof StyledText)._styleProperties;
+    for (let i = 0; i < styleProps.length; i++) {
+      const prop = styleProps[i];
```
ShaMan123, line 275: can be refactored to use `pickBy`?

ShaMan123 (commented):
- `{ [R in K]: T[K] }` should have been `{ [P in K]: T[P] }` => `Pick<T, K>`
- fine tuning `toObject`/`fromObject` types
- Rect, Ellipse need a `fromObject` declaration
- `pickBy` fixed Shadow TS error and is cleaner, now used in Object as well

asturur:
> * Rect, Ellipse need a `fromObject` declaration

In case declaration yes, code no.

src/shapes/Image.ts
asturur: i prefer it clearly states what is going on here.
If we need it to enlive filters, i want to enlive also the resizeFilter manually. is more clear

src/shapes/Polyline.ts
```diff
@@ -21,12 +21,18 @@ export const polylineDefaultValues: Partial<TClassProperties<Polyline>> = {
...
-export interface SerializedPolylineProps extends SerializedObjectProps {
+export interface UniquePolylineProps {
   points: XY[];
 }
 
+export interface SerializedPolylineProps
+  extends SerializedObjectProps,
+    UniquePolylineProps {}
+
+export interface PolylineProps extends FabricObjectProps, UniquePolylineProps {}
```
asturur, line 33: this is not correct, i didn't do it on purpose because points is not part of polylineProps. only of the serialized one.

src/Shadow.ts
```diff
@@ -19,12 +20,12 @@ export const shadowDefaultValues: Partial<TClassProperties<Shadow>> = {
...
-  offsetX?: number;
-  offsetY?: number;
-  affectStroke?: boolean;
-  nonScaling?: boolean;
+  color: string;
+  blur: number;
+  offsetX: number;
+  offsetY: number;
+  affectStroke: boolean;
+  nonScaling: boolean;
```
asturur, line 28: Those needs to be optional because toObject needs a return type, and it doesn't return the full object every time

src/util/misc/pick.ts
```diff
@@ -15,3 +15,15 @@ export const pick = <T extends Record<string, any>>(
...
+export const pickBy = <T extends Record<string, any>>(
+  source: T,
+  predicate: <K extends keyof T>(value: T[K], key: K, collection: T) => boolean
+) => {
+  return (Object.keys(source) as (keyof T)[])
+    .filter((key) => predicate(source[key], key, source))
+    .reduce((o, key) => {
+      o[key] = source[key];
+      return o;
+    }, {} as Partial<T>);
```
asturur, line 28: why the double loop rather than ust usin the if.
```
if(predicate(source[key], key, source)){
  o[key] = source[key];
}
```

src/shapes/Image.ts
ShaMan123: That can be said regarding the rest of the enlivenables...
I don't mind but it reminds me of that open PR #8507 that touches that logic

src/shapes/Polyline.ts
ShaMan123: Good point. Same for objects in group

src/Shadow.ts
ShaMan123: All serialized props are defined as non optional and used with Partial
So I made this the same

ShaMan123: And in toObject is does return a partial in case of `includeDefaultValues===false`

src/util/misc/pick.ts
ShaMan123: true

ShaMan123:
> > * Rect, Ellipse need a `fromObject` declaration
> 
> In case declaration yes, code no.

Circle as well (it is coded right now)

src/shapes/Polyline.ts
```diff
@@ -86,8 +101,9 @@ export class Polyline extends FabricObject {
    *   top: 100
    * });
    */
-  constructor(points: XY[] = [], { left, top, ...options }: any = {}) {
+  constructor(points: XY[] = [], options: Props = {} as Props) {
     super({ points, ...options });
+    const { left, top } = options;
```
ShaMan123, line 106: This change is unwanted IMO
Why should we set position before bbox if we do it durring/after bbox calc?

ShaMan123 (commented):
I am done interfering with this PR

asturur:
ok i will fix the tests asap.

src/shapes/Polyline.ts
asturur: because types.

src/shapes/Text/StyledText.ts
asturur: maybe/probably but i would love to keep this as types only as much as possible.
Also this function in particular, differently from the toObject we modified with pickBy is render code.

asturur:
ok this is done and i m happy everything looks cleaner

ShaMan123:
Nice!
I will look if I have follow ups

ShaMan123:
IMO this can be beta2 or an official release
 After we decide how to export types

PR #8757, opened 2023-03-05
WIP Use the mouse or pointer offsetX/offsetY instead of clientX/clientY

asturur:
## Motivation

Code simplification.

## Description

Took a break from types to tackle this thing it was in my head since long.

Today for historical reasons we check for mouse coordinates in client space, and then we subtract the current position of the canvas recursively calculating our position inspecting various css rules.
This code is old, and probably was done in thise way because older browsers didn't suppor getBoundingClientRect.

Unfortunately touch events ( that are still supported ) do not offer this property and for those browsers we will need to do a similar calculation, possibly using non custom code but a couple of browser api.

It is also to be investigated the whole `calcOffset` code that as of now would be needed only for text area positioning.

It would be great to remove `calcOffset` from canvas domain and simply relegate it to a hiddenTextarea and editable text concern.

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -1144,9 +1144,6 @@ export class SelectableCanvas<
       }
     }
 
-    this.calcOffset();
-    pointer.x = pointer.x - this._offset.left;
-    pointer.y = pointer.y - this._offset.top;
     if (!ignoreVpt) {
       pointer = this.restorePointerVpt(pointer);
     }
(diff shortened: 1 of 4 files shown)


asturur:
And of course JSDOM added support for those only recently but always set to 0,0
This won't be a 20 minutes job as i hoped.

ShaMan123:
truth is we don't need to bother about jsdom because no events fire in node.
it is a matter of having a dedicated index for node.

asturur:
this PR will wait for playwright testing working, with real mouse events and real layout, and divisions of tests from browser and node.

Right now i don't even remember all the layout cases that this code is trying to account for, i think that there was also the case of a canvas scrolling inside a div.

PR #8759, opened 2023-03-06
ci(): disallow circular deps

ShaMan123:
## Motivation

Make build fail on circular deps

## Description

Disallowing circular deps sounds like a good CD decision
All tests are failing because there is a circular dep in src that is solved by #8758


ShaMan123 (commented):
ready
makes sense merging #8758 and then this PR so it passes all checks

asturur:
Yes this is important, especially if we keep the mjs or node version unbundled.

asturur (approved):

PR #8760, opened 2023-03-06
lint(): fix errors

ShaMan123:
## Description

A no brainer that makes lint pass


asturur:
i didn't know this was all that was left.

PR #8766, opened 2023-03-07
refactor(): vector utils into Vector class

ShaMan123:
## Motivation

Working on coords and found myself wanting/needing this
Not crucial for me but I think it might make things less confusing in terms of a Point that is actually a vector. True they are very close in concept and impl both in fabric and in math in general.
And it makes logic a bit more scoped IMO

## Description

consolidate all vector utils into a subclass of Point

## Changes

renamed `IPoint` to `XY` - this I want in anyways. `IPoint` is a bad name

### DIFF
--- a/src/Point.ts
+++ b/src/Point.ts
@@ -29,10 +29,10 @@ export class Point implements XY {
     }
   }
 
-  create(): this;
-  create(x: number, y: number): this;
-  create(point: XY): this;
-  create(arg0: number | XY = 0, y = 0): this {
+  protected create(): this;
+  protected create(x: number, y: number): this;
+  protected create(point: XY): this;
+  protected create(arg0: number | XY = 0, y = 0): this {
     // @ts-expect-error constructor is not recognized
     return new this.constructor(arg0, y);
   }
--- a/src/Point.ts
+++ b/src/Point.ts
@@ -34,7 +34,7 @@ export class Point implements IPoint {
    * @return {Point} new Point instance with added values
    */
   add(that: IPoint): Point {
-    return new Point(this.x + that.x, this.y + that.y);
+    return this.create(this.x + that.x, this.y + that.y);
   }
 
   /**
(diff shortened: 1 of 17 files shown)


ShaMan123 (commented):
ready

ShaMan123:
Though it isn't crucial to me I would like to point out that encapsulation is awesome and I have matrix on my target doing the same.
Class methods are convenient and easy to discover.

asturur:
```
renamed IPoint to XY - this I want in anyways. IPoint is a bad name
```
meeehhh Why IPoint is a bad name? it would be PointInterface.

asturur:
Rename can be done with a replace in the codebase i guess? if yes open a PR for the rename and do it.

asturur:
It make me total mad that the code does the same things but is unminifiable.
Not sure what you mean by encapsulation.
I don't think that typing vector and discover code by auto completition is something i want to give value to compared to bundle size.
Someone should open the doc and read what vector utils can do.

If i m not understanding correctly, what do you mean by encapsulation?

asturur:
also please do not refactor matrix code, there are like 2 matrix functions, we  are good.

src/Point.ts
```diff
@@ -28,23 +29,31 @@ export class Point implements IPoint {
     }
   }
 
+  protected create(): this;
+  protected create(x: number, y: number): this;
+  protected create(point: XY): this;
+  protected create(arg0: number | XY = 0, y = 0): this {
+    // @ts-expect-error constructor is not recognized
+    return new this.constructor(arg0, y);
+  }
```
asturur, line 38: Why do we want this?

ShaMan123: legacy, passing new Point(x,y) is legacy. You want it dead?

ShaMan123:
In encapsulating I mean having all vector logic reside under a single class. Initializing it is easy and you do not need to import helpers, all are part of the prototype.
Also it seems reasonable to differentiate between point and vector (e.g. `pointer.transform(t, true for vector and false for point)`, same for rotate and another I think)
I am for

PR #8767, opened 2023-03-07
Redo coords

ShaMan123:
## Motivation

coords have been longing for a rewrite.
Ever since I rewrote Group I knew that but was afraid of the daunting task.
aCoords, lineCoords and oCoords are badly named and badly calculated, don't really fit Group, or edge cases like `strokeUniform`.

## Description
The goal is to provide a single set of coords that represent the bounding box and a set of transformations that output matching bounding boxes. This will be done using change of basis.
In addition, we must take into account the fact that because of the infamous `strokeUniform` option we must do bbox calculation in the viewport. Once we do that we can transform the bbox anywhere, however the opposite is incorrect. `strokeUniform` is not linear.
As part of this effort, I will try to impl @asturur's suggestion: no need for generic width/height option for all shapes.

Previously, control positioning didn't respect flipping. We discussed this and decided to leave as is for now, leaving this as a possibility to be handled by `Control#positionHandler`.
~~I want to change the signature of `positionHandler` so it accepts a set of transformations that will make it easy to position controls in different circumstances (e.g. the default rotation matrix, a bbox that isn't rotated, a transformed bbox, a bbox respecting flipping etc.).
If we must not break the Control API I will add a calling method that will handle legacy behavior.~~
`positionHandler` can use any BBox to position controls.

POC has uncovered a lot of bugs in control handlers. ~~I hope to find a way to complete the first iteration without touching that.~~

Another issue I have noticed is the conditional usage of `qrDecompose` (e.g. `getTotalAngle` in case of `group`). That should be considered a bug from now on since it will produce major inconsistencies in transformations. `qrDecompose` is the baseline for transformations. Under this scope, flipping and rotation might be a mess. I need to look into that a bit more.

Another thought in this field which is drastic.... Transform options is causing us a lot of headache. Fact is HTMLCanvas2DContext has nothing similar, rather, it exposes methods such as `rotate`, `skew`, `scale`, `preMultiply` etc. IMO so should we. And move all calculation to use the raw transform matrix. ~~Can it be done? Should it? I have no idea and won't be trying it for the near future but I think the change of basis approach is the solution for transform options in general.~~

**EDITED**: I have exposed `ObjectTransformations` (`qrDecomposed` can be avoid in layout/bbox calculations, see next paragraph).

For the same reasons mentioned above control action handlers must receive a vpt pointer and bbox infos

Considering performance and `setCoords`:
Only objects with `strokeUniform` or `padding` must be calculated in the viewport. The rest can be transformed/sent to the viewport while keeping accuracy.
Coords are used to draw controls and to detect pointer hits. If we are handling an object that must be calculated in the viewport we have no other option and must calculate coords once the viewport changes. However, if the object respects the viewport we can save on calculations by skipping `setCoords` in this case and sending the pointer to the relative object plane. The same applies to nested objects under groups.

There are several bugs/issues this PR aims to fix.
- [x] flipping + controls (and under group)
- [x] stroke uniform and padding bbox 
- [x] isOnScreen under group
- [x] removing geometry signatures `absolute, calculate`
- [x] proofing geometry for 100% of cases
- [x] make coords respect the viewport (skewing, rotation)
- [x] complex transforms under group (https://codesandbox.io/s/fabric-vanillajs-sandbox-forked-3rykyj)

## Changes

- Object origin has been **REMOVED** in favor of BBox. I did this because object origin respected only the containing plane so it was worthless. However the concept is brilliant so I adapted it into BBox, allowing us the work with origins on any plane. 
BBox is basically a transform matrix that represents a plane. It accepts an origin point and transforms it into the point in the plane, making it possible to transform the context and work with origin etc. The motivation of BBox is double: we must compute coords and geometry in the viewport because of things like strokeUniform and padding and we don't want to continuously calculate and figure out hard non intuitive linear algebra stuff.
- ObjectGeometry was split into:
   - ObjectLayout - purely layout logic, not related to bbox/coords. There was a lot of confusion about this in the code. stroke/padding do not belong here. This is the breaking change of the PR. Positioning is not affected by stroke any longer. Think of strokeUniform - how can layout know the size of stroke beforehand? And it is IMO the expected behavior. If I change stroke I don't expect my object to move from its center.
   - ObjectBBox - bbox/coords logic, strokeUniform and padding go here, the bbox is created in `setCoords`
   - ObjectPosition - exposing getters/setters for x,y instead of ObjectOrigin (sme methods might be redundant)
   - ObjectTransformation - **new** class mimicking canvas transform API for objects because using `set` is not valid for the same reason ObjectOrigin isn't (and as a bonus it starts moving away from qrDecompose). Exposed `scale`, `skew`, `rotate`, `scaleBy`, `skewBy`, `rotateBy`, `translate` and more. It will replace some control logic.
   - ObjectGeometry - pure geometry methods, all use bbox now, so most of them are aliases
- I want to drop decomposition in favor of holding the own matrix and exposing setters and getters for the transition or forever. As I see it this is the reason the canvas API doesn't decompose, it exposes a lot of friction. But things like limiting scale etc will probably need decomposition anyways. 
- This PR should be considered part of the Group rewrite. I hacked controls by sending the pointer to the containing plane but it is wrong to do that (strokeUniform, vpt for example are not accounted for properly). I have decided to make controls work in the viewport because by definition they interact with the event which exists in the viewport but it is a design call. We need to discuss it. Because with BBox we could accurately send the points to the containing plane and set apply changes to object transformation in the vpt.
- BBox will use stroke projection after we agree on how to move forward

- Misc
  - This PR has made me aware that rendering doesn't respect strokeUniform as well. Skewing an object or the vpt will distort the stroke and shouldn't. The stroke should be rendered in a post effect hook IMO
  - We can expose a util or option in `fromObject` that consumes v5 data and outputs v6 data, since the version is present as part of the data

I am not touching tests yet because I don't want to waste time. First we need to decide and discuss.

https://github.com/fabricjs/fabric.js/pull/8767#pullrequestreview-1342025596

## Gist

Change of basis

## In Action

https://user-images.githubusercontent.com/34343793/225823225-089b6fd0-1e7f-4881-9969-0b6f203a6993.mp4

This is a case that is broken on master:
https://codesandbox.io/s/fabric-vanillajs-sandbox-forked-3rykyj?file=/package.json

### 3D rotation

I have accidently managed to make the skewing control a 3d rotation 🤯😆

https://user-images.githubusercontent.com/34343793/225829816-eb92a0b3-2f67-4085-814f-53b46ef1e902.mp4

#### `activeObject.rotate3D(1, -1, 1, true)`

https://user-images.githubusercontent.com/34343793/225849059-d6a16688-5cfd-4dec-b6be-f787e00f3252.mp4

#### 3D controls

https://user-images.githubusercontent.com/34343793/225880489-1d00df09-e581-437b-ac2f-daf366a72743.mp4


ShaMan123:
@luizzappa I am rewriting coords in order to fix a few core issues and want to use stroke projection for calculations.
I need to repurpose how stroke projection is used. It will get the final transformed coords as they are visible to the user.
Can you give me a hand?

luizzappa:
Hey, @ShaMan123! There's a lot here!

It is not yet clear to me how this change will address these bugs. But speaking specifically of what you want to implement.

> The goal is to provide a single set of coords that represent the bounding box and a set of transformations that output matching bounding boxes. This will be done using change of basis.
> In addition, we must take into account the fact that because of the infamous `strokeUniform` option we must do bbox calculation in the viewport. Once we do that we can transform the bbox anywhere, however the opposite is incorrect. `strokeUniform` is not linear.

Regarding `polygons` and `polylines`, the `projectStrokeOnPoints` function returns the points on the object plane, which are used by `_calcDimensions` function to calculate the BBOX. It would be enough to take these points and send them to any other plane with the corresponding transformation matrix. Why would `storkeUniform` interfere with that?

> no need for generic width/height option for all shapes.

For `poly` it doesn't make much sense to have a generic `width`/`height`, because these dimensions end up being defined by the points we pass when creating them.. But the way `Rect` and `Triangles` currently exist, the `width`/`height` end up defining  the points of the geometric figure. For example, when creating a `Rect` with `width` = `height` = 100, in the background I am setting the following original points = [ { x:-50, y: -50}, { x: 50, y: -50}, { x : 50, y:
 , 50}, { x:-50, y: 50}]. 
In short, what we currently have: the polly points define the generic width/height, and in the case of Rect/Triangle it is the opposite (the generic width/height defines the underlying points).
Does it make sense for the Rect and Triangle classes as they currently are? Maybe one way is for them to inherit from polygon

luizzappa:
**qrDecompose** -> since that PR (decompose transformation matrix) I opened, I delved more on the subject. In my opinion it's more of a design decision, regardless of which way to go. I'll come back to that PR later for more details.

luizzappa:
It seems to me that this calculation change will enable a new feature: rotating the canvas (vpt) without messing with the controls.

ShaMan123:
> It seems to me that this calculation change will enable a new feature: rotating the canvas (vpt) without messing with the controls.

yep. I will allow more even. I am playing around with controls and see more possibilities. 
If I manage to remove the need of transform options for control it might be possible to remove the need from fabric entirely and then `qrDecompose` will become obsolete

What you wrote about stroke projection I remember and understand.
seems to me that `strokeUniform` projections are inaccurate this way since the viewport/group will distort this calculation. Truth is rendering in this case is wrong as well.
image

I tried using stroke projection round to define the bbox coords for circle w/o success

ShaMan123:
In other words I want to change stroke projection to use the viewport coords but don't know how to with skewing

luizzappa:
Create a codesand box with this problem and I'll take a look. 

I cloned this PR, but I don't understand how the build is currently working, it doesn't generate a fabric.js bundle anymore? It seems to be generating an index.js bundle that no longer exports the fabric namespace.

ShaMan123:
I tried something like 
```ts
// Circle.ts
  calcCoords(): TCornerPoint {
    return cornersFromPoints(
      [
        new Point(0, -0.5),
        new Point(0.5, 0),
        new Point(0, 0.5),
        new Point(-0.5, 0),
      ].map((origin) =>
        this.calcDimensionsVector(origin, {
          applyViewportTransform: true,
        }).add(this.getCenterPoint())
      )
    );
    return cornersFromPoints(
      projectStrokeOnPoints(
        [
          new Point(0, -0.5),
          new Point(0.5, 0),
          new Point(0, 0.5),
          new Point(-0.5, 0),
        ].map((origin) =>
          this.calcDimensionsVector(origin, {
            applyViewportTransform: true,
          }).add(this.getCenterPoint())
        ),
        {
          ...this,
          ...qrDecompose(
            multiplyTransformMatrices(
              this.getViewportTransform(),
              this.calcTransformMatrix()
            )
          ),
        }
      ).map((projection) => projection.projectedPoint)
    );
  }

```

luizzappa:
> ```ts
>   projectStrokeOnPoints(
>         [
>           new Point(0, -0.5),
>           new Point(0.5, 0),
>           new Point(0, 0.5),
>           new Point(-0.5, 0),
>         ].map((origin) =>
> ```

Here you are projecting a square. In the case of the circle I believe you could take advantage of the `projectStrokeOnPoints` passing a single `point [{ x: 0, y: 0 }]` and in options `strokeLineCap = "round"`, this way it will compute for the case of a circle.

luizzappa:
With skew must not work because its fix hasn't been merged yet: https://github.com/fabricjs/fabric.js/pull/8374

src/shapes/Object/BBox.ts
```diff
@@ -0,0 +1,313 @@
...
+      tr: this.pointFromOrigin(new Point(0.5, -0.5)),
+      br: this.pointFromOrigin(new Point(0.5, 0.5)),
+      bl: this.pointFromOrigin(new Point(-0.5, 0.5)),
+    };
+  }
+
+  getCoords() {
+    const { tl, tr, br, bl } = this.getCoordMap();
+    return [tl, tr, br, bl];
+  }
```
asturur, line 70: just a note or i forget. Is this `Object.values(this.getCoordMap())`?

src/Collection.ts
```diff
@@ -325,11 +325,11 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
...
           object.visible &&
-          ((includeIntersecting && object.intersectsWithRect(tl, br, true)) ||
-            object.isContainedWithinRect(tl, br, true) ||
+          (object.isContainedWithinRect(tl, br, true) ||
             (includeIntersecting &&
-              object.containsPoint(tl, undefined, true)) ||
-            (includeIntersecting && object.containsPoint(br, undefined, true)))
+              (object.intersectsWithRect(tl, br, true) ||
+                object.containsPoint(tl, true) ||
+                object.containsPoint(br, true))))
```
asturur, line 332: this seems a change that is unrelated. is it?

src/canvas/SelectableCanvas.ts
```diff
@@ -1052,11 +1051,8 @@ export class SelectableCanvas<
     // until we call this function specifically to search inside the activeGroup
     while (i--) {
       const objToCheck = objects[i];
-      const pointerToUse = objToCheck.group
-        ? this._normalizePointer(objToCheck.group, pointer)
-        : pointer;
-      if (this._checkTarget(pointerToUse, objToCheck, pointer)) {
-        target = objects[i];
+      if (this._checkTarget(objToCheck, pointer)) {
+        target = objToCheck;
```
asturur, line 1055: ok this is cool, i wanted to do it since long.

asturur: 1 over 32 files i have nothing to say

src/util/misc/matrix.ts
```diff
@@ -112,16 +132,35 @@ export const qrDecompose = (a: TMat2D): TQrDecomposeOut => {
...
+}: {
+  skewX?: TDegree;
+  skewY?: TDegree;
+  shearX?: number;
+  shearY?: number;
+}) =>
+  multiplyTransformMatrices(
+    [1, 0, shearX ?? (skewX ? Math.tan(degreesToRadians(skewX)) : 0), 1, 0, 0],
+    [1, shearY ?? (skewY ? Math.tan(degreesToRadians(skewY)) : 0), 0, 1, 0, 0]
+  );
```
asturur, line 162: unused? also what is the difference between skew and shear ( in english or in geometry ) ?
I thought they were the same thing

src/util/misc/matrix.ts
```diff
@@ -82,6 +84,24 @@ export const multiplyTransformMatrices = (
...
+ * @returns
+ */
+export const multiplyTransformMatrixChain = (
+  matrices: TMat2D[],
+  is2x2?: boolean
+): TMat2D =>
+  matrices.reduceRight(
+    (product, curr) => multiplyTransformMatrices(curr, product, is2x2),
+    iMatrix
+  );
```
asturur, line 103: unused?

src/util/misc/matrix.ts
ShaMan123: shear is `tan(skew)`

src/canvas/SelectableCanvas.ts
ShaMan123: > 1 over 32 files i have nothing to say

Don't understand

src/shapes/Object/BBox.ts
ShaMan123: yes
I should redo that

ShaMan123 (commented):
I am far from reviewing this PR

asturur:
I started reading and before you try to finish i need to block you.
Reduce the scope.
Is there need to touch oCoords and controls? they are separated from geometry right now, they are already dynicamically calculated.
Is there any need to do immediately the boudning box class? we can validate the approach using the sendToPlane functions.

A pr so large will stay here so long and just be a mess for conflicts.

I don't believe there aren't progressive steps that can have us going in this direction but one step at time.
Can we take the changes you made to vptCoords and have just those changes and start measuring the impact?

vptCoord changes seems ot me:
### vptCoords
vptCoords do not exist anymore.
isOnScreen now instead of comparing aCoords with vptCoords, compares the coordinates in canvas space with 0,0,width,height.
in order to obtain the coords in viewport space we do the following:
... ( i have no idea because what happens in bbox classes and planeChangeMatrix is not that much better than what we have today at the point of not requiring explanations )

ShaMan123:
This PR has taken a turn ad is almost stable
Very exciting
When I review it I will update the description and list the fixed bugs. Which are numerous.
Just for fun, all viewport transformations are now possible

asturur:
> This PR has taken a turn ad is almost stable Very exciting When I review it I will update the description and list the fixed bugs. Which are numerous. A spoiler: all viewport transformations are now possible

We still needs to break into pieces somehow.
The scope of change is too large to accepted.

ShaMan123:
Once I have it stable I will think how to split.
That is part of why I want tochat

ShaMan123 (commented):
- Object origin has been **REMOVED** in favor of BBox. I did this because object origin respected only the containing plane so it was worthless. However the concept is brilliant so I adapted it into BBox, allowing us the work with origins on any plane. BBox is basically a transform matrix that represents a plane. It accepts an origin point and transforms it into the point in the plane, making it possible to transform the context and work with origin etc. The motivation of BBox is that we must compute coords and geometry in the viewport because of things like strokeUniform and padding.
- ObjectGeometry was split into:
   - ObjectLayout - purely layout logic, not related to bbox/coords. There was a lot of confusion about this in the code. stroke/padding do not belong here. This is the breaking change of the PR. Positioning is not affected by stroke any longer. Think of strokeUniform - how can layout know the size of stroke beforehand? And it is IMO the expected behavior. If I change stroke I don't expected my object to move from its center.
   - ObjectBBox - bbox/coords logic, strokeUniform and padding go herer
   - ObjectPosition - exposing getters/setters for x,y instead of ObjectOrigin
   - ObjectTransformation - **new** class in mimicking canvas transform API for objects because using `set` is not valid for the same reason ObjectOrigin isn't. (and as a bonus it starts moving away from qrDecompose). Exposed `scale`, `skew`, `rotate`, `scaleBy`, `skewBy`, `rotateBy` and `translate` and more. It will replace some control logic.
   - ObjectGeometry - pure geometry methods, all use bbox now, so most of them are aliases
- I want to drop decomposition in favor of holding the own matrix and exposing setters and getters for the transition or forever. As I see it this is the reason the canvas API doesn't decompose, it exposes a lot of friction. But things like limiting scale etc will probably need decomposition anyways. 
- This PR should be considered part of the Group rewrite. I hacked controls by sending the pointer to the containing plane but it is wrong to do that (strokeUniform, vpt for example are not accounted for properly). I have decided to make controls work in the viewport but that is what they are but it is a design call. We need to discuss it. Because with BBox we could accurately send the points to the containing plane and set the transformation in the vpt.
- BBox will use stroke projection after we agree on how to move forward


- Misc
  - This PR has made me aware that rendering doesn't respect strokeUniform as well. Skewing an object or the vpt will distort the stroke and shouldn't. The stroke should be rendered in a post effect hook IMO




I am not touching tests yet because I don't want to waste time. First we need to decide and discuss.
- Are you ok with stroke moving away for layout? It is a must. We can expose a util or option in `fromObject` that consumes v5 data and outputs v6 data, since the version is present as part of the data



I know this is a huge PR. I want to think with you how it is best to move forward because it fixes most of the known transform/bbox/coord issues

ShaMan123:
updated description

src/shapes/Group.ts
```diff
@@ -682,13 +679,13 @@ export class Group extends createCollectionMixin(FabricObject<GroupEvents>) {
       return this.prepareBoundingBox(layoutDirective, objects, context);
     } else if (layoutDirective === 'clip-path' && this.clipPath) {
       const clipPath = this.clipPath;
-      const clipPathSizeAfter = clipPath._getTransformedDimensions();
+      const clipPathSizeAfter = clipPath.getDimensionsVectorForLayout();
```
ShaMan123, line 682: not sure

ShaMan123 (commented):
controls almost stable, resize width + skew remain

src/controls/controlRendering.ts
```diff
@@ -123,10 +122,9 @@ export function renderSquareControl(
     styleOverride.cornerStrokeColor || fabricObject.cornerStrokeColor || '';
   // this is still wrong
   ctx.lineWidth = 1;
-  ctx.translate(left, top);
-  //  angle is relative to canvas plane
-  const angle = fabricObject.getTotalAngle();
-  ctx.rotate(degreesToRadians(angle));
+  ctx.translate(x, y);
+  //  angle is relative to viewport
+  ctx.rotate(fabricObject.getTotalAngle());
```
ShaMan123, line 127: should be handled differently

src/controls/skew.ts
ShaMan123, line 1: skew is different than before
I can try to make it as it was

src/shapes/Group.ts
ShaMan123, line 1: Group has regressions to fix

src/shapes/Object/ObjectBBox.ts
```diff
@@ -0,0 +1,228 @@
...
+    //   {
+    //     ...this,
+    //     ...qrDecompose(
+    //       multiplyTransformMatrices(
+    //         this.needsViewportCoords() ? this.getViewportTransform() : iMatrix,
+    //         this.calcTransformMatrix()
+    //       )
+    //     ),
+    //   }
+    // );
```
ShaMan123, line 161: @luizzappa look here

ShaMan123 (commented):
Starting to stabilize
@asturur I think you can start reviewing, ping me if you need
I suggest starting as follows:
- BBox classes:
   - PlaneBBox
   - ViewportBBox
   - BBox
   - **skip** OwnBBox
- Object classes
  - ObjectLayot
  - ObjectBBox
  - ObjectGeometry
  - Object
  - InteractiveObject
  - ObjectTransformations (or after ObjectBBox to read in inheritance order)
- shapes
- canvas
- controls

Test files aren't changed apart from unverified minors and skipping so I suggest not to go there at all


@luizzappa I think the PR is stable enough for stroke projection
The concept is that all objects expose their viewport coords.
Therefore, I want us to modify stroke projection to accept viewport coords.
The polys and ovals can override `calcCoords` and `needViewportCoords` and that should be it. Then we could remove the workarounds we did in polys and remove `setDimensions`, `width`/`height` props etc.
I suggest you fork this and PR back to this branch.

Also if you guys understand what is up with group I am all ears

src/shapes/Group.ts
ShaMan123: done!

src/shapes/Group.ts
```diff
@@ -503,20 +523,22 @@ export class Group extends createCollectionMixin(FabricObject<GroupEvents>) {
...
         ctx.transform(...invertTransform(this.calcTransformMatrix()));
-        this._objects[i].render(ctx);
+        object.render(ctx);
         ctx.restore();
-      } else if (this._objects[i].group === this) {
-        this._objects[i].render(ctx);
+      } else if (
+        object.group === this &&
+        (preserve || !this._activeObjects.includes(object))
+        // && (!object.skipOffscreen || object.isOverlapping(this))
```
ShaMan123, line 537: a design decision

ShaMan123 (commented):
I believe this PR is now stable and ready for the next step

ShaMan123:
IMO we should make ownMatrix what we store instead of the decomposed values and expose getters/setters for them

src/shapes/Group.ts
ShaMan123: initial layout + rotation

luizzappa:
> The concept is that all objects expose their viewport coords.
> Therefore, I want us to modify stroke projection to accept those coords.

@ShaMan123  ,  did not quite understand. Your expectation is instead of passing the points (`this.points`) of the polygons, passing viewport coords (bbox in vpt plane) to the `projectStrokeOnPoints` function? If so, you won't get the result you expect. Points are necessary because they define the geometric figure, BBOX is just a square.

Something like this could be done within `Polyline.ts` to take advantage of `ObjectBBox.calcCoord`:

```
  calcCoords() {
    const pointsInObjectPlane = this._projectStrokeOnPoints().map(
        (proj) => proj.projectedPoint
      ),
      bboxInObjectPlane = makeBoundingBoxFromPoints(pointsInObjectPlane), // <--
      factor = new Point(bboxInObjectPlane.width, bboxInObjectPlane.height),
      offset = new Point(
        -(bboxInObjectPlane.left + bboxInObjectPlane.width / 2),
        -(bboxInObjectPlane.top + bboxInObjectPlane.height / 2)
      );
    // That top part would need a cache

    return mapValues(
      {
        tl: new Point(bboxInObjectPlane.left, bboxInObjectPlane.top),
        tr: new Point(
          bboxInObjectPlane.left + bboxInObjectPlane.width,
          bboxInObjectPlane.top
        ),
        bl: new Point(
          bboxInObjectPlane.left,
          bboxInObjectPlane.top + bboxInObjectPlane.height
        ),
        br: new Point(
          bboxInObjectPlane.left + bboxInObjectPlane.width,
          bboxInObjectPlane.top + bboxInObjectPlane.height
        )
      },
      (origin) =>
        this.calcCoord(origin.divide(factor), {
          offset: offset.multiply(new Point(this.scaleX, this.scaleY))
        })
    );
  }
```

Doing a test works fine, still need to adjust the offset when VPT has some kind of transformation. If that's the way to go, I'll proceed with that tweak.

image

ShaMan123:
> Your expectation is instead of passing the points (`this.points`) of the polygons, passing viewport coords (bbox in vpt plane) to the `projectStrokeOnPoints` function? If so, you won't get the result you expect. Points are necessary because they define the geometric figure, BBOX is just a square.

No. The poly or any other shape is responsible of calculating its coordinates, how many they are, in the viewport. This has nothing to do with the bounding box. Polys will use stroke projection to do that. 
BBox receives the calculated viewport coords and is in charge of producing a transformation that is used to define the shape's bounding box (hence, there are several options: `rotated`, `transformed`). The transformation can be used to position controls, draw borders etc.
True that currently BBox receives 4 corners of a rect. I think it should accept n points and generate a bbox from that. This is crucial for the `rotated` bbox. So adjusting that part should be done (`BBox.getViewportCoords` should return an array of points I suppose).
So seems stroke projection doesn't need a lot of adjusting. I think it doesn't take into account the viewport at all, therefore stroke uniform is wrong. Additionally, if there is a way to apply the transformation matrices without decomposition that would be great. I am aiming for that, see `src/shapes/Object/ObjectTransformations`.
This way any shape can define its coords, expose them in the viewport and let the BBox do the rest.

ShaMan123:
If you think differently do tell. It might be that we must expose the 4 corners and that is it. I looked at `BBox.transformed` for example and am not sure how to do it with an array of n coords

asturur:
@ShaMan123 this PR balooned out of coords a lot.
I will check the bounding box part and start with that, but at this level of changes is unusable for me.

ShaMan123:
It is huge
We can discuss how to split it into pieces

asturur:
i m trying to extract all the utils changes and merge those, since those will be useful anyway and will make additional changes easier.

ShaMan123:
Ported to https://github.com/ShaMan123/fabric.js/pull/4
Contributions are welcome

nounder:
@ShaMan123 @asturur  Is there any reason why this PR wasn't merged and was ported to a fork instead? This piece of work is rad.

ShaMan123:
Disagreements
I left fabric

PR #8315, opened 2022-09-25
ci(): bug report form

ShaMan123:
### First merge #8189 

- Added a github action to have the user check some boxes before opening an issue
- Added a github action to automatically update the list of available versions for the form to choose from

https://github.com/fabricjs/fabric.js/compare/readme...bug-report-form

PR #8770, opened 2023-03-09
perf(): optimize `perPixelTargetFind`

ShaMan123:
## Motivation

Making perPixelTargetFind faster

## Description

limited the canvas size 
view the diff w/o ws

## Changes

renamed `cacheCanvasEl` => `pixelFindCanvasEl` because it is no longer a generic cache canvas

## Gist

I don't really know how to benchmark this.

Before

image

After 

image


ShaMan123 (commented):
Seems ready
Could you help me benchmark the change?

jiayihu:
@ShaMan123 I'll gladly have a look personally in the week-end. Do you have any documentation about how to setup the fabric repo locally? What are the steps to open the project that you have used in your screenshots in localhost?

Regarding how to analyse the performance, you typically need to record actions that trigger the method a lot `isTransparent` in this case, e.g. several transparent objects all above each other and then move the mouse above them. 
Then you need to look for the method in the flame graph and see how it behaves. The most basic evaluation is just observing the time and compare it to a new recording with the latest PR changes.

Once you select a method in the flame graph, you can also click on the bottom to jump to its file and see the timings of each line. There you can also have a quick evaluation of where most of the time is spent and see if there are significant improvements. This is an example of what I did last week, where I replaced the dynamically created styles declaration with a static one which is significantly more performant for V8 because it has a static shape (we can talk about stuff like this that I noticed in fabric in another time)

## Before

Screenshot 2023-03-06 at 12 06 07

## After

Screenshot 2023-03-06 at 12 34 00

ShaMan123:
I use the net app from #8135 
- clone `.codesandbox/templates/next` 
- install deps
- link fabric to your local repo or install this branch

OR you can squash #8315 into this locally and run `npm run sandbox start next`

danielcrisp:
Interesting! Will take a look later. Thanks @ShaMan123

src/canvas/SelectableCanvas.ts
```diff
@@ -667,10 +667,23 @@ export class SelectableCanvas<
...
+      // TODO: handle in set
+      this.pixelFindCanvasEl.width = this.pixelFindCanvasEl.height = size;
+    } else {
+      this.clearContext(ctx);
+    }
     // in case the target is the activeObject, we cannot execute this optimization
     // because we need to draw controls too.
-    if (isFabricObjectCached(target) && target !== this._activeObject) {
-      // optimizatio: we can reuse the cache
+    if (
```
asturur, line 681: it happened that i thought a couple of times of this:
```
// in case the target is the activeObject, we cannot execute this optimization
    // because we need to draw controls too.
```
And then i asked myself, why do i have to draw the controls?
The controls can be found with the coordinates.

src/canvas/SelectableCanvas.ts
```diff
@@ -667,10 +667,23 @@ export class SelectableCanvas<
    * @return {Boolean}
    */
   isTargetTransparent(target: FabricObject, x: number, y: number): boolean {
+    const ctx = this.contextCache;
+    const retina = this.getRetinaScaling();
+    const size = Math.max(2 * this.targetFindTolerance * retina, 10);
```
asturur, line 672: Why would you max it out?

asturur: i would rather have a method `setTargetFindTolerance` that creates this mini canvas, i m not even sure retina scaling is a requirement since the mouse pixels ( the so called css pixels ) are always without retina.

ShaMan123: > i would rather have a method `setTargetFindTolerance` that creates this mini canvas,

will do

I wasn't sure about retina. What will happen in case of ay scaled and blurry image? Will the hit test return a false positive?

src/canvas/SelectableCanvas.ts
ShaMan123: absolutely
I had no idea what was that.
And I think active selection never caches, or at least it didn't in v5. Now I don't remember.

ShaMan123:
Truth be told IMO this method doesn't belong to canvas. It is a standalone util. The canvas element can be a global const and we could even make it offscreen I believe.

src/canvas/SelectableCanvas.ts
ShaMan123: I thought I removed the max value of 10

asturur:
> Truth be told IMO this method doesn't belong to canvas. It is a standalone util. The canvas element can be a const and we could even make it offscreen I believe.

Maybe it doesn't belong, but if this is an optimization, the rename of the canvas is already a stretch, 

Yes could be offscreen but that would make target detection async, no?
Per pixel target selection is expensive anyway and make sense only in certain cases,

src/canvas/SelectableCanvas.ts
```diff
@@ -1237,8 +1255,13 @@ export class SelectableCanvas<
...
   protected _createCacheCanvas() {
-    this.cacheCanvasEl = this._createCanvasElement();
-    this.contextCache = this.cacheCanvasEl.getContext('2d')!;
+    this.pixelFindCanvasEl = this._createCanvasElement();
+    this.pixelFindContext = this.pixelFindCanvasEl.getContext('2d', {
+      willReadFrequently: true,
+    })!;
+    const retina = this.getRetinaScaling();
+    this.pixelFindContext.scale(retina, retina);
+    this.setTargetFindTolerance(this.targetFindTolerance);
```
asturur, line 1264: setting width/height with this methof after setting the scale will kill the transform? or kill just the pixels? i don't remember

asturur:
> @ShaMan123 I'll gladly have a look personally in the week-end. Do you have any documentation about how to setup the fabric repo locally? What are the steps to open the project that you have used in your screenshots in localhost?
> 
> Regarding how to analyse the performance, you typically need to record actions that trigger the method a lot `isTransparent` in this case, e.g. several transparent objects all above each other and then move the mouse above them. Then you need to look for the method in the flame graph and see how it behaves. The most basic evaluation is just observing the time and compare it to a new recording with the latest PR changes.
> 
> Once you select a method in the flame graph, you can also click on the bottom to jump to its file and see the timings of each line. There you can also have a quick evaluation of where most of the time is spent and see if there are significant improvements. This is an example of what I did last week, where I replaced the dynamically created styles declaration with a static one which is significantly more performant for V8 because it has a static shape (we can talk about stuff like this that I noticed in fabric in another time)
> 

AND

> I don't really know how to benchmark this.

Is true we can get as deep as we want with perf measurements and edge cases and find an improvement or not depending what we do.
this PR has the advantage of swapping a large canvas as big as the application canvas with one that is around 10-20px large.
There are 2 ways in which you can benchmark it.
You can add 2 performance timings at the start and the end of findTarget function ( yes wrapping all the process and not just this ) make a counter of how many time the function run over a bunch of objects and sum up the time and the compare the 2 cases.
Then if you enable the memory in the profiling you can check if in the same case total memory consumption is lower.

A not on the `willReadFrequently`. In my opinion that property will make getImageData faster, but will make drawImage slower. Since we do one and one for each pixelfind i m not sure is an exact choice.

src/canvas/SelectableCanvas.ts
```diff
@@ -680,31 +684,46 @@ export class SelectableCanvas<
           target.cacheTranslationY + normalizedPointer.y * target.zoomY,
           0
         );
+      // transform tolerance according to vpt
+      // TODO: use sendVectorToPlane
+      const tolerance = new Point()
+        .scalarAdd(this.targetFindTolerance)
+        .transform(invertTransform(this.viewportTransform), true);
+      const size = tolerance.scalarMultiply(2).max(new Point(1, 1));
```
asturur, line 692: i need to think about this. this put ourselves in the case of transforming the tolerance with the vpt when what we want is really just canvas pixels. So if there is a small text or a thin line and the zoom level makes it super thin, we were not composensating for this, if the line was so thinned by the zoom out that was nearly transparent, we would consider it transparent before.

src/canvas/SelectableCanvas.ts
```diff
@@ -680,31 +684,46 @@ export class SelectableCanvas<
...
       );
     }
 
-    const ctx = this.contextCache,
-      originalColor = target.selectionBackgroundColor,
-      v = this.viewportTransform;
-
-    target.selectionBackgroundColor = '';
-
-    this.clearContext(ctx);
```
asturur, line 698: who is clearing the context now between different checks? and where the `selectionBackgroundColor` handling code went?

src/canvas/SelectableCanvas.ts
asturur: @shaman can you share the app code used to test this so i can reproduce the testing case? i m curious about trying a couple of things.

src/canvas/SelectableCanvas.ts
ShaMan123: excellent question...
I will reorder to be safe
I have no idea

src/canvas/SelectableCanvas.ts
ShaMan123: oops.

src/canvas/SelectableCanvas.ts
ShaMan123: The method accepts a global pointer, living in the viewport, so 1 is the what is perceived as 1 by the user, not by the canvas.

asturur: yes 1 viewport pixel. 1 of what are you looking at in that moment and also 1 of how much you can move your mouse.
The need to use the max between scalarMultiply and 1, and the invert of the viewportTransform is something you added, i think it change the meaning of what tolerance does.

ShaMan123: I think I fixed it because previously it was checking tolerance in the canvas plane and not in the viewport. I will add a dedicated test.

ShaMan123: The app... I will share a link?

src/canvas/SelectableCanvas.ts
ShaMan123: I will restore selection background... 
The context must be scaled because we set dimensions of canvas as well and never re init retina, right?

asturur: not sure. i m building a little demo based on this concept so we can talk with examples.
A fun diversion from types/tests

ShaMan123: > A fun diversion from types/tests

cool

Since you merged the sandbox pr just run
`npm run sandbox start -- -t next [path/to/folder or nothing]`
if you provide a path the app will build there, if not it will build on the template. Which is fine as well.

ShaMan123: This is the deployed version
https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-k53evn?welcome=true&file=%2FREADME.md

Can you look at the sandbox github app and see it can operate in the repo? It should post links for us to reproduce with ease using the templates I built that are now merged into master

ShaMan123: The app I posted is nothing really, just objects on canvas using the next template

ShaMan123:
@asturur you were right about the tolerance, it is in canvas element space, so not affected by transformations.
IMO we should drop the so called optimization using the cache because `render` is optimized to use the cache and it is a head to make correct and extremely hard

ShaMan123:
You are also right about setting size resets transforms

src/canvas/SelectableCanvas.ts
```diff
@@ -680,31 +693,46 @@ export class SelectableCanvas<
...
+        target._cacheCanvas,
+        Math.floor(targetRelativeX - tolerance.x),
+        Math.floor(targetRelativeY - tolerance.y),
+        Math.ceil(size.x),
+        Math.ceil(size.y),
+        0,
+        0,
+        Math.ceil(size.x * retina),
+        Math.ceil(size.y * retina)
+      );
```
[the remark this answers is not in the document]
ShaMan123: doing this is redundant in terms of having this block of code, better off to use render

src/canvas/SelectableCanvas.ts
asturur: i made mine, i had this time to work on it and i just played a bit
image

I made this canvas where i can see what is in the pixelFind canvas in realtime, as soon as i m done highlighting the difference i can push up the branch

ShaMan123:
also what happens when the target is under a group. The optimization won't work, am I right? cause we need to transform the context.
Removing, you can revert if you think differently

src/canvas/SelectableCanvas.ts
```diff
@@ -680,31 +686,49 @@ export class SelectableCanvas<
...
-    this.clearContext(ctx);
-
     ctx.save();
-    ctx.transform(v[0], v[1], v[2], v[3], v[4], v[5]);
+    ctx.translate(-x + this.targetFindTolerance, -y + this.targetFindTolerance);
+    ctx.transform(...this.viewportTransform);
+    const selectionBgc = target.selectionBackgroundColor;
+    target.selectionBackgroundColor = '';
     target.render(ctx);
+    target.selectionBackgroundColor = selectionBgc;
```
asturur, line 724: to be honest i think `selectionBackgroundColor` is just a terrible feature and that it could be deleted, but we shouldn't do it here.
Is a property that pertain the control bounding box and was added to the objects because rush/inexperience

ShaMan123: I met it redoing controls. Managed to get it working there but I agree it is weird because it seems not part of object rendering scope, same as controls, but because it needs to be behind and object stacking I understand why you did it.

src/util/misc/isTransparent.ts
```diff
@@ -13,39 +15,26 @@ export const isTransparent = (
...
+  const start = point.scalarSubtract(tolerance).floor();
+  const end = point
+    .scalarAdd(Math.max(tolerance, 1))
+    .ceil()
+    .min(new Point(ctx.canvas.width - 1, ctx.canvas.height - 1));
+  const boundStart = start.max(new Point());
+  const size = end.subtract(boundStart);
+  if (size.x <= 0 || size.y <= 0) {
+    // out of bounds
+    return true;
```
asturur, line 29: i notice now this change, i don't think we should do it. the old loop is so much clear and smaller.
There is something weird in the old code and is that if we set x or y to 0 we do not adjust the tolerance to avoid considering extra pixels, but this change here is definitely extra complication.

src/canvas/SelectableCanvas.ts
asturur: i will look into this sandbox as soon as i can.
The goal is that for each PR there is a sandbox with the branch built?

asturur:
> @asturur you were right about the tolerance, it is in canvas element space, so not affected by transformations. IMO we should drop the so called optimization using the cache because `render` is optimized to use the cache and it is a head to make correct and extremely hard

I think the cache canvas optimization ( reusing the cache ) is still ok and usable for top level obects at least.
render will reuse the cache correct, so is probalby doing the exact same thing?
Right now my code has drifet apart i have a fork ( it drifted becuase i was trying to recover this KB added ) but let's focus on the fact that this PR is introducing the optiization of the small canvas, if we change too much it means we got confused along the road. It shouldn't be hard to just make the operational canvas small.

asturur:
It would be also ok to remove the cacheCanvas optimization if we can see that for already cached object the change in performance is minimal ( maybe is 0 ) since with interactive groups lot of things could be different.
But there are also other considerations to do here and there. I have to stop now, but i was enjoying this exploration so i ll do some more later.

asturur:
my code is here `fork-pixel-detection` and is probably broken right now because i left in half

src/util/misc/isTransparent.ts
ShaMan123: I will try to make simple but seems to me there are edge cases in the logic.
I can't understand the logic

ShaMan123: please try to explain it

src/canvas/SelectableCanvas.ts
ShaMan123: correct

asturur:
Let's try tomorrow to finish this.
I ll show what i have and i ll see if i can pull a benchmark.
The only reason why i want to pull a benchmark out is because i m curious about the effects of `willReadFrequently` when we want to paint over the `willReadFrequently` canvas some canvases that are in the GPU.

asturur:
Collecting some data from the fork at: `fork-pixel-detection`

cached object
image

normal active selection uncached
image

image

Now i ll compare to master asap just for curiosity and compare the data extraction logic

src/canvas/SelectableCanvas.ts
```diff
@@ -667,44 +672,22 @@ export class SelectableCanvas<
...
     ctx.restore();
-
-    target.selectionBackgroundColor = originalColor;
-
-    return isTransparent(ctx, x, y, this.targetFindTolerance);
+    return isTransparent(
+      ctx,
+      this.targetFindTolerance,
+      this.targetFindTolerance,
+      this.targetFindTolerance
```
ShaMan123, line 689: looking at this again shouldn't it be multiplied by retina?

asturur: in my branch i completely removed retina from the logic of target transparent.
I don't think it should be involved at all.

asturur:
Is not very easy to benchmark this, but i see like improvents here.
Sometimes my laptop goes in slow mode and the results all over the place.

image

image

Master is so higher up right now, sometimes it get the full 16ms to get back a result, and i also notice we run findTarget twice when not needed ( i ll open a PR for that )

image

src/canvas/SelectableCanvas.ts
```diff
@@ -656,6 +649,18 @@ export class SelectableCanvas<
...
+  setTargetFindTolerance(value: number) {
+    this.targetFindTolerance = value;
+    const size = Math.ceil(
+      2 * Math.max(this.targetFindTolerance, 1) * this.getRetinaScaling()
+    );
+    if (this.pixelFindCanvasEl.width < size) {
+      this.pixelFindCanvasEl.width = this.pixelFindCanvasEl.height = size;
+      const retina = this.getRetinaScaling();
+      this.pixelFindContext.scale(retina, retina);
+    }
```
asturur, line 661: after playing around with this branch a bit and the feature, i think this should be slightly different:
```suggestion
  setTargetFindTolerance(value: number) {
    value = Math.floor(value); // floats aren't supported. but just in case get rid of them
    this.targetFindTolerance = value;
    const size = 2 * value + 1;
    this.pixelFindCanvasEl.width = this.pixelFindCanvasEl.height = size;
  }
```
notice that size is  2 * tolerance + 1. 1 is the pixel that we are hoverin and tolerance 3 for example, would mean we always take 3 pixels each side in addition to the original pixel.
So the pixelTargetFind canvas is always an odd number of pixels ( 1,3,5,7... )

ShaMan123: sounds correct

src/canvas/SelectableCanvas.ts
ShaMan123: I thought it is needed for precision in case of blur

src/canvas/SelectableCanvas.ts
```diff
@@ -667,44 +672,22 @@ export class SelectableCanvas<
...
-
-    target.selectionBackgroundColor = originalColor;
-
-    return isTransparent(ctx, x, y, this.targetFindTolerance);
+    return isTransparent(
+      ctx,
+      this.targetFindTolerance,
+      this.targetFindTolerance,
+      this.targetFindTolerance
+    );
```
asturur, line 690: This is the version of isTargetTransparent i ended up with, for the following reasons.
Still trying to figure out if retina should be included or not, my gut feeling is no, because is not part of the mouse interactions, is just a visual enhancement.
I left in the cache optimization because i was worried we could invalidate the cache with that render executed outside of context, didn't investigate deeply.
```js
  isTargetTransparent(target: FabricObject, x: number, y: number): boolean {
    const pixelCtx = this.pixelFindContext;
    const retina = this.getRetinaScaling();
    pixelCtx.resetTransform();
    this.clearContext(pixelCtx);
    const tolerance = this.targetFindTolerance;
    // normalized pointer is independendent from zoom and scale and retina.
    const normalizedPointer = this._normalizePointer(target, new Point(x, y));
    if (isFabricObjectCached(target) && !target.dirty) {
      // optimization: use the cache
      pixelCtx.translate(
        -target.cacheTranslationX / retina -
          (normalizedPointer.x * target.zoomX) / retina +
          tolerance,
        -target.cacheTranslationY / retina -
          (normalizedPointer.y * target.zoomY) / retina +
          tolerance
      );
      // the cache is retina scaled. we want to counter that.
      pixelCtx.scale(1 / retina, 1 / retina);
      // performance optimization:
      // we draw the hit area to the dedicated canvas instead of using `getImageData` on the target's cache canvas
      pixelCtx.drawImage(target._cacheCanvas, 0, 0);
    } else {
      pixelCtx.translate(-x + tolerance, -y + tolerance);
      pixelCtx.transform(...this.viewportTransform);
      target.render(pixelCtx);
    }
    return isTransparent(pixelCtx, tolerance, tolerance, tolerance);
  }
```

asturur: this is just for reference, is missing the selection background color that you added back.

ShaMan123: My concern is testing a cached nested object. I don't think the drawing is correct in that situation. That is why I removed it completely. Also in case of a dirty cache the method will return a wrong value. So I thought it best to run and redraw cache if it needs

src/util/misc/isTransparent.ts
```diff
@@ -28,24 +29,19 @@ export const isTransparent = (
     }
   }
 
-  let _isTransparent = true;
   const { data } = ctx.getImageData(
     x,
     y,
     tolerance * 2 || 1,
     tolerance * 2 || 1
   );
```
asturur, line 37:
```suggestion
  const { data } = ctx.getImageData(
    x - tolerance,
    y - tolerance,
    tolerance * 2 + 1,
    tolerance * 2 + 1,
  );
```

If we write this in this way we can get rid of  the if condition above.
Reading getImageData out of bounds seems in spec

```
The getImageData(sx, sy, sw, sh, settings) method steps are:

If either the sw or sh arguments are zero, then throw an ["IndexSizeError"](https://webidl.spec.whatwg.org/#indexsizeerror) [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException).

If the [CanvasRenderingContext2D](https://html.spec.whatwg.org/multipage/canvas.html#canvasrenderingcontext2d)'s [origin-clean](https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-origin-clean) flag is set to false, then throw a ["SecurityError"](https://webidl.spec.whatwg.org/#securityerror) [DOMException](https://webidl.spec.whatwg.org/#dfn-DOMException).

Let imageData be a [new](https://webidl.spec.whatwg.org/#new) [ImageData](https://html.spec.whatwg.org/multipage/canvas.html#imagedata) object.

[Initialize](https://html.spec.whatwg.org/multipage/canvas.html#initialize-an-imagedata-object) imageData given sw, sh, [settings](https://html.spec.whatwg.org/multipage/canvas.html#initialize-imagedata-settings) set to settings, and [defaultColorSpace](https://html.spec.whatwg.org/multipage/canvas.html#initialize-imagedata-defaultcolorspace) set to [this](https://webidl.spec.whatwg.org/#this)'s [color space](https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-color-space).

Let the source rectangle be the rectangle whose corners are the four points (sx, sy), (sx+sw, sy), (sx+sw, sy+sh), (sx, sy+sh).

Set the pixel values of imageData to be the pixels of [this](https://webidl.spec.whatwg.org/#this)'s [output bitmap](https://html.spec.whatwg.org/multipage/canvas.html#output-bitmap) in the area specified by the source rectangle in the bitmap's coordinate space units, converted from [this](https://webidl.spec.whatwg.org/#this)'s [color space](https://html.spec.whatwg.org/multipage/canvas.html#concept-canvas-color-space) to imageData's [colorSpace](https://html.spec.whatwg.org/multipage/canvas.html#dom-imagedata-colorspace) using ['relative-colorimetric'](https://drafts.csswg.org/css-color/#valdef-color-profile-rendering-intent-relative-colorimetric) rendering intent.

Set the pixels values of imageData for areas of the source rectangle that are outside of the [output bitmap](https://html.spec.whatwg.org/multipage/canvas.html#output-bitmap) to [transparent black](https://drafts.csswg.org/css-color/#transparent-black).

Return imageData.
```

src/canvas/SelectableCanvas.ts
asturur: i ll try to figure out if we can ever break the object cache by re-rendering a cached object.

src/util/misc/isTransparent.ts
ShaMan123: - `document.createElement('canvas').getContext('2d').getImageData(0,0,1,1)` returns [0,0,0,0] though IMO it should return null

ShaMan123: so that is fine as well

src/canvas/SelectableCanvas.ts
asturur: considering retina makes the cached/noncached case very similar and cut differences there.
But force us to continuosly round numbers

ShaMan123: That isn't too bad, is it?
I am thinking that this feature is used with text and text is heavily impacted by retina

asturur: yes i think it will win the smaller code at the end.
traget transparency isn't meant to be a pixel sniper effect. It has just to guess if is completely transparent or not.

src/util/misc/isTransparent.ts
asturur: getImageData(0,0,1,1) should give you a pixel correct? why would you have null?

ShaMan123: I think the spec is confusing. I am asking for a non existing pixel and I get one. But for our sake it is fine since it is transparent

ShaMan123:
As far as I am concerned you have taken over this PR and are in charge

src/util/misc/isTransparent.ts
asturur: oh i see what you mean.
I think is a point of view. Having always pixels is definitely more ergonomic.
Of course if you are hunting for transparent pixels and you go accidentally out of bounds, you are confused.

ShaMan123 (commented):
Looks great, unit tests seem to prove accuracy has been reached

ShaMan123 (commented):
maybe we should a test under a transformed group

asturur:
Ok i ll merge now.
I m half sad that code readibility affect code size so much.
I wonder if there is some terser option that can help.
Like protected methods do not need to be maintained in full size in naming for example.
We removed a lot in this PR yet the final bundle size growns.

asturur (approved):

asturur:
Kind of we forgot that now you can't read how much tolerance you set.

ShaMan123:
> Kind of we forgot that now you can't read how much tolerance you set.

You mean the enhanced value?


I wonder about this bundle size. Is it really accurate? We dropped 50 lines of code and didn't add anything that is compiled down. So what is this data? Whatever @asturur we made it better.

Issue #7451, opened 2021-10-24
feat(): web-worker/offscreen support


ajil-paul:
I am using webpack for a React.js application. I need to do a heavy work with fabric. So I tried to use web worker for that purpose. I used worker-loader plugin for running web-worker. And my woker is working fine. This is my worker file

```
const fabric = require('fabric')
self.onmessage = async ({ data: { canvasInfo } }) => {

    const canvas = new fabric.StaticCanvas('canvas', {
      width:1024,
      height: 1024,
      selection: false,
    })

    // Doing heavy work with fabric canvas fn()

    self.postMessage({
      message:"success"
    })
}
```

I am getting the following error.
```
Uncaught TypeError: jsdom.JSDOM is not a constructor
    at Object../node_modules/fabric/dist/fabric.js (fabric.js:25)
    at __webpack_require__ (bootstrap:21)
    at web.worker.js:5
    at web.worker.bundle.worker.js:140722
    at web.worker.bundle.worker.js:140724
```

AJIL-PAUL:
Yes, I have discussed the same thing in [stackoverflow](https://stackoverflow.com/questions/69467398/unable-to-import-fabric-js-in-web-worker). I am still stuck on the issue.

kingschnulli:
this should be addressed with #7596 as well as other node detection related issues

ShaMan123:
see #8208

PR #8774, opened 2023-03-11
feat(): Offscreen env POC

ShaMan123:
## Motivation

Performance
A stab at offscreen env support
#7451

## Description

We should consider splitting logic into:
- events
- rendering
- rest

This will allow to hook fabric to offscreen with ease and make node env better as well
I suggest we reorder the code base into
- core
- bindings
  - web
  - offscreen
  - node

## Gist

I think we should discuss if fabric should exist on a single thread or manage threads for caches etc.
Thinking of something we discussed, before scaling an object (on selected or on corner hover) we can render the cache offscreen 10X scaled without it affecting the main thread, and actually render more than one to optimize scaling.

- [ ] [Offscreen image](https://ezrogha.hashnode.dev/how-to-render-images-in-offscreencanvas)

https://github.com/mdn/dom-examples/tree/main/web-workers/offscreen-canvas-worker
https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/getContext

https://mdn.github.io/dom-examples/web-workers/offscreen-canvas-worker/

## In Action

https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-bb6sne?file=%2Fpackage.json&selection=%5B%7B%22endColumn%22%3A60%2C%22endLineNumber%22%3A8%2C%22startColumn%22%3A60%2C%22startLineNumber%22%3A8%7D%5D

https://user-images.githubusercontent.com/34343793/224482564-169baa83-34c6-4afe-82b5-762a5308b68e.mp4


asturur:
If we can get some directive to developers with offscrren support that is great.
But we are not going into specialized app behavior combining the modes to obtain faster caching at least not in the library.
If you want to make a repo that shows how to combine offscreen with normal to speed up cache, great.
I talked with you ( if i m not mistaken )

I can see 2 usages of the offScreen, the one that you shown in the demo and another where you use post message to send the json data and get the blob of the resulting render from the post message.

asturur:
I have been thinkin of this on and off since is out,
i don't get how we would support somenthing like
`canvas.getObjects()` for example and then object manipulation.
If thereisn't a way to share access to the data across threads, every api call becomes a postMessage?

asturur:
Did you check if using offscreen canvas in the main thread offer any way any slight performance improcement?

ShaMan123:
Every API call will become a postMessage. But I think that only rendering (including filtering and animating) should be supported offscreen with some kind of syncing using postMessage.
We could have an object exist on each thread and sync them.... don't know how bad that might be.
I didn't try using offscreen on the main thread but I read there is a way if I am not mistaken. Arguably, less performant.
Will look into it.

ShaMan123:
It can.
It can be its own entry point, fabric/offscreen

PR #8775, opened 2023-03-12
fix(): Text broken when doing inputs

asturur:
## Motivation

A recent change for removing the stateful mix has broken text input

## Description

Every input is going to clear the dimension cache.
For the few cases in which we swap and `a` with another `a` in the same place, well we will pay the price of recalculation.
Not a big deal.

I could have added `this._forceClearCache = true` in onInput, but since `updateTextarea` is the part that immediately consumes it, and is also the only place from where that function is called, i opted to have `updateTextarea` always clear the dimension cache and set as dirty in case the initDimensions would return the exact same size ( swapping a letter with one with the same size )

Added 2 tests for onInput ( a small test but better than nothing ) and one for updateFromTextarea.


asturur:
I need to write a test and verify that things like insertChars are working ro not

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -188,16 +201,6 @@ export abstract class ITextKeyBehavior<
...
-    if (this.hiddenTextarea.value === '') {
-      this.styles = {};
-      this.updateFromTextArea();
-      this.fire('changed');
-      if (this.canvas) {
-        this.canvas.fire('text:changed', { target: this });
-        this.canvas.requestRenderAll();
-      }
-      return;
-    }
```
asturur, line 200: moved this up since it doesn't depend from any of the let/const.

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -265,12 +268,7 @@ export abstract class ITextKeyBehavior<
       }
       this.insertNewStyleBlock(insertedText, selectionStart, copiedStyle);
     }
-    this.updateFromTextArea();
-    this.fire('changed');
-    if (this.canvas) {
-      this.canvas.fire('text:changed', { target: this });
-      this.canvas.requestRenderAll();
-    }
```
asturur, line 273: deduped this

asturur:
this is ready

asturur:
@ShaMan123 have a look tomorrow i want to have master fixed

.codesandbox/templates/next/pages/index.tsx
ShaMan123, line 1: Is there a reason for removing the animation? I thought it is good as a showcase but it is annoying when developing

ShaMan123 (commented):
This looks fine but is it enough?
What about #8711 ? It should be merged into this PR as well

.codesandbox/templates/next/pages/index.tsx
asturur: that is also an internal tool, not a showcase thing.
I needed to test text, i started with my first experience with something movinng that i couldn't click, when i managed to click it i figured out it wasn't editable.
I trashed it.
There is also the issue that when running it invalidates the cache every frame and is just more work on a laptop that is rebuilding the library in the background and rebuilding the next pages.
Everyone will use the template to add something that is needed in that occasion, maybe the best default is a rect, not even a textbox

asturur:
https://github.com/fabricjs/fabric.js/pull/8711 probably.
As often happens the description of your pr are cryptic.

```
_shouldClearDimensionCache was changed in https://github.com/fabricjs/fabric.js/pull/8663, not taking into account state changes any longer including changes to text which should trigger a layout - text was changed but layout was skipped
This PR moderates those calls
```

Isn't clear what is broken and what is being fixed.
What does it means moderates? is removeChar broken? is insertChar broken?

I will merge both

PR #8777, opened 2023-03-12
ci() fix node-14 by moving sandbox out of test script

asturur:


ShaMan123:
you fixed it by regenerating the lock file to an old version not by moving the script

ShaMan123:
https://github.com/npm/cli/issues/4787

PR #8778, opened 2023-03-13
Fix index issue

Lazauya:
Added a line to make sure I can run the index script. I have no clue what was causing this bug, but for some reason on my machine the `version` function wasn't returning `this`, even though the docs say it should. This change shouldn't change any functionality, only a small bug fix. Specifically, the error was saying: 
```
file:///<redacted>/fabric.js/scripts/index.mjs:490
  .showSuggestionAfterError();
  ^

TypeError: Cannot read properties of undefined (reading 'showSuggestionAfterError')
    at file:///<redacted>/fabric.js/scripts/index.mjs:490:3
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

PR #8779, opened 2023-03-13
ci(): restore `package-lock.json` after regression

ShaMan123:
## Motivation

#8777 regression 

## Description

@asturur what version of node/npm are you using? Mine warned me about the lock file and regenerated it
I bet that is the reason for #8778


ShaMan123:
image

asturur:
so node14 wasn't working anymore because after the sandbox pr they package-lock was too new and was breaking the npm version that comes by default node14.
So i installed node14 npm 6 and generate a node14 package-lock.json to have the pipeline pass.

ShaMan123:
My system complains and won't let me use v1

asturur:
yes mine tries to update it all the time. i think this fix is exactly ok as any other. But before wasn't an issue. i was supposed to look at how we broke it before merging and i forgot.

asturur:
ok so the reason is that when you updated the ci sandbox you moved the package lock from version 2 to version 3.
Probably we can just have version 2 till node14 is around since node14 is supported, and then when we deprecate node 14 we can move to version 3.
Otherwise all ( few node14 users will have to use this flag now to make it work ).

asturur:
Probably you recently moved to node18?

asturur:
https://github.com/fabricjs/fabric.js/pull/8781 here it is

PR #8782, opened 2023-03-14
reflect NUM_FRACTION_DIGITS to SVG path data

### LINKED ISSUE #5734: NUM_FRACTION_DIGITS doesn't work for SVG export
## Version
3.3.0

## Test Case
https://www.prima-cms.com/2629.svg

Nodejs or browser?
Which browsers?

## Steps to reproduce
Export a canvas to SVG

## Expected Behavior
Rounded number

## Actual Behavior
Decimal number

I run Fabric like this 👍 
`custengine.canvas = new fabric.Canvas($(custengine.container).children('canvas').attr('id'), {enableRetinaScaling: true});
    custengine.canvas.preserveObjectStacking = true;
    custengine.canvas.selection = false;
    fabric.Object.NUM_FRACTION_DIGITS = 0;`

But the SVG export contain big float number (see sample)

And the problem is with background image pattern, floatting number create half pixel shift in Imagick render.

Example: see this file https://www.prima-cms.com/imagick.png that is a conversion of the SVG to png and you can see 1 pixel in background patterns.

How works NUM_FRACTION_DIGITS ? Is there anyway solution to export SVG without floating numbers ?

Thanks

tushuhei:
## Motivation
Fix #5734. `toSVG` does not consider `NUM_FRACTION_DIGITS` in exporting path data, which makes the resulting SVG (and JSON eventually) unnecessarily large.

## Description
`util.path.joinPath()` seriarizes SVG path data by joining commands. This change is to extend the function by additing the `fractionDigits` argument, which represents the number of fraction digits for the serialized SVG path data.

## Changes
- **BREAKING**: Added a new argument to the `joinPath` function. If no argument is passed `toFixed` will round it to an int
- Call the `joinPath` function with `config.NUM_FRACTION_DIGITS`.
- Add a unit test for the `joinPath` function.

## In Action
```html
<canvas id="fabric"></canvas>
<p id="output"></p>
<div id="reconstructed"></div>
<script src="https://unpkg.com/fabric@5.3.0/dist/fabric.js"></script>
<script>
const canvas = new fabric.Canvas('fabric');
const path = new fabric.Path('M55.23 82.29a11.13 11.13 0 0 0-10.46 0L23.55 95.12c-2.88 1.74-4.61.49-3.84-2.78l5.65-24.15a11.16 11.16 0 0 0-3.23-9.95L3.36 42c-2.54-2.2-1.88-4.23 1.47-4.51l24.71-2.09A11.11 11.11 0 0 0 38 29.27l9.63-22.85c1.3-3.1 3.44-3.1 4.74 0L62 29.27a11.11 11.11 0 0 0 8.46 6.15l24.71 2.09c3.35.29 4 2.31 1.47 4.51L77.87 58.24a11.16 11.16 0 0 0-3.23 9.95l5.65 24.15c.76 3.27-1 4.52-3.84 2.78Z');
canvas.add(path);
fabric.Object.NUM_FRACTION_DIGITS = 0;
document.getElementById("output").textContent = canvas.toSVG();
document.getElementById("reconstructed").innerHTML = canvas.toSVG();
</script>
```

**Before:**
Screenshot 2023-03-14 21 49 37

**After:**
Screenshot 2023-03-14 21 47 18


ShaMan123 (commented):
Thanks
Good job
Add a changelog please

ShaMan123:
And run `npm run prettier:write` and commit the changes

Read the CONTRIBUTING.md for more details on how to finish up the PR

src/util/path.ts
```diff
@@ -935,7 +936,17 @@ export const getRegularPolygonPath = (numVertexes, radius) => {
 /**
  * Join path commands to go back to svg format
  * @param {Array} pathData fabricJS parsed path commands
+ * @param {number} fractionDigits  number of fraction digits to "leave"
  * @return {String} joined path 'M 0 0 L 20 30'
  */
-export const joinPath = (pathData) =>
-  pathData.map((segment) => segment.join(' ')).join(' ');
+export const joinPath = (pathData: PathData, fractionDigits?: number) =>
```
ShaMan123, line 942:
```suggestion
export const joinPath = (pathData: PathData, fractionDigits: number) =>
```

ShaMan123 (commented):
Passing undefined to `toFixed` rounds the number entirely so that doesn't support what was before.
So I guess we should make the arg non optional as you did but I am not sure.
It is used only but the `_toSVG` method so I think it doesn't really matter.
@asturur thoughts?

src/shapes/Path.ts
```diff
@@ -194,7 +194,8 @@ export class Path extends FabricObject {
    * of the instance
    */
   _toSVG() {
-    const path = joinPath(this.path);
+    const digits = config.NUM_FRACTION_DIGITS;
+    const path = joinPath(this.path, digits);
```
ShaMan123, line 198:
```suggestion
    const path = joinPath(this.path, config.NUM_FRACTION_DIGITS);
```

inline

ShaMan123 (approved):

asturur:
This is fine.
But in the changelog we need to write BREAKING.

asturur (changes requested):
Please add BREAKING in front of the changelog line

tushuhei:
Sure thing, done.

asturur:
@ShaMan123 your question was collapsed for me and i didn't see.
Sorry @tushuhei @ShaMan123 is right. While joinPath is used once, is still an external utility. Forcing the use of toFixed means making impossible to keep the original values.
(please corret me if i m mistaken)

so for joinPath toFixed was good as an optional parameter, and if is isn't passed should not truncte the numbers

src/util/path.ts
```diff
@@ -935,7 +936,17 @@ export const getRegularPolygonPath = (numVertexes, radius) => {
...
  */
-export const joinPath = (pathData) =>
-  pathData.map((segment) => segment.join(' ')).join(' ');
+export const joinPath = (pathData: PathData, fractionDigits: number) =>
+  pathData
+    .map((segment) => {
+      return segment
+        .map((arg, i) => {
+          if (i === 0) return arg;
+          return toFixed(arg, fractionDigits);
```
asturur, line 948: fractionDigits should go back to optional and

```suggestion
          return isNumber(fractionDigits) ? toFixed(arg, fractionDigits) : arg;
```

Or something similar to this.

ShaMan123: That is what I thought but then again looking in a broader context it is an svg util and everything in svg output forces the config value without giving a way out

tushuhei: Made `fractionDigits` optional.

asturur: i used `joinPath` a bunch of time on my own to do things in json export for example or manual things. I use it outside the context of svg export.

asturur (approved):

Exlord:
Hi, I am using the custom build version which is 5.2. Is this update available there ?

Exlord:
Is version 6 production ready ?

ShaMan123:
beta
much better than v5

PR #8787, opened 2023-03-18
chore(TS): Path util typings and refactoring

Lazauya:
Things in this PR:
- Created a folder for path utils (`path`) that includes a type file and a functions file
- Added types for parsed and string commands, both relative and absolute
    - Functions for determining command type may be redundant, requires review
    - Created regex to match each (cleaned) path command
    - Simplified and "complex" (unsimplified) command types
- Refactored path parsing to be simpler and more general
    - Removed specific arc edge case
- Refactored `makePathSimpler` slightly to be more type friendly
    - Remove support for lowercase z (changed tests to reflect this)
- Refactored other various other types in `path.ts` to make code more type-friendly (e.g. TPathSegmentInfo)
- Changes the `PathData` type to the more specific `TSimplePathData` type elsewhere in the code

Rationale:
Properly typing the path parsing is a step towards a fully typed SVG parser. I also want to have real-time feedback for invalid/valid path commands. I had this particular issue when using the DefinitelyTyped library; not only were the signatures for path-related things screwed up, but the path command types were very unhelpful.


asturur:
hi @Lazauya  ll try to be quick and effective with this PR.
I didn't read it yet, but i have been trying to be super clear with @ShaMan123 too that continuos refactoring is making us stale in a beta stage too long.

I have a couple of things to merge the next hours and then i ll give an eye to this one

Lazauya:
@ShaMan123 @asturur This PR only substantially changes one file (and adds another). There is a minor functionality change: lowercase z is not supported. This was to make the code simpler and more uniform. Other than that, all functionality should remain.

src/shapes/Path.ts
```diff
@@ -36,12 +37,14 @@ export class Path extends FabricObject {
 
   /**
    * Constructor
-   * @param {Array|String} path Path data (sequence of coordinates and corresponding "command" tokens)
+   * @param {TSimplePathData} path Path data (sequence of coordinates and corresponding "command" tokens)
+   * @param {Number} left
+   * @param {Number} top
```
asturur, line 42: mostly sure this is incorrect as JSDOC.

Lazauya: You're right, fixed

Lazauya: Maybe I'll go through and replace all the incorrect elementary typed jsdoc comments like this in another PR, because they're a lot of them everywhere

src/brushes/PencilBrush.ts
```diff
@@ -2,17 +2,17 @@ import { ModifierKey, TEvent } from '../EventTypeDefs';
 import { Point } from '../Point';
 import { Shadow } from '../Shadow';
 import { Path } from '../shapes/Path';
-import { PathData } from '../typedefs';
-import { getSmoothPathFromPoints, joinPath } from '../util/path';
+import { getSmoothPathFromPoints, joinPath } from '../util/path/path';
 import type { Canvas } from '../canvas/Canvas';
 import { BaseBrush } from './BaseBrush';
+import { TSimplePathData } from '../util/path/path_types';
```
asturur, line 8: i think the new pattern is to use camelCase for files.
I say i think because it wasn't really planned it just happened.
What about `../util/path/typedefs` ?

src/cache.ts
```diff
@@ -1,4 +1,5 @@
 import { config } from './config';
+import {TBounds} from "./typedefs";
```
asturur, line 2: mmm missing spaces around curly brackets should trigger prettier

src/shapes/Path.ts
asturur: just remove this 2 new lines you added line 41 and 42

src/brushes/PencilBrush.ts
Lazauya: done

src/shapes/Path.ts
```diff
@@ -3,7 +3,7 @@ import { config } from '../config';
 import { SHARED_ATTRIBUTES } from '../parser/attributes';
 import { parseAttributes } from '../parser/parseAttributes';
 import { Point } from '../Point';
-import { PathData } from '../typedefs';
+import { TClassProperties } from '../typedefs';
```
ShaMan123, line 6: unused

src/typedefs.ts
```diff
@@ -100,6 +95,8 @@ export type TCacheCanvasDimensions = {
   y: number;
 };
 
+export type TBounds = IPoint[];
```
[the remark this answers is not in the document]
ShaMan123: no IPoint is used

src/util/path/path.ts
ShaMan123, line 1: I would rename this file to index

src/util/path/path.ts
```diff
@@ -156,10 +170,20 @@ const arcToSegments = (toX, toY, rx, ry, large, sweep, rotateX) => {
...
+/**
+ * @private
+ * Calculate the angle between two vectors
+ * @param ux u endpoint x
+ * @param uy u endpoint y
+ * @param vx v endpoint x
+ * @param vy v endpoint y
  */
-const calcVectorAngle = (ux, uy, vx, vy) => {
+const calcVectorAngle = (
```
ShaMan123, line 181: should be removed and use calcAngleBetweenVectors

src/util/path/path.ts
```diff
@@ -297,64 +333,64 @@ export const fromArcToBeziers = (
...
-  for (let i = 0; i < len; ++i) {
-    let converted = false;
-    const current = path[i].slice(0);
+  const destinationPath: TSimplePathData = [];
+  let previous: string | undefined,
+    // placeholders
+    controlX = 0,
+    controlY = 0;
+  for (const parsedCommand of path) {
+    const current: TComplexParsedCommand = Object.assign([], parsedCommand);
```
asturur, line 365: `[...parsedCommand]` vs `Object.assign([], parsedCommand)`. slice(0) or concat() are good too.

src/util/path/path.ts
```diff
@@ -297,64 +333,64 @@ export const fromArcToBeziers = (
...
     ) {
       case 'l': // lineto, relative
-        current[0] = 'L';
         current[1] += x;
         current[2] += y;
       // falls through
       case 'L':
         x = current[1];
         y = current[2];
+        converted = ['L', current[1], current[2]];
```
asturur, line 377: i have read through the changes of `makePathSimpler` and i want to ask you what makes you prefer this converted array approach vs the mutation of the previous array.
Is more verbose.
current was a new array every command so is not a mutation issue. What is it?

src/util/path/path.ts
```diff
@@ -486,18 +526,43 @@ export const makePathSimpler = (path: PathData): PathData => {
  * @param {Number} y2 starting point y
  * @return {Number} length of segment
  */
-const calcLineLength = (x1, y1, x2, y2) =>
-  Math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2);
+const calcLineLength = (
```
ShaMan123, line 529: use Point#distanceFrom

src/util/path/path.ts
```diff
@@ -651,9 +751,12 @@ export const getPathSegmentsInfo = (path: PathData): TPathSegmentsInfo[] => {
           current[5],
           current[6]
         );
-        tempInfo.iterator = iterator;
-        tempInfo.angleFinder = angleFinder;
-        tempInfo.length = pathIterator(iterator, x1, y1);
+        tempInfo = <TCurveInfo<'C'>>{
+          ...basicInfo,
```
asturur, line 755: in my opinion there is no need to spread basicInfo every time.
basicInfo is a new object every loop, can just be added of new properties.

src/util/path/path.ts
```diff
@@ -560,20 +652,23 @@ const pathIterator = (
...
     nextStep = 0.01,
-    lastPerc;
+    lastPerc = 0;
   // nextStep > 0.0001 covers 0.00015625 that 1/64th of 1/100
   // the path
   const iterator = segInfo.iterator,
     angleFinder = segInfo.angleFinder;
   while (tmpLen < distance && nextStep > 0.0001) {
-    p = iterator(perc);
+    p = { ...iterator(perc), angle: 0 };
```
ShaMan123, line 671: can you explain this?

src/util/path/path.ts
Lazauya: if we left it alone then it complains about not being able to assign 'm' to 'M' and things of that nature, which meant that either we'd need a ton of ugly casting or just plain ts-ignores. You can see the approach with types "just slapped on" more or less in the second commit.

src/util/path/path.ts
Lazauya: I tried that, but it was causing similar issues to what I said [here](https://github.com/fabricjs/fabric.js/pull/8787#discussion_r1141161099). I agree it would be prettier but I couldn't make the types work nicely without a bunch of annoying casting/"as" stuff.

src/util/path/path.ts
```diff
@@ -697,48 +805,67 @@ export const getPathSegmentsInfo = (path: PathData): TPathSegmentsInfo[] => {
...
+          ),
+          angle: 0,
+        };
+        info.angle = Math.atan2(segment[2] - segInfo.y, segment[1] - segInfo.x);
+      } else {
+        // TODO: provide a proper error
+        throw Error(
+          `Segment/path info mismatch, expected L, got ${segment[0]}`
+        );
+      }
```
asturur, line 861: why this change is necessary?

src/util/path/path.ts
```diff
@@ -697,48 +805,67 @@ export const getPathSegmentsInfo = (path: PathData): TPathSegmentsInfo[] => {
...
-        segPercent
-      );
+      info = {
+        ...new Point(segInfo.x, segInfo.y).lerp(
+          new Point(segInfo.destX, segInfo.destY),
+          segPercent
+        ),
+        angle: 0,
+      };
       info.angle = Math.atan2(
```
ShaMan123, line 841: this can also go into the info object

src/util/path/path.ts
```diff
@@ -697,48 +805,67 @@ export const getPathSegmentsInfo = (path: PathData): TPathSegmentsInfo[] => {
...
-      info.angle = Math.atan2(segment[2] - segInfo.y, segment[1] - segInfo.x);
+      if (isAbsLineCmd(segment)) {
+        info = {
+          ...new Point(segInfo.x, segInfo.y).lerp(
+            new Point(segment[1], segment[2]),
+            segPercent
+          ),
+          angle: 0,
+        };
+        info.angle = Math.atan2(segment[2] - segInfo.y, segment[1] - segInfo.x);
```
ShaMan123, line 855: same

src/util/path/path.ts
Lazauya: Cursed typing stuff again. If you don't verify it's a line, TS thinks it could be a z command which makes segment[1] be undefined. It SHOULD never happen, so the error is probably superfluous, but again I couldn't figure out how to make TS understand that in an easier way.

src/util/path/path.ts
```diff
@@ -754,81 +881,78 @@ export const getPointOnPath = (
...
+export const parsePath = (pathString: string): TComplexPathData => {
+  // clean the string
+  // add spaces around the numbers
+  pathString = pathString.replace(
+    new RegExp(`(${numberRegExStr})`, 'gi'),
+    ' $1 '
+  );
+  // replace annoying commas and arbitrary whitespace with single spaces
+  pathString = pathString.replace(/,/gi, ' ');
+  pathString = pathString.replace(/\s+/gi, ' ');
```
asturur, line 893:
```suggestion
  pathString = pathString.replace(
    new RegExp(`(${numberRegExStr})`, 'gi'),
    ' $1 '
  )
  // replace annoying commas and arbitrary whitespace with single spaces
  .replace(/,/gi, ' ')
  .replace(/\s+/gi, ' ');
```

would this work?

src/util/path/path.ts
Lazauya: Oh I forgot about `[...parsed]`, I like that a lot more and I'll make that change really quick

src/util/path/typedefs.ts
```diff
@@ -0,0 +1,480 @@
...
+  | TParsedAbsoluteCubicCurveCommand
+  | TParsedAbsoluteQuadraticCurveCommand;
+
+/**
+ * A series of simple paths
+ */
+export type TSimplePathData = TSimpleParsedCommand[];
+
+/**
+ * A point and the of the vector to the point and the x-axis
```
ShaMan123, line 478: unclear

src/util/path/typedefs.ts
```diff
@@ -0,0 +1,480 @@
...
+    `|(?:${reHorizontalLineCommand.source})` +
+    `|(?:${reVerticalLineCommand.source})` +
+    `|(?:${reClosePathCommand.source})` +
+    `|(?:${reCubicCurveCommand.source})` +
+    `|(?:${reCubicCurveShortcutCommand.source})` +
+    `|(?:${reQuadraticCurveCommand.source})` +
+    `|(?:${reQuadraticCurveShortcutCommand.source})` +
+    `|(?:${reArcCommand.source}))`,
+  'dgi'
+);
```
ShaMan123, line 454: 🤕
I hate regexp
Nice job I guess, I have no idea

src/util/path/path.ts
```diff
@@ -697,48 +805,67 @@ export const getPathSegmentsInfo = (path: PathData): TPathSegmentsInfo[] => {
   return info;
```
ShaMan123, line 805: can anyone tell me why the previous line exists?
I remember it was one of many pains fabric inflicted upon me

src/util/path/path.ts
Lazauya: I think so but I'll run the tests really quick and check

src/util/path/path.ts
ShaMan123: I see that in line https://github.com/fabricjs/fabric.js/pull/8787/files#diff-d309b6b9d88030a0c982bdc0536f31a874f2441c726e9dc7f293ff2dfbed2992R685 it is set 
Is it that why?

src/util/path/path.ts
ShaMan123: I did my best reviewing it, can't say I did a good job

ShaMan123 (commented):
Did my best reviewing, but didn't do a good job

ShaMan123:
And I see node 14 is failing because of regexp. I don't know what version supports what, grouping has never seemed cross platform to me

asturur:
@Lazauya there is a lot typing stuff here that is good.
A couple of things we need to change:
- the typedef files has regex, functions and files. keep the types there, make a file for regexes, make another file for that type check functions ( most of them are unused )
- use import type where appropriate
- isAbsCubicCurveShortcutCmd and many like functions are there here are unused. Why should they stay? they can be useful to someone yes. Do you have specific use case for those? I m ok with keeping them as long as those are not surfaced in the main bundle. ( developers will eventually find them in the right file from an import path. Otherwise remove them )

src/parser/constants.ts
```diff
@@ -15,7 +15,7 @@ export const reNum = '(?:[-+]?(?:\\d+|\\d*\\.\\d+)(?:[eE][-+]?\\d+)?)';
 
 export const svgNS = 'http://www.w3.org/2000/svg';
 
-export const commaWsp = '(?:\\s+,?\\s*|,\\s*)';
+export const commaWsp = String.raw`(?:\s+,?\s*|,\s*|$)`;
```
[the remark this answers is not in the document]
Lazauya: the linter actually complains if I don't use it, weird. In any case its just a fancy version of ` `` `

src/util/path/path.ts
ShaMan123: I don't understand this bit but casting at the top to something loose and then recasting at the bottom to something strict should do the job w/o too much ugliness

ShaMan123: You can always create a type util that does that for you inline I guess

src/typedefs.ts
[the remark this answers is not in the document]
Lazauya: I'm keeping this for two reasons: IPoint is just a generalized version of Point, and the bounds might be more than a pair (think general polygon bounds)

asturur:
My only big issue is added code because typescript isn't smart enough

src/util/path/path.ts
asturur: we should at least reuse the named variables where possible.
`converted = ['L', x, y]` and there are other cases like this down the file
This is added code for nothing but shouldn't be much and most of it is minifiable

src/util/path/path.ts
Lazauya: I wanted to return `p` but I couldn't set `p` to `iterator` there because it only returned a `Point` and not a `TPointAngle`. I was just trying to make ts happy

asturur:
```
abogazzi@abogazzi-mbp > /abogazzi/fabric.js > ts-path-util $ ls -l  dist/index.min.js
-rw-r--r--  1 abogazzi  staff  313127 Mar 18 22:47 dist/index.min.js
```
vs master

```
abogazzi@abogazzi-mbp > /abogazzi/fabric.js > master $ ls -l  dist/index.min.js
-rw-r--r--  1 abogazzi  staff  311559 Mar 18 22:48 dist/index.min.js
```

1600 bytes added for code style changes. i have issues accepting it.
Let's start remove the unused functions or keep them in a file that is not imported ( but still transpiled ) and let's see where we go

asturur:
and node14 issue as @ShaMan123 pointed out needs to be fixed

src/util/path/path.ts
Lazauya: ¯\_(ツ)_/¯

src/util/path/path.ts
ShaMan123:
```ts
type AbsPathCommad<CmdArr extends [string,...I],I extends number[] = number[]> =
 [Uppercase<CmdArr[0]>,...I]
 
type test = AbsPathCommad<['m', 1, 3]>



function a(arg:['m',number,number]|['M',number,number]){
    switch(arg[0]){
        case 'm':
        case 'M':
        //  TS accepts this
         const upperM = (arg as AbsPathCommad<typeof arg>)[0]
            
    }
}



```



```ts
"use strict";
function a(arg) {
    switch (arg[0]) {
        case 'm':
        case 'M':
            //  TS accepts this
            const upperM = arg[0];
    }
}

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggRgZwAoENgAsDCB7AtrlAEwB5NdCYAnSqCAD2AgDtCEoBtBYSgSyYHMANADpRASQC6gsbQbNWUJgFdccCJXYSoAXkUq1GiQD4dAKA4BVMJEoBjFAgilyVDQAZjI8RPOnQkKEYuHVhEVAwcfCJidgByXFjBKABGJIBmY1Ms0wAzJSZbYB5sJigUAAoUSn4ALjiEwWVVdUb9dQkAHziAWUSmg1bmygkASgBvcygphAB3HmBbdErq9g9xyanNqHtHKHjYmo2t7YdoWN7D482AemupgBUAZTLbWwgwYDYMHgQj49sSsElNZ1N0Qst+GU2PBkGgsHgCCR-BBsDkytUjCNVj4rlcjgBfUyErJAA)

src/util/path/path.ts
Lazauya: fixed

src/util/path/path.ts
Lazauya: using `as` is bad practice and should be avoided as much as possible. I really don't like this solution but I'm not set in stone, if both of you think it's better I can revert it I guess.

Lazauya:
> ```
> abogazzi@abogazzi-mbp > /abogazzi/fabric.js > ts-path-util $ ls -l  dist/index.min.js
> -rw-r--r--  1 abogazzi  staff  313127 Mar 18 22:47 dist/index.min.js
> ```
> 
> 
>      
>   
> 
> vs master
> 
> 
> ```
> abogazzi@abogazzi-mbp > /abogazzi/fabric.js > master $ ls -l  dist/index.min.js
> -rw-r--r--  1 abogazzi  staff  311559 Mar 18 22:48 dist/index.min.js
> ```
> 
> 
>      
>   
> 
> 1600 bytes added for code style changes. i have issues accepting it.
> Let's start remove the unused functions or keep them in a file that is not imported ( but still transpiled ) and let's see where we go

I figured this would be an issue. There was a brief time when I thought I needed them then realized I didn't. But I thought it was a waste to get rid of them. How can we make them not be bundled?

src/util/path/path.ts
ShaMan123: Perhaps, but not using as makes us slaves to TS and its disabilities
I am for TS all along but we must draw a line when it starts driving us mad

ShaMan123: If it offends you using as then write a casting function (that uses as) but is written once


```ts
type AbsPathCommad<CmdArr extends [string,...I],I extends number[] = number[]> =
 [Uppercase<CmdArr[0]>,...I]
 
type test = AbsPathCommad<['m', 1, 3]>

function cast<S extends T, T = unknown>(a: T): S {
    return a as S
}


function a(arg:['m',number,number]|['M',number,number]){
    switch(arg[0]){
        case 'm':
        case 'M':
        //  TS accepts this
         const upperM = (arg as AbsPathCommad<typeof arg>)[0]
         cast<AbsPathCommad<typeof arg>>(arg)[0]
            
    }
}



```



```ts
"use strict";
function cast(a) {
    return a;
}
function a(arg) {
    switch (arg[0]) {
        case 'm':
        case 'M':
            //  TS accepts this
            const upperM = arg[0];
            cast(arg)[0];
    }
}

```


</details>


```json
{
  "compilerOptions": {
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "strictBindCallApply": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "alwaysStrict": true,
    "esModuleInterop": true,
    "declaration": true,
    "target": "ES2017",
    "jsx": "react",
    "module": "ESNext",
    "moduleResolution": "node"
  }
}
```


</details>

**Playground Link:** [Provided](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggRgZwAoENgAsDCB7AtrlAEwB5NdCYAnSqCAD2AgDtCEoBtBYSgSyYHMANADpRASQC6gsbQbNWUJgFdccCJXYSoAXkUq1GiQD4dAKA4BVMJEoBjFAgilyVDQAZjI8RPOnQkKEYuHVhEVAwcfCJidgByXFjBKABGJIBmY1NTADMlJltgHmwmKHsuYgBlWUYWNgAVJLqQvIBrJmwAdyYjAAoUAC4oOoBKQaqAb3MoacoIYCVKEpQoBygK0wBfLJy8gqKlvsp+friEwWVVdXP9dQkAHziAWUSLg2vLyglhyenphA6eMBbOhDvx2B5vlNfr8ytB4rF+lDoaUHHDnojkb8APRY6Z1KooWy2CBgYBsDA8BBI5G2YrBJTWdSPEKglZseDINBYPAEEj+CDYbIrI5GYbgnyYmEOYDEDnhblRPngAVClAi3pq-hijzUyVIrZbLJAA)

ShaMan123: actually cast is a good util for fabric in general IMO

ShaMan123: > Perhaps, but not using as makes us slaves to TS and its disabilities I am for TS all along but we must draw a line when it starts driving us mad

And not damage perf for it to be happy

src/typedefs.ts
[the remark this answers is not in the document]
ShaMan123: It is not used anywhere else. 
TBounds is not needed at all. I want it removed. Just adding confusion to naming. 
No one will use bounds of curve in other cases because we have stroke projection for that
Please remove it. The method understand the return type so you only need to type 
`boundsOfCurveCache: Record<string, [Point,Point]> = {};` and remove the rest

src/util/path/path.ts
ShaMan123: So the same applies here.
Making a crazy thing happy is to be avoided 😵

src/util/path/path.ts
ShaMan123: Another guiding point for me I would like to point out.
The main purpose of TS is to make it easy to consume fabric.
Then to develop fabric with ease. 
So what we say is: you are aiming for ease of development in these lines but you are working harder... so the purpose is missed.
The rest of the PR you fulfill the main purpose.
TS is a tool, not a purpose on its own

ShaMan123 (commented):
Wow what a difference once you split the files up! A lot more readable

src/typedefs.ts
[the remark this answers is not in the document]
ShaMan123: and perhaps add `as const` to the bounds
```
  const result = [
    new Point(Math.min(...bounds[0]), Math.min(...bounds[1])),
    new Point(Math.max(...bounds[0]), Math.max(...bounds[1])),
  ] as const;
```

asturur:
once the functions are split up in their own file, we can keep them in and document where those are. let me see where are we now

asturur:
Even exacluding the typecheck regexes i get
```
-rw-r--r--  1 abogazzi  staff  312885 Mar 19 14:30 dist/index.min.js
```

Now bundle aside i want to make a real point.
Regexes with the /g/ (global) flag have a special behaviour with test and exec.
They restart from the last match.
We build N regexp with global state but then we use ust one of them.
Now i think is more proficuos if we just create the regexp when needed and we keep in our file the source only.
What do you think? i can make that change if you agree.

Example, all of those becomes

```diff
- export const reArcCommand = new RegExp(
+ export const reArcCommand = 
  String.raw`(A) (?:${p} ${p} ${p} ([01]) ?([01]) ${p} ${p} ?)+`,
- 'gi'
- );
```

 and in the path code, in the only place where we use

```js
  for (const match of pathString.matchAll(rePathCommand)) {
```

we do
```js
  const rePathCommandExp = new Regexp(rePathCommand)
  for (const match of pathString.matchAll(rePathCommandExp)) {
```

thoughts?

asturur:
The only real blocker so far is that node 14 doens't like `dgi` and so we need to find an alternative.
It doesn't seem for the main regex is an issue, so i removed in favor of gi, but we are still using `d` in one place.

Lazauya:
> What do you think? i can make that change if you agree.

I think that makes a lot of sense and I agree with the change. I'll look into `d` alternatives today.

src/util/path/index.ts
```diff
@@ -755,81 +866,76 @@ export const getPointOnPath = (
...
-      const parsed = parseFloat(coords[j]);
-      if (!isNaN(parsed)) {
-        coordsParsed.push(parsed);
+  const res: TComplexPathData = [];
+  for (const match of pathString.matchAll(rePathCommand)) {
+    let matchStr = match[0];
+    const chain: TComplexPathData = [];
+    let paramArr: RegExpExecArray | null;
+    do {
+      paramArr = new RegExp(rePathCommand.source, 'id').exec(matchStr);
```
asturur, line 884: ok let's focus on this code for a second.
match and paramArr are always identical to me.
Can you explai why we regex twice?

Lazauya: on line 909 we change `matchStr` by removing the command we just found. This is to account for chained commands.

src/util/path/index.ts
```diff
@@ -755,81 +866,76 @@ export const getPointOnPath = (
...
+      });
+      chain.push(command as any);
+      // chop off that last set of params
+      if (filteredIndices.length <= 1) {
+        break;
       }
-    } else {
-      result.push(coordsParsed);
-    }
+      matchStr = matchStr.substring(0, filteredIndices[1][0]);
```
asturur, line 909: what is this doing? to me is keeping the first part of the matchString, so basicaly the command like 'Q', 'M'...
But then what are we doing with it?
Why do we think we need this?

Has anything to do with the fact that those regex return always an array of 40 items to me, and i have no idea why?
Is it like that if the string is long we need more parses?

Lazauya: It's for chained commands

src/util/path/path.ts
asturur: Yes in general i don't like the idea of changing code because TS doesn't like it.
TS main purpose is to help avoid errors, it helps with discovery too, but we shouldn't restrict JS at what TS understand, at least i didn't sign up for that when i decided to try a conversion.

src/util/path/index.ts
asturur: oh OK because `L 10 23 23 32 32 32 12 32 32 12 32 12 32 21 32 112 32` it gets parsed initially as a single command

Lazauya: Correct. I'm going to resolve these convos, unreslove them in necessary

src/util/path/path.ts
Lazauya: IDK I'm obviously biased but I still think that the current method is easier to follow. Also I'd like to point out that the refactored function is the same number of lines.

ShaMan123: Rejoining
I do prefer reconstructing over mutation if nothing is damaging
And here it is much rradable IMO

src/typedefs.ts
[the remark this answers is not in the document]
Lazauya: I went ahead and re did the type for the bounds to `TRectBounds = [min: IPoint, max: IPoint]`. I was going to remove it completely but then I realized that I'd need to define it twice (once for the `boundsOfCurve` in the cache and once for the function return signature) which is just bad practice. I think this is probably the best compromise.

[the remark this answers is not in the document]
Lazauya: I'm going to close this convo. Feel free to reopen if necessary

src/util/path/path.ts
Lazauya: So I went ahead and made it use the vars instead of `current[n]` where possible. I'm still adamant that I think this refactoring is better than doing casting, especially since it's the same number of lines. I agree that sometimes TS is overbearing but I think it's perfectly reasonable in THIS particular case

src/util/path/path.ts
Lazauya: I went ahead and changed it to non-spread since I realized I was already doing a cast, derp. It's actually less lines than before, so I'm going to resolve this convo.

src/util/path/path.ts
Lazauya: OK I went ahead and changed it. I realized I wasn't thinking when I was doing this one lol, it's much simpler now.

src/util/path/path.ts
Lazauya: So I think this is the last open conversation. Is this a blocker? Or are we good to go?

src/typedefs.ts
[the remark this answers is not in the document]
ShaMan123: I an live with this now though if you use:


> and perhaps add `as const` to the bounds
> 
> ```
>   const result = [
>     new Point(Math.min(...bounds[0]), Math.min(...bounds[1])),
>     new Point(Math.max(...bounds[0]), Math.max(...bounds[1])),
>   ] as const;
> ```

you don't need to define the return type or `result` at all and you are left with defining boundsOfCurves only.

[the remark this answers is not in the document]
Lazauya: Sorry for the delay, I finally had time to revisit this. When you do `as const` it infers the type as `readonly [Point, Point]` not `[IPoint, IPoint]` like we actually want. I could change the signature to `readonly [Point, Point]` but that's unnecessarily strict. I checked the code and it being a `Point` specifically is not relevant, since the `Point` methods aren't used, so `IPoint` is definitely the right type in this case. I'm not going to touch it but if it irks you I don't actually have a strong opinion so feel free to change it. I just want to be done with this PR, haha.

Lazauya:
> @Lazauya let me know when you think the code is done, that i will give another read.

@asturur I've looked at everything and I think it's ready, feel free to read over again.

src/typedefs.ts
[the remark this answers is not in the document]
ShaMan123: it is good enough

.codesandbox/templates/next/pages/index.tsx
ShaMan123, line 1: I would love hearing about CD/DX, how is the app? Was it easy to set up and use?

asturur:
@Lazauya i had a busy week, i ll get back here asap.
But the PR looks good

asturur (approved):

PR #8802, opened 2023-03-23
Add a way to start an animation at any arbitrary time

Ivanca:
Adding an argument called startOffset so you can start an animation at any given percentage (so not always at the beguining), this way the users of this library can do stuff like kill an animation and later on re-create it where it left (and the easing function will continue right where it left, instead of wrongly assuming t=0)

### DIFF
--- a/src/util/animation/AnimationBase.ts
+++ b/src/util/animation/AnimationBase.ts
@@ -18,6 +18,7 @@ export abstract class AnimationBase<
   declare readonly startValue: T;
   declare readonly endValue: T;
   declare readonly duration: number;
+  declare readonly startOffset: number;
   declare readonly delay: number;
 
   protected declare readonly byValue: T;
@@ -58,6 +59,7 @@ export abstract class AnimationBase<
     byValue,
     duration = 500,
     delay = 0,
+    startOffset = 0,
     easing = defaultEasing,
     onStart = noop,
     onChange = noop,
@@ -69,6 +71,7 @@ export abstract class AnimationBase<
 
     this.duration = duration;
     this.delay = delay;
+    this.startOffset = startOffset;
     this.easing = easing;
     this._onStart = onStart;
     this._onChange = onChange;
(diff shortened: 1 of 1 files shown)


ShaMan123:
Why not?
```ts
const { value, endValue, durationProgress, duration } = anim.abort()
const resumed = animate({ startValue:value, endValue, duration: duration * (1 - durationProgress) })
```

ShaMan123:
I prefer adding a pause/resume than an ambiguous `startOffset`

Ivanca:
There are many uses for starting at arbitrary positions than just pausing and resuming, such as going back one second in time in an animation; my opinion is that both things are pretty useful; meaning that there should be a way to set a starting point and also a way to call _pause()_ and _resume()_ 

> Why not?
> 
> ```ts
> const { value, endValue, durationProgress, duration } = anim.abort()
> const resumed = animate({ startValue: value, endValue, duration: duration * (1 - durationProgress) })
> ```

This would not work in any complex easing function, for example an easing function that bounces 3 times would wrongly bounce again 3 times instead of just continiung where it left.

ShaMan123:
Good point.
So I suggest adding pause/resume methods

ShaMan123:
Going back in time and such, hmm. Didn't consider that actually.
There is a PR dedicated to declarative object animation state https://github.com/fabricjs/fabric.js/pull/6390. Is that what you are looking for?

ShaMan123:
I now understand what you want to achieve. 
I don't like the naming `startOffset`.
If I understand correctly it only affects easing as you mentioned.
So in my example in the previous comment if I could pass the original start time to the easing function that would have been correct, right?

So I can revise

```ts
const { startValue, endValue, durationProgress, duration } = anim.abort()
const resumed = animate({ startValue, endValue, duration, onStart(){ this.startTime = duration * durationProgress } })
```

I don't have a strong opinion about this. 
Add tests under `test/unit/animation.js`.

asturur:
Sorry but what is today the difference between startOffset and delay?

Ivanca:
@asturur  Here startOffset means the opposite of delay, it would be like giving a negative delay (but fabric doesn't support that)

Ivanca:
I don't have the time to fix this PR so anyone who wants to take over is free to do so; in my project I'm using an old version of fabric hastily modified to fit my needs instead, but yeah, my dream animation api for fabricjs includes "pause(), resume(), restart(), restart([startTime])", anything like that or close enough would be great. Thank you all.

ShaMan123:
We all want fabric moving forward and must dedicate time for that to happen.
Such is the nature of open source.
Until then closing.

ShaMan123:
@asturur this is kind of like your proposal of declarative animation start but as part of the animation api

PR #8806, opened 2023-03-25
rename(): `IPoint` => `XY`

ShaMan123:
## Motivation

IPoint is a bad name

## Description

A non brainer rename, nothing else

## Gist

Thankfully VSCODE is a reliable tool

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- rename(): `IPoint` => `XY` [#8806](https://github.com/fabricjs/fabric.js/pull/8806)
 - perf(): optimize `perPixelTargetFind` [#8770](https://github.com/fabricjs/fabric.js/pull/8770)
 - BREAKING fix(): reflect NUM_FRACTION_DIGITS to SVG path data [#8782] (https://github.com/fabricjs/fabric.js/pull/8782)
 - fix(IText): layout change regression caused by #8663 (`text` was changed but layout was skipped) [#8711](https://github.com/fabricjs/fabric.js/pull/8711)
--- a/src/Point.ts
+++ b/src/Point.ts
@@ -2,7 +2,7 @@ import { TMat2D, TRadian } from './typedefs';
 import { cos } from './util/misc/cos';
 import { sin } from './util/misc/sin';
 
-export interface IPoint {
+export interface XY {
   x: number;
   y: number;
 }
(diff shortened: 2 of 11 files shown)


asturur:
I have still to understand why iPoint is bad, interface for point.

asturur (approved):

asturur:
at some point XY will become IXY if we keep consistency where I are interfaces and T are types.

If you have 5 minutes, make an effort and explain why IPoint is bad.

ShaMan123:
I think that as a dev using fabric it is unnscessary overhead to have types starting with T and interface starting with I when all I want is to find the type easily and intuitively by guessing the name and autocomplete. I don't know if it is an interface or type and I don't want to know because I don't care. It is annoying to find it like that and opens up discussions like "should this type be an interface or a type?", which is a waste of time and effort. We just want to get it done.
IPoint is a name from the definately typed package, hitorical remains. 
XY is a good name, self explanatory. All names should be like that.
As I mentioned previously, I don't think prefixing with T is a good idea as well for the reasons I just stated. But if we must lets stick to a single prefix for all types.

ShaMan123:
I think that as a dev using fabric it is unnscessary overhead to have types starting with T and interface starting with I when all I want is to find the type easily and intuitively by guessing the name and autocomplete. I don't know if it is an interface or type and I don't want to know because I don't care. It is annoying to find it like that and opens up discussions like "should this type be an interface or a type?", which is a waste of time and effort. We just want to get it done.
IPoint is a name from the definately typed package, hitorical remains. 
XY is a good name, self explanatory. All names should be like that.
As I mentioned previously, I don't think prefixing with T is a good idea as well for the reasons I just stated. But if we must lets stick to a single prefix for all types.

ShaMan123:
But TXY is silly and not self explanatory

asturur:
So your answer is not that IPoint was a bad name, but that you don't like the prefixing with I for interfaces and so since Point was already taken, for the class itself, XY is an alternative.

Regarding this:
```
I think that as a dev using fabric it is unnscessary overhead to have types starting with T and interface starting with I when all I want is to find the type easily and intuitively by guessing the name and autocomplete. I don't know if it is an interface or type and I don't want to know because I don't care. It is annoying to find it like that and opens up discussions like "should this type be an interface or a type?", which is a waste of time and effort. We just want to get it done.
```
Guessing the name and autocomplete is maybe your way to finding things that is fine. pressing T and knowing that most of the auto complete will be a type is another good way to do things.
The fact that you don't care if is an interfact or a type it means that either TS is badly done or that this is a personal preference.

ShaMan123:
True, I don't understand the difference between type and interface. I do understand that interface defines an API but I think type can be used fir that as well.

Pressing T is a way to find types, absolutely. That is why I am against I. I don't think it is a good idea to press T and then press I. I won't be sure I didn't miss what I am looking for if there are 2 places to lookup instead of 1. Or if the type exists or not.
Another option is to export types as a named export. I prefer that. Then we don't need a prefix, we import **types** from fabric. 
What do you think?

PR #8807, opened 2023-03-26
chore(TS): Group types

ShaMan123:
## Description

types, props and some patching


ShaMan123 (commented):
ready

src/shapes/Group.ts
```diff
@@ -821,8 +840,8 @@ export class Group extends createCollectionMixin(FabricObject<GroupEvents>) {
         strokeWidth: 0,
       }),
       bboxSizeAfter = this._getTransformedDimensions({
-        width: bbox.width,
-        height: bbox.height,
+        width: w,
+        height: h,
```
ShaMan123, line 844: might be this is an unwanted change?
bbox.width/height could have been undefined if group has no objects and then `_getTransformedDimensions` would use this.width/height

ShaMan123 (commented):
1 thing

src/shapes/Group.ts
asturur: i don't even know how we eneded up pass width and height to `_getTransformedDimensions` , it accepted skewX and skeW for the only reason to cheat during skew control support, but shouldn't take anything to be honest.

Regarding your question:

```
const width = hasWidth ? this.width : w,
      height = hasHeight ? this.height : h,
```
would using width and height local vars effectively sort the issue?
If bbox does not return width/height ( but why???) it would have this.width and this.height.

ShaMan123: In #8767 I changed this piece of logic. And I want a bit more so we shouldn't dig into this here IMO.
The only reason it wouldn't return is if group has no children.
Really, this part is not good enough in general.
width/height from props should take affect only if the layout type is `fixed`. All other conditions should ignore it.
The center point however is affected by left/top for all cases.
What I want to do is first set the center point and after that preform the initial layout.
I think we can move forward with this even if we commit a regression since obviously we need to look into this anyways and the BBox approach should reduce the layout logic by half and enable 10 times of options.

asturur: Ok so i ll remove this change here.

asturur (approved):

asturur:
Groups has still a bunch of type violations all around the file, i ll put back the `// @ts-nocheck` directive.

PR #8816, opened 2023-04-02
typo(): percision => precision

timgates42:
There is a small typo in test/unit/util.js.

Should read `precision` rather than `percision`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


ShaMan123 (approved):
You should consider creating a github action out of your repo
Is that the only typo we have? Did you run your tool on fabric?

PR #8817, opened 2023-04-02
fix(): move blocking browser logic to env

ShaMan123:
## Motivation

#8774 

## Description

move browser logic to env/index file

## Changes

- `createCanvasElement` + `createImage` moved to env
- renamed `createImage` to `createImageElement`
- removed `util/misc/dom`


src/shapes/Text/Text.ts
```diff
@@ -1278,13 +1278,11 @@ export class Text<
    * @return {CanvasPattern} a pattern to use as fill/stroke style
    */
   _applyPatternGradientTransformText(filler: TFiller) {
-    const pCanvas = createCanvasElement(),
-      // TODO: verify compatibility with strokeUniform
-      width = this.width + this.strokeWidth,
+    // TODO: verify compatibility with strokeUniform
+    const width = this.width + this.strokeWidth,
```
ShaMan123, line 1282: I am sure this is incorrect but we have a different approach in the pipeline anyways

ShaMan123 (commented):
progress

ShaMan123:
fabric isn't ready for this

PR #8820, opened 2023-04-04
fix(): `stylesToArray` edge case

ShaMan123:
## Motivation

This edge case came to my knowledge thanks to @jiayihu

## Description

A simple fix and test addition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): text styles [#8820](https://github.com/fabricjs/fabric.js/pull/8820)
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
@asturur this looks ready
I am beginning to think that all text style logic should be a standalone class **detached** from the text instance. Less noise on the already noisy Text class and more scoping is a true benefit.
Since you wish to release I will await a green light

ShaMan123:
@melchiar since this is your baby....

jiayihu:
LGTM, the unit tests is exactly what I remember and I trust the unit test more than my memory 👍🏼

Issue #8588, opened 2023-01-10
[Feature]: _fireSelectionEvents easy performance win


SLKnutson:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have searched and referenced existing issues, feature requests and discussions
- [X] I am filing a **FEATURE** request.

### Description

While profiling, I noticed there is an easy performance win within `_fireSelectionEvents`. This function shows up in profiling when you are selecting/deselecting a large number of objects (low 1000s). Right now, it determines what is added and what is removed by looping each array and checking if the object exists in the other array. If the existence check was against a set instead of searching the whole list, the performance would be improved, especially when you have large numbers of objects.

I don't have time to do a PR, but I think it would look something like this:
```
let somethingChanged = false,
  invalidate = false;
const objects = this.getActiveObjects(),
  added: FabricObject[] = [],
  removed: FabricObject[] = [];
const objectsSet = new Set(objects);
const oldObjectsSet = new Set(oldObjects);

oldObjects.forEach((target) => {
  if (!objectsSet.has(target)) {
    somethingChanged = true;
    target.fire('deselected', {
      e,
      target,
    });
    removed.push(target);
  }
});

objects.forEach((target) => {
  if (!oldObjectsSet.has(target)) {
    somethingChanged = true;
    target.fire('selected', {
      e,
      target,
    });
    added.push(target);
  }
});

```

### Current State

Works, but slow for large numbers of objects.
Thanks!

ShaMan123:
> I don't have time to do a PR, but I think it would look something like this:

Hope you find the time, seems good

proke03:
In my opinion, it seems to be an invalid method due to the overhead caused by the operation new Set(arr).

PR #8822, opened 2023-04-05
feat: improve _fireSelectionEvents performance

proke03:
## Motivation
#8588 

## Changes
Reduce two forEach to one for

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -1335,9 +1335,11 @@ export class SelectableCanvas<
     const objects = this.getActiveObjects(),
       added: FabricObject[] = [],
       removed: FabricObject[] = [];
+    const objectsSet = new Set(objects);
+    const oldObjectsSet = new Set(oldObjects);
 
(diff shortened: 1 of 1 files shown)


ShaMan123:
@proke03 thanks for PRing
However, I am not sure the change is worth the trouble. Does it optimize performance? You need to prove it.
Because it makes the code less readable IMO it must have a gain.
The only case this PR has an impact is if oldObjects and objects have many shared objects (then the code loops almost half of what it loops currently), which doesn't really happen in any case.
I would try a different approach if this is a pressing matter, or prove me wrong.

proke03:
Thank you for your review. I think you're right.
By the way, there is an error in the part that I did not modify. Do you know the cause? It seems to be the cause of the build failure.

캡처_2023_04_05_13_10_19_804

ShaMan123:
> By the way, there is an error in the part that I did not modify. Do you know the cause? It seems to be the cause of the build failure.

#8756

proke03:
Thank you!

Issue #8852, opened 2023-04-21
[BUG]: Can't parse SVG that was working with 5.3 but fails with 6.0


asturur:
Svg string:

```xml
[...]

PR #8853, opened 2023-04-22
fix(Path) Issue 8852 fix a bug in path parsing

asturur:
## Motivation

By using fabric 6 i found a non parseable path.
The error seems to come from how we reorganized the code creating an array from `new Array(results)` but sometimes results is NaN.
Since the case that can make results NaN is a rx or ry set to 0, i added a shortcut on top o the function to avoid parsing in that case.

The old code would still run the function but then output an empty array.

The second bug is present in 5.3 too, and is because sometimes the code to render special gradients that gets transformed in patterns can generate canvases that are smallre than 1 px. This would eventually round them to 0 and error out.
This fix makes the cnavas size at least 1.


asturur:
closes #8852

src/util/path/index.ts
```diff
@@ -94,6 +94,9 @@ const arcToSegments = (
   sweep: number,
   rotateX: TRadian
 ): TParsedAbsoluteCubicCurveCommand[] => {
+  if (rx === 0 || ry === 0) {
```
ShaMan123, line 97: Just so I understand...
What does rx/ry stand for in arc?

Issue #8868, opened 2023-04-28
chore(Types): export types


ShaMan123:
@asturur we need to address this
The community is starting to complain and use types in a wrong way via VSCODE importing scheme.
As mentioned before, I suggest to remove all the T prefixes and export all types as a named export
`import { types } from 'fabric'`
What do you think?

ShaMan123:
Of course we can leave the T prefix for now.
The main thing is the exporting.
No types are currently exported. 
Why BTW?

asturur:
aren't type exported automatically from the build, all the one that are used in the functions?

asturur:
Or link the complains if the information is there.
I gave a quick look at newer discussion but didn't find much

asturur:
> Of course we can leave the T prefix discussion for now. The main thing is the exporting. No types are currently exported. Why BTW?

I think the WHY is becuase we discoverd doing.
We didn't have a TS long term plan to act upon.
As far as i can see we generate a fabric.d.ts file that exports all the types and package.json knows about it. So i m not sure we are not exporting them.

You can't export types from a generated JS file because i don't think such thing exists.

gloriousjob:
Thanks, @ShaMan123! This is addressing the concern I had.

So I wouldnt' say no types are exported. For example, in my subclass example, import { Rect } from 'fabric' works. However, I can't get to the other exports inside of Rect to use the parts in my subclass.

The reason for this issue appears to be because index.d.ts references fabric.d.ts, which only has a subset of items (e.g., Rect is listed in there but not the other exports in Rect). These two files are basically copies of their normal counterparts (index.ts and fabric.ts). These aren't autogenerated.

ShaMan123:
Notice that no types are exported from the index files, that is what I mean
I will PR soon

bladerunner2020:
Any update on this? We need other types (like GroupProps).

PR #8875, opened 2023-04-29
chore(TS): Patterns types

ShaMan123:
## Motivation

Follow up from an issue described in #8868 

## Description

Exporting missing types from classes.
Apparently is not enough to have exported types from files for typescript to make them available, they need to be exported from the package interface.

## Changes

> * extracted types and renamed to match pattern
> * new export pattern
> * `Abortable` type (separate commit, can extract if needed)


ShaMan123 (commented):
@asturur please review the pattern
I suggest an index file is to expose the class (Pattern) and the types as a named export (Types or PatternTypes).
Then we reexport in `fabric.ts`

- extracted types and renamed to match pattern
- new export pattern
- `Abortable` type (separate commit, can extract if needed)

Thoughts?

ShaMan123:
If you agree I can apply the pattern on the repo

ShaMan123:
@tushuhei I am interested in some feedback regarding importing from fabric
What do you think of
```
import { Pattern, PatternTypes } from 'fabric'
```
OR
```
import { Pattern, PatternProps, OtherType } from 'fabric'
```
OR
```
import { Pattern } from 'fabric'
import { PatternProps, OtherType } from 'fabric/types'
```
Do you have other suggestions?

ShaMan123:
@asturur @melchiar take a look at the prev comment

src/shapes/Object/Object.ts
```diff
@@ -1518,15 +1519,11 @@ export class FabricObject<
    * @param {object} object
    * @param {object} [options]
    * @param {string} [options.extraParam] property to pass as first argument to the constructor
-   * @param {AbortSignal} [options.signal] handle aborting, see https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal
```
asturur, line 1521: this is still valid

src/Pattern/Pattern.ts
```diff
@@ -111,15 +85,15 @@ export class Pattern {
    * @param {Object} [options] Options object
    * @param {option.source} [source] the pattern source, eventually empty or a drawable
    */
-  constructor(options: TPatternOptions = {}) {
+  constructor(options: PatternProps = {}) {
```
asturur, line 88: in general i really dislike the word `props` as abbreviation of properties.
Not sure if is because react js popularity or something else, but those are options.

asturur:
you likely didn't touch the code. where the extra 52 bytes come from? how this stuff works

> @tushuhei I am interested in some feedback regarding importing from fabric What do you think of
> 
> ```
> import { Pattern, PatternTypes } from 'fabric'
> let foo: PatternTypes.PatternProps
> ```
> 
> OR
> 
> ```
> import { Pattern, PatternProps, OtherType } from 'fabric'
> ```
> 
> OR
> 
> ```
> import { Pattern } from 'fabric'
> import { PatternProps, OtherType } from 'fabric/types'
> ```
> 
> Do you have other suggestions?

I think the only issue is discovery.
How do i get to know all the things i can import and what they mean.
How do other large TS base libraries solve this? Do we have any good example to look at? pixiJS? React?

ShaMan123:
None prefix types
All export along side the rest of the code
https://github.com/pixijs/pixijs/blob/dev/packages/canvas-renderer/src/index.ts

ShaMan123:
This is what I wanted from the first place

tushuhei:
I personally like

```
import { Pattern } from 'fabric'
import { PatternProps, OtherType } from 'fabric/types'
```

because it infers that I can do type-only import for the latter. (i.e. `import type { PatternProps, OtherType } from 'fabric/types'`)

Regarding discovery, probably Three.js might be a good reference. When I work on Three.js projects, I read the documentation or just remember important class names, then I just let VScode autocomplete the import statements as I instantiate objects from those classes.

asturur:
> None prefix types 
I will solve this here, and i really don't want any answer because at some point i get tired too being human.
I wanted something that namespaced the types, and i picked up a prefix. I made clear that i wanted a pattern that made easy  divide classes from types. A pattern like any other and i picked up a T. Some people prefix constants with K or use a different casing for constants, those are just patterns. You made it clear that you didn't like it, and you continued pushing for removing it. `don't like` can't be a metric. You also said we can leave the discussion of T for later but regardless you renamed some types in this PR to remove the T.
Fine remove the T i don't care anymore and i don't want to talk about it.
The fault at the end is mine becuase i could have just closed PRs that came with types without T but i didn't do it.
I'm weak.

@tushuhei has an alternative, to namespace the types with `fabric/types`. Not sure what it takes but maybe we can try. Is probably just a package.json gimmick, i don't think we need a different index or something like that

@tushuhei can you point us to a library that does that so we can look at how is achieved?

asturur:
so the extra bytes come from this:
```
var types = /*#__PURE__*/Object.freeze({
  __proto__: null
});
```

Some side effect of bundling coming from the export * from that we need to find a way to avoid or they will build up with all the files that will use this pattern.

ShaMan123:
I suggested that pattern of fabric/types.
I am suggesting options.
The T can remain. I trnamed same as ObjectProps.
All this PR is trying to do is come up with a pattern.
That is why I didn't keep on, but stopped so we dicuss.
I will back off about the T prefix once we have a pattern of exporting types.
All in all I just want to promote exporting types because the community is starting to us imports via vscode in a WRONG way.
And I need it as well.
This is a matter of DX, that is all.
You are welcome to commit, revert, close, whatever.

asturur:
> I suggested that pattern of fabric/types. I am suggesting options. The T can remain. I trnamed same as ObjectProps. All this PR is trying to do is come up with a pattern. That is why I didn't keep on, but stopped so we dicuss. I will back off about the T prefix once we have a pattern of exporting types. All in all I just want to promote exporting types because the community is starting to us imports via vscode in a WRONG way. And I need it as well. This is a matter of DX, that is all. You are welcome to commit, revert, close, whatever.

No no, as i said i m past willing to discuss the T anymore. i got tired.

I didn't see it was one of the 3 options you proposed, i just thought it came from @tushuhei.

Still i don't understand what is the wrong versus good way to import types, this has been said twice, but which is it?
Also what does it mean imports via vscode? you mean just in general using the wrong imports or there is some special tool in vscode that let you reach missing types?

asturur:
ok using `export type` is good enough to not generate extra objects
image

The pattern is fine by me, between the 3 i would prefer to namespace the types with something, so fabric/types would work for me.

I m not sure how it would work. Did you seen it used somewhere?

ShaMan123:
I didn't see it been done. No.
It is the same concept as `@types/fabric`.
The only reason I am suggesting this (and taking part in this discussion) is because you want a mechanism to distinguish between importing an object and a type. It is not something I require/want/need or think is a good idea. More loss than gain IMO if it is because of maintaining index files, forgetting to export types and the fact that it is fairly easy to distinguish between the two via vscode (show a different icon for interfaces, types and classes have the same unfortunately) or whatever. But it is not a blocker for me. However I do want you to consider my points.
I am fine with `fabric/types`, depending on discovery (will vscode be able to import it automatically?). It should be fairly simple to set up with an additional index file exporting types (I don't see a different way). 
Mind that moving in this direction makes the T prefix redundant in this context. I have an entry point for types only and everything is prefixed with a T. I am repeating myself and probably annoying us both because I think you will regret this one day, as you talk about `strokeUniform`. Types can be prefixed in the codebase and re-exported without it. Is that something you are willing to consider?
Anyways, at the end I just want fabric to move forward without casualties from our side (while being rather future proof) so please forgive me if I wronged you.



> Still i don't understand what is the wrong versus good way to import types, this has been said twice, but which is it?
> Also what does it mean imports via vscode? you mean just in general using the wrong imports or there is some special tool in vscode that let you reach missing types?

VSCODE looks for exports not only in the official exported files (so it seems, no special tool). So it crawls fabric and finds the type somewhere and imports it directly from that file, which causes a TS error because we did not declare the path as an entry point in `package.json` (and we shouldn't because of build). Bad DX. Nothing more, unless someone starts importing actual classes from those files.
I don't know what is good and bad. I know what is common nowadays, so that is what guides me, because we want fabric to be top notch and follow standards (that are not defined of course). There are many libraries that have several entry points but I didn't see types being one of them. It doesn't seem too extreme to do that.


so options is better?
fine with me

> so the extra bytes come from this:
> 
> ```
> var types = /*#__PURE__*/Object.freeze({
>   __proto__: null
> });
> ```
> 
> Some side effect of bundling coming from the export * from that we need to find a way to avoid or they will build up with all the files that will use this pattern.

looks like a bug in rollup to me

asturur:
So when you say `All in all I just want to promote exporting types because the community is starting to us imports via vscode in a WRONG way.` you intend that they are able to reach types import because vscode scrapes folders that shouldn't.

Regarding the extra code i m not sure if rollup has a bug or is extremely on spec, it doesn't matter since using export type is also more clear and solve the issue.

> have an entry point for types only and everything is prefixed with a T. I am repeating myself and probably annoying us both because I think you will regret this one day, as you talk about `strokeUniform`. Types can be prefixed in the codebase and re-exported without it. Is that something you are willing to consider? 

Do you know that you don't want to add a T but you are proposing to do PatternTypes.type1, PatternTypes.type2, RectTypes.type1, RectType.type2 ... and so on? Is way uglier, longer, and inside codebase you don't even use the same names that you use outside. Seems the worst option to me.

So all the types are reacheable but directly from the files where those are created, and that doesn't seem right.

creating PatternTypes as a container with types has the downside that you have to know which types belong to who before finding them and the usual breakage if you have to move a type because becomes generic.

Isn't better to just list the types one by one ? That would also force us to not use duplicates names and seems like the less risky and more normal pattern.

ShaMan123:
Agreed. Was an option I wanted to look at.
I am more than fine to close this PR.
The discussion was what I was after, agreeing and shaping type exports is the import thing.

ShaMan123:
I suggest you PR this as you see now and we move from there. Or reshape this PR.

asturur:
The patterns are just those 3.
All considered this seems the only one that is practical:

```
import { Pattern, PatternProps, OtherType } from 'fabric'
```

asturur:
Apart the obligations we have from work, this kind of effort seems a blocker for release and should take precedence over others.
Seems to me that 

```
import { Pattern, PatternProps, OtherType } from 'fabric'
```
is the only approach that doesn't require exploring new patterns or new tools.

What do we want to do?

ShaMan123:
I agree it is the major blocker
Let's move forward

```ts
import { Pattern, PatternProps, OtherType } from 'fabric'
```

Is fine.
Where are you with the T prefix? So I know, it is unclear by your example.

asturur:
i don't need to talk anymore about the T prefix. I got angry i scraped it from my list and i said i don't want to talk about it anymore, indeed in my example there is not T.

So how can we divide the effort? if we do one file per class, being sure we just export what we have right now and we don't add new code just because we are there, should we be able to create quick fast PRs?

Do we want to do something like

```
export { Pattern } from './pattern'
expor type { PatternProps, OtherType } from './pattern
```
rigorously one after the other so that is clear what is the JS interface at a glance?

ShaMan123:
I knew you will get angry 🫣
But it was unclear.
We don't have to add more index files if you are against

The export type directive is good practice so it seems.
Should be 1-2 prs and left overs.

asturur:
I wouldn't mind an index file for  files, but it seems we don't have a 100% reference to d that. do we ? I would like to do this following a manual rather than discovering

ShaMan123:
So let's close this pr and use pattern to create the pattern 🙃

ShaMan123:
Or modify it

fabric.ts
```diff
@@ -15,7 +15,7 @@ export { Intersection } from './src/Intersection';
 export { Color } from './src/color/Color';
 
 export { Gradient } from './src/gradient/Gradient';
-export { Pattern } from './src/Pattern';
+export * from './src/pattern';
```
ShaMan123: I prefer this because we have the pattern index file to manage exports

src/Pattern/index.ts
ShaMan123, line 1: github doesn't pickup I renamed the folder to lowercase pattern.
Can you see if after you pull it is correct?
`src/Pattern/index.ts` => `src/pattern/index.ts`

src/Pattern/Pattern.ts
ShaMan123: renamed to options

src/Pattern/index.ts
ShaMan123: https://github.com/fabricjs/fabric.js/actions/runs/4920171612/jobs/8788677597?pr=8875

ShaMan123: reverted, it is a bug on git/github

ShaMan123 (commented):
ready

src/shapes/Object/Object.ts
ShaMan123: a211c25cd

asturur (approved):

PR #8499, opened 2022-12-08
chore(TS): migrate Eraser

ShaMan123:
## Description

Migrate eraser and refactor to a module.
Adapt fabric entities so there is no mutation done (the mixin used to mutate the prototypes of Object, Group, Canvas)
Can be regarded as Eraser v3, see v2 #7470 for reference

## Changes

- comply EraserBrush to brush pattern
- perf(): doesn't need to do a full redraw of path each frame/step
- perf(): optimize `needsFullRender` - move alpha checking to PencilBrush and descendants not including eraser (I think we should try masking PencilBrush, same as the eraser does, but with `fillRect` of `color`, that will make redrawing the entire path redundant in case of opacity, am I right?)
- perf(PencilBrush): fixed straight line used to redraw until mouse up regardless of the straight line state
- fix(): erasing + straight line
- feat(): alpha erasing
- fix(): clipPath, changed rendering logic so that first the brush draws over top context, then it is masked and then it is used to clip the main context. Makes #7991 #7984 obsolete and `won't fix` and enhances perf
- feat(): `updating`, eraser will subscribe to canvas rendering and update its pattern 
- fix(): utilize `renderCanvas` to render pattern correctly (was missing out canvas#clipPath)

## Gist

- [ ] first merge #8476 https://github.com/fabricjs/fabric.js/compare/clip-brush...ts-eraser
- [x] add tests
   - [ ] no idea why `top_ctx` goldens are wrong... missing the opaic red rect that is visible when actually erasing
(actual) freedrawing_eraser_custom_stack_alpha_clipped_top_ctx
- [ ] toSVG tests
- [ ] opaic background destroys the eraser pattern... no idea why

## In Action

[Edit fabric-nextjs-sandbox (forked)](https://codesandbox.io/s/fabric-nextjs-sandbox-forked-njtsgy?fontsize=14&hidenavigation=1&module=%2Fpages%2Findex.tsx&theme=dark)

https://user-images.githubusercontent.com/34343793/206911085-06bad37d-801f-4cc6-8583-1c17ba6ac220.mp4

**With clip path**

https://user-images.githubusercontent.com/34343793/206911099-af39d7b9-3005-45d3-b061-5b32eff218da.mp4

PR #8876, opened 2023-04-29
chore(TS): Enable strictPropertyInitialization

tushuhei:
## Motivation
Nit. Enable `strictPropertyInitialization`, which should be true by default when the `strict` mode is on.


asturur:
This has unintend consequences, we went over back and forth on this and how the code gets transpiled is not compatible with what we do.

The issue is that we relegate full option settings to the base class.
Those properties declared as public properties in the class syntax end up being a constructor assignment.
if this happen outside the base class, those values are overwriting the options passed by the user.

What is this achieving and why on your side?

src/canvas/Canvas.ts
```diff
@@ -114,7 +114,7 @@ export class Canvas extends SelectableCanvas {
 
   declare currentSubTargets?: FabricObject[];
 
-  private _isClick: boolean;
+  private _isClick = false;
```
asturur, line 117: this line here actually is:
```
    _defineProperty(this, "_isClick", void 0);
```
with your change becomes `false` instead of void 0.

If the issue for you is this being undefine, just add `this._isClcik = false` in the constructor.

If that is not an issue what we want here is a declare.

I'm also not sure if the correct default value is false or true, since we seem to make at first event it a  `true` unless you change it to a `false` with a drag or a move.

tushuhei:
Thanks for taking a look. I haven't found the root causation, but for some reason my app's `tsc` gives build errors on theses lines even though I'm excluding node_modules and the skipLibCheck option is enabled.

Looking at the errors, it seemed that the required changes to make it comply with `strictPropertyInitialization` were not so bug, so I made this PR. I thought it's helpful for child projects to share the default tsconfig settings (i.e. tsc --init) with fabric.js anyway.

asturur:
Well you shouldn't get an error, it means we left out initializing it.
Either we put declare in front of it, or we add it in the constructor right away.

We are not using the public properties in classes for the fabric classes ( but we can do it for normal unrelated classes that run some logic )

so for example we can to do this:

```js
class FabricShape {

  label: string;

}
```

nor 

```js
class FabricShape {

  label: string = '';

}
```

but we have to do

```js
class FabricShape {

  declare label?: string;

  constructor() {
    
  }

}
```

or 

```js
class FabricShape {

  declare label: string;

  constructor() {
     // we are 100% sure label will be set, or we set it ourselves
     this.label = '';
  }

}
```

So if you are just getting a compilation error and the value of isClick doesn't bother you, just add a declare in front of it, but don't modify our ts configuration

tushuhei:
Thanks for the explanation. Yes, I agree it's more of a problem for an application-side. Let me close this PR for now.

tushuhei:
I found the causation. I was trying to use the eraser brush, and I was importing from like `../node_modules/fabric/src/mixins/eraser_brush.mixin`, which forced my `tsc` to check if `fabric` is complying with my project's `tsconfig.json`.
I also tried to import it from `fabric/src/mixins/eraser_brush.mixin` as follows, but it causes a compilation error because it's not exposed yet.
https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-es0d03?file=%2Fpages%2Findex.tsx

Probably this is related to #8499, but what would be the best way to import a eraser brush for v6? Is it still simply WIP?

PR #8877, opened 2023-04-29
refactor() BREAKING: align line to other classes for positioning

asturur:
## Motivation

Line has some peculiar code that could be removed in favor of standard methods we use everyewhere else.

## Description

Trying to remove the custom code for the standard setPostionByOrigin and makeBoundingBoxFromPoints, leaving open the issue with changes in positioning.

This won't affect old saved data but affect lines that are created without left,top and just points.
I would seem normal that if i have a line that extends from `5,15 to 15,5` it's center is in `10, 10` regardless of strokeWidth and that the left, top and originX or originY are assigned depending from the options or the defaults but not from the points themselves

Today you get different originX and originY depending on the order of your points, like if x1 or y1 are the origin, and unless you specified a different origin.

Make sense to remove those contraddictions but there are cases in which the lines aren't anymore in the previous places when initialized with the same points.

### DIFF
--- a/src/shapes/Line.ts
+++ b/src/shapes/Line.ts
@@ -103,7 +103,7 @@ export class Line<
   _set(key: string, value: any) {
     super._set(key, value);
     if (coordProps.includes(key as keyof UniqueLineProps)) {
-      // this doesn't make sense anymore, since setting x1 when top or left
+      // this doesn't make sense very much, since setting x1 when top or left
       // are already set, is just going to show a strange result since the
       // line will move way more than the developer expect
       this._setWidthHeight();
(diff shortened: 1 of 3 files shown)


ShaMan123:
I know you don't want more changes...
However, why not to completely remove this or make a subclass of Polyline, overriding `toObject/fromObject/fromElement`?

asturur:
Because you can already have a polyline with 2 points.
I don't like this class but i don't want to risk svg import now.
We would need to keep everything and what would go away is basically the render code for the line and nothing more.

asturur:
Ideally both polyon, polyline and line just extend Path limiting some functionality, but we need a smart way to handle the points handling afterwards.

asturur:
I m amazed that the more es6 i use, the more the bundle grow, i think is all safari fault.

ShaMan123:
It is really strange the stats from the last prs.
I don't pay enough attention to that.
Just see readable es6 code and smile...

asturur:
is just because every time we add a deconstruct in the arguments safari freaks out.
Is ok one day we can move to safari 13 and the bundle will shrink.

asturur:
you should see the template with parameters what does it do:
```js
    return ['<line ', 'COMMON_PARTS', "x1=\"".concat(x1, "\" y1=\"").concat(y1, "\" x2=\"").concat(x2, "\" y2=\"").concat(y2, "\" />\n")];

```

ShaMan123:
Web dev can just go smoothly even when things start to workout

ShaMan123:
> Web dev can just go smoothly even when things start to workout

Can't

src/shapes/Line.ts
```diff
@@ -70,24 +71,28 @@ export class Line<
...
-    this.left = left ?? this._getLeftToOriginX();
-    this.top = top ?? this._getTopToOriginY();
+  _setWidthHeight() {
+    const { x1, y1, x2, y2 } = this;
+    this.width = Math.abs(x2 - x1);
+    this.height = Math.abs(y2 - y1);
+    const { left, top, width, height } = makeBoundingBoxFromPoints([
+      { x: x1, y: y1 },
+      { x: x2, y: y2 },
+    ]);
```
ShaMan123, line 93: redundant calc of width/height
obtain from `makeBoundingBoxFromPoints`

PR #8878, opened 2023-04-30
chore(TS): Parse transform attribute typing

Lazauya:
Some minor typing to move towards a fully-typed parser.
Includes some minor logic changes, meaning simpler regex and smarter parsing.
Removed the superfluous `numberRegExStr` constant and spruced up the `reNum` type.


src/parser/parseTransformAttribute.ts
```diff
@@ -1,105 +1,50 @@
...
+const p = String.raw`(${reNum})`;
+const skewX = String.raw`(skewX)\(${p}\)`;
+const skewY = String.raw`(skewY)\(${p}\)`;
+const rotate = String.raw`(rotate)\(${p}(?: ${p} ${p})?\)`;
+const scale = String.raw`(scale)\(${p}(?: ${p})?\)`;
+const translate = String.raw`(translate)\(${p}(?: ${p})?\)`;
+const matrix = String.raw`(matrix)\(${p} ${p} ${p} ${p} ${p} ${p}\)`;
+const transform = String.raw`(?:${matrix}|${translate}|${rotate}|${scale}|${skewX}|${skewY})`;
+const transforms = String.raw`(?:${transform}*)`;
+const transformList = String.raw`^\s*(?:${transforms}?)\s*$`;
```
asturur, line 21: please don't use String.raw where is no need for it. 1( not sure where here )
Also to ease review, can you explain how you get rid of the commaWsp?

asturur:
Seems good so me apart that minor comment.

src/parser/parseTransformAttribute.ts
Lazauya: Similar to my other work on the path parser, I clean the string in the parse function to remove commas and arbitrary whitespace.

Lazauya:
OK @asturur should be ready now.

asturur (approved):

asturur:
Good to me, let's wait @ShaMan123 too.

src/parser/parseTransformAttribute.ts
```diff
@@ -1,105 +1,50 @@
...
+ * @return {TTransformMatrix} Array of 6 elements representing transformation matrix
  */
-export function parseTransformAttribute(attributeValue) {
+export function parseTransformAttribute(attributeValue: string): TMat2D {
+  // first we clean the string
+  attributeValue = attributeValue
+    .replace(new RegExp(`(${reNum})`, 'gi'), ' $1 ')
+    // replace annoying commas and arbitrary whitespace with single spaces
+    .replace(/,/gi, ' ')
+    .replace(/\s+/gi, ' ')
```
asturur, line 41: actually, sorry, this part here is duplicated in path. Do we want to do a function util, under utils/internals that clean up the string? so we can also test that the input and output of those 3 asserts combined are always what we want.

src/parser/parseTransformAttribute.ts
```diff
@@ -110,13 +55,14 @@ export function parseTransformAttribute(attributeValue) {
...
-      operation = m[1],
-      args = m.slice(2).map(parseFloat);
+  for (const match of attributeValue.matchAll(reTransform)) {
+    const transformMatch = new RegExp(transform).exec(match[0]);
+    if (!transformMatch) {
+      continue;
+    }
+    const matchedParams = transformMatch.filter((m) => !!m);
+    const operation = matchedParams[1];
+    const args = matchedParams.slice(2).map(parseFloat);
```
asturur, line 65:
```suggestion
    const [ _, operation, ...rawArgs ] = matchedParams;
```
Would this work?

then

```js
const args = rawArgs.map((arg) => parseFloat(arg));
```

There is a risk using parseFloat inside a map like that, that we are calling parseFloat as


```js
(value, index, wholeArray) => parseFloat(value, index, wholeArray)
```

And you never know what happens

src/parser/parseTransformAttribute.ts
```diff
@@ -110,13 +55,14 @@ export function parseTransformAttribute(attributeValue) {
...
 
-  attributeValue.replace(reTransform, function (match) {
-    const m = new RegExp(transform).exec(match).filter(function (match) {
-        // match !== '' && match != null
-        return !!match;
-      }),
-      operation = m[1],
-      args = m.slice(2).map(parseFloat);
+  for (const match of attributeValue.matchAll(reTransform)) {
+    const transformMatch = new RegExp(transform).exec(match[0]);
```
asturur, line 59: i never remember how regexp work, in this case can we create the regexp before the loop and just use it for each match[0]?

asturur:
Sorry i got some additional minor comments, feel free to comment if you think some are unnecessary

src/parser/parseTransformAttribute.ts
asturur: it can totally be called `removeExtraSpaceAndCommasFromSVGAttributes` it doesn't have to be generic or anything like that.

ShaMan123: If it is used more than once, sure

ShaMan123: Even if used once

src/parser/parseTransformAttribute.ts
ShaMan123: Dido.
It never does what you want it to.
And mdn says one thing while chrome does another from what I remember

src/parser/parseTransformAttribute.ts
ShaMan123: I agree
Risky  
Should not abbreviate

src/parser/parseTransformAttribute.ts
asturur: i don't know, it could be because we replace first the comma with a 'space' and only then we can clean up multiple spaces. Probably this is safer to follow and a single regexp is a bit more hard to read and be sure it does what it has to do.

src/parser/scaleMatrix.ts
```diff
@@ -1,6 +1,17 @@
...
+ * Takes form
+ * [x 0 0]
+ * [0 y 0]
+ * [0 0 1]
+ * For more info, see
+ * @link https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#scale
+ */
+type TMatScale = [x: number] | [x: number, y: number];
+
+export function scaleMatrix(matrix: TMat2D, args: TMatScale | number[]) {
```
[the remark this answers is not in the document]
asturur: So i think the code is very old and bad.
I would argue we fix it in a followup since @Lazauya made a work on types and regex and i think we can focus on that.
But yes, definetely this matrix loop is wrong, it has a mutable array that gets renewed evey loop with a new one, in a let. Is just bad code.

asturur:
@Lazauya i want to publish a new beta i m merging this and following up myself for the minor changes requested

PR #8881, opened 2023-05-04
chore(Path): Clean up from Lazuyia PR

asturur:
## Motivation

Just code cleanup
Function reuse, different es version code, no functional changes

### DIFF
--- a/src/parser/skewMatrix.ts
+++ b/src/parser/skewMatrix.ts
@@ -10,7 +10,7 @@ import { TDegree, TMat2D } from '../typedefs';
  * @link https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#skewx
  */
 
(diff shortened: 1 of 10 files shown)


src/parser/skewMatrix.ts
```diff
@@ -1,20 +1,47 @@
...
+ *
+ * [1 0 0]
+ * [y 1 0]
+ * [0 0 1]
+ *
  * For more info, see
  * @link https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform#skewx
  */
-type TMatSkew = [xy: number];
+const fromAngleToSkew = (angle: TDegree) => Math.tan(degreesToRadians(angle));
```
ShaMan123, line 19: @asturur this is a valid util, should belong to utils IMO
I will move it in #8893

PR #8884, opened 2023-05-04
chore(TS) BREAKING: svg parsing function are promises instead of using callbacks

asturur:
## Motivation

Fabric abandoned the callbacks in favor of promises or async functions, SVG parsing wasn't converted yet.

## Description

LoadSVGFromString and LoadSVGFromUrl and parseSVGDocument lose the callback argument and become an async function that return an object with 4 properties that containes the 4 arguments that the callback had.

## Changes

```js
LoadSVGFromString(string, (objects, options, elements, allElements) => {

}, reviver, options)
```
becomes
```
LoadSVGFromString(string, reviver, options).then(({objects, options, elements, allElements}) => {

})
```
or

```js
const {objects, options, elements, allElements} = await LoadSVGFromString(string, reviver, options);
```


asturur:
i hope to finish this in the weekend or monday

src/parser/parseSVGDocument.ts
```diff
@@ -62,8 +70,11 @@ export function parseSVGDocument(
     ); // http://www.w3.org/TR/SVG/struct.html#DefsElement
   });
   if (!elements || (elements && !elements.length)) {
-    callback([], {}, [], descendants);
-    return;
+    return {
+      ...emptyResponse,
```
ShaMan123, line 74: even spreading isn't safe

src/parser/parseSVGDocument.ts
```diff
@@ -74,22 +85,33 @@ export function parseSVGDocument(
         (el) => svgValidTagNamesRegEx.test(el.nodeName.replace('svg:', ''))
       );
     });
+
+  // thos are like globals we need to fix
```
ShaMan123, line 89: I see your comment now

ShaMan123: passing as a context should do the job w/o too much ink

src/parser/parseSVGDocument.ts
```diff
@@ -74,22 +85,33 @@ export function parseSVGDocument(
...
-    (instances, elements) => {
-      if (callback) {
-        callback(instances, options, elements, descendants);
-        delete gradientDefs[svgUid];
-        delete cssRules[svgUid];
-        delete clipPaths[svgUid];
-      }
-    },
-    { ...options },
+    options,
```
ShaMan123, line 97: was there are reason it was spread before?

src/shapes/Ellipse.ts
```diff
@@ -158,18 +158,14 @@ export class Ellipse<
...
-    element: SVGElement,
-    callback: (ellipse: Ellipse) => void
-  ) {
+  static async fromElement(element: SVGElement) {
     const parsedAttributes = parseAttributes(element, this.ATTRIBUTE_NAMES);
 
     parsedAttributes.left = (parsedAttributes.left || 0) - parsedAttributes.rx;
     parsedAttributes.top = (parsedAttributes.top || 0) - parsedAttributes.ry;
-    callback(new this(parsedAttributes));
+    return new this(parsedAttributes);
```
ShaMan123, line 168: prefer to spread here as done in Circle

src/shapes/Image.ts
```diff
@@ -824,16 +824,18 @@ export class Image<
...
   ) {
     const parsedAttributes = parseAttributes(element, this.ATTRIBUTE_NAMES);
-    this.fromURL(parsedAttributes['xlink:href'], {
+    return this.fromURL(parsedAttributes['xlink:href'], {
       ...options,
       ...parsedAttributes,
-    }).then(callback);
+    }).catch((err) => {
+      console.log(err);
+      return null;
```
ShaMan123, line 837: don't know what I think about silent errors

src/shapes/Image.ts
```diff
@@ -824,16 +824,18 @@ export class Image<
    * @param {AbortSignal} [options.signal] handle aborting, see https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal
    * @param {Function} callback Callback to execute when Image object is created
    */
-  static fromElement(
+  static async fromElement(
```
ShaMan123, line 827: no need for async here

src/shapes/Image.ts
ShaMan123: especially since the rest of the elements don't have a catch block

src/shapes/Path.ts
```diff
@@ -411,25 +411,18 @@ export class Path<
...
-   * @param {(path: Path) => void} callback Callback to invoke after the element has been parsed
    * @param {Partial<PathProps>} [options] Options object
    */
-  static fromElement(
-    element: SVGElement,
-    callback: (path: Path) => void,
-    options: Partial<PathProps>
-  ) {
+  static async fromElement(element: SVGElement, options: Partial<PathProps>) {
     const parsedAttributes = parseAttributes(element, this.ATTRIBUTE_NAMES);
```
ShaMan123, line 417: can we spread this so that `d` isn't passed into options?

src/shapes/Polyline.ts
```diff
@@ -323,31 +323,21 @@ export class Polyline<
...
    */
-  static fromElement(
-    element: SVGElement,
-    callback: (poly: Polyline | null) => any,
-    options?: any
-  ) {
-    if (!element) {
-      return callback(null);
-    }
+  static async fromElement(element: SVGElement, options?: any) {
```
ShaMan123, line 328:
```suggestion
  static async fromElement(element: SVGElement, options?: Partial<SerializedPolylineProps>) {
```

src/shapes/Polyline.ts
```diff
@@ -323,31 +323,21 @@ export class Polyline<
...
     const points = parsePointsAttribute(element.getAttribute('points')),
       // we omit left and top to instruct the constructor to position the object using the bbox
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       { left, top, ...parsedAttributes } = parseAttributes(
```
ShaMan123, line 332: same, spread out points so they are not passed to options

src/shapes/Rect.ts
```diff
@@ -223,7 +219,6 @@ export class Rect<
       height,
       visible: Boolean(visible && width && height),
```
ShaMan123, line 220: this is unusual

src/shapes/Text/Text.ts
```diff
@@ -1766,18 +1766,9 @@ export class Text<
...
-    options: object
-  ) {
-    if (!element) {
-      return callback(null);
-    }
-
+  static async fromElement(element: SVGElement, options: object) {
     const parsedAttributes = parseAttributes(element, Text.ATTRIBUTE_NAMES),
       parsedAnchor = parsedAttributes.textAnchor || 'left';
     options = Object.assign({}, options, parsedAttributes);
```
ShaMan123, line 1774: spread

test/unit/path.js
```diff
@@ -48,12 +48,16 @@
...
-    getPathObject('M 100 100 L 300 100 L 200 300 z', callback);
+    const path = new fabric.Path('M 100 100 L 300 100 L 200 300 z', {
+      fill: 'red',
+      stroke: 'blue',
+      strokeLineCap: 'butt',
+      strokeLineJoin: 'miter',
+      strokeMiterLimit: 4,
+      strokeWidth: 0,
+    });
+    callback(path);
```
ShaMan123, line 60: no reason for this to be a cb

test/unit/path.js
```diff
@@ -172,7 +176,7 @@
     var done = assert.async();
     makePathObject(function(path) {
       assert.ok(typeof path.toSVG === 'function');
-      assert.equalSVG(path.toSVG(), '<g transform=\"matrix(1 0 0 1 200.5 200.5)\"  >\n<path style=\"stroke: rgb(0,0,255); stroke-width: 1; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(255,0,0); fill-rule: nonzero; opacity: 1;\"  transform=\" translate(-200, -200)\" d=\"M 100 100 L 300 100 L 200 300 Z\" stroke-linecap=\"round\" />\n</g>\n');
+      assert.equalSVG(path.toSVG(), '<g transform=\"matrix(1 0 0 1 200 200)\"  >\n<path style=\"stroke: rgb(0,0,255); stroke-width: 0; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(255,0,0); fill-rule: nonzero; opacity: 1;\"  transform=\" translate(-200, -200)\" d=\"M 100 100 L 300 100 L 200 300 Z\" stroke-linecap=\"round\" />\n</g>\n');
```
ShaMan123, line 179: snapshot testing where are you??

test/unit/util.js
```diff
@@ -250,25 +250,20 @@
...
-    fabric.loadSVGFromString('   \n\n  ' + SVG_DOC_AS_STRING + '  ', function(objects) {
-      loadedObjects = objects;
-    });
-
-    setTimeout(function() {
-      assert.ok(loadedObjects[0] instanceof fabric.Polygon);
-      assert.equal('red', loadedObjects[0] && loadedObjects[0].fill);
+    fabric.loadSVGFromString('   \n\n  ' + SVG_DOC_AS_STRING + '  ').then(({ objects }) => {
+      assert.ok(objects[0] instanceof fabric.Polygon);
+      assert.equal('red', objects[0] && objects[0].fill);
```
ShaMan123, line 264:
```suggestion
      assert.equal(objects[0].fill, 'red');
```

ShaMan123 (changes requested):
This is a lot better than it was
Many comments

src/parser/elements_parser.ts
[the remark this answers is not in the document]
asturur: Yes i considered it, and i really said, i don't need that.
The only exposed method is parse all the rest is internal.
i could change it in a class only for a net gain in code size. ( and in another PR anyway )

src/parser/elements_parser.ts
```diff
@@ -30,52 +32,35 @@ const ElementsParser = function (
...
-    this.createObjects();
-  };
-
-  proto.createObjects = function () {
-    this.elements.forEach((element, i) => {
-      element.setAttribute('svgUid', this.svgUid);
-      this.createObject(element, i);
-    });
+  proto.parse = function (): Promise<FabricObject[]> {
+    return this.createObjects();
```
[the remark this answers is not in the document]
asturur: yes, good since it parse only objects as well and not a whole doc.

src/parser/elements_parser.ts
```diff
@@ -136,37 +110,41 @@ const ElementsParser = function (
...
+            .fromElement(clipPathElement, this.options)
+            .then((_newObj) => {
+              removeTransformMatrixForSvgParsing(_newObj);
+              _newObj.fillRule = _newObj.clipRule;
+              return _newObj;
+            });
+        })
+      );
+      const clipPath =
+        container.length === 1 ? container[0] : new Group(container);
```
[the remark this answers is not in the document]
asturur: i don't get it, explain again with a different sentence

test/unit/path.js
asturur: i think the issue was changing too many tests

test/unit/path.js
asturur: Look when we have time to migrate to jest we will get them  maybe. I don't have the mental strength now.
But those are a class of tests that can be totally be node only and jest will be good.

src/parser/elements_parser.ts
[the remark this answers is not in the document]
asturur: That requires a bit more of rework i ll fix it later in another PR.
I noticed it while reworking the code but wasn't straight forward to fix.
I wanted to focus on callback vs promises only.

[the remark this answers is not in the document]
asturur: Also is the only blocker to remove tsnocheck from that file

src/parser/loadSVGFromURL.ts
```diff
@@ -20,23 +20,27 @@ import type { LoadImageOptions } from '../util/misc/objectEnlive';
...
-  });
+    request(url.replace(/^\n\s*/, '').trim(), {
+      method: 'get',
+      onComplete,
+      signal: options.signal,
+    });
+  })
+    .then((parsedDoc) => parseSVGDocument(parsedDoc, reviver, options))
+    .catch(() => {
+      return emptyResponse;
```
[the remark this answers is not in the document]
asturur: i can spread it.

For the catch part i didn't want to change the previous behaviour of returning null.
That needs to be documented by the way

src/parser/parseSVGDocument.ts
asturur: i ll use clone

src/parser/parseSVGDocument.ts
asturur: ah good point i don;t know.

src/shapes/Path.ts
asturur: yes corret

src/parser/elements_parser.ts
```diff
@@ -136,37 +110,41 @@ const ElementsParser = function (
...
+      clipPath.set({
+        scaleX,
+        scaleY,
+        angle,
+        skewX,
+        skewY: 0,
+      });
       clipPath.setPositionByOrigin(
-        { x: options.translateX, y: options.translateY },
+        { x: translateX, y: translateY },
```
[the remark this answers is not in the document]
asturur: mehhh. I would love a micro benchmark. We assume creating a class is faster or equal to an object, in reality we don't know and we assume is better to use the class than the object.

src/parser/loadSVGFromURL.ts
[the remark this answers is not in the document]
asturur: I think i prefer to have all of them behave the same rather than catching and not catching.
If the SVG isn't parsable you get the empty response.

src/shapes/Text/Text.ts
asturur: i shouldn't have digged in this function and dismiss as out of scope, i refactored most of it hopefully i didn't break it.
There is more to cut away there, but i just really wanted to expose those as promises.

src/shapes/Ellipse.ts
asturur: is not useful spreading here, the options are then assigned to the class in the constructor.
Assigning the object or a copy of it should be the same.

src/shapes/Polyline.ts
asturur: you know i have NO IDEA what that options are for. Not gonna type till i discover.
Everything comes from the element, 2-3 classes have options, the others have no.
Seems to me options are those:
```js
    options = {
      ...applyViewboxTransform(doc),
      crossOrigin,
      svgUid,
      signal,
    };
```
And they have nothing to do with the partial props of the instance itslef.
I ll figure it out in another PR

src/shapes/Polyline.ts
asturur: points are not in parsedAttributes here.

src/parser/loadSVGFromURL.ts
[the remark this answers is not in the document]
ShaMan123: Checking for an empty response is less straight forward than catching

src/parser/parseSVGDocument.ts
ShaMan123: createEmptyRespose sounds straight forward.
Cloning means someine needs to remember that

src/shapes/Text/Text.ts
```diff
@@ -1816,10 +1801,18 @@ export class Text<
     textContent = textContent
       .replace(/^\s+|\s+$|\n+/g, '')
       .replace(/\s+/g, ' ');
```
ShaMan123, line 1803: can the remove svg spaces util be used here?

src/parser/elements_parser.ts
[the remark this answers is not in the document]
ShaMan123: > groupSVGElements?

should we reuse `groupSVGElements`  - NO. leave as is.


> Not related to PR: You know I have been thinking of this... we use Canvas/Group etc. w/o obtaining them from the registry. What if I want fabric to use my group impl?

We are using Group for example across fabric. Say in my app I override group essentials and I want it to reflect everywhere so I register it in the class registry. Fabric should use that. I remember you mentioned this before.

ShaMan123 (commented):
I prefer `createEmptyResponse` instead of `deepClone` simply because it is safer for future use.
Looks good.

src/parser/elements_parser.ts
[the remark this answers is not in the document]
asturur: yes this maybe should use getKlass at that point

src/shapes/Text/Text.ts
asturur: no as said before ths function is out of scope.

asturur:
> I prefer `createEmptyResponse` instead of `deepClone` simply because it is safer for future use. Looks good.

doens't cost much to address

PR #8894, opened 2023-05-07
chore(): Matrix util cleanup

ShaMan123:
## Motivation

continues #8881 

## Description

Matrix utils belong to `utils`, not to `parser`
It is wrong to have `utils` import from `parser` since `utils` should be pure as much as possible and these methods have value for more than just the parser


src/util/misc/matrix.ts
ShaMan123, line 1: I don't like the naimg `calcXXXMatrix` but I left it as is because `calcRotateMatrix` was already here.
I think `createXXXMatrix` is better.
Thoughts?

ShaMan123 (commented):
cleanup done!
ready

src/util/misc/angleSkewConversion.ts
```diff
@@ -0,0 +1,8 @@
+import type { TRadian, TDegree } from '../../typedefs';
+import { degreesToRadians, radiansToDegrees } from './radiansDegreesConversion';
+
+export const angleToSkew = (angle: TDegree) =>
```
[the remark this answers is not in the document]
asturur: This was put in its on function just to save on code, i think it should stay in the matrix files and doesn't need to be referred as an export.

src/util/misc/matrix.ts
```diff
@@ -107,22 +106,126 @@ export const qrDecompose = (a: TMat2D): TQrDecomposeOut => {
...
+    sinValue = sin(angleRadiant);
+  return [
+    cosValue,
+    sinValue,
+    -sinValue,
+    cosValue,
+    x - (cosValue * x - sinValue * y),
+    y - (sinValue * x + cosValue * y),
+  ];
+}
```
asturur, line 161: The parser calcRotateMatrix and the old calcRotateMatrix are the only one i don't want to mix.
I m not sure the pivot point in our geometry is identical to the one in the svg context.

ShaMan123: I will add a test comparing this and rotatePoint

src/util/misc/angleSkewConversion.ts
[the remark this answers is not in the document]
ShaMan123: It has value IMO but I didn't export it publicly

src/util/misc/matrix.ts
ShaMan123: take a look at 5d3642811 with critical eyes

src/util/misc/matrix.ts
asturur: i don't think it makes a difference for me.
They weren't a big part of the interface so doensn't matter but the one from the object method should not chage

src/util/misc/matrix.ts
asturur: i m not worried about that use case.
Those matrices were the building block of calcTransformMatrix.
I would like to have the calcRotateMatrix used for calcTransformMatrix, different from the function that has to only serve the SVG parsing abilities.

Is ok if the svg parser call the old `calcRotateMatrix` and then adds the missin 2 values and we name that parseSVGRotateTransformAttribute.

I don't know of svg notations that may need different operation to the one that we have and that are identical to the one we were already using, so scale/skew/translate do not bother me.

ShaMan123: I don't really understand
I think I proved it is identical to how we use origin and since there aren't other use cases seems ok.
But I understand the mistrust.
I do think it is useful outside the parsing context.

src/util/misc/matrix.ts
ShaMan123: > one from the object method should not chage

What is that?

asturur: that calcTransfromMatrix should stay calcTransformMatrix, because now is memonic

src/util/misc/matrix.ts
asturur: i just want the pivot point, that was part of the svg parsing and nothing else, to stay part of the svg parsing.
I don't want to mix them up because we save a bunch of code and because if we pass 0,0 at the pivot point then the function is functionally he same.
Let's leave the previous calcRotateMatrix and the svg one divided.
We may want to change the svg parsing code independently from the canvas code.

ShaMan123: I still don't get your point. As a standalone util it is great, isn't it?
Are you worried it does more calculations than before? Is that the point?

> We may want to change the svg parsing code independently from the canvas code.

Then we change it accordingly.

I have encountered a number of times the need of this util and had to use `rotatePoint` to figure out the translation. That is why I want it exposed.

src/util/misc/matrix.ts
ShaMan123: what about `calcDimensionsMatrix`?

src/util/misc/matrix.ts
```diff
@@ -84,6 +82,46 @@ export const multiplyTransformMatrices = (
...
+/**
+ * Multiplies {@link matrices} such that a matrix defines the plane for the rest of the matrices **after** it
+ *
+ * `multiplyTransformMatrixChain([A, B, C, D])` is equivalent to `A(B(C(D)))`
+ *
+ * @param matrices an array of matrices
+ * @param [is2x2] flag to multiply matrices as 2x2 matrices
+ * @returns the multiplication product
+ */
+export const multiplyTransformMatricesFromEnd = (
```
ShaMan123, line 94: what do you think about naming?

src/util/misc/matrix.ts
```diff
@@ -84,6 +82,46 @@ export const multiplyTransformMatrices = (
...
+ *
+ * `multiplyTransformMatrixChain([A, B, C, D])` is equivalent to `D(C(B(A)))`
+ *
+ * **Note**: this form of multiplication is considered as reversed mathematically
+ *
+ * @param matrices an array of matrices
+ * @param [is2x2] flag to multiply matrices as 2x2 matrices
+ * @returns the multiplication product
+ */
+export const multiplyTransformMatricesFromStart = (
```
ShaMan123, line 115: maybe this is a bad idea because it is equal to `multiplyTransformMatricesFromEnd(matrices.reverse())`

ShaMan123 (commented):
added
~~`multiplyTransformMatricesFromEnd`~~ renamed
~~`multiplyTransformMatricesFromStart`~~ removed
`multiplyTransformMatrices2` - do you a good name for it?

to multiply a chain of matrices

src/util/misc/matrix.ts
asturur: That does calculations no? it has more matrices in. si calc seems appropriate

src/parser/parseTransformAttribute.ts
```diff
@@ -87,8 +90,5 @@ export function parseTransformAttribute(attributeValue: string): TMat2D {
     matrices.push(matrix);
   }
 
-  return matrices.reduce(
-    (acc, matrix) => multiplyTransformMatrices(acc, matrix),
-    iMatrix
-  );
```
asturur, line 93: we are changing a reduce to reduce right here nad i don't understand why.
I m incline to say don't touche the current transformation code at all, is not a cleanup anymore at that point

src/util/misc/matrix.ts
```diff
@@ -84,6 +82,25 @@ export const multiplyTransformMatrices = (
...
+ */
+export const multiplyTransformMatrices2 = (
+  matrices: (TMat2D | undefined | null | false)[],
+  is2x2?: boolean
+) =>
+  matrices.reduceRight(
+    (product: TMat2D, curr) =>
+      curr ? multiplyTransformMatrices(curr, product, is2x2) : product,
+    iMatrix
+  );
```
asturur, line 102: maybe i don't understan reduceRigth but it seems to me we are inverting the order of multiplication of matrices.

asturur (changes requested):
- multiplyTransformMatrices2 doesn't seems a good name.
- multiplyTransformMatrices2 uses reduce and not reduceRight
- calcRotateMatrix does not get the extra calculation

Started as a cleanup is drifing away from that concept

src/util/misc/matrix.ts
asturur: Well the 4 addition and 4 multiplication are also unwelcome an extra point to have divided in 2 stages that is all i m asking.

src/parser/parseTransformAttribute.ts
ShaMan123: I agree.
Touched it and became uneasy but the math is correct and I double checked that there is a test to cover it.
This proves it is all right: https://en.wikipedia.org/wiki/Matrix_multiplication#Associativity

Notice the args order passed to `multiplyTransformMatrices`

src/util/misc/matrix.ts
ShaMan123: This is the right order. 
All tests pass.
https://en.wikipedia.org/wiki/Matrix_multiplication#Associativity

Notice the args order passed to `multiplyTransformMatrices`

ShaMan123:
> multiplyTransformMatrices2

 what name do you suggest?
`multiplyTransformMatrixArray`?

asturur:
> > multiplyTransformMatrices2
> 
> what name do you suggest? `multiplyTransformMatrixArray`?

That works for me

src/util/misc/matrix.ts
asturur: i know the result doesn't change because i know you know math. ( on top of that i know that property too ).
I opened the article anyway to see if maybe there was explained a mathematical property that was the reason for this change, but i couldn't find one.

So i don't know why you think that is the most correct order.
Why do you think is important to change it? I always have express preferences for don't touch what is not getting measurably better and i think you know it by now.

( apart that given the flawed float system of js, changing mutliplication order of floats ( the one inside the matrices ) isn't exactly the same thing. It could be, but sometimes it won't be)
( and apart that most of developers, inclding me, when they see reduceRight, they have to go and check how ti differs from reduce )

ShaMan123: This is how it is done mathmatically
Because of [Associativity](https://en.wikipedia.org/wiki/Matrix_multiplication#Associativity) it doesn't matter. I wasn't sure at the time of writing but now I am.
So it can be turned into reduce but IMO this is the correct manner.

ShaMan123: added a test and renamed

src/util/misc/matrix.ts
ShaMan123: can I just have an early return instead if the origin is 0,0?

src/util/misc/matrix.ts
ShaMan123: I will better explain.
Without [Associativity](https://en.wikipedia.org/wiki/Matrix_multiplication#Associativity) it would matter.
Meaning, that using reduce multiplies left to right and is correct only because of [Associativity](https://en.wikipedia.org/wiki/Matrix_multiplication#Associativity). 
Multiplying RTL is how matrix multiplication is defined

ShaMan123: It is confusing

asturur: after much thinking i will write you what i think.
I think that even removed any personal preference, the rule of no gain no change should be easy to agree with for everyone.
There is no gain here ( just minor losses ) ( i can't even find references about one order being more canon that the other ) and just the will to have this change.
This was hours ago, now you take this reduceRight and you keep it, my time and mood is not worth this change.
When i get to this point i the never want to talk about it anymore so please help me there.

asturur (dismissed):

PR #8897, opened 2023-05-07
feat(env): support iframe with relative window/document

### LINKED ISSUE #4805: Does fabric.js support canvas inside iframe?
I got an iframe inside a html and a canvas inside the iframe. When I add fabric.Text to the canvas from html, the text did appear on the canvas but cannot drag around or do any interaction. Please help here:)
```
var iframeDocument = document.getElementById('myIframe').contentWindow.document;
var canvas = iframeDocument.getElementById('myCanvas');
var fabricCanvas = new fabric.Canvas(canvas);

var text = new fabric.Text('Hello from parent', {
   left: 100,
   top: 100
});

fabricCanvas.add(text);
```

ShaMan123:
## Motivation

closes #4805 

## Description

Changes the event targets to the window/document containing the canvas element of the fabric canvas so it should work inside multiple iframes

## In Action

https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-6pkg55

With images: https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-7y55kw?file=%2FREADME.md

https://user-images.githubusercontent.com/34343793/236675903-d18cb3ec-f5b8-419b-8ace-1aa7e12d8995.mp4


src/canvas/StaticCanvas.ts
```diff
@@ -417,20 +417,6 @@ export class StaticCanvas<
...
-  protected _createCanvasElement() {
-    const element = createCanvasElement();
-    if (!element) {
-      throw new Error(CANVAS_INIT_ERROR);
-    }
-    if (typeof element.getContext === 'undefined') {
-      throw new Error(CANVAS_INIT_ERROR);
-    }
-    return element;
-  }
```
asturur, line 432: This will error later with a different less clear error
something like `can't use <insertFunctionName> on null` or similar.

Do we want to remove this wrapper and why?

asturur:
I m mostly onboard apart naming.
getWindow and getElementWindow at this point are so much similar and is unclear when i should use wich one.

getWindow at this point is the window where fabric lives and would need to reflect that.

Instatiating a fabric element across an iframe seems so weird to me and dangerous.
For exampe images, if you use the loadImage functionality you have an image loaded in fabric window, and then you add it on canvas instance that is in the fabric window, but the draw command is going to act on a context that is in the iframe referencing pixels that are in the main window.

Would that work?


Should those be method of the canvas rather than utilities?

ShaMan123:
good point
I have expose `getWindow/Document` on canvas but didn't commit
Maybe we should have an env for each canvas and pass that around

ShaMan123:
https://codesandbox.io/s/hybc13 works, need to check all cases

asturur:
Alos disclaimers about cross origin work probably being impossible.
Do not expect to iframe a website with a canvas and take over their canvas from another domain.

An env per canvas seems to much.
But if those functions always need the canvas element, they can live on the canvas instance.
Maybe is a good idea to differentiate them.

ShaMan123:
> Also let's add disclaimers about cross origin work probably being impossible.
> Developers should not expect to iframe a website with a canvas and take over their canvas from another domain.

this is why I think we should have an env scoped to canvas. That way when everything belongs to same domain.
I will try that out

ShaMan123:
I think best to do it in 2 steps. This PR as is and another one for the scoped env concept.

ShaMan123:
looks good: https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-7y55kw?file=%2FREADME.md

ShaMan123:
I think this is good enough as a first step
Eventually we will make a guide for all envs

ShaMan123:
> Instatiating a fabric element across an iframe seems so weird to me and dangerous. For exampe images, if you use the loadImage functionality you have an image loaded in fabric window, and then you add it on canvas instance that is in the fabric window, but the draw command is going to act on a context that is in the iframe referencing pixels that are in the main window.
> 
> Would that work?

https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-sy6726

src/canvas/StaticCanvas.ts
ShaMan123: I will make it part of createCanvasElement util

ShaMan123: 6e48007d2

ShaMan123 (commented):
Naming is left

ShaMan123:
I would like to add tests but I don't know how to tests that an iframe receives events etc.

asturur:
> I would like to add tests but I don't know how to tests that an iframe receives events etc.

We can do later with that new test suite probably. I didn't have time to write proper tests but seems exactly useful for those kind of high level interactions

asturur:
Do not open follow up prs for this concept because if this works, all we need is just docs.

My initial comment still stand

> I m mostly onboard apart naming.
getWindow and getElementWindow at this point are so much similar and is unclear when i should use wich one.

> getWindow at this point is the window where fabric lives and would need to reflect that.

getWindow => getFabricWindow
getDocument => getFabricDocument
getElementDocument => getDocumentForElement (for/of)
getElementWindow => getWindowForElement (for/of)

asturur (changes requested):
would like to see a different naming scheme that makes the function apart.
The difference between elementDocument ( i think intented as element's document ) and the documentElement ( the root node of the document ) is too subtle

ShaMan123:
renamed used for instead of for/of

test/lib/event.simulate.js
ShaMan123, line 1: I am pretty sure this is not correct

ShaMan123: but it is outdated anyways.
What is it?

ShaMan123 (commented):
renamed, noisy diff

asturur (approved):

asturur (approved):

Issue #8887, opened 2023-05-06
[Bug]: Overlay disappears when panned out of the canvas, even if overlayVpt is set to false


JPLassala-bix:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

5.3.0

### In What environments are you experiencing the problem?

Firefox, Chrome, Safari

### Node Version (if applicable)

None

### Link To Reproduction

Overlay disappears when panning even with overlayVpt set to false

### Steps To Reproduce

1. Create a new canvas
2. Create a new shape
3. Add the shape as an overlay, making sure it is displayed on the canvas
4. Pan on the canvas using `relativePan()` until the internal overlay coordinates get out of the screen.

I made a working demo reproducing the error:
https://codesandbox.io/s/long-frost-2dp3u4?file=/src/index.js

To pan in that canvas, one should press and hold the space bar, and then click and drag to any direction. When dragging, a clone of the circle and a blue square (used as a guide) will move until the circle goes out of the screen.
When that happens, the overlay disappears.

I included some console output that indicates when the overlay ends up disappearing.

### Expected Behavior

When `overlayVpt` is set to `false`, the overlay shouldn't be affected by viewPort transformations.

### Actual Behavior

When panning on the canvas enough that the viewport transform goes beyond the supposed overlay location, the overlay is counted as `isOffScreen` and it doesn't render anymore 

### Error Message & Stack Trace

```bash
When the console says:

> Overlay in screen:  false

The object disappears.
```

ShaMan123:
v6 working: https://codesandbox.io/s/keen-shannon-eigcee?file=/src/index.js
I believe the error is in your code.
You shouldn't add `circle` to canvas. But there is something else there.

ShaMan123:
I think there is a bug here regarding bbox calc.
[...]

PR #8898, opened 2023-05-07
fix(): Disable offscreen check for bg and overlay when not needed 

asturur:
## Motivation

closes https://github.com/fabricjs/fabric.js/issues/8887

the setting to make bg and overlay independent from zoom and panning are not respected by isOffscreen.
There isn't an easy way to solve this, and this one line fix is probably a way to block the bug for now.

We could have pulled the onscreen check outside of the render function but that could have unintended consequences on existing subclasses and also share the responsibility of offscreen check to more players ( canvas and group at least ).

## Changes

The function to render backround or overlay will disable the offscreen check if needed and restore it later


src/canvas/StaticCanvas.ts
```diff
@@ -885,10 +885,15 @@ export class StaticCanvas<
     }
     if (object) {
       ctx.save();
+      const { skipOffscreen } = this;
+      // if the object doesn't move with the viewport,
+      // the offscreen concept does not apply;
+      this.skipOffscreen = needsVpt;
```
ShaMan123, line 891: I am so confused.
Shouldn't it be the other way around? If it respects vpt it should be subject to offscreen optimizations.
Anyways I think tests will not pick this up because the bg/overlay objects have no canvas ref (which is dangerous as well), what makes the offscreen test skip entirely.

ShaMan123 (commented):
I think you should fine tune tests and put in the description this regards only rendering

src/canvas/StaticCanvas.ts
asturur: Eh because skipOffscreen means skip rendering of offscreen object, not skip the check.
So if needsVpt, skip the rendering when offscreen.
For a moment i thought i got confused.

I can disable the line and see if tests fail

asturur:
I see how the necessity of adding canvas to backgroundImage ( or overlay ) is unhandled.
This will need to be addressed somehow. Can't be left to the dev

ShaMan123 (approved):

PR #8901, opened 2023-05-08
fix(Canvas): make `dispose` sync - revert async logic

ShaMan123:
## Motivation

closes #8899 
closes #8903 

## Description

DOM cleanup of `Cavas#dispose` needs to be a sync operation for compatibility with other tools (React useEffect)
In fact making it async was cumbersome. This approach is much more straight forward.
It uses a simply try-catch block.
It is as safe as before.

The difference is in concept.
**After calling dispose the dev should not expect any pending rendering tasks to succeed.**

## Changes

- wrap `renderCanvas` with try-catch (logic resides in `renderCanvasUnsafe`
- split dispose logic into `cleanupDOM` and `destroy`

## In Action

This branch: https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-584cvb
Previous commit (broken in useEffect): https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-lowfw5

PR #8903, opened 2023-05-08
fix(Canvas): sync cleanup of dom elements in dispose

### LINKED ISSUE #8899: [Bug]: Can't unmount Canvas in react useEffect
### CheckList

### Version

6.0.0-beta5

_No response_

None

### Link To Reproduction

https://github.com/vinkovsky/fabric-demo/tree/main

### Steps To Reproduce

1. clone repo or move to codesandbox link
2. pnpm i 
3. pnpm start
 

### Expected Behavior

The component should mount and unmount without errors

### Actual Behavior

The canvas does not unmount properly.

```bash
Uncaught Error: fabric.js: trying to initialize a canvas that has already been initialized
    at Canvas._createLowerCanvas (StaticCanvas.ts:450:1)
    at Canvas.initElements (StaticCanvas.ts:335:1)
    at Canvas.initElements (SelectableCanvas.ts:517:1)
    at new StaticCanvas (StaticCanvas.ts:325:1)
    at Canvas._createSuperInternal (createSuper.js:11:1)
    at new SelectableCanvas (SelectableCanvas.ts:512:1)
    at Canvas._createSuperInternal (createSuper.js:11:1)
    at new Canvas (Canvas.ts:122:1)
    at App.js:9:1
    at commitHookEffectListMount (react-dom.development.js:23150:1)
_createLowerCanvas @ StaticCanvas.ts:450
initElements @ StaticCanvas.ts:335
initElements @ SelectableCanvas.ts:517
StaticCanvas @ StaticCanvas.ts:325
_createSuperInternal @ createSuper.js:11
SelectableCanvas @ SelectableCanvas.ts:512
_createSuperInternal @ createSuper.js:11
Canvas @ Canvas.ts:122
(anonymous) @ App.js:9
commitHookEffectListMount @ react-dom.development.js:23150
invokePassiveEffectMountInDEV @ react-dom.development.js:25154
invokeEffectsInDev @ react-dom.development.js:27351
commitDoubleInvokeEffectsInDEV @ react-dom.development.js:27330
flushPassiveEffectsImpl @ react-dom.development.js:27056
flushPassiveEffects @ react-dom.development.js:26984
(anonymous) @ react-dom.development.js:26769
workLoop @ scheduler.development.js:266
flushWork @ scheduler.development.js:239
performWorkUntilDeadline @ scheduler.development.js:533
react-dom.development.js:22839 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at removeChildFromContainer (http://localhost:3001/static/js/bundle.js:16707:19)
    at commitDeletionEffectsOnFiber (http://localhost:3001/static/js/bundle.js:27734:19)
    at recursivelyTraverseDeletionEffects (http://localhost:3001/static/js/bundle.js:27700:9)
    at commitDeletionEffectsOnFiber (http://localhost:3001/static/js/bundle.js:27811:13)
    at recursivelyTraverseDeletionEffects (http://localhost:3001/static/js/bundle.js:27700:9)
    at commitDeletionEffectsOnFiber (http://localhost:3001/static/js/bundle.js:27856:13)
    at commitDeletionEffects (http://localhost:3001/static/js/bundle.js:27690:9)
    at recursivelyTraverseMutationEffects (http://localhost:3001/static/js/bundle.js:27928:13)
    at commitMutationEffectsOnFiber (http://localhost:3001/static/js/bundle.js:28075:13)
    at commitMutationEffects (http://localhost:3001/static/js/bundle.js:27915:7)
(
```

ShaMan123:
## Motivation

closes #8901 
closes #8899

## Changes

- expose `cleanupDOM` on canvas that runs from dispoe (sync)
- moved `dispose` to node env and use only in `index.node`


ShaMan123 (commented):
- expose `cleanupDOM` on canvas that runs from dispoe (sync)
- moved `dispose` to node env and use only in `index.node`
-

src/Pattern/index.ts
```diff
@@ -1,2 +1,2 @@
 export { Pattern } from './Pattern';
-export type * from './types';
```
ShaMan123, line 2: ts warning

ShaMan123 (commented):
Shouldn't we dispose of all cache canvases in Object#dispose??

ShaMan123 (commented):
I think this ok
It is better than #8901 for the case of running an async `toCanvasElement` during disposing but I don't like the jsdom dispose thing

ShaMan123 (commented):
added disposing to object...
I have to say I don't like it and prefer to have that specific dispose logic in the node entry only like I did with image before I reverted.
But it means we need to apply a mixin to fabricObject to patch destroy

ShaMan123:
But I can live with it.
It is strange ow that canvas disposing is done in the node entry file though.
But my point for doing that is say one day we use the env dispose in the browser as well or a different env that reuses the element. We cleanup the dom, lets say hot relaoding occured so everything is great. But then destroy kicks in a cripples the element and no one know what happened.
That is why I left it in the node entry. I would feel better if it were part of the dom cleanup function but then the pending rendering will throw so we missed out the entire point of this effort.
All this is why I am not convinced this is better than #8901 that says - once you called dispose don't expect a pending render to succeed. 
So all in all I prefer #8901 and wish do better understand why you don't

ShaMan123:
Well, yes. Thinking of it I dislike this.

ShaMan123:
I would argue that we should combine the 2.
Have dispose be sync and expose the cleanupDOM method

ShaMan123:
just did that in #8901

asturur:
i don't like the idea of try catching in a  render because regardless of typescript requiring us a context, then we know there are situation in which will fail anyway because somethin could fire a render after dispose.
requestAnimationFrame is async, it doesn't cancel imperatively as we would like to, we have to handle it asyncronously.
There is no reason to force it to be sync after all the efforts we did to cancel it properly.

On top of that again, the issue here is not to be sync or async, is the interaction with react and we need a foolproof react wrapper at somepoint, and cleaning up the dom elements is just a part of it.

Probably ( didn't test ) you don't even need to dispoe the canvas on hot reload, you can even just call cleanupDOM and let the instance be garbage collected naturally.

asturur:
Is also important that before reverting and changing idea on something we think twice or three times.
We can't waste hours on a pr and the at the first issue remove it without thinking of other ways to solve the issue.
this approach just change the order of operations and fixes the issue.

asturur:
Please give an explanation why you dislike this approach and then close the pr.
You don't have to do things you don't like

src/Pattern/index.ts
```diff
@@ -1,2 +1,2 @@
...
+export * from './types';
```
asturur: this is unrelated and as far as i understand will bring back the extra dumb bundling object. Don't mix up this change, we will check up this during the TS prs

ShaMan123:
I saw you commented on garbage collection.
Why then not leave all of the disposing part accept the dom cleanup to gc and remove this logic from fabric altogether?
A different approach can be throwing when dispose is called during a requested/active render - That is better IMO because it forces the dev to own the cleanup of animations, requested renders, etc. and it can be easily caught with try-catch

ShaMan123:
What do you think? 
Is disposing is a node only thing?
We can merge this and reiterate if needed.

What I have against it:
If we instruct users to use `cleanupDOM` and not dispose, then why have the dispose logic?
If `dispose` meant for node only?

asturur:
> I saw you commented on garbage collection. Why then not leave all of the disposing part except the dom cleanup to gc and remove this logic from fabric altogether? A different approach can be throwing when dispose is called during a requested/active render - ~That is better IMO because~ it forces the dev to own the cleanup of animations, requested renders, etc. and it can be easily caught by the dev with try-catch

A one issue that we got in the past is that node was accumulating canvases buffer till would give an exception.
Node GC would eventually clear up the JS part of node-canvas and JSDOM but somehow those buffers needed to be removed manually or node would create issues.
There are issues in the repo on this matter and unless we can prove that a correct code initialization and disposal will prevent any memory leak, i would leave the code in place at least to say, `we are trying t make your server not die`.

asturur (approved):

PR #8904, opened 2023-05-09
fix(TS): some left overs

ShaMan123:
## Motivation

left overs for recent merges


src/Pattern/index.ts
```diff
@@ -1,2 +1,2 @@
 export { Pattern } from './Pattern';
-export type * from './types';
+export * from './types';
```
ShaMan123, line 2: TS complains.
I see the bundle grew so this might not be good

ShaMan123: and rather annoying

ShaMan123: I have no idea why the bundle grows

asturur (approved):

asturur:
Probably every ts warning apart sourcemaps should be considering a blocking error

ShaMan123:
> Member

ShaMan123:
> Probably every ts warning apart sourcemaps should be considering a blocking error

agreed, why not source maps?

ShaMan123:
We just need to throw when a ts warning is fired
https://github.com/fabricjs/fabric.js/blob/d7bb027058fc0ce30dbc2b610336a7bcc4756e7a/rollup.config.mjs#L35

PR #8906, opened 2023-05-09
fix: :bug: typescript compiler error

seal7766:
## Motivation

fix typescript compiler error

## Description

Fix typescript compiler error
- In the case of the '_isClick' property, an error occurred because it was declared in the class property but not initialized. To correct this, it was changed to optional. The reason is that the value is a boolean value, and even if it is not assigned, it is recognized as undefined, so it is recognized as a negative value.

- The 'draggableTextDelegate' property is the same as the above, but since it is an instance, I did not change it to optional and thought that the initBehavior method was called necessarily, so I modified it using a Non-null assertion operator.

- For the reason why 'SerializedObjectProps' and 'FabricObjectProps' types were added, please refer to the link below.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#type-on-import-names

## Changes

- Modified the '_isClick' property to be optional.
- Modified the draggableTextDelegate' property to be Non-null assertion operator.
- Add type export to 'SerializedObjectProps', 'FabricObjectProps'

### DIFF
--- a/src/shapes/Object/types/index.ts
+++ b/src/shapes/Object/types/index.ts
@@ -1,7 +1,7 @@
 import { FabricObjectProps } from './FabricObjectProps';
 
-export { SerializedObjectProps } from './SerializedObjectProps';
-export { FabricObjectProps };
+export { type SerializedObjectProps } from './SerializedObjectProps';
+export { type FabricObjectProps };
 
 export type TProps<T> = Partial<T> & Record<string, any>;
 
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -114,7 +114,7 @@ export class Canvas extends SelectableCanvas {
 
   declare currentSubTargets?: FabricObject[];
 
-  private _isClick: boolean;
+  private _isClick?: boolean;
 
   textEditingManager = new TextEditingManager();
 
--- a/src/shapes/IText/ITextClickBehavior.ts
+++ b/src/shapes/IText/ITextClickBehavior.ts
@@ -30,7 +30,7 @@ export abstract class ITextClickBehavior<
   private declare __lastPointer: XY | Record<string, never>;
   private declare __newClickTime: number;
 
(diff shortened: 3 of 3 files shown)


asturur:
Hello, thanks for the pr, we discussed this change here:
https://github.com/fabricjs/fabric.js/pull/8876

Please take your time to read the explanation, for some reason the origina author then decided to close the PR.
What we want is probably a declare there.
Those 2 properties are being unitialized as an undefined now by tanspilation and in the range of classes between canvases and shapes we don't want that to happen

asturur:
Hello, thanks for the pr, we discussed this change here:
https://github.com/fabricjs/fabric.js/pull/8876

Please take your time to read the explanation, for some reason the origina author then decided to close the PR.
What we want is probably a declare there.
Those 2 properties are being unitialized as an undefined now by tanspilation and in the range of classes between canvases and shapes we don't want that to happen

asturur:
we are in the preocess of exporting types because we miss the fact that had to be explicit.
So you can safely add export to any type you know you could be using in your app.
We are about to do specific prs for that tho, so not sure if is worth doing in this PR

src/shapes/Object/types/index.ts
```diff
@@ -1,7 +1,7 @@
 import { FabricObjectProps } from './FabricObjectProps';
 
-export { SerializedObjectProps } from './SerializedObjectProps';
-export { FabricObjectProps };
+export { type SerializedObjectProps } from './SerializedObjectProps';
+export { type FabricObjectProps };
```
asturur, line 4: we are adding a lint rule that *SHOULD* catch this, but for now it doesn't work for some reason.
It catches imports but not exports yet.
https://github.com/fabricjs/fabric.js/pull/8907

seal7766: @asturur 
Thanks for the explanation.

This PR will be closed!

asturur:
the other 2 lines were good.
We need to add declare in front of those props. Someone has to do it at some point

PR #8907, opened 2023-05-09
chore(linting) apply a rule that fixes import vs import type when is possible

asturur:
## Motivation

This adds import consistency and probably speeds up aware bundler/parsers
The changes are all made with the lint --fix command and no human intervention.
## Changes

import type will be enforced when possible


asturur:
ok export type works in vscode but for some reason not with the lint command

.eslintrc.js
```diff
@@ -13,6 +19,8 @@ module.exports = {
     },
   ],
   rules: {
+    '@typescript-eslint/consistent-type-exports': 'error',
+    '@typescript-eslint/consistent-type-imports': 'error',
```
ShaMan123, line 23:
```suggestion
    '@typescript-eslint/consistent-type-exports': [
      'error',
      {
        fixMixedExportsWithInlineTypeSpecifier: true,
      },
    ],
    '@typescript-eslint/consistent-type-imports': [
      'error',
      {
        fixStyle: 'inline-type-imports',
      },
    ],
```

ShaMan123: what do you think?

https://typescript-eslint.io/rules/consistent-type-imports/#fixstyle
https://typescript-eslint.io/rules/consistent-type-exports/#fixmixedexportswithinlinetypespecifier

ShaMan123: image

ShaMan123 (commented):
I see there are more actionable code paths that lint didn't fix.
This is a good tool.
Since we are going in this direction how about https://eslint.org/docs/latest/rules/sort-imports?
Before, I wanted it but I thought it was part of the scope of prettier, but of course it isn't. It is the scope of lint. Do you agree?

asturur:
i don't want to sort import.
We don't sort vars, we don't sort methods in a class nor property in objects, i don't see why i have to sort imports.

asturur:
so inline type import has been introduced in typescript 4.5
i would skip it for now, not sure all tools support it yet

.eslintrc.js
asturur: aren't those catched with default values?

asturur: even applying your config i don't get the extra changes.

package.json
```diff
@@ -57,7 +57,7 @@
     "test:coverage": "nyc --silent qunit test/node_test_setup.js test/lib test/unit",
     "test:visual:coverage": "nyc --silent --no-clean qunit test/node_test_setup.js test/lib test/visual",
     "coverage:report": "nyc report --reporter=lcov --reporter=text",
-    "lint": "eslint --config .eslintrc.js src/**/*.ts",
+    "lint": "eslint --config .eslintrc.js src",
```
asturur, line 60: this was my issue. 
i was getting only files inside the first folders.
VSCode indeed was looking and eslint.rc and workin correctly.

asturur:
I updated lint and tsplugins for lint to latest since i was trying to figure out why i didn't see the things

asturur:
@ShaMan123 regarding inline vs non inline.
For me separated import exports are very clear to see and follow. the `type` notation get lost for me in the import / export string if is mixed with the rest.
I also like the idea to stick with default configuration when possible.

ShaMan123:
So that is why lint ran on my machine and produced a big diff? Because of the merge conflict and the versuons of the tool?

I don't mind about the syntax. Wanted to highlight the options so we make an informed decision.

ShaMan123 (approved):
looks valid to me
I didn't check every file though

PR #8909, opened 2023-05-10
reflect NUM_FRACTION_DIGITS to toObject path data

tushuhei:
## Motivation
Related to #5734.
#8782 reflects `NUM_FRACTION_DIGITS` to the `toSVG` outputs, but not `toObject`.
This PR reflects `NUM_FRACTION_DIGITS` to the `toObject` outputs.

## Description
- Add the `roundCommand` function to `src/util/path/index.ts`.
- Use the method in `toObject`.

## Changes
Repro: https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-es0d03?file=%2Fpages%2Findex.tsx&selection=%5B%7B%22endColumn%22%3A39%2C%22endLineNumber%22%3A18%2C%22startColumn%22%3A39%2C%22startLineNumber%22%3A18%7D%5D


asturur:
@tushuhei we can do this but we need to do it in a different way.
Way you do here is rounding runtinme numbers in order to affect toObject/ SVGoutput.
For https://github.com/fabricjs/fabric.js/pull/8782 you applied the fix in the toSVG code branch while in this case you are changing how the path looks like in the canvas.

So there are 2 things at stake here.

1) The semplification logic needs ot be in toObject in some way.

2) The other thing could be a different feature ( that we won't address here in the same PR ) that let the developer have a path sticking to a grid. so it would round yes, but not in the sense of decimal but in the sense of units you want the path to conform to ( like only integers, or only decimals, or only values that increment by 0.2 )

asturur:
Let me know for (1) if you want some guidance or you wnat to enjoy it yourself.

tushuhei:
@asturur  Thanks for taking a look. I fixed the code accordingly.

PR #8910, opened 2023-05-11
fix(Object): dirty unflagging inconsistency

ShaMan123:
## Motivation

Work issue

## Description

Object should unflag dirty after `drawObject`, see ref:
https://github.com/fabricjs/fabric.js/blob/d7bb027058fc0ce30dbc2b610336a7bcc4756e7a/src/shapes/Object/Object.ts#L768-L769

## Changes

Consistently mark object as not dirty **after** calling `drawObject`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
-- fix(Object): dirty flagging cleanup inconsistency [#8910](https://github.com/fabricjs/fabric.js/pull/8910)
+- fix(Object): dirty unflagging inconsistency [#8910](https://github.com/fabricjs/fabric.js/pull/8910)
 - chore(TS): minor type/import fixes [#8904](https://github.com/fabricjs/fabric.js/pull/8904)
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
ready

asturur (approved):

asturur:
Can you explain the issue in the PR please?
PRs need to be a bit self explanatory.
Work issue is clear between me and you, the average reader that could read that change has no clue.
Consistency is clear, but what issue is actually causing isn't.

ShaMan123:
Code is heavily overriding the render cycle. There is a state check involving the dirty flag.
Since it wasn't consistent, non cache rendering was unflagging after drawObject while cache rendering was unflagging before it made the code cumbersome.
Regardless I am looking into the impl but as a first step this seems to make sense in the eyes of a maintainer.
Did I manage to explain?

asturur:
yes i just want to have a PR description that is different from `i m having some issue and this solves it for me`

ShaMan123:
Yeah, I get that. For sure.

PR #8913, opened 2023-05-12
chore(TS) export filter types

asturur:
## Motivation

More type exporting

### DIFF
--- a/fabric.ts
+++ b/fabric.ts
@@ -49,3 +49,14 @@ export { Control } from './src/controls/Control';
 export * as controlsUtils from './src/controls';
 
 export * from './src/filters';
+export type {
+  TProgramCache,
+  TTextureCache,
+  TPipelineResources,
+  TWebGLPipelineState,
+  T2DPipelineState,
+  TWebGLUniformLocationMap,
+  TWebGLAttributeLocationMap,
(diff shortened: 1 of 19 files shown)


ShaMan123 (approved):

fabric.ts
```diff
@@ -13,12 +13,15 @@ export { Canvas } from './src/canvas/Canvas';
...
 export { Gradient } from './src/gradient/Gradient';
 export * from './src/Pattern';
 export { Shadow } from './src/Shadow';
 
 export { BaseBrush } from './src/brushes/BaseBrush';
+export * from './src/brushes/typedefs';
+
 export { PencilBrush } from './src/brushes/PencilBrush';
 export { CircleBrush } from './src/brushes/CircleBrush';
 export { SprayBrush } from './src/brushes/SprayBrush';
```
ShaMan123, line 27: what is the decision regarding index files?

asturur: If it doesn't make my life simpler i don't add one.
in this case i could write a line to export types and i did.

asturur: i think i can see a typedef file as the index file of the types alone.
Since i try to differentiate types from code exports, in this  case a file alone serve the purpose well

PR #8914, opened 2023-05-13
chore(TS): Export gradient and shadow types

ShaMan123:
## Description

Makes gradient and shadow types available as exports from the main package

### DIFF
--- a/fabric.ts
+++ b/fabric.ts
@@ -10,6 +10,7 @@ export { Observable } from './src/Observable';
 export { StaticCanvas } from './src/canvas/StaticCanvas';
 export { Canvas } from './src/canvas/Canvas';
 
+export type { XY } from './src/Point';
 export { Point } from './src/Point';
 export { Intersection } from './src/Intersection';
 export { Color } from './src/color/Color';
--- a/fabric.ts
+++ b/fabric.ts
@@ -14,7 +14,7 @@ export { Point } from './src/Point';
 export { Intersection } from './src/Intersection';
 export { Color } from './src/color/Color';
 
-export { Gradient } from './src/gradient/Gradient';
+export { Gradient } from './src/gradient';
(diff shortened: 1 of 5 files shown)


fabric.ts
```diff
@@ -10,13 +10,15 @@ export { Observable } from './src/Observable';
...
+export type { XY } from './src/Point';
 export { Point } from './src/Point';
 export { Intersection } from './src/Intersection';
 export { Color } from './src/color/Color';
 
-export { Gradient } from './src/gradient/Gradient';
+export * from './src/gradient';
 export * from './src/Pattern';
 export { Shadow } from './src/Shadow';
+export type { SerializedShadowOptions } from './src/Shadow';
```
ShaMan123, line 21: Naming and exporting are inconsistent.
Should we capitalize or not? `./src/gradient` OR `./src/Gradient`, ame goes to shadow and pattern.
Do you think it is ok to nest these 3 under a folder? called Drawables or renderables? RenderingEffects?
I think they should.

ShaMan123: should shadow have n index file and be turned into a folder?

asturur: I mean are inconsistent yes, but just in the index file. i can survive with that.
I m used that classes and classes files are capitalized, not sure if is universal.
Folders can be lowercase or uppercase i guess, i don't really mind if a class folder is not upper case.
other folders should start with lower case.

I would love still to differentiate export of code from export of types.
I don't mind exporting extra types, but i want to see what i export as code, in this case `export * from './src/gradient';` masks that and make me feel uncomfortable.

asturur (approved):

fabric.ts
ShaMan123: I am used yo the same
Let's enforce it then.

PR #8915, opened 2023-05-13
chore(TS): export util types

ShaMan123:
## Motivation

More type exports

## Changes

- move `src/util/fireEvent` =>  `src/controls/fireEvent`
- rename `src/util/type` =>  `src/controls/typeAssertions`


fabric.ts
```diff
@@ -5,12 +5,16 @@ export { config } from './src/config';
...
+
+export * from './src/EventTypeDefs';
 export { Observable } from './src/Observable';
 
 export { StaticCanvas } from './src/canvas/StaticCanvas';
 export { Canvas } from './src/canvas/Canvas';
 
 export { Point } from './src/Point';
+export type { IntersectionType } from './src/Intersection';
 export { Intersection } from './src/Intersection';
```
ShaMan123, line 18:
```suggestion
export * from './src/Intersection';
```

ShaMan123 (commented):
What about * exports?
You said you are against, right?

ShaMan123 (commented):
look at `src/util/misc/svgParsing.ts`
What should be exported etc.?
Please decide about enums. Do we want/need them?
IMO string types are more straight forward so I don't see a need for them.

ShaMan123 (commented):
If you disapprove * exports I will refactor

ShaMan123:
Feel free to commit changes

asturur:
> What about * exports? You said you are against, right?

I m not against, i would like to divide exports of types from the one of code.
That can be done with a typedef file, a named export type, or 2 export from the same file

asturur:
Trying to figure out if again code bloat is because of what.
I can't see any new code

Still the bundle grows. New types shouldn't even introduce new variables that could lead to move some commonly used variable from 2 letter to 3 letter because we finished the combinations of 2.

How this stuff works?

asturur:
ok we are exporting some new stuff as code that shouldn't be part of the interface probably. i m not sure.

src/util/index.ts
```diff
@@ -65,12 +66,17 @@ export {
   sendObjectToPlane,
 } from './misc/planeChange';
 export * as string from './lang_string';
+export type {
+  EnlivenObjectOptions,
+  LoadImageOptions,
+} from './misc/objectEnlive';
```
asturur, line 72: here is where i start to have some doubts.
How do i access those types then? Are they on the main export root or do i need and grab them from inside util?

asturur:
For how we are use enums, enums are useless.
I don't think they are supposed to be compared to string unions, but they need to replace unreadable code.

So for example:
```ts
const enum configurationBits = {
  vgaScaler: 1,
  hdmi96k: 2,
  usbHub: 4,
  stereomode: 8,
  syncOnGreen: 16,
}
```

So that then in the code you can use

```ts
configurationBits.vgaScaler & configurationBits.hdmi96k
```

And then you know that at compilation time it goes back to

```
1 & 2
```

That we would consider unreadable and would make every configuration that uses numbers the same to the other.

Enums are more straight forward to use to me compared to strings union, but are the wrong tool for the job to pick up from a choice of strings, so we shouldn't use them to differentiate between jpg/png for example.

I don't think we have a valid use case for enums in our codebase right now

asturur:
@ShaMan123 do you think it make sense ot use the type or typedef file as index file for types?

Should types be all first level exports even if we have utils and controls objects exports?

asturur:
so the extra code was indeed us exporting the enums as code.
And that is the reason why i prefer to have types separated from code in import / exports.
We want to export every type we add, we don't always want to export all the code because we don't want to support anything we export once and keep it there because then is breaking changes.

asturur:
let's merge as it is, since it has no bad side effects.
But i think when we have all types exported i want to give a stab at proposing a more strict organization, with types all at first level exports.

asturur (approved):

ShaMan123:
> @ShaMan123 do you think it make sense ot use the type or typedef file as index file for types?
> 
> Should types be all first level exports even if we have utils and controls objects exports?
> 


I am not sure about top level type exports. IMO it makes sense to have types at the same level as what they are describing. For discovery.
However, I am not sure how to import a type nested under a named export like util. Can I spread the import and import it directly? If not, it is inconvinient so I would prefer top level exports it that case or expose utils as an entry point 'fabric/util'. Actually that might be the best solution.
Entry points for 
- shapes
- Brushes
- Util
- Controls
- Filters
- Etc.

But then we need to think how 'fabric/node' fits into this pattern.

What are your thoughts and preferences?

ShaMan123:
> so the extra code was indeed us exporting the enums as code.
> And that is the reason why i prefer to have types separated from code in import / exports.
> We want to export every type we add, we don't always want to export all the code because we don't want to support anything we export once and keep it there because then is breaking changes.

This is a good point.

ShaMan123:
> But i think when we have all types exported i want to give a stab at proposing a more strict organization, with types all at first level exports.

Ok.

ShaMan123:
Thinking a bit more about entry points.
If we go in that direction we can split fabric into submodules. Even in npm. That might prove to be good. And also promote plugins etc.

PR #8916, opened 2023-05-13
feat(Color) Regex and color code cleanup and new color format support

asturur:
## Motivation

Improve the regexes and the documentation around them.

So the part with strings and composition was definitely the most readable.
But took 400 bytes of extra code where usually we fit a small feature or a fix or 2.
Now we don't read regexes as a flow, we want to be able to know what they do and i wrote extensive docs for them

## Changes
Following

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb
and
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl
- improve the regex by adding the ability to parse the alpha channel from the format `80%`
- add the ability to recognize the format with spaces and not commas and the `/` for the alpha channel
- add comments, lot of them


src/color/Color.ts
```diff
@@ -279,7 +279,7 @@ export class Color {
    * @return {TRGBAColorSource | undefined} source
    */
   static sourceFromRgb(color: string): TRGBAColorSource | undefined {
-    const match = color.match(reRGBa);
+    const match = color.match(new RegExp(reRGBa, 'i'));
```
asturur, line 282: the truth is that the best code is the literal regex inside the function here.
The constants outside the file are an organization thing that we pay with having those in the scope of the library, always initialized.

asturur:
changed my mind.
I ll overcome the readibility with chatgpt aided comments

asturur:
So the part with strings and composition was definitely the most readable.
But took 400 bytes of extra code where usually we fit a small feature or a fix or 2.
Now we don't read regexes as a flow, we want to be able to know what they do and i wrote extensive docs for them

src/color/Color.ts
```diff
@@ -326,14 +324,14 @@ export class Color {
...
+    const match = color.match(reHSLa());
     if (!match) {
       return;
     }
 
     const h = (((parseFloat(match[1]) % 360) + 360) % 360) / 360,
-      s = parseFloat(match[2]) / (/%$/.test(match[2]) ? 100 : 1),
-      l = parseFloat(match[3]) / (/%$/.test(match[3]) ? 100 : 1);
+      s = parseFloat(match[2]) / 100,
+      l = parseFloat(match[3]) / 100;
```
asturur, line 334: s and l are always percentages only.
Non percentages aren't ammitted.

asturur:
At this point this needs some extra test for the new format covered

src/color/Color.ts
```diff
@@ -53,9 +56,10 @@ export class Color {
    * @param {Number} r Red color value
    * @param {Number} g Green color value
    * @param {Number} b Blue color value
+   * @param {Number} a Alpha color value pass through
    * @return {TRGBColorSource} Hsl color
    */
-  _rgbToHsl(r: number, g: number, b: number): TRGBColorSource {
+  _rgbToHsl(r: number, g: number, b: number, a: number): TRGBAColorSource {
```
asturur, line 62: this should probably be an utility like hsltorgb

asturur:
ok i can stop here with this.
Ready to review if some code choice isn't great

asturur:
@Lazauya since you had previously worked with regexes, have a look for suggestions

src/color/constants.ts
ShaMan123, line 1: This is great!

src/color/constants.ts
```diff
@@ -1,26 +1,107 @@
...
+ * The alpha channel can be in the format 0.4 .7 or 1 or 73%
+ *
+ * WARNING this regex doesn't fail on off spec colors. it matches everything that could be a color.
+ * So the spec does not allow for rgba(30 , 45%  35, 49%) but this will work anyway for us
  */
-// eslint-disable-next-line max-len
-export const reRGBa =
-  /^rgba?\(\s*(\d{1,3}(?:\.\d+)?%?)\s*,\s*(\d{1,3}(?:\.\d+)?%?)\s*,\s*(\d{1,3}(?:\.\d+)?%?)\s*(?:\s*,\s*((?:\d*\.?\d+)?)\s*)?\)$/i;
+export const reRGBa = () =>
+  /^rgba?\(\s*(\d{0,3}(?:\.\d+)?%?)\s*[\s|,]\s*(\d{0,3}(?:\.\d+)?%?)\s*[\s|,]\s*(\d{0,3}(?:\.\d+)?%?)\s*(?:\s*[,/]\s*(\d{0,3}(?:\.\d+)?%?)\s*)?\)$/i;
```
ShaMan123, line 55: does it over complicate reusing some string e.g. `\(\s*(\d{0,3}(?:\.\d+)?%?)\s*[\s|,]\s*` in a string template?

src/color/util.ts
```diff
@@ -21,12 +23,58 @@ export function hue2rgb(p: number, q: number, t: number): number {
...
+  }
+
+  return [Math.round(h * 360), Math.round(s * 100), Math.round(l * 100), a];
+};
+
+export const fromAlphaToFloat = (value = '1') =>
+  parseFloat(value) / (value.endsWith('%') ? 100 : 1);
 
 /**
  * Convert a value ∈ [0, 255] to hex
```
ShaMan123, line 77:
```suggestion
 * Convert a value in the inclusive range [0, 255] to hex
```

You said in the past this is too mathy

src/color/util.ts
```diff
@@ -21,12 +23,58 @@ export function hue2rgb(p: number, q: number, t: number): number {
...
  */
-export function hexify(value: number) {
-  const hexValue = value.toString(16).toUpperCase();
-  return hexValue.length === 1 ? `0${hexValue}` : hexValue;
-}
+export const hexify = (value: number) =>
+  Math.min(value, 255).toString(16).toUpperCase().padStart(2, '0');
```
ShaMan123, line 80: padStart, who knew. 👍

test/unit/color.js
```diff
@@ -133,187 +133,109 @@
...
+  }].forEach(({ name, stringToParse, expectedSource }) => {
+    QUnit.test(name, function(assert) {
+      var oColor = fabric.Color.fromRgba(stringToParse);
+      assert.ok(oColor);
+      assert.ok(oColor instanceof fabric.Color);
+      assert.deepEqual(oColor.getSource(), expectedSource);
+      var oColorUppercase = fabric.Color.fromRgb(stringToParse.toUpperCase());
+      assert.ok(oColorUppercase);
+      assert.ok(oColorUppercase instanceof fabric.Color);
+      assert.deepEqual(oColorUppercase.getSource(), expectedSource);
```
ShaMan123, line 202: seems this function can be reused in the hsl loop as well

ShaMan123 (commented):
This PR is a GREAT cleanup!
Some recommended changes

src/color/Color.ts
```diff
@@ -108,57 +68,52 @@ export class Color {
...
 
   /**
    * Returns color representation in HEX format
    * @return {String} ex: FF5555
    */
   toHex() {
-    const [r, g, b] = this.getSource();
-    return `${hexify(r)}${hexify(g)}${hexify(b)}`;
+    const fullHex = this.toHexa();
+    return fullHex.slice(0, 6);
```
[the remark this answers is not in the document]
asturur: was like this before, trying to not change behaviour since it doesn't seems important.
I don't know the reasons, we should dig into git history

src/color/Color.ts
```diff
@@ -186,13 +139,9 @@ export class Color {
...
-      average = parseInt(
-        (source[0] * 0.3 + source[1] * 0.59 + source[2] * 0.11).toFixed(0),
-        10
-      ),
-      currentAlpha = source[3];
-    this.setSource([average, average, average, currentAlpha]);
+    const [r, g, b, a] = this.getSource(),
+      average = Math.round(r * 0.3 + g * 0.59 + b * 0.11);
+    this.setSource([average, average, average, a]);
     return this;
```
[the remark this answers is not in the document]
asturur: this would be a possible change, and we can do that outside the cleanup/refactor scope.
Something that also should be done maybe that is part of cleanup is consolidathing this average code in a function util

src/color/Color.ts
```diff
@@ -310,14 +251,14 @@ export class Color {
    * @see http://http://www.w3.org/TR/css3-color/#hsl-color
    */
   static sourceFromHsl(color: string): TRGBAColorSource | undefined {
-    const match = color.match(reHSLa);
+    const match = color.match(reHSLa());
     if (!match) {
       return;
     }
 
     const h = (((parseFloat(match[1]) % 360) + 360) % 360) / 360,
```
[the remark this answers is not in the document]
asturur: i wonder... this seems correct and i noticed was off.
I m trying to think of some non obvious value that could break.
But i prefer to discover in this case and in case fix later because this seems wrong

src/color/Color.ts
```diff
@@ -358,31 +299,19 @@ export class Color {
...
-      return [
-        parseInt(r, 16),
-        parseInt(g, 16),
-        parseInt(b, 16),
-        parseFloat((parseInt(a, 16) / 255).toFixed(2)),
-      ];
+        isShortNotation = value.length <= 4;
+      let expandedValue: string[];
+      if (isShortNotation) {
+        expandedValue = value.split('').map((hex) => hex + hex);
```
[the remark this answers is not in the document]
asturur: nope. that is longer for no reason.
those are 2 strings, and added are concatenated, is legit and s easier to see than a template that is 50% operators and 50% var names.

src/color/Color.ts
```diff
@@ -358,31 +299,19 @@ export class Color {
...
+      } else {
+        expandedValue = value.match(/.{2}/g)!;
+      }
+      const [r, g, b, a = 255] = expandedValue.map((hexCouple) =>
+        parseInt(hexCouple, 16)
+      );
+      return [r, g, b, a / 255];
```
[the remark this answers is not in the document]
asturur: Yes but do you think that is functional? why we want alpha reduced in precision? all the output function take care of that.
We need to round in export for hex maybe but not internally when parsing doesn't make sense.
I ll take as an action item that outputs are never wrong

src/color/constants.ts
asturur: in my head there was the fun experiment that 150 colors can be represented by 2 byte in base64 encoding. And so that should be possible to reduce this map to have custom hashed keys to which color names resolve univocally. But i m not sure is easy or hard, but for sure would be fun.

src/color/constants.ts
```diff
@@ -1,26 +1,107 @@
...
+  /^hsla?\(\s*(\d{1,3})\s*[\s|,]\s*(\d{1,3}%)\s*[\s|,]\s*(\d{1,3}%)\s*(?:\s*[,/]\s*(\d*(?:\.\d+)?%?)\s*)?\)$/i;
 
 /**
  * Regex matching color in HEX format (ex: #FF5544CC, #FF5555, 010155, aff)
- * @static
- * @field
- * @memberOf Color
  */
-export const reHex = /^#?([0-9a-f]{8}|[0-9a-f]{6}|[0-9a-f]{4}|[0-9a-f]{3})$/i;
+export const reHex = () => /^#?(([0-9a-f]){3,4}|([0-9a-f]{2}){3,4})$/i;
```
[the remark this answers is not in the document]
asturur: i didn't do that i was tired. I should follow up

test/unit/color.js
asturur: could but i didn't becuase  `fromRgb` vs `fromHsl` and didn't want to overcomplicate it. Still better than all different tests.

src/color/util.ts
asturur: This is exactly where we should verify rounding is not an issue.

src/color/constants.ts
asturur: the issue with that is that i loose the ability to use the literal regex and i have to go with newRegexp.
I started with that if you look at the first commits. I got 400 extra bytes of code and i was pissed off.
I should try again now that i made the regex slightly different.

asturur:
action items:
- [ ] average function
- [ ] see if i can fit the template in the rgba regex ( hsla not worth )
- [ ] verify output is always trimmed correctly if alpha isn't rounded during parsing

src/color/Color.ts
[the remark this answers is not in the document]
ShaMan123: Ahhh I see.
What if the value is < -360. Now is fails, before it was safe.
Please revert this.
I didn't think of that.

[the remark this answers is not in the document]
asturur: Yeah that is fine, we discovered.

src/color/constants.ts
ShaMan123: So desu, ne?!

test/unit/color.js
ShaMan123: thought of 
```
function runColorTest({ name, stringToParse, expectedSource }, from, fromAlpha) {
      var oColor = fromAlpha(stringToParse);
      assert.ok(oColor);
      assert.ok(oColor instanceof fabric.Color);
      assert.deepEqual(oColor.getSource(), expectedSource);
      var oColorUppercase = from(stringToParse.toUpperCase());
      assert.ok(oColorUppercase);
      assert.ok(oColorUppercase instanceof fabric.Color);
      assert.deepEqual(oColorUppercase.getSource(), expectedSource);
}

```

src/color/util.ts
```diff
@@ -74,7 +74,20 @@ export const fromAlphaToFloat = (value = '1') =>
...
+ * Calculate the grey average value for rgb and pass through alpha
+ */
+export const greyAverage = (
+  r: number,
+  g: number,
+  b: number,
+  a = 1
+): [average: number, alpha: number] => [
+  Math.round(r * 0.3 + g * 0.59 + b * 0.11),
+  a,
```
ShaMan123: seems alpha doesn't belong here
but I understand you want it the usage to be simple to use (less lines of code)
I suggest then returning a source value.
`[greyAvg, greyAvg, greyAvg, alpha]`

It is a bit weird like this, but not a big deal.

ShaMan123 (dismissed):

src/color/util.ts
asturur: yes maybe we can return the source format, is an internal util, it can do whatever it needs to do to make the code efficient

asturur: we can make it from source to source.
Easier.

asturur:
ok toHex and toHexa had a decimal bug, but now is forced as rounded

asturur:
I think i addressed all comments

PR #8917, opened 2023-05-13
[DRAFT] This is just a reminder of safari 13 extra size

asturur:
Just a built lib to see how much extra code we have for safari 13
This is not mean to be merged for release 6


asturur:
There are still 50 occurrences of 
```js
let aValue = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : 1;
```
That could be removed if i could understand which browser introduces them

ShaMan123:
nice

PR #8918, opened 2023-05-14
chore(TS) remove enums for string unions when enum is not necessary

asturur:
## Motivation

As per discussion, our use of enums isn't justified.
https://github.com/fabricjs/fabric.js/pull/8915#issuecomment-1546829630

Enums aren't supposed to be named dictionary of strings, we have union types for that.

There was also a possible bug in the glPrecision code in which we were transforming an object into an array, without knowing in which order we were loopin through precisions test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- chore(TS): change enums with types [#8918](https://github.com/fabricjs/fabric.js/pull/8918)
 - chore(TS): export gradient types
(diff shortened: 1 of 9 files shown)


asturur:
We could have a second PR in which we explore if common strings used over and over should be a constant.

asturur:
This PR is also probably fixing most of your extra code in the utils types @ShaMan123

asturur:
i m moving forward with this, i don't see possible disagreemts since it spawned from your question.
In case i m fine reverting and re-opening.

ShaMan123 (approved):

PR #8919, opened 2023-05-14
ci() Introducing JEST and basic tests

asturur:
## Motivation

Today Qunit it is what it is and in order to move to testing functions rather than the whole fabricJS package, we need to refactor our test files.

Is probably possible to better undertand QUnit and just import the specific file we want to test and run scoped tests, but at work we mostly use JEST and we are probably more productive with it.

Now this brings in patterns and decisions and opinions.

I opened this PR with the same pattern i use every day at work

- test file located with the code file
- files named .test.ts or .spec.ts
- simple unit tests when possible
- using babel as transpiler compared to node-ts

i followed the jest tutorial to get this setup runnig.

My proposal is to:
- new tests are just written in JEST
- tests run only in node
- priority is given to unit tests that do not need to use the canvas, but i configured jest with jest-JSDOM anyway.
- tests that requires rendering and browser check will still run in QUNIT and probably are only the visual test suite
- find a way to combine the 2 coverages reports ( or deactivate the one from jest and cover everythin with nyc )
- phase out the old unit test suite slowly
- remember the name of that Test suite that is fully JEST compatible but way faster and use that instead

PROS:
- everyone knows JEST
- this is a plain vanilla setup, no overhead
- a way like another to migrate way from QUNIT

CONS:
- JEST gets slow


asturur:
This is here for discussion and best practices.
I know Jest get slow fast, and there is an alternative that uses esbuild i would like to check on that

asturur:
Ok this is up for discussion.
I m not attached to it but i would like something like this with a pretty much vanilla setup for ts and node.
It doesn matter is Jest as long as i don't need a new runner and as long as the api is popular and estabilished ( no new shills )

asturur:
Well coverage didn't work. i suppose runInBand?

asturur:
i don't know on my local works, coverage bumbs to 83.92 while on github action something is off.
I m picking up coverage in a temp dire and then merging it in the old folder nyc_ouput.

Is not supoer important before evaluating the approach anyway

File                                   | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
---------------------------------------|---------|----------|---------|---------|-----------------------------------------------------------------------------------------------------------------------------
All files                              |   83.92 |    74.06 |   84.44 |   83.76 |
 fabric.js                             |   14.28 |      100 |       0 |   14.28 |
  index.node.ts                        |   14.28 |      100 |       0 |   14.28 | 17-35
 fabric.js/src                         |   96.46 |    80.57 |   92.62 |   97.14 |
  ClassRegistry.ts                     |     100 |      100 |     100 |     100 |

src/brushes/CircleBrush.test.ts
ShaMan123, line 1: This is very nice

ShaMan123 (commented):
I am not for jest because it is falling behind in esm support.
I am worried it will reach its course because of this issue.
IMO this is a no go for us.

Slow when analyzing what to run?

Can you, for the sake of discussion think of a need to mock other modules in fabric for some test? Because that is one of the thigs jest will not be able to do if I understand correctly. Or does babel fix that?
At least we should have a test that mocks some class methods w/o using mocks and not mutatuion.

ShaMan123:
I heard of vitest this week (asked people at work since they use jest heavily with mocking etc.)
Have no idea of its worth.
I can't find from the back of my mind a test that needs/uses stubbing. Might be since it wasn't an option so we did other stuff instead. 
Looked around a bit. This list is not what you asked for unfortunately.


- A simple test: `test/unit/cache.js` mocking importing cache
- Mocking a method on an instance: https://github.com/fabricjs/fabric.js/blob/9e856bca06fcfd9234d4fc4e1b9a9de189935267/test/unit/canvas_static.js#L443
- Spying a class method (search for `sinon` in the tests): https://github.com/fabricjs/fabric.js/blob/9e856bca06fcfd9234d4fc4e1b9a9de189935267/test/visual/text.js#L616-L623
- Mocking an object for a specific purpose (svg output): https://github.com/fabricjs/fabric.js/blob/9e856bca06fcfd9234d4fc4e1b9a9de189935267/test/unit/canvas_static.js#L130-L132
- `test/unit/env.js`
- Stubbing canvas: https://github.com/fabricjs/fabric.js/blob/9e856bca06fcfd9234d4fc4e1b9a9de189935267/test/unit/object_interactivity.js#L207-L209
- In the visual cb so not a good candidate for jest to test: https://github.com/fabricjs/fabric.js/blob/9e856bca06fcfd9234d4fc4e1b9a9de189935267/test/lib/visualTestLoop.js#L137

ShaMan123:
So maybe it is non blocking...

ShaMan123:
could you (or should I) make #8738 pick these tests up as well and see it works.
I understand that you want to create a separation between pure logical stuff to visual/interaction stuff but for the sake of the POC and in case we decide otherwise.

ShaMan123:
importing and snapshots in tests is extremely exciting

ShaMan123:
Also I would us to come up with a convention for visual test etc.
Should they remain in different files? Since now we can use imports they don't have to. However for test separation they should. But it can be nice to have logical, visual and interaction tests related to the same subject being in the same folder/file.
Can expose a global `describe.visual` and/or `test.visual` instead of how visual tests are written today?
Should we have a different ext name? e.g. `.visual.test.js`?
Should we write tests in TS? Will it have any value for type checking? I think we can write in TS but NOT enforce it so we stay sane. What does that mean regarding type tests?

asturur:
> could you (or should I) make #8738 pick these tests up as well and see it works. I understand that you want to create a separation between pure logical stuff to visual/interaction stuff but for the sake of the POC and in case we decide otherwise.

You can, but i don't want change test runner.
I want remove test runners.
JEST + PLAYWRIGTH would cover both bases and avoid us having to deal with a test runner

asturur:
> importing and snapshots in tests is extremely exciting

Yes another important point is that we want to test functions without having to put them in the bundle or test them from the bundle.
While current visual tests will be our sort of higher level test, that run from the package

asturur:
i have no idea what `types test` are in general, because i don't understand why they are needed on top of TSC.
If you think are important, you can totally add them as long as the setup is simple.
There is limited amount of information my brain can process per day and i don't want to focus on those, but you can add them if you want them.

asturur:
The issue that remains with visual tests is that we need to run them in 2 different suites.
In jest for node, and in playwright for the browser.
This makes it harder because the tests are written in a different way.
In that regard qunit is still useful but the intention is to phase it away.

With this PR i don't want to settle the conventions for all tests, i just want to see if JEST is good to replace QUNIT for all tests that are logic only and do not relate to events or canvas, and remove those from the browser.

This is the scope with which i opened this PR.

ShaMan123:
I am ok with merging this and moving forward.
We handle whatever comes next.

ShaMan123:
Regarding visual tests, I am hoping we will find a decent way to solve it and believe we will.

asturur:
In order to merge this i need to understand what is going on with coverage.
Where is stored the action @ShaMan123 so i can look at what it does?

ShaMan123:
.github you mean? In the build action I think

ShaMan123:
https://github.com/fabricjs/fabric.js/blob/9414434d845ea01594f2f588d64367e8f99ea5b4/.github/workflows/build.yml#L92-L109

ShaMan123:
This test is wasteful. We can use the existing test action and upload the coverage outputs as an artifact and then run the report to combine both browser and node coverage

jest.config.js
```diff
@@ -0,0 +1,194 @@
...
+  // bail: 0,
+
+  // The directory where Jest should store its cached dependency information
+  // cacheDirectory: "/private/var/folders/gt/ttb4nqb55fg11kc65xjd54r40000gp/T/jest_dy",
+
+  // Automatically clear mock calls, instances, contexts and results before every test
+  clearMocks: true,
+
+  // Indicates whether the coverage information should be collected while executing the test
+  collectCoverage: false,
```
ShaMan123, line 20: why is this false?

jest.config.js
```diff
@@ -0,0 +1,194 @@
...
+  // ],
+
+  // Indicates which provider should be used to instrument code for coverage
+  // coverageProvider: "babel",
+
+  // A list of reporter names that Jest uses when writing coverage reports
+  coverageReporters: [
+    'json',
+    // "text",
+    // 'lcov',
```
ShaMan123, line 40: did you try lcov? I think the action uses that

asturur:
i have to add the globals or JEST before being able to merge

jest.config.js
asturur: becasue we collect coverage with nyc for qunit, and is the only way to mix it it seems. So i disabled the internal coverage reporter of Jest

src/util/internals/cloneDeep.test.ts
```diff
@@ -0,0 +1,15 @@
...
+  it('clones an object deeply', () => {
+    const testObject = {
+      a: {
+        b: 3,
+      },
+      array: [1, 2, 3, 4],
+      test: '123',
+    };
+    const cloned = cloneDeep(testObject);
+    expect(cloned).toEqual(testObject);
```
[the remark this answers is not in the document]
ShaMan123: how does it output a diff?
I want to see it is better than QUnit dumps (I am sure it is)

src/util/internals/getRandomInt.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    global.Math.random = jest.fn(() => 0.1);
+  });
+  afterAll(() => {
+    global.Math.random = originalMathRandom;
+  });
+  it('return a number between min and max', () => {
+    (global.Math.random as jest.Mock).mockReturnValue(0.1);
+    const semiRandom = getRandomInt(0, 10);
+    expect(semiRandom).toBe(1);
+  });
```
ShaMan123, line 16: I don't understand why the beforeAll/afterAll if this test is using mocks
I think we should not do that anymore and use only mocks, stubs, spies whatever

ShaMan123 (commented):
What is the status with coverge? Did you try lcov as well?
The action needs lcov reports.

asturur:
> What is the status with coverge? Did you try lcov as well? The action needs lcov reports.

For now we don't have it.
I ll see if i do that or more TS or more website.

src/util/internals/cloneDeep.test.ts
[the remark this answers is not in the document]
asturur: toBe is the reference and toEqual is deepEqual, i can add the toBe

src/util/internals/getRandomInt.test.ts
asturur: i don't think you can mock random differently than this.
Because Math is not an import.

asturur:
> What is the status with coverge? Did you try lcov as well? The action needs lcov reports.
All the coverage is generated with NYC, so the format is the same as before ( LCOV ).
Locally it produces good coverage but in the GHA it doesn't.

src/util/internals/getRandomInt.test.ts
ShaMan123: So i would like to add sinon
Sinon dies that so easily

PR #8930, opened 2023-05-18
refactor(): Canvas dom handling delegation to utility class

ShaMan123:
## Motivation

Mine/Our discomfort of what is going on with prototype chains

## Description

This is a POC here for discussion, show casing the concept of flattening the prototype chain of canvas into delegates/services/managers (call them how'd you like).
Each of them should have a well defined scope.
The final purpose being not having a prototype chain at all. Not for canvas nor for objects. Or at least minimizing significantly. So that subclassing, customizing etc. is more straight forward.
It should also make the code more easy to understand because of scoping.
I chose to begin with canvas deliberately to prove it can be done and prove it has value.
Took a lot more time than I thought.
Part of the concept is to make these changes only internal and non breaking. Then when the time is right swap everything, break and bump a version.

## Changes

**No changes to exposed interfaces, only to internals - NOT BREAKING**

- extracted all elements logic to ElementsManager
- extracted target find logic to `FindTarget` and exposed `renderTarget`
- initializing retina scaling now happens only on backstore size change (was firing always before) - Is that correct?
- removed redundant size initializing since it happens when canvas calls `setDimensions`
- Naming is not final


src/canvas/ElementsManager/util.ts
ShaMan123, line 1: This util might belong under the util dir

ShaMan123 (commented):
Suggest order of review:
1. `src/canvas/ElementsManager/StaticCanvasElements.ts`
2. `src/canvas/ElementsManager/util.ts`
3. `src/canvas/StaticCanvas.ts`
4. `src/canvas/ElementsManager/CanvasElements.ts`
5. `src/canvas/SelectableCanvas.ts`
6. the rest

ShaMan123 (commented):
I am happy with this.
I believe it is ready for reviewing.

ShaMan123 (commented):
done

ShaMan123:
> **Build Stats**
> 
> fabric	917.475 (**+2.257**)	304.881 (**+0.805**)

image

ShaMan123:
master
image

src/canvas/ElementsManager/CanvasElements.ts
```diff
@@ -0,0 +1,119 @@
...
+  setStyle,
+} from '../../util';
+import type { CSSDimensions } from '../../util/dom_misc';
+import { allowTouchScrolling } from '../../util/dom_misc';
+import { setCSSDimensions } from '../../util/dom_misc';
+import type { CanvasItem } from './StaticCanvasElements';
+import { StaticCanvasElements } from './StaticCanvasElements';
+import { setCanvasDimensions } from './util';
+
+export class CanvasElements extends StaticCanvasElements {
```
asturur, line 15: CanvasElement would be like a CanvasDOMManager?

ShaMan123: Yes
A better name indeed

asturur:
I didn't read the code deeply, but apart naming changes i think the targetFind part is rushed since it could fit in other kinds of changes, like the selection code and is not worth doing now.
Better one topic at time imho.
The DOM one is small enough to be tested, but we can't do more of this kind of changes now, i really want to go to a release.

The layout logic of group is exactly something i wanted to refactor in this style.

ShaMan123:
great
yeah I thought to leave out target find. It will be easy to revert:
- [0360770](https://github.com/fabricjs/fabric.js/pull/8930/commits/036077076fd4a70010255a90008f499ba222a011)
- [bca2355](https://github.com/fabricjs/fabric.js/pull/8930/commits/bca235587e57ef7aaea4d9b6b8bc29f37d436d01)

ShaMan123 (commented):
- reverted target find
- renamed to DOMManager
- exported 
- added dep warning on `Canvas#containerClass`

asturur:
I need to figure out a couple of things then i can merge.
1) why this code can't be minified as before, ok 2 new getters and some chained property access ( this.that.those.these ) but 576 byte extra just for code shuffling seems strange to me

2) some util that is used exclusively for canvas dom manipulation can still be moved

I ll look at it tomorrow or later today

asturur:
I moved all the utilities that are used only by the dom manager in the dom manager folder.
Removed makeELemenentSelectable/unselectable from exports

PR #8931, opened 2023-05-18
fix(): `typeAssertions` not respecting subclasses

ShaMan123:
## Motivation

Was working on a subclass of IText when selection rendering started having bugs, being cleared because passing this condition (should pass ` isInteractiveTextObject` and `isEditing`):
https://github.com/fabricjs/fabric.js/blob/0f0ed963b959969e1968563b176e9e67a3d0feb5/src/canvas/Canvas.ts#L890-L896

https://github.com/fabricjs/fabric.js/blob/0f0ed963b959969e1968563b176e9e67a3d0feb5/src/util/typeAssertions.ts#L54-L60

Of course it will because types are deprecated and the name of the class doesn't match.

Indeed this line is more types than logic but the rest aren't

## Description

Came up with a solution in the form of static TAGS and an `is` method that is extremely similar to `isType`

## Gist

There is no decent way for me to test this

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): `typeAssertions` not respecting subclasses [#8931](https://github.com/fabricjs/fabric.js/pull/8931)
 - fix(TS): extending canvas and object event types (`type` => `interface`) [#8926](https://github.com/fabricjs/fabric.js/pull/8926)
 - chore(build) simple deps update [#8929](https://github.com/fabricjs/fabric.js/pull/8929)
 - fix(Canvas): sync cleanup of dom elements in dispose [#8903](https://github.com/fabricjs/fabric.js/pull/8903)
--- a/src/gradient/Gradient.ts
+++ b/src/gradient/Gradient.ts
@@ -203,7 +203,7 @@ export class Gradient<
     }
     if (
       object instanceof BaseFabricObject &&
-      object.isType('Path') &&
+      object.is('path') &&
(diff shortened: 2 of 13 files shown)


asturur:
mmm this is like isType but different.
The solution would be instanceOf that we are not using because it makes impossible to split the classes for which we have a type assertion.

I would prefer we detect the classes by a specific class property instead of having our own tag system.

So the bug is that this isType can't work because of course your subclass name is different.
an text interactive object has the `enterEditing` method for example and i m sure the activeSelection has something like that.
Can't we go on that direction?

I don't want to add custom systems to do this.

ShaMan123:
Hmm...
But then a class can have enterEditing on it but not be a text object.
What about a dummy method then?
`isInteractiveText():void` so we check for that?

ShaMan123:
Well if we manage one day to flatten the object tree instanceof will be possible. But until then...

asturur:
the empty methods could be good enough but is just like a tag.
Right now we have an issue that is that legit subclasses won't work.
So we need to fix.
The class that will have enterEditing but won't be a text class is not an issue of today. We can also take something less generic than enterEditing.

ShaMan123:
I have a different concept, using class registry. I will POC

ShaMan123:
So the concept isn't that good
Simply moving the tag concept to the registry.
I don't like the methods thing because it feels unsafe but I don't have a better concept.

asturur:
@ShaMan123 it feels unsafe but is not.
When is really unsafe we find a fix.
Fix you detection issue using a method, don't refactor all the code around for it.

PR #8933, opened 2023-05-18
chore() Try to use constants for repeated strings.

asturur:
Hopefully less repetitions, less typin errors, less packed code.

This PR will cover `center`, `left`, `top`, `right`, `bottom`, `none`.
I m not sure there is much else

- [x] center
- [x] left
- [x] right
- [x] top
- [x] bottom
- [x] justification


asturur:
ok for now this is done like this.
There are more opportunities for scale and other common strings, but this is already a small cleanup with no functional changes

ShaMan123:
I don't mind
Enforcing it will be hard though, in future prs.
And I think the min build will die in the near far future, since umd will be opt in while es will be defaul.
so I am not sure we should consider minification a priority any more.
That being said, if you think it should enter let it, perhaps add a lint rule for these pieces of text.

asturur:
minification will be done at some other level, by the importing appliacation maybe, but the concept doesn't change, what improve minification here will improve it there.
We will continue to minify as a benchmark at least.

asturur:
Also we don't need to enforce it, if is seen in code review good, if is not seen will be addressed another day

PR #8934, opened 2023-05-19
patch(): `Control#shouldActivate`

ShaMan123:
## Motivation

I need the control api to be more flexible

## Description

Introduce the ability for a control to decide if it should be activated or not.
Before this was handled by the canvas in an hardcoded way with:

```js
      !this.hasControls ||
      !this.canvas ||
      (this.canvas._activeObject as unknown as this) !== this
```

as an early return and later on on the control check if the control is visible.

Now the first early return doesn't take care of the object being selected, but the control later has a new method `shouldActivate` that by default return true only if the object is the active one and if the control is visible.

The developer can override this in a custom control and apply the logic that prefers.

## Changes

- `Control#shouldActivate` - for overriding
- now the control loop and the visibility check run regardless if the object is active or not.

### DIFF
--- a/src/controls/Control.ts
+++ b/src/controls/Control.ts
@@ -169,6 +169,14 @@ export class Control {
    */
   declare mouseUpHandler?: ControlActionHandler;
 
+  shouldActivate(controlKey: string, fabricObject: InteractiveFabricObject) {
+    // TODO: locking logic can be handled here instead of in the control handler logic
+    return (
+      fabricObject.canvas?.getActiveObject() === fabricObject &&
+      fabricObject.isControlVisible(controlKey)
+    );
+  }
+
   /**
    * Returns control actionHandler
    * @param {Event} eventData the native mouse event
--- a/src/shapes/Object/InteractiveObject.ts
+++ b/src/shapes/Object/InteractiveObject.ts
@@ -184,31 +184,27 @@ export class InteractiveFabricObject<
    * @param {boolean} forTouch indicates if we are looking for interaction area with a touch action
    * @return {String|Boolean} corner code (tl, tr, bl, br, etc.), or 0 if nothing is found.
    */
-  _findTargetCorner(pointer: Point, forTouch = false): 0 | string {
-    if (
-      !this.hasControls ||
-      !this.canvas ||
-      (this.canvas._activeObject as unknown as this) !== this
-    ) {
-      return 0;
+  _findTargetCorner(pointer: Point, forTouch = false): string {
+    if (!this.hasControls || !this.canvas) {
+      return '';
     }
 
     this.__corner = undefined;
     // had to keep the reverse loop because was breaking tests
     const cornerEntries = Object.entries(this.oCoords);
     for (let i = cornerEntries.length - 1; i >= 0; i--) {
-      const [cornerKey, corner] = cornerEntries[i];
-      if (!this.isControlVisible(cornerKey)) {
-        continue;
+      const [key, corner] = cornerEntries[i];
+      if (this.controls[key].shouldActivate(key, this)) {
+        const lines = this._getImageLines(
+          forTouch ? corner.touchCorner : corner.corner
+        );
+        const xPoints = this._findCrossPoints(pointer, lines);
+        if (xPoints !== 0 && xPoints % 2 === 1) {
+          this.__corner = key;
+          return key;
+        }
       }
-      const lines = this._getImageLines(
-        forTouch ? corner.touchCorner : corner.corner
-      );
-      const xPoints = this._findCrossPoints(pointer, lines);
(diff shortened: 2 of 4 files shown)


src/shapes/Object/InteractiveObject.ts
```diff
@@ -322,6 +319,15 @@ export class InteractiveFabricObject<
...
   }
 
+  isPointerOverControl(pointer: Point, controlKey: string, forTouch = false) {
+    const corner = this.oCoords[controlKey];
+    const lines = this._getImageLines(
+      forTouch ? corner.touchCorner : corner.corner
+    );
+    const xPoints = this._findCrossPoints(pointer, lines);
+    return xPoints !== 0 && xPoints % 2 === 1;
+  }
```
ShaMan123, line 329: #8767 TODO

ShaMan123 (commented):
Is there a need for a test?
I feel like not adding non crucial tests until jest is in

ShaMan123 (commented):
made it simpler by reverting `isPointerOverControl`

src/shapes/Object/InteractiveObject.ts
```diff
@@ -184,31 +184,28 @@ export class InteractiveFabricObject<
...
-        continue;
-      }
-      const lines = this._getImageLines(
-        forTouch ? corner.touchCorner : corner.corner
-      );
-      const xPoints = this._findCrossPoints(pointer, lines);
-      if (xPoints !== 0 && xPoints % 2 === 1) {
-        this.__corner = cornerKey;
-        return cornerKey;
+    const controlKeys = Object.keys(this.oCoords);
```
asturur, line 194: the idea is fine, but Object.keys vs Object.entries seems a downgrade.
What is the reason for switching to key?

ShaMan123: good point

ShaMan123:
One day I will refactor the control visibility awkward code. But not soon.

src/shapes/Object/InteractiveObject.ts
ShaMan123: a4c0f6282

ShaMan123 (commented):
DONE

src/shapes/Object/InteractiveObject.ts
```diff
@@ -184,31 +184,27 @@ export class InteractiveFabricObject<
...
+      const [key, corner] = cornerEntries[i];
+      if (this.controls[key].shouldActivate(key, this)) {
+        const lines = this._getImageLines(
+          forTouch ? corner.touchCorner : corner.corner
+        );
+        const xPoints = this._findCrossPoints(pointer, lines);
+        if (xPoints !== 0 && xPoints % 2 === 1) {
+          this.__corner = key;
+          return key;
+        }
```
asturur, line 205: looks like that both getImageLines and findCrossPoints are object method that are more utils than anyhitng else.
That some parameter in, do not used `this` at all and give a result.

If you want to have the control class bein the owner of the decision if the cursor is in the control, you can do that moving those to utils and exposing a method that uses the utils.

asturur:
What made you change the idea around the pointer thing? something bad happens when doing that?

ShaMan123:
I am looking at #8767 to guide me.
I don't want to introduce changes here, I prefer to do it there.
I do think the control should be in charge but I am happy with this limited scope and not worrying about merge conflict in #8767

ShaMan123:
https://github.com/fabricjs/fabric.js/commit/a4c0f62827ac8ca3d98624d7891558ac11906c60#diff-fe6477e182894fdb5db40e4d3bffa1f12e02af91a158b9619752aca781441332R199-R201

This can easily be exposed on the control

asturur:
> One day I will refactor the control visibility awkward code. But not soon.

That is there just for backward compatibility not because we like it.
Is probably a good idea to deprecate the current functions so that we inform developers those are going to be changed/removed.

src/controls/Control.ts
```diff
@@ -169,6 +169,14 @@ export class Control {
    */
   declare mouseUpHandler?: ControlActionHandler;
 
+  shouldActivate(controlKey: string, fabricObject: InteractiveFabricObject) {
+    // TODO: locking logic can be handled here instead of in the control handler logic
```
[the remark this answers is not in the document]
ShaMan123: yes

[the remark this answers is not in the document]
ShaMan123: or better off, move the lock property to the control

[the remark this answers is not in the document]
asturur: well lock is by object and controls could be generic.
That is the same issue we have with visibility.

asturur:
The PR is fine, i m merging.
The description is bad i fixed it.

PR #8938, opened 2023-05-20
ci(): automate PR changelog

ShaMan123:
## Motivation

I don't want to keep copying a pr title for each PR I open and commit it.
It is very good to have it scoped to the PR.
So let github do it.

## Description

This bit of CI does the following:
When a PR is opened or the title is changed it updates the changelog as following:

  - check if there is a log
  - if there is and it is identical to the PR title remove it
  - add title in the format of `## [next]` if necessary
  - if there is no log add a log in the format of `- ${pr_title} ${pr_md_link}\n`

This means that we have full control over this action. We can still edit the changelog w/o the action interfering. Once the log doesn't match the PR title, updating the title will not affect the changelog.
We can also checkout the changelog and do whatever we like because once opened the only trigger to this action will be a title update.
Also it is still visible in the PR diff so no surprises.


ShaMan123 (commented):
This bit of CI does the following:
When a PR is opened or the title is changed it updates the changelog as following:

  - check if there is a log
  - if there is and it is identical to the PR title remove it
  - add title in the format of `## [next]` if necessary
  - if there is no log add a log in the format of `- ${pr_title} ${pr_md_link}\n`

This means that we have full control over this action. We can still edit the changelog w/o the action interfering. Once the log doesn't match the PR title, updating the title will not affect the changelog.
We can also checkout the changelog and do whatever we like because once opened the only trigger to this action will be a title update.
Also it is still visible in the PR diff so no surprises.

This is ready for review.
I am happy with it because it keeps to good concept of handling the changelog in each PR and automates it if needed. I have found myself annoyed with the process of opening a PR and then checking the PR number, copying the title, committing, waiting again for test etc. 

Do you any concerns/thoughts?
Updated desc

ShaMan123:
Now I am happy.
You can test it works by changing the PR title, checking out the file etc.

ShaMan123 (commented):
done

asturur:
I think the cost of having another github action to potentially have to look at for issues is bigger than the issue of updating your changelog when your PR is done ( since anyway intent and final PR often end up diverging ).

I m ok adding this even if i think is unnecessary overhead.
But at the first issue that doesn't work or do something it wasn't supposed to do, i ll revert it.

asturur (approved):

ShaMan123:
Fair enough
Thanks

PR #8940, opened 2023-05-20
ci(): Multi chunk build

ShaMan123:
## Description

Rollup config and more to allow dynamic imports if needed.
Leaving this here for reference

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       minified: dist/index.min.js
-      bundled: dist/index.mjs
+      bundled: dist/es/index.js
     steps:
       - uses: actions/checkout@v3
         with:
--- a/package.json
+++ b/package.json
@@ -117,8 +117,8 @@
   "engines": {
     "node": ">=14.0.0"
   },
-  "module": "./dist/index.mjs",
-  "main": "./dist/index.node.cjs",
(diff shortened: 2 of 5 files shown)


asturur:
This PR reminds me i really want to stop rolling up node. Seems completely unnecessary.
What do you think?
reference: https://github.com/fabricjs/fabric.js/pull/8649/

ShaMan123:
I think you are right.
And as you said in that PR we don't need to bundle es for the browser as well. Only transpile to js.
Then one day the umd build is not needed at all.

ShaMan123:
@asturur what do you think about submodules?

ShaMan123:
controls, brushes, filters are great candidates.
Github has a decent concept for it.

ShaMan123:
And it means we could expose them to npm as standalones so a dev can do something like
```ts
import { Canvas } from '@fabric/web' // '@fabric/node'
import { Control } from '@fabric/controls
import { Text } from '@fabric/Text'
import { Circle } from '@fabric/core'
```

What do you think?
Might be an overkill once we stop bundling.

asturur:
i m not sure we ever stop bundling, bundling is not going away anywhere soon. There are benefits of various kinds even when using http2, apart that it doesn't cost us anything to keep a bundle rolling.

At some point a way to import single files from the node_modules folder will be super common and adopted and we can just jump on that.

Right now offering transpiled single files that import each other + a bundle seems the best we can do.

PR #8941, opened 2023-05-20
chore(TS): minor style

ShaMan123:
## Motivation

Improve some typings

### DIFF
--- a/src/shapes/Text/constants.ts
+++ b/src/shapes/Text/constants.ts
@@ -8,7 +8,7 @@ const fontProperties = [
   'fontStyle',
 ] as const;
 
-const textDecorationProperties = [
+export const textDecorationProperties = [
   'underline',
   'overline',
   'linethrough',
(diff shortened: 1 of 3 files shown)


src/shapes/Text/constants.ts
```diff
@@ -8,7 +8,7 @@ const fontProperties = [
   'fontStyle',
 ] as const;
 
-const textDecorationProperties = [
+export const textDecorationProperties = [
```
asturur, line 11: do we need to export this?

asturur: I think maybe is a leftover of trying to get the type automatically from this array?

ShaMan123: Something like that
I used it for the HTML logic

asturur (approved):

asturur:
I added the changelog manually here because for some reason the action can't add it here.
If is all green i ll merge

src/shapes/Text/constants.ts
ShaMan123: Not really

PR #8942, opened 2023-05-20
feat(): support native copy/paste

ShaMan123:

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -53,16 +53,6 @@ class BaseConfiguration {
   minCacheSideLimit = 256;
 
   /**
-   * When 'true', style information is not retained when copy/pasting text, making
-   * pasted text use destination style.
-   * Defaults to 'false'.
-   * @type Boolean
-   * @default
-   * @deprecated
-   */
-  disableStyleCopyPaste = false;
-
-  /**
    * Enable webgl for filtering picture is available
    * A filtering backend will be initialized, this will both take memory and
    * time since a default 2048x2048 canvas will be created for the gl context
--- a/src/shapes/IText/ITextBehavior.ts
+++ b/src/shapes/IText/ITextBehavior.ts
@@ -35,6 +35,9 @@ export type ITextEvents = ObjectEvents & {
   tripleclick: TPointerEventInfo;
   'editing:entered': never | { e: TPointerEvent };
   'editing:exited': never;
+  copy: { e: ClipboardEvent };
+  cut: { e: ClipboardEvent };
+  paste: { e: ClipboardEvent };
 };
 
 export abstract class ITextBehavior<
@@ -917,6 +920,11 @@ export abstract class ITextBehavior<
     start: number,
     copiedStyle?: TextStyleDeclaration[]
   ) {
+    if (copiedStyle && insertedText.length !== copiedStyle.length) {
+      throw new Error(
+        `fabric.js: expected ${insertedText.length} style declarations, found ${copiedStyle.length}`
+      );
+    }
     const cursorLoc = this.get2DCursorLocation(start, true),
       addedLines = [0];
     let linesLength = 0;
@@ -1000,13 +1008,13 @@ export abstract class ITextBehavior<
    * start/end ar per grapheme position in _text array.
    *
    * @param {String} text text to insert
-   * @param {Array} style array of style objects
+   * @param {Array} styles array of style objects
    * @param {Number} start
    * @param {Number} end default to start + 1
    */
   insertChars(
     text: string,
-    style: TextStyleDeclaration[] | undefined,
+    styles: TextStyleDeclaration[] | undefined,
     start: number,
     end: number = start
   ) {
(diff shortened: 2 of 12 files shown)


ShaMan123 (commented):
I have no idea what is going in with firefox tests but I tested it live and it works

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -87,7 +87,7 @@ export abstract class ITextKeyBehavior<
     this.hiddenTextarea.addEventListener('keyup', this.onKeyUp.bind(this));
     this.hiddenTextarea.addEventListener('input', this.onInput.bind(this));
     this.hiddenTextarea.addEventListener('copy', this.copy.bind(this));
-    this.hiddenTextarea.addEventListener('cut', this.copy.bind(this));
+    this.hiddenTextarea.addEventListener('cut', this.cut.bind(this));
```
asturur, line 90: i think there is a precise reason why we sued copy and manual deletion for cut.
I think it was to centralize the logic on onInput.

asturur:
@ShaMan123 copy/paste is not a big part of the core functionalities that i m scared to touch or attached to a particular implementation.

Is unclear to me with this PR what is actually getting better. If we are not gettin tangible measurable improvements ( ex: things that didn't work before now work, or developers can now do things that couldn' do before, or added functionalities ) we should at least try to reduce the code growth

src/config.ts
```diff
@@ -52,16 +52,6 @@ class BaseConfiguration {
...
 
-  /**
-   * When 'true', style information is not retained when copy/pasting text, making
-   * pasted text use destination style.
-   * Defaults to 'false'.
-   * @type Boolean
-   * @default
-   * @deprecated
-   */
-  disableStyleCopyPaste = false;
```
asturur, line 63: this is the second time you try to delete this but you never say why.

src/shapes/IText/ITextBehavior.ts
```diff
@@ -917,6 +920,11 @@ export abstract class ITextBehavior<
     start: number,
     copiedStyle?: TextStyleDeclaration[]
   ) {
+    if (copiedStyle && insertedText.length !== copiedStyle.length) {
+      throw new Error(
+        `fabric.js: expected ${insertedText.length} style declarations, found ${copiedStyle.length}`
+      );
+    }
```
asturur, line 927: This seems something to cover for developers that decide to mess with te copy/paste event and i don't want to do it.
is there a legit reason for this to happen?

asturur:
I need to understand better the preventDefault issue.
If calling preventDefault is necessary and require us to duplicate the onInput logic is harder to see this an improvement.

src/shapes/IText/ITextKeyBehavior.ts
ShaMan123: Might be.
Indeed it breaks off from having the textarea as the source of change.
We can and should break up the onInput logic to a function that generates the diff and has not side effects and the function that applies the diff IMO.

src/shapes/IText/ITextBehavior.ts
ShaMan123: A left over. Can be removed.
Was heard to understand what was going on without it.
But still that is what the method expects and if it receives something else it doesn't throw but the text object will be a mess

ShaMan123:
Again, my fault of a bad description.
> This PR adds support to native copy/cut/paste events

This PR makes it possible to copy and paste things from the clipboard (such as html, meaning copying text with styles from other text editors, copying links etc.). And aligns these events to drag/drop.
I don't recall us touching copy/paste, I wanted this as part of draggable text and extract it for scoping.
Currently you can only copy and paste text in the same fabric instance, the same tab. This is not good UX.
Hence the need to `preventDefault`, w/o that the clipboard data doesn't change (from what I read and tried).
The point of this effort is to make Text feel native, same as textarea. The same reason for draggable text.
Using these events gives the dev full control of what is being copied/pasted, as should be, and support different use cases w/o the need to hack anything:
```ts
text.on('copy', ({ e }) => e.clipboardData.setData('mimeType', data))
```

If you set `data.text` or `data.styles` then they will be handled out of the box. Elsewise, you can listen to the paste event.
That is why I also removed `disableStyleCopyPaste`. It doesn't need to be an app level config. Though it can. In my mind the dev can do the following instead:
```ts
text.on('copy', ({ e }) => e.clipboardData.clearData('application/fabric'))
```
Then the only mime type left is plain text. I think it is better because it is fine grained and not app level.
I wrote copy event, but mean copy/cut.
Regarding using the textarea to keep track of changes... I tried firing a custom event. That has no effect.
I would also argue the need for the textarea is to capture the events and for history. It isn't really needed to keep track of state. IText has enough state and logic in place for that. I would use only to fire events and not for text state syncing. Also since history is broken due to draggable text (unless you find a way to somehow trigger an event on textarea) it seems even less useful using it for state/diff.

I will update desc.

PR #8943, opened 2023-05-21
chore(): reduce redundant code size

asturur:
## Motivation

Removing redundant code in non performance critical spot.
This code could perform worse than the one before but this isn't really a performance impacting area

Just less bulky code in general

### DIFF
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1585,10 +1585,6 @@ export class Canvas extends SelectableCanvas {
     return true;
   }
 
-  exitTextEditing() {
-    this.textEditingManager.exitTextEditing();
-  }
-
   /**
    * @override clear {@link textEditingManager}
    */
--- a/src/shapes/IText/ITextKeyBehavior.ts
+++ b/src/shapes/IText/ITextKeyBehavior.ts
@@ -98,6 +98,7 @@ export abstract class ITextKeyBehavior<
     Object.entries(eventsHandlers).map(([eventName, handler]) =>
       textarea.addEventListener(eventName, this[handler].bind(this))
     );
+    this.hiddenTextarea = textarea;
   }
 
   /**
--- a/src/shapes/IText/ITextKeyBehavior.ts
+++ b/src/shapes/IText/ITextKeyBehavior.ts
@@ -76,10 +76,10 @@ export abstract class ITextKeyBehavior<
     Object.entries(attributes).map(([attribute, value]) =>
       textarea.setAttribute(attribute, value)
     );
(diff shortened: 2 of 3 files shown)


src/canvas/Canvas.ts
```diff
@@ -1585,10 +1585,6 @@ export class Canvas extends SelectableCanvas {
     return true;
   }
 
-  exitTextEditing() {
-    this.textEditingManager.exitTextEditing();
-  }
```
asturur, line 1590: this doesn't seem used by anyone and wasn't in jsdocs either.

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -64,43 +64,41 @@ export abstract class ITextKeyBehavior<
...
-      this.onCompositionStart.bind(this)
-    );
-    this.hiddenTextarea.addEventListener(
-      'compositionupdate',
-      this.onCompositionUpdate.bind(this)
-    );
-    this.hiddenTextarea.addEventListener(
-      'compositionend',
-      this.onCompositionEnd.bind(this)
+    textarea.style.cssText = `position: absolute; top: ${top}; left: ${left}; z-index: -999; opacity: 0; width: 1px; height: 1px; font-size: 1px; padding-top: ${fontSize};`;
```
ShaMan123, line 82: I think we should setStyles here

src/shapes/IText/ITextKeyBehavior.ts
```diff
@@ -64,43 +64,41 @@ export abstract class ITextKeyBehavior<
...
       getFabricDocument();
-    this.hiddenTextarea = doc.createElement('textarea');
-    this.hiddenTextarea.setAttribute('autocapitalize', 'off');
-    this.hiddenTextarea.setAttribute('autocorrect', 'off');
-    this.hiddenTextarea.setAttribute('autocomplete', 'off');
-    this.hiddenTextarea.setAttribute('spellcheck', 'false');
-    this.hiddenTextarea.setAttribute('data-fabric', 'textarea');
-    this.hiddenTextarea.setAttribute('wrap', 'off');
-    const style = this._calcTextareaPosition();
+    const attributes = {
```
ShaMan123, line 67: don't you prefer to inline the arg if it is used only once?

ShaMan123 (dismissed):

src/shapes/IText/ITextKeyBehavior.ts
asturur: sometimes i have some readibility concenrn but here seems fine to do so.

PR #8947, opened 2023-05-22
refactor(): create Filler base class for Pattern/Gradient

ShaMan123:
## Motivation

1. allow `instance of Filler` as part of #8931 
2. share some interface
3. port some of #8263 

## Changes

- moved `isFiller`, `TFiller` under `filler` => caused a lot of import changes
- rm `isPattern` in favor of instance of used only by canvas. We can add a dummy super class if you prefer to do `instanceof` checks against it but importing Pattern is ok, not cycles.
- export Filler base class with min logic/interface

### DIFF
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -5,7 +5,7 @@ import type { CanvasEvents, StaticCanvasEvents } from '../EventTypeDefs';
 import type { Gradient } from '../gradient/Gradient';
 import { createCollectionMixin } from '../Collection';
 import { CommonMethods } from '../CommonMethods';
-import type { Pattern } from '../Pattern';
+import { Pattern } from '../Pattern';
 import { Point } from '../Point';
 import type { BaseFabricObject as FabricObject } from '../EventTypeDefs';
 import type { TCachedFabricObject } from '../shapes/Object/Object';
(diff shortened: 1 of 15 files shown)


ShaMan123 (commented):
Reviewing:
1. `src/fillers/Filler.ts`
2. `src/gradient/Gradient.ts`, `src/Pattern/Pattern.ts`
3. `src/util/typeAssertions.ts`
4. the rest

src/canvas/StaticCanvas.ts
```diff
@@ -1405,20 +1401,27 @@ export class StaticCanvas<
         additionalTransform = shouldInvert
           ? matrixToSVG(invertTransform(this.viewportTransform))
           : '';
+      const { width = finalWidth, height = finalHeight } =
+        filler instanceof Pattern
```
ShaMan123, line 1405: change, importing Pattern

src/gradient/Gradient.ts
```diff
@@ -150,9 +137,9 @@ export class Gradient<
    * @param {string[]} [propertiesToInclude] Any properties that you might want to additionally include in the output
    * @return {object}
    */
-  toObject(propertiesToInclude?: (keyof this | string)[]) {
+  toObject<T extends keyof this>(propertiesToInclude?: T[]) {
```
ShaMan123, line 140: I am not attached to this at all

ShaMan123 (commented):
Not a lot of changes, most files are import changes

ShaMan123:
I experienced a glitch in the PR changelog action. If you open a PR and see it committing twice please inform me (I believe it is a small race condition I am able to fix easily)
After that I will accept it should die

ShaMan123:
Also I wanted to move pattern and gradient under filler but didn't to keep this PR clean

PR #8949, opened 2023-05-22
ci(): fix changelog action race condition

ShaMan123:
## Motivation

I want to keep this action alive and not kill it yet
Last chance

## Description

The action would run twice on open (in case I edited the description/title before the opened event would finish)
Seems to because of an opened and an edited event being fired at the same time, strange that git allowed to commit on top of a changed origin w/o needing to fetch. 

Anyways it should be fixed now.
It also means the action is stricter
It will write a log on PR opened ONLY if there is none existing.
After that it will only **update** (and not try to create the log) in case it is identical to the PR title.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -80,7 +80,6 @@ jobs:
       - name: Commit & Push
         if: fromJson(steps.update.outputs.result)
         run: |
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
ready

PR #8950, opened 2023-05-22
chore(): cleanup Type assertions

ShaMan123:
## Motivation

#8931 

## Description

After inspecting type assertions in depth I found some could be removed
continues #8947 

## Changes

Removed the following:
- `isFabricObjectWithDragSupport` - moved methods to InteractiveObject
- `isFabricObjectCached` - unused
- `isInteractiveTextObject` - see comments below for each call (all were removed)
- ~~`isActiveSelection`~~

Moved `isCollection` to `Collection.ts`


src/canvas/Canvas.ts
```diff
@@ -765,21 +757,13 @@ export class Canvas extends SelectableCanvas {
...
       (activeObject && target && activeObject !== target)
-    ) {
-      // this covers: switch of target, from target to no target, selection of target
-      // multiSelection with key and mouse
-      return true;
-    } else if (isInteractiveTextObject(activeObject)) {
-      // if we mouse up/down over a editing textbox a cursor change,
-      // there is no need to re render
-      return false;
-    }
```
ShaMan123, line 781: redundant, since we are retuning false anyways

src/shapes/Object/InteractiveObject.ts
```diff
@@ -631,12 +624,29 @@ export class InteractiveFabricObject<
...
+
+  /**
+   * Override to customize Drag behavior\
+   * Fired once a drag session has started
+   * @returns true to handle the drag event
+   */
+  onDragStart(e: DragEvent) {
+    return false;
+  }
+
```
ShaMan123, line 644: exposed on instance instead of checking for method existence

src/util/typeAssertions.ts
```diff
@@ -50,29 +44,3 @@ export const isTextObject = (
...
-  // we could use instanceof but that would mean pulling in Text code for a simple check
-  // @todo discuss what to do and how to do
-  return !!fabricObject && fabricObject.isType('IText', 'Textbox');
-};
-
-export const isFabricObjectCached = (
-  fabricObject: FabricObject
-): fabricObject is TCachedFabricObject => {
-  return fabricObject.shouldCache() && !!fabricObject._cacheCanvas;
-};
```
ShaMan123, line 66: unused

ShaMan123 (commented):
looks alright

ShaMan123 (commented):
looks alright

src/canvas/Canvas.ts
```diff
@@ -1516,7 +1494,6 @@ export class Canvas extends SelectableCanvas {
         }
         this._fireSelectionEvents(prevActiveObjects, e);
       } else {
-        isInteractiveTextObject(activeObject) && activeObject.exitEditing();
```
ShaMan123, line 1519: should be handled by

https://github.com/fabricjs/fabric.js/blob/8eb642460556afea542e50537e1bcf0de3cb79bb/src/canvas/SelectableCanvas.ts#L1435-L1438


https://github.com/fabricjs/fabric.js/blob/8eb642460556afea542e50537e1bcf0de3cb79bb/src/shapes/IText/ITextBehavior.ts#L109-L113

ShaMan123: And is more correct since if onDeselect returns true then text editing should remain active... and perhaps multi selection should abort.

src/shapes/Object/StackedObject.ts
```diff
@@ -50,8 +49,7 @@ export class StackedObject<
    */
   getParent<T extends boolean>(strict?: T): TAncestor | undefined {
     return (
-      (isActiveSelection(this.group) ? this.__owningGroup : this.group) ||
-      (strict ? undefined : this.canvas)
+      this.__owningGroup || this.group || (strict ? undefined : this.canvas)
```
ShaMan123, line 52: simplified

ShaMan123 (commented):
moved `isCollection` to `Collection.ts`

src/shapes/Object/StackedObject.ts
ShaMan123: this is wrong

ShaMan123: reverted

ShaMan123 (commented):
DONE

asturur:
image

i still stand on my opinion.
There was an issue, subclasses detection, and instead of fixing it you worked around changing the code around the affected methods in order to remove them.

The change to the iText exit editing is unclear to me. I m not sure the onDeselect case you point out cover all the bases, i would keep the class check there.

src/canvas/Canvas.ts
ShaMan123: so to male things clear, is this the only thing you disapprove of?

zhe-he:
You might consider being more daring by using capability detection instead of type detection. Replace all type checks within the program with capability checks. For example, the reason you want to know if something is a TextBox is because TextBox can focus the cursor. Instead, you can directly check if it has the ability to focus the cursor (using unique variables or methods), which also solves the problem of type checking loops with instanceof. Similarly, if you want to determine if something is a Path in order to manipulate its unique property pathOffset, you can directly check if it has the pathOffset property and modify it accordingly.
Instead of using isInteractiveTextObject, you can use canEdit, canFocus, can... to represent the capabilities. Similarly, instead of using isFiller, you can break it down into canGradient, hasSource, can...

The above is just my personal suggestion and represents my own viewpoint. I am not familiar with the entire project, so there may be other factors to consider that I might have overlooked. If there are any shortcomings, please understand.

src/canvas/Canvas.ts
asturur: this line i would put back

asturur:
```ts
(activeObject as IText).exitEditing && activeObject.exitEditing();
```
it think something like this would just work fine and not introduce a change we don't want to think about now

asturur (changes requested):
I would avoid removing that exitEditing change

asturur:
tried to solve merge conflicts, hopefully i didn't did mistakes.
This PR also reminds me that text drag support is not easily deactivable

asturur:
> You might consider being more daring by using capability detection instead of type detection. Replace all type checks within the program with capability checks. For example, the reason you want to know if something is a TextBox is because TextBox can focus the cursor. Instead, you can directly check if it has the ability to focus the cursor (using unique variables or methods), which also solves the problem of type checking loops with instanceof. Similarly, if you want to determine if something is a Path in order to manipulate its unique property pathOffset, you can directly check if it has the pathOffset property and modify it accordingly. Instead of using `isInteractiveTextObject`, you can use `canEdit`, `canFocus`, `can...` to represent the capabilities. Similarly, instead of using `isFiller`, you can break it down into `canGradient`, `hasSource`, `can...`
> 
> The above is just my personal suggestion and represents my own viewpoint. I am not familiar with the entire project, so there may be other factors to consider that I might have overlooked. If there are any shortcomings, please understand.

Yeah this we discussed also, to keep type checks in place and keep subclassing normal, we should avoid strings checks and check something more fundamental of the instance.

We also need to say that those were only really typechecks, to avoid using as everywhere.
Apparently TS is annoyed when you want just to check if a property exists.

Those are probably safe to go away

asturur:
@ShaMan123 if you have time let me know what you think of my change request so that we can move forward or reiterate

src/canvas/Canvas.ts
ShaMan123: You want to put it back because you are afraid of a regression?
I understand your point I think.
But disagree.
Cleaning up has many gains. And leaving a dead/redundant line of code is a future bug in potential. And may cause developing other stuff seem harded because it will be viewed as a constrain when actually it should not be there.
I am willing the test it deeply.

Since you are unwilling then we should at least add a TODO comment explaining this. I am OK with that.

ShaMan123 (commented):
I am not sure but I think you reinstated `isPath` which should also be removed, maybe in a follow up,

I am ok with your changes as long as we add a TODO comment and address that in the future when we look into side effects running in onSelect/onDeselect

src/canvas/Canvas.ts
asturur: > You want to put it back because you are afraid of a regression?

Yes removing a single line and adding a possible regression when we are trying to clean up is risky.
Right now in the worst case we are calling exit editing twice. Is not a big deal.
As we discussed we can keep this, and just make a comprehensive plan on who should handle the deselection side effect and how and let's think of this.

asturur (approved):

PR #8951, opened 2023-05-22
chore(): cleanup type assertions, expose `Object#parent` => rm `isActiveSelection`

ShaMan123:
## Motivation

#8931 
continues #8950 #8947 

closed #9349 

I was looking into the patch I made into #9336 and understood that `getParent` was a design mistake.
It is confusing and silly. I wrote it in a haste, not wanting to create too many changes.
It is a design mistake IMO mainly because there is a confusion about active selection - should it be considered a parent or not.
Apps might need to access both the the current `group` of an object and both the owning group.
Since `group` and `canvas` are used it seems bad that to get the owning parent you do `getParent(true)`.
I regret it.
This PR makes it simpler both for fabric and for devs and removes ambiguity of the `group` property.

## Description

This PR introduces the `parent` property on `FabricObject` and removes `getParent`.
The `parent` property is the owning group of an object, meaning not the active selection.

While testing the PR I realized I didn't take into consideration another part of transferring an object between active selections that I missed in #9336
The object should be removed from the current active selection and not be removed from its owning group.

## Changes

Exposes `Object#parent` instead of `__owningGroup`, which is something I wanted to do ever since #7670 since it makes it simpler to handle the object tree.
This removes the need for `isActiveSelection`


ShaMan123 (commented):
a lot more straight forward 
DONE

ShaMan123:
ridiculous
This is waiting here and I redid it #9349

src/shapes/Object/StackedObject.ts
```diff
@@ -42,17 +41,14 @@ export class StackedObject<
...
    * Used to keep the original parent ref when the object has been added to an ActiveSelection, hence loosing the `group` ref
    */
-  declare __owningGroup?: Group;
+  declare parent?: Group;
 
   /**
    * Returns instance's parent **EXCLUDING** `ActiveSelection`
    * @param {boolean} [strict] exclude canvas as well
    */
   getParent<T extends boolean>(strict?: T): TAncestor | undefined {
```
ShaMan123, line 50: remove this method because the dev can access `group` or `parent` or check equality

ShaMan123, line 50: we should rename or reassess the need of this method

src/shapes/ActiveSelection.ts
```diff
@@ -82,14 +82,7 @@ export class ActiveSelection extends Group {
...
   enterGroup(object: FabricObject, removeParentTransform?: boolean) {
-    if (object.group) {
-      //  save ref to group for later in order to return to it
-      const parent = object.group;
-      parent._exitGroup(object);
-      // make sure we are setting the correct owning group
-      // in case `object` is transferred between active selections
-      !(parent instanceof ActiveSelection) && (object.__owningGroup = parent);
-    }
+    object.parent?._exitGroup(object);
```
ShaMan123, line 85: I need to revisit after rebasing - this is clearly wrong

ShaMan123 (commented):
looks correct

asturur:
Looking at this PR i see the confusion between add/remove and enter/exit group.
Which are the one that the developer is supposed to use?

asturur:
ok i see add/remove are still the principals one and enter/exit Group are just tools for add/remove to happen.
I ll follow up masking things accordingly, as of now the code for add and remove is more complex that it needs.

do you think `_onRelativeObjectAdded` is used anywhere? I can't see it used anywhere and is undocumented.

ShaMan123:
It isn't
You can and should remove it

src/shapes/ActiveSelection.ts
```diff
@@ -76,22 +76,29 @@ export class ActiveSelection extends Group {
...
+    // is also its parent, meaning that is not in an active selection, but is
+    // in a normal group.
+    if (object.parent && object.parent === object.group) {
+      // Disconnect the object from the group functionalities, but keep the ref parent intact
+      // for later re-enter
+      object.parent._exitGroup(object);
+      // in this case the object is probably inside an active selection.
+    } else if (object.group && object.parent !== object.group) {
+      // in this case group.remove will also clear the old parent reference.
+      object.group.remove(object);
```
asturur, line 96: please check this comment here @ShaMan123 .
I do not know if the deletion of object.parent in this case is wanted.
On line 92 that doesn't happen, and i don't understand why this should be different.

src/shapes/ActiveSelection.ts
```diff
@@ -76,22 +76,29 @@ export class ActiveSelection extends Group {
...
-   * @returns {boolean} true if object entered group
    */
   enterGroup(object: FabricObject, removeParentTransform?: boolean) {
-    if (object.group) {
-      //  save ref to group for later in order to return to it
-      const parent = object.group;
-      parent._exitGroup(object);
-      // make sure we are setting the correct owning group
-      // in case `object` is transferred between active selections
-      !(parent instanceof ActiveSelection) && (object.__owningGroup = parent);
```
asturur, line 91: in my opinion the instanceof check was more explicit of what we are after here.

asturur:
Sorry for this:
image

But i tried to be clear many times, i don't want to see all those prs, it just get me slow and away from reviewing them.
The more we open, the more open topics, i get exponentially slower.

asturur (approved):

src/shapes/ActiveSelection.ts
```diff
@@ -76,22 +76,29 @@ export class ActiveSelection extends Group {
...
+    // This condition check that the object has currently a group, and the group
+    // is also its parent, meaning that is not in an active selection, but is
+    // in a normal group.
+    if (object.parent && object.parent === object.group) {
+      // Disconnect the object from the group functionalities, but keep the ref parent intact
+      // for later re-enter
+      object.parent._exitGroup(object);
+      // in this case the object is probably inside an active selection.
```
ShaMan123, line 93: not probably, for sure
and it should be inside the else if block IMO because it is confusing to where it belongs to

src/shapes/ActiveSelection.ts
ShaMan123: object.group === ActiveSelection instance so remove will not remove the parent ref
I agree instaof ActiveSeletion will make this code eaasier

src/shapes/ActiveSelection.ts
```diff
@@ -76,22 +76,29 @@ export class ActiveSelection extends Group {
...
+    } else if (object.group && object.parent !== object.group) {
+      // in this case group.remove will also clear the old parent reference.
+      object.group.remove(object);
     }
+    // enter the active selection from a render perspective
+    // the object will be in the objects array of both the ActiveSelection and the Group
+    // but referenced in the group's _activeObjects so that it won't be rendered twice.
```
ShaMan123, line 100: not percise
ActiveSelection is not rendered

PR #8952, opened 2023-05-22
refactor(): cache state => rm `isTextObject`

ShaMan123:
## Motivation

#8931 

## Description

The last part of type assertion to handle.
I had to refactor the messy logic of `_updateCacheCanvas` and extract from it a method that has no side effects `getCacheState` so Text can override it.

Apart from that is is only one more use of `isTextObject`
https://github.com/fabricjs/fabric.js/blob/a1f94b29dda272f3dff6b2eda92f29d6d2b8c3ad/src/canvas/StaticCanvas.ts#L1319-L1321 which I don't understand but I guess we can check for an existence of an svg method that text has and we are done

### DIFF
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -8,7 +8,6 @@ import { Shadow } from '../../Shadow';
 import type {
   TDegree,
   TFiller,
-  TSize,
   TCacheCanvasDimensions,
   Abortable,
 } from '../../typedefs';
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
DONE

src/shapes/Text/TextSVGExportMixin.ts
```diff
@@ -286,4 +286,15 @@ export class TextSVGExportMixin extends FabricObjectSVGExportMixin {
...
+  getSVGFontList() {
+    const fontList: Record<string, boolean> = { [this.fontFamily]: true };
+    this.styles &&
+      Object.values(this.styles).forEach((styleRow) => {
+        Object.values(styleRow).forEach(({ fontFamily = '' }) => {
+          fontList[fontFamily] = true;
+        });
+      });
+    return fontList;
+  }
```
ShaMan123, line 299: ported from #8954

ShaMan123 (commented):
DONE after porting

ShaMan123:
We disagreed on this so closing

PR #8953, opened 2023-05-22
chore() remove unnecessary typeof === 'undefined'

asturur:
## Motivation

For variables that are declared, we don't need to check the typeof === 'undefined' to know if they are undefined. We can just check the strict equality with undefined.

`typeof variable === 'undefined'` is needed when you don't know if the variable is in scope at all

Please someone correct me if i m wrong


asturur:
Eventually the minifier does know this already.

ShaMan123:
What is the benefit?

PR #8955, opened 2023-05-24
feat(): Extended New Snapping Guidelines for objects

CODE-AXION:
### _1) The feature was added using the existing code methods mentioned in aligning_guidelines.js file_ 

**New Features I Added In Snapping Guidelines:**

-   // snaps if the top of the active object touches the bottom of the other object
-   // snaps if the bottom of the active object touches the top of the other object
-   // snaps if the left side of the active object touches the right side of the other object
-   // snaps if the right side of the active object touches the left side of the other object

**Existing Snapping Guidelines:**
- snap by the horizontal center line
- snap by the left edge
- snap by the right edge
- snap by the vertical center line
- snap by the top edge
- snap by the bottom edge

**_New Snapping Guidelines screenshots:_**
_snaps if the left side of the active object touches the right side of the other object:_
image
_snaps if the right side of the active object touches the left side of the other object_
image
_snaps if the top of the active object touches the bottom of the other object:_
image
_snaps if the bottom of the active object touches the top of the other object:_
image

### _2) Fixed a bug_

bug: Throws an error when i try to download the canvas in different formats like png , jpg etc.. 

How i fixed: added condition check on _lib/centering_guidelines.js_ and _lib/aligning_guidelines.js_

Error: 
image


asturur:
@CODE-AXION thank you for the PR!
Today till monday i will be out and i won't be able to look at it.

I will back to you soon

CODE-AXION:
> @CODE-AXION thank you for the PR! Today till monday i will be out and i won't be able to look at it.
> 
> I will back to you soon

CODE-AXION:
> @CODE-AXION thank you for the PR! Today till monday i will be out and i won't be able to look at it.
> 
> I will back to you soon

hey when you will checkout my PR ?

CODE-AXION:
@asturur Hi its been a long time can you review this PR ?

asturur:
Those will need to be updated to TS and exported properly for v6, but i see no harm in merging as is.
I m sorry i couldn't pay attention to you for years but the repo is higly congested and things are a bit stuck and i don't want to work on fabricjs 8 hours a day.

Smrtnyk:
This is the feature I am very interested in
Does this piece of code work also when canvas is zoomed in?

asturur:
> This is the feature I am very interested in Does this piece of code work also when canvas is zoomed in?

Not sure, but if it doesn't it should pretty easy to fix.

asturur:
You have also to consider this code hasn't been tested with v6 and more importantly has no tests at all

lib/aligning_guidelines.js
```diff
@@ -89,6 +89,79 @@ function initAligningGuidelines(canvas) {
...
+          x: objectLeft - objectWidth / 2,
+          y1: (objectTop < activeObjectTop)
+            ? (objectTop - objectHeight / 2 - aligningLineOffset)
+            : (objectTop + objectHeight / 2 + aligningLineOffset),
+          y2: (activeObjectTop > objectTop)
+            ? (activeObjectTop + activeObjectHeight / 2 + aligningLineOffset)
+            : (activeObjectTop - activeObjectHeight / 2 - aligningLineOffset)
+        });
+
+        activeObject.setPositionByOrigin(new fabric.Point(objectLeft - objectWidth / 2 - activeObjectWidth / 2, activeObjectTop), 'center', 'center');
```
ShaMan123, line 107: do not use setPositionByOrigin, it is wrong for a nested object

asturur (approved):

lib/aligning_guidelines.js
ShaMan123: use setXY since the point is in the scene

lib/aligning_guidelines.js
```diff
@@ -89,6 +89,79 @@ function initAligningGuidelines(canvas) {
           objectHeight = objectBoundingRect.height / viewportTransform[3],
           objectWidth = objectBoundingRect.width / viewportTransform[0];
 
+
+      // snaps if the right side of the active object touches the left side of the object
+      if (isInRange(activeObjectLeft + activeObjectWidth / 2, objectLeft - objectWidth / 2)) {
+
+        verticalInTheRange = true;
+        verticalLines.push({
+          x: objectLeft - objectWidth / 2,
```
ShaMan123, line 98: objectLeft is a wrong name

lib/aligning_guidelines.js
```diff
@@ -190,7 +263,9 @@ function initAligningGuidelines(canvas) {
   });
 
   canvas.on('before:render', function() {
-    canvas.clearContext(canvas.contextTop);
+    if (canvas.contextTop) {
+      canvas.clearContext(canvas.contextTop);
+    }
```
ShaMan123, line 268: not needed if v6

ShaMan123: the right approach would be
canvas.contextTopDirty = true

asturur:
@ShaMan123 This code won't probably work with nested selections inside groups, but also it wasn't in scope at the time of changes.
This whole guidelines code has probably a large number of incompatibilities with v6 and needs to be converted and evaluated again from scratch.

It hasn't been reviewed for ages and i m not going to ask changes now isn't really helpful.
We have to assume it works as is for the author and that anyway we would have to fix guidelines for v6 and typescript and the little amount of code he added makes no difference for us.

asturur:
The real issue with this code is that it assumes that there is an object called 'fabric' that doesn't exist in that context.

asturur:
@Smrtnyk i suggest you do not dive converting this exact code in TS.
There are things to consider

scaling support
exposing this feature in the module world we are now


I hope we can close 6.0 soon and then fit this in a roadmap with some specs, then if you think you can take it on, you can take it on, with specs

asturur:
@CODE-AXION on which version of fabricJS are you using this exact code?

CODE-AXION:
> @CODE-AXION on which version of fabricJS are you using this exact code?

@asturur ok so i opened the project on my local and at that time i was using "version": "6.0.0-beta6" (from package.json)

asturur:
So it works for you also if it references fabric.Point? do you have fabric object on the window object?

CODE-AXION:
> So it works for you also if it references fabric.Point? do you have fabric object on the window object?

@asturur yes i am using fabric as a window object

asturur:
@CODE-AXION someone refactored all the guidelines in TS and i worked to make them an extension package in fabric.
I never came back to this PR but i merged today because i wanted to leave track of your effort of fixing them with the commit.
I also need to compare both implementation and make a review of both approaches, probably the best will be a mix of the 2

PR #8939, opened 2023-05-20
chore(): move canvas click handler to TextManager

ShaMan123:
## Motivation

I missed the click handler that IText attaches to canvas when I wrote the TextEditingManager

## Description

Move this click handler to the manager, where it belongs

PR #8970, opened 2023-06-01
fix(): regression to itext focusing from #8939

ShaMan123:
## Description

I made a mistake in #8939 and introduced a bug.
Editing text and then clicking with the mouse on the itext will not gain focus (because TextEditingManager loses `target` on mouse up).

## Changes

Use the active object for focusing the hiddenTextarea

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
please review

asturur (approved):

PR #8971, opened 2023-06-01
ci(): warn build errors in dev mode

ShaMan123:
## Motivation

Currently the process exits when there is a build error.
That should happen only in CI.
When we develop we want rollup to keep watching and building to fix the error (we don't want to restart the build each time)

## In Action

https://github.com/fabricjs/fabric.js/actions/runs/5147975815/jobs/9269055157?pr=8971

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
ready

asturur (dismissed):

asturur (approved):

PR #8972, opened 2023-06-01
fix(IText): cursor detection

ShaMan123:
## Motivation

Working on a subclass that has vertical text alignment.
This class overrides `_getTopOffset` => rendering is perfect.
However, pointer detection is off.
Traced it back to this line.

## Description

The local pointer is relative to the left/top point of where the text begins.

## Gist

I hate the naming of the methods in Text and subclasses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
-- fix(IText): cursor detection  [#8972](https://github.com/fabricjs/fabric.js/pull/8972)
+- fix(IText): cursor detection [#8972](https://github.com/fabricjs/fabric.js/pull/8972)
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
No tests...
However I did test it in the app I am working on.
Also, I can add vertical alignment into fabric, using the impl of the app. It is a minimal change.
Thoughts?

asturur:
> No tests... However I did test it in the app I am working on. Also, I can add vertical alignment into fabric, using the impl of the app. It is a minimal change. Thoughts?

We are not adding anything for now.
I have to make an hard cut because i don't feel i m moving forward to a release.

I can add tests for this PR.

I'm thorn for removing the method, but i understand that the localPointer would become a local pointerd depending on the text orientation and align and doesn't make much sense anymore.

PR #8973, opened 2023-06-02
fix(Text): styles line break

ShaMan123:
## Description

Line break consistency in style parsing

### DIFF
--- a/src/shapes/Text/constants.ts
+++ b/src/shapes/Text/constants.ts
@@ -1,4 +1,4 @@
-import { LEFT } from '../../constants';
+import { LEFT, reNewline } from '../../constants';
 import type { TClassProperties } from '../../typedefs';
 import type { Text } from './Text';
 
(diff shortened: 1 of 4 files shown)


ShaMan123 (commented):
done

src/shapes/Text/constants.ts
```diff
@@ -63,7 +63,7 @@ export const styleProperties: Readonly<StylePropertiesType[]> = [
 // regexes, list of properties that are not suppose to change by instances, magic consts.
 // this will be a separated effort
 export const textDefaultValues: Partial<TClassProperties<Text>> = {
-  _reNewline: /\r?\n/,
+  _reNewline: reNewline,
   _reSpacesAndTabs: /[ \t\r]/g,
   _reSpaceAndTab: /[ \t\r]/,
   _reWords: /\S+/g,
```
asturur, line 69: thos imho should be all const. They don't seem fit for class properties

asturur (approved):

src/shapes/Text/constants.ts
ShaMan123: I agree

PR #8974, opened 2023-06-02
ci(): invoke tests after changelog action

ShaMan123:
## Motivation

~~Since I have disabled tests when a change is made to the changelog they don't show up in the PR UI after the changelog finishes and it is vital they do.
It is used for one commit PRs because if you continue pushing then tests will start and everything is fine. There are other options I have been using (closing and re-opening a PR and I think I can invoke manually but it is annoying)
We could simply remove the `ignore-paths` directive instead of what I did in this PR.
Whatever you decide is good for me.~~

It didn't work so I removed `paths-ignore`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,6 @@ name: '🩺'
 on:
   pull_request:
     branches: [master]
-    paths-ignore: [CHANGELOG.md]
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,6 @@ on:
   pull_request:
     branches: [master]
     paths-ignore: [CHANGELOG.md]
-  workflow_call:
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


ShaMan123:
Well it didn't work so I am going for the second option

ShaMan123 (commented):
DONE

asturur:
I think in general if we reduce the amount of tests/checks we run is better, especially the CI infrastructure is free until people abuse it.
Something should be possible and i didn't find a way to do, is that it should be possible to specify the merge strategy for changelog.MD to specify that we always keep both changes and so autoresolve the confilct

asturur (approved):

ShaMan123:
> I think in general if we reduce the amount of tests/checks we run is better, especially the CI infrastructure is free until people abuse it.

I agree, that is why I added the ignore paths but the UI is so annoying. Once the changelog action commits the tests are skipped and the UI doesn't show the previous run.

> Something should be possible and i didn't find a way to do, is that it should be possible to specify the merge strategy for changelog.MD to specify that we always keep both changes and so autoresolve the confilct

We have this already in .gitattributes

ShaMan123:
I don't mind reverting this
All it means is that if the changelog action commits I close and re-open the PR so the UI shows the test status

PR #8980, opened 2023-06-05
feat(): get/set proxy traps

ShaMan123:
## Motivation

#8682

## Description

Using proxy traps aligns us with class spec (assigning values in the constructors goes through the `defineProperty` trap and not the `set` trap) so by default, once we migrate `set`, no side effects will run during initialization. This will force us to initialize behavior explicitly in the constructor, which is good practice IMO.

I added traps for side effects to replace `set` (as you suggested in the past) that call `transformValue` and `onChange` if needed. 
`transformValue` runs also from the `get` trap and can return a different value that will be used instead (e.g. use for `scaleX`, `flipX` - `flipX` get become a getter that is declared only and `scaleX` can have a negative value but return a positive value always).
`onChange` is for side effects and can return false to revert the change that was made.

## Changes

- `createHybrid`
- `CommonMethods#transformValue` pre get/set hook
- `CommonMethods#onChange` side effect hook that is invoked every time a change to a value occurs, returning false reverts the change
- `Circle#set` - changed to prove the side effect logic

## Unlocked Possibilities
The set traps allows us to do more state syncs:
- `setCoords` - can virtually become an under the hood method that will be used rarely
- `calcTransformMatrix` - setting it in the trap and removing the need to have a cache key


src/util/internals/Proxy/Proxy.ts
```diff
@@ -0,0 +1,75 @@
...
+            target
+          )
+        : value;
+    },
+
+    set(target, p, newValue) {
+      const has = Reflect.has(target, p);
+      const prevValue = Reflect.get(target, p);
+      const descriptor = Reflect.getOwnPropertyDescriptor(target, p);
+      const enumerable = !descriptor || descriptor.enumerable;
```
ShaMan123, line 24: wee can decide to allow even non enumerable props. I allow the deepest overriding there is but that is dangerous

ShaMan123 (commented):
Ready for reviewing

ShaMan123:
I really don't know how to make TS understand this simple type

ShaMan123 (commented):
This is good and safe

PR #8981, opened 2023-06-05
refactor(): migrate `Object#_set` method to `onChange` hook

ShaMan123:
## Description

continues #8980 , merge that first
https://github.com/fabricjs/fabric.js/compare/set-proxy...set-proxy-2

## Changes

- `Object#_set` => `onChange`
- BlendImage filter didn't respect flipX/Y - was that intentional or is that a bug?
- 

## Gist

blocked by #8507 for shadow parsing


asturur:
@ShaMan123 i can't accept any more any kind of refactor till we don't release 6.0 with proper documentation.
zero changes that are not plain bugfixes.

ShaMan123:
I agree, put it here for the future.
Sorry for not stating that

PR #8987, opened 2023-06-06
patch(): Text styles minor patch

ShaMan123:
## Description

a tiny cleanup and a change to the next/vanilla app adding text styles

### DIFF
--- a/src/shapes/Text/StyledText.ts
+++ b/src/shapes/Text/StyledText.ts
@@ -8,6 +8,7 @@ import { FabricObject } from '../Object/FabricObject';
 import { styleProperties } from './constants';
 import type { StylePropertiesType } from './constants';
 import type { Text } from './Text';
+import { pick } from '../../util';
 
 export type TextStyleDeclaration = Partial<Pick<Text, StylePropertiesType>>;
 
@@ -252,10 +253,7 @@ export abstract class StyledText<
    */
   _getStyleDeclaration(lineIndex: number, charIndex: number) {
     const lineStyle = this.styles && this.styles[lineIndex];
-    if (!lineStyle) {
-      return null;
-    }
-    return lineStyle[charIndex];
+    return lineStyle ? lineStyle[charIndex] : null;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/shapes/Text/StyledText.ts
```diff
@@ -266,18 +264,11 @@ export abstract class StyledText<
...
-      styleObject[prop] =
-        typeof style[prop] === 'undefined'
-          ? this[prop as keyof this]
-          : style[prop];
-    }
-    return styleObject;
+    return {
+      // @ts-expect-error readonly
+      ...pick(this, (this.constructor as typeof StyledText)._styleProperties),
+      ...(this._getStyleDeclaration(lineIndex, charIndex) || {}),
```
ShaMan123, line 270: I don't think there is a need to pick the props from this object because it is defined with these props

src/shapes/Text/StyledText.ts
```diff
@@ -266,18 +264,11 @@ export abstract class StyledText<
...
-      // @ts-expect-error TS complains even when we serve everything on a silver plate.
-      styleObject[prop] =
-        typeof style[prop] === 'undefined'
-          ? this[prop as keyof this]
-          : style[prop];
-    }
-    return styleObject;
+    return {
+      // @ts-expect-error readonly
+      ...pick(this, (this.constructor as typeof StyledText)._styleProperties),
```
ShaMan123, line 269: I think we can removed `_styleProperties`, used only here

ShaMan123 (commented):
DONE

src/shapes/Text/StyledText.ts
asturur: While this change is not wrong and is fine, tidier, we still need to validate the comment of that user that suggested verbose code over a loop in this part because of performance issues.
Just keep in mind that we may end up just putting a blunt list of properties in the style of:

```
return {
  fontsSize: style.fontSize ?? this.fontSize,
  ...
}
```

if performance concerns are true

melchiar: performance aside, I would also opt for the more human readable code

asturur (approved):

src/shapes/Text/StyledText.ts
jiayihu: Yes, as reported before, this function is a hot one as it's called **very** frequently when rendering, especially without caching enabled, so any performance issue is multiplied by a big factor.

In my findings with fabric v3.6, changing from a dynamically created object to a static one would improve the performance significantly:

## Before

Screenshot 2023-03-06 at 12 06 07

## After

Screenshot 2023-03-06 at 12 34 00

Looping through the list of properties has a cost, but more significantly, creating an object by dynamically assigning the properties has a much greater impact I reckon. It means that each time the method is called, V8 or any other JS engine cannot optimize this code as it's too dynamic. Using an object with a static shape allows the engine to optimize the code and speed it up: https://mathiasbynens.be/notes/shapes-ics#shapes

ShaMan123: So for the sake of discussion
```ts
{
   fontSize: this.fontSize,
   ...,
   ...(this._getStyleDeclaration(lineIndex, charIndex) || {})
}
```

Is not good but what @asturur suggested is?

jiayihu: I found today this tool that can maybe solve any doubt about what works and what doesn't: https://mathiasbynens.be/notes/shapes-ics#shapes

PR #8990, opened 2023-06-06
fix(Textbox): `splitByGrapheme` measurements

ShaMan123:
closes #8989

## Motivation

closes #8989 

## Description

Text measuring was adding too much offset because I made a mistake thinking it was words that have space in between them which is incorrect when flagging `splitByGrapheme`

image

PR #8994, opened 2023-06-07
fix(Textbox): minWidth of text, measure lines before wrapping

ShaMan123:
## Motivation

#8990 

Wrapping line logic is unreadable and very hard to trace back.
It is needed (by me at least) for overriding so it must be clear and scoped.

## Description

`minWidth` should be shared between lines.
In case there is a text line that is long after a short one, the short line does not know it has more available size to grow to.
This is visible with the resizing control, see attached videos.

I suggest you read this PR commit by commit since they are self explanatory IMO, or not, you can simply read as is.
I consolidated all measuring logic (moved infix measuring to the measuring loop) and extracted it to `measureLinesForWrapping`, now running before wrapping and not as part of wrapping.
Besides that I cleaned up logic to be readable and to use as much consts and as less switches as possible.

## Changes

No changes to functionality.
Exposed `measureLinesForWrapping`

**BREAKING**: `_wrapLine` signature

#### Master

https://github.com/fabricjs/fabric.js/assets/34343793/1c10fbb5-0e88-4c08-b29e-02d9922f8e69

#### Fixed

https://github.com/fabricjs/fabric.js/assets/34343793/7dee0614-fe20-40bf-8d86-ecb5018c7665


ShaMan123 (commented):
stable and clean

ShaMan123 (commented):
DONE

ShaMan123 (commented):
DONE DONE DONE!

src/shapes/Textbox.ts
```diff
@@ -314,84 +370,56 @@ export class Textbox extends IText {
...
-    return graphemeLines;
+    // layout
+    return data.reduce(
+      ({ lines, lineWidth }, { graphemes, infix, infixWidth, width }, i) => {
+        // `i === 0` => `infixWidth === 0`
+        const lineWidthAfter = lineWidth + infixWidth + width;
+        if (i === 0 || lineWidthAfter - additionalSpace > maxWidth) {
+          // push a new line, we spread to protect `data` from mutation
+          lines.push([...graphemes]);
+          lineWidth = width;
```
ShaMan123, line 410:
```suggestion
          lineWidth = infixWidth + width;
```

If someone decides to override `measureLineForWrapping` and return an infix we should respect it.

something like:
```
some words some words some words some words 
some words some words ....
    a line with an infix at the start
```

ShaMan123: and then we should push the infix if it exists

src/shapes/Textbox.ts
```diff
@@ -306,6 +310,58 @@ export class Textbox extends IText {
     return value.split(this._wordJoiners);
   }
 
+  measureLineForWrapping(line: string, lineIndex: number) {
+    const splitByGrapheme = this.splitByGrapheme,
```
melchiar, line 314: is this assignment really necessary as opposed to just using `this.splitByGrapheme`?

src/shapes/Textbox.ts
```diff
@@ -306,6 +310,58 @@ export class Textbox extends IText {
...
+          // measure infix at offset to respect styling etc.
+          infixWidth = this._measureWord(infix, lineIndex, offset);
+          // move cursor after infix
+          offset += infix.length;
+        }
+        // split words if necessary
+        const graphemes = !splitByGrapheme
+          ? this.graphemeSplit(part)
+          : // we must use concat for compatibility
+            [].concat(part);
```
melchiar, line 336: not sure I'm understanding this. Why not just use `[ part ]`?

ShaMan123: I tried that.
some tests were failing but like this they were not.
It is really strange

src/shapes/Textbox.ts
ShaMan123: agreed, it is a left over

ShaMan123: 76eef16e2

ShaMan123:
I now understand what I dislike about the text logic.
The lifecycle of measuring, layout, rendering is a complete mess.
Some measuring is done here some there.
It should be strict measuring, layout, rendering.
But that is a big effort.

melchiar:
> I now understand what I dislike about text logic. The lifecycle of measuring, layout, rendering is a complete mess. Some measuring is done here some there. It should be strict measuring, layout and finally rendering. But that is a big effort.

In an ideal world measurements would also be cached, though that mean adding cache invalidation and a lot more work

ShaMan123:
Not really,
`initDimensions` is the place everything originates from and it calls `clearCache`

melchiar:
I guess the real question then is how much of a performance hit measuring actually causes. Even if it's small these methods get called many times especially when there's lots of text

ShaMan123:
It is not a question of adding measuring, simply doing it it one place, returning all data needed.
A single pass.
And then the layout and rendering consume it.

src/shapes/Textbox.ts
asturur: if part is an array, [].concat(part) is part and not [part]
I think that is the the difference.

asturur:
i don't think the measurements today are done in many places.
init dimensions does all of it, and stores different stages of measuring in different cache objects.

Apart that, ugly or bad code is sometimes subjective.

I see all those changes are concentrated in 2 methods, so that is mostly fine by me, but still i have to go and check if the methods are as before or not because is not really clear just going over the code.

What i want to know is the improvement this pr brings.
Yes now we have 2 methods instead of one, what do you do by overriding the line wrapper? Because developers asked more than once for example to have break-word or bullet list ( that is possible today ).

ShaMan123:
It makes it easier (but not easy) to override line wrapping without having to also measure everything (which is absolutely obscure)
#8526 has been waiting for a long time...
Why didn't we merge it BTW?
It should make it easier to wrap differently, yes.


> i don't think the measurements today are done in many places.
init dimensions does all of it, and stores different stages of measuring in different cache objects.

I don't think so. Because of `initDimensions` I was expecting `__charBounds` to be populated but it is not.
Also `__lineWidth`, `__lineHeight` all that mess, why not measure everything and store it?

src/shapes/Textbox.ts
ShaMan123: Yes, let's abandon this idea.

> measureLines shouldn't change the lines imho, if someone does that is going on its own road with no guarantee of support

I don't understand this

asturur:
> I don't think so. I was expecting `__charBounds` to be populated after running `initDimensions`, but it is not.
> Also `__lineWidth`, `__lineHeight` all that mess, why not measure everything and store it?

Because charbounds has been added years after __lineWidth and __lineHeight already existed.
I would expect charBounds to be populated, not sure why isn't.
It could be that we measure the chars only if kerning information is necessary.
If a word is all of the same color and we don't need to do kerning manually we measure the whole word and we don't get charbounds.
I don't remember sincerly.


Many PRs aren't merged because in order:
- i m not adding features easily
- when we focused on TS we lost track of a lot of old work
- we are not adding anything anyway now
- too many changes all together in general i just get more not wanting changes

asturur:
@ShaMan123 going back to the focus of merging this PR, there are some red flags for me.
Together with the split you changed a bunch of for loops in reduce, added some safeguard mutation, and things i don't see yet.
Are those needed by you? if you need to split the method, split the method and remove the rest. We just fixed a bug in text yesterday because we changed some code with some code code we considered equivalent and we didn't see the bug for weeks

If the changes are required because you are splitting the class method not in half, but in many parts and taking some of them in a different order from the original, then we are not simply extracting some code in a method.

asturur:
> I understand. My point is it is really hard to work on text. It is a no touch thing. And that is a shame. Because it is really powerful and has lots of potential.

yes Text is hard, is fragile and the main focus after 6.0 should be RTL and vertical text ( that you started long ago ), and proper emoji support.
All the rest is really minor imho.

ShaMan123:
I have found and fixed a bug so now I have a good excuse to merge this
Updated description and added a test that is broken on master

src/shapes/Textbox.ts
```diff
@@ -306,6 +324,77 @@ export class Textbox extends IText {
     return value.split(this._wordJoiners);
   }
 
+  measureLinesForWrapping(lines: string[]) {
```
ShaMan123, line 327: I don't like the naming

src/shapes/Textbox.ts
```diff
@@ -306,6 +324,77 @@ export class Textbox extends IText {
...
+            width: number;
+            height: number;
+            data: GraphemeBBox<false>[];
+            /**
+             * space to insert before `graphemes`
+             */
+            infix: string | null;
+            infixWidth: number;
+          }[],
+          minWidth: 0,
```
ShaMan123, line 389: not sure this is useful

test/unit/textbox.js
```diff
@@ -406,12 +406,24 @@
...
     });
-    var line1 = textbox._wrapLine('', 0, 100, 0);
+    var line1 = textbox._wrapText([''], 100, 0);
     assert.deepEqual(line1, [[]], 'wrapping without splitByGrapheme');
     textbox.splitByGrapheme = true;
-    var line2 = textbox._wrapLine('', 0, 100, 0);
+    var line2 = textbox._wrapText([''], 100, 0);
     assert.deepEqual(line2, [[]], 'wrapping with splitByGrapheme');
   });
+  QUnit.test('wrapping respects max line width', function (assert) {
```
ShaMan123, line 415: broken on master!

ShaMan123 (commented):
moved the measuring logic and managed to fix a bug due to it

ShaMan123:
If you don't like reduce and prefer for loops I will do that.

src/shapes/Textbox.ts
```diff
@@ -256,11 +258,26 @@ export class Textbox extends IText {
    * @param {Number} desiredWidth width you want to wrap to
    * @returns {Array} Array of lines
    */
-  _wrapText(lines: Array<any>, desiredWidth: number): Array<any> {
-    const wrapped = [];
+  _wrapText(lines: string[], desiredWidth: number, reservedSpace = 0) {
+    const wrapped: string[][] = [];
     this.isWrapping = true;
+    const { lines: data, minWidth } = this.measureLinesForWrapping(lines);
```
ShaMan123, line 264: We should cache this and recalc only when text/styles change.
This will make resizing a lot faster

src/shapes/Textbox.ts
```diff
@@ -256,11 +258,23 @@ export class Textbox extends IText {
...
+    const { data, minWidth } = this.measureLinesForWrapping(lines);
+    const additionalSpace = this._getWidthOfCharSpacing();
+    const maxWidth = Math.max(
+      desiredWidth - reservedSpace,
+      minWidth,
+      this.dynamicMinWidth
+    );
+    if (minWidth + reservedSpace > this.dynamicMinWidth) {
+      this.dynamicMinWidth = minWidth - additionalSpace + reservedSpace;
+    }
```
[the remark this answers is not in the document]
asturur: if i remember correctly maxWidth is the largest space we have at our disposal.
The larges between what we want the texbox to be, or the size of the smallest word.

asturur:
So i put the master code side to side with this PR an i can't understand wich part you extracted.
It seems to me you didn't. I checked the old code and it does this:

It wrap a single line at time.
it takes the line, split all the words, measure each word.
now that knows all the words length in a single line ( but not across lines ) split the single line in one or more line for the textbox wrapper.

the pr change more between last 2 days, and now is completely changing the process of text wrapping in order to early catch the minimum word width to avoid extra breaking in lines, since a long word on the last line could make you wrapping in the early lines not optimal.

I wouldn't call it even a bug, but i understand we want to fix it.

That is a bug and can be fixed, and that is the only part i understand of this PR.
All the other changes are obscure to me.

I can propose, let's see where the bugfix bring us, let's check again the need of dividing the measurement with the word measurement exposed when the bugfix is merged

asturur:
let me know i i understood the largest word width / min width bug 
i tried to fix it here just moving code around.
https://github.com/fabricjs/fabric.js/pull/9004
My laptopm is discharing and i dont have a charger ( issue i should fix tomorrow ).
Let's fix the bug  first and then try to explain what you want from the extra measure method

ShaMan123:
You understand the width breaking line issue correctly

ShaMan123:
Not sure I agree with closing
We should extract more stuff from here

PR #8995, opened 2023-06-07
fix(IText): support control interaction in text editing mode

ShaMan123:
## Motivation

I want to leave text transformable while being edited.

Overriding `_setEditingProps` and removing the part that remove controls when enter editing, gives me a transformable text object, but the selection event handler run togheter with the trasnform when i use the controls.

This exposes the need to disable cursor/selection logic while interacting with a control.

This is seems possible with two simple changes on the handlers.

Needed for work.

## Description

I am not adding tests because this requires an interaction test.

## In Action

Enter text editing and then use the controls to resize the text object
https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-d9r99p?file=%2Fpages%2Findex.tsx%3A13%2C18

https://github.com/fabricjs/fabric.js/assets/34343793/1eff5054-0e05-4b8d-a057-9084fa71f798

### DIFF
--- a/src/shapes/IText/ITextBehavior.ts
+++ b/src/shapes/IText/ITextBehavior.ts
@@ -404,6 +404,10 @@ export abstract class ITextBehavior<
    * called by {@link canvas#textEditingManager}
    */
   updateSelectionOnMouseMove(e: TPointerEvent) {
+    if (this.__corner) {
+      return;
+    }
+
     const el = this.hiddenTextarea!;
     // regain focus
     getDocumentFromElement(el).activeElement !== el && el.focus();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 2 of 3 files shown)


ShaMan123 (commented):
very simple and get the job done!

src/shapes/IText/ITextBehavior.ts
```diff
@@ -404,6 +404,10 @@ export abstract class ITextBehavior<
    * called by {@link canvas#textEditingManager}
    */
   updateSelectionOnMouseMove(e: TPointerEvent) {
+    if (this.__corner) {
+      return;
+    }
```
asturur, line 409: i kind of understand the need o this particular return, but it would be nice if you can add an approximate picture of how your control look like, even a fake one without having to disclose work things, and explain which interaction you make possible that interacts with the mouse move of the selection.

ShaMan123:
done

asturur:
ok now i undertstand. I thought it was some more custom stuff

asturur (approved):

PR #9004, opened 2023-06-09
Fix minimum word width calculation

asturur:
## Motivation

While working on text measurement reorganization, @shaman123 found some possible improvemente in the handlin of the minimum text width/ largest word width.

This PR is the extraction of that improvement from his PR that has too many changes.
reference:
#8994 

## Changes

- words are measured all together before layout.
- adds 2 loops around lines rather than 1 as the recent PR that added looping over the whole line before layouting the single line, before fabric was trying to do measurements while progressing on the line, but this of course came with compromises.

### DIFF
--- a/src/shapes/Textbox.ts
+++ b/src/shapes/Textbox.ts
@@ -426,6 +426,9 @@ export class Textbox extends IText {
 
     i && graphemeLines.push(line);
 
+    // TODO: this code is probably not necessary anymore.
+    // it can be moved out of this function since largestWordWidth is now
+    // known in advance
     if (largestWordWidth + reservedSpace > this.dynamicMinWidth) {
       this.dynamicMinWidth = largestWordWidth - additionalSpace + reservedSpace;
     }


ShaMan123 (commented):
This looks good
I would like to rename the use of words to parts or something else because in case of `splitByGrapheme` it is not words and very confusing.
I also think we should change the signature of `_wrapLine` to be more orderly
`_wrapLine(lie, lineIndex, context)`

Pull in the test from #8994 

Regardless I think we should hold the entire measurement data in `wordsData` and reuse it in case of resizing, it optimizes perf significantly.
If we do that overriding, `measureWords` allows easy customization of wrapping (this is what I am doing) because it will contain of the data needed.

asturur:
> This looks good I would like to rename the use of words to parts or something else because in case of `splitByGrapheme` it is not words and very confusing. I also think we should change the signature of `_wrapLine` to be more orderly `_wrapLine(lie, lineIndex, context)`
> 
> Pull in the test from #8994
> 
> Regardless I think we should hold the entire measurement data in `wordsData` and reuse it in case of resizing, it optimizes perf significantly. If we do that overriding, `measureWords` allows easy customization of wrapping (this is what I am doing) because it will contain of the data needed.

Let me verify if this fix what i think it fixes ( i think is just a case in initialization ), let's see then what is missing for your overrides need, then we can also rename and change signature of private methods in another PR.

src/shapes/Textbox.ts
```diff
@@ -257,15 +265,59 @@ export class Textbox extends IText {
...
     this.isWrapping = false;
     return wrapped;
   }
 
+  /**
+   * For each line of text terminated by an hard line stop,
+   * measure each word width and extract the largest word from all.
+   * @param {string[]} lines the lines we need to measure
+   */
+  measureWords(lines: string[]): WordsWidthData {
```
asturur, line 284: @ShaMan123 this code now works as your for what regards min word width.
You said that this would be enough to also let you override the code, but in that case we need to be specific of the requirements of this functions.
Seems to me that this on top of measuring is also deciding the words that are going to be rendered?
like are you returning different words from this function from the one that are in input?

in that case we need to  rename it accordingly and also specify this in the returned type and description.

test/unit/textbox.js
```diff
@@ -353,6 +353,14 @@
...
     );
   });
+  QUnit.test('Measure words', function(assert) {
+    const textbox = new fabric.Textbox('word word\nword\nword', { width: 300 });
+    const { wordsData, largestWordWidth } = textbox.measureWords(textbox.textLines);
+    assert.deepEqual(
+      wordsData[0],
+      [{ word: 'word', width: largestWordWidth }, { word: 'word', width: largestWordWidth }]
+    );
+  });
```
asturur, line 363: started to add this test but i don't understand why the word comes out as split `['w', 'o', 'r', 'd']`

asturur: oh i see why, to measure we need to split.

src/shapes/Textbox.ts
asturur: this function so far would be something like:

GetGraphemeDataForRender

MeasureAndSplitWords

getSplittedWords

getSplitWordsInfo

src/shapes/Textbox.ts
```diff
@@ -319,47 +372,35 @@ export class Textbox extends IText {
...
       desiredWidth,
       largestWordWidth,
       this.dynamicMinWidth
     );
     // layout words
+    const data = wordsData[lineIndex];
     offset = 0;
     let i;
-    for (i = 0; i < words.length; i++) {
+    for (i = 0; i < data.length; i++) {
```
asturur, line 403: there is probably something else that can be done down here to bring infix width calculation in measure words.
but i m intentionally skipping it for now.
Seems like line 419 could be pused in the else condition after 415.

i m also not convinced at all from the offset++ at line 426, but eventually it works

src/shapes/Textbox.ts
```diff
@@ -257,15 +265,57 @@ export class Textbox extends IText {
...
+        ? this.graphemeSplit(lines[i])
+        : this.wordSplit(lines[i]);
+
+      // fix a difference between split and graphemeSplit
+      if (words.length === 0) {
+        words.push([]);
+      }
+
+      // measure words
+      data[i] = words.map((word) => {
```
ShaMan123, line 303: why not use lines.map?

src/shapes/Textbox.ts
ShaMan123: So I have a few things.
> (ex: swapping last word for ellipsis)

Yes. I am doing that exactly. And I can PR it if you wish.
I am also breaking words, this is why it is crucial to manage the infix here as well because then I can add and measure the infix, or not if I am simply breaking a word. It makes sense to have all measuring in one place, another reason to move infix measuring to here.
We should move this method to Text, it belongs there IMO.
Then the measureChars function should use it instead of remeasuring.
Positioning data (left, top => `getGraphemeDataForRender`) should be applied separately because these props are invalidated more frequently (always actually) where as the words data (width, height) could be cached to improve perf.
I have something POCing this and the perf diff is extremely significant when I apply caching.

src/shapes/Textbox.ts
ShaMan123: Yes, all this is correct.
I did that in the other PR.

ShaMan123: Now it will make sense to you.

src/shapes/Textbox.ts
ShaMan123: And actually it means that `_wrapLine` becomes a very standard method

asturur: Don't poc anything i m not accepting new features, i m trying to let you get the method you need without flipping textbox now.
How can we get to what you need without revolutionin the code now?

src/shapes/Textbox.ts
asturur: welll the previus for loop didn't exist, so i can use lines.map yes

asturur:
@ShaMan123 let me know if this is helpful and if you want to merge it.

ShaMan123 (commented):
This is almost there.
I want the infix as part of this change
Also please rename:
- All the use of words to something else because when splitting by grapheme words are not words and it is awfully confusing.
- getGraphemeDataForRender => this is not for rendering, it is for layout. The rendering measurements are in renderChar

ShaMan123:
without handling the infix I can't control word breaking because `_wrapLine` will push the infix it decides instead of the one I want so this PR will not be good enough.
However it is a step forward.
Moving the infix logic will make both this and #8994 identical.

src/shapes/Textbox.ts
ShaMan123: I am POCing in my app.

ShaMan123:
this part doesn't make sense in `_wrapLine`
https://github.com/fabricjs/fabric.js/pull/9004/files#diff-25ec98d446b9f1934b6a36a8784e77fed6b0bf5304a0f95fc7155ad93dfd6d33R427-R429

ShaMan123:
@asturur please merge this before you publish.
Thus is good enough bevause the lifecycle is fixed. I can subclass now.
We can get back to infix later and then I remove my logic in favor of what we decide

asturur:
> this part doesn't make sense in `_wrapLine` https://github.com/fabricjs/fabric.js/pull/9004/files#diff-25ec98d446b9f1934b6a36a8784e77fed6b0bf5304a0f95fc7155ad93dfd6d33R427-R429

It Doesn't make sense anymore because we moved the calculation up, and this code is now sort of unnecessary.
But i really wanted to do the minimum possible here.

src/shapes/Textbox.ts
```diff
@@ -258,16 +266,58 @@ export class Textbox extends IText {
...
+        word = splitByGrapheme ? word : this.graphemeSplit(word);
+        const width = this._measureWord(word, lineIndex, offset);
+        largestWordWidth = Math.max(width, largestWordWidth);
+        offset += word.length + infix.length;
+        return { word, width };
+      });
+    });
+
+    return {
+      wordsData: data,
```
ShaMan123, line 316: I am not happy about naming

ShaMan123:
> > this part doesn't make sense in `_wrapLine` https://github.com/fabricjs/fabric.js/pull/9004/files#diff-25ec98d446b9f1934b6a36a8784e77fed6b0bf5304a0f95fc7155ad93dfd6d33R427-R429
> 
> It Doesn't make sense anymore because we moved the calculation up, and this code is now sort of unnecessary. But i really wanted to do the minimum possible here.

#8994!! That is why I changed so much, because of all this ugliness

ShaMan123:
I am not 100% happy of naming and overall state of this PR

PR #9007, opened 2023-06-11
fix(ClassRegistry) add the extended names to the registry

### LINKED ISSUE #9006: Minification breaks classRegistry
When using the minified version of fabricJS, the class registry looks like this:

image

This is wrong and comes from the fact that we are relying on constructor name to handle the label.

it should look like this

image

asturur:
## Motivation

This is a pain because it brings back the topic of uppercase vs not.
Now that the class.constructor.name is not usable, there are 2 big issues:

- type / a string in the class  is again needed
- the old values for type are good enough but we can still move to the version similar to the class name becuase of consistency with filters that were always in that way.

This fixes the class registry:
image

closes #9006

### DIFF
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -1,10 +1,8 @@
-import * as fabric from 'fabric/dist/index.min.js';
+import * as fabric from 'fabric';
 import './styles.css';
 
 const el = document.getElementById('canvas');
 const canvas = new fabric.Canvas(el);
-
-window.fabric = fabric;
 
 //  edit from here
 canvas.setDimensions({
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -1,8 +1,10 @@
-import * as fabric from 'fabric';
+import * as fabric from 'fabric/dist/index.min.js';
 import './styles.css';
 
 const el = document.getElementById('canvas');
 const canvas = new fabric.Canvas(el);
+
+window.fabric = fabric;
 
 //  edit from here
 canvas.setDimensions({
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(classRegistry): make the registry name mandatory [#9007](https://github.com/fabricjs/fabric.js/pull/9007)
 - fix(lib): fix aligning_guideline zoom [#8998](https://github.com/fabricjs/fabric.js/pull/8998)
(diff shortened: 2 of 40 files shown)


asturur:
Now this takes care of #9006 but #9008 is very important.

The plan would be to write simple code where the smart code failed.
Every class gets its own `get type` that return the correct string.

@ShaMan123 unless you have some simpler idea ( not smarter, not better design, but just `more simplest` ) i will proceed next days putting back a type string per class, uppercase version for consistency with filters.

asturur:
@melchiar the invite to pull out a simple idea is of course for you to.

CODE-AXION:
hey brother you said you would come back to my PR ?

ShaMan123 (commented):
I would remove the excessive registering of lowercase and have the registry handle that when it tries to resolve the class.

**Can we agree that `type` is a static prop and not a getter??**
This way the registry can register classes automatically as it does now, replacing `this.contructor.name` with `this.contructor.type` 
**The instance should not be able to change its type IMO.**
However, If a getter is a must then it should be
```
get type() { return this.constructor.type; }`
```

Could you add a test instead of changing `.codesandbox/templates/vanilla/src/index.ts`? refer to `unit/env.js` or add to class registry.

ShaMan123:
> ## Motivation
> This is a pain because it brings back the topic of uppercase vs not. Now that the class.constructor.name is not usable, there are 2 big issues:
> 
> * type / a string in the class  is again needed
> * the old values for type are good enough but we can still move to the version similar to the class name becuase of consistency with filters that were always in that way.
> 
> This fixes the class registry: image
> 
> closes #9006

no 30 is undefined

asturur:
The template change was for me to verify and screenshot, i will revert it.
We currently do not have tests for the minified version.

Type can't be static or not, it does not matter imho.
I can also just start from the type PR and see if i can reliably register from the type, and close this one it works.

asturur:
> > closes #9006
> 
> no 30 is undefined

i noticed that and i fixed it with one of the previous commits ( i think it was texbox.ts ) but i couldn't do another screenshot

asturur:
> @asturur hey brother you said you would come back to my PR ?

Can you link it to me? i have thousands of fabricJS email rom the repo and is hard to keep track while trying to finish 6.0 and do normal things.

asturur:
started #9009 but i m on the train is shaking all the time i can't really progress

CODE-AXION:
> > @asturur hey brother you said you would come back to my PR ?
> 
> Can you link it to me? i have thousands of fabricJS email rom the repo and is hard to keep track while trying to finish 6.0 and do normal things.

@asturur 
Here is the link: https://github.com/fabricjs/fabric.js/pull/8955

asturur:
replaced by #9009

PR #9009, opened 2023-06-12
fix(serialization): restore explict class type

### LINKED ISSUE #9006: Minification breaks classRegistry

### LINKED ISSUE #9008: Minification breaks serialization
Currently serialization is broken.
While serialization and deserialization may work in the same app on the same build o fabricJS, there is no guarantee it will work across different builds.

Currenty when exporting an object we assign as type the name of the minified variable of the class:

image

That name is not stable and won't work across differnt builds.

Serialization in the current state is broken

asturur:
## Motivation

During the migration to "modern js" we thought that having a `type` property on an instance was kind of a duplicate because classes have costructor.name that is basically what type was.

Dumbly enough we didn't think that minification would have changed variable names and so constructor.name.
This wasn't caught by tests since tests run on the non minified version of the code.

This PR aims at fixing:
#9006 and #9008 making #9007 obsolete.

It restores class type, as a static property that will take the place o constructor.name ( constructor.type ).
The compatibility getter for type on the instances is still there and will return the static type.

closes #9006 
closes #9008

### DIFF
--- a/src/filters/ColorMatrixFilters.ts
+++ b/src/filters/ColorMatrixFilters.ts
@@ -7,6 +7,8 @@ export function createColorMatrixFilter(key: string, matrix: number[]) {
       return key;
     }
 
+    static type = key;
+
     static defaults = {
       ...colorMatrixDefaultValues,
       /**
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -269,6 +269,15 @@ export class FabricObject<
   }
 
   /**
+   * The class type. Used to identify which class this is.
+   * This is used for serialization purposes and internally it can be used
+   * to identify classes. As a developer you could use `instance of Class`
+   * but to avoid importing all the code and blocking tree shaking we try
+   * to avoid doing that.
+   */
+  static type = 'FabricObject';
+
+  /**
    * Legacy identifier of the class. Prefer using utils like isType or instanceOf
    * Will be removed in fabric 7 or 8.
    * The setter exists because is very hard to catch all the ways in which a type value
@@ -278,7 +287,7 @@ export class FabricObject<
    * @deprecated
    */
   get type() {
-    const name = this.constructor.name;
+    const name = this.constructor.type;
     if (name === 'FabricObject') {
       return 'object';
     }
@@ -515,7 +524,7 @@ export class FabricObject<
           : null,
       object = {
         ...pick(this, propertiesToInclude as (keyof this)[]),
-        type: this.constructor.name,
+        type: this.constructor.type,
         version: VERSION,
         originX: this.originX,
         originY: this.originY,
@@ -609,7 +618,7 @@ export class FabricObject<
    * @return {String}
    */
   toString() {
-    return `#<${this.constructor.name}>`;
+    return `#<${this.constructor.type}>`;
   }
 
   /**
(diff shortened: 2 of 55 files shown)


ShaMan123 (commented):
Looks good

test/unit/object.js
```diff
@@ -293,7 +293,7 @@
 
   QUnit.test('toString', function (assert) {
     class Moo extends fabric.Object {
-      static type = 'moo'
+      static type = 'Moo'
```
asturur, line 296: NO IDEA why this class already had a static type and what was testing.

src/filters/ColorMatrixFilters.ts
```diff
@@ -7,6 +7,8 @@ export function createColorMatrixFilter(key: string, matrix: number[]) {
       return key;
     }
```
ShaMan123, line 8: should be removed, extending ColorMatrix is enough

src/shapes/Object/Object.ts
```diff
@@ -1454,7 +1463,10 @@ export class FabricObject<
    * @return {Boolean}
    */
   isType(...types: string[]) {
```
ShaMan123, line 1465: should we deprecate this also??

ShaMan123 (commented):
This looks good.
This will also solve my Bug since we disconnected class name and type (but I still think my PRs are valid but they an wait now).
I significant comment and another one for discussion

CaptEmulation:
don't we need to update `ClassRegistry.ts` as well to use `.type`?

asturur:
> don't we need to update `ClassRegistry.ts` as well to use `.type`?

i think i did it? i m sure i did it. Let me double check.

asturur:
@CaptEmulation 
image

Is because tabs designs sucks.
We need vertical tabs since most of the screens are now 16:9.

At this point i need a test.

asturur:
@ShaMan123  we verified this unblock my teammates at work.
So please for us is good, give it a review, and if there are no hiccups  i ll try to publish this tomorrow at the first occasion.

src/shapes/Object/Object.ts
asturur: not part of this PR.
isType is not directly related to the incident. nor `isType` is wrong per se.

src/filters/ColorMatrixFilters.ts
asturur: those had their own name before, will keep them.
I m not even sure is possible with the different defualts. Without their type class they would serialize to ColorMatrix and then to the next deserialize wouldn't work anymore

ShaMan123:
> > don't we need to update `ClassRegistry.ts` as well to use `.type`?
> 
> i think i did it? i m sure i did it. Let me double check.

How did tests pass then?

ShaMan123 (dismissed):
I understand now what happened with tests.
Good

src/filters/ColorMatrixFilters.ts
ShaMan123: I want this removed because I think in the future it is a potential bug
Once we remove type getter this thing will become an artifact

ShaMan123: > I m not even sure is possible with the different defualts

??

PR #9018, opened 2023-06-14
chore(TS) Add more types to Text, BREAKING: change return type of _getStyleDeclaration

asturur:
## Motivation

I start to not feel confindent with large classes having ts-nocheck still in place.
Especially text.

Those types i did are not perfect, but better than nothing.

Before merging, discussing if we should get rid of stroke and fill being null since it doesn't makes much sense.
We should understand what null does today ( skill fill? ) and use maybe `transparent` in place of it and skip fill when `transparent` is used.
It simplify types a lot and makes every fill not skipped safe anyway.


src/shapes/IText/IText.ts
```diff
@@ -497,10 +497,14 @@ export class IText<
       lineIndex = cursorLocation.lineIndex,
       charIndex =
         cursorLocation.charIndex > 0 ? cursorLocation.charIndex - 1 : 0,
-      charHeight = this.getValueOfPropertyAt(lineIndex, charIndex, 'fontSize'),
+      charHeight = this.getValueOfPropertyAt<'fontSize'>(
```
asturur, line 500: Before it gets asked, i tried to derive the type from the third argument right away from inside the function  `getValueOfPropertyAt` , but it does not work.
This return string/number reliably so far.

src/shapes/IText/IText.ts
```diff
@@ -668,9 +676,9 @@ export class IText<
    * Unused by the library, is for the end user
    * @return {String | TFiller} Character color (fill)
    */
-  getCurrentCharColor(): string | TFiller {
+  getCurrentCharColor(): string | TFiller | null {
```
asturur, line 679: fill and stroke being null is the cause of many issue with types and a lot of extra null checks.

src/shapes/Text/StyledText.ts
```diff
@@ -179,21 +181,18 @@ export abstract class StyledText<
...
-  private _extendStyles(index: number, styles: TextStyleDeclaration) {
+  private _extendStyles(index: number, styles: TextStyleDeclaration): void {
     const { lineIndex, charIndex } = this.get2DCursorLocation(index);
 
     if (!this._getLineStyle(lineIndex)) {
       this._setLineStyle(lineIndex);
     }
 
-    if (!this._getStyleDeclaration(lineIndex, charIndex)) {
+    if (!Object.keys(this._getStyleDeclaration(lineIndex, charIndex)).length) {
```
asturur, line 191: this is the only code change i made.
there were 5 occurrencies in which returning undefined from `_getStyleDeclaration` needed to be fixed with an `|| {}` while just one where returning undefined ( or null ) was useful, to know if there was no style.

src/shapes/Text/StyledText.ts
```diff
@@ -246,14 +243,18 @@ export abstract class StyledText<
...
+   * @return {Object} style object a REFERENCE to the existing one or a new empty object
    */
-  _getStyleDeclaration(lineIndex: number, charIndex: number) {
+  _getStyleDeclaration(
+    lineIndex: number,
+    charIndex: number
+  ): TextStyleDeclaration {
     const lineStyle = this.styles && this.styles[lineIndex];
-    return lineStyle ? lineStyle[charIndex] : null;
+    return lineStyle ? lineStyle[charIndex] : {};
```
asturur, line 257: this is the code change that caused the object.keys above.

src/shapes/Text/Text.ts
```diff
@@ -63,20 +69,16 @@ type TPathAlign = 'baseline' | 'center' | 'ascender' | 'descender';
  * needs the the info of previous graphemes already filled
  * Override to customize measuring
  */
-export type GraphemeBBox<onPath = false> = {
+export type GraphemeBBox = {
```
asturur, line 72: this was too hard to be flexible ( or simply i can't use it )
I m thinking that maybe we just keep al the properties required and we just add the extra 3 everywhere.

src/shapes/Text/Text.ts
```diff
@@ -1528,22 +1566,23 @@ export class Text<
       const lineLeftOffset = this._getLineLeftOffset(i);
       let boxStart = 0;
       let boxWidth = 0;
-      let lastDecoration = this.getValueOfPropertyAt(i, 0, type);
-      let lastFill = this.getValueOfPropertyAt(i, 0, 'fill');
+      let lastDecoration = this.getValueOfPropertyAt<typeof type>(i, 0, type);
```
asturur, line 1569: typeof type is fantastic

src/shapes/Text/constants.ts
```diff
@@ -87,7 +87,7 @@ export const textDefaultValues: Partial<TClassProperties<Text>> = {
   textBackgroundColor: '',
   stroke: null,
   shadow: null,
-  path: null,
+  path: undefined,
```
asturur, line 90: not sure why null was chose in first place.

src/shapes/Textbox.ts
```diff
@@ -244,7 +252,7 @@ export class Textbox extends IText {
    * @param {Object} style
    * @private
    */
-  _setLineStyle(lineIndex: number) {
+  protected _setLineStyle(lineIndex: number) {
```
asturur, line 255: those are protected in the other classes

asturur:
This is ready to be discussed @ShaMan123  @melchiar also @jiayihu since you are around TS.
This is not meant to be perfect, the goal is to remove ts-nocheck from as much files as possible

test/unit/text.js
```diff
@@ -59,7 +59,7 @@
     skewY:                      0,
     charSpacing:                0,
     styles:                     [],
-    path:                       null,
+    path:                       undefined,
```
asturur, line 62: i wonder if i need to make a backward compatibilitu statement or change somewhere.
The code never check if path is undefined, all it does is checking if path is truthy or not.
So null or undeined is the same also when reloading old data

src/shapes/IText/IText.ts
ShaMan123: if we removing null why this this needed?

src/shapes/Text/StyledText.ts
ShaMan123: I guess this is fine but BREAKING if anyone used it before

src/shapes/Text/StyledText.ts
```diff
@@ -246,14 +243,18 @@ export abstract class StyledText<
...
+   * @return {Object} style object a REFERENCE to the existing one or a new empty object
    */
-  _getStyleDeclaration(lineIndex: number, charIndex: number) {
+  _getStyleDeclaration(
+    lineIndex: number,
+    charIndex: number
+  ): TextStyleDeclaration {
     const lineStyle = this.styles && this.styles[lineIndex];
-    return lineStyle ? lineStyle[charIndex] : null;
+    return lineStyle ? lineStyle[charIndex] ?? {} : {};
```
ShaMan123, line 257: Are we ok with abbreviations now or not yet?
```suggestion
    return lineStyle?.[charIndex] ?? {};
```

src/shapes/Text/Text.ts
```diff
@@ -40,6 +44,8 @@ import {
   JUSTIFY_RIGHT,
 } from './constants';
 import { CENTER, LEFT, RIGHT, TOP, BOTTOM } from '../../constants';
+import { isFiller } from '../../util/typeAssertions';
+import type { Gradient, Pattern } from 'fabric/node';
```
ShaMan123, line 48: Hello

ShaMan123: I hate this
I thought of adding a step to publishing (altering `package.json`) so this won't happen to us.

src/shapes/Text/Text.ts
```diff
@@ -773,9 +784,9 @@ export class Text<
    */
   _measureChar(
     _char: string,
-    charStyle: TextStyleDeclaration,
+    charStyle: CompleteTextStyleDeclaration,
     previousChar: string | undefined,
-    prevCharStyle: any
+    prevCharStyle: CompleteTextStyleDeclaration | Record<string, never>
```
ShaMan123, line 789: what does this mean?
Or a style or any empty object?

src/shapes/Text/Text.ts
```diff
@@ -818,12 +829,13 @@ export class Text<
         // we can measure the kerning couple and subtract the width of the previous character
         coupleWidth = ctx.measureText(couple).width;
         fontCache[couple] = coupleWidth;
-        kernedWidth = coupleWidth - previousWidth;
+        // safe to use the non-null since if undefined we defined it before.
```
ShaMan123, line 832: this belongs on line 838 I think

src/shapes/Text/Text.ts
```diff
@@ -1492,12 +1529,13 @@ export class Text<
    * @param {String} property the property name
    * @returns the value of 'property'
    */
-  getValueOfPropertyAt(lineIndex: number, charIndex: number, property: string) {
+  getValueOfPropertyAt<T extends StylePropertiesType>(
+    lineIndex: number,
+    charIndex: number,
+    property: StylePropertiesType
+  ): this[T] {
```
ShaMan123, line 1536:
```suggestion
  getValueOfPropertyAt<T extends StylePropertiesType>(
    lineIndex: number,
    charIndex: number,
    property: T
  ): this[T] {
```

Seems this is what you want so TS infers it instead of you explicitly telling it.


image

image

Autocomplete
image

src/shapes/Text/Text.ts
```diff
@@ -1780,7 +1820,7 @@ export class Text<
     const parsedAttributes = parseAttributes(element, Text.ATTRIBUTE_NAMES);
 
     const {
-      textAnchor = LEFT,
+      textAnchor = LEFT as typeof LEFT | typeof CENTER | typeof RIGHT,
```
ShaMan123, line 1823: Maybe you should add this as top level type `HAlign` `VAlign`

src/shapes/Textbox.ts
```diff
@@ -4,6 +4,7 @@ import { IText } from './IText/IText';
 import { classRegistry } from '../ClassRegistry';
 import { createTextboxDefaultControls } from '../controls/commonControls';
 import { JUSTIFY } from './Text/constants';
+import type { TextStyleDeclaration } from 'fabric/node';
```
ShaMan123, line 7: Second time makes me itch
I will PR that fix probably...

src/shapes/Textbox.ts
```diff
@@ -187,11 +188,14 @@ export class Textbox extends IText {
...
+    charIndex: number
+  ): TextStyleDeclaration {
     if (this._styleMap && !this.isWrapping) {
       const map = this._styleMap[lineIndex];
       if (!map) {
-        return null;
+        return {};
       }
       lineIndex = map.line;
       charIndex = map.offset + charIndex;
```
ShaMan123, line 201: though this is not related I would like to say that this piece of logic has a lot more usages than just in style.
It maps between a wrapped position to a non wrapped position.
It is confusing when to use the wrapped/non wrapped position when working on textbox, error prone and not friendly.
I would have a method that returns the offset give a line, char, isWrapped since it is not changed by wrapping (I hope there is no edge case for that) and use that across all methods as a standard.
This will make working on Text and Textbox easier or else we must keep in mind this constant difference that is applied at a certain point in the lifecycle. e.g. measuring uses to non wrapped, rendering uses the wrapped.
All should us the offset.

ShaMan123 (commented):
- I want to understand the problem with `GraphemeBBox<true | false>`

Apart from that looks good

src/shapes/IText/IText.ts
asturur: because i m not removing null in this PR.
But this PR brought up the issue again

src/shapes/Text/StyledText.ts
asturur: is so much longer the abbreviation because safari.

asturur: i m also not sure about performance in general of that abbrevieation

src/shapes/Text/Text.ts
asturur: oopppsss i ll fix it.
This is the stupid VSCODE.
The automatic import MICROSOFT NEVER WORK things

src/shapes/Text/Text.ts
asturur: yes. or complete or empty.
Not sure why, i just satisfied it.

src/shapes/Text/Text.ts
asturur: i tried this as i stated before, it doesn't work.

asturur: oh no what i tried was slightly different.
This works.
I didn't remove styleProperty from the argument i tried to use it.

src/shapes/Text/Text.ts
asturur: meh, parseAttributes needs to be typed at some point.
Not wasting time on it just for this particular use case

src/shapes/Textbox.ts
asturur: as i said there are so many things that have been bolted on top of normal text, and this stylemap was one.
This works, is confusing, but it works,

src/parser/applyViewboxTransform.ts
```diff
@@ -1,4 +1,4 @@
-//@ts-nocheck
+// @ts-nocheck
```
asturur, line 1: this i did half, then i had to stop. too hard

src/parser/doesSomeParentMatch.ts
```diff
@@ -1,19 +1,18 @@
-//@ts-nocheck
```
asturur, line 1: this was an easy one.
parentNode was used instead o parentElement because IE i think

src/shapes/Textbox.ts
asturur: fixed

src/shapes/Text/Text.ts
asturur: I m commenting on the `!` for ` previousWidth`

asturur:
> * I want to understand the problem with `GraphemeBBox<true | false>`

GraphemeBBox starts as <false> but then later is true, and there is no easy way to inform the code that it has been updated because is done with mutation through different methods

asturur:
Needs to decide what to do with the breaking change.

ShaMan123:
> > * I want to understand the problem with `GraphemeBBox<true | false>`
> 
> GraphemeBBox starts as <false> but then later is true, and there is no easy way to inform the code that it has been updated because is done with mutation through different methods

Ie ould make that no mutate but return a new object regardless of types 
But that is for another cleanup

src/canvas/DOMManagers/CanvasDOMManager.ts
```diff
@@ -82,8 +82,8 @@ export class CanvasDOMManager extends StaticCanvasDOMManager {
   ) {
     setStyle(element, {
       position: 'absolute',
-      left: 0,
-      top: 0,
+      left: '0',
+      top: '0',
     });
```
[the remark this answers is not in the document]
ShaMan123: hmm I see that in CCSStyleDeclration that is string anyways.

src/util/dom_style.ts
```diff
@@ -1,20 +1,15 @@
-//@ts-nocheck
 // TODO this file needs to go away, cross browser style support is not fabricjs domain.
-
 /**
  * wrapper for setting element's style
  * @param {HTMLElement} element
  * @param {Object | string} styles
  */
-export function setStyle(element, styles) {
+export function setStyle(element: HTMLElement, styles: Record<string, string>) {
```
ShaMan123, line 7:
```suggestion
export function setStyle(element: HTMLElement, styles: Partial<CSSStyleDeclaration>) {
```

src/util/dom_style.ts
```diff
@@ -1,20 +1,15 @@
...
   const elementStyle = element.style;
   if (!elementStyle) {
     return;
-  } else if (typeof styles === 'string') {
-    element.style.cssText += ';' + styles;
```
ShaMan123, line 14: I remember a bug that was solved with this....
I might be wrong.
It was to do with restoring the lower canvas style.

ShaMan123 (commented):
left overs

src/util/dom_style.ts
asturur: ok... let me try to look at the history and see if i figure out

src/canvas/DOMManagers/CanvasDOMManager.ts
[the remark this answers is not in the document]
asturur: setAttribute wants strings.
is call setStyles but we are not setting the CSS we are setting the attributes

src/util/dom_style.ts
asturur: i ll revert the changes to this function, this deserve a mini PR and an extra mini clarificaiton.

src/shapes/Text/StyledText.ts
asturur: i would make this breaking, and i someone asks, i would add the hasStyleDeclaration method that return true/false, or maybe i can add it right away

asturur:
i want to merge this and move forward with more types.

src/shapes/Text/StyledText.ts
ShaMan123: I think you should
Then there is a clear migration path (Object.keys is pretty clear also but...)

ShaMan123:
Go ahead!

PR #8616, opened 2023-01-20
test(): add playwright for UI testing

luciemars:
## Motivation

Add Playwright UI testing framework.

## Description

Add playwright dependency with config file and examples of test (1 passing and 1 failing).
Running the tests will open the test report in a browser.

PR #9019, opened 2023-06-14
fix(IText): empty line selection

ShaMan123:
## Description

Selecting an empty line was buggy.
I will upload a video and a repro and hopefully a test proving it

Steps to reproduce:
- create Textbox
- add new lines
- click across an empty line => will jump a line sometimes
- deselect textbox
- double click textbox => will not show cursor because selectionEnd < selectionStart

## In Action

Fixed: https://codesandbox.io/s/jlnq83 => I don't know what happened to the parcel codesandbox. It is not working

BUG repro:

https://github.com/fabricjs/fabric.js/assets/34343793/70f0a5a4-93d1-4857-aaa3-9269ab3ddd57


src/shapes/IText/ITextBehavior.ts
```diff
@@ -313,19 +313,23 @@ export abstract class ITextBehavior<
    * @param {Number} direction 1 or -1
    * @return {Number} Index of the beginning or end of a word
    */
-  searchWordBoundary(selectionStart: number, direction: number): number {
+  searchWordBoundary(selectionStart: number, direction: 1 | -1): number {
```
ShaMan123, line 316: we may want to safeguard `direction = Math.sign(direction )`

src/shapes/IText/ITextBehavior.ts
```diff
@@ -336,14 +340,13 @@ export abstract class ITextBehavior<
...
-        selectionStart,
-        1
-      ); /* search forward */
+    // search backwards
+    const newSelectionStart = this.searchWordBoundary(selectionStart, -1),
+      // search forward
+      newSelectionEnd = Math.max(
+        newSelectionStart,
+        this.searchWordBoundary(selectionStart, 1)
+      );
```
ShaMan123, line 349: added a safeguard

ShaMan123 (commented):
squashed `_getNewSelectionStartFromOffset` into `getSelectionStartFromPointer`

test/unit/itext.js
```diff
@@ -426,6 +426,10 @@
       iText.selectWord(7);
       assert.equal(iText.selectionStart, 5); // |foo|
       assert.equal(iText.selectionEnd, 8);
+
+      iText.selectWord(17);
+      assert.equal(iText.selectionStart, 17); // |\n|
+      assert.equal(iText.selectionEnd, 17);
```
ShaMan123, line 432: broken on master!

test/unit/itext_click_behaviour.js
ShaMan123, line 1: Sorry for this ugly test
I need playwright to do this properly.
You are welcome to revert this

ShaMan123 (commented):
Ready for review

src/shapes/IText/IText.ts
```diff
@@ -39,7 +39,7 @@ export const iTextDefaultValues = {
   caching: true,
   hiddenTextareaContainer: null,
   _selectionDirection: null,
-  _reSpace: /\s|\n/,
+  _reSpace: /\s|\r?\n/,
```
asturur, line 42: Since you are not so old, do you know what \r is and why is there together with \n ?

ShaMan123: I don't...
Return carriage?
Something from the past?

ShaMan123: I don't care too much about it but maybe for compatibility

test/unit/itext_click_behaviour.js
```diff
@@ -124,33 +124,4316 @@
...
+          49,
+          49,
+          49,
+          49,
+          49,
+          49,
+          49,
+          49
+        ]
+      ]);
```
asturur, line 4436: this seems the wrong way to test this.
What is wrong with the old tests?

asturur:
i thought it was simpler to read, i ll come back to this PR next.

test/unit/itext_click_behaviour.js
ShaMan123: The old test is too hard to fix

> this seems the wrong way to test this.

I agree, let's merge #8616 and then I add tests

src/shapes/IText/IText.ts
asturur: it was the special code that made the printer head go back to start of the line, while the '\n' was advancing the paper o ne line.
If you didn't add `\r` you would print on the same line of paper over and over.

Or sometimes you would do

```
BOLD TEXT\n
BOLD TEXT\r\n
```
to print twice on purpose and have the print darker.

things i learned working with old dot matrix printers

src/shapes/IText/ITextBehavior.ts
asturur: maybe no, here is type there is doc, you are welcome to break your own software if you really don't want to know how this is supposed to work.

src/shapes/IText/ITextClickBehavior.ts
```diff
@@ -261,54 +262,31 @@ export abstract class ITextClickBehavior<
...
-          : 1;
-    let newSelectionStart = index + offset;
-    // if object is horizontally flipped, mirror cursor location from the end
-    if (this.flipX) {
-      newSelectionStart = jlen - newSelectionStart;
-    }
-
-    if (newSelectionStart > this._text.length) {
-      newSelectionStart = this._text.length;
-    }
```
asturur, line 310: i want to do some very minor tweaks but before accepting delething this i want to be sure this code here didn't hide some RTL logic that with Math.abs we loose? Did you try in RTL ?

src/shapes/IText/ITextBehavior.ts
```diff
@@ -313,19 +313,23 @@ export abstract class ITextBehavior<
...
     const text = this._text;
-    let index = this._reSpace.test(text[selectionStart])
-        ? selectionStart - 1
-        : selectionStart,
+    // if we land on a space we move the cursor backwards
+    // if we are searching boundary end we move the cursor backwards ONLY if we don't land on a line break
+    let index =
+        selectionStart > 0 &&
+        this._reSpace.test(text[selectionStart]) &&
+        (direction === -1 || !reNewline.test(text[selectionStart - 1]))
```
asturur, line 323: i compare direction to -1 rather than to 0. and i inverted the IF condition order, from easiest to calculate to hardest

asturur:
@ShaMan123 please follow up with verifying RTL selection still works good ( I think so because RTL selection happens in LTR way )

asturur (approved):

src/shapes/IText/IText.ts
ShaMan123: interesting
so \r should be added, right?

src/shapes/IText/ITextBehavior.ts
ShaMan123: because it move back if it encounters \n and then if that is the only char on line it does move forward
The test proves it

src/shapes/IText/IText.ts
ShaMan123: > interesting ~so \r should be added, right?~

so \r should NOT be added, right?

ShaMan123 (commented):
RTL is broken on v5, v6
#7864 needs porting, that fixes selection.
Since this logic regards char offset I think it is safe

asturur:
RTL on v5 is broken? i remember the basic stuff was working

asturur:
> RTL is broken on v5, v6 #7864 needs porting, that fixes selection. Since this logic regards char offset I think it is safe

not for v6.

asturur:
image

Something happened, it wasn't so broken in v5 nor in v6.
Selection in v6 seems more broken tho.
i need to try without this commit

asturur:
Regardless of this PR v6 selection is more broken than v5 for RTL.

ShaMan123:
I can port that pr the upcoming days and see if it fixes this

asturur:
no please not as of now is just a distraction

ShaMan123:
I noted to myself that changing the cursor calc in get2DCursor has changed rtl because I think now it is accounted for twice. But I forgot to test it.
Anyways...

asturur:
i didn't even know we changed `get2DCursor`

asturur:
Do you remember what PR was?

ShaMan123:
I was confused, I meant #8972 
So it should not affect rendering...

PR #9020, opened 2023-06-14
fix(IText): `exitEditing` should clear contextTop

ShaMan123:
## Description

Reproduce:
- select text
- deselect text by pressing the canvas

Expected selection to be cleared.
It doesn't
This is a regression caused IMO by making `abortCursorAnimation` effecient.
We need to think about this.
I am not sure if there are more cases we want to clear selection.
The only path that I am not sure about is this:

https://github.com/fabricjs/fabric.js/blob/00bbc7ce2081d37c273fcdf4804ea2459fc2736b/src/shapes/IText/ITextKeyBehavior.ts#L106-L108

The rest seem fine.
I am quite sure that draggable text needs blur NOT to clear selection, but that might be incorrect. Due to `_renderDragEffects` handling that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [next]
 
 - fix(IText): `exitEditing` should clear contextTop [#9020](https://github.com/fabricjs/fabric.js/pull/9020)
+
 ## [6.0.0-beta9]
 
 - fix(fabric): Fix the serialization and registry dependency from minification [#9009](https://github.com/fabricjs/fabric.js/pull/9009)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(IText): `exitEditing` should clear contextTop [#9020](https://github.com/fabricjs/fabric.js/pull/9020)
 ## [6.0.0-beta9]
 
 - fix(fabric): Fix the serialization and registry dependency from minification [#9009](https://github.com/fabricjs/fabric.js/pull/9009)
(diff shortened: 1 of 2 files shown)


asturur:
> This is a regression caused IMO by making `abortCursorAnimation` effecient.

you mean making `abortCursorAnimation` broken.
Those are just points to my stance, don't touch.

asturur (dismissed):

asturur (approved):

asturur:
Are you verified since you enabled controller on textbox that when text-selection is active, scaling down the textbox works correctly and doesn't leave behind a trail of uncleaned selection?

asturur (approved):

ShaMan123:
> Are you verified since you enabled controller on textbox that when text-selection is active, scaling down the textbox works correctly and doesn't leave behind a trail of uncleaned selection?

You mean this?


https://github.com/fabricjs/fabric.js/assets/34343793/9d9b5155-d675-4cc0-87a8-429a05636c06

asturur:
yes because we are clearing up the top layer only on the texbox area, i m not sure if that means sometimes the 2 sizes aren't in sync. 

I would expect it maybe as an edge case while shrinking equally from the BR quicky

ShaMan123:
These are the paths which call clearContextTop:
- `_renderDragEffects`, `dragEndHandler`: drag effects clear context top and are tested I think
- `abortCursorAnimation`
- `onKeyDown` if navigating with arrows etc.
- `_exitEditing`
- `renderCursorOrSelection`: this is the case you are referring to, right?

I am not sure how it is handled but take notice that in the video I resized using `set`, which is how the resize control does it and since that worked as expected I think we are good.
I checked the stack trace and indeed it is handled by `renderCursorOrSelection` called from `render`.
I think I worked on this before but I can't recall how it works correctly if the size changes.
But it does

PR #9022, opened 2023-06-14
fix(): #9009 artifacts

ShaMan123:
## Motivation

> I want this removed because I think in the future it is a potential bug
Once we remove type getter this thing will become an artifact

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/pull/9009#discussion_r1227187374_
            

## Description

I did not imagine that would happen so fast.


ShaMan123 (commented):
Fixes `textbox.type` and removes future bug potential with the filters.

asturur:
mmmmm
those get type are there for backaward compat?
What is the bug actually?

ShaMan123:
The bug is textbox.type==='i-text'
Filters have a grtter from super class, enogh just to set static type

asturur:
in fabric6 instances need to have a type property that return the fabric 5 style type.
We stopped using internally and we will remove if forward in time.
That `i-text' is on purpose.

src/filters/ColorMatrixFilters.ts
```diff
@@ -3,10 +3,6 @@ import { classRegistry } from '../ClassRegistry';
 
 export function createColorMatrixFilter(key: string, matrix: number[]) {
   const newClass = class extends ColorMatrix {
-    get type() {
-      return key;
-    }
```
asturur, line 8: if this still works because of BaseFilter, this we can remove

src/shapes/IText/IText.ts
```diff
@@ -204,10 +204,6 @@ export class IText<
 
   static type = 'IText';
 
-  get type() {
-    return 'i-text';
-  }
```
asturur, line 209: this should stay.

ShaMan123: But is breaks subclassing
Textbox is broken

ShaMan123: We should instead override isType like you did with object

ShaMan123: Fixed like you did with object

.codesandbox/templates/vanilla/src/index.ts
```diff
@@ -2,7 +2,7 @@ import * as fabric from 'fabric';
 import './styles.css';
 
 const el = document.getElementById('canvas');
-const canvas = new fabric.Canvas(el);
+const canvas = (window.canvas = new fabric.Canvas(el));
```
ShaMan123, line 5: for ease of dev

ShaMan123 (commented):
This is ready and satisfies all requirements

asturur (approved):

asturur (approved):

PR #9030, opened 2023-06-19
refactor(): Remove storage + cssRules, gradientDefs, clipPathDefs globals

asturur:
## Motivation

I wanted to remove the 3 globals cited in the title.

## Description

gradientDefs, clipPathDefs and Cssrules were global because with the old code we couldn't await anything.
We assigned every SVG parsing an id ( svguid ), and then we would keep a registry called storage, where we had 3 properties:
clippaths, gradients, and css rules.
In each of those 3 properties we had an object of which a property assigned to svguid would contain the data for our parsin sessions.

Now this data is local to the parser class.
The parser injects the css rules in the fromElement function that then pushes it in the parseAttributes function.
ClipPaths and gradientDefs are consume in the parser so they didn't need much work.

The old code was just bad, was one of the first things i did for fabricJS in 2014 and i just extended existing things without really diving deep in how/why.

## Changes

fromElement has now 3 arguments.


src/parser/getCSSRules.ts
```diff
@@ -23,39 +24,40 @@ export function getCSSRules(doc: Document) {
...
+          if (_rule === '') {
+            return;
+          }
+          allRules[_rule] = {
+            ...(allRules[_rule] || {}),
+            ...ruleObj,
+          };
+        });
       });
-    });
```
asturur, line 58: looks bad but i really just removed the var  `rules` and chained 3 array methods of which we didn't need the return value

src/parser/parseAttributes.ts
```diff
@@ -17,59 +17,58 @@ import { setStrokeFillOpacity } from './setStrokeFillOpacity';
...
-    ownAttributes[fSize] = fontSize = parseUnit(
-      ownAttributes[fSize],
-      parentFontSize
-    );
+    fontSize = parseUnit(ownAttributes[fSize], parentFontSize);
+    ownAttributes[fSize] = `${fontSize}`;
   }
 
-  const normalizedStyle = {};
+  const normalizedStyle: Record<string, string> = {};
```
asturur, line 71: i m not sure what happened here, i will rely on tests.

asturur:
No tests failing, super shady.

asturur:
i m sure it require more tests

PR #9036, opened 2023-06-19
chore(TS) remove ts-nocheck from image class 

asturur:
## Motivation

Continue to enable type-checks.

## Description

Breaking changes:
- divided the option object and the loadingOption object in the static FromURL method.


src/shapes/Image.ts
```diff
@@ -199,7 +198,7 @@ export class Image<
     this.setElement(
       typeof arg0 === 'string'
         ? ((
-            (this.canvas && getElementDocument(this.canvas.getElement())) ||
+            (this.canvas && getDocumentFromElement(this.canvas.getElement())) ||
```
asturur, line 201: buuuug

ShaMan123: TS helps sometimes

src/shapes/Image.ts
```diff
@@ -794,16 +794,14 @@ export class Image<
...
+      f && enlivenObjects<BaseFilter>(f, options),
       // TODO: redundant - handled by enlivenObjectEnlivables
-      rf && enlivenObjects([rf], options),
+      rf && enlivenObjects<BaseFilter>([rf], options),
       enlivenObjectEnlivables(object, options),
     ]).then(([el, filters = [], [resizeFilter] = [], hydratedProps = {}]) => {
       return new this(el, {
         ...object,
-        src,
-        crossOrigin,
```
asturur, line 806: i suspect this is the cause of some breakage in image.clone i can't explain.

Issue #9010, opened 2023-06-12
Text style bug?


ShaMan123:
https://github.com/fabricjs/fabric.js/blob/8536cb6149337ac30443d3931b648f136a85dea1/src/shapes/Text/Text.ts#L1609-L1629

Is this a bug?

```diff

- family = this.fontFamily, 
+ family = style.fontFamily, 

```

@asturur ?

asturur:
eh...
so the first bug seems that even if we pass a `styleObject`, this `styleObject` could not have a fontFamily.
For me the correct logic is that fontFamily is either `styleObject.fontFamily` || `this.fontFamily`, and this is valid for everything else, fontStyle, fontWeight and fontSize.

I don't remember the story of this function, please do a deep git blame and try to recover the history of it.

asturur:
i would rather have `styleObject` not optional
and do something like
```
const { fontSize = this.fontSize, fontFamily = this.fontFamily, fontWeight = this.fontWeight, fontStyle = this.fontStyle } = styleObject;
```

But recovering the history is better.

It could totally be me that i have seen an opportunity to save some repetition and i did this `style = styleObject || this`

VardaQuraishi:
@ShaMan123 is this issue available to work on?

VardaQuraishi:
I have made the changes and opened a pull request to it.

asturur:
```
I don't remember the story of this function, please do a deep git blame and try to recover the history of it.
```

Did anyone did a deep git blame to see how this function changed in time?

ShaMan123:
I didn't

ShaMan123:
A bug

BEFORE
https://github.com/fabricjs/fabric.js/commit/b112b3405f75702c5bd1a20dc0e756321d0f9a31#diff-9ba978ec9931806800697f52edd5e63429cd10aad6a5da2d6081cc0d46e44881L832-L839

BLAME
https://github.com/fabricjs/fabric.js/compare/0c65644..abe813f#diff-9ba978ec9931806800697f52edd5e63429cd10aad6a5da2d6081cc0d46e44881R1205-R1210

PR #9041, opened 2023-06-21
Update Text.ts

VardaQuraishi:
This is a fix to Text style bug? #9010 issue.
I have tested the code and it is passing all checks.

### DIFF
--- a/src/shapes/Text/Text.ts
+++ b/src/shapes/Text/Text.ts
@@ -100,21 +100,22 @@ interface UniqueTextProps {
 
 export interface SerializedTextProps
   extends SerializedObjectProps,
-  UniqueTextProps { }
+    UniqueTextProps {}
 
-export interface TextProps extends FabricObjectProps, UniqueTextProps { }
+export interface TextProps extends FabricObjectProps, UniqueTextProps {}
 
 /**
  * Text class
  * @tutorial {@link http://fabricjs.com/fabric-intro-part-2#text}
  */
 export class Text<
-  Props extends TProps<TextProps> = Partial<TextProps>,
-  SProps extends SerializedTextProps = SerializedTextProps,
-  EventSpec extends ObjectEvents = ObjectEvents
->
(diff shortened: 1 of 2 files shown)


src/shapes/Text/Text.ts
```diff
@@ -1640,23 +1638,23 @@ export class Text<
...
-      family.indexOf(',') > -1 ||
-      family.indexOf('"') > -1 ||
-      fontIsGeneric
-        ? style.fontFamily
+        family.indexOf("'") > -1 ||
+        family.indexOf(",") > -1 ||
+        family.indexOf('"') > -1 ||
+        fontIsGeneric
+        ? this.fontFamily
         : `"${style.fontFamily}"`;
```
ShaMan123, line 1651: I thought the point was to use `family`

ShaMan123 (changes requested):
Please run `npm run prettier:write` and commit 
After that the diff will be meaningful

The fact that tests pass is meaningless because they are passing without this PR, meaning this is not covered by a test so you should add a test that is failing on master and passing here

src/shapes/Text/Text.ts
VardaQuraishi: Thank you @ShaMan123 for guiding me through.

The highlighted line (const family = style.fontFamily || this.fontFamily;) checks if the style object has a valid fontFamily property. If it does, family will be assigned the value of style.fontFamily. Otherwise, it will fall back to the this.fontFamily property.

I have run `npm run prettier:write`. Are there any other commands or tests that needs to be run?

Could you help against this 
image

VardaQuraishi: I have also added a unit test in the text/unit file and all tests are passing

asturur:
Please check the request for the git blame verification in the original issue.

ShaMan123 (changes requested):
You also need to refine this pr

PR #9043, opened 2023-06-21
BREAKING(IText): remove `_editingProps`

ShaMan123:
## Motivation

#8995 made me want to do the same and have the drag action not trigger when editing without the need for `lockMovementX/Y` 
Then once I saw it was simple I decided to remove the hack of `_editingProps`, which was long desired.
I consider it a hack because it changes the public state/props of an object, meaning that if a dev changes one of these props while editing without using set that will be discarded after exit editing. 
It is very annoying for a dev to waste a few hours on this, and of course it will require them to dig deep into the code and dislike fabric a bit more.
We do not want that.

## Description

This PR removes the hack of `_editingProps` and substitutes it with simple method overriding.
This means that any subclass can override these methods to restore controls, borderColor etc.
It is BREAKING for whoever had to override this hack, but I am sure everyone will appreciate it.
Also it removes the change done to the canvas cursor.
I do not understand why it was there to begin with.
I expect the cursor to be text only if I am hovering the text object while editing.
Current behavior is that the cursor is text on the entire canvas while editing.
My preferences don't really matter rather, fabric should not decide this for the dev.
They can simply:
```ts
    canvas.on('text:editing:entered', () => {
      canvas.setCursor('text')
    })
    canvas.on('text:editing:exited', () => {
      canvas.setCursor('pointer');
    });
```

## Changes

- removed `_editingProps`, `_setEditingProps`,


ShaMan123 (commented):
- exposed `Object#getCursor`,
- exposed `Canvas#getTransformAction` for overriding, replacing `getActionFromCorner`
- removed `_savedProps`, `_saveEditingProps`, `_setEditingProps`, `_restoreEditingProps`

ShaMan123:
I should add a playwright test for this
Can you instruct me how to use it with the cool test generator?

ShaMan123 (commented):
DONE apart from playwright tests

src/shapes/IText/ITextBehavior.ts
```diff
@@ -442,21 +471,6 @@ export abstract class ITextBehavior<
...
   }
 
-  /**
-   * @private
-   */
-  _setEditingProps() {
-    this.hoverCursor = 'text';
-
-    if (this.canvas) {
-      this.canvas.defaultCursor = this.canvas.moveCursor = 'text';
```
ShaMan123, line 452: handled by getCursor

src/shapes/IText/ITextBehavior.ts
```diff
@@ -442,21 +471,6 @@ export abstract class ITextBehavior<
...
-    }
-
-    this.borderColor = this.editingBorderColor;
-    this.hasControls = this.selectable = false;
```
ShaMan123, line 456: handled by _rednerControls

src/canvas/SelectableCanvas.ts
```diff
@@ -796,6 +796,38 @@ export class SelectableCanvas<
...
+  } {
+    if (corner && isSelected) {
+      const control = target.controls[corner];
+      return {
+        action: control.getActionName(e, control, target),
+        actionHandler: control.getActionHandler(e, target, control),
+      };
+    } else if ((target as IText).isEditing) {
+      return {
+        action: 'text',
```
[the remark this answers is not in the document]
ShaMan123: added this action so that the default darg action is not kicked in

src/shapes/IText/ITextBehavior.ts
```diff
@@ -442,21 +471,6 @@ export abstract class ITextBehavior<
...
-    this.lockMovementX = this.lockMovementY = true;
```
ShaMan123, line 457: handled by the action name or handler

ShaMan123 (commented):
explanations

PR #9045, opened 2023-06-23
Fix main export

jiayihu:
According to the [Node.js documentation](https://nodejs.org/api/packages.html#package-entry-points), `main` is the legacy property supported by all Node.js whereas `exports` is supported only by 12.x

What I've noticed with Create-React-App v5.0.1 (latest CRA btw) is that it internally uses Jest v27. Unfortunately Jest added [full support for `exports` field only in Jest 28](https://jestjs.io/blog/2022/04/25/jest-28#packagejson-exports) and we cannot update CRA to use Jest 28 because it's an internal dependency. This means that Jest 27 and CRA is currently using the `main` field, which however is exporting the node version.

This results in fabric defining its own canvas mock using jsdom, overriding the one globally defined in the project. @ShaMan123 fixed this temporarily by writing `fabric.setEnv({ ...fabric.getEnv(), window, document })` so that it overrides again the fabric jsdom mock with the global one, but even he didn't fully understand the root cause 😄 So it could be extremely confusing for other people. I've instead tried that updating the `main` field indeed solves the issue.

I believe that the `main` field should export the browser version because it's the legacy property supported by any Node version and tooling in the community. So it's safer to keep it referencing the `dist/index.js` as previous versions of fabric did, then modern tools can read `exports` and pick the best export.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [next]
 
+- chore(): fix main export [#9045](https://github.com/fabricjs/fabric.js/pull/9045)
+
 ## [6.0.0-beta10]
 
 - chore(TS): Remove @ts-nocheck from Text class. [#9018](https://github.com/fabricjs/fabric.js/pull/9018)
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "node": ">=14.0.0"
   },
   "module": "./dist/index.mjs",
-  "main": "./dist/index.node.cjs",
+  "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
   "typesVersions": {
     ">=4.2": {


asturur:
Why not ejecting the CRA and updating the dependencies?
This is a runtime change for a test issue, and the reason why we created setEnv is exactly to support all of those needs to configure a slightly different environment.

jiayihu:
> Why not ejecting the CRA and updating the dependencies?

Ejecting CRA is hardly recommended as you can't update to future versions anymore and you are in charge of maintaining the dependencies at that point.

> This is a runtime change for a test issue, and the reason why we created setEnv is exactly to support all of those needs to configure a slightly different environment.

Yes there is the workaround, but why would you want to change the legacy `main` export to point to the Node.js build to begin with? This change in fabric v6 could cause issues with other tools as well, like browserify. The way I see it, `main` should not point to the Node build at all.

ShaMan123:
I understood this exactly, which is why I PRed (not in front of a laptop, so no ref) to allow setting env as @asturur stated.

I understand your point but disagree with the propsed solution.

I think we should block it completely, not allowing any ambiguity in env/importing, if possible. 
It is rather confusing. 

If an import error shows I think it is better.
Legacy tools can use dynamic import if no other option can be found to import them in legacy node in such a case.

Before v6 fabric would guess the env, now it doesn't. So it shouldn't guess any longer.

CRA is a dead and horrible tool and hardly mainstream nowadays.
Node14 is considered end of life I think.
And fabric officially supports node14 up to begin with.
Hence I see no reason to allow importing via the old scheme.

ShaMan123:
> why would you want to change the legacy main export to point to the Node.js build to begin with?

Because docs state it is for node env consumption so old apps using this would have used fabric in node.
Which is correct in the jest use case, since it operates on node, but incorrect for you, since you want it to operate as if in a browser env...
Meaning it is on your test config to handle this IMO.

ShaMan123:
We can make things even more straight forward by having a browser/web entry point, equiv to node, and not have a default one

jiayihu:
> We can make things even more straight forward by having a browser/web entry point, equiv to node, and not have a default one

Hmm...I'll give it a try using the `browser` field then

jiayihu:
> Because docs state it is for node env consumption so old apps using this would have used fabric in node.

I don't see any old app that would use fabric as node, since old apps are less isomorphic than new ones definitely.

> Which is correct in the jest use case, since it operates on node, but incorrect for you, since you want it to operate as if in a browser env

It's not that **I** want to operate it as a browser, Jest has always fully supported testing in browser mode. It has been only after several years that they made it opt-in, but it's still a first class behaviour. So I'm not using an edge case I believe.

I've also tried the `browser` field a try but support for it, turns out, was removed in Jest 26 since it was mainly a browserify thing: https://github.com/jestjs/jest/issues/10036

I still believe that it doesn't make sense for `main` to point to the `node.cjs` (so Node + CJS extension) since it's the legacy field, but we can agree to disagree.

PR #9059, opened 2023-07-04
Added patternTransform to Pattern's toSVG method

iocoker:
## Description

The `patternTransform` attribute is missing in Pattern SVG export, even though the Pattern fill has a `patternTransform`
If there's no `patternTransform` on the Pattern, it provides a default identity matrix.

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/patternTransform

### DIFF
--- a/src/Pattern/Pattern.ts
+++ b/src/Pattern/Pattern.ts
@@ -177,10 +177,13 @@ export class Pattern {
       patternHeight =
         repeat === 'repeat-x' || repeat === 'no-repeat'
           ? 1 + Math.abs(patternOffsetY || 0)
-          : ifNaN((patternSource.height as number) / height, 0);
+          : ifNaN((patternSource.height as number) / height, 0),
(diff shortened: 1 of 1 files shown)


ShaMan123 (commented):
This looks good
I don't remember if I handled this in the pattern transform pr which is now stale, I think not.
You should handle fromSVG as well.
Please add snapshot tests in jest and visual tests, in playwright or simply with the existing tooling (there is an svg test file).

asturur:
I have vague memories about this.
There was something fishy with appling the transform in that way.
Do we have a reproduction test?

ShaMan123:
I remember pattern transform became part of canvas API not a long time ago. Could that be it?

iocoker:
@ShaMan123, thanks I will add the tests.

@asturur are you referring to using `patternTransform="scale(1 0.5) rotate(20) skewX(30)"` instead of `patternTransform="matrix(1 0 0 1 0 0)"` ??

PR #9068, opened 2023-07-06
cd(node): ban `package.json` main entry

ShaMan123:
## Description

Blocks the main (legacy) entry by setting it to null
This will remove all ambiguous resolution and force the dev to use `fabric/node`

related to #9045

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@
 - ci(): properly checkout head for stats [#9080](https://github.com/fabricjs/fabric.js/pull/9080)
 - fix(Text): `_getFontDeclaration` wasn't considering fontFamily from the style object [#9082](https://github.com/fabricjs/fabric.js/pull/9082)
 - chore(TS): Fix ITextBehaviour enterEditing type [#9075](https://github.com/fabricjs/fabric.js/pull/9075)
+- cd(node): ban `package.json` main entry [#9068](https://github.com/fabricjs/fabric.js/pull/9068)
 - chore(TS): export FabricObjectProps and GroupProps [#9025](https://github.com/fabricjs/fabric.js/pull/9025)
 - chore(TS): Replace BaseFabricObject with FabricObject [#9016](https://github.com/fabricjs/fabric.js/pull/9016)
 - refactor(svgImport): remove the css/gradient/clipPath global definitions [#9030](https://github.com/fabricjs/fabric.js/pull/9030)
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
ready

asturur:
This is acceptable since we have now 2 entry points, and definetely we want to have people using the correct entry points
Now my point is what happens if you are in jest+jsdom importing browser fabric under node?
That needs to work.

ShaMan123:
I wanted to test that but I am pretty sure using dynamic import will always work

asturur:
What do you mean dynamic import? i want to merge this but i need to be sure i understand what is changing, and with this description i m not sure i get it

asturur:
I also got a bug yesterday with webglprobe and i m trying to understand if it is because of this.

ShaMan123:
```
const fabric = await import('fabric'  or 'path/to/dist')
```

ShaMan123 (commented):
TODO: add import/require tests in node

ShaMan123:
can you point out the docs
Maybe it changed? I remember reading something about a proposal that might have not been accepted

ShaMan123:
And also we should check in MDN how browser modules work

asturur:
Here is the npm doc page i was referring to:
https://docs.npmjs.com/cli/v10/configuring-npm/package-json#main

it has main and browser but doesn't mention the others.

The other one we use for sub-folders style imports i found them here:
https://webpack.js.org/guides/package-exports/

And today searching again i also found this one:
https://nodejs.org/api/packages.html#packages_exports so exports is indeed a node feature

asturur:
i would like to find an example/explanations for main equal null to understand which use cases is solving

asturur:
The node documentation clearly says that exports is an alternative to main.

Also found this PR to the proposal that fix the description because of a bug with the absence of a main entry when using "." export entrypoint.

https://github.com/jkrems/proposal-pkg-exports/pull/38/files

We do use "." instead of "./" and eventually we can change that, but it doesn't seem that is causing an error to us

asturur:
@ShaMan123 my question is still open, why main needs to be null and not removed from the package.json ?
I can't find a list of valid values for the main entry, where did you find null as a good value?

ShaMan123:
I don't rememeber
Maybe it doesn't
I read something and opened the pr and added a test.
Try to remove and see what happens to test

asturur (approved):

PR #9070, opened 2023-07-06
cd(TS): fix internal type importing

ShaMan123:
## Motivation

closes #9051 

## Changes

Exposes `dist/src` types to the consumer

## In Action

image

https://github.com/fabricjs/fabric.js/assets/34343793/19f09622-8fb4-445c-baae-f963ffe46712

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- cd(TS): fix internal type importing [#9070](https://github.com/fabricjs/fabric.js/pull/9070)
 - chore(TS): export FabricObjectProps and GroupProps [#9025](https://github.com/fabricjs/fabric.js/pull/9025)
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
The best solution

jiayihu:
It works but I think it's actually the worst solution: the Language Service Protocol suggests this path as auto-import but it doesn't actually work

Screenshot 2023-07-07 at 09 49 40

You would need to debug the types resolution to know to change it to `fabric/types/shapes/Text/Text`. Also, it doesn't avoid the consumer to use the source path of the type. So it's the worst combination of the drawbacks we discussed in #9051:
- As consumer, you need to know the types resolution configuration from fabric and the internal file path
- As fabric maintainer, you will break the type if you move the file or the type to another file

ShaMan123:
Ahhvery bad indeed
I thought it resolved properly

ShaMan123:
So we export as much types as possible

PR #9075, opened 2023-07-07
fix(TS): ITextBehaviour enterEditing type

jiayihu:
Very very minor change. The event was type correctly: `'editing:entered': never | { e: TPointerEvent }`, but the method required `e` to be defined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- chore(TS): Fix ITextBehaviour enterEditing type [#9075](https://github.com/fabricjs/fabric.js/pull/9075)
 - chore(TS): export FabricObjectProps and GroupProps [#9025](https://github.com/fabricjs/fabric.js/pull/9025)
 - chore(TS): Replace BaseFabricObject with FabricObject [#9016](https://github.com/fabricjs/fabric.js/pull/9016)
 - refactor(svgImport): remove the css/gradient/clipPath global definitions [#9030](https://github.com/fabricjs/fabric.js/pull/9030)
--- a/src/shapes/IText/ITextBehavior.ts
+++ b/src/shapes/IText/ITextBehavior.ts
@@ -374,7 +374,7 @@ export abstract class ITextBehavior<
   /**
    * Enters editing state
    */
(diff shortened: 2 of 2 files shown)


ShaMan123 (approved):
Makes sense
Needed when calling the method imperatively

ShaMan123:
We cab still pass the same object to the event because the fire method passes an empty object anyways

jiayihu:
> We cab still pass the same object to the event because the fire method passes an empty object anyways

The type would be different, the passed object would be of type `{ e: TPointerEvent | undefined }`, whereas the event has type `never | { e: TPointerEvent }`

ShaMan123:
It doesn't matter
https://github.com/fabricjs/fabric.js/blob/2a484bf7a911e72e4dc70b9e866c168688063540/src/Observable.ts#L177

jiayihu:
`{ e: undefined }` is not the same as `undefined`. Anyway I'm talking about the type

ShaMan123 (approved):

ShaMan123:
I know
I wrote the type
`fire` passes an object anyways so it can be `{ e?: TPointerEvent }` instead of `never | { e: TPointerEvent }`

ShaMan123:
I am meticulous only because I was instructed to keep type PR pure
I know it can be annoying

jiayihu:
> fire passes an object anyways so it can be { e?: TPointerEvent } instead of never | { e: TPointerEvent }

Aahh so you were suggesting to change the event type to be `{ e?: TPointerEvent }`? In that case, you need to be clear in your comments, most of the time I can hardly understand what you mean you exactly 😄

> I am meticulous only because I was instructed to keep type PRs pure

For instance, here by "pure" you mean keeping the PR type-only without runtime JS changes? I can't every time have to guess what's in your mind. For me "pure" has a different meaning, typically associated with "pure functions", i.e. without side effects.

ShaMan123:
I looked again at the code
There are other instances of this and they use an optional event so this should as well
https://github.com/fabricjs/fabric.js/blob/d0d98dbf82a6ce2d9d16b55f95812df9b236331f/src/EventTypeDefs.ts#L180-L193

PR #9076, opened 2023-07-08
fix(): path iterator M cmd

ShaMan123:
## Description

Was working on path animation and found a bug.
The M cmd was using stale values.
Added a test that fails on master.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): path iterator M cmd [#9076](https://github.com/fabricjs/fabric.js/pull/9076)
 - chore(TS): export FabricObjectProps and GroupProps [#9025](https://github.com/fabricjs/fabric.js/pull/9025)
 - chore(TS): Replace BaseFabricObject with FabricObject [#9016](https://github.com/fabricjs/fabric.js/pull/9016)
 - refactor(svgImport): remove the css/gradient/clipPath global definitions [#9030](https://github.com/fabricjs/fabric.js/pull/9030)
--- a/src/util/path/__snapshots__/index.spec.ts.snap
+++ b/src/util/path/__snapshots__/index.spec.ts.snap
@@ -1,0 +1,55 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Path Utils getPathSegmentsInfo operates as expected 1`] = `
+[
+  {
+    "command": "M",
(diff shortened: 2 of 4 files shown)


ShaMan123:
question is if other cmds were broken

ShaMan123:
Also it seems better not to have `tempInfo` and instead just push the data in each case

ShaMan123 (commented):
seems that text path rendering worked around this bug w/o knowing it was a bug

asturur:
> Looking at history and blame seems #6525 broke the M cmd

Can you explain this? Seems to me #6525 added the code to measure the path, so how could it break it if it introduced it in that moment?

asturur:
oh i see the pathInfo was not populated of a point for the case of M.

src/shapes/Text/Text.ts
```diff
@@ -941,18 +937,14 @@ export class Text<
...
+  _setGraphemeOnPath(positionInPath: number, graphemeInfo: GraphemeBBox) {
     const centerPosition = positionInPath + graphemeInfo.kernedWidth / 2,
       path = this.path!;
 
     // we are at currentPositionOnPath. we want to know what point on the path is.
     const info = getPointOnPath(path.path, centerPosition, path.segmentsInfo)!;
-    graphemeInfo.renderLeft = info.x - startingPoint.x;
-    graphemeInfo.renderTop = info.y - startingPoint.y;
+    graphemeInfo.renderLeft = info.x - path.pathOffset.x;
+    graphemeInfo.renderTop = info.y - path.pathOffset.y;
```
asturur, line 947: this change is not 1:1.
before we were subtracting the first path point all the time, while now we are just substracting the offset?

ShaMan123: because of the bug startingPoint was 0,0 if I am not mistaken (at least for paths starting with M cmd)

asturur: shouldn't it have been undefined/undefined?

asturur: ok i couldn't read this: `tempInfo = <TPathSegmentInfoCommon<'M'>>basicInfo;`

asturur: i need a color scheme where the TS syntax is rgba (255, 255, 255, 0.02) fontSize: 1

ShaMan123: Explain

PR #9077, opened 2023-07-08
feat(): path animation

ShaMan123:
## Motivation

#9071 
I use this kind of animation in my apps and this is 4x cleaner

## Description

```
  fabric.util.animate({
    path,
    duration: 3000,
    endValue: '85%',
    easing: fabric.util.ease.easeInOutBack,
    onChange({ x, y }, v, t) {
      circle.setXY(new fabric.Point(x, y), 'center', 'center');
      canvas.renderAll();
    },
  });
```

## Changes

Apart from `PathAnimation`:
- refactored animation types:
  - moved XXXAnimationOptions to its file
  - removed TAnimation, AnimationOptions in favor of maps => AnimationRegistry, AimationOptionsRegistry

## In Action

https://codesandbox.io/p/sandbox/fabric-nextjs-sandbox-forked-4vqpvd?file=%2Fpages%2Findex.tsx%3A52%2C5-77%2C6

https://github.com/fabricjs/fabric.js/assets/34343793/44fbe52b-627e-4e4d-bbff-75806ef3cbb5


src/parser/percent.ts
```diff
@@ -3,8 +3,8 @@ import { capValue } from '../util/misc/capValue';
 
 const RE_PERCENT = /^(\d+\.\d+)%|(\d+)%$/;
 
-export function isPercent(value: string | null) {
-  return value && RE_PERCENT.test(value);
```
ShaMan123, line 7: checked that all usages are with a defined value

src/util/animation/types.ts
```diff
@@ -117,20 +115,3 @@ export type TAnimationOptions<T, TCallback = T, TEasing = T> = Partial<
...
-  string,
-  number[]
->;
-
-export type AnimationOptions<T extends number | number[] | TColorArg> =
-  T extends TColorArg
-    ? ColorAnimationOptions
-    : T extends number[]
-    ? ArrayAnimationOptions
-    : ValueAnimationOptions;
```
ShaMan123, line 136: moved to the files

ShaMan123 (commented):
updated desc

ShaMan123:
I want to write test with playwright but not sure of the best way to do so

ShaMan123 (commented):
ready

.github/workflows/build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
               size: { fabric: { minified: fs.statSync('${{ env.minified }}').size, bundled: fs.statSync('${{ env.bundled }}').size } }
             });
       - name: checkout src files
-        run: git checkout origin/master -- src fabric.ts index.ts index.node.ts .browserslistrc rollup.config.mjs
+        run: git checkout ${{ github.event.pull_request.base.sha }}
```
ShaMan123, line 45: fixed - didn't checkout added files for some reason

ShaMan123:
I also want animateColor to be removed and intergrated into `animate` - for that to happen we can make start/endValue a Color instance so we can check that in the condition in `animate`

src/parser/percent.ts
asturur: can we put `const RE_PERCENT = /^(\d+\.\d+)%|(\d+)%$/;` inside the function `isPercent` ? i m not sure there is reason to keep a regex alive all the time. The parser is not hot code

ShaMan123: I thought you wanted all regex to be const to be created once.
Can we not hold back on these things for now?

src/util/animation/PathAnimation.ts
```diff
@@ -0,0 +1,72 @@
...
+      >
+    | undefined,
+  context: ReturnType<typeof parsePathForIteration>
+) =>
+  callback &&
+  ((pathProgress: number, valueProgress: number, durationProgress: number) => {
+    const distance = pathProgress * context.length;
+    return callback(
+      {
+        ...context.getPointOnPath(distance),
```
ShaMan123, line 30: maybe we need to respect a transform matrix line under group?

asturur:
let me  see if i can unblock this today

asturur:
ok for me here there is a real divergent idea on how features needs to be exposed.
We will need to sync on this.

The smartest part of this pr is the callback wrapper that changes a value animation to a point on path context.
All the rest to me seems unnecessary construct to hide everything as an option of a generic utility.

This is exactly what i see as feature creep.

To explain myself at best, something like this:

```js
  fabric.util.animate({
    duration: 3000,
    endValue: '85%',
    easing: fabric.util.ease.easeInOutBack,
    // path or pathContext as first argument depending if we think parsing the path is a cost needs to be avoided
    onChange: positionOnPathCallbackWrapper(path, ({ x, y }, v, t) {
      circle.setXY(new fabric.Point(x, y), 'center', 'center');
      canvas.renderAll();
    }),
  });
```

would be equally clean and require far less code to maintain and doesn't need to be part of the bundle either.


On top of that a Path animation for me would be taking path A, path B and move smoothly between path A and path B and not animating a point over a path.

All the other animation types do exactly that, change a color from a to b, a number from a to b and so on.

ShaMan123:
I tried to align to existing patterns.
Your point is valid but makes color and array animations as unwanted as the point on path animation.
I can rename it to PointOnPathAnimation
Animating a path is not something I used, I did not experience this UX
The real problem in the code is the the path iterator has many weird stuff in it, the worst is it pushes as the last object a sum or whatever which is junk if you want to iterate over it and that the angle values of the first and therefore the last entries are wrong.
Fixing that and making the path iterator more freindly will remove this need entirely I believe.
But it is breaking so if you agree I will reak and rename

asturur:
I was not advocating for a path animation utility indeed.
I also don't like array animation, correct. Instead color is a bit of different because color is a first level property on many objects, so if you support object animation by properties, at some point someone needed the color and added it. It could be espressed as a callback wrapper too, but would make not possible to animate objects with group of properties.
Color is also a value and is not exactly clear how to animate it. Was also added many years ago when i had way less comprehensive view of the library.

The path iterator was born only for supporting text on a path, so is probable that may not support other use cases.
The initial angle of the first point is probably broken because text is never on the initial point since no glyph as a zero width, and if it has is invisible, so no one see the wrong angle, and so no one checked for it, but the solution is probably very simple for that.

I m not sure what you mean by `weird stuff and junk`. I really don't like to discuss code on that terms because i have to guess what are you referring to instead of you telling me what is that you don't like.

Here there is a callback wrapper that imho can be useful and light weight if couple with an example on how to use it.

If the path iterator can be done better, as long as it supports text on a path as it is today, we have no issue changing it, is not part of the public api per se, it was there because there was no distinction between public and private in fabric 5 because of how the code was written.

Then apart the callback wrapper, and the improvement on the path iterator, path iteration is not a fabricJS feature.
If it can be, point out the part that in your opinion need fixing before going head down fixing it.

ShaMan123:
I don't think the color animation is useful because it allow only rgb animation, where as hsl animation is what I am used to do.
I can open another PR targeting the path iterator.
It is a very good piece of code I would never have access to anywhere else and allows outstanding UX/UI but is extremely hard to use.
I will show you how I use it in my projects so you understand my intention.
I want it to become a class with methods exposing 1-2 methods and hiding all the iterator context as protected properties

```ts

export const parsePathForIteration = (
  path: string | fabric.TSimplePathData
) => {
  const pathData = fabric.util.makePathSimpler(
    Array.isArray(path) ? path : fabric.util.parsePath(path)
  );
  const data = fabric.util.getPathSegmentsInfo(pathData);
  return {
    path: pathData,
    data,
    // this is because in the code the last entry is a sum 
    length: data[data.length - 1]!.length,
    // unintuitve to understand this
    getPointOnPath: (distance: number) =>
      fabric.util.getPointOnPath(pathData, distance, data)!,
  };
};

```

ShaMan123:
so fixing the angles and removing the last entry + refactoring it to expose `getPointOnPath` etc.

asturur:
Please i don't want to add any more classes.

Can we just add an iterate over to fabric.Path so that the intuitive part of:

```ts
    getPointOnPath: (distance: number) =>
      fabric.util.getPointOnPath(pathData, distance, data)!,
```

can be saved from the state of the path?

```ts
getPointOnPath: (distance: number) => {
  return fabric.util.getPointOnPath(this.path, distance, this.pathSegmentInfo)!,
}
```

so instead of creating a path iterator new class you just create a fabric path.

ShaMan123:
I like that

ShaMan123:
I don't understand the transform part

asturur:
we can talk through it, the transform part is optional.

Like i have a path long 10, how long it is when my path has scaleX by 2? it could be 20, unless the path is a vertical line so is still 10.

this is nice for a relaxing chat maybe

ShaMan123:
I see, I would transform the points and that is it

PR #9078, opened 2023-07-08
test(playwright): add snapshots, refactor utils, coverage

ShaMan123:
## Motivation

playing around with playwright cause I wanted to write another test

## Description

Made the exisitng test create visual snapshots
Refactored util to be more useful
Added coverage reporting
Added an app concept to each test

## Changes

Each test is a folder containing:
- `*.spec.ts` - playwright test
- `index.ts` - fabric test setup:
   - `beforeAll`
      - a simple hook that creates the canvas, attaches and runs `index.ts`
      - return a map of objects that playwright can access via eval
      - it is awaited by playwright to avoid race conditions
   - `before`: same as `beforeAll` only that is accepts a selector to select the canvas element to operate on
   - `afterAll` 
      - executed by playwright at the end of a test
      - also awaited
   - `after`: same as `afterAll` only that is accepts a selector to select the canvas element to operate on
- `index.html`
  - used to override the default body (e.g. a test with multiple canvases, offset etc.)


ShaMan123 (commented):
extracted logic from util/objects to ObjectUtil

ShaMan123:
I really want to use jest inside playwright for object snapshots, is there a way to do that?

.github/workflows/tests.yml
```diff
@@ -109,6 +107,19 @@ jobs:
...
+    name: Playwright tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/cached-install
+      - name: Build fabric.js
+        uses: ./.github/actions/build-fabric-cached
+      - name: Run playwright tests
+        run: |
+          npx playwright install
```
asturur, line 121: doesn't npm ci install this too?

e2e/utils/TestUtil.ts
```diff
@@ -0,0 +1,29 @@
+import { Page } from '@playwright/test';
+import { ObjectUtil } from './ObjectUtil';
+
+let ID = 0;
+const uuid = () => `test${ID++}`;
+
+export class TestUtil {
+  constructor(readonly page: Page) {}
```
asturur, line 8: What is that allow us to not assign page explicitly to the constructor?

.github/workflows/tests.yml
ShaMan123: seems not... at least before #9086 
I didn't check after

ShaMan123: I think it is needed for browsers etc

e2e/utils/TestUtil.ts
ShaMan123: Yes, it is an abbreviation

.github/workflows/tests.yml
ShaMan123: https://playwright.dev/docs/ci-intro

ShaMan123:
@asturur I have installed the playwright VSCODE extension. Looks good.

ShaMan123:
Also the --ui option is good, instead of the extension
But still I don't understand how to use codegen

ShaMan123:
@luciemars ?

ShaMan123:
https://medium.com/@novyludek/code-coverage-of-e2e-tests-with-playwright-6f8b4c0b56e1

https://github.com/playwright-community/jest-playwright??
Allows us to use playwright within jest... they say however
image

ShaMan123:
BEFORE
image

e2e/site/index.html
```diff
@@ -5,9 +5,10 @@
       body {
         padding: 0;
         margin: 0;
+        background-color: black;
       }
-      .canvas-container {
-        border: 1px solid black;
```
ShaMan123, line 10: was causing snapshot size mismatch

ShaMan123 (commented):
READY!
Playwright is caching so this is optimized!

e2e/tests/text/text-editing.spec.ts
```diff
@@ -1,53 +1,89 @@
...
   // click the object to select it
-  await clickCanvas(page, {
+  await util.clickCanvas({
     position: textCenter,
   });
 
-  // drag the control
-  await page.mouse.move(controlPoint.x, controlPoint.y);
+  // drag the mr control
+  await page.mouse.move(mrControlPoint.x, mrControlPoint.y);
```
jiayihu, line 54: In our project we create in Cypress a utility to drag from-to in the canvas because that's the most common interaction almost

e2e/tests/text/text-editing.spec.ts-snapshots/decrease-width-ml.png
jiayihu, line 1: I recomment also creating a utility that snapshots the canvas extractly with the bounding box of the fabric Object, although I'm not sure how much that's easy. That's because having useless pixels makes the diff less reliable, at least in Cypress. 
Also in this image, not the whole object is screenshotted and some pixels are cut. That's why using a static screen size doesn't work really well with the canvas.

e2e/utils/TestUtil.ts
```diff
@@ -0,0 +1,35 @@
...
+    return this.page.click('canvas.upper-canvas', clickProperties);
+  }
+
+  screenshot(options: LocatorScreenshotOptions = {}) {
+    return this.page
+      .locator('[data-fabric="wrapper"]')
+      .screenshot({ omitBackground: true, ...options });
+  }
+
+  async addTextbox(text: string, properties) {
```
jiayihu, line 20: Since you can have many types of objects it's more flexible to have `add(() => new Textbox())` as API, i.e. leaving the caller the decision about what to create as object.

ShaMan123: That can be done using eval since it must run in the browser context

ShaMan123: This is still a POC

e2e/tests/text/text-editing.spec.ts-snapshots/decrease-width-ml.png
ShaMan123: Yes, we experience it as well in fabric tests, I wonder how playwright is in this term
Once #8767 is merged I will feel confident doing something like you have suggested

.github/workflows/tests.yml
```diff
@@ -109,8 +107,40 @@ jobs:
...
+      - name: Cache Playwright Browsers
+        id: cache-browsers
+        uses: actions/cache@v3
+        with:
+          path: /home/runner/.cache/ms-playwright
+          key: playwright-${{ hashFiles('package-lock.json') }}
+          restore-keys: playwright
+      - name: Install Playwright Browsers
+        if: steps.cache-browsers.outputs.cache-hit != 'true'
+        run: npx playwright install --with-deps
```
mxschmitt, line 128:
```suggestion
        run: npx playwright install --with-deps chromium
```

Since you are only testing Chromium, you can do this and it will be probably around 3x faster.

ShaMan123: Thanks man

e2e/site/index.html
```diff
@@ -1,25 +1,19 @@
 <html>
   <head>
-    <script type="text/javascript" src="/dist/index.js"></script>
+    <title>Fabric e2e Testing Suite</title>
+    <link rel="icon" type="image/x-icon" href="./favicon.ico" />
```
ShaMan123, line 4: to silence playwright raising an error

ShaMan123 (commented):
### Concept

Each test is a folder containing:
- `*.spec.ts` - playwright test
- `index.ts` - fabric test setup:
   - `beforeAll`
      - a simple hook that creates the canvas, attaches and runs `index.ts`
      - return a map of objects that playwright can access via eval
      - it is awaited by playwright to avoid race conditions
   - `afterAll` 
      - executed by playwright at the end of a test
      - also awaited

ShaMan123 (commented):
I am DONE

ShaMan123 (commented):
For the last time
DONE
Not touching, I sware

Playwright is a powerful tool indeed, and this PR makes it easy for us to use it and to migrate tests to it.
If it had data snapshots that would have been amazing, testing state during interaction

e2e/utils/TestUtil.ts
asturur: can you link it to me? i want to learn what is it.

.github/workflows/tests.yml
```diff
@@ -109,8 +107,33 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/cached-install
+      - name: Build fabric.js
+        uses: ./.github/actions/build-fabric-cached
+        # Playwright suggests against caching the browser install
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps chromium
+      - name: Run Playwright tests
+        run: xvfb-run npm run test:e2e
```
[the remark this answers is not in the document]
asturur: For browsers test as unit tests in browsers is not needed.
For playwright i don't know the framework so well yet.
i assume that to take screenshots you need xvfb ready unless you are doing them through the browser with an internal browser buffer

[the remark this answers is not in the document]
ShaMan123: If you look before this commit you will see I didn't use the headed option thus nkt needing xvfb
But I don't know what the diff is, tests pass in both cases

asturur (approved):

asturur:
Let's try this setup.
There is a lot ot absorb tho.

.github/workflows/tests.yml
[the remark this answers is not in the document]
asturur: Let's keep what works, at some point we will optimize instance usage and strain, for now starting the X server on linux who cares

e2e/utils/TestUtil.ts
ShaMan123: https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd

PR #9080, opened 2023-07-09
ci(): properly checkout head for stats

ShaMan123:
## Description

didn't checkout files in a PR for some reason

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,7 +42,7 @@ jobs:
               size: { fabric: { minified: fs.statSync('${{ env.minified }}').size, bundled: fs.statSync('${{ env.bundled }}').size } }
             });
       - name: checkout src files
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -42,7 +42,7 @@ jobs:
               size: { fabric: { minified: fs.statSync('${{ env.minified }}').size, bundled: fs.statSync('${{ env.bundled }}').size } }
             });
       - name: checkout src files
-        run: git checkout origin/master -- src fabric.ts index.ts index.node.ts .browserslistrc rollup.config.mjs
+        run: git checkout ${{ github.event.pull_request.base.sha }}
```
asturur, line 45: not sure why we were checking out only some files. Probably because before dist was in the git history

asturur (approved):

PR #9081, opened 2023-07-09
fix(text): styles

### LINKED ISSUE #9028: [BUG]: text style onInput
@asturur please help if you can

**It seems that when text is wrapped styles is not updated correctly from onInput**

I have to say, I do understand why styles was chosen to be a map but I don't think that was a good choice.
So much easier to maintain and update it as a flat array corresponding to char offset.
Then wrapping is not related and syncing char changes from `onInput` to styles or where ever is as simple as splicing 
I do think that fixing this bug or making it an array will cost the same amount of time.
I understand it is breaking so you will probably not want this now.
The logic for me is unapproachable.

Suspected `insertNewlineStyleObject` and `removeStyleFromTo`:

https://github.com/fabricjs/fabric.js/assets/34343793/35b086e7-7ae3-4858-b471-01a4bb91b890

https://github.com/fabricjs/fabric.js/assets/34343793/a66cc38a-b33e-4a5f-9b51-18d6c70abb5a

ShaMan123:
## Motivation

closes #9028

## Description

Extracts style logic to a manager
Styles has been refactored into an array mirroring the chars to make our lives easier.
In addition I have exposed a method `getDiffFromInputEvent` to make our lives easier.
I have left the interface untouched so not to break, though I think we should remove all that bad interface.

## Gist

TODO:
- [ ] test


asturur:
@ShaMan123 you are welcome to fix the bug, but not accepting those kind of refactors right now.
Did you find the bug?

ShaMan123:
The mess is unbearable

asturur:
Let's keep the conversation to the PR, since the related issue doesn't really say much else apart showing to video to be interpreted.

ShaMan123:
The bug is to do with bad design
Somewhere style is used with unwrapped text, and in another place with the wrapped value
I will not dig into the style logic - did you see the amount of code I removed just by moving to a simple array corresponding to the chars???? That fixed the bug

ShaMan123:
text is amess @asturur 
I don't see how to move forward without completely breaking it
It can remain broken for a long time and then it dies...
Too much work to fix fabric

ShaMan123:
## Why is this bad design?

Styles depend on layout (with no good reason) instead of depending on char offset from start.
Now let's say I have an app and I need to change the default size of textbox.
I can't iterate over the server, changing the width value because it will break all styles.
So I will have to run some ugly code on the client or spawn a node env to simply do `textbox.width=150`.
That makes no sense and is a nightmare.
Considering responsive layouts... I might want to restrict the size of the textbox in some cases and again I am in this pain of breaking styles.
Initializing textbox with a style is also impossible because you need to know the layout beforehand but the style depends on the layout and the layout depends on the style so you can't! 

Styles should not depend on layout.

ShaMan123:
You are right
I tried and got frustrated
It cannot be fixed IMO because of style depending on layout and input changing all that
Backing off

asturur:
Yeah but if you give me reproduction steps i can try.
I don't see what you do in the videos.

ShaMan123:
Start the vanilla app
Click on the bold text (I clicked before the .)
Hit enter a few times
Hit delete a few times
Styles shift badly

PR #9082, opened 2023-07-09
fix(Text): `_getFontDeclaration`

### LINKED ISSUE #9010: Text style bug?

ShaMan123:
## Motivation
closes #9010
closes #9041

## Description

The function **_getFontDeclaration** was suppose to take the font information from either the style object or the instance itself, but for fontFamily there is a bug in which the fontFamily is only taken from the instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
-- fix(): `_getFontDeclaration` [#9082](https://github.com/fabricjs/fabric.js/pull/9082)
+- fix(Text): `_getFontDeclaration` [#9082](https://github.com/fabricjs/fabric.js/pull/9082)
 - chore(TS): Fix ITextBehaviour enterEditing type [#9075](https://github.com/fabricjs/fabric.js/pull/9075)
 - chore(TS): export FabricObjectProps and GroupProps [#9025](https://github.com/fabricjs/fabric.js/pull/9025)
 - chore(TS): Replace BaseFabricObject with FabricObject [#9016](https://github.com/fabricjs/fabric.js/pull/9016)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): `_getFontDeclaration` [#9082](https://github.com/fabricjs/fabric.js/pull/9082)
 - chore(TS): Fix ITextBehaviour enterEditing type [#9075](https://github.com/fabricjs/fabric.js/pull/9075)
 - chore(TS): export FabricObjectProps and GroupProps [#9025](https://github.com/fabricjs/fabric.js/pull/9025)
 - chore(TS): Replace BaseFabricObject with FabricObject [#9016](https://github.com/fabricjs/fabric.js/pull/9016)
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
fixed

ShaMan123 (commented):
done

asturur (approved):

src/shapes/Text/Text.ts
```diff
@@ -1640,25 +1641,31 @@ export class Text<
...
-        ? style.fontFamily
-        : `"${style.fontFamily}"`;
+    const parsedFontFamily =
+      fontFamily.includes("'") ||
+      fontFamily.includes('"') ||
+      fontFamily.includes(',') ||
+      Text.genericFonts.includes(fontFamily.toLowerCase())
+        ? fontFamily
+        : `"${fontFamily}"`;
     return [
```
jiayihu, line 1664: Is this a hot path method? (e.g. called when rendering) In that case you should check if a string interpolation `${fontStyle} ${fontWeight} etc.` is significantly less expensive than creating an array on the fly and doing `join`

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(Text): `_getFontDeclaration` [#9082](https://github.com/fabricjs/fabric.js/pull/9082)
```
jiayihu, line 5: This changelog doesn't tell much about the change. It should be more `_getFontDeclaration: passed style should take precedence` as the test case.

src/shapes/Text/Text.ts
ShaMan123: I thought we should do that. It is called more than once in rendering

CHANGELOG.md
ShaMan123: point noted

asturur: Yeah i also changed the pr description and the merge commit, but forgot to update the changelog.

asturur:
Just for the record:

```js
_getFontDeclaration() {
    let {
      fontFamily = this.fontFamily,
      fontStyle = this.fontStyle,
      fontWeight = this.fontWeight,
      fontSize = this.fontSize
    } = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : {};
    let forMeasuring = arguments.length > 1 ? arguments[1] : undefined;
    const parsedFontFamily = fontFamily.includes("'") || fontFamily.includes('"') || fontFamily.includes(',') || Text.genericFonts.includes(fontFamily.toLowerCase()) ? fontFamily : "\"".concat(fontFamily, "\"");
    return [fontStyle, fontWeight, "".concat(forMeasuring ? this.CACHE_FONT_SIZE : fontSize, "px"), parsedFontFamily].join(' ');
  }
```

This is how the code gets transpiled so far.
template string become concats and the optional argument becomes horrible chain of argumentss keyword.

asturur:
Seems like no browser supports optional arguments correctly, unless you ask babel to support only chrome 100 onward. that seems very strange to me.

PR #9097, opened 2023-07-15
chore(TS): Add more type-checks for remaining files

asturur:
## Motivation

Continue to extend type checks

## Changes

- moved 2 text related svg mixins from object to text.
- added an empty toSVG method because TS


src/shapes/Object/FabricObjectSVGExportMixin.ts
```diff
@@ -65,80 +69,29 @@ export class FabricObjectSVGExportMixin {
...
-  /**
-   * Returns text-decoration property for svg-export
-   * @param {Object} style the object from which to retrieve style properties
-   * @return {String}
-   */
-  getSvgTextDecoration(style) {
-    return ['overline', 'underline', 'line-through']
-      .filter((decoration) => style[decoration.replace('-', '')])
-      .join(' ');
-  }
```
asturur, line 124: moved to the text svg export

src/shapes/Textbox.ts
```diff
@@ -302,22 +334,29 @@ export class Textbox extends IText {
...
-        word = splitByGrapheme ? word : this.graphemeSplit(word);
-        const width = this._measureWord(word, lineIndex, offset);
+        const graphemeArray = splitByGrapheme
+          ? [word]
+          : this.graphemeSplit(word);
+        const width = this._measureWord(graphemeArray, lineIndex, offset);
         largestWordWidth = Math.max(width, largestWordWidth);
         offset += word.length + infix.length;
-        return { word, width };
+        return { word: graphemeArray, width };
```
asturur, line 359: @ShaMan123 we change this function recently.
Word is a string, but then it becomes string or string[] ( because graphemeSplit returns string[];

GraphemeData is defined as wanting string, but definetely here is getting also string[];

I have tried to make this function more straight forward but i m blocked.

ShaMan123: I noticed it when working on it but was afraid to touch it
Your fix looks valid, change `GraphemeData` type to accept both but it can since it does `value[index]` only if I recall correctly

asturur: but it doesn't pass type check :)

ShaMan123 (commented):
https://github.com/fabricjs/fabric.js/pull/9097/files#diff-201e45ac35afd37fa7c0095d9fde230dbc5844ac57a0d4d157a17e6837816496R50

this line fails typing and seems it should because I think it `applyViewboxTransform` return immediatly so no need to call it...
please check it, svg logic is a black box to me

asturur:
i wanted to keep element parser part of its on pr, i ll extract that commit in a new branch

asturur:
i will merge this and work on the 2 remainin big files

PR #9102, opened 2023-07-17
chore(): expose `__corner` => `getActiveControl()`

ShaMan123:
## Motivation

I need to access `__corner` for many cases, sync interaction managers, selection etc.

## Description

expose `getActiveControl()` - naming might be better


ShaMan123 (commented):
naming suggestion are welcome (corener is misleading, a control can be anywhere)
besides that DONE

asturur (approved):

asturur (dismissed):

asturur (approved):

jiayihu:
> !!canvas.getActiveObject().getActiveControl()

Sometimes you can have an active control without an active object, e.g. a hover button that can be clicked without first selecting the object. So ideally it should be `canvas.getActiveControl()` IMO. @ShaMan123

ShaMan123:
Good point


> Sometimes you can have an active control without an active object, e.g. a hover button that can be clicked without first selecting the object. So ideally it should be `canvas.getActiveControl()` IMO

PR #9108, opened 2023-07-19
ci(e2e): support relative imports

ShaMan123:
## Motivation

We want to use `index.ts` as a regular module so relative imports are a must

## Description

Using a babel plugin to transform relative imports to point to `e2e/dist` folder in needed


ShaMan123 (commented):
The test report (when failing) uploads only the `index.html` which is useless - could not configure it to succeed

ShaMan123 (commented):
added matchers

asturur:
The issue i m about to have with those pr for e2e is that in this way it became super complex.
The initial implementation was simpler imho.

asturur (approved):

PR #9110, opened 2023-07-19
test(): stroke projection

ShaMan123:
## Motivation

testing #8374 

## Description

First merge #9108 

@luizzappa you rock!
I have written super strict tests and all of them pass except one test that group seems to be the blame for.
GREAT JOB!
Thanks again

## Changes

I have made a strict test the samples the borders around each object in the test!


src/util/internals/promiseSequence.ts
```diff
@@ -0,0 +1,2 @@
+export const promiseSequence = <T>(tasks: (() => Promise<T>)[]) =>
+  tasks.reduce((prev, curr) => prev.finally(curr), Promise.resolve());
```
asturur, line 2: this doesn't belong here.

test/visual/stroke_projection.js
```diff
@@ -1,307 +0,0 @@
...
-            fabricClass: 'Canvas',
-            disabled: fabric.isLikelyNode,
-          });
-        });
-      });
-    });
-  });
-
-  tests.forEach(visualTestLoop(QUnit));
-});
```
asturur, line 307: i m not 100% sure those tests should be migrated. If we check visual rendering for visual parity, the old visual test suite is the one to keep. How else are you going to validate node canvas regressions?

ShaMan123: All these tests were disabled for node canvas

test/visual/stroke_projection.js
```diff
@@ -1,307 +0,0 @@
...
-            golden: `stroke-projection/${
-              isPolygon ? 'strokeLineJoin' : 'strokeLineCap'
-            }/${testSinglePoint.name}-${strokeLineCap}${
-              strokeUniform ? '-uniform-' : ''
-            }-${skewX}skewX-${skewY}skewY.png`,
-            percentage: 0.001,
-            width: 600,
-            height: 900,
-            fabricClass: 'Canvas',
-            disabled: fabric.isLikelyNode,
```
ShaMan123, line 299: See

asturur:
i wanted to merge this PR but i don't really understand how those tests works and why are better than previous one.

ShaMan123:
I tests the outer and inner border of the bbox to makes sure the caclualtion is correct.
Outer border should have no colored pixels, inner should have. The border itself is not checked if the object is selected and or due to aliasing

ShaMan123:
I tests the outer and inner border of the bbox to makes sure the caclualtion is correct.
Outer border should have no colored pixels, inner should have. The border itself is not checked if the object is selected and or due to aliasing

asturur:
Can we just have a visual test like before?
We can use the node snapshot pr with explicit `do not match` where they do not match between browser and node.

For pixel accurate checks we should just check bounding box numbers from calculations no?
checking borders of transparent pixels seems like the wrong tool for the job to me

ShaMan123:
I do not agree.
Testing the math with math sounds not good enough or too hard and incomplete.
This is not perfect but it does validate that the bbox is correct. 
Being a test unrelated to impl it is a strong validation.
Regardless we can save bbox snapshots but that does not help us find errors in the impl itself. That was why I wrote it and since the cases are too many to consider.

asturur:
> I do not agree. Testing the math with math sounds not good enough or too hard and incomplete. This is not perfect but it does validate that the bbox is correct. Being a test unrelated to impl it is a strong validation. Regardless we can save bbox snapshots but that does not help us find errors in the impl itself. That was why I wrote it and since the cases are too many to consider.

Maybe you do not agree, but we are testing bounding box by measuring transparent pixels to know a value of a bounding box by a render rather than inspecting the result of the bounding box function.

This is a unit test for me not an e2e, and i really do not understand why you think this is good.

playwright.config.ts
```diff
@@ -28,7 +28,7 @@ const config: PlaywrightTestConfig = {
   /* Retry on CI only */
   retries: process.env.CI ? 1 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
+  // workers: process.env.CI ? 1 : undefined,
```
asturur, line 31: not sold on this regardless of those test implementation.

ShaMan123: What exactly?

asturur:
Going to see if i can slap comments on those tests to overcome the fact i found them unreadable

asturur:
I didn't remember this was the PR that was using visual pixels to detected bounding boxes, and i disagree with that.
Stroke calculation fits ok in unit tests for me and i don't want to run them in e2e tests.
What was the issue with running them as normal unit tests?

PR #9113, opened 2023-07-20
chore(IText): cleanup `_renderCursor`

ShaMan123:
## Motivation

Exposing the cursor measurements it crucial for devs

## Description
 - extracted the mesuring logic to `getCursorRenderData`
- moved the `inCompositionMode` logic to top level where it should be:
   - logically it seems fine since `renderCursor` is called for the same place and `renderCursorAt` is imperative

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- chore(IText): cleanup `_renderCursor` [#9113](https://github.com/fabricjs/fabric.js/pull/9113)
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
pretty clean

asturur:
Can we define crucial / how is used / use cases?
I see this is not breaking probably we can look at it.

is not chore() is feat() anyway.

ShaMan123:
Not crucial
It can wait
I thought it was in between chore/feat so I wanted ti make ut smaller 🫣
I am actually more into the playwright prs, very old group stuff and the eraser that I want in v6...

PR #9119, opened 2023-07-24
chore(TS): complete type check

asturur:

### DIFF
--- a/src/shapes/Text/TextSVGExportMixin.ts
+++ b/src/shapes/Text/TextSVGExportMixin.ts
@@ -293,6 +293,7 @@ export class TextSVGExportMixin extends FabricObjectSVGExportMixin {
    * @return {String}
    */
   getSvgStyles(this: TextSVGExportMixin & Text, skipShadow?: boolean) {
+    // @ts-expect-error TS doesn't respect this type casting
     return `${super.getSvgStyles(skipShadow)} white-space: pre;`;
   }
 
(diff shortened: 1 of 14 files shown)


ShaMan123 (commented):
good enough I think

ShaMan123:
coverage dropped

ShaMan123:
The remaining 4 `@ts-nocheck` files are dead files or unmigrated ones

ShaMan123 (approved):

src/shapes/Text/TextSVGExportMixin.ts
```diff
@@ -293,7 +293,7 @@ export class TextSVGExportMixin extends FabricObjectSVGExportMixin {
    * @return {String}
    */
   getSvgStyles(this: TextSVGExportMixin & Text, skipShadow?: boolean) {
-    // @ts-ignore this line is not error free. ( super is not the same type as this )
+    // @ts-expect-error TS doesn't respect this type casting
```
asturur, line 296: to my vscode thise line wasn't an error and LINT was complaining i was using an unused expect-error. Don't you have that?

asturur: here it is in current master 
image

ShaMan123: Same
But the rollup ts plugin throws

PR #9129, opened 2023-07-28
chore(TS): export types

ShaMan123:
## Changes

A few more export types

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -39,8 +39,6 @@ export {
   multiplyTransformMatrices,
   multiplyTransformMatrixArray,
   isIdentityMatrix,
-  angleToSkew,
-  skewToAngle,
 } from './misc/matrix';
 export type { TextStyleArray } from './misc/textStyles';
 export {
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -18,6 +18,13 @@ export {
 } from './misc/radiansDegreesConversion';
 export { rotatePoint } from './misc/rotatePoint';
 export * from './misc/projectStroke';
+export type {
+  TComposeMatrixArgs,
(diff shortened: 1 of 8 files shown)


ShaMan123 (commented):
looks ready

ShaMan123 (commented):
Now is the time to decide about naming
You said you do not like `Props` and prefer `Options`: `CircleProps` => `CircleOptions`
I am fine with that

jiayihu:
`CircleOptions` is more consistent with the JS variable name indeed

src/util/index.ts
```diff
@@ -32,6 +39,8 @@ export {
   multiplyTransformMatrices,
   multiplyTransformMatrixArray,
   isIdentityMatrix,
+  angleToSkew,
+  skewToAngle,
```
asturur, line 43: ae those useful?

ShaMan123: Do not know
But might be
Will save a dev frim reading about skew
Not important for me

ShaMan123:
@asturur change this PR as you wish and merge when possible
I am not attached to anything apart from our long lasting buf about `T` prefixing 🤫🤐

ShaMan123:
> `CircleOptions` is more consistent with the JS variable name indeed

I think we should do it in a separate PR - it is huge diff (34 files)

src/util/index.ts
asturur: I removed them for now, when we have time to put a description over those and explain what they do and what is the value returned useful for, we can expose them.

asturur:
Yes in general i don't like that we call PROPS everything than before was options because of... react i suppose.
But let's expose the types, and circle back on that if we want

asturur (approved):

PR #9130, opened 2023-07-28
Fix event types and .once this binding

jiayihu:
- Made event type definitions interfaces to allow again declaration merging
- Fixed the types for the `on, once, off` event registering. It was particularly not working for registering with an object of event handlers, since the type of `EventRegistryObject` was `Record<K, TEventCallback<E>>`, meaning that **every** event has the same callback type, whereas each event has its own.
- Fixed a bug with `obj.once("event", function() { console.log(this) })` where `this === undefined`, whereas `obj.on("event", function() { console.log(this) })` would correctly have `this === object`

### DIFF
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -1,6 +1,6 @@
 export type TEventCallback<T = any> = (options: T) => any;
 
-type EventRegistryObject<E = any> = {
+type EventRegistryObject<E> = {
   [K in keyof E]?: TEventCallback<E[K]>;
 };
 
@@ -38,11 +38,9 @@ export class Observable<EventSpec> {
     }
     if (typeof arg0 === 'object') {
       // one object with key/value pairs was passed
-      for (const eventName in arg0) {
-        const event = eventName as string as K; // Technically eventName can be also a number or symbol
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        this.on<K, EventSpec[K]>(event, arg0[event]!);
-      }
+      Object.entries(arg0).forEach(([eventName, handler]) => {
+        this.on(eventName, handler as TEventCallback);
+      });
       return () => this.off(arg0);
     } else if (handler) {
       const eventName = arg0;
@@ -81,16 +79,14 @@ export class Observable<EventSpec> {
     if (typeof arg0 === 'object') {
       // one object with key/value pairs was passed
       const disposers: VoidFunction[] = [];
-      for (const eventName in arg0) {
-        const event = eventName as string as K; // Technically eventName can be also a number or symbol
-        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-        disposers.push(this.once(event, arg0[event]!));
-      }
+      Object.entries(arg0).forEach(([eventName, handler]) => {
+        disposers.push(this.once(eventName, handler as TEventCallback));
+      });
       return () => disposers.forEach((d) => d());
     } else if (handler) {
       const disposer = this.on<K, E>(
         arg0,
-        function (this: Observable<EventSpec>, ...args) {
+        function onceHandler(this: Observable<EventSpec>, ...args) {
           handler.call(this, ...args);
           disposer();
         }
@@ -155,10 +151,9 @@ export class Observable<EventSpec> {
     }
     // one object with key/value pairs was passed
     else if (typeof arg0 === 'object') {
-      for (const eventName in arg0) {
-        const event = eventName as string as K; // Technically eventName can be also a number or symbol
-        this._removeEventListener(event, arg0[event]);
-      }
+      Object.entries(arg0).forEach(([eventName, handler]) => {
+        this._removeEventListener(eventName as K, handler as TEventCallback);
+      });
     } else {
       this._removeEventListener(arg0, handler);
     }
--- a/src/Observable.ts
+++ b/src/Observable.ts
@@ -24,10 +24,6 @@ export class Observable<EventSpec> {
     eventName: K,
     handler: TEventCallback<E>
   ): VoidFunction;
-  on<K extends string, E>(
-    eventName: K,
-    handler: TEventCallback<E>
-  ): VoidFunction;
   on(handlers: EventRegistryObject<EventSpec>): VoidFunction;
   on<K extends keyof EventSpec, E extends EventSpec[K]>(
     arg0: K | EventRegistryObject<EventSpec>,
@@ -39,7 +35,7 @@ export class Observable<EventSpec> {
     if (typeof arg0 === 'object') {
       // one object with key/value pairs was passed
       Object.entries(arg0).forEach(([eventName, handler]) => {
-        this.on(eventName, handler as TEventCallback);
+        this.on(eventName as K, handler as TEventCallback);
       });
       return () => this.off(arg0);
     } else if (handler) {
@@ -67,10 +63,6 @@ export class Observable<EventSpec> {
     eventName: K,
     handler: TEventCallback<E>
   ): VoidFunction;
-  once<K extends string, E>(
-    eventName: K,
-    handler: TEventCallback<E>
-  ): VoidFunction;
   once(handlers: EventRegistryObject<EventSpec>): VoidFunction;
   once<K extends keyof EventSpec, E extends EventSpec[K]>(
     arg0: K | EventRegistryObject<EventSpec>,
@@ -80,7 +72,7 @@ export class Observable<EventSpec> {
       // one object with key/value pairs was passed
       const disposers: VoidFunction[] = [];
       Object.entries(arg0).forEach(([eventName, handler]) => {
-        disposers.push(this.once(eventName, handler as TEventCallback));
+        disposers.push(this.once(eventName as K, handler as TEventCallback));
       });
       return () => disposers.forEach((d) => d());
     } else if (handler) {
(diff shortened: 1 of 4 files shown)


src/Observable.ts
```diff
@@ -1,9 +1,8 @@
 export type TEventCallback<T = any> = (options: T) => any;
 
-type EventRegistryObject<
-  K extends string | number | symbol = string,
-  E = any
-> = Record<K, TEventCallback<E>>;
+type EventRegistryObject<E = any> = {
```
ShaMan123, line 3: should E extend `Record<string, any>`?
Then `K` should always be string and no symbol etc.

src/Observable.ts
```diff
@@ -29,11 +28,9 @@ export class Observable<EventSpec> {
     eventName: K,
     handler: TEventCallback<E>
   ): VoidFunction;
+  on(handlers: EventRegistryObject<EventSpec>): VoidFunction;
   on<K extends keyof EventSpec, E extends EventSpec[K]>(
-    handlers: EventRegistryObject<K, E>
```
ShaMan123, line 33: Why did you remove this overload?
The base signature is not exposed by TS to the consumer, is it?

src/Observable.ts
```diff
@@ -42,7 +39,9 @@ export class Observable<EventSpec> {
     if (typeof arg0 === 'object') {
       // one object with key/value pairs was passed
       for (const eventName in arg0) {
-        this.on(eventName as K, arg0[eventName]);
+        const event = eventName as string as K; // Technically eventName can be also a number or symbol
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        this.on<K, EventSpec[K]>(event, arg0[event]!);
```
ShaMan123, line 44: I dislike this syntax.
`event` is a confusing name in this context, I also prefer inlining it of possible.
Does `K` become a string after applying my first comment? If so it resolves all this.

src/Observable.ts
```diff
@@ -74,25 +73,28 @@ export class Observable<EventSpec> {
...
       }
       return () => disposers.forEach((d) => d());
     } else if (handler) {
-      const disposer = this.on<K, E>(arg0, (...args) => {
-        handler(...args);
-        disposer();
-      });
+      const disposer = this.on<K, E>(
+        arg0,
+        function (this: Observable<EventSpec>, ...args) {
```
ShaMan123, line 93: maybe we should give it a name for debugging
`onceHandler`?

ShaMan123 (commented):
Looks good, see comments
Could yuo add a test for the once handler, so we cover it?

ShaMan123:
Also, what do you think about a TS test suite? #8392

src/Observable.ts
jiayihu: > should E extend Record<string, any>?

I tried it but it doesn't seem to help to restrict the key type to `string`. In turn it would force `EventSpec` to be a `Record` it self, which is okay, I just didn't do it to avoid in turn affecting maybe other types as well. But if you're okay with it, I can try.

src/Observable.ts
jiayihu: It looks removed but it isn't, look at the final code on the right
Screenshot 2023-07-29 at 13 00 54

src/Observable.ts
jiayihu: > Does K become a string after applying my first comment? If so it resolves all this.

When I tried yesterday, no. I didn't investigate further though. 

Anyway it would still be:

```ts
this.on<K, EventSpec[K]>(eventName as K, arg0[eventName as K]!);
```

Are you okay with that? For me all that typing makes the code less readable than `this.on<K, EventSpec[K]>(event, arg0[event]!)` but it's up to you.

src/Observable.ts
jiayihu: Okay I've tried it and it still gives this error

Screenshot 2023-07-29 at 13 09 09

Using `Object.entries` fixes the issue, probably `for (key in obj)` is hardcoded to have type `string | number | symbol` because you're looping over all the keys, even non-own properties.

However, `extends Record<string, any>` brought to `CommonMethods<CanvasEvents>` giving error because `CanvasEvents` cannot be assigned to `Record<string, any>`. It works only with `type`, not `interface`. But we need `interface` for declaration merging.

See:
- https://github.com/microsoft/TypeScript/issues/15300
- https://www.typescriptlang.org/play?#code/C4TwDgpgBAogbhAdsAShA5gSwM7AE4gA8MUEAHsEgCbZRoDGA9nlYbnpougDRSICuAWwBGEPAD5xUALxQA3lADaAaSicoAawghGAM1gBdAFx8hovFAC+AKGuhIsBMgDKkejLoQmLNvk49TETFxa05KPF0AQ3poeCRgVy8AJnlrKHSoSJMk6xs7cGgAQQ845DQsdiJShLcQ+2gAIRKnVAwcfCqWxPok8SA

ShaMan123: Does [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogbhAdsAShA5gSwM7AE4gA8MUEAHsEgCbZRoDGA9nlYbnpougDRSICuAWwBGEPAD5xUALxQA3lADaAaSicoAawghGAM1gBdAFx8hovFAC+AKGuhIsBMgDKkejLoQmLNvk49TETFxa05KPF0AQ3poeCRgVy8AJlIKalo4lzd5ayg8qEiTJOsbO3BoAEEPTNQMHHwiGsT6EPtoACFqp1qsdkbu5qTxIA) solve the issue?

I didn't fully understand the interface limitation. Didn't read it thoroughly.
This is when I say ohh TS... why???

ShaMan123: I am ok using Object.entries
Very annoying the for in loop wrong stqtic types

ShaMan123: Another question about declaration merging
Why not pass an augmented type to the class? That was my intention we I wrote it. I didn't think of declaratiin mergign because I don't use it so much.
I am used to generics

ShaMan123: So if i understand correctly using types is better for maintainers because TS is simpler to use and interfaces for devs/consumers for declarstiin merging but that takes a toll from maintaining?

ShaMan123: > I am ok using Object.entries Very annoying the for in loop wrong static types Just for discussion, do you know of a perf diff between them?

Or we just `@ts-expect-error`

ShaMan123 (approved):
I noticed this

https://github.com/jiayihu/fabric.js/blob/35a27e37606db22fb857b05891d55e422cb99a8d/src/Observable.ts#L27-L30

https://github.com/jiayihu/fabric.js/blob/35a27e37606db22fb857b05891d55e422cb99a8d/src/Observable.ts#L70-L73

Or on master:

https://github.com/fabricjs/fabric.js/blob/d2c7b897664c6821339d142b41e7a8023bed965f/src/Observable.ts#L28-L31

https://github.com/fabricjs/fabric.js/blob/d2c7b897664c6821339d142b41e7a8023bed965f/src/Observable.ts#L73-L76

Why is it there? I am to blame... Maybe the prev type needed it.
Seems that it should be removed, right?

src/Observable.ts
jiayihu: > Does [this](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogbhAdsAShA5gSwM7AE4gA8MUEAHsEgCbZRoDGA9nlYbnpougDRSICuAWwBGEPAD5xUALxQA3lADaAaSicoAawghGAM1gBdAFx8hovFAC+AKGuhIsBMgDKkejLoQmLNvk49TETFxa05KPF0AQ3poeCRgVy8AJlIKalo4lzd5ayg8qEiTJOsbO3BoAEEPTNQMHHwiGsT6EPtoACFqp1qsdkbu5qTxIA) solve the issue?

Not really, because at that point you can do `const value: EventSpec2 = { a: 2, b: 3 }`, e.g. add extra properties without error. And that's exactly why TS gives the error in my example, because technically you're saying that something with "well-defined properties" is assignable to a type where "any string index can be used", which is unsafe.

> Just for discussion, do you know of a perf diff between them?

Screenshot 2023-07-30 at 12 43 07

There is definitely an overhead in using functions, but not significant. Also not relevant for an object of event handlers, whose number of entries is only a few and the code is not in a critical hot path (rendering).

> Why not pass an augmented type to the class? That was my intention when I wrote it. I didn't think of declaration merging because I don't use it so much. I am used to generics

Declaration merging is different than generics. Generic is for a specific instance, a declaration merging extends an existing interface wherever it's used. It can also significantly harder for the lib to maintain good generics. For instance the following would need to be fixed as the `EventSpec` generic is currently not used:

Screenshot 2023-07-30 at 12 45 29

But if you fix everything and make it work, it's definitely a doable alternative. At that point not sure if there are still other use cases where declaration merging for canvas/object events might be helpful.

jiayihu:
> Why is it there? I am to blame... Maybe the prev type needed it.
Seems that it should be removed, right?

I don't know but I noticed it to be weird indeed. I can remove it if you want, no idea why it was there though.

ShaMan123:
> > Why is it there? I am to blame... Maybe the prev type needed it.
> > Seems that it should be removed, right?
> 
> I don't know but I noticed it to be weird indeed. I can remove it if you want, no idea why it was there though.

lets remove the wierd artifact

jiayihu:
Done

ShaMan123 (approved):
I am good with this PR, it also fixes autocomplete of `.on({ ... })`
Declaration merging, interface vs. type, poses a problem that we should decide about soon.
@asturur do you accept it for 6.0?

asturur (dismissed):

PR #9133, opened 2023-07-30
ci(playwright): better eval

ShaMan123:
## Description

Use playwright internals (`evaluateHandle`) to evaluate stuff in the browser instead of re implementing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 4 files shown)


ShaMan123 (commented):
more a cleanup than anything else

asturur:
i think test() is more appropriate for this kind of changes rather than ci()

asturur (approved):

PR #9112, opened 2023-07-20
test(): text dnd

ShaMan123:
## Motivation

Testing the formerly untestable drag and drop feature.
After a lot of hours I finally managed to get it to work.
Eraser is next
Playwright is powerful - it was a very good decision

## Changes

- removed `e2e/imports.ts` in favor of a full importmap produced in runtime
- made babelrc esm
- added tests

PR #9134, opened 2023-07-30
ci(e2e): node canvas visual tests

ShaMan123:
## Motivation

Use playwright to do visual diffs with browser and node.

This PR add a test example that is structured as follow:
- index.ts does not do anything if not running a function that setup the canvas, returns an empty map because no interaction is planned for those tests during the test code.
- the setup function code is shared in a file, since it needs to be imported also somewhere else
- the test code has 2 steps: browser and node and they strictly just match the snapshot of the canvas with the golden

to make this possible an util has been added `createNodeSnapshot` that will take as argument the setup function from the test, running the same function in node, and returning the buffer of the canvas that is what the matchSnapshot function would expect

## Description

Added a template that demonstrates how to do use e2e with node canvas.
Fairly simple.

TODO:
- [ ] register fonts in node canvas - first merge #9112 
- [ ] fabric/node coverage


ShaMan123 (commented):
not bad...
could be better.
for now I think we can move forward
Thoughts?

ShaMan123:
Ahh coverage

e2e/tests/template-with-node/index.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Runs in the **BROWSER**
+ */
```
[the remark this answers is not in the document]
asturur: ok now i get it

asturur:
tests are becoming flaky already we need to be careful of how we run them.
i think 1-2 at time is maximum if we have many tests.
on my machine they already get stuck easily because macos is eventually unhappy of opening 5  chrome all together

asturur:
There is something that doesn't make this test pass and i don't understand what it is.

ShaMan123:
> tests are becoming flaky already we need to be careful of how we run them.
> i think 1-2 at time is maximum if we have many tests.
> on my machine they already get stuck easily because macos is eventually unhappy of opening 5  chrome all together

Did you try to change the parallel config?

ShaMan123:
I also saw jest has a plugin for image diffs...

asturur:
the reporting features of playwright are great imho, but anyway, why is this failing? did it ever pass to you in the CI?

ShaMan123:
I will look into it when I am back
Seems the setup is running twice for some reason

e2e/setup/index.ts
ShaMan123, line 1: seems that importing these files is what made tests flaky.
Calling the setup as a method seems to fix it

e2e/setup/setupSelectors.ts
ShaMan123, line 1: this is also flaky, fixed by trycatch

ShaMan123 (commented):
FIX FLAKINESS

ShaMan123:
@asturur please merge this asap

asturur:
i wonder if this also solves the issue i was having on text tests.
One issue with copy paste test is that it SEEMS to me that 2 text tests running in parallel are sharing the same instance of fabricjs and so one consumes the clipboard data ( styles ) of the other.

asturur (approved):

PR #9139, opened 2023-08-03
chore(TS): mv + rename `TProps` => `TOptions`

ShaMan123:
## Motivation

I am typing canvas options and need `TProps`.
It was under `Obejct` which doesn't make sense anymore if using in canvas files.
So I moved it to typedefs and renamed to TOptions to get started with renaming

## Description

A lot of ink but nothing happening here


ShaMan123 (commented):
Seems ready

ShaMan123 (commented):
READY

asturur (approved):

PR #9140, opened 2023-08-03
chore(TS): canvas options

ShaMan123:
## Description

First merge #9139 

Typing canvas options

## Changes

- rm `interactive` flag 0d0ce13
- typed `backgroundImage`, `overlay` as undefined instead of null and removed from default values with clipPath

## Gist

2 difficult commits 86927bd 6791d74


ShaMan123 (commented):
stable

test/unit/canvas_static.js
```diff
@@ -192,8 +192,6 @@
     var canvas = new fabric.StaticCanvas();
     assert.ok('backgroundColor' in canvas);
     assert.ok('overlayColor' in canvas);
-    assert.ok('backgroundImage' in canvas);
-    assert.ok('overlayImage' in canvas);
```
asturur, line 196: do you know why those are not true anymore?

asturur: maybe `canvas.hasOwnProperty` would work? how do wecheck we don't accidentally remove stuff from the classes? this was meant to do this only

src/canvas/CanvasOptions.ts
```diff
@@ -0,0 +1,296 @@
...
+
+export interface CanvasTransformOptions {
+  /**
+   * When true, objects can be transformed by one side (unproportionally)
+   * when dragged on the corners that normally would not do that.
+   * @type Boolean
+   * @default
+   * @since fabric 4.0 // changed name and default value
+   */
+  uniformScaling: boolean;
```
asturur, line 13: In retrospective i don't like the idea of moving the documentation out of the classes themselves. I understand it makes files shorter but it seems the wrong solution.
\I know we did this already and is not new, but now i really dislike it.

Opening a class and being able to read is important to me

test/unit/canvas_static.js
asturur: just making a test....

src/canvas/CanvasOptions.ts
ShaMan123: If you hover over a prop it shows the jsdoc.
I like it but that is why I kept it in separate commits... because I wanted to give you an easy way out.

test/unit/canvas_static.js
ShaMan123: No need
I wrote in the description that I changed the type from null to undefined.
So there was no sense in setting these field to undefined in default props so I removed them.
That is all.

src/canvas/CanvasOptions.ts
asturur: no if we go out we change them all. For now we keep consistency.
I know hovering works but reading is reading. It is also a vscode thing not an universal one

asturur (approved):

asturur (approved):

src/canvas/CanvasOptions.ts
ShaMan123: So should I revert or not?

ShaMan123: Should the jsdoc exist on the options as well?
For a dev the important place is on options so when creating a canvas ir object they see what it can do.

test/unit/canvas_static.js
ShaMan123: Not sure.
But that is the point of declare, isn't it?
Without declare it is defined as void 0.
But then we have problems in initialization.
That can be fixed of coursr but will require a lot if refactoring

src/canvas/CanvasOptions.ts
jiayihu: I agree with @asturur, better a big module with everything related than death-by-thousand-files. While vscode hover documentation helps, there are many other occasions where it's useful to have things colocated in the same file, e.g. PR reviewing or debugging.

jiayihu: I'd recommend this book where the topic of "deep" modules is explained: https://www.amazon.com/Philosophy-Software-Design-John-Ousterhout/dp/1732102201, i.e. a **deep** module with the relevant implementations all together (as far as the exposed public API is small) is better than many **shallow** modules that you need to put together in your mind in order to understand.

ShaMan123:
I removed the `interactive` flag because fabric doesn't use it or need. 
It has expectations we do not want to accomodate and was there in the first place because of old patterns (mainly usage of canvas mixins) which are dead. In other words, this is part of the v6 cleanup.
Regarding moving back all the jsdoc, 2 vs. 1, I lost. Is there a third option?

jiayihu:
> Moving it back makes the interface concept a bit missing IMO.

What do you mean? In TS (and JS) the `interface` keyword has no strong OOP meaning and indeed I believe some naming in TS is misleading people to bad OOP habits. For me an `interface` is treated like a `type`, just with some technical differences in what you can do with them (declaration merging, type manipulation etc), but not in their semantic.
To make another example, Generics is also a bad name compared to "Type parameters" which much better represent the correct view.

PR #9141, opened 2023-08-03
fix(ActiveSelection): expose `selectAll`

ShaMan123:
## Motivation

I need to override adding objects to active selection.
`multiSelectAdd` is great but we need another method for the group selector.
I thought it is best to merge them together but I am not sure it is backward compatible and there are different use cases that may need the different context.

## Changes

- renamed `multiSelectAdd` to `multiSelect` - since we are in beta I think it is fine
- exposed `selectAll`

### DIFF
--- a/src/shapes/ActiveSelection.ts
+++ b/src/shapes/ActiveSelection.ts
@@ -43,6 +43,10 @@ export class ActiveSelection extends Group {
     //  noop
   }
 
+  batchSelectAdd(...targets: FabricObject[]) {
+    this.add(...targets);
+  }
+
   /**
    * Adds objects with respect to {@link multiSelectionStacking}
    * @param targets object to add to selection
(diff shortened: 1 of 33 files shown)


ShaMan123 (commented):
DONE

src/shapes/ActiveSelection.ts
```diff
@@ -43,11 +43,15 @@ export class ActiveSelection extends Group {
     //  noop
   }
 
+  selectAll(...targets: FabricObject[]) {
+    this.add(...targets);
+  }
```
asturur, line 48: if we have multiSelect already i don't see to make sense.
This is a selectAll that takes targets as an input? so what the ALL part does? all of what?

ShaMan123: All of the objects in the selection...
Naming can be modified.
Also we can use multiSelect and pass more context so that the method knows what calle it

jiayihu: It's weird to me that a method called `selectAll` doesn't replace the existing selection, instead it adds to the selection. Should be called `addAll` then? Also not clear what's its usage exactly if one doesn't need to override it.

ShaMan123:
closed => #9142

PR #9142, opened 2023-08-04
Patch selection methods

ShaMan123:
## Motivation

closes #9141 

## Description

A different impl

## Changes

1 signature `addToSelection` with enough context - naming can be changed of course

### DIFF
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1570,7 +1570,7 @@ export class Canvas extends SelectableCanvas {
       this.setActiveObject(objects[0], e);
     } else if (objects.length > 1) {
       // add to active selection and make it the active object
-      this._activeSelection.select(...objects);
+      this._activeSelection.selectAll(...objects);
       this.setActiveObject(this._activeSelection, e);
     }
 
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1506,7 +1506,7 @@ export class Canvas extends SelectableCanvas {
           }
         } else {
           //  `target` isn't part of active selection => add it
-          activeSelection.multiSelectAdd(target);
+          activeSelection.multiSelect(target);
           this._hoveredTarget = activeSelection;
           this._hoveredTargets = [...this.targets];
         }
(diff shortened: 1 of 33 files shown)


ShaMan123:
@asturur take a lot if yu prefer this instead of #9141

src/shapes/ActiveSelection.ts
```diff
@@ -47,8 +47,11 @@ export class ActiveSelection extends Group {
    * Adds objects with respect to {@link multiSelectionStacking}
    * @param targets object to add to selection
    */
-  multiSelectAdd(...targets: FabricObject[]) {
-    if (this.multiSelectionStacking === 'selection-order') {
+  addToSelection(mode: 'multi-select' | 'default', ...targets: FabricObject[]) {
+    if (
+      mode === 'default' ||
```
jiayihu, line 52: `default` isn't really a good name. What's default? It doesn't seem to do much, the only real difference in the implementation is `multiSelectionStacking`. So the use cases:

```ts
activeSelection.addToSelection('multi-select', activeObject, target);
this._activeSelection.addToSelection('default', ...objects);
```

can be changed to:

```ts
activeSelection.addToSelection(activeObject, target, { multiSelectionStacking: true });
this._activeSelection.addToSelection(...objects,  { multiSelectionStacking: false }); // equivalent to mode default
```

By adding a `mode` you're making the method much harder to reason about as the technically combinations have cardinality `mode x multiSelectionStacking`, 2x2=4 in this case. At the end of the day, the difference is in how objects are added to the selection, not the "mode" right?

ShaMan123:
This is a better option imo as well.
Naming is not good for sure. I didn't know what to choose. Default sucks in must cases.
Options is fine instead of mode, indeed more versitle, but then the signature is not a tupple and that differs from add, insertAt but is not a big deal.

ShaMan123:
I think the caller context is important for apps to decide how to customize selection

ShaMan123:
After discussion on the topic:
- Suggested way forward is to use `onSelect` to block objects from being selected and overrride exisiting `collectObjects`, `multiSelectAdd`.
- patch `collectObjects` to support the stacking option
- It would be good that all logic resided on the same entity as opposed to now, part on Canvas, part on ActiveSelection.
- `multiSelectAdd` is a bad name

src/canvas/Canvas.ts
```diff
@@ -1570,7 +1570,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       this.setActiveObject(objects[0], e);
     } else if (objects.length > 1) {
       // add to active selection and make it the active object
-      this._activeSelection.add(...objects);
+      this._activeSelection.addToSelection('default', ...objects);
```
ShaMan123, line 1573: this line is the real change

ShaMan123 (commented):
This PR is not ready for merging and needs more discussion

PR #9144, opened 2023-08-04
fix(): class registry

ShaMan123:
## Motivation

closes #8507 

## Changes

- changed signature of `getXXXClass` accepting a `strict` param indicating if it should throw or not
- use registry to resolve `enlivenObjectEnlivenables`
- changed `getClass` signature to accept an object, needed to resolve Gradient in `enlivenObjectEnlivenables`
- refactored `enlivenObjectEnlivenables` to return an object **only** with enlived stuff, previously was returning everything
- changed Gradient to fit the registry
-  enliven Shadow in `enlivenObjectEnlivables` not in `set` (I don't think it is breaking but devs can add `type: shadow` to data in case they fall into an edge case)
- fix `fromObject` regression 3fd06607d


ShaMan123 (commented):
stable

ShaMan123 (commented):
seems ready

src/util/misc/objectEnlive.ts
```diff
@@ -118,54 +115,39 @@ export const enlivenObjects = <
...
-            return pattern;
-          });
-      }
-      return value;
+    const promises = Object.values(serializedObject).map(async (value) => {
+      const klass =
+        value &&
+        typeof value === 'object' &&
+        classRegistry.getClass(value, false);
+      if (!klass) return;
```
ShaMan123, line 139: instead of passing all the options and overriding with enlivened ones now this method returns only what was enlivened by the method

asturur:
I m sorry but this PR has been open so long because it was never explained what it wants to achieve.
And still i don't get it.

Does it want to align shadow to be used to the registry? that is fine
Does it want to align Gradient to be used in the registry? that is fine

But the fixes to the class registry, that is the title of the PR is unclear what they want to achieve and how.

asturur:
Let's start with
shadow: https://github.com/fabricjs/fabric.js/pull/9626
gradient: https://github.com/fabricjs/fabric.js/pull/9627

That seems the easy parts, and let's see what is left after.

asturur:
@ShaMan123 let's focus on this since you said this was your last important pr for v6

PR #9151, opened 2023-08-11
Avoid unwanted mutation in Group

jiayihu:
Minor fix: since `Collection` mutates the `_objects` array, this can cause really hard-to-debug bugs if the passed array of objects is used for other purposes as well by the application. To avoid any complication for the unaware developer, a copy should be made.

### DIFF
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -173,7 +173,7 @@ export class Group extends createCollectionMixin(
     objectsRelativeToGroup?: boolean
   ) {
     super();
-    this._objects = objects;
+    this._objects = objects.slice(); // Avoid unwanted mutations of Collection to affect the caller
     this.__objectMonitor = this.__objectMonitor.bind(this);
     this.__objectSelectionTracker = this.__objectSelectionMonitor.bind(
       this,


asturur (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- patch(): Avoid unwanted mutation in Group [#9151](https://github.com/fabricjs/fabric.js/pull/9151)
```
asturur, line 5: I think the correct commit would be:
patch(Group): Avoid unwanted mutation to passed objects array to Group constructor

asturur:
maybe a dumb unit test , an new group JEST file, that tests that given an array and doing a collection mutation on it, the original array is still intact, and that the 2 arrays do not pass the strict equality test.

If we wanted to go full mutation protection we should clone the object, that would make the constructor async and i don't think we should do it at all.

jiayihu:
Done. Renamed changelog text and added a Jest test.

src/shapes/Group.ts
```diff
@@ -173,7 +173,7 @@ export class Group extends createCollectionMixin(
     objectsRelativeToGroup?: boolean
   ) {
     super();
-    this._objects = objects;
+    this._objects = objects.slice(); // Avoid unwanted mutations of Collection to affect the caller
```
ShaMan123, line 176:
```suggestion
    this._objects = objects.slice(); // Avoid mutations
```

ShaMan123 (commented):
the comment can be simpler but it is not a deal breaker

ShaMan123:
@asturur can you look into giving the workflows permissions so that forks will be able to pass all tests?
I remember I opened an issue for it

src/shapes/Group.ts
jiayihu: I think it should be mentioned that the mutations come from Collection as it's not evident since the constructor is not mutating it

ShaMan123: ok

ShaMan123 (approved):

asturur (approved):

PR #9152, opened 2023-08-14
refactor(): Layout Mananger

ShaMan123:
## Motivation

#9148 

## Changes

TODO:

- fix that strange initialization logic with rotation, skewing


ShaMan123 (commented):
This is ready for a review

Extracted logic to LayoutManager that accepts a LayoutResolver which is what the dev should override.

There is this issue with skewing and rotation on first layout which I can't get pass but this is good enough for v6

Exposed on Group:
triggerLayout, onBeforeLayout, onLayout and layout:before, layout events.

A lot of renaming

src/LayoutManager/LayoutManager.ts
```diff
@@ -115,11 +92,15 @@ export class LayoutManager {
...
-      true
+    const relativeCorrection = new Point(
+      result.relativeCorrectionX ?? 0,
+      result.relativeCorrectionY ?? 0
     );
+    const offset = prevCenter
+      .subtract(nextCenter)
+      .add(correction)
+      .transform(invertTransform(target.calcOwnMatrix()), true)
+      .add(relativeCorrection);
```
ShaMan123, line 103: added for ease of use

ShaMan123 (commented):
This is looking good

ShaMan123 (commented):
I think I am done for this PR
next PR can try to fix initialization with rotation, skew

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    const prevCenter = target.getRelativeCenterPoint();
+    const result = context.strategy.calcLayoutResult(
+      context,
+      target.getObjects()
+    );
+    if (!result) {
+      return;
+    }
+    const nextCenter = new Point(result.centerX, result.centerY);
+    const correction = new Point(
```
jiayihu, line 92: What's the correction?

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    };
+
+    this.onBeforeLayout(strictContext);
+
+    const layoutResult = this.getLayoutResult(strictContext);
+    let bubblingContext: LayoutResult | undefined;
+
+    if (!this._firstLayoutDone) {
+      if (layoutResult) {
+        this.commitLayout(strictContext, layoutResult);
```
jiayihu, line 41: It seems like:

```ts
this.commitLayout(strictContext, layoutResult);
bubblingContext = layoutResult;
```

is called regardless of `this._firstLayoutDone` as far as `layoutResult` is defined. So it could be extracted before?

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,184 @@
...
+    this._prevLayoutStrategy = strictContext.strategy;
+  }
+
+  protected onBeforeLayout(context: StrictLayoutContext) {
+    const { target } = context;
+
+    //  fire layout hook and event (event will fire only for layouts after initialization layout)
+    target.onBeforeLayout({
+      context,
+    });
```
jiayihu, line 73: Maybe the layout strategy needs to be called as well if it needs to do some preparation? Just wondering

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,184 @@
...
+   */
+  protected layoutObject(
+    context: StrictLayoutContext,
+    { offset }: Required<LayoutResult>,
+    object: FabricObject
+  ) {
+    object.setRelativeXY(object.getRelativeXY().add(offset));
+  }
+
+  protected onLayout(context: StrictLayoutContext, layoutResult: LayoutResult) {
```
jiayihu, line 159: For consistency with `onBeforeLayout` it should be `onAfterLayout`

src/LayoutManager/LayoutStrategies/FitContentLayout.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { Group } from '../../shapes/Group';
+import type { FabricObject } from '../../shapes/Object/FabricObject';
+import { makeBoundingBoxFromPoints } from '../../util/misc/boundingBoxFromPoints';
+import type { LayoutStrategyResult, StrictLayoutContext } from '../types';
+import { LayoutStrategy } from './LayoutStrategy';
+import { getObjectBounds, getObjectSizeVector } from './utils';
+
+export class FitContentLayout extends LayoutStrategy {
+  /**
+   * optimize layout when possible
```
jiayihu, line 11: Should explain what's exactly is delayed. Code should always make clear "what" and "why".

src/LayoutManager/LayoutStrategies/LayoutStrategy.ts
```diff
@@ -0,0 +1,186 @@
...
+        height: height,
+        strokeWidth: 0,
+      }),
+      bboxSizeAfter = target._getTransformedDimensions({
+        width: bbox.width,
+        height: bbox.height,
+        strokeWidth: 0,
+      });
+
+    //  calculate center and correction
```
jiayihu, line 100: What's this correction? It looks like you want to handle when the group has initial position specified, but the objects are absolute so you calculate a correction for the objects so that they end where the group position is? It look however a use case whose handling is leaking the implementation detail `correction`, which makes the rest more complicated as well.

src/LayoutManager/LayoutStrategies/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  const sizeVector = getObjectSizeVector(object);
+  return [objCenter.subtract(sizeVector), objCenter.add(sizeVector)];
+};
+
+export const shouldPerformLayout = (context: StrictLayoutContext) => {
+  return (
+    context.type === 'initialization' ||
+    context.type === 'imperative' ||
+    context.strategyChange
+  );
```
jiayihu, line 32: What about when objects are added/deleted/changed?

src/LayoutManager/LayoutManager.ts
ShaMan123: will do

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: I think the name is misleading... this is a minor util for layout that normally do not change when objects are changed

src/LayoutManager/LayoutStrategies/LayoutStrategy.ts
ShaMan123: The correction value is the nastiest bit of logic of the layout manager.
It is needed because objects are relative to the group center but then the group changes size so an object that is expected to align to the top left corner of group needs to apply a correction to its position so it relates to the new center - I will add this to the docs

src/LayoutManager/LayoutStrategies/FitContentLayout.ts
ShaMan123: I will add comments
I was thinking of this use case after our chat today...
Adding scaling, rotating, resizing, moving triggers makes sense to me.
In fact I want to move the trigger logic to the manager, to an intialize/attach method so you can also fine tune them.
Back to topic - if we add these triggers we can use the lazy concept to calc layout of the transforming object and cache the rest of the objects' bbox for quick computation.

ShaMan123: you think `optimize`` is a better name??

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: please suggest a better name

ShaMan123:
I will paste this on my laptop ;)
As mentioned in a prev comment I want to extract the triggers to the manager as well - look at `Group#_watchObject`

src/LayoutManager/LayoutStrategies/LayoutStrategy.ts
ShaMan123: I have an elaborate solution for dropping the need for the correction value... but it might degrade perf and limit customization.
I might expose a layout strategy that does that for the common good

ShaMan123 (commented):
applied some review comments + tidy up

ShaMan123 (commented):
remae FitContentLayout#lazy => optimized
shouldPerformLayout => to class method

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: moved to class

src/LayoutManager/types.ts
```diff
@@ -1,11 +1,17 @@
...
 import type { Point } from '../Point';
 import type { Group, GroupProps } from '../shapes/Group';
+import type { ITextEvents } from '../shapes/IText/ITextBehavior';
 import type { FabricObject } from '../shapes/Object/FabricObject';
 import type { LayoutStrategy } from './LayoutStrategies/LayoutStrategy';
 
 export type LayoutTrigger =
   | 'initialization'
+  | 'object_modifying'
   | 'object_modified'
```
ShaMan123, line 15: I ant to rename these

src/LayoutManager/types.ts
```diff
@@ -59,38 +79,63 @@ export type LayoutResult = {
...
+  options?: Partial<GroupProps>;
+  objectsRelativeToGroup?: boolean;
+};
+
+export type CollectionChangeLayoutContext = CommonLayoutContext & {
+  type: 'added' | 'removed';
+  targets: FabricObject[];
+};
+
+export type ObjectModifiedLayoutContext = CommonLayoutContext & {
```
ShaMan123, line 108: also

src/LayoutManager/types.ts
```diff
@@ -59,38 +79,63 @@ export type LayoutResult = {
...
+  type: 'object_modified';
+  trigger: 'modified';
+  e: ModifiedEvent;
+};
+
+export type ObjectModifyingLayoutContext =
```
ShaMan123, line 114: also

src/LayoutManager/types.ts
```diff
@@ -59,38 +79,63 @@ export type LayoutResult = {
...
+export type InitializationLayoutContext = CommonLayoutContext & {
+  type: 'initialization';
+  options?: Partial<GroupProps>;
+  objectsRelativeToGroup?: boolean;
+};
+
+export type CollectionChangeLayoutContext = CommonLayoutContext & {
+  type: 'added' | 'removed';
+  targets: FabricObject[];
+};
```
ShaMan123, line 106: todo: supporting adding relative objects

ShaMan123 (commented):
move object triggers to LayoutManager + added modifying triggers

ShaMan123 (commented):
refactored the initialization logic. horrible work

ShaMan123:
I think the current failing tests are an unrelkated regression in polyline initial position

asturur:
@ShaMan123  is the pr stable enough for review?

ShaMan123:
wait until I ping you

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,177 @@
...
+      prevCenter,
+      nextCenter,
+    } = layoutResult;
+    // set dimensions
+    target.set({ width, height });
+    // layout descendants
+    this.layoutChildren(context, layoutResult);
+    //  set position
+    !nextCenter.eq(prevCenter) &&
+      target.setPositionByOrigin(nextCenter, CENTER, CENTER);
```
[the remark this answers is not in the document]
ShaMan123: fixed

ShaMan123 (commented):
almost

ShaMan123 (commented):
@asturur this is stable and ready for a review
We need to consider serialization

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,277 @@
...
+      context,
+      result: layoutResult,
+    });
+    target.fire('layout', {
+      context,
+      result: layoutResult,
+    });
+
+    //  bubble
+    if (target.group?.layoutManager) {
```
jiayihu, line 266: I just had a case today where I traced a bug back to this. When applying syncing xy changes from the server to a group and its children, which are also groups, I don't want a `child.triggerLayout()` to bubble up. I've not finished yet updating all the objects with the final relative xy, so bubbling up will use a mix of old and new relative xy. I will manually call `parent.triggerLayout()`.

So in brief: there should be a flag in the `context` to stop the bubbling

ShaMan123: I think I understand the need but not sure how you are using `triggerLayout` in this use case.
Could you show me?
I do understand that you want to stop bubbling for perf reasons - so the top group will be layed out only once instead of each time a child lays out.

ShaMan123: Do you think we should add a method that triggers a deep layout on all its childen?
Then you call `triggerLayout({deep:true})` and you are done.

jiayihu: > I do understand that you want to stop bubbling for perf reasons

It's not about perf, rather it's a bug to bubble up the layout invalidation when in the middle of updating several children since you need to wait for *all* of them to be updated first, because the xy values are relative to the parent center.
Use case: syncing changing strokeWidth of subselected group (shape in group) => results in syncing the new xy values of the whole group and children because the parent has changed size and center. The change of strokeWidth was significant.
- For each child, we set the new xy values and for one of them the new strokeWidth. We also call `child.triggerLayout()` 
- If the layout invalidation is bubbled up, the parent will use incorrect values. Some children are sill with the new properties, some are with new ones. This results in slightly wrong positions, you may even not notice them at first. I can show you in our next call.

> Do you think we should add a method that triggers a deep layout on all its childen?
Then you call triggerLayout({deep:true}) and you are done.

Not sure, I don't think so. A child changing position invalidates the parent layout, but not vice-versa.

asturur: guys keep it minimal as minimal as possible

jiayihu: Minimal: allow to skip bubbling. Don't assume it's always needed.

ShaMan123:
Regarding tests, yes indeed we should add extensive tests.
I can do that is this pr or a follow up.
We do have the visual tests but that does not cover the lifecycle for sure.

src/LayoutManager/LayoutManager.ts
ShaMan123: The word that guides me is useful.
I want it to serve me/devs and ease developing.
Of course it should be straight forward as possible not only to use but to write and maintain.
I want to avoid deving stuff that to use in a conventional app I need to break into pieces, override and patch to a point that I almost rewrite it all and whatever fabric put there became useless for me.

src/shapes/Group.ts
```diff
@@ -385,17 +313,15 @@ export class Group extends createCollectionMixin(
...
-    object[directive]('changed', this.__objectMonitor);
-    // @ts-expect-error TS limitations
-    object[directive]('modified', this.__objectMonitor);
-    // @ts-expect-error TS limitations
-    object[directive]('selected', this.__objectSelectionTracker);
-    // @ts-expect-error TS limitations
-    object[directive]('deselected', this.__objectSelectionDisposer);
+    if (watch) {
+      object.on('selected', this.__objectSelectionTracker);
+      object.on('deselected', this.__objectSelectionDisposer);
```
asturur, line 320: we probably need internals on/off to avoid the developer can kill those events without understanding how, like doing an object.off('deselected').
Not scope of this pr.

src/shapes/Group.ts
```diff
@@ -431,7 +357,7 @@ export class Group extends createCollectionMixin(
     this._shouldSetNestedCoords() && object.setCoords();
     object._set('group', this);
     object._set('canvas', this.canvas);
-    this.interactive && this._watchObject(true, object);
+    this._watchObject(true, object);
```
asturur, line 360: why watching if the group is not interactive?

src/LayoutManager/LayoutStrategies/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  if (decomposition?.angle ?? object.angle) {
+    const rad = degreesToRadians(object.angle),
+      sine = Math.abs(sin(rad)),
+      cosine = Math.abs(cos(rad)),
+      rx = sizeVector.x * cosine + sizeVector.y * sine,
+      ry = sizeVector.x * sine + sizeVector.y * cosine;
+    return new Point(rx, ry);
+  }
+  return sizeVector;
+};
```
asturur, line 38: is this a way to find the object bounding box aligned to group axis? or what does it do?

src/shapes/Group.ts
```diff
@@ -995,7 +560,6 @@ export class Group extends createCollectionMixin(
   >(propertiesToInclude: K[] = []): Pick<T, K> & SerializedGroupProps {
     return {
       ...super.toObject([
-        'layout',
```
asturur, line 998: the first question i have is how someone persist the layout on object restore.
The layout influence how you draw the objects too and where they are, so in theory needs to be part of the restore.

One way i can think of is suclassing of course.
Another is making the layoutManager a  hydratable option with a string, but you have to go through some gymanstic since what you want to restore is the strategy and not the manager itself that is always the same

asturur:
i will take some notes and then sort the question in order of importance. I will need some hours multiple times to get comfortable with all this code.
At a first glance is on the right track, right now i have a mix of questions that is not useful to put down one by one would just make noise

src/shapes/Group.ts
ShaMan123: I agree, we can add an options object with a flag for that.
Trurth is I think this bit can be removed

src/shapes/Group.ts
ShaMan123: 1. I think we can remove the watching functionality because it is not used and can be replaced by `canvas.getActiveObjects().includes(obj)`
2. the constructor was throwing because it was setting this option before populationg `_objects` - but that can be fixed if we do not pass options to super and set them afterwards.
3. the selected/deselected events will fire only if the group is interactive so this is a redundant check making the use of interactive more annoying because if you toggle it without using `set` the group will be in a broken state.

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: caclulating objects size in group plane and taking into account the fact that an object can be selected, so it may need a plane change matrix

src/shapes/Group.ts
ShaMan123: I agree.
What about?
```
toJSON() {
  return { strategy: this.stategy.name }
}
```

I will try it out.
Do you want me to freeze this PR or can I push changes?

ShaMan123:
I always prefer a chat.

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: Resurfacing #8767... this is here because that PR/concept should redo this entire section which has bugs with rotation of selected objects

src/LayoutManager/LayoutManager.spec.ts
ShaMan123, line 1: added this massive time consuming test

src/LayoutManager/__snapshots__/LayoutManager.spec.ts.snap
```diff
@@ -0,0 +1,82 @@
...
+exports[`Layout Manager getLayoutResult #type trigger 3`] = `
+{
+  "nextCenter": Point {
+    "x": 50,
+    "y": 100,
+  },
+  "offset": Point {
+    "x": -70,
+    "y": -120,
+  },
```
ShaMan123, line 66: I do not understand why it is the same as the first call

ShaMan123 (commented):
- added static type on LayoutStrategy for persistence and toJSON on the strategy... how we move with this TBD
- added a test to LayoutManager, I think the strategies should be tested visually and not in unit tests. It has no value IMO
- patched here and there thanks to tests
- rm `strategyChanged`
- disposing

ShaMan123:
I am done apart from naming and your review

src/LayoutManager/LayoutStrategies/utils.ts
asturur: we shouldn't do this. _getTransformedDimensions wasn't meant to extend in this way and accepting skew values was already wrong.
This is a copy of code we have also somewhere else.
So this is calculating the object size but with which kind of rotation? is this aligned with the group bounding box?

src/shapes/Group.ts
asturur: let's freeze the PR and let's address the things we may disagree on, serialization support can be another

asturur:
> I always prefer a chat.

That is fine, i still have to test it and understand it, reading all methods of a class does not give an idea of what is going on.
I do not know what trigger what and how a developer is supposed to use it.

I have to pull it down, put console logs and play

asturur:
image
now 1.929.

The more you touch it the worse it get.

jiayihu:
Better merge it asap then before he touches it again 😄 Jokes aside this PR is finally making the Group layouting more sensible in fabric, since v3 it was dark magic, with v6beta a mess. This PR makes it easier to understand and customize/extend. I'm always in favour of more lines of code if the quality improves.
At the end of the day, a single image in the page is gonna cost more bytes then 100k lines.

ShaMan123:
Dark magic it was. And a mess indeed.
I suggest looking at the jest test which tests the entire API.

asturur:
Spent some time on the pr digging the code to get familiar with it.
I have a bunch of direct changes i think we should do, while i will find others maybe, those can be addressed and discussed.

- context.target in the various performLayout and other operations is always the group and is not supposed to change.
It should be a property of the layout manager rather than be passed every time? i ll comment right away in the code where it happens.

- array.flat() can't be used since is supported only by recent browsers and none of the ones we list as supported:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat
So we need to concat those results differently.

- all those 'imperative' and 'initialization' strings need to become const ( but i can simply open a pr on your branch for that since that is not even code )

- onBeforeLayout and onAfterLayout methods on the group instances do not do anything since they are hooked up to be called but not to intervene on the process. Are as useful as an event and so should be deleted since we have the event already

- comment on LayoutObject

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,294 @@
...
+  /**
+   * @param {FabricObject} object
+   * @param {Point} offset
+   */
+  protected layoutObject(
+    context: StrictLayoutContext,
+    { offset }: Required<LayoutResult>,
+    object: FabricObject
+  ) {
+    object.setRelativeXY(object.getRelativeXY().add(offset));
```
asturur, line 232: please this is exactly:
```suggestion
    object.left += offset.x;
    object.top += offset.y
```
let's not make it more complicated than it is until is really needed

src/shapes/Group.ts
```diff
@@ -183,15 +125,20 @@ export class Group extends createCollectionMixin(
...
-    this.set({ ...options, angle: 0, skewX: 0, skewY: 0 });
+
     this.forEachObject((object) => {
       this.enterGroup(object, false);
     });
-    this._applyLayoutStrategy({
+
+    // perform initial layout
+    this.layoutManager = layoutManager;
+    this.layoutManager.performLayout({
```
asturur, line 135: can we change this in:

```js
layoutManager.target = this; // so we don't have to pass it each time
layoutManager.performLayout({ ... });
this.layoutManager = layoutManager;
```

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,294 @@
...
+    }
+    const nextCenter = new Point(result.centerX, result.centerY);
+    const correction = new Point(
+      result.correctionX ?? 0,
+      result.correctionY ?? 0
+    );
+    const relativeCorrection = new Point(
+      result.relativeCorrectionX ?? 0,
+      result.relativeCorrectionY ?? 0
+    );
```
asturur, line 150: this is verbose for no reason.
calcLayoutResult should return center, correction, relativeCorrection as points and we should use them right away instead of recreating them.

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,294 @@
...
+      // TODO: what about strokeWidth?
+      const origin = nextCenter.add(
+        new Point(width, height).multiply(
+          new Point(
+            resolveOrigin(target.originX),
+            resolveOrigin(target.originY)
+          )
+        )
+      );
+      target.set({ left: context.x ?? origin.x, top: context.y ?? origin.y });
```
asturur, line 198:
```suggestion
      const left = context.x ?? nextCenter.x + width * resolveOrigin(target.originX);
      const top = context.y ?? nextCenter.y + height * resolveOrigin(target.originY);
      target.set({ left, top });
```
is more concise, calculates only what we need and in a less expensive way and is equally readable

asturur:
https://github.com/fabricjs/fabric.js/pull/9352
this is the constant swap

src/LayoutManager/LayoutManager.ts
ShaMan123: true, but I want to move away from left/top

asturur: let's be clear.
Future plans not discussed do not count.
As of today we have left and top.
If we had x and y this suggestion would have been:

```js
    object.x += offset.x;
    object.y += offset.y
```

ShaMan123:
> * context.target in the various performLayout and other operations is always the group and is not supposed to change.
>   It should be a property of the layout manager rather than be passed every time? i ll comment right away in the code where it happens.

I don't mind... had in mind the Controls API that always passes around the object 

> onBeforeLayout and onAfterLayout methods on the group instances do not do anything since they are hooked up to be called but not to intervene on the process. Are as useful as an event and so should be deleted since we have the event already

For the first layout you will not be able to listen to the event
```
const group = new Group()
// ran layout in constructor
group.on('layout')


class G extends Group {
  onAfterLayout() {
     // runs in intialization and everytime
  }
}
```

src/LayoutManager/LayoutManager.ts
ShaMan123: true

src/LayoutManager/LayoutManager.ts
ShaMan123: for now I will live with it

asturur:
> > * context.target in the various performLayout and other operations is always the group and is not supposed to change.
> >   It should be a property of the layout manager rather than be passed every time? i ll comment right away in the code where it happens.
> 
> I don't mind... had in mind the Controls API that always passes around the object
> 
> > onBeforeLayout and onAfterLayout methods on the group instances do not do anything since they are hooked up to be called but not to intervene on the process. Are as useful as an event and so should be deleted since we have the event already
> 
> For the first layout you will not be able to listen to the event
> 
> ```
> // runs layout in constructor before event subscribed
> const group = new Group()
> group.on('layout')
> 
> 
> 
> class G extends Group {
>   onAfterLayout() {
>      // runs in intialization and everytime
>   }
> }
> ```

The first layout is not an event indeed.
You create the group with your own code, no even needed you know is layouting, do what you want to do in your function that creates the group.
The active selection is one and if you act by clicks you are using always the same one so is always not the first layout.
Still convinced that we don't need to put that in for use cases that are not an issue.

asturur:
> * context.target in the various performLayout and other operations is always the group and is not supposed to change.
>   It should be a property of the layout manager rather than be passed every time? i ll comment right away in the code where it happens.

The control api wasn't mean to be attached to an object.
It build on the idea of controls sets you would keep around  which all objects would reference.
Is there a safe way to use the LayoutManager in the same way? in that case we can put it in the groupDefaults rather than passing it in the constructor.

ShaMan123:
> * array.flat() can't be used since is supported only by recent browsers and none of the ones we list as supported:
>   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat
>   So we need to concat those results differently.

shouldn't the build transform this? Anyways it is not a big deal. I'll remove

src/LayoutManager/LayoutManager.ts
ShaMan123: I dislike it but OK

ShaMan123:
> > * context.target in the various performLayout and other operations is always the group and is not supposed to change.
> >   It should be a property of the layout manager rather than be passed every time? i ll comment right away in the code where it happens.
> 
> The control api wasn't mean to be attached to an object. It build on the idea of controls sets you would keep around which all objects would reference. Is there a safe way to use the LayoutManager in the same way? in that case we can put it in the groupDefaults rather than passing it in the constructor.

I designed the LayoutManager to be independant from Group as much as possible.
So it is possible.
It is a design question.
Should it be like that? I am not against. I want to hear it out.

> in that case we can put it in the groupDefaults rather than passing it in the constructor.

If we do it as you suggested it can become an optional property and then the Group will do a single layout on intialization and drop the ref

ShaMan123 (commented):
rm `onBeforeLayout`, `onAfterLayout` hooks
renamed `layout` event to `layout:after`

ShaMan123:
> in that case we can put it in the groupDefaults rather than passing it in the constructor.

Anyways I do not agree to put an instance in the defaults. Then it is shared and the whole nightmare of the shared instance is back.
I agree to have it undefined by default or not. And then someone with an urge of a shared instance can pass it to the constructor. This is possible aready as it is now so I think I will not touch it.

ShaMan123:
I was just think I have a different approach without using correction, rather storing all oject center points in the canvas plane before layout and reapplying them after layout. Same concept as with the position handler of the control API

src/LayoutManager/LayoutStrategies/FixedLayout.ts
```diff
@@ -16,14 +17,16 @@ export class FixedLayout extends LayoutStrategy {
...
+        .multiply(new Point(-resolveOrigin(originX), -resolveOrigin(originY)))
+        .transform(context.target.calcOwnMatrix(), true);
 
       return {
         ...result,
-        center: result.center.add(
-          new Point(width, height).subtract(result.size).scalarDivide(2)
-        ),
+        // center: result.center.add(sizeCorrection),
+        // correction: new Point(result.correction).add(sizeCorrection),
```
ShaMan123, line 29: I am struggling with supporting a fixed size

ShaMan123 (commented):
separated the layout manager from the group

asturur:
> > * array.flat() can't be used since is supported only by recent browsers and none of the ones we list as supported:
> >   https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat
> >   So we need to concat those results differently.
> 
> shouldn't the build transform this? Anyways it is not a big deal. I'll remove

no this would be a polyfill and not a syntax issue. We don't have polyfills so far, we avoid what we don't support

src/shapes/Group.ts
```diff
@@ -133,7 +131,13 @@ export class Group
 
     // perform initial layout
     this.layoutManager = layoutManager;
-    this.layoutManager.performLayout({
+    (
+      layoutManager ||
+      // legacy
+      new LayoutManager(
+        this.clipPath ? new ClipPathLayout() : new FixedLayout()
+      )
```
asturur: i don't understand this change here

asturur:
I don't understand some changes in eac81fb396500674ffaf3f0c35d74c53b6f4ac74

asturur:
i think i have to relook at everything from top down again.

Regarding the first layout lost issue, is also possible do like we do with other events, we also fire a canvas level event with the target of the event set to the group, so you are already registered
Or the event could be at canvas level only too if we don't want to have 2

src/LayoutManager/LayoutStrategies/FixedLayout.ts
asturur: But this is a newly discovered bug or something that came out because of changes requested?

src/shapes/Group.ts
ShaMan123: aparently some tests were failing because I made FitContentLayout not respect width/height of constructor options
Add to that that `layoutManager` is now optional (which is a good decision IMO) then use this fallback for the initial layout

ShaMan123: But this part of the initial fixed layout is still not stable

ShaMan123:
I wish you would comment so at least I understand where you are at.
Even if we decide to continue.
Better to do it right and not drag bad design forward.
I didn't dig into my suggestion but it seems to be possible with the current API an da few overrides.
Meaning, it can become a layout strategy that can replace the existing one if need be.

> I don't understand some changes in [eac81fb](https://github.com/fabricjs/fabric.js/commit/eac81fb396500674ffaf3f0c35d74c53b6f4ac74)

What don't you understand?

ShaMan123 (commented):
seems stable
the enum thing is annoying

asturur:
i thought about the enum then and i don't think is good.
If is typed properly the enum is just overhead.
We are not hiding values like '0', '1' behind user friendly names 'types.initialize' is '0' and 'types.layout' is '1' for example.
We still need user friendly values for developers, so enum simply are not the tool for this job.

jiayihu:
Jumping in to just say that indeed enums are not well seen in the TS community because:
- Inconvenient to use for library consumers as it needs to be imported
- Are not type-only, as everything else in TS. They have runtime effects most people are unaware, especially const enums

For internal use however they're more okayish, as they make easy to refactor the values or use a different semantic name, while keeping the same runtime value

asturur:
> Jumping in to just say that indeed enums are not well seen in the TS community because:
> 
> * Inconvenient to use for library consumers as it needs to be imported
> * Are not type-only, as everything else in TS. They have runtime effects most people are unaware, especially const enums
> 
> For internal use however they're more okayish, as they make easy to refactor the values or use a different semantic name, while keeping the same runtime value

But that is useful when the runtime value is either not under your control ( an external api ) or really unfriendly to remember ( a bit mask 1,2,4,8,16,32,64 for example ). All i see being done across projects i work on is:

```
enum composite { 
  sourceOver: 'source-over',
  darken: 'darken',
}
```

And i find it a misuse

asturur:
i m doing the const thing and reviewing again to see if we can go.
I have some issues with calcbounds using getTransformedDimensions but that i will leave as last thing.

asturur:
> I wish you would comment so at least I understand where you are at. Even if we decide to continue. Better to do it right and not drag bad design forward. I didn't dig into my suggestion but it seems to be possible with the current API an da few overrides. Meaning, it can become a layout strategy that can replace the existing one if need be.
> 
> > I don't understand some changes in [eac81fb](https://github.com/fabricjs/fabric.js/commit/eac81fb396500674ffaf3f0c35d74c53b6f4ac74)
> 
> What don't you understand?

I don't have much to comment, i pointed out all the things i wanted to change and we changed them. I left the bounds for last because is more annoying but is very focused in one file only.

There are things i wish weren't there like mandatory registration of events for every static non interactive group but i m not sure i have a solution for that right now. Unless i don't get an idea that i know would cover 90% of the work i m not putting it on the table because the scope of this PR is extracting from group in a dedicated place and not to rethink all the system from scratch.

ShaMan123:
I understand that point
But since the layout manager is undefined by default creating a group will behave more as it did in v5.
If you want it to layout continuously you pass a layout manager.
And always you can override

src/shapes/Group.ts
asturur: in general we use a lot of internal .on/off.
We should just have a sort of _on and _off that we can setup and can't be broken from outside doing a simple off() that kills everything

ShaMan123: True
Specifically fir group I think these can be removed

src/LayoutManager/types.ts
```diff
@@ -0,0 +1,148 @@
...
+  prevCenter: Point;
+  nextCenter: Point;
+  /**
+   * The vector used to offset objects by, as measured by the plane
+   */
+  offset: Point;
+};
+
+type ImperativeLayoutCommonOptions = {
+  overrides?: LayoutStrategyResult;
```
asturur, line 66: While i look at calcBounding box can you explain me what overrides is doing?
Seems like something put there but fabricJS is not using and there is no explanation on how could be used and from who.

ShaMan123: It forces the layout the use these values instead of the calculated ones

src/LayoutManager/LayoutStrategies/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  planeChange?: TMat2D
+) => {
+  const decomposition = planeChange
+    ? qrDecompose(
+        multiplyTransformMatrices(planeChange, object.calcOwnMatrix())
+      )
+    : undefined;
+  const sizeVector = object
+    ._getTransformedDimensions(decomposition)
```
asturur, line 27: One part of my issue with the calcBounding box starts from here.
_getTransformedDimensions should not have options and was a mistake to put them there.
The options object is an any and in this case we are putting down a full decomposition, including an angle.
In the case this happens the _getTransformedDimensions will 'secretly' push the angle down to recompose a matrix that will factor in both the planeChange angle and the object angle, the bounding box will be rotated.

Later down it seems like  we want to compensate this angle we get from sizeVector but we do that only with object.angle

In the case of a planeChange the box includes both angles
in the case without planeChange the box includes NO angles at all

But if the object has an angle we compensate it anyway.

This seems a bug lurking to me.

Can you try to explain why are we compensating the object angle in general? i imagined that it was because we want the bounding box of the object aligned with the group angle, but those calculations do not seems to reflect that

ShaMan123: There is a bug in this code and I left a comment for.
_getTransformedDimensions is a bad impk regardless of the options.
It can't move planes once calculated, that is why I had tonadd the options.
I will look into what you mentioned but I want to say clearly that this is not worth it.
It must be rewritten, stroke calc is wrong. That is how I started with redoing coords.
Put aside what you thibk of it, it fixes this very issue

src/LayoutManager/LayoutManager.spec.ts
```diff
@@ -309,7 +319,7 @@ describe('Layout Manager', () => {
   describe('commitLayout', () => {
     const prepareTest = (
       contextOptions: {
-        type: 'initialization' | 'added';
+        type: INITIALIZATION | ADDED;
```
ShaMan123, line 322:
```suggestion
        type: typeof INITIALIZATION | typeof ADDED;
```

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: I read the code again, what do you mean about the angle?
I am not attempting to fix this also because stroke uniform is broken all across fabric so I have no way to reliably measure objects w/o #8767

asturur: the angle compensation code is hidden behind 
```
qrDecomposition?.angle ?? object.angle
```

that means if we have a qrDecomposition it triggers if the angle of the qr decomposition is not 0, otherwise it doesn't
if the angle of qrDecomposition is different than 0 then we compensate by object.angle.

But this qrDecomposition angle can be different than 0 for a series of reasons:
- planeChange has an angle and object doesn't
- planeChange and object they both have an angle
- planeChange does not have an angle and the object has

OR

they could be both have an angle that compensate to 0, and the plane change includes the angle from both groups that create the the plane change matrix.

I m saying it seems strange that if any angle is present, the compensation has to happen with the object angle.
I wish there were a set of use cases to try

asturur: > There is a bug in this code and I left a comment for.

Do you remember where this comment his? or what the bug is?

asturur: i split the code in this example for better discussion

```ts
const getObjectSizeVector = (object: FabricObject) => {
  const sizeVector = object._getTransformedDimensions().scalarDivide(2);
  if (object.angle) {
    return compensateAngle(object.angle, sizeVector);
  }
  return sizeVector;
};
```

```ts
const getObjectSizeVectorWithPlaneChange = (
  object: FabricObject,
  /**
   * account for selected objects in active selection
   */
  planeChange: TMat2D
) => {
  const decomposition = qrDecompose(
    multiplyTransformMatrices(planeChange, object.calcOwnMatrix())
  );
  const sizeVector = object
    ._getTransformedDimensions(decomposition)
    .scalarDivide(2);
  if (decomposition.angle !== 0) {
    return compensateAngle(object.angle, sizeVector);
  }
  return sizeVector;
};
```

this is how i read the 2 different codepath.
The part in which we do 
```
  if (decomposition.angle !== 0) {
    return compensateAngle(object.angle, sizeVector);
  }
```
is unclear to me.

The first question i made and of which i m not sure yet, is if this code is supposed to counter rotate the size vector to repsect the current group angle

asturur: In the case with no plane change, you get sizeVector that is not influenced by object angle, and then you eventually apply/remove object angle from it.

in the case with the plane change, you get size vector that is influenced by all the angles, group, target group and object itself and then only if the resulting angle from all the angles is different from 0, you apply /remove object angle from it.

In case 2 the size vector is already influenced by object angle in some way

asturur: image

There is something that i don't get on how this calculation wants to work. It tries to reduce to 2 points something that before we were doing with 4 points.
I tried to add some debug code to rendering in order to visualize what happens and i see that while the result is correct, the logic behind it is unclear.

src/LayoutManager/LayoutStrategies/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+    .scalarDivide(2);
+  if (decomposition?.angle ?? object.angle) {
+    const rad = degreesToRadians(object.angle),
+      sine = Math.abs(sin(rad)),
+      cosine = Math.abs(cos(rad)),
+      rx = sizeVector.x * cosine + sizeVector.y * sine,
+      ry = sizeVector.x * sine + sizeVector.y * cosine;
```
asturur, line 34: this is not a rotation, what kind of transformation it is?
Is a projection of what kind?

src/LayoutManager/LayoutManager.ts
```diff
@@ -0,0 +1,299 @@
...
+  LAYOUT_TYPE_REMOVED,
+  LAYOUT_TYPE_IMPERATIVE,
+  LAYOUT_TYPE_OBJECT_MODIFIED,
+  LAYOUT_TYPE_OBJECT_MODIFYING,
+} from './constants';
+import type { LayoutContext, LayoutResult, StrictLayoutContext } from './types';
+
+export class LayoutManager {
+  private _prevLayoutStrategy?: LayoutStrategy;
+  private _subscriptions: Map<FabricObject, VoidFunction[]>;
```
asturur, line 22: this should be a WeakMap or we may create memory leaks of some sort.
If the object is used as a key i have no idea if that counts as a reference for the GC, but a WeakMap is probably safer

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: Always we have the issue of strokeUniform I didn't address here because I don't think it is stable enough in the rest of the code

asturur: yeah forget strokeUniform.
_getTransformedDimensions does account for object angle if used improperly.
What i am pointing out is that in one case it is accounting for angle, and for another it isn't.

When you do `getTransformedDimensions(decomposition)` is accounting for the decomposition angle, object and plane change matrix. while when you do `getTransformedDimensions()` it is not.

The issue lies in the destructuring of the options here:

```
  _getTransformedDimensions(options: any = {}): Point {
    const dimOptions = {
      scaleX: this.scaleX,
      scaleY: this.scaleY,
      skewX: this.skewX,
      skewY: this.skewY,
      width: this.width,
      height: this.height,
      strokeWidth: this.strokeWidth,
      ...options,
    };
```
the angle from options is being put back in dimOptions, that then is used to transform the points.

And i wouldn't even be worried about that, the fact is that we are doing that only in one case of 3.

This is complicated and i wanted to talk about it. i got a grasp of this PR since  i m looking at it since long, and there are things i want to talk about

src/LayoutManager/LayoutStrategies/utils.ts
asturur: this have to explain me, i really don't get it why this is here.

src/LayoutManager/LayoutStrategies/utils.ts
asturur: But yes in general i really prefer sticking to the 4 point system to create the bounding box rather than those 2 points. The main issue is that i can't relate the code to anything i know about algebra/geometry and so i tend to prefer something that i think is more common knowledge

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: bbox

Take a look here, this is the math behind the bbox transform which takes a point from the rotated plane and transforms it to the non rotated plane.
`compenstaeAngle` which is a wrong name is an application of the bbox transform matrix

`α = object.angle`

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: > the angle from options is being put back in dimOptions, that then is used to transform the points.
And i wouldn't even be worried about that, the fact is that we are doing that only in one case of 3.

I don't get this. Let's talk.

ShaMan123: `dimOptions` does not do anything with the angle

asturur: ok ok is true, dimOptions takes the angle but then it creates a matrix that doesn't take in account of the angle. i was sure it did.

The picture is explanative and needs to go in the docs somehow. i ll commit it to this PR. What did you use to draw it?

compensate angle is for sure a wrong name, i knew when it was all addition that it wasn't a rotation, but here and in the original PR there wasn't an explanation for that logic, and i couldn't make one up.

I still think the 4 points are easier to manage, but i still do not understand, what is the scope of this calculation. It get me the bounding box, and if you don't say anything else i would argue is the same as getBoundingRect(absolute=true)? So i was asking if this bounding box is actually counter rotated as the group, but i can't see that in the code.

On top of that I still think there is an issue, because if the group is rotated -30 and the object is rotated 30, the decomposition has an angle of 0, but the object is rotated anyway, doesn't it count in that case?

Maybe i have to try how this behave when adding an object to a rotated group and how it gets there.

ShaMan123: I used to kitchensink - it was too tough
I can commit svg to the readme but then I will make it nicer
I don't mind, but we can expose a planechange matrix with a good name and then it become useful and comprehensive

src/LayoutManager/LayoutManager.ts
ShaMan123: I am not sure this is wise
How will we be able to dispose/unsubscribe?

src/LayoutManager/LayoutStrategies/utils.ts
ShaMan123: though this is correct something was wrong  in the code, refactored to a simpler solution

src/LayoutManager/LayoutManager.ts
ShaMan123: and fixed strokeUniform in group layout
The solution can be used all across fabric to fix stroke uniform issue and remove the wrong `_getTransformedDimensions`

asturur: a weakMap is still a map, it just does not hold the object in memory if the object gets sent away.
Is not a valid reference to stop garbage collector

asturur:
the change on the bounding is definetely better becuase it keeps the 4 points.
I wanted to do something really similar with a wrapper for strokeUniform.

Do you know why those unit tests with the array transform matrix changed?

src/Point.ts
```diff
@@ -385,4 +385,4 @@ export class Point implements XY {
   }
 }
 
-const originZero = new Point(0, 0);
+export const ZERO = new Point(0, 0);
```
asturur, line 388: this has never to become a public export.

asturur:
i will add more docs and then merge.
There are things i want to change in general about this new functionalities ( group layouts and group interactivity ), but i won't touch anything till we can sync.

src/LayoutManager/LayoutStrategies/utils.ts
```diff
@@ -1,56 +1,53 @@
...
+  const objectCenter = t
+    ? object.getRelativeCenterPoint().transform(t)
     : object.getRelativeCenterPoint();
-  const sizeVector = getObjectSizeVector(object, planeChange);
-  return [objCenter.subtract(sizeVector), objCenter.add(sizeVector)];
+  const strokeUniformVector = strokeUniform
+    ? sendVectorToPlane(
+        new Point(strokeWidth, strokeWidth),
+        undefined,
+        destinationGroup.calcTransformMatrix()
```
ShaMan123, line 41: this can't be right

asturur:
I have some changes i have proposed in a branch and then there would be some followup to change some code for handlign the `fromObject` case.

I m not sure this PR is solving 1) tho.
We don't user case with code we can run over.

In theory we should be working just on this PR and i should insta-close any other PR, but then that never happen and this get pushed back

ShaMan123:
`_getTransformedDimensions` is to blame for strokeUniform and bbox
The error is easy to see, the storke uniform vector is relative to the parent instead of to the canvas.
Now the geometry is clean I can approach a partial fix. Note that all control action handlers use this ***** method and moving them away is an effort

ShaMan123:
The bright side is that I think that will end all strokeUniform bugs all across fabric making it a stable feature

asturur:
ok i patched the fromObject function to get this merged, fromObject and toObject is what i want to follow up on.
If is all green i m merging

asturur (approved):

PR #9161, opened 2023-08-18
fix(): border rendering with padding under group

ShaMan123:
## Motivation

Rendering an active object under group with padding and found a bug

## Description

Look at the diff - self explanatory


ShaMan123 (commented):
ready to merge
@asturur  if you can make it enter beta11 that will be great

ShaMan123 (commented):
added another test to make sure padding is not affected by group transform

src/shapes/Object/InteractiveObject.ts
```diff
@@ -462,7 +462,10 @@ export class InteractiveFabricObject<
               // and is the qrDecompose of a matrix that takes in account zoom too
               new Point(options.scaleX, options.scaleY)
         ).scalarMultiply(this.strokeWidth);
-      size = bbox.add(stroke).scalarAdd(this.borderScaleFactor);
+      size = bbox
+        .add(stroke)
+        .scalarAdd(this.borderScaleFactor)
+        .scalarAdd(this.padding * 2);
```
asturur, line 468: padding does not change with zoom. Is another PITA like stroke uniform.
Which is the CTX state here? is it multiplied already by the viewport transform? or is it scaled back to neutral?

Can you please add a visual test like you did but with also a bunch of zoom?

asturur: i will publish another release after this fix.

asturur: Also, just to be clear, padding shouldn't enter in the layout of the group, so if an object inside the group has padding 2000, it will blow out of the layout when selected and that is fine.

ShaMan123: I have added a test with scaling, did you look at it?
It does not affect group layout

ShaMan123: > Which is the CTX state here? 

Same as for rendering controls

ShaMan123: actually I have to say I think we should use the coords to render the border but that is not related to here

ShaMan123: Ahh I missed vpt, sorry

ShaMan123:
Yes

PR #9164, opened 2023-08-18
test(): fix snapshots by removing version

ShaMan123:
## Description

beta11 exposed this issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [next]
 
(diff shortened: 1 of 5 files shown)


asturur:
merging this.
I see strange things in the snapshot diffs, maybe is just github.
I was trying to regenerate the snapshot but for some reason my new laptop doesn't want to install canvas at all, i think because i don't have an intel arch, so i have to compile it by myself.

asturur (approved):

PR #9165, opened 2023-08-19
CD(): website submodule

ShaMan123:
## Motivation

have the website repo visible

## Description

`git submodule add https://github.com/fabricjs/fabricjs.github.io website`

## In Action

image


ShaMan123:
I wonder if there is a way not to point to a specific commit

ShaMan123:
reading https://stackoverflow.com/questions/8191299/update-a-submodule-to-the-latest-commit seems it must be updated...
we can add an action for that

asturur:
i think it needs to be updated because submodules are meant to be imported as code dependencies?

asturur:
What we do with it, what does it enable?

asturur:
we can have it of course, but we need to write an action to write a commit everytime we want to update it, no thanks.

asturur (approved):

PR #9167, opened 2023-08-21
feat(`perPixelTargetFind`): support `not-selected` option

ShaMan123:
## Motivation

I was looking into this and it doesn't make sense IMO.

If an object is selected I think the user expectation is to iteract with the bbox, whereas if it is not selected `perPixelTargetFind` should be respected

## Description

added an option `not-selected` to `perPixelTargetFind`


jiayihu:
This sounds a strong assumption on the desired UX. Fabric should probably be agnostic about it, leaving the decision to the application. Maybe fabric should just tell that both the children and the activeObject are targets of the event.

ShaMan123:
> Maybe fabric should just tell that both the children and the activeObject are targets of the event.

Don't get it

ShaMan123:
if we turn the usage from a flag check to a method it can be easily overridden

```
object.isPerPixelTargetFind()
```

jiayihu:
> Don't get it

I mean that `_searchPossibleTargets` maybe should return a list of targets, instead of only one. Then depending on the case, evaluate what to do if multiple targets are detected, leaving possibly some decisions to the application

> if we turn the usage from a flag check to a method it can be easily overridden

In that case you can already override the property with a getter

ShaMan123:
We talked about refactoring `_searchPossibleTargets` to return context instead of setting `targets` on canvas etc. => that should be done

asturur:
i remember when we needed it years ago, was mostly for text, people would see things behind text objects and they were angry they couldn't  be selected if text was selected.

It could be an option in 6.1, now this shouldn't be touched.

src/canvas/SelectableCanvas.ts
```diff
@@ -772,17 +771,11 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     obj: FabricObject,
     globalPointer: Point
   ): boolean {
-    if (
-      obj &&
-      obj.visible &&
-      obj.evented &&
-      // http://www.geog.ubc.ca/courses/klink/gis.notes/ncgia/u32.html
-      // http://idav.ucdavis.edu/~okreylos/TAship/Spring2000/PointInPolygon.html
```
asturur, line 780: we should preserve this comment maybe over the containsPoint function

src/canvas/SelectableCanvas.ts
```diff
@@ -772,17 +771,11 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
-      obj.containsPoint(pointer)
-    ) {
+    if (obj && obj.visible && obj.evented && obj.containsPoint(pointer)) {
       if (
-        (this.perPixelTargetFind || obj.perPixelTargetFind) &&
-        !(obj as unknown as IText).isEditing
+        // if `obj` is selected we want it to be easy to interact with it
```
asturur, line 776: i can see both ways being reasonable.
this will be a boolean option maybe called `preferBoundingBoxWhenActive: boolean` to add.

ShaMan123: IMO options less booleans more deleagtes/managers

src/canvas/SelectableCanvas.ts
ShaMan123: both are dead, and also the impl has moved forward more than a decade

asturur:
> We talked about refactoring `_searchPossibleTargets` to return context instead of setting `targets` on canvas etc. => that should be done

In this way we take away from simplicity for the purpose of trying to write the perfect flexible code that doesn't exist.

You were about to change an IF only, that means that a boolean option is fine.

ShaMan123:
So if going in the direction I would change `perPixelTargetFind: boolean | 'not-selected'`

src/canvas/SelectableCanvas.ts
```diff
@@ -772,18 +774,19 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
-        }
-      } else {
-        return true;
-      }
+      const shouldPerformPixelFind =
+        !(obj as unknown as IText).isEditing &&
+        (this.perPixelTargetFind === true ||
+          obj.perPixelTargetFind === true ||
+          ((this.perPixelTargetFind === 'not-selected' ||
+            obj.perPixelTargetFind === 'not-selected') &&
```
ShaMan123, line 782: I am not sure about combinations between `canvas.perPixelTargetFind` and `obj.perPixelTargetFind`
Shouldn't the object take presedence?
I will update tests once we decide

ShaMan123 (commented):
added `not-selected` option

ShaMan123:
I am not sure about this PR.
Perhaps it is best to expose a method on object instead of the new option.

src/canvas/SelectableCanvas.ts
asturur: We should collect the current rules around the code.
But there shouldn't be such precedence is just confusing.
There must a be a documented way to handle that for all objects or just some.
If you can set it on a single object or an entire subclass through default values is good enough imho

asturur:
how this would behave when you have selected an object inside a group with not-selected? what is the ux like?

ShaMan123:
I think we should close this pr.
And jnstead move the logic to containsPoint.
Then each object can override at will and no confusing ux.
Thoughts?

asturur:
my question on the ux is still useful to understand what this features would do.

contains point won't be the center of the selection anymore since it doesn't take padding in account. containsPoint can be used for selection with a pointer, but is not a selection function so it should not have selection logic inside. If a point is contained it has to return true, regardless of selection ux.

asturur:
i still want to know this if you know:

how this would behave when you have selected an object inside a group with 'not-selected'? what is the ux like?

PR #9169, opened 2023-08-22
ci(): generate docs

ShaMan123:
## Description

Added some config + patched some types for typedoc to work
It looks promising - tried to put it on the website but seems the md files output something that gatsby is not happy about
If you remove the plugin it outputs a standalone website rather than md files

We can add an action to the publish action that somehow generates docs into the website

## Gist

```bash
npm run docs
```


typedoc.json
```diff
@@ -0,0 +1,10 @@
+{
+  // https://typedoc.org/options/
+  "$schema": "https://typedoc.org/schema.json",
+  "entryPoints": ["fabric.ts"],
```
ShaMan123, line 4:
```suggestion
   "entryPoints": ["index.ts", "index.node.ts"],
```

support node entry?

ShaMan123 (commented):
this works

typedoc.json
asturur: let's start with one and let's see how it works

asturur (approved):

PR #9172, opened 2023-08-23
Force users to use fabricObject instead of object

zhe-he:
Object is a keyword. Force users to use FabricObject instead of Object

### DIFF
--- a/fabric.ts
+++ b/fabric.ts
@@ -41,7 +41,11 @@ export { CircleBrush } from './src/brushes/CircleBrush';
 export { SprayBrush } from './src/brushes/SprayBrush';
 export { PatternBrush } from './src/brushes/PatternBrush';
 
-export { FabricObject as Object } from './src/shapes/Object/FabricObject';
+import { FabricObject } from './src/shapes/Object/FabricObject';
+/** @deprecated Object is a keyword. Please use FabricObject instead */
+class Object extends FabricObject { };
+export { FabricObject, Object };
+
 export type {
   TFabricObjectProps,
   FabricObjectProps,


ShaMan123:
we have discussed this.
Image and Text are the same
you should do what you did when you import fabric
But I am not against renaming

fabric.ts
```diff
@@ -41,7 +41,11 @@ export { CircleBrush } from './src/brushes/CircleBrush';
 export { SprayBrush } from './src/brushes/SprayBrush';
 export { PatternBrush } from './src/brushes/PatternBrush';
 
-export { FabricObject as Object } from './src/shapes/Object/FabricObject';
+import { FabricObject } from './src/shapes/Object/FabricObject';
+/** @deprecated Object is a keyword. Please use FabricObject instead */
+class Object extends FabricObject { };
```
jiayihu, line 46: Should be `const Object = FabricObject` otherwise this adds a runtime class level in the hierarchy

ShaMan123: Could you also?
```
export { FabricObject, FabricObject as Object }
```

ShaMan123:
I suggest we hear from @asturur 
The problem might be inconsistency.
Now we have FabricText and IText - weird
After that you should make tests pass

src/shapes/Image.ts
```diff
@@ -856,5 +851,8 @@ export class Image<
   }
 }
 
-classRegistry.setClass(Image);
-classRegistry.setSVGClass(Image);
+classRegistry.setClass(FabricImage);
+classRegistry.setSVGClass(FabricImage);
+
+/** @deprecated Image is a keyword. Please use FabricImage instead */
```
asturur, line 857: i would extend this:

WARNING the old fabric.Image class can't be imported as Image because of a reserved word clash.
For this reason it has been renamed to FabricImage. Please use FabricImage in place of Image.

src/shapes/Image.ts
```diff
@@ -856,5 +851,8 @@ export class Image<
   }
 }
 
-classRegistry.setClass(Image);
-classRegistry.setSVGClass(Image);
+classRegistry.setClass(FabricImage);
+classRegistry.setSVGClass(FabricImage);
```
asturur, line 855: What happens to JSON exports old and new? <--- this is important.

zhe-he:
There are not many names left for us. If for brevity, can we just add F in front?
Image -> FImage ?
Text -> FText ?
Because we still have Itext TextBox, if the text is renamed a lot, how can the two be compatible?
I don't have any great ideas, if you guys have a better name that can accommodate Itext, textbox at the same time, please let me know and I will submit

ShaMan123:
I do not like it but we can rename to StaticText like StaticCanvas
Object can become BaseObject
Image... Img
All very ugly

asturur:
if we don't have ideas FabricText and FabricImage are all ok.
But still i don't know, what happens to FabricImage.type ? what happens to exports and registry?
in case we have to change the getter and fix the registry.

ShaMan123:
We moved away from type being the class name because of minifying, right?
So it shouldn't be a problem, should it?
It has the same `static type = 'type'` so it should be fine.
But tests are failing

asturur:
anyway as soon as test passes we are good to go since we don't have groundbreaking names

asturur:
Gave a stab at fixing tests

asturur:
If you are ok with class names this is fine for me.
I extended the deprecation notice to be as clear as possible.

asturur (dismissed):

asturur (approved):

asturur (approved):

src/shapes/Image.ts
```diff
@@ -856,5 +856,13 @@ export class Image<
...
+
+/**
+ * @deprecated The old fabric.Image class can't be imported as Image because of conflict with Web API.
+ * https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image
+ * For this reason it has been renamed to FabricImage.
+ * Please use `import { FabricImage }` in place of `import { Image as FabricImage }`
+ */
+export class Image extends FabricImage {}
```
ShaMan123, line 868: I think these 3 should reside in the main `fabric.ts` file and not in the def files

ShaMan123 (commented):
Do we want to rename the files as well?

asturur:
> Do we want to rename the files as well?

for me no, i just learned the order of the new files in the tree

asturur (approved):

asturur:
ok do not know how to fix jest now

zhe-he:
it's OK now for jest.

asturur (dismissed):

asturur (approved):

asturur:
Leaving this to you guys, you had some open topic on those names.

asturur:
Is there any other decision to make on this PR? or is good as it is?

asturur (approved):

asturur:
@zhe-he @ShaMan123 i don't understand if you still have some internal discussion here or if we are done

ShaMan123:
Nope

asturur (approved):

asturur (approved):

ragul0417:
activeObject._findTargetCorner is not a function
TypeError: activeObject._findTargetCorner is not a function
    at klass.findTarget (http://localhost:3000/static/js/bundle.js:236207:47)
    at klass._onMouseEnter (http://localhost:3000/static/js/bundle.js:236908:44)
ERROR
activeObject._findTargetCorner is not a function
TypeError: activeObject._findTargetCorner is not a function
    at klass.findTarget (http://localhost:3000/static/js/bundle.js:236207:47)
    at klass._cacheTransformEventData (http://localhost:3000/static/js/bundle.js:237424:84)
    at klass.__onMouseMove (http://localhost:3000/static/js/bundle.js:237448:12)
    at klass._onMouseMove (http://localhost:3000/static/js/bundle.js:237096:12)
    
    how to solve this error it will occur on while update a image in fabric

PR #9175, opened 2023-08-25
feat(Canvas): provide correctly named canvas pointer methods. BREAKING: rm `_normalizePointer`, `restorePointerVpt`

ShaMan123:
## Motivation

This PR adds 2 new methods to get a point from a mouse event and tries to remove the confusion about pointer and absolutePointer.

After much discussion we ended up renaming:

-scenePoint (before absolutePointer) a point that is the fabricCanvas plane, the one in which you measure object.top and object.left of an object added in the canvas.

-viewportPoint (before pointer) a point in the viewport coordinate, for which the point 0,0 is the top left corner of the canvas container, while canvas.width, canvas.height is the bottom right corner of the canvas container ( because of retina scaling i avoid on purpose of talking about the HTMLCanvasElement )

`canvas.getPointer(event, ignoreVpt)` becomes `canvas.getPointer(event, fromViewport)` and becomes **deprecated**, will be moved to protected or private next major version.
Developers are directed to use:
`canvas.getViewportPoint(event)` for `canvas.getPointer(event, true)`
and
`canvas.getScenePoint(event)` for `canvas.getPointer(event, false)`

All the events that were emitted with pointer and absolutePointer still are doing so, those event properties are deprecated and the new `scenePoint` takes the place of absolutePointer and `viewportPoint` takes the place of pointer.

`restorePointerVpt` has been removed and also `transformPointRelativeToCanvas` in favor of `sendPointToPlane`


ShaMan123 (commented):
OMG that slipped under the radar
@jiayihu could this be the reason for the very strange outputs in tests?

ShaMan123 (commented):
@asturur this is critical before v6
I imagine we should add tests

src/canvas/Canvas.ts
```diff
@@ -402,7 +402,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
     this.targets = [];
     const target = this._searchPossibleTargets(
       this._objects,
-      this.getPointer(e, true)
+      this.getHTMLPointFromEvent(e)
```
jiayihu, line 405: Weird, why do you use the HTML point to search the targets?

src/canvas/SelectableCanvas.ts
```diff
@@ -854,38 +839,67 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+   * @example
+   * const pointInHTMLPlane = sendPointToPlane(
+   *  this.getCanvasPointFromEvent(e),
+   *  canvas.viewportTransform
+   * );
+   *
+   */
+  getCanvasPointFromEvent(e: TPointerEvent) {
+    if (this._absolutePointer) {
+      return this._absolutePointer;
```
jiayihu, line 877: Risky business IMO, the method sounds a pure calculation but this could return some stale state

jiayihu: Also this is also handled internally in `getPointer`

jiayihu:
> I imagine we should add tests

Yes, tests for something so important should be there. This is the base of many interactions an application can build on top of the canvas.

src/canvas/SelectableCanvas.ts
ShaMan123: I forgot to remove from there
but yes, caching here seems...

ShaMan123: however all events run or at least should run `_cacheTransformEventData` and `_resetTransformEventData`
I agree it is far from perfect

src/canvas/Canvas.ts
ShaMan123: Now it is visible that there are very strange usages and some buggy ones

ShaMan123 (commented):
the sanpshot show clearly we need to expose a method that prepares the event data for consistency

ShaMan123 (commented):
event data is TESTED
This PR can merge

jest.setup.ts
```diff
@@ -0,0 +1,4 @@
+import { setEnv, getEnv } from './src/env';
+
+// set custom env
+beforeAll(() => setEnv({ ...getEnv(), window, document }));
```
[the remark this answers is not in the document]
ShaMan123: this is also good for devs wanting to configure jest + fabric
now they can see how

src/canvas/SelectableCanvas.ts
```diff
@@ -812,8 +791,13 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
       const objToCheck = objects[i];
+      // this strange logic indicates a severe bug in containsPoint
       const pointerToUse = objToCheck.group
-        ? this._normalizePointer(objToCheck.group, pointer)
+        ? sendPointToPlane(
+            pointer,
+            invertTransform(this.viewportTransform),
+            objToCheck.group.calcTransformMatrix()
+          )
         : pointer;
```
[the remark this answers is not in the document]
asturur: How the bug manifest? The fact that the code is not your liking doesn't mean is bugged.

asturur:
Please add a description showcase of the bug. What the bug was? how did you find it?
containspoint was doing stuff that you assumed was other stuff, and now we scream BUG BUG BUG MESS MESS MESS the code is not understandable.
The truth is that all the changes that have been done in the name of improving something else changed the meaning of those functions, containsPoint was strictly used for checking targets on mouse.

src/canvas/SelectableCanvas.ts
```diff
@@ -854,41 +839,62 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+   * however, from the viewer's perspective the point is changed.
+   *
+   * @example
+   * const pointInHTMLPlane = sendPointToPlane(
+   *  this.getCanvasPointFromEvent(e),
+   *  canvas.viewportTransform
+   * );
+   *
    */
-  getPointer(e: TPointerEvent, ignoreVpt = false): Point {
```
asturur, line 883: please restore getPointer.

src/canvas/SelectableCanvas.ts
```diff
@@ -854,41 +839,62 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+   *
+   * @param {Event} e
+   * @param {Boolean} inHTMLPlane
+   * `true` returns a point that from the viewer's perspective remains untouched
+   * by {@link viewportTransform} changes\
+   * `false` returns a point in the canvas plane,
+   * the same plane as the plane {@link FabricObject#getCenterPoint} exists in
+   * @return {Point}
+   */
+  protected getPointer(e: TPointerEvent, inHTMLPlane = false): Point {
```
asturur, line 897: please restore getPointer as normal method, we are not adding another migration headache to devs.

src/canvas/SelectableCanvas.ts
asturur: getting the pointer traversing the dom can be expensive, that is why caching was added.
An event chain can do that multiple times, the risky businness of not having stale cache is on us. The events are syncronous and they create and delete the cache, a user could save an event object from yesterday, and use it today on an event handler and yes get a wrong result. We should document that getPointer inside an event handler is meant to return you the pointer for that event and not a previous mouse event.

src/canvas/Canvas.ts
```diff
@@ -245,8 +245,8 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
     const shared = {
       e,
       isClick: false,
-      pointer: this.getPointer(e),
-      absolutePointer: this.getPointer(e, true),
+      pointer: this.getHTMLPointFromEvent(e),
+      absolutePointer: this.getCanvasPointFromEvent(e),
```
asturur, line 249: getCanvasPoint won't work because Canvas is also an html term.
Also HTML doesn't really say the whole story because is still from the top-left corner of the canvas and not the screen for example.

I usually call them viewportPoint and designPoint but i want to hear more opinions

asturur:
> I now understand that containsPoint needs fixing - it uses a complete mess of different coords, not taking group into account

subtarget checks with multiple groups has always worked, that was its job.
If now has a new job, it may not be fit for the job.

src/canvas/Canvas.ts
ShaMan123: viewport is risky because the page has a viewport also
I agree the names must be very acurate.

src/canvas/SelectableCanvas.ts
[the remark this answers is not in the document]
ShaMan123: It does.
The method accepts points in different cases, under group with viewportTransform removed from the point and if not under group then the point in the html plane.
If that is not a bug in design I do not now what a bug is.
I do not plan to touch this because #8767 fixed that and I am tired of getting back to thing I fixed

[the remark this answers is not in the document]
ShaMan123: It is a bug because I can't use containsPoint, no one can without copying this code because it makes no sense

ShaMan123:
> Please add a description showcase of the bug. What the bug was? how did you find it? containspoint was doing stuff that you assumed was other stuff, and now we scream BUG BUG BUG MESS MESS MESS the code is not understandable. The truth is that all the changes that have been done in the name of improving something else changed the meaning of those functions, containsPoint was strictly used for checking targets on mouse.

This proves to me how dangerous `getPointer` is.
Did you miss https://github.com/fabricjs/fabric.js/pull/9175#discussion_r1305437169?
You are welcome to revert or cherry pick, I see how dangerous it is from our discussion and from the fact that every time I use `getPointer` I must read again the entire comment and start imagining what it does.
You might understand it, but I don't. And I assume others won't as well.
There is no overhead or headache, just a rename. And fact is I could find the bug easily once I renamed.
Nothing is breaking.
If you dislike the name let's rename.
If you dislike the PR, close it and fix it in  a different PR. I won't be happy, but that's life.
If we want fabric to be robust it must be easier.
Even doing a simple `find all occurances` is now a lot easier than with `getPointer`

src/canvas/Canvas.ts
```diff
@@ -930,8 +930,8 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         subTargets: targets,
         button,
         isClick,
-        pointer: this.getPointer(e),
-        absolutePointer: this.getPointer(e, true),
```
ShaMan123, line 934: an example of the bug

src/canvas/Canvas.ts
```diff
@@ -930,8 +930,8 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
+        pointer: this.getHTMLPointFromEvent(e),
+        absolutePointer: this.getCanvasPointFromEvent(e),
```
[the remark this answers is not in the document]
ShaMan123: https://github.com/fabricjs/fabric.js/pull/9175/files#r1306300610

ShaMan123:
Also cherry pick the test against master and you will see the diff

src/canvas/Canvas.ts
ShaMan123: should have been
```
```

src/canvas/SelectableCanvas.ts
ShaMan123: you mean remove the protected keyword?

ShaMan123: I can do that.
As I see it it is a present to the dev.
TS will warn them about it. Then they can read the comment and decide if to change it or not. It is not a must. But it might help devs when debugging a hard bug.

ShaMan123: image

ShaMan123: Apart from saving the devs, having it protected saves fabric devs - I care about that a lot.
But it is not a must if you disapprove

asturur: we can let it public and deprecated and later it will be protected with an internal name. we can't remove it now, i think i have hundreds of getPointer usage and so have others.

ShaMan123: Very well

asturur:
the confusing methods were only a way to reduce code duplication across fabric, and have no value so is fine they can go.

In all this discussion is unclear to me how many instances of getPointer we swapped during refactors. one or more?

We already discussed that all the absolute:true, ignoreVpt:true and also the definition of absolutePointer are confusing.

Said so be worried of dev not understanding getPointer and then not being able to get it after a console.log is an excess of worry.

This is a change we can merge after we settle on names and i ll try to avoid the size bump that maybe can be avoided

src/canvas/Canvas.ts
asturur: maybe getHTMLCanvasRelativePoint(e) and getDesignCanvasRelativePoint, or for this second one, getDocument, getPage CanvasRelativePoint.

src/canvas/Canvas.ts
asturur: @asturur bug 1.

asturur: sorry but this is how is in other places too:
image

are you sure this is one instance of the bug?

asturur:
@ShaMan123  i m sorry but i m having an hard time understanding the changes in this PR.
you swapped a bunch of pointers for a html to non html and the reasons are not clear to me.

src/canvas/__tests__/eventData.test.ts
```diff
@@ -0,0 +1,119 @@
...
+        'mouseover',
+        'dblclick',
+        'wheel',
+        'contextmenu',
+      ] as (keyof WindowEventMap)[]
+    )
+      .map(
+        (type) =>
+          [
+            [type, iMatrix],
```
asturur, line 36: we don't need the test with iMatrix, do we?

src/canvas/__tests__/eventData.test.ts
```diff
@@ -0,0 +1,119 @@
...
+            [type, [2, Math.PI, 0, 0.5, 50, 50] as TMat2D],
```
asturur, line 37: the choice for this generic matrix but with clientX and clientY matching the translation bring the absolutePointer always to 0,0 nullifing the effect of zoom and that unused skew/rotation extra parameter.

Let's have a matrix that cover our use cases, with 0 and 3 same number and 1 ad 2 set as zero, and the translation that is not the same as the mouse client coord.

asturur:
So i spente last hour carefully reading this pr, i think there is a big issue here.
today the code returns:

getPointer(e) => the default behaviour of getPointer, the fabric design canvas coordinate, stored in `_absolutePointer` and returned in the events in pointer ( the default ).

getPointer(e, true) => the html canvas coordinates, stored in `_pointer` internal cache and returned in the event with the name absolutePointer ( added later in the game ).

The names are bad ( or swapped if we want), but this PR is swapping the data too by assigning the `getPointer(e, true) to the function called CanvasCoordinates and assigning to getPointer(e) the one called HTML, so it is swapping A with B.
Or am i seeing this wrong?
We have to approach this differently.

So first of all i ask you, which is the bug you got that led you to create this PR? because in all of this conversation i m still confused.
If we identified a clear bug we should fix it, and the approach assigning better names in some way.

asturur:
ok now after spending more time i figured out in 5,x the values were as you are putting them here, let's search the commit

asturur:
https://github.com/fabricjs/fabric.js/commit/b6f076cc0c817cd7f70abc7028c9d04295d121a8
8 months ago when we converted the mixin.
I wonder why we had to change it to a function call when we had the the this._absolutePointer and this._pointer.
Some TS complain maybe that forced us to not use the cached values?

ShaMan123:
So naming....
I think we should avoid using the words `viewport`,  `absolute`

ShaMan123 (commented):
Fixed toRoundedSnapshot - I can extract it to a standalone PR if you prefer.
changed `getPointer` signature from `protected` to `deprecated`.

ShaMan123:
What about:
`getHTMLPointFromEvent` => `getPointAsElement` / `getEventPointAsElement` / `getPointFromEventAsElement`
`getCanvasPointFromEvent` => `getPointFromEventAsObject` etc. (I don't like this option)

??

asturur:
i would happily break this PR in a way that we can identify the breaking commits and fix those, and leave this for renaming / deprecation.
As it is now this pr i identified that breaking commit i did, but i still have to look at the other changes that are hard to follow for me.

```
getHTMLPointFromEvent => getPointAsElement / getEventPointAsElement / getPointFromEventAsElement
getCanvasPointFromEvent => getPointFromEventAsObject etc. (I don't like this option)
```
those sound worse to me.

i would call the fabric space either design or doc, so getDesignRelativePoint, getDocRelativePoint, are the only one not mixing with html naming counterparts.

ShaMan123:
The doc option sounds ok

ShaMan123:
but what is the counterpart?

asturur:
ok now the pr is smaller and we need to decide on naming only

jiayihu:
> i would call the fabric space either design or doc, so getDesignRelativePoint, getDocRelativePoint, are the only one not mixing with html naming counterparts.

I find the "design" naming extremely unintuitive and abstract. Also bear in mind that much of the v6 codebase already uses "canvas plane" as term, so I'd personally pick `getCanvasPointFromEvent` and `getHTMLPointFromEvent`. Alternatively, `getDocumentPointFromEvent` sounds also good to me, like `getDocRelativePoint`. The difference is avoiding to use the words relative/absolute from this point on in favour of "document/html plane, canvas plane, group plane" etc.

ShaMan123:
What about using external vs. internal?
getExternalPoint, getInternalPoint?
Or something like that?
I agree using html is indeed misleading as well

asturur:
i don't like the plane reference at all in general, for me the sendPointToPlane is a regrettable name also.
for me the best 2 options so far are:
getDocPointFromEvent() and getHMTLCanvasPointFromEvent(). are extremely long and i really dislike that.

I prefer getViewportPoint instead of getHMTLCanvasPointFromEvent, @ShaMan123 said something about the viewport being confusing but i don't understand why.

pointer and absolutePointer needs to stay in the event data object but needs to be near the new names, deprecating the old one, that in this case would be:
docPoint and HTMLCanvasPoint

ShaMan123:
Viewport is confusing because the window/browser also has a viewport.

ShaMan123:
Doc is almost there, not quite IMO

ShaMan123:
Why is sendPointToPlane not good?
We can still rename.
I want to understand your POV

asturur:
> Viewport is confusing because the window/browser also has a viewport.

true, but no one knows about that, also if we call it canvasViewport or fabricViewport or docViewport there is no chance of confusing.

Let's focus on the main one, the `Doc` one. Doc is quite there, but not entirely, let's have proposal for this one.

Doc
Page
Design
Drawing
~Document~ too much html
layout
basicPlane
~sheet~ too hard to pronounce correctly when speaking


Thesaurus isn't helping with canvas, page, document synonims

ShaMan123:
So you dislike the word plane? U prefer space?
canvasViewportSpace/Plane is a good name IMo BUT what do you call the html space? EventSpace? We can call it the nonTransformed space... or something.

ShaMan123:
I saw also usage of `scene` in konva and others

ShaMan123:
canvasPlane/Space vs. canvasElementSpace/Plane?

ShaMan123:
BTW there is a weird thing when using sendPointToPlane with vpt... it feels a bit reversed
```
cnst pointInCanvasElPlane = sendPointToPlane(pointInCanvasPlane, canvas.viewprtTransform)
```

I make sense but still hard to grasp

asturur:
> So you dislike the word plane? U prefer space?
> canvasViewportSpace/Plane is a good name IMo BUT what do you call the html space? EventSpace? We can call it the nonTransformed space... or something.

forget plane for this pr let's focus on the names of those pointers

asturur:
> I saw also usage of `scene` in konva and others

scene is good. Scene is better than document or page. 
we still need the other that for me can be HTMLCanvasPoint that is just ok.

ShaMan123:
My question is why introduce another name?
If we use HTMLCanvasPoint/Plane/Space why not use CanvasPoint/Plane/Space and document it clearly?

asturur:
the second name is for the events.
we suggest to not use getPointer and in the long run it gets private.
Even if we just add getScenePointFromEvent and nothing else, we still need to call getPointer() to backfill pointer and absolutePointer in the event data for release.
regardless of what we choose to do we have to pick up new names for pointer and absolutePointer in the event data to put near the old one that we will remove at some point, since pointer and absolutePointer are confusing as getPointer.

ShaMan123:
canvasPointer, canvasElementPointer?

asturur:
> canvasPointer, canvasElementPointer?

those are points, so we need to remove pointer name.
if we call the first method getScenePointFromElement() the event could probably have scenePoint and HTMLCanvasPoint in the event. I still think it has value to provide the HTMLCanvasPoint until we still have some duality in coordinates in fabric internals, or at least to move away from absolute pointer and give a way out of it.

We do not need to provide an alternate method like getHTMLCanvasPointFromElement if we think they should use the the scene + transformation, but we still need to have a way to add it to the event using the cached  value as we do now

ShaMan123:
In #8767 there is some mechanism that can be used for this purpose.
Are you sure scene is good? I am not
Regarding caching... I think it is enough to cache a single point and use `sendPointToPlane` - It is a single matrix multplication. Doesn't sound damaging to perf.

ShaMan123:
I gave it more thought.
I am aginst scene.
Viable options:
| canvas el | canvas plane | 
| --------- | -------------- | 
| identity plane | canvas plane | 
| viewer/viewing plane | renderer/rendering/object plane |

The canvas el plane is the identity plane for the dev. 
If a chess player tells a pawn that exists in the identity plane to move a step of (1, 1) both will measure the same distance. 
However if the player tells a pawn exisiting in the canvas plane both will mesure a differnet thing if the vpt is not the **identity** matrix.
I will use this term from now on, hopefully we can agree on something here.

asturur:
what is wrong with scene? seems a commonly used name.
If viewingPlane is good for you we keep viewportPoint and getviewportPoint.... Since viewport is that same concept.

ShaMan123:
Viewport has too much load on it.
I already said that.
Tell me, what is a viewport point?
A point in the browser viewport? Of course not, but still confusing.
A point in the canvas viewport? Or in the canvas viewport transform plane?????
The first is the identity llane and the second is not.

ShaMan123:
Ok, scene is fine
We have a scene playing and a viewer viewing it.
Can you live with that?
Viewport is a no go, just asks for more confusion

ShaMan123:
Which other names can convey the fact we are looking through a perspective? Like a camera, lens etc.

asturur:
My mind goes to viewport because is a term i have seen always around in graphic, sometimes referred as camera sometimes viewport. While viewer not.

i think that eventually we could also not use scene and explain that when an event or an object has a property called point, that is just in the scene/design/document/fabric coordinate. No ambiguity. If we give back a point in a different coordinate system then we would have to name it.

asturur:
If i google viewport it seems to confirm is exactly the viewer:

https://en.wikipedia.org/wiki/Viewport

image

https://docs.unity3d.com/ScriptReference/Camera.WorldToViewportPoint.html

image

Some common points that come up is that one viewport point is always a unit in the device that display ( in our case we use css pixels and not screen pixels )


Regarding the fact that viewport has too much load i disagree, we have a viewport transform that is a concept that is long in fabric, and we have a method like viewportCenter object.

We may like it or not but is a common way to call that kind of coordinates and is well known to whoever does graphics.

ShaMan123:
Ok 
I just find it confusing that when sending a point from the viewport to the canvas plane and vice versa then the viewport transform is applied in the opposite way in relation to the name.
So viewport is ehat the element measures, and scene is what an object measures?

jiayihu:
Just call it `clientViewport`, `windowViewport` or `documentViewport` guys, then it's clear compared to `canvas.viewportTransform` or "canvas viewport" which indeed is correctly meaning the area in canvas plane coordinates. Whereas `clientViewport` means the area in client coordinates

ShaMan123:
renamed.
final thing I need to understand - I will add a test for it.
Seems the calculation of the scene point is wrong because in `getPointer` it skips the last part of the retina normalization whereas in `_cacheTransformEventData` it does not

ShaMan123 (commented):
renamed, rm `transformPointRelativeToCanvas` which is related to this mess

src/canvas/Canvas.ts
```diff
@@ -855,7 +855,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
             transform.target,
             originalControl
           );
-      pointer = pointer || this.getPointer(e);
+      pointer = pointer || this.getScenePoint(e);
```
ShaMan123, line 858: seems redundant since we cache `getScenePoint`

src/canvas/Canvas.ts
```diff
@@ -984,8 +985,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       this.discardActiveObject(e);
       this.requestRenderAll();
     }
-    // this is an absolute pointer, the naming is wrong
-    const pointer = this.getPointer(e);
+    const pointer = this.getScenePoint(e);
     this.freeDrawingBrush &&
       this.freeDrawingBrush.onMouseDown(pointer, { e, pointer });
```
ShaMan123: we should rename this as well when doing the absolutePointer rename

src/canvas/Canvas.ts
```diff
@@ -1150,8 +1150,12 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
   _cacheTransformEventData(e: TPointerEvent) {
     // reset in order to avoid stale caching
     this._resetTransformEventData();
-    this._pointer = this.getPointer(e, true);
-    this._absolutePointer = this.restorePointerVpt(this._pointer);
+    this._pointer = this.getViewportPoint(e);
+    this._absolutePointer = sendPointToPlane(
```
ShaMan123, line 1155: also these should be renamed and maybe put into an object

src/canvas/Canvas.ts
```diff
@@ -1343,7 +1347,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
    * @param {Event} e Event fired on mousemove
    */
   _transformObject(e: TPointerEvent) {
-    const pointer = this.getPointer(e),
+    const pointer = this.getScenePoint(e),
```
ShaMan123: rename

src/canvas/__tests__/eventData.test.ts
ShaMan123, line 1: this tests does not seem very useful

ShaMan123 (commented):
seems ready apart for the useless tests

ShaMan123:
and now I pushed a regression

src/canvas/SelectableCanvas.ts
```diff
@@ -765,7 +764,10 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
       obj &&
       obj.visible &&
       obj.evented &&
-      obj.containsPoint(this.restorePointerVpt(pointer), true)
+      obj.containsPoint(
+        sendPointToPlane(pointer, undefined, this.viewportTransform),
+        true
```
ShaMan123, line 769: keep this in mind for conflict resolution with #9395

ShaMan123 (commented):
test should pass now

asturur:
ok i m merging.

asturur (approved):

src/EventTypeDefs.ts
```diff
@@ -129,9 +129,21 @@ export interface TPointerEventInfo<E extends TPointerEvent = TPointerEvent>
...
+   */
+  pointer: Point;
+  /**
+   * @deprecated
+   * use scenePoint instead.
+   * Kept for compatibility
+   */
   absolutePointer: Point;
+  scenePoint: Point;
+  viewportPoint: Point;
```
ShaMan123, line 146: the order is confusing 


should have been 

```
pointer,
absolutePointer,
viewportPoint,
scenePoint
```

OR

```
pointer,
viewportPoint,
absolutePointer,
scenePoint
```


It might seem petty or insignificant but IMO it isn't
A dev reading this code will scratch their hand

ShaMan123 (commented):
You forgot the brushes

src/canvas/Canvas.ts
```diff
@@ -24,6 +24,17 @@ import { TextEditingManager } from './TextEditingManager';
...
 
+const getEventPoints = (canvas: Canvas, e: TPointerEvent) => {
+  const viewportPoint = canvas.getViewportPoint(e);
+  const scenePoint = canvas.getScenePoint(e);
+  return {
+    viewportPoint,
+    scenePoint,
+    pointer: viewportPoint,
+    absolutePointer: scenePoint,
+  };
```
ShaMan123, line 35: there is anotherthing I have been thinking of lately
Since we cache the points an event can do
```
scenePoint.x += 5
```
And cause a bug all across fabric so we should return a clone
I will PR

jiayihu:
FWIW I think the final decision to use scene/viewport as names adds confusion because other parts of the codebase will use a different terminology. So at least another PR is needed so that the whole codebase uses a consistent naming for the same concept.

Screenshot 2023-11-09 at 11 40 18

ShaMan123:
#9483 has some jsdocs fixes including what you pointed out
After searching there are 4 usages of `canvas plane` that can be turned into scene plane but that seems minor

PR #9178, opened 2023-08-25
fix(utils) Fixes the code for the anchor point in point controls for polygons

asturur:
## Description

The demo code was using `_getNonTransformedDimensions` while in the util we settled for width/height.
What i can think of is that we did it because we were working with exact bounding box in mind


src/controls/polyControl.ts
```diff
@@ -80,28 +80,26 @@ const anchorWrapper = (
...
       adjustFlip = new Point(poly.flipX ? -1 : 1, poly.flipY ? -1 : 1);
 
-    const newPosition = new Point(
-      poly.points[anchorIndex].x,
-      poly.points[anchorIndex].y
-    )
+    const newPositionNormalized = anchorPoint
       .subtract(poly.pathOffset)
-      .divide(polygonBaseSize)
+      .divide(poly._getNonTransformedDimensions())
```
asturur, line 94: this is the fix.
All the rest is trying to reduce code footprint

ShaMan123: You are right, I changed the demo to use `exactBoundingBox` and it works.
I forgot about it.
But then I saw that origin is wrong `exactBoundingBox`

ShaMan123 (commented):
I think we should a playwright test

ShaMan123 (approved):

ShaMan123:
Since we plan to deprecate `exactBoundingBox: false` the demo should use `exactBoundingBox: true`

hexiaokang (approved):

asturur:
> Since we plan to deprecate `exactBoundingBox: false` the demo should use `exactBoundingBox: true`

The demo could use both and yes a test is needed i ll add it.

For deprecation or defaults, the plan is also to reduce the bundle, so we need to have this bounding code in a place we can exclude from a functional minimal code bundle. If we switch that to true, when is an optional import we need to switch it back to false. So we need to get there with less breaking changes as possible

ShaMan123:
I think most if not all devs will want this on always.
But it might be it could be integrated into the redo coords somehow

asturur:
I refrain from using all, since for example we are 2 developers and we have pretty much opposed views. So given the fact we are both normal devs, the `all devs` rarely happens.

If you are doing something pretty much geometric you need  it, but for example for most of the shapes used in applications that i have seen around those end up being paths ( with M and L and V and H ) and so this code won't help.

The code similar to this but aimed at fixing also Paths is probably even larger.

trying to write a test now

asturur:
ok the test is in, it was easy with the textbox example.

ShaMan123:
Should we add a strokeUniform variant as well?

ShaMan123:
I am beginning to love playwright

PR #9179, opened 2023-08-26
change(): rm 

asturur:
## Motivation

Remove some old code rediscovered by chance.

## Description

`_setOriginToCenter` and `_resetOrigin` are 2 methods that other than setting the origin are also changing top and left.
They were written for rotate, are private and used only for rotate.
They add 2 properties to every object ( `_originalOriginX` and `_originalOriginY` ) that keep track of the state.
All of those things can be handled by the rotate method, that is a public facing utility, by using simple var scope.

This change should have no change in functionality and remove 2 methods and 2 class members.

### DIFF
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -1491,18 +1491,26 @@ export class FabricObject<
    * @param {TDegree} angle Angle value (in degrees)
    */
   rotate(angle: TDegree) {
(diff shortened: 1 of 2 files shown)


ShaMan123 (dismissed):
looks good, do we have a test for this case?

asturur:
Covered with a test all the 3 cases.
Non of them was covered before apart angle itself.

PR #9185, opened 2023-08-30
feat(Node): pdf/svg output

ShaMan123:
## Motivation

I was in a need for pdf output

## Description

Utilizing node canvas' native pdf/svg output by accepting a `ctx` as arg for `toCanvasElement`

first merge #9112 for registering fonts

I have come up with a different approach that can be achieved by 2 subclasses: https://codesandbox.io/p/devbox/fabric-node-sandbox-forked-mg9222?workspaceId=d29ca40f-0679-47bb-8fc8-429da561f9e5

## Changes

- added additonal `ctx` optional arg/option
- types
- tests

## In Action

https://codesandbox.io/p/devbox/modern-waterfall-glhq62


ShaMan123 (commented):
need to add fonts in node

ShaMan123:
@melchiar I think you would be interested in this

src/canvas/StaticCanvas.ts
```diff
@@ -1363,16 +1363,16 @@ export class StaticCanvas<
...
-      multiplier = 1,
-      enableRetinaScaling = false,
-    } = options;
+  toDataURL({
+    format = 'png',
+    quality = 1,
+    multiplier = 1,
+    enableRetinaScaling = false,
+    ...options
+  }: TDataUrlOptions = {}): string {
```
asturur, line 1375: to me it doens't make sense to specify a ctx here.
Can we releage this to a specific node item?
If we need to export something in PDF/SVG using node-canvas, shouldn't we add a method that is specific to that and return something related to that?

We don't have code size issues in node and we are not bound to bundling issues.
Having toCanvasElement that return a canvas element that  is good for pdf/svg export buy only if we previously created a specific ctx sounds complicated.

Let's have a toNodePDF and a toNodeSVG ( or whatever it can be called to avoid confusion with current toSVG ) in the node classes.

Take also care that moving away from node-canvas is still a remote option and if there are other libraries that do not require building or that are more canvas compliant, is not granted they will have a pdf or svg export.

ShaMan123: That is why I didn't expose toPDF etc. So we are not bound to node canvas.

Suggest a different way?

ShaMan123:
> I have come up with a different approach that can be achieved by 2 subclasses: https://codesandbox.io/p/sandbox/fabric-node-sandbox-forked-8hwny3?file=%2Fsrc%2Findex.mjs%3A9%2C1-22%2C2

.

ShaMan123 (commented):
rebased to head

ShaMan123:
Anyone interested in this great solution bear in mind that it is also possible to achieve it on the client by using canvas API svg renderers.
This will not merge into core so I am closing

hamzawain7:
Hi @ShaMan123 any progress on this? Do you think this works perfectly? I want this feature for my website. Can I manually merge these changes in my fork? Are there any known issues?

hamzawain7:
I love this. Can we please add this feature? Can I manually merge these changes in my fork till then? Are there any known issues?

hamzawain7:
I love & need this. Can we please add this feature? Can I manually merge these changes in my fork till then? Are there any known issues?

asturur:
This should be node only and not on the generic methods. Those will work only under node and there should be.

PR #9186, opened 2023-08-30
fix(): regression of `getPointer` usages + BREAKING: drop event data

ShaMan123:
## Motivation

prepare for #9175 

fixes bug from https://github.com/fabricjs/fabric.js/commit/b6f076cc0c817cd7f70abc7028c9d04295d121a8

## Description

Somewehre in v6 `_pointer` and `_abolutePointer` were changed to `getPointer` and many ~done WRONG~ were swapped in values.
Not a single test catch it so I have added tests

`pointer` => `getPointer(e, true)`
`absolutePointer`  => `getPointer(e)`


ShaMan123 (commented):
step

asturur:
@ShaMan123 those are the only swap instances i could find. Do you think there are more?

I removed the identity matrix test and i added a more generic matrix that doesn't put the absolute pointer in 0,0

src/canvas/Canvas.ts
```diff
@@ -512,6 +512,8 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       target,
       subTargets: targets,
       dragSource: this._dragSource,
+      // this is an absolute pointer, the name convention is wrong.
+      // such was in 5.x
```
asturur, line 516: What do we want to do with this? do we make consistency and we fix it?

asturur (dismissed):
let with my proposal of normalization. If this is good for you is approved you can merge.

src/canvas/Canvas.ts
asturur: update: i normalized it. Do not merge if you don't agree.

asturur (dismissed):

asturur (approved):

src/canvas/Canvas.ts
ShaMan123: Yes
I think we should expose a protected method that build the event data so everything calls it

src/canvas/Canvas.ts
```diff
@@ -513,8 +513,8 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       subTargets: targets,
       dragSource: this._dragSource,
       // this is an absolute pointer, the name convention is wrong.
```
ShaMan123, line 515:
```suggestion
```

ShaMan123 (commented):
Brush API also uses the wrong naming... but that was like that before v6 => to be done in  #8476 IMO
Group selector also uses the absolute pointer, which is weird cause it is not affected by vpt

getPointer usages seem correct

ShaMan123:
you anyways need to approve it since I created it, so merge at will

ShaMan123 (commented):
done?

asturur:
I can't find any getPointer usage in the brush files, can you point to me?
with this PR any usage of getPointer function is associated to the correct name, or at least that seems  to me

asturur (approved):

asturur:
At least the pointers in brush are exposed to the brush functionality only, that is an issue for who is creating custom brushes that should be way less than who is just listening to events

PR #9189, opened 2023-08-30
Avoid Group transform when calculating the initial objects bb

jiayihu:
Not 100% this is a bug, but it looks like it to me. If you create a Group and you specify the `width + height`, but not `top/left`, the Group will calculate the initial children bounding box. Hower the calculation is wrong IMO, because it sends the `relativeCenter` to the Group plane, which I'm pretty sure should not be done during initialization. The group is calculating its center by computing the children bb, which in turn uses the group center ??? 

In the test I added for instance:
- the `relativeCenter` is `x: 100, y: 100`
- it gets sent to the Group plane, whose centerPoint is also `x: 100, y: 100` because I've specified a width and height

The end result is that the child is at `x: -200, y: -200`, whereas it should be `x: -100, y: -100`.

Screenshot 2023-08-30 at 13 25 05

### DIFF
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -826,7 +826,8 @@ export class Group extends createCollectionMixin(
       return;
     }
 
-    const bbox = this.getObjectsBoundingBox(objects) || ({} as LayoutResult);
+    const bbox =
+      this.getObjectsBoundingBox(objects, true) || ({} as LayoutResult);
     const { centerX = 0, centerY = 0, width: w = 0, height: h = 0 } = bbox;
     const width = hasWidth ? this.width : w,
       height = hasHeight ? this.height : h,
@@ -897,14 +898,14 @@ export class Group extends createCollectionMixin(
   }
 
   /**
-   * Calculate the bbox of objects relative to instance's containing plane
+   * Calculate the bbox of objects, by default relative to instance's containing plane
    * @public
    * @param {FabricObject[]} objects
    * @returns {LayoutResult | null} bounding box
    */
   getObjectsBoundingBox(
     objects: FabricObject[],
-    ignoreOffset?: boolean
+    ignoreContainingPlane?: boolean
   ): LayoutResult | null {
     if (objects.length === 0) {
       return null;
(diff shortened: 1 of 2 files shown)


src/shapes/Group.ts
```diff
@@ -897,14 +898,14 @@ export class Group extends createCollectionMixin(
...
   /**
-   * Calculate the bbox of objects relative to instance's containing plane
+   * Calculate the bbox of objects, by default relative to instance's containing plane
    * @public
    * @param {FabricObject[]} objects
    * @returns {LayoutResult | null} bounding box
    */
   getObjectsBoundingBox(
     objects: FabricObject[],
-    ignoreOffset?: boolean
```
jiayihu, line 907: Removed and replaced because it was not used by any call actually

ShaMan123:
I will give this a deep look next week.
Though we might want to port this on top of #9148

jiayihu:
> Why do you want to specify width/height but not a layout strategy?

For loading SVGs as shapes in our application, where a Shape is a Group comprising of the fabric.Paths from the loaded SVG and the shape text. This means that the Shape Group has width/height speficied as it's known for each SVG.

Also generally speaking, since fabric supports the `fixed` layout strategy, it seems to me very likely that one might want to specify width/height. And indeed I've actually specified the layout strategy as "fixed" in my application and the bug still is reproduced.

asturur:
> > Why do you want to specify width/height but not a layout strategy?
> 
> For loading SVGs as shapes in our application, where a Shape is a Group comprising of the fabric.Paths from the loaded SVG and the shape text. This means that the Shape Group has width/height speficied as it's known for each SVG.
> 
> Also generally speaking, since fabric supports the `fixed` layout strategy, it seems to me very likely that one might want to specify width/height. And indeed I've actually specified the layout strategy as "fixed" in my application and the bug still is reproduced.

But that would make it specifically an svg import issue  if the intention is to respect the svg viewbox property.
Can you share the svg that brought you to this investigation?

jiayihu:
Forget about the SVG, it's probably only adding confusion. It was a bad example from me. Suppose instead I just want a fixed layout of 100x100, and I initialize it with a Rect of 100x100. This looks a fair legit case IMO, like I'm implementing some sort of calendar layout. What do you think?

Screenshot 2023-09-01 at 15 44 58

ShaMan123:
> the Group will calculate the initial children bounding box. Hower the calculation is wrong IMO, because it sends the `relativeCenter` to the Group plane, which I'm pretty sure should not be done during initialization. The group is calculating its center by computing the children bb, which in turn uses the group center ???

I think you have pointed a real issue in the logic
I am looking into it

ShaMan123:
closing in favor of recent fixes in #9148

asturur:
but are unmerged yet. the pr merge will close the issue automatically

asturur:
i got confused i thought this was an issue

PR #9096, opened 2023-07-15
BREAKING fix(Text): styles `onInput` 

ShaMan123:
## Motivation

closes #9028 

## Description

I had an Eureka about text styles and saw it to be true
There was a confusion in the code, styles correlates to the UNWRAPPED lines so I decided to make it strict.
This was found across a lot of logic

## Changes

**BERAKING**: `get2DCursorLocation` => `getCursorPosition`, `getStyleCursorPosition` to avoid confusion
The dev will be happy to find this breaking because it will uncover bugs.
  **BREAKING**: rm `getCurrentCharFontSize`, `getCurrentCharColor`, `_getCurrentCharIndex`
  **BREAKING**: rm `cleanStyle`, `removeStyle`, `_extendStyles`, `_setStyleDeclaration`, `_deleteStyleDeclaration`, `_getLineStyle`, `_setLineStyle`, `_deleteLineStyle`
  **BREAKING**: rm `_getStyleDeclaration`, `getCompleteStyleDeclaration` => `getStyleDeclaration`

## In Action

https://github.com/fabricjs/fabric.js/assets/34343793/e31635bb-906c-432b-bf01-d7524b995d35
[video.webm](https://github.com/fabricjs/fabric.js/assets/34343793/ec3049ed-33ef-4943-bef0-b0fbdce16a86)

#### FIXED

[video.webm](https://github.com/fabricjs/fabric.js/assets/34343793/234c4849-a3ee-41fe-add2-7929717f3106)

test-finished-1

### MASTER

[video.webm](https://github.com/fabricjs/fabric.js/assets/34343793/89e06273-a852-44a1-85fe-e40ca7f1a5cb)

test-finished-1

PR #9192, opened 2023-08-31
fix(Textbox) Textbox inputs with new lines

### LINKED ISSUE #9190: [Bug]: Textbox input seems completely off
In current master textbox input seems completely broken.
Adding new lines does not add  new lines after the first one.

- double check that is actually broken after a night of sleep
In case is broken:
- identify the commit that broke it
- revert the commit
- re-evaluate what to do

asturur:
## Motivation

This PR reverts the part of broken code in https://github.com/fabricjs/fabric.js/pull/9097.
The code doesn't make sense, and we need a test to be sure we don't break it again.

The regression is happening only for splitByGrapheme set to true.

There is still a style issue that i still don't know from which commits originates, also only for splitByGrapheme, and i m looking into it

closes #9190 

To add a test the following addition have been done to the CanvasUtil:

- add a press method that takes a key
- add a ctrlC and ctrlV method.

ctrlC and ctrlV require permissions for clipboard handling and need to treat macos differently.


src/shapes/Textbox.ts
```diff
@@ -339,17 +345,16 @@ export class Textbox<
...
+        // @ts-expect-error
+        wordsOrGraphemes.push([]);
       }
 
       return wordsOrGraphemes.map((word: string) => {
         // if using splitByGrapheme words are already in graphemes.
-        const graphemeArray = splitByGrapheme
-          ? [word]
-          : this.graphemeSplit(word);
+        const graphemeArray = splitByGrapheme ? word : this.graphemeSplit(word);
```
asturur, line 354: this line and line 349 are both necessary to unbreak the textbox.

ShaMan123:
Please try the fix I did to styles and onInput

asturur:
i found this trying to merge the test from exactly that PR, that is the only thing that will likely get merged.
That pr won't be merged since:

- removes methods without explaining why
- renames methods together with the fix
- does not explain where the fix is.

i would gladly merge the fix if i could individuate it

asturur:
but yes the style bug is the same that is reported in #9028, so i don't need to investigate it.

asturur:
added a test that covers inputing new lines with both splitByGrapheme true and false

ShaMan123:
If the description is the problem I can amend that.

asturur:
no @ShaMan123 that pr is a large mix refactor that is not going anywhere for me. if you identified the style bug we can fix it, otherwise we will have to find the bugged code first.

ShaMan123:
I wrote what the bug is.
A mix between the wrapped and non wrapped use of line/char index
That is why I renamed stuff.
A lot like with getPointer.
And removed the methods that existed to work around all the mix.

asturur:
We talked about this more than once.
What you do is `this code is terrible, i ll write a bunch of new code that works`
I can't accept PRs like that. Not to fix a bug.
And anyway is unrelated from this PR.

ShaMan123:
In the weekend I will get to it

asturur:
ok i ll let text rest this week then. I ll move to look at the layout pr.

ShaMan123 (changes requested):
Sorry @asturur but it seems to me you are redoing #9096 (and not fixing all it fixes) instead of merging it.
I really do not understand why you are against that PR. Yes, it removes many methods because they are not needed, all seem to have been written to patch the exisiting situation.
The test in #9096 is more extensive than the tests you wrote here because it also deletes lines.

e2e/tests/text/adding-text/index.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        textbox.splitByGrapheme = context.splitByGrapheme;
+        textbox.set('dirty', true);
+        textbox.initDimensions();
+        textbox.canvas.renderAll();
+      },
+      { splitByGrapheme }
+    );
+
+    await expect(await canvasUtil.screenshot()).toMatchSnapshot({
+      name: `1-initial-splitByGrapheme-${splitByGrapheme}.png`,
```
ShaMan123, line 27: Please rename:
```suggestion
      name: `1-initial-${splitByGrapheme ? 'splitByGrapheme ': ''}.png`,
```

ShaMan123 (changes requested):
Please explain why snapshot 3 is so different when splitByGrapheme is true/false
The style is not applied differently to the pasted text, right?

e2e/tests/text/adding-text/index.spec.ts
ShaMan123: also perhaps create a function to generate the name?

ShaMan123:
I have commited 4803a1951 - feel free to revert it.
It is a PR review suggestion I wanted to make.

src/shapes/Textbox.ts
```diff
@@ -339,17 +345,16 @@ export class Textbox<
         : this.wordSplit(line);
 
       if (wordsOrGraphemes.length === 0) {
-        return [];
+        // @ts-expect-error
+        wordsOrGraphemes.push([]);
```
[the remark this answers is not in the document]
ShaMan123: see https://github.com/fabricjs/fabric.js/pull/9192#issuecomment-1712401466

asturur:
@ShaMan123 because i m not doing refactors on text now or later to fix a bug that is a regression and that gets fixed removing the code that introduce the bug. This pr reverts the offending code that broke the functionality. period.

i will check your commit and if is not another refactor we can keep it.

the methods you deleted in your pr are user facing functionalities, are not used by fabricjs ( like cleanStyle )

ShaMan123:
I was sure cleanStyle was internal.
I do not see a use for it since stylesToArray has come but I might be missing something

ShaMan123:
> @ShaMan123 because i m not doing refactors on text now or later to fix a bug that is a regression and that gets fixed removing the code that introduce the bug. This pr reverts the offending code that broke the functionality. period.

Yes, but there are other bugs with style that PR fixes and these do not

asturur:
> I was sure cleanStyle was internal. I do not see a use for it since stylesToArray has come but I might be missing something

cleanStyle allows you to remove styles that are identical to the object main properties. It reduce the quantity of style applied to an object.

asturur:
> > @ShaMan123 because i m not doing refactors on text now or later to fix a bug that is a regression and that gets fixed removing the code that introduce the bug. This pr reverts the offending code that broke the functionality. period.
> 
> Yes, but there are other bugs with style that PR fixes and these do not and are not regressions

I m sure i talked about this with you already.
There is a bug, we find the bug we fix the bug. Rewriting a bunch of code to fix a bug is wrong.
I can't accept it.
I m not going to put in discussion a complicated piece of code like onInput because there is a style shift somewhere else ( that is fixed by another PR ).
When there are bugs we fix them as patches on the current codebase.

In this case we should have remove the breaking PR entirely but is not possible and remove a lot of TS work that is just types. So i just undid the lines that introduced the issue.

asturur:
> Please explain why snapshot 3 is so different when splitByGrapheme is true/false The style is applied differently to the pasted text, right?

snapshot 3 is different because this PR doesn't fix the style shift.
That will be fixed by the other PR and the snapshot will have to be updated

e2e/tests/text/adding-text/index.spec.ts
asturur: i think snapshots name are fine as they are
image

They specify if splitByGrapheme is true or false explicitly,  without leaving to interpretation the one that doesn't say it what it is

src/shapes/Textbox.ts
```diff
@@ -22,7 +22,7 @@ export const textboxDefaultValues: Partial<TClassProperties<Textbox>> = {
 
 export type GraphemeData = {
   wordsData: {
-    word: string[];
+    word: string[] | string;
```
ShaMan123, line 25:
```suggestion
    word: string[];
```

???

src/shapes/Textbox.ts
```diff
@@ -327,6 +327,12 @@ export class Textbox<
    *
    */
   getGraphemeDataForRender(lines: string[]): GraphemeData {
+    // this method has issues.
+    // it has been typed after some small refactors and is unclear how is working
+    // graphemeArray is either a string or a string[].
```
ShaMan123, line 332: this is also wrong now

asturur:
the line you changed seems to unbreak the issue anyway and is more consistent with how we planned that function in #9097 so we keep it.

asturur:
> The test in #9096 is more extensive than the tests you wrote here because it also deletes lines.

Yes because this PR is fixing a line adding bug, so it is testing that.

Regarding redoing #9096 in the other PR, i don't think it is true.
I m fixing the different bugs you found in the issue, on the current codebase, patching the current code.
There are 3 different bugs, so i ll need 3 smalls prs.
The third i didn't do yet.

ShaMan123 (approved):

PR #9198, opened 2023-09-03
Update publish.js

RohitPaul0007:


ShaMan123:
Sorry,
This is not pr material.
If you are looking to create meaningful prs you can ping us but better you look into bug reports and move from there.
Also, you should read the contribution guide.
Mind that changing var to let is more noise for us and than anything else.

PR #9322, opened 2023-09-09
fix(ActiveSelection): make sure canvas is in charge of setting initial coords

ShaMan123:
## Description

Moved the call to `setCoords` from the constructor.
It should not be there because the canvas ref is being set after that possibly, leading to wrong coords due to vpt.


ShaMan123 (commented):
ready for review

ShaMan123 (commented):
this fix is ready
the test is broken on master

asturur:
The changes is ok for me, but we should comunicate somewhere that using ActiveSelection manually is becoming highly not suggested.

src/shapes/ActiveSelection.spec.ts
```diff
@@ -55,4 +56,20 @@ describe('ActiveSelection', () => {
...
+        new FabricObject({
+          left: 100,
+          top: 100,
+          width: 100,
+          height: 100,
+        }),
+      ]),
+      viewportTransform: [2, 0, 0, 0.5, 400, 150],
+    });
+    expect(canvas.getActiveSelection().getCoords()).toMatchSnapshot();
```
asturur, line 72: let's test by picking aCoords and lineCoords that is what are we testing today.
We are testing that those are correct, without the need of being recalculated.
getCoords return an array and is mostly for internal usage of calculations, if we snapshot aCoords and lineCoords we also see tl,tr,bl,br.

src/shapes/ActiveSelection.spec.ts
```diff
@@ -55,4 +56,20 @@ describe('ActiveSelection', () => {
     selection.add(new FabricObject({ left: 50, top: 50, strokeWidth: 0 }));
     expect(selection.item(0).getCenterPoint()).toEqual({ x: 50, y: 50 });
   });
+
+  it('sets coords after attaching to canvas', () => {
```
asturur, line 60: Isn't this a canvas test? is the canvas executing the new code no?

ShaMan123: true, I have no idea where it should belong

src/shapes/ActiveSelection.spec.ts
ShaMan123: > We are testing that those are correct, without the need of being recalculated.

They are not recalcuated.

I just have in mind #8767 so I avoid lineCoords already but whatever

ShaMan123:
> The changes is ok for me, but we should comunicate somewhere that using ActiveSelection manually is becoming highly not suggested.

I was thinking to make the constructor protected. Then you get a TS error you can ignore.

ShaMan123 (commented):
ok

src/shapes/ActiveSelection.spec.ts
asturur: they are not recalculated no, but no owe knows it, is not written the test is not explcit.
if getCoords change or recalculation becomes default, this test doesn't break but also becomes uneffective.

asturur (approved):

asturur (approved):

src/shapes/ActiveSelection.spec.ts
ShaMan123: I do not follow
What should be changed?

asturur: oh i thought you changed because i have seen outdated tag on the discussion
i ll change it never mind

PR #9323, opened 2023-09-09
test(): Migrate json control tests

ShaMan123:
## Motivation

~I was working on #9148 and the tests drove me mad~

EDIT:

We are switching to playwright also for snapshot testing since the manually maintained visual testing library isn't great.
~~While we are probably loosing node snapshot with this change, we do know that node renders boxes and lines very good and the point of this test was always just to keep an eye on control positions.~~

**EDITED**: There doesn't seem to be a reason to test node because controls have no use in that env, however we can add also node snapshots once #9134 is merged

For the work i m doing around #9148 i m more comfortable with those tests in playwright


e2e/tests/controls/rendering/index.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+              object.padding = padding;
+              object.borderScaleFactor = 3;
+              object.transparentCorners = false;
+              const color = object.fill;
+              object._renderControls(canvas.contextContainer, {
+                borderColor: color,
+                cornerColor: color,
+              });
+              object.getObjects?.().forEach((subTarget) => {
+                subTarget.padding = padding;
```
ShaMan123, line 31: added this line

ShaMan123 (commented):
this is a simple test migration.
controls13/14 golden is data
controls12 golden is data2 and I added a test for it with padding

e2e/tests/controls/rendering/index.spec.ts
ShaMan123: the rest is unchanged

ShaMan123 (commented):
done

asturur:
At this point i want to know which part made you mad

e2e/tests/controls/rendering/index.spec.ts
asturur: can you do 2 * padding or 3 * padding in this subTarget? i want to see the  controls inside the group getting out of the group with a too large padding. This will ensure we cover weird situations we could break the code to

ShaMan123:
> At this point i want to know which part made you mad

The tests were failing and I wanted to inspect the json. That made me made.
Now inspecting it is easy

e2e/tests/controls/rendering/index.spec.ts
ShaMan123: done

ShaMan123 (commented):
good, updated description

asturur (approved):

PR #9325, opened 2023-09-09
Remove subtype in method parameters

jiayihu:
Remove type specialization in methods that made them non-assignable to the base class `FabricObject`. If you make `onDeselect` in InteractiveObject to require `InteractiveObject`, it makes the object not assignable to `FabricObject` anymore.

```ts
// Assuming for InteractiveObject
onDeselect(options?: {
    e?: TPointerEvent;
    object?: InteractiveFabricObject;
}

const callOnDeselect = (o: FabricObject) => o.onDeselect({ object: new FabricObject() })

// Unsafe, InteractiveObject technically expects InteractiveObject on onDeselect. 
// May use properties or methods not defined in FabricObject
callOnDeselect(new InteractiveObject())
```

### DIFF
--- a/src/shapes/IText/ITextBehavior.ts
+++ b/src/shapes/IText/ITextBehavior.ts
@@ -4,7 +4,7 @@ import type {
   TPointerEventInfo,
 } from '../../EventTypeDefs';
 import { Point } from '../../Point';
-import type { FabricObject } from '../Object/FabricObject';
+import type { FabricObject } from '../Object/Object';
 import { Text } from '../Text/Text';
 import { animate } from '../../util/animation/animate';
 import type { TOnAnimationChangeCallback } from '../../util/animation/types';
--- a/src/shapes/Image.ts
+++ b/src/shapes/Image.ts
@@ -29,7 +29,6 @@ import { NONE } from '../constants';
 import { getDocumentFromElement } from '../util/dom_misc';
 import type { CSSRules } from '../parser/typedefs';
 import type { Resize } from '../filters/Resize';
-import type { TCachedFabricObject } from './Object/Object';
 
 // @todo Would be nice to have filtering code not imported directly.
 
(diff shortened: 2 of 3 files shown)


src/shapes/IText/ITextBehavior.ts
```diff
@@ -4,7 +4,7 @@ import type {
   TPointerEventInfo,
 } from '../../EventTypeDefs';
 import { Point } from '../../Point';
-import type { FabricObject } from '../Object/FabricObject';
```
jiayihu, line 7: Gets later used in `onDeselect`. This means the problem occurred already in fabric itself, but the fix was not to use InteractiveObject here as well. The fix is to remove the type specialization in InteractiveObject's `onDeselect`

ShaMan123: This is incorrect, what is the problem it is causing?

ShaMan123 (changes requested):

src/shapes/IText/ITextBehavior.ts
jiayihu: Screenshot 2023-09-10 at 11 26 22

As I said, it's used for onDeselect

asturur:
So weird that TS has issue understanding classes with ease.
Is never easy to tell it `this argument is anything above class X in the prototype chain`

jiayihu:
To defend TS, I think the error makes sense as you can't assign InteractiveObject to FabricObject anymore once you've restricted onDeselect to accept only InteractiveObject. If I call `onDeselect({ object: new FabricObject() })` you're gonna run into an exception if you use any `InteractiveObject` property/method.
As I explained to Shachar, this is how typing works in any type system. 

In this case, there was no need for the restriction to begin with, so easy to fix. Otherwise, if you're really sure you're never gonna pass a FabricObject as InteractiveObject, then a type assertion `object as InteractiveObject` can be used. Indeed type assertions are unsafe.

asturur:
yes but in this case the FabricObject is the
complete class with also interactiveObject inside no? the basic one, the non interactive one should have no businnes with onDeselect. Does it?

ShaMan123:
> yes but in this case the FabricObject is the
> complete class with also interactiveObject inside no? the basic one, the non interactive one should have no businnes with onDeselect. Does it?

Definitely.

asturur:
i think understanding where the issues are is very important.
This is useful

ShaMan123:
hard to grasp...
can you explain why this doesn't solve the issue
```ts

// InteractiveObject says it accepts a InteractiveObject
onDeselect(options?: {
        e?: TPointerEvent;
        object?: FabricObject;
    }): boolean;

```

asturur:
But a type union won't fix it? or like multiple declarations of the method with different types?

jiayihu:
Using a union could bring to other issues and make reasoning harder. In the end, both types in InteractiveFabricObject and ITextBehaviour are reasoanable, so I pushed a different solution, that is explicitly overriding FabricObject's onDeselect to use FabricObject instead of falling back to InteractiveFabricObject's one.

PR #9326, opened 2023-09-09
Fix a regression in `wrap with fixed anchor`

### LINKED ISSUE #9258: [Bug]: The Shape's left position changed when I flip it from right side
### CheckList

### Version

6.0.0-beta12

Chrome

None

### Link To Reproduction

https://codepen.io/ritali4912/pen/JjwbPYG

### Steps To Reproduce

1.  Create a rectangle
2.  Click the mr control point then flip the rectangle **several times**. Notice: keep holding the mouse down during your several time flip from right to left then back to right then to left again, something like that.

P.S, it is correct on V5.

### Expected Behavior

The Left position should not change during I flip it from right side.

### Actual Behavior

At beginning:
image

After several times flip:

image

_No response_

asturur:
## Motivation

This is a regression fix from v5.
During PR https://github.com/fabricjs/fabric.js/pull/8400 we changed a transform.originX/Y consumed before and after a custom function, with a version destructured.
The custom function gets transform in the arguments and can mutate the originX and originY, so we need to stick to use the values that are on the transform object.

closes #9258

### DIFF
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -44,4 +44,17 @@ function animate(toState) {
     }
   );
 }
+
+canvas.add(
+  new fabric.Line([30, 30, 30, 150], { strokeWidth: 2, stroke: 'red' })
+);
+canvas.add(
+  new fabric.Rect({
+    strokeWidth: 2,
+    stroke: 'green',
+    width: 400,
+    height: 100,
+    fill: 'blue',
+  })
+);
 // animate(1);
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -44,17 +44,4 @@ function animate(toState) {
     }
   );
 }
-
-canvas.add(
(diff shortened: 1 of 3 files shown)


asturur:
requires a unit test

ShaMan123 (commented):
nice catch
I was sure I saw it when I tested in fabricjs.com

asturur:
unit test added, this is ready to go

ShaMan123 (commented):
looks fine, 
the test tells nothing to me tho.
playwright?

ShaMan123 (dismissed):

asturur:
I don't think we can't test everything with playwright.
Unit tests have their value to test function implementation, and we changed the implementation of this function without knowing.
wrapWithFixedAnchor does exactly 2 things and this tests checks it.
The ability to write messages on single expect assertion is lacking, probably i should add that plugin that adds it

asturur:
i think i know how i can improve the test to be more explicit on what we are testing

ShaMan123:
> The ability to write messages on single expect assertion is lacking, probably i should add that plugin that adds it

YES!!!
I was so annoyed by this.

asturur:
The test now is testing more clearly what is expecting.
Is not about the comments but the properties we are testing compared to what we are testing them against.

PR #9327, opened 2023-09-10
test(e2e): improve test case for line shifting and style with more colors

asturur:
## Motivation

Improve the test case with more colors and deletion to showcase the text issues that needs to be fixed in #9197 

I extended a the ObjectUtil class creating TextUtil class that contains a method to click the text at a specific point rather than guessing where.


e2e/tests/test.ts
```diff
@@ -13,6 +14,9 @@ type AsyncReturnValue<T> = T | Promise<T>;
 const setupTasks: Promise<void>[] = [];
 const teardownTasks: Awaited<VoidFunction>[] = [];
 
+// makes possible call things in browser context.
+window.fabric = fabric;
```
asturur, line 18: not going to do anything fancier than this. this works

asturur:
nothing to see here, improved a test.

e2e/tests/test.ts
ShaMan123: I am interestred where it is used

e2e/utils/TextUtil.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { IText, Point } from 'fabric';
+
+export class TextUtil<T extends IText = IText> extends ObjectUtil<T> {
+  async getCanvasCursorPositionAt(index: number): Promise<Point> {
+    return this.executeInBrowser(
+      (object, { index }) => {
+        const { left, top, topOffset, leftOffset } =
+          object._getCursorBoundaries(index, true);
+        // add extra 10 pixels to avoid controls
+        const point = new fabric.Point(
```
asturur, line 11: @ShaMan123 here i needed to use it.
I can't import Point here nor pass it as a context in executeInBrowser

ShaMan123: I see, thanks

PR #9329, opened 2023-09-12
fix(): Avoid firing twice in subselection

jiayihu:
In Group subselection, mouse events are fired twice because the target is both `target` and in `subTargets`. This is the quick fix, the proper one should probably be evaluating whether it's intended for the target to appear in both states.

cc @ShaMan123


ShaMan123:
Great catch, thanks! (saying it again)
In fabric terms this happens when group is  flagged with `interaction` and `subTargetCheck`.
It is part of the group rewrite I did not consider.
I will add a failing test so we have something to refer to.

Now question is how we should handle this.
IMO the target should be removed from `subTargets`.
I want to understand how` subTargets` is used.
In this case  `subTargets` wouldn't `subTargets contain only `target`? 
Running `searchPossibleTargets` on `target` might make sense. Then we align to current behavior.
If `subTargetCheck` is true `subTargets` will be the result or an empty array if not.

ShaMan123:
This is an opportunity to define what should happen and refactor `_searchPossibleTargets` so we stop using `Canvas#targets`

ShaMan123:
this helped find another severe bug

src/canvas/SelectableCanvas.ts
```diff
@@ -796,62 +796,59 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
-    const target = this._searchPossibleTargets(objects, pointer);
-    // if we found something in this.targets, and the group is interactive, return that subTarget
-    // TODO: reverify why interactive. the target should be returned always, but selected only
-    // if interactive.
-    return target &&
-      isCollection(target) &&
-      target.interactive &&
-      this.targets[0]
-      ? this.targets[0]
-      : target;
```
ShaMan123, line 854: this was wrong, added a test that fails on master

ShaMan123 (approved):
- `_searchPossibleTargets` => `fintTargetsTraversal`, now returns an array of hits
- fixed `searchPossibleTargets` bug cause d by confusion of `subTargetCheck` and `interactive`
- added tests

src/canvas/SelectableCanvas.ts
```diff
@@ -796,62 +796,59 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
    * @param {Array} [objects] objects array to look into
-   * @param {Object} [pointer] x,y object of point coordinates we want to check.
-   * @return {FabricObject} **top most object from given `objects`** that contains pointer
-   * @private
+   * @param {Object} [pointer] point coordinates to check
+   * @param {boolean} [param2.searchStrategy] strategy
+   * @returns {FabricObject[]} path of hits starting from **top most** object on screen.
    */
-  _searchPossibleTargets(
+  protected findTargetsTraversal(
```
jiayihu, line 806: What's the difference between this and `searchPossibleTargets`

ShaMan123: changed name to break itr cleanly

src/canvas/__tests__/eventData.test.ts
ShaMan123, line 1: dreadful work
but at least now when these tests fail it will have ameaning for a human being and it will take a second to understand what failed instead of trying to find in the massive dump something worthwhile

test/unit/canvas.js
ShaMan123, line 1: moved all findTarget tests to jest

test/unit/canvas_events.js
ShaMan123, line 1: moved to jest

ShaMan123 (commented):
OK
This is stabilizing.
I now think that `subTargets` should be renamed into `path` or `hits` or something since `target` can indeed be part of it, and the name doesn't reflect the fact that target's parent are part of the array as well.
The fix is done.

src/canvas/Canvas.ts
```diff
@@ -1500,27 +1483,22 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
         }
         this._fireSelectionEvents(prevActiveObjects, e);
       } else {
         isInteractiveTextObject(activeObject) && activeObject.exitEditing();
         // add the active object and the target to the active selection and set it as the active object
         activeSelection.multiSelectAdd(activeObject, target);
-        this._hoveredTarget = activeSelection;
-        // ISSUE 4115: should we consider subTargets here?
-        // this._hoveredTargets = [];
-        // this._hoveredTargets = this.targets.concat();
```
ShaMan123, line 1523: what is this?

src/canvas/Canvas.ts
```diff
@@ -82,13 +83,6 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
    */
   private declare _willAddMouseDown: number;
 
-  /**
-   * Holds a reference to an object on the canvas that is receiving the drag over event.
-   * @type FabricObject
-   * @private
-   */
-  private declare _draggedoverTarget?: FabricObject;
```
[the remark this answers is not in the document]
ShaMan123: I found where it was set, thought it was only read. but still redundant

ShaMan123 (commented):
refactored `_hovreredTarget`, `_hovreredTargets` => `hoveringState`
squashed `_fireEnterLeaveEvents`, `_fireOverOutEvents` => `handleSyntheticInOutEvents`

asturur:
For me this is on hold until the layout manager refactor, and after considering a dumb fix to avoid the double firing before considering other changes.

asturur:
> this helped find another severe bug

Which one?

src/canvas/SelectableCanvas.ts
asturur: Can you define in which way this is wrong?

ShaMan123:
I think first I will extract the migrated tests to a PR because it is impossible working with the old draggable text tests

src/canvas/SelectableCanvas.ts
ShaMan123: @asturur this.
If there is a target under a subTargetChec group under and interactive group the middle group should be selected but actually the target will be wrongly selected
see the test

asturur:
I would like to understand why this one:
https://github.com/fabricjs/fabric.js/pull/9329/commits/d64f256968d15aa1eba9e06f03a24b1bd54d18a7
the perfect commit, wasn't good enough.
Was this solving the bug or not?
Can't we have this and a test for this?

ShaMan123:
You can if you want to monkey patch and leave the real bug in the code base

asturur:
There are no real bugs and fake bugs.
@jiayihu described a bug and proposed a fix.
The other bug which is it? and why cannot be fixed in a different PR?

ShaMan123:
it can

ShaMan123:
Why can't we talk about the problem?
`searchPossibleTargets` populates `targets` and that was never considered with the group rewrite.
So now it returns targets including possibly the `target` that will be returned by the method and that is not a 2 line fix.
It is all across the event logic and there are no tests for it

asturur:
> Why can't we talk about the problem? 

It is exactly what i was asking, what the other bug is.

asturur:
So what is the impact of the bug and why you consider it severe? One is a double event firing, the other is an incorrect target? another double firing? Really isn't clear.

e2e/tests/text/drag&drop/index.spec.ts-snapshots/disabling-drop-A-events.json
ShaMan123, line 1: look here at the events recorded by the test
dragenter is called too much and dragleave too little - might be it is caused by this PR.

ShaMan123: I couldn't know what was wrong because of the draggable_text test that just fails without outputting something decent

ShaMan123:
Whatever relies on `targets` (e.g. synthetic events) is incompatible with v6.
I asked what is the definition of `subTargets`, can you answer that?

ShaMan123:
I reverted everything

ShaMan123 (commented):
restored

jiayihu:
Maybe we should go ahead with my simple and harmless fix, then take time to understand the implications of `subTargets` being populated with the target. That is the root logical inconsistency, that could provoke other bugs. At the same time it's better to avoid risky refactors until the situation is clear.

Once we all have a clear picture of what `subTargets` should and should not contain, then we can address properly. Spoken simply, we can wait to see if there are further bugs. The more bugs come in, the more we understand.

ShaMan123:
I have found more bugs.
That is why I dived in.
I am against monkey patching, especially in large repos like this.
This is a symptom, not the problem.
We need to fix the problem.
You say justly that chasing bugs is not the right way forward.
Or else as a dev I can't trust fabric. It is a nightmare to dev like this. Not being sure you can trust a simple method or an event behaving as expected.

As I said, this is part of the group rewrite. (Redoing coords is as well).
Since changes must be incremental this is the nature of changing. 

I never understood this part until now because how it is written. Scatterred changes in so many place I could not wrap my head around it.

I am now convinced subTargets should be removed in favor of a different namre first of all, hits explains best what that thing is.

Also take into account that since the group rewrite there is a valid case where you will want to hit a sibling of what was found. Currently, subTargets will contain only the parents. That is why I made it overridable.
I other words I find subTargets useless as it is.

ShaMan123:
## Rethinking and Reconsidering `subTargets`

If I go by the name and from what I understand was the original intention I believe `subTargets` should populate the sub targets of the found target.
Parents of the found target should not be part of `subTargets`, that is what makes the name confusing. They should not be part mainly because it is clear that they are in the hit region or else the found target would not be found. If anyone including fabric needs the parents it is very simple to walk up the tree.

The main use in fabric of `subTargets` is for events. So we need to fix that.
I think it is unclear what is expected from the synthetic events.
Should enter/exit events fire on parents as well and on descendants? I think they should.
That is an easy fix.

ShaMan123:
Take a look at #9343 
It is a good and small fix for the underlying issue.

jiayihu:
> I have found more bugs.
> That is why I dived in.

I'll ask it for @asturur: what are the bugs?

ShaMan123:
- where events are fired on subtargets this will happen (e.g. drag events)
- synthetic events are wrong, firing incosistentantly
- event data using `subTargets` is wrong - who knows how apps you it - the name itself it wrong the way it is currently. I would expect subTargets to be the subTargets of the event target and they are not, so why `subTargets`. Call them `targetsFoundWhileSearching`, `path, `hits`
- `_setCursorFromEvent` uses it

asturur (approved):

PR #9333, opened 2023-09-13
test(): migrate tests to playwright/jest

ShaMan123:
## Motivation

#9329 

## Description

Working on #9329 unconvered many ubgs.
Trying to fix them made tests fail.
These tests are not human friendly, they dump endless output onto the console so they are useless.
To be percise `draggable_text` was written by me while grinding my teeth. It was hard and I knew it was incorrect to write it like that.
So I have moved  `draggable_text` to e2e as it should be, removing redundancies.
And I have moved all the `findTarget` related tests to jest.


e2e/tests/text/drag&drop/index.spec.ts-snapshots/4-events.json
```diff
@@ -0,0 +1,191 @@
...
+        "isTrusted": true
+      },
+      "target": "b",
+      "subTargets": [],
+      "dragSource": "a"
+    },
+    "canvas"
+  ],
+  [
+    "dragleave",
```
ShaMan123, line 35: weird, a bug.
after dragstart nothing has entered yet

e2e/tests/text/drag&drop/index.spec.ts-snapshots/5-events.json
```diff
@@ -0,0 +1,191 @@
...
+        "x": 120,
+        "y": 55
+      },
+      "absolutePointer": {
+        "x": 120,
+        "y": 55
+      }
+    },
+    "b"
+  ],
```
ShaMan123, line 55: here canvas says b is entering while b is saying it is leaving

ShaMan123 (commented):
reverted, now only tests are here and #9329

ShaMan123 (commented):
@asturur please merge this
The tests in this PR are much more freindly and reliable - I have PRs with tests that are failing, tests that are migrated here.

ShaMan123 (commented):
It seems that tests are failing becuase of font differences.
We need to register the font - the code you removed from the other CI PR

e2e/tests/text/drag&drop/index.ts
```diff
@@ -77,41 +77,100 @@ const styles = {
...
+  caller: fabric.Textbox | fabric.Canvas
+) =>
+  JSON.parse(
+    JSON.stringify([
+      type,
+      {
+        ...ev,
+        ...(pointer ? { pointer: roundPoint(pointer) } : {}),
+        ...(absolutePointer
+          ? { absolutePointer: roundPoint(absolutePointer) }
```
ShaMan123, line 97: scenePoint/viewportPoint

asturur (approved):

asturur:
I ll make an issue with a followup for this PR since the the position of the mouse should be refactored with the proper text utils where possible.
Also one test for me doesn't pass locally

ShaMan123:
> I ll make an issue with a followup for this PR since the the position of the mouse should be refactored with the proper text utils where possible. Also one test for me doesn't pass locally

My point of view on this is that we should use hardcoded position because it makes the test strict.
If we use a function that depends on what we are testing then the test becomes weaker, don't you think?

PR #9341, opened 2023-09-14
fix(IText): cursor width under group

ShaMan123:
## Motivation

@jiayihu 

Found while editing text in a scaled group

## Description

Text cursor width doesn't take into account group transform.
Another miss of the group rewrite.
I saw that draggable text has the same issue

### DIFF
--- a/src/shapes/IText/__snapshots__/IText.test.ts.snap
+++ b/src/shapes/IText/__snapshots__/IText.test.ts.snap
@@ -1,0 +1,56 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`IText cursor width under a group scaled by 1 and canvas zoomed by 1 1`] = `
+[
+  [
+    -92.2,
+    -18.6,
+    100,
+    40,
+  ],
+]
+`;
+
+exports[`IText cursor width under a group scaled by 1 and canvas zoomed by 50 1`] = `
+[
+  [
+    -43.2,
+    -18.6,
+    2,
+    40,
+  ],
+]
+`;
+
+exports[`IText cursor width under a group scaled by 200 and canvas zoomed by 0.005 1`] = `
+[
+  [
+    -92.2,
+    -18.6,
+    100,
+    40,
+  ],
+]
+`;
+
+exports[`IText cursor width under a group scaled by 200 and canvas zoomed by 1 1`] = `
+[
+  [
+    -42.5,
+    -18.6,
+    0.5,
+    40,
+  ],
+]
+`;
+
+exports[`IText cursor width under a group scaled by 200 and canvas zoomed by 50 1`] = `
+[
+  [
+    -42.2,
+    -18.6,
+    0,
+    40,
+  ],
+]
+`;
--- a/src/shapes/IText/__snapshots__/IText.test.ts.snap
+++ b/src/shapes/IText/__snapshots__/IText.test.ts.snap
@@ -1,56 +1,50 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`IText cursor width under a group scaled by 1 and canvas zoomed by 1 1`] = `
-[
-  [
-    -92.2,
-    -18.6,
-    100,
-    40,
-  ],
-]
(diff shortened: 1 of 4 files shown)


src/shapes/IText/__snapshots__/IText.test.ts.snap
```diff
@@ -0,0 +1,56 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`IText cursor width under a group scaled by 1 and canvas zoomed by 1 1`] = `
+[
+  [
+    -92.2,
```
jiayihu, line 6: Aren't you a fan of Playwright? At least you could catch a visual regression. These kind of tests instead easily break if the value changes only a little. A better test, in Jest, would be checking that cursors at different group scale have the same size, whatever is the value.

asturur: I agree snapshots tests shouldn't be used so largely.

asturur:
Those are all the artifacts of interactive group that insist was a bad idea and should be marked beta also with 6.0 release.

asturur:
The fix is totally fine, i want to think of a nicer way to test this, also with more cases.
This should be quick and is a bug, so i ll bump it up in my list

src/shapes/IText/__snapshots__/IText.test.ts.snap
ShaMan123: Well guys I first tried the playwright test and it won't be percise enough to catch regression I belive.
I can add it, it is stashed on my local branch and try to refine it... but I don't see how.



> I agree snapshots tests shouldn't be used so largely.

I prefer snapshots instead of hardcoding because then it is easy to update and see the diff. The main pro of jest IMO.

ShaMan123:
> Those are all the artifacts of interactive group that insist was a bad idea and should be marked beta also with 6.0 release.

@jiayihu can maybe share what v6 is doing to the project we are working on.
You should have seen the amount of impossible workarounds removed due to interactive groups.
I can't defend it more than I have.

jiayihu:
Group subselection is indeed a common feature of canvas applications, think about draw.io, Miro, tldraw.com etc where you can group objects but still want to subselect. Think about photoshop/sketch/Figma layers as well, all groups. 
The way I see a Group is basically like a DIV, you have many reasons to use it as wrapper while still wanting the items to be selectable/interactive.

asturur:
pulling in and out object in and from groups was easy enough after the first tweak @ShaMan123 did to the group that persisted the group transform when adding/removing.
All the rest is superflous

asturur:
now that i added the angles and i don't see the results i expect, i want to see it in action

asturur:
It behaves mostly ok also with weird situations of angle and scale. It never look broken

asturur (approved):

src/shapes/IText/__snapshots__/IText.test.ts.snap
jiayihu: > I prefer snapshots instead of hardcoding because then it is easy to update and see the diff. The main pro of jest IMO.

I suggested something like `expect(getCursoWidth(renderWithScale1)).toEqual(getCursoWidth(renderWithScale2))`, then you don't need to hardcode anything or even care about what is the value

> it is easy to update and see the diff. The main pro of jest IMO.

Jest snapshots are definitely a nice feature, but as everything you have to use it properly. They have drawbacks:
1. Artifacts like the snapshot files pollute the codebase
2. People tend to **too easily** just update a diff thinking it's okay because you can just press a keystroke `u` and voilà, the test passes. A normal test forces you to look into the problem and really assert whether it was a false negative.

I've been using Jest since React 0.14, trust me a bit 😄

asturur: yes my concern is indeed the easyness of update.
In this case i slimmed down what we are snasphotting so that it makes clearer what are we update.
In the case like before with an array of 4 numbers, if we change the way we draw cursor, like we just center it in place, the first 2 numbers of the snapshot would have changed, but with no indication that those are the values of x,y of fillRect.

In generale is preferable to have tests that says:
expect(mockCalls[0][2], 'width is exactly 100').toBe(text.cursorWidth);

or something in that ballpark

PR #9342, opened 2023-09-15
feat(FabricObject): attached/detached events

ShaMan123:
## Motivation

There are cases where an objects needs the canvas ref to do some work, such as subscribing to events.
The `added` event is not sufficient if we are dealing with nested objects.

This PR is well related to the active selection initialization process because before recent changes initialization would be done with the canvas ref passed in options and now you don't need to do that and maybe even can't (if you pass the active selection ref to the canvas constructor as options) so the event is needed for the change to be less breaking for apps that relied on the canvas ref. Meaning you simply add an `attached` event listener and do your work there,. That would be the migration path.

Also I have discussed this with @jiayihu who was strongly urging fabric to have some kind of trigger for object initialization that is not the constructor. This is needed for several use cases.
Using subclassing sometimes makes the constructor very hard to deal with because you can't hook into it. Having a lifecycle in the contructor is not an accepted practice so it seems, I am guessing because of the order of execution some properties might not be set yet etc., let alone the TS `declare` nightmare.
A good example is Text classes. They call `initDimensions` in the constructor. It is reasonable to assume that a subclass of Textbox will want to do some work before calling that but it can't. This PR doesn't fix this use case but it provides a workaround saying something like "The first call is a waste, nevermind, I won't call it or use it until the object is attached to the canvas, then I am sure I can subclass even more and not face this issue".
BTW, fabric faces this issue (caused also by using `set` in the super class) and deals with it by adding a flag `initialized` which I know @asturur does not like.
This is another good example of a use case that can't be fixed with this approach but can be worked around somehow.

## Description

When the canvas ref changes the attached/detached events are fired.
Making it easy to do related work.
These events have a meaning also in terms of rendering.
When they fire you know for sure the object will render or will stop rendering.

## Changes

I patched an active selection test to make it more complete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [next]
 
+- feat(FabricObject): attached/detached events [#9342](https://github.com/fabricjs/fabric.js/pull/9342)
+
 ## [6.0.0-b3]
 
 - fix(Textbox): implemente a fix for the style shifting issues on new lines [#9197](https://github.com/fabricjs/fabric.js/pull/9197)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - feat(FabricObject): attached/detached events [#9342](https://github.com/fabricjs/fabric.js/pull/9342)
 
-## [6.0.0-b3]
+## [6.0.0-beta13]
 
 - fix(Textbox): implemente a fix for the style shifting issues on new lines [#9197](https://github.com/fabricjs/fabric.js/pull/9197)
 - Fix(Control) fix a regression in `wrap with fixed anchor`, regression from #8400 [#9326](https://github.com/fabricjs/fabric.js/pull/9326)
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -250,6 +250,8 @@ export interface ObjectEvents
   // tree
   added: { target: Group | Canvas | StaticCanvas };
   removed: { target: Group | Canvas | StaticCanvas };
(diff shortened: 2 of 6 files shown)


ShaMan123 (commented):
ready

ShaMan123 (commented):
now tests should pass

asturur:
sublassing is bad, fiddling with the prototype is great.
Now after this pearl of wisdom, i ll review this PR as soon as possible, but after the others

asturur:
I read only the description and i think i would like to know which are the user driven actions that can change a canvas reference to an object.
Can you elaborate on that?

jiayihu:
I'll try to explain my original rational @asturur : running side effects in the constructor is very inconvenient:

1. From a good programming POV, constructors should not have side effects and should be pure. Example: registering to events could be dangerous as the object could be thrown away without being added to the canvas, but your listeners are now in the memory forever if you're not careful
2. From a practical POV, it's hard for a subclass to customise side-effects like `initDimensions` because the subclasss properties are not defined yet during the parent constructor call. This is not only non-evident, especially for a junior/mid-distracted dev, but even results in dark magic tricks to make it work

ShaMan123:
> I read only the description and i think i would like to know which are the user driven actions that can change a canvas reference to an object. Can you elaborate on that?

No user actions.
I have an object. It will be added to a group/canvas in an uncertain point of time.
I want to add an event listener subscribing to a canvas event and run some scoped logic on the opbject.
I do not want to run logic checking when this object is added because I am not sure where it will be added, it makes my app very messy.

ShaMan123:
We can rename to mounted/unmounted
related to #9550

PR #9343, opened 2023-09-15
fix(): `searchPossibleTargets` `targets` value

ShaMan123:
## Motivation

Trying to impl a fix for the problem presented in #9329 
Being that the group rewrite made `searchPossibleTargets` return a sub target instead of the found taregt to allow deep selection. This surfaced an issue that the use of `Canvas#targets` aka `subTargets` is wrong and confusing.

closes #9329 

### Rethinking and Reconsidering `subTargets`

If I go by the name and from what I understand was the original intention I believe `subTargets` should populate the sub targets of the found target.
Parents of the found target should not be part of `subTargets`, that is what makes the name confusing. They should not be part mainly because it is clear that they are in the hit region or else the found target would not be found. If anyone including fabric needs the parents it is very simple to walk up the tree.

The main use in fabric of `subTargets` is for events. So we need to fix that.
I think it is unclear what is expected from the synthetic events.

## Changes

- `_searchPossibleTargets` => `findTargetsTraversal` + added a flag to allow searching for all hits
- fix `searchPossibleTargets` return value to consider an case where `subTargetCheck: true, interactive: false` group were travesred, so returning the first hit is wrong
- dep  `searchPossibleTargets` => `findTargets`
- move setting `targets` to `findTarget` because it varies on the case and is a step towards refactoring the use of `Canvas#targets`

## Gist

The tests came from #9333 if you wish to merge that first, conflicts should resolve to `ours`

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -712,7 +712,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
 
       // check pointer is over active selection and possibly perform `subTargetCheck`
       const { target: selectedTarget, targets: selectedTargets } =
-        this.searchPossibleTargets([activeObject], pointer);
+        this.findTargets([activeObject], pointer);
 
       if (selectedTarget && aObjects.length > 1) {
         // active selection does not select sub targets like normal groups
@@ -728,7 +728,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
           );
           return activeObject;
         } else {
-          const { target, targets: canvasTargets } = this.searchPossibleTargets(
+          const { target, targets: canvasTargets } = this.findTargets(
             this._objects,
             pointer
           );
@@ -754,10 +754,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
       }
     }
 
-    const { target, targets } = this.searchPossibleTargets(
-      this._objects,
-      pointer
-    );
+    const { target, targets } = this.findTargets(this._objects, pointer);
     this.targets = target ? targets.slice(0, targets.indexOf(target)) : targets;
     return target;
   }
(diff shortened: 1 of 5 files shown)


src/canvas/__tests__/eventData.test.ts
ShaMan123, line 1: Migrated all `findTarget` tests, see 0c0429255
I have decided not to remove the old ones yet so we prove the PR passes (eventhough I did a thorough job of migrating the tests). I kept the names of the tests the same to make it easy to trace back.
c1448b000 contains new tests about my concerns raised in the review

src/canvas/__tests__/eventData.test.ts
```diff
@@ -100,3 +112,790 @@ describe('Canvas event data', () => {
...
+      canvas.add(other, target);
+      activeSelection.add(target, other);
+      canvas.setActiveObject(activeSelection);
+
+      jest.spyOn(canvas, '_checkTarget').mockReturnValue(true);
+      const found = canvas.searchPossibleTargets([target, other], new Point());
+      expect(found).toBe(other);
+      expect(canvas.targets).toEqual([subTarget, target, other]);
+    });
+
```
ShaMan123, line 399: this is my concern

ShaMan123 (commented):
This is a good incremental fix but there are a few more things to address here.
Even though all tests pass it does not mean that event firing is correct.
Seems there are no tests covering these cases.

ShaMan123 (commented):
- dep  `searchPossibleTargets` => `findTargets`
- move setting `targets` to `findTarget` because it varies on the case and is a step towards refactoring the use of `Canvas#targets`

Now I am happy with this PR

src/canvas/__tests__/eventData.test.ts
ShaMan123: fixed by 40c725b

src/canvas/SelectableCanvas.ts
```diff
@@ -796,62 +793,72 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
-      : target;
+    const targets = this.findTargetsTraversal(objects, pointer, {
+      searchStrategy: 'first-hit',
+    });
+
+    const found = targets.findIndex((target) => {
+      return (
+        !target.group ||
+        (!(target.group instanceof ActiveSelection) && target.group.interactive)
+      );
```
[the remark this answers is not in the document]
ShaMan123: fixed and tested by 40c725b

ShaMan123:
Ported to https://github.com/ShaMan123/fabric.js/pull/5
Contributions are welcome

PR #9349, opened 2023-09-16
refactor(Group): `parent` + fix(ActiveSelection): transferring object

ShaMan123:
## Motivation

I was looking into the patch I made into #9336 and understood that `getParent` was a design mistake.
It is confusing and silly. I wrote it in a haste, not wanting to create too many changes.
It is a design mistake IMO mainly because there is a confusion about active selection - should it be considered a parent or not.
Apps might need to access both the the current `group` of an object and both the owning group.
Since `group` and `canvas` are used it seems bad that to get the owning parent you do `getParent(true)`.
I regret it.
This PR makes it simpler both for fabric and for devs and removes ambiguity of the `group` property.
**Please accept it before v6 so it only breaks beta.**

## Description

This PR introduces the `parent` property on `Object` and removes `getParent`.
The `parent` property is the owning group of an object, meaning not the active selection.

While testing the PR I realized I didn't take into consideration another part of transferring an object between active selections that I missed in #9336
The object should be removed from the active selection and not be reomved from its owning group.

related to #8951

## Changes

BREAKING beta: rm `getParent`


src/shapes/ActiveSelection.ts
```diff
@@ -79,17 +79,18 @@ export class ActiveSelection extends Group {
...
-      const parent = object.group;
-      parent._exitGroup(object);
-      object.__owningGroup = parent;
+    // make sure we exit the parent only once
+    if (object.parent && object.parent === object.group) {
+      // keep the object part of the parent
+      object.parent._exitGroup(object);
+    } else if (object.group && object.parent !== object.group) {
+      // in case `object` is transferred between active selections we remove it from the prev
+      object.group.remove(object);
```
ShaMan123, line 90: this is this fix to active selection

ShaMan123:
I need to assess if this should be merged into #8951 or done on top of it but they are duplicates to some extent

ShaMan123:
closing => sqaushed into #8951

PR #9350, opened 2023-09-16
CD(): playwright codegen

ShaMan123:
## Motivation

We want tests to be easy to write.
I saw you commented on this https://github.com/microsoft/playwright/issues/21504 that requests support for mouse events etc.
I looked into it and decided to impl a simple solution for fabric

## Description

setupCodegen is fabric's version for playwright's codegen.
It adds event listeners and generates playwright code when the test is done and attachs it to the test results.

### Usage

- start the template ` npm run test:e2e -- e2e/tests/template/index.spec.ts`
- in the chrome devtools exec `startRecording()`
- play with the canvas
- add a step by calling `step('step name')`
- take a screenshot by calling `captureScreenshot()`
- hit the play button of the playwright devtools
- the generated code will be logged to the console and attached to the test results

## In Action

image


ShaMan123 (commented):
This is a game changer

package.json
```diff
@@ -144,6 +144,12 @@
       ],
       "src/*": [
         "src/*"
+      ],
+      "test/playwright": [
+        "e2e/setup/index.ts"
+      ],
+      "test/jest": [
+        "jest.setup.ts"
```
ShaMan123, line 152: now consumers can use it

ShaMan123: but I don't knowe how to make typescipt auto complete `fabric/tes` => `fabric/test/..`

ShaMan123 (commented):
I love this.
Testing will become a lot easier with this.

ShaMan123:
Now that I understand better playwright I think we should configure fixtures if we can, instead on before/after hooks

asturur:
Sorry  i was sure i posted a comment here weeks ago.
Can't this become a plugin and leave as its own library? isn't this reusable outside fabricJS?

ShaMan123:
I guess,
So then I install it on the repo?

ShaMan123:
Do you want it on the org or shall I put it on my account?

asturur:
In my opinion this is very useful for brush tests or generic test for which we can't use center of objects or text positions.
In my opinion you should make a plugin, your repo or the org repo is your preference.
There is chance this is very useful to more people.

ShaMan123:
closed in favor of https://github.com/ShaMan123/playwright-codegen

PR #9352, opened 2023-09-17
Layout constants

asturur:
Swapping constants on layout-manager

### DIFF
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -29,30 +29,8 @@ const text = new fabric.Textbox(textValue, {
     textValue
   ),
 });
-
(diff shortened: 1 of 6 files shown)


ShaMan123 (changes requested):
Why not doing
```
isImperative(context)
isInitialization(context)
```

clear and saves more

asturur:
i thought about it but i wanted to keep the code as it is it doesn't matter.
What it matters is to do not repeat the same constants around all the time. Function or constant doesn't change much to me

ShaMan123:
I prefer an enum instead.
Any objections?

asturur:
Actually
 ill show you my ideal enum at this point.

asturur:
i ll reopen the PR

PR #9356, opened 2023-09-19
docs() enable typedocs to run again

asturur:
## Motivation

typedoc does not run anymore.
It complains of some code inside tests.
I changed the code of tests and jest extend and removed the protected from toObject.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,6 +36,7 @@ module.exports = {
     ],
     'no-restricted-syntax': [
       'error',
+      // explore how to define the selector: https://astexplorer.net/
       {
         selector: '[callee.object.name="Math"][callee.property.name="hypot"]',
         message:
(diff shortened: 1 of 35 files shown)


ShaMan123 (dismissed):
can't it ignore `*.test.ts` `*.spec.ts`?

ShaMan123 (commented):
what about an action for a PR merge to push to the website?
Then we see it actually runs and the website is kept to date

asturur:
> can't it ignore `*.test.ts` `*.spec.ts`?

Dosen't work.
It wants all TS to be good for some dumb reason.

asturur:
> what about an action for a PR merge to push to the website? Then we see it actually runs and the website is kept to date

When it works it can be done, for now it doesn't work at all, the pages are ugly, creating a custom theme for mark down is not very well documented ( or docs are hard to find ) and gatsby adds its own issues with mdx and ts syntax overlap.

asturur:
probably we should add a step that run the docs and that we check ti can run at least. So we can detect breakage waiting for the time that we can dig into why excluding doesn't really work

ShaMan123 (approved):
fixed the config, it skips the excluded files now - tested in a branch w/o the fixes from this branch

Issue #9355, opened 2023-09-18
[Bug]: Invisible text when a gradient or pattern is applied


u-blusky:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

5.3.0

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

None

### Link To Reproduction

https://jsfiddle.net/yrpuqvec/

### Steps To Reproduce

1. Create Text object
2. Apply gradient
3. The text is invisible

### Expected Behavior

Text fill with gradient and visible

### Actual Behavior

![gradient_text_chrome_problem](https://github.com/fabricjs/fabric.js/assets/131244197/100904da-91de-4b60-a4ed-cc100e9f708d)
![gradient_text_firefox_ok](https://github.com/fabricjs/fabric.js/assets/131244197/1d1124f5-78f7-4a0e-9aff-83245875a0cb)

ShaMan123:
I think it is a priblem is chrome, our tests also fail in this aspect.
Try seeing if toggling objectCaching makes adifference

u-blusky:
Hi,
I tried with objectCaching, statefullCache, dirty but, unfortunately, it doesn't solve the problem.

u-blusky:
Maybe I found:
In Chrome, Use hardware acceleration when available = false
and the text is drawn correctly.

tonycaso:
changing gradientUnits to pixels makes text visible(no need to disable objectCaching), but gradient isn't displayed correctly

ShaMan123:
Thanks for reporting
Reproduction will be great, a pr even more

u-blusky:
@tonycaso, @ShaMan123 
thanks a lot.
Using "pixels" works.
I created a new fiddle to test this:
https://jsfiddle.net/y30ge8fx/1/

u-blusky:
Unfortunately the "linear" type is not right for how I set up the code.
[...]

PR #9359, opened 2023-09-20
fix(): gradient transform + rm workarounds

ShaMan123:
## Motivation

#9355 

I was invetigating the bug and saw that the cause was the `_applyPatternGradientTransformText`
When I wrote #8263 I tried to fix this thing with a math solution and now I managed to.

## Description

Previously in order to apply transform on a pattern/gradient fabric would create a canvas, transform it acoordingly and render a rect.
Then it would use the canvas to create a pattern and set that to the rendering context as fillStyle/strokeStyle.

- CanvasPattern supports natively `setTransform` so I used that instead
- `gradientTransform` is supported only for linear gradients (don't know why but I am fine with it because then my solution is valid) so I applied the transform to the coords
- offsetX/Y are a transform so I applied that also when setting the trasform on the pattern/gradient

This is a very good change.
It removes a workaround that was complex, had bugs and was mentioned as a perf hit

## Changes

Moved all pattern/gradient transform logic to `toLive`
BREAKING:
- `toLive(ctx) => `toLive(ctx, target)`
-


ShaMan123 (commented):
progress

ShaMan123:
radial gradients must use a pattern :(

asturur:
radial gradients don't support transform?

ShaMan123:
not natively

ShaMan123:
all gradients do not but I have found a math solution for the linear gradient transform i think

ShaMan123:
Correct me if I am wrong but it seems (radial) gradient needs context isolation to be transformed or else if transforming the context text for example will be transformed as well.
So, I am not sure fabric is ready for me to propose my solution.
And this is why a pattern was used.
In general I think we should pass a context in rendering and supply a canvas for context isolation operations

asturur:
for sure there is a math solution to transorm linear gradients because are just points in a line and we either squeeze the points or change the angle and it will look identical.

But for radial gradients when circle becomes ellipses no you need to transform those.
Radial gradients are not so popular, if we have to keep around the old code for radial gradients is a pain but at least we solve the slow path of the linear gradients that are more common.

PR #9361, opened 2023-09-20
fix(StaticCanvas): disposing animations

ShaMan123:
## Motivation

> Why not instead abort any pending request? After all, there may even be other async stuff pending, such as animations scheduled with `requestAnimationFrame`, other than `canvas.requestRenderAll()`. I was thinking today that maybe we could dispose any pending request done with `fabric.util.requestAnimFrame`, to also have a reason to use that vs the native one.

This is done in dispose. There is an option when creating an animation to attach a target property, later on dispose calls the `cancelByCanvas` method on the running animations array. **It should have at least. Now Isee it doesn't.**

_Originally posted by @ShaMan123 in https://github.com/fabricjs/fabric.js/issues/8299#issuecomment-1728542235_
            

## Description

Dispose running animations that are targetted to the canvas

### DIFF
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -24,6 +24,7 @@ import {
   cancelAnimFrame,
   requestAnimFrame,
 } from '../util/animation/AnimationFrameProvider';
+import { runningAnimations } from '../util/animation/AnimationRegistry';
 import { uid } from '../util/internals/uid';
 import { createCanvasElement, toDataURL } from '../util/misc/dom';
 import { invertTransform, transformPoint } from '../util/misc/matrix';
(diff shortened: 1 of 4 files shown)


ShaMan123 (commented):
DONE

src/canvas/StaticCanvas.ts
```diff
@@ -1440,6 +1441,7 @@ export class StaticCanvas<
   dispose() {
     !this.disposed &&
       this.elements.cleanupDOM({ width: this.width, height: this.height });
+    runningAnimations.cancelByCanvas(this);
```
asturur, line 1444: isn't better to cancel animation before disposing the dom?

asturur:
well is a line we can swap later is not worth holding up

Issue #9324, opened 2023-09-09
Finalize object.type


jiayihu:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have searched and referenced existing issues, feature requests and discussions
- [X] I am filing a **FEATURE** request.

### Description

With fabric v6, `object.type` has been deprecated in favour of `object.constructor.type` but I believe this was a bad decision. `object.type` remains a very handy property:
- To know the object type without going through the constructor class. I find it ugly and very OOP-style, whereas in JS it's more natural and solid to check "by value". If you have `object.type === "duck"`, you are a Duck
  - Also in TS, `constructor` can be any runtime function and the type system can't know it so it's just `Function`, resulting in uglu `(this.constructor as typeof FabricObject).type`
-  More importantly, when an object is serialized you still need the `type` of course since all the constructor info is lost. Then when you create a serialized object, you have to manually filter out `type` to avoid the warning `Setting type has no effect`. This is very annoying and ironically even fabric doesn't do it. Just try `object.clone()`
- Confusion in the codebase as both variants are used. The fabric code itself still uses a mix.

So I propose that we revert to the good old `object.type`. I don't see any special advantage with the static property in the class.

### Current State

.

### Additional Context

cc @ShaMan123

asturur:
We did many bad decisions on the trail of modern code and TS.
And TS is ugly no matter what you do or how you write it, it makes JS ugly

The only thing we wanted to obtain was to have type read only, and the setter was added  to avoid the breaking error.
We should remove type from every Object fromObject toObject cycle, and in some we did already.

I agree the warning is annoying but is just a warning, and it can be fixed.
Bringing back type as normal property is not convenient because no one should change it
the getter seems the best compromise, the warning can be minimized a lot by making sure fabric doesn't try to set type.

one thing is sure, the old type for release of 6.0 needs to be there because many application do type inspection by doing instance.type and removing it is a big headache.
And the class.name can't be used because minification kills it.

ShaMan123:
We can use `defineProperty` with writable set to false. Then is throws. Even more annoying but a good breaking change. Easy to migrate.
I think fabric should stop using type comlpetely and acess the static type directly. It is needed only in serialization so not a lot of places.
Then we should remove things like `isType` and getting objects by type.
[...]

PR #9368, opened 2023-09-23
refactor(v6): remove instance `type` property + `isType`

ShaMan123:
## Motivation

Previous discussions about `type` in v6.
At the beginning we changed `type` to a getter (and added a spamming setter to warn apps of the change) returning the class name for 2 reasons:
- the class regisrty would pick it up without the dev needing to configure anything when creating a subclass
- because we wanted to save time and patch the change instead of going throw the code base

Then minifying made us use the static `type` property because it mangled class names.
However we didn't remove the `type` getter/setter, though we could have.
Apparently apps use the instance property. I am against.
The migration path is to set the static `type` value for serialization.
The rest is deprecated.

closes https://github.com/fabricjs/fabric.js/issues/9324

## Description

I have made fabric use only the static `type` property.

## Changes

BREAKING:
- rm `isType` - if a dev wants it they can create a util for it.
- rm `getObjects(...type)` signature => `getObjects()`
- Serializable fabric classes do not accept `type` as an option, it will be dicarded in the constructor. Again, devs can solve this problem or we can safegaurd all fromObject instead of the constructor - that sounds like a viable option


src/util/typeAssertions.ts
ShaMan123, line 1: I have PRs open to remove this bad design.
Why is it so bad? I subclass one of these classes - I MUST keep the type value the same so to for serialization to work as expected I will have to override `toObject`. This is an unneeded gotcha.

ShaMan123 (commented):
Once all tests pass on d57969989 it will mean that fabric is indeed safe and I will remove the following safeguard
```
  get type() {
    throw new Error('get type');
  }

  set type(value) {
    throw new Error('set type');
  }
```

src/gradient/Gradient.ts
```diff
@@ -193,16 +193,16 @@ export class Gradient<
...
-      offsetX += object.width / 2;
-      offsetY += object.height / 2;
+      offsetX += width / 2;
+      offsetY += height / 2;
     }
     // todo what about polygon/polyline?
-    if (isPath(object) && this.gradientUnits !== 'percentage') {
-      offsetX -= object.pathOffset.x;
-      offsetY -= object.pathOffset.y;
+    if (pathOffset && this.gradientUnits !== 'percentage') {
```
ShaMan123, line 203: what about this comment?

ShaMan123 (commented):
tests should pass now

src/gradient/Gradient.ts
ShaMan123: ` // todo what about polygon/polyline?`

ShaMan123 (commented):
All tests have passed with the strict safeguard throwing errors - aa243ed
Now removed it

ShaMan123:
Actually now we can discuss and decide if to allow devs to use `type` on an instance because fabric doesn't use it anymore and I wrote an extensive test covering all initialization usages.

In v5 changing the instance `type` prop would affect `toObject` and maybe even `fromObject` because it was on the prototype.
No it will not.
That is why I think we should not allow passing down `type`.
In such a case it is needed the app should override and subclass anyways so better we are strict about it.

ShaMan123:
anyways it is easy to deal with now and should be done in a follow if we decide to.
So this is ready to merge

asturur:
i m sure this wasn't the outcome of the discussion.
we can't remove the getter, nor kill an app because is trying to set something that doesn't have a setter.
so the getter and the setter should stay.

asturur:
Where is the original chat? i can't find it anymore

ShaMan123:
> i m sure this wasn't the outcome of the discussion. we can't remove the getter, nor kill an app because is trying to set something that doesn't have a setter. so the getter and the setter should stay.

If fabric doesn't use type anymore what diffreence does it make. It becomes as any property.
What was the outcome then?
At least we can agree that fromObject should remove the type property before passing options to constructor?
And why kill an app? the current situation kills apss that use type.

ShaMan123 (commented):
x

asturur:
Do you have a link to the old thread? is hard to remember what was said about this being an issue.

ShaMan123:
https://github.com/fabricjs/fabric.js/issues/9324

src/util/typeAssertions.ts
```diff
@@ -41,29 +40,36 @@ export const isCollection = (
...
 };
 
 export const isInteractiveTextObject = (
   fabricObject?: FabricObject
 ): fabricObject is IText | Textbox => {
   // we could use instanceof but that would mean pulling in Text code for a simple check
   // @todo discuss what to do and how to do
-  return !!fabricObject && fabricObject.isType('IText', 'Textbox');
+  return (
+    !!fabricObject &&
```
zhe-he, line 68: If I create a class TextBox2 that inherits from TextBox, it won't match here.  isPath and isTextObject also have the same issue. Would it be better to write it like this?
```
fabricObject instanceof IText && !(fabricObject instanceof Text)
```

ShaMan123: instanceof is the best solution not doubt
but it causes import cycles so we can't use it
this is part of the reason why I want to remove all these utils as done in #8950

Smrtnyk:
so basically when we are subclassing and have custom toObject, we should do it like this
```typescript
toObject() {
    return {
        ...super.toObject(),
        type: "custom-type"
    }
}
```
so that when reviving serialized object it gets picked up by this type `custom-type` we set
but we should not set `type` class property anymore?

ShaMan123:
That is correct
However the best practice is
```
class CustomObject extends FabricObject {
  static type = 'custom-type'
}
classRegistry.setClass(CustomObject)
```

It will handle both `toObject` and `fromObject` out of the box

PR #9369, opened 2023-09-23
chore(): cleanup logs and error messages

ShaMan123:
## Motivation

After another dev wrote about the duplicate canvas init error I uderstood the message was not good enough so I decided to open a patch for that.
Then I decided to have all logs and errors in one place, I remember @asturur said something about that.

## Changes

- exposed a `log` internal util
- exposed Error subclasses
- added lint to force usage

### DIFF
--- a/src/canvas/DOMManagers/StaticCanvasDOMManager.ts
+++ b/src/canvas/DOMManagers/StaticCanvasDOMManager.ts
@@ -4,6 +4,7 @@ import type { CSSDimensions } from './util';
 import { setCSSDimensions, getElementOffset } from './util';
 import { createCanvasElement, isHTMLCanvas } from '../../util/misc/dom';
 import { setCanvasDimensions } from './util';
+import { FabricError } from '../../util/internals/console';
 
 export type CanvasItem = {
   el: HTMLCanvasElement;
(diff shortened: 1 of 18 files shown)


ShaMan123 (commented):
ready

ShaMan123 (commented):
ok

src/canvas/DOMManagers/StaticCanvasDOMManager.ts
```diff
@@ -33,11 +34,9 @@ export class StaticCanvasDOMManager {
           (getFabricDocument().getElementById(arg0) as HTMLCanvasElement)) ||
         createCanvasElement();
     if (el.hasAttribute('data-fabric')) {
-      /* _DEV_MODE_START_ */
-      throw new Error(
-        'fabric.js: trying to initialize a canvas that has already been initialized'
+      throw new FabricError(
+        'Trying to initialize a canvas that has already been initialized. Did you forget to dispose the canvas?'
       );
-      /* _DEV_MODE_END_ */
```
asturur, line 40: i will still do this one day. I can re-add them later

asturur (dismissed):

asturur:
Found 68 of the extra 192 bytes, i m won't look any longer but from the diff i was expecting from the diffs this could easily be a pr that was removing some payload.

asturur (approved):

PR #9370, opened 2023-09-23
fix(Control): method binding for mouseUpHandler, mouseDownHandler, and actionHandler

ShaMan123:
## Motivation

Creating controls and subclassing them I found that I can't override actionHandler, mouseDownhandle, mouseUpHandler for 2 reasons:
1. The methods do not get called bound to the instance
2. They are not defined as method by TS but as properties so it troubles me 

## Description

Bind the handler when getting them.
I must say I am not sure I understand the need og `getActionHandler` etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - chore(): cleanup logs and error messages [#9369](https://github.com/fabricjs/fabric.js/pull/9369)
 - feature(Object) BREAKING: Remove lines parameter from object.containsPoint [#9375](https://github.com/fabricjs/fabric.js/pull/9375)
 - patch(Control): move hit detection to shouldActivate [#9374](https://github.com/fabricjs/fabric.js/pull/9374)
-- fix(Control): method binding [#9370](https://github.com/fabricjs/fabric.js/pull/9370)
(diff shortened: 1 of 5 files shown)


ShaMan123 (commented):
done apart for the TS thing that I am not sure I want to dig into - having an optional method that can be passed in option or defined on the instance

asturur:
The methods do not get called bound to instance because the methods get the instance passed in as a parameter.
If you bind them on request you get always a different function every time you call for getActionHandler for example.
Can't you bind them when you assign them?

asturur:
> I must say I am not sure I understand the need og getActionHandler etc.

I think the point was only to let you return a different function or let you run your own logic if for some reason you add 2 - 3 different functions you wanted to run depending on event status, object.

It was just a possibility, if you have an actionHandler that does multiple things is not useful.

ShaMan123:
> Can't you bind them when you assign them?

I don't assign them.
That is the thing, I use a class and expections are for a method to be bound.
I can handle it for sure but I didn't expect this to happen.
Changed the binding to call where possible.

asturur:
This still feels weird.
I think the best thing would be to remove that getActionHandler after being sure we know why is in that way.

asturur (dismissed):

asturur (approved):

ShaMan123:
I thought so as well but wasn't sure about breaking the api

asturur:
yes let's keep this fix, and let's see later how important is this extra layer on top of the action themselves

PR #9371, opened 2023-09-23
BREAKING refactor(Intersection): support ray intersection

ShaMan123:


ShaMan123 (commented):
done, please merge this. It will make my work easier

src/controls/Control.ts
```diff
@@ -169,11 +170,17 @@ export class Control {
...
+    fabricObject: InteractiveFabricObject,
+    pointer: Point,
+    { tl, tr, br, bl }: TCornerPoint
+  ) {
     // TODO: locking logic can be handled here instead of in the control handler logic
     return (
       fabricObject.canvas?.getActiveObject() === fabricObject &&
-      fabricObject.isControlVisible(controlKey)
+      fabricObject.isControlVisible(controlKey) &&
+      Intersection.isPointInPolygon(pointer, [tl, tr, br, bl])
```
ShaMan123, line 183: I am very pleased from this.
Now if I want to create a control for scaling a side I can simply replace the last call with `intersectLinePolygon`

ShaMan123 (commented):
This is READY
I will refactor `containsPoint` to use `isPointInpolygon` and rm `_getImageLine`

src/Intersection.ts
```diff
@@ -81,6 +81,36 @@ export class Intersection {
...
+   * Use the ray casting algorithm to determine if {@link point} is in the polygon defined by {@link points}
+   * @see https://en.wikipedia.org/wiki/Point_in_polygon
+   * @param point
+   * @param points polygon points
+   * @returns
+   */
+  static isPointInPolygon(point: Point, points: Point[]) {
+    const other = new Point(point).setX(
+      Math.min(point.x - 1, ...points.map((p) => p.x))
+    );
```
ShaMan123, line 94: this is a simple workaround to the need to cast a ray - I could have refactored intersectsLineLine for that but decided not to, keeping things simple

asturur:
@ShaMan123 i want to merge this PR, i see i don't manage to explain how i would like to see it finished.
I will swap all the occurrences of the old raycasting with the new one, and add a readme with the code for the benchmark for the new and old code, delete the old code.

Is that ok?

ShaMan123:
absolutely

asturur:
So i created a benchmark folder with a script runnable in node.

For my machine i had to bump to 1 milion runs to get meaningful results.

```
andrea@mbp-m1pro > /fabric.js/src > controls-point-in-poly * $ node ./benchmarks/raycasting.mjs
{
  benchmark1: 107,
  benchmark2: 327,
  bench1_run: 0.000107,
  bench2_run: 0.000327
}
```

this implementation is 3 times slower than the original, but since we are talking about less than one micro second per run we don't care for controls at all.

There is a reason why for the new implementation i used points and for the old one i used XY interface, i wanted to include in the measurement the fact that one runs on Point only, while the other could run on XY too. If i make the new one running on XY i shave off 10% from the result. that means that creating a point account for 10% of the time spent more or less.

It also must be noted, while it is common knowledge, that micro benchmarking isn't useful, but i think still is important that when we decide to ditch old code for new we try at least to understand what are the implications.

src/shapes/Object/ObjectGeometry.ts
```diff
@@ -343,10 +342,8 @@ export class ObjectGeometry<EventSpec extends ObjectEvents = ObjectEvents>
    * @return {Boolean} true if point is inside the object
    */
   containsPoint(point: Point, absolute = false, calculate = false): boolean {
-    return cornerPointContainsPoint(
-      point,
-      this._getCoords(absolute, calculate)
-    );
+    const { tl, tr, br, bl } = this._getCoords(absolute, calculate);
```
asturur, line 345: @ShaMan123 please not that your test implementation was based on this.getCoords, while for compatibility reason this is still using _getCoords.

There is a lot to say about moving this to getCoords, and i don't want to do it now

src/Intersection.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    ],
+  ] as const)('%p is in polygon %p, case index %#', (point, result) => {
+    const polygon = [
+      new Point(),
+      new Point(1, 0),
+      new Point(1, 1),
+      new Point(0, 1),
+    ];
+    expect(Intersection.isPointInPolygon(point, polygon)).toBe(result);
+  });
```
asturur, line 66: those tests are hard to read and do not cover use cases this new implementation should cover, i will propose some different tests later.

src/Intersection.ts
asturur: Is the scope of this loop to define a point on the left that is for sure outside the polygon?

asturur: If the answer is yes, i would rather use -Number.MAX_VALUE and ditch the loop.

ShaMan123: yes, instead of casting an infinite ray.
but you are right. we must reduce it by another step

ShaMan123: I have tried to change the intersection methods to support ray casting. Not a lot of change. and I think it is better than this

asturur: Half of the speed hit goes away, just to say.
map on 4 points means creating 4 function stacks, creating a new array, then spread it, and run Math.min on it. on those things that do pure calculations is all wasted time.

asturur, line 94:
```suggestion
    const other = new Point(-Number.MAX_VALUE, point.y);
```

ShaMan123: this is great, it can be also positive, no need for minus

src/benchmarks/README.md
ShaMan123, line 1: why not with jest and a separare cmd?

src/shapes/Object/ObjectGeometry.ts
ShaMan123: I fixed in in #9372

ShaMan123 (commented):
I did not think of MAX_VALUE - good point

src/Intersection.ts
```diff
@@ -81,6 +81,36 @@ export class Intersection {
...
+        points[index],
+        points[(index + 1) % points.length],
+        // ray
+        point,
+        other
+      );
+      if (inter.includes(point)) {
+        // point is on the polygon side
+        return true;
+      }
```
asturur, line 108: is this an optimization or is needed to cover extra use cases?
This would help if the polygon had a tons of points and we are checking for containing of one of the early points in the list. removing it on a 4 point polygon removes another 20% of speed loss

ShaMan123: If the ray is casted in this case it will return an even number of points and fail the ray casting test

ShaMan123: but you can concat all the results and check at the end if that is better

ShaMan123: I bet if we make the array of points a Set it will help also

ShaMan123: I think we should move to Set wereever possible

asturur: if having the point exactly over the segment does not count as intersection we can leave this code as it is. I wanted to be sure it wasn't a speed optimization of some sort. It should be written in a comment that being on the segment doesn't count as intersecting, so we don't forget.

src/Intersection.ts
asturur: ok i ll commit this change then

asturur:
well everything broke with the change.
Not sure why it doesn't work.

src/benchmarks/README.md
asturur: Just because jest is for tests, it does its own things, i do not know which they are and how they change version by version, it doesn't minify the code for example, it just transpile it, i m not sure if it transpile for node rather than browser, is just simpler to run a function, we don't need to have a framework to do that.

asturur:
somehow the implementations are different. Ok interesting to understand where.

asturur:
the MAX_VALUE thing breaks something. that is interesting too, but some tests aren't working anyway

asturur:
the reason why MAX_VALUE breaks is because when we do segment segment intersection we do multiplications around MAX_VALUE that send stuff to NUMBER.Infinity level.
So we could take a smaller number.

MAX_VALUE is around the 10 ^ 308 interval.
MAX_SAFE_INTEGER is 9007199254740991 around 10 ^ 15

now if you work with very large objects i would argue MAX_SAFE_INTEGER is large enough and if it isn't probably the rounding error that happens after that start to be noticeable.

This fall under the opinions umbrella but i would argue developers and projects working with numbers that are inside the trillion is enough? and so MAX_SAFE_INTEGER is safe enough for our ray casting?

We could also take a stance and document fabricJS working inside the MAX_SAFE_INTEGER range for accuracy reasons and assume MAX_SAFE_INTEGER + 1 ( a float ) is our infinite ray casting.

ShaMan123:
Seems it is better to patch the line intersection
I will try that

asturur:
we can keep the point loop let's not derail.
Either a big enough but not problematic number or the loop.
Digging in changes for the other function will open its own issues.

ShaMan123 (commented):
Added support for `ray` intersection, cut down to x2 on my machine
Added some tests for it.
It is a breaking change because passing `true` instead of `line` or `false` instead of `segement` is not supported but can be easily with a small
```
if(type === true) type = 'line';
else if(type === false) type = 'segement';
``` 

For me this is good enough.
I also tried to use map instead of array but that was distrasterous, x7 slower

ShaMan123:
The remaining failing test are solved in #9333

asturur:
the issue is not the breaking change is stacking changes on top of others.
We can merge commit https://github.com/fabricjs/fabric.js/pull/9371/commits/2cf63e68c59140a2dacd88d8f9ca6383f232cde6 and keep the rest for the future.

asturur:
But do we know why the tests in unit tests are failing? i understand the new tests are better, but what did it change in the output now?

ShaMan123 (commented):
merge this please

asturur:
What is the scope of this PR now? isn't it just an improvement in speed functionally identical?

ShaMan123:
I am not sure
Didn't you say there was a problem with Number.MAX_VALUE?
This is the right and safe way of doing it

ShaMan123:
Now I need  this in production

asturur:
My point was ( and is ) that this PR started as something entirely different that was hit testing a control.
Then that was added with the necessity of hit testing not against a square but as a generic polygon and we did it.

What this PR is achieving now?
It is improving how we run the code for intersection making it faster, or is there more?
Have we upgraded the intersection class as something the user is supposed to use outside object collisions and internal calculations?

src/Intersection.ts
```diff
@@ -89,21 +99,25 @@ export class Intersection {
    * @returns
    */
   static isPointInPolygon(point: Point, points: Point[]) {
-    const other = new Point(point).setX(
-      Math.min(point.x - 1, ...points.map((p) => p.x))
-    );
```
ShaMan123, line 94: you said this is a major perf hit so it is removed in favor of solid math

asturur: Since you pointed out that you need this in production, i m asking, what is this doing more than fabric already does? my question is very simple If it was just a perf bump you wouldn't need it.

i have no issue merging it, i just need to try to get out some extra bytes, but my question is really, how a performance improvement become needed? it means that is more than performances no?

ShaMan123: Yes
Connecting lines
It is needed for it. I need to get an intersectiin for the side of the control I am connecting to. That is a ray intersection.
Needed especially for elbow (90deg) connectors

ShaMan123:
Who ever needs this can copy it over or use my fork

PR #9372, opened 2023-09-24
fix(Geometry): `containsPoint` - BREAKING `Canvas#_checkTarget`

ShaMan123:
## Motivation

`containsPoint` is not user freinly and that is a shame. 
It accepts points in different planes depending on use cases and is to me at least unintuitive.

This is actually a part of #8767 - Since you have freezed it I have decided to extract this important part immediately.

## Description

First merge #9333 

## Changes

**BREAKING**
- `containsPoint` signature `containsPoint(point, lines, absolute, calculate)` => `containsPoint(point, absolute, calculate)`
- `_checkTarget` signature `_checkTarget(pointer, target, globalPointer)` => `_checkTarget(target, globalPointer = pointerFromViewport)`

## Gist

I believe it is now possible to remove lineCoords


zhe-he:
Excuse me for interrupting. Is this case overwritten?
```javascript
const rect = new Rect({ left: 0, top: 0, width: 100, height: 100 });
const g = new Group([rect], { left: 0, top: 0 });
const g2 = new Group([g], { left: 10, top 10 });
canvas.add(g2);
const p = new Point(5, 5);

const flag = g.containsPoint(p);
console.log(flag);
```

Or do I need to manually remove the child elements from the group and then evaluate, and once evaluated, restore them, like this?
```javascript
const rect = new Rect({ left: 0, top: 0, width: 100, height: 100 });
const g = new Group([rect], { left: 0, top: 0 });
const g2 = new Group([g], { left: 10, top 10 });
canvas.add(g2);
const p = new Point(5, 5);

+ const gg = g.group;
+ gg.remove(g);

const flag = g.containsPoint(p);

+ gg.add(g);
console.log(flag); // Is this the correct way to achieve the desired result?
```

src/canvas/SelectableCanvas.ts
```diff
@@ -807,28 +803,22 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
     // Cache all targets where their bounding box contains point.
-    let target,
-      i = objects.length;
+    let i = objects.length;
     // Do not check for currently grouped objects, since we check the parent group itself.
     // until we call this function specifically to search inside the activeGroup
     while (i--) {
-      const objToCheck = objects[i];
-      const pointerToUse = objToCheck.group
-        ? this._normalizePointer(objToCheck.group, pointer)
```
asturur, line 817: _normalizePointer can go away now. It was used only for this and was explicitly private

asturur: i removed it

src/canvas/SelectableCanvas.ts
```diff
@@ -762,30 +750,26 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+  _checkTarget(obj: FabricObject, pointer: Point): boolean {
     if (
       obj &&
       obj.visible &&
       obj.evented &&
-      // http://www.geog.ubc.ca/courses/klink/gis.notes/ncgia/u32.html
-      // http://idav.ucdavis.edu/~okreylos/TAship/Spring2000/PointInPolygon.html
-      obj.containsPoint(pointer)
+      obj.containsPoint(
+        sendPointToPlane(pointer, undefined, this.viewportTransform),
```
asturur, line 764: can we use this.restorePointerVpt(pointer) here?

We are taking the pointer in html coordinates and we are transitioning it to scene coordinates correct?

asturur:
#9333 has 2000 lines to read and requires 2 screens to do it comfortably, on the small notebook screen is a pain to compare old and new tests. 
I think the dependency of  #9333 has been removed yesterday when i fixed the failing tests on the qunit side.

src/canvas/SelectableCanvas.ts
asturur: this change and looking at possibly adding tests is the only thing i need to merge this

asturur:
i ll add specific tests for the new functionality in a following PR.
We need some small test that can cover nested stuff properly for containsPoint and checkTarget

asturur (approved):

src/canvas/SelectableCanvas.ts
ShaMan123: I deliberately removed this usage
You removed restoreVpt so why leave this?
It is removed in the getPointer renaming pr, is it not?

PR #9373, opened 2023-09-24
BREAKING refactor(Geometry): rm line coords

ShaMan123:
## Motivation

#8767 #9372 

It it something we are talking about for a long time and now it is possible.

## Description

First merge #9372 (removes the need for `lineCoords`)
Removes `lineCoords` from geometry and the `absolute` concept/arg as well as the `calculate` arg.
Meaning that in this PR all calculation are done in the scene plane, point args are expected to be in that plane.
This is incorrect because of stroke uniform, padding etc. but is a big step forward. I made this PR incremental and minimal as possible.
This should be considered part of the group rewrite because it make Geometry 100% safe to use with groups as opposed to before there were a number of valunrabilities (`_getCoords` returning own coords regardless of group, `containsPoint` doing something still unclear to me etc.)
Also if I am not mistaken, `lineCoords` were wrong when under group because they applied the viewportTransform before the group transform, it should have been vice versa.

## Changes

- `oCoords` => `controlCoords`
- `aCoords` => `ownCoords`
- `calcOCoords` => `protected calcControlCoords`
- `calcACoords` => `protected calcCornerCoords`
- rm `_getCoords`, `_containsCenterOfCanvas`, `calcLineCoords`
- rm `absolute, calculate` for all geometry methods
- rm `scaleToWidth`, `scaleToHeight` because it is not supported under group and expections are unclear regarding usage + it is a fairly simple method to write


src/shapes/Object/Object.ts
```diff
@@ -1375,9 +1375,10 @@ export class FabricObject<
       sendObjectToPlane(this, this.getViewportTransform());
     }
 
+    // skip canvas zoom and calculate with setCoords now.
+    this.setCoords();
```
ShaMan123, line 1379: i think there is aredundant call to setCoords in this method

ShaMan123 (commented):
- `oCoords` => `controlCoords`
- `aCoords` => `ownCoords`
progress

ShaMan123 (commented):
- `calcOCoords` => `protected calcControlCoords`
- `calcACoords` => `protected calcCornerCoords`
- rm `_getCoords`, `_containsCenterOfCanvas`, `calcLineCoords`
- rm `absolute, calculate` for all geometry methods
- rm `scaleToWidth`, `scaleToHeight` because it is not supported under group and expections are unclear regarding usage + it is a fairly simple method to write

test/unit/object_geometry.js
```diff
@@ -218,7 +214,7 @@
     assert.equal(object.containsPoint(point5), false);
   });
 
-  QUnit.test('containsPoint with padding', function(assert) {
+  QUnit.test.skip('containsPoint with padding', function(assert) {
```
ShaMan123, line 217: we should discuss this.
It is not a problem respecting padding in `containsPoint`
I think it should be an option though.

test/unit/object_geometry.js
ShaMan123, line 1: moved all the getCoords tests to `src/shapes/Object/Object.spec.ts`

ShaMan123 (commented):
good to go

ShaMan123 (commented):
u

ShaMan123:
Nope
Go ahead

asturur:
i would like to remove the absolute parameter in the codebase without any rename first.
Then deciding how the things should be named looking at how they can be used differently now.
I want to do that for the pleasure of coding a bit because i can do just reviews and tests and i m not having any fun

asturur:
i made a different branch where i tackle the entirity of the absolute arguments in the functions ( at lesat all the one i found ), but i didn't rename anything and i still didn't test it properly.

But that is how i would like to tackle the removal of a layer of coordinates that we shouldn't have.
As soon as i manage to get all uts passing i ll open a pr
branch is: `remove-absolute-globally` if you are curious but is not ready for review/comments

PR #9376, opened 2023-09-25
patch(Control): pass object to `calcCornerCoords`

ShaMan123:
## Motivation

Same as #9374 
Working on custom controls.
I want to calc the control size according to the object size

## Description

Pass the object ref to `calcCornerCoords`
It is the only method that doesn't receive the ref


ShaMan123 (commented):
very simple
but crucial for me - please merge ASAP and release a beta

asturur:
this is ok.
I had some issue when we were passing also this.angle, but at this point is fine as it is.

asturur (approved):

PR #9380, opened 2023-09-25
patch(Coords): calc oCoords only with canvas ref

ShaMan123:
## Motivation

~Still working on controls and finding too many edgecases
I have a control that needs to access the canvas, so trying to do `fabricObject.canvas` needa to be safeguarded.
And it makes absolutely no sense.
Because controls exist only on the canvas and need the canvas for vpt calculations etc. so without the canvas the controls are not used, calculated for no reason and produce wrong values~

Since the introduction of layout in groups and nested targets, calls to setCoords are happening before adding an object to the canvas.
With the recent addition of https://github.com/fabricjs/fabric.js/pull/9376 you can use the object itself when calculating objects coordinates and this brings you to eventually evaluate the canvas of that object that is undefined in that moment.
For objects that do not do that we are still setting oCoords ignoring the canvas VPT and so potentially wrong.

## Description

calc oCoords only when the canvas is available to the object
Since fabric calls setCoords when an object is added to canvas this PR makes no difference and is not breaking, but when working with custom logic the result will predictable and will save the developer from annoying errors and bugs.


ShaMan123 (commented):
can't be simpler than this
done

asturur:
can't accept a PR anymore that cites annoying error and bugs without specifing what is going on.
Is the issue that viewportTransform is the identity matrix when canvas is not defined?
Why canvas is note defined in your use case?

ShaMan123:
Yes
Not canvas means no vpt
Also group layout calls setCoords and that can happen before being added to canvas

hexiaokang (approved):

asturur (approved):

PR #9381, opened 2023-09-25
feat(Intersection) add implementation of pointInPolygon

asturur:
## Motivation

I have a control that is not defined by a point such as scaling an object for the sides.
Currently I need to override `_findTargetCorner` which is OK but not great
However it makes sense moving the hit test to the control IMO and I can override `shouldActive` with ease

## Description

Perform the control hit test in `Control#shouldActivate` and use `isPointInPolygon` instead of `containsPoint` to do that (because it doesn't make sense passing those lines around, see #9372 )
Then overriding is easy and straight forward because the dev sees they can use the `Intersection` class

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -36,6 +36,7 @@ module.exports = {
     ],
     'no-restricted-syntax': [
       'error',
+      // explore how to define the selector: https://astexplorer.net/
       {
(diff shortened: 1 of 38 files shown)


asturur:
Picture of the test for reference
image

PR #9383, opened 2023-09-26
Breaking: Remove node 14

asturur:
## Motivation

It doesn't get any security update anymore since long
We can keep node16 unless some security issue comes out, but it faded out this month too.


asturur:
i realize now our node build isn't targeting any specific node version. i ll open an issue for that.

PR #7726, opened 2022-02-21
chore(v6): prepare fabric for Group

ShaMan123:

PR #9387, opened 2023-09-27
fix(Canvas): invalidate `_objectsToRender` on stack change 

ShaMan123:
## Motivation

@jiayihu found a bug

## Description

We forgot to invalidate `_objectsToRender` on stack change

## Changes

the fix itself and a test


ShaMan123 (commented):
done

jiayihu:
IMO this is a bad design and caching should never be implicit, rather left opt-in for the developer or even as override if needed. As a developer, I would need to dig into the source code to find out why my objects are not correctly rendered even **after** maybe spending hours investigating that the `_objects` array has been correctly modified.

ShaMan123:
I am not sure I agree.
This is a pure bug.
As long as caching works it is great, once it doesn't it is awful to debug. You said that.
Your experience was that exactly.

ShaMan123:
But maybe we could add a flag to renderAll

asturur:
We set _objectsToRender as undefined most of the time when we need to invalidate,
but when the canvas start we assign it an empty array

```ts
  /**
   * hold the list of objects to render
   * @type FabricObject[]
   * @private
   */
  _objectsToRender?: FabricObject[] = [];
```

Why?

ShaMan123:
Nice catch

asturur:
Can we set it as `declare` there and avoid the empty array? the first render with or without objects will be created.

ShaMan123:
Sounds the right way to it.
Also adding an assertion to the test I have added

asturur:
> IMO this is a bad design and caching should never be implicit, rather left opt-in for the developer or even as override if needed. As a developer, I would need to dig into the source code to find out why my objects are not correctly rendered even **after** maybe spending hours investigating that the `_objects` array has been correctly modified.

Regarding this i don't have an opinion about this being bad or not bad.
The change of caching the result of that calculation was added here:
https://github.com/fabricjs/fabric.js/pull/7726
as part of 'neutral group changes'
It hasn't given much thought, but as with each caching it brings complexity and possible bugs.

Regarding a developer digging hours and not understanding why what, that is the exact issue that i mention and why i m not releasing 6.0 without a full set of docs and explanations.
The code auto documenting itself is not true and people should have document available to understand how a framework works under the hood.

src/canvas/Canvas.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+      canvas._objectsToRender = [];
+      canvas.bringObjectToFront(object);
+      expect(canvas._objectsToRender).toBeUndefined();
+
+      canvas._objectsToRender = [];
+      canvas.sendObjectToBack(object);
+      expect(canvas._objectsToRender).toBeUndefined();
+    });
+  });
+});
```
asturur, line 55: can we call this 'SelectableCanvas.spec.ts' so it matches the part of code we are writing the test for? with classes and tests covering multiple levels of classes some files will grow enormous and we will need at some point to divide them differently, but for now we can stick to a file matching the name of the file we are testing.

ShaMan123:
I didn't add `declare` because there is no need for it since it is undefined by default

ShaMan123 (commented):
applied review comments

asturur (approved):

asturur:
> I didn't add `declare` because there is no need for it since it is undefined by default

is the way it gets assigned that is unfriendly with setOptions, we can add declare another time.

PR #9388, opened 2023-09-27
ci(e2e): fix babel compiling error

ShaMan123:
## Motivation

Babel has a silent bug (on windows at least) - it doesn't comiple the files if using babel cli from a spawned call or something like that. It does not throw.
What it means is that when I run e2e tests they will fail ungracefully and I will be the only one to know why.
When I run e2e tests I must copy the cmd from `playwright.setup.ts` and run it with npx for babel to pick it up correctly.

## Description

Move away from using babel cli via a spawn call to **using the API directly**.


ShaMan123 (commented):
DONE
please merge asap

playwright.setup.ts
```diff
@@ -1,16 +1,84 @@
...
+  console.log(
+    `Successfully compiled ${tasks.length} files from ${path.relative(
+      process.cwd(),
+      src
+    )} to ${path.relative(process.cwd(), dist)}`
+  );
+
+  // watch
+  if (process.argv.includes('--ui')) {
+    const watcher = watch(
```
ShaMan123, line 74: super flaky on windows but good enough to move on

asturur:
if this removes local development headaches i m all for it

Also i think windows is a pain for node development, does it work properly if you run it on the linux subsystem? or is flaky there?

asturur (approved):

ShaMan123:
I don't know
Seems that windows is a pain full stop

PR #9395, opened 2023-10-01
BREAKING: remove absolute true/false from the api.

asturur:
## Motivation

Fabric is currently maintaining 2 sets of coordinates.
One that is a scene/design coordinate space and another that is the screen/viewport space.
The scene is called aCoords, while the screen/viewport are both oCoords and lineCoords.
Those 2 sets aren't simply connected by the viewport trasnformation but also they have the addition of padding that always threw a wrench in calculations and predictability of the api.

Each api of the geometry topic ( containsPoint, containsObject, getBoundingBox ) had an option to be called with a boolean called `absolute`. In case the boolean was true, calculation were made around aCoords, while with false around lineCoords.

The issue was that lineCoords included padding, and that meant that depending on the padding level of controls those intersection would trigger also over the padding area.
This created a bleeding between 2 different piece of logic, pure object geometry and selection/targeting functionalities.

Padding adds some pixels around the natural bounding box of the object either for aesthetic reasons or in order to facilitate the object selection when an object is too small.
Adding padding to an object doesn't make it larger or covering more area or intersecting with more objects.

**So now padding is out of all the geometries.**

to make an example, scaleToWidth would scale an object to X pixels, on screen pixels, including padding.
If you had padding set to 10 to an object, you would get the bounding box of the object including padding, then you would divide that by the scaledWidth of the objet to get a scale factor.
Imagine the scale factor was 0.5, you would scale down your object by half, but padding never scales, so again your object would always be either too small ( because you don't expect and you don't see padding ) or too large ( because you expect padding but padding doesn't get scaled by object scaling ).

**object.lineCoords have no reason to exists and are gone**
All the api that required absolute=true to be specified are loosing absolute option and that is true by default.
We are slowly stopping referring to those coordinates as absolute since it was also confusing.

List of affect apis:

- _getCoords
- getCoords
- intersectsWithRect
- intersectsWithObject
- isContainedWithinObject
- isContainedWithinRect
- containsPoint
- getBoundingRect
- scaleToWidth
- scaleToHeight

Padding calculation is now a concern of the canvas and for now is not cached.
If you have an object with padding, every time you are selecting it or running a target check over it, you will run this calculation on top of the normal ones:

```ts
    const padding = obj.padding / viewportZoom;
    if (padding) {
      const [tl, tr, br, bl] = coords;
      // what is the angle of the object?
      // we could use getTotalAngle, but is way easier to look at it
      // from how coords are oriented, since if something went wrong
      // at least we are consistent.
      const angleRadians = Math.atan2(tr.y - tl.y, tr.x - tl.y),
        cosP = cos(angleRadians) * padding,
        sinP = sin(angleRadians) * padding,
        cosPSinP = cosP + sinP,
        cosPMinusSinP = cosP - sinP;

      coords = [
        new Point(tl.x - cosPMinusSinP, tl.y - cosPSinP),
        new Point(tr.x + cosPSinP, tr.y - cosPMinusSinP),
        new Point(br.x + cosPMinusSinP, br.y + cosPSinP),
        new Point(bl.x - cosPSinP, bl.y + cosPMinusSinP),
      ];
```

The code has been written with the idea of minimizing the calculation and to be consistent with the status of aCoords.
If aCoords are out of sync, padding will be out of sync too, that is ok.

## Testing

This pr adds a new method that could be tested a bit more rigorously and requires manual verification since i conducted manual tests with objects and group but not too deeply


asturur:
replace #9373

asturur:
i need to add back some debug code so we can see against what we are targeting objects.

asturur:
Found the small error, now i have to write a test that is clear in the visual and the intent

asturur:
Ok this is ready for review.

ShaMan123:
> adding adds some pixels around the natural bounding box of the object either for aesthetic reasons or in order to facilitate the object selection when an object is too small.
Adding padding to an object doesn't make it larger or covering more area or intersecting with more objects.

Rearding `facilitate the object selection when an object is too small` I don't think this is supported any longer with this PR

All the description is correct only for canvas level objects.
For objects under groups lineCoords were always wrong because the padding that as added t othem would be then transformed by the group plane

src/canvas/SelectableCanvas.ts
```diff
@@ -754,7 +755,49 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+      // otherwise we have to maintain 2 sets of coordinates for everything.
+      // we can reiterate on storing those on something similar to lineCoords
+      // if this is slow, for now the semplification is large and doesn't impact
+      // rendering.
+      // the idea behind this is that outside target check we don't need ot know
+      // where those coords are
+    }
+    return Intersection.isPointInPolygon(point, coords);
+  }
+
```
ShaMan123, line 797: Do you mind if I simplify the logic here using a padding vector?
Does it have to be private?
Can we instead change `containsPoint` to accept a 3rd param `padding` defaulting to 0?
I think that is much better both for fabric and for the dev

src/canvas/SelectableCanvas.spec.ts
ShaMan123, line 1: I am not sure this is extensive enough
What about using the same tehnique as the e2e test? Iterating over all points and calling the method and storing the result somehow?
Then it is a pixel perfect test and the test can e reduced in size using `each` I guess

ShaMan123 (changes requested):
Added comments, some asking for changes

src/Collection.ts
```diff
@@ -317,6 +317,7 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
       { left, top, width, height }: TBBox,
       { includeIntersecting = true }: { includeIntersecting?: boolean } = {}
     ) {
+      // TO DO VERIFY WHAT ARE WE PASSING IN
```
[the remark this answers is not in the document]
asturur: i forgot, i don't know. I ll try to remember to be sure this wasn't something extremely important

src/mixins/eraser_brush.mixin.ts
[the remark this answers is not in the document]
asturur: ok i can do that.

src/shapes/Object/ObjectGeometry.ts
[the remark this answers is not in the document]
asturur: i don't want to do that now.
We can do that when those new concepts are settled.
Is easier to say, from now own care just about aCoords only, and forget lineCoords, oCoords, rather than pulling a new name.

If geometry handling will change strongly later, we can give a new name
Is like renaming top and left, is doable but too much mess right now.
I do need some continuity.

src/canvas/SelectableCanvas.ts
asturur: my issue here is that Intersection is not a selection thing.
This is a method entirely done to check the selection and padding is selection only.
Why would you use containsPoint with padding? use checkTarget right away no?
if is padding is a selection thing.

Why would you use containsPoint instead of checkTarget?


Regarding private, i did that because i thought, let's see how it shapes out, become public is easier, or maybe then we need to change it radically or change its interface, and being private is a lesser deal.

Regarding the logic, i would like to see it first, so that we can compare is actually simpler.
This was just porting over what was before somewhere else, i m not attached to it.

Just dump how you would write this whole method in a comment.

test/unit/object_geometry.js
```diff
@@ -529,19 +502,19 @@
...
+    // assert.equal(boundingRect.width, 10, 'gives the bounding rect width with zoom 1');
+    // assert.equal(boundingRect.height, 10, 'gives the bounding rect height with zoom 1');
+    // cObj.canvas = {
+    //   viewportTransform: [2, 0, 0, 2, 0, 0]
+    // };
+    // cObj.setCoords();
+    // boundingRect = cObj.getBoundingRect();
+    // assert.equal(boundingRect.left, 10, 'gives the bounding rect left with zoom 2');
+    // assert.equal(boundingRect.width, 20, 'gives the bounding rect width with zoom 2');
+    // assert.equal(boundingRect.height, 20, 'gives the bounding rect height with zoom 2');
```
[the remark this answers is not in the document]
asturur: i have to follow up on this. I don't remember if i wanted just to move it in jest or there was some issue.
If i moved it i will just delete it

test/unit/object_geometry.js
```diff
@@ -480,12 +453,12 @@
...
-    cObj.scaleToWidth(100, true);
+    cObj.scaleToWidth(100);
     assert.equal(cObj.getScaledWidth(), 100, 'is not influenced by zoom - width');
     assert.equal(cObj.get('scaleX'), 100 / 560);
-    cObj.scaleToWidth(100);
-    assert.equal(cObj.getScaledWidth(), 50, 'is influenced by zoom - width');
-    assert.equal(cObj.get('scaleX'), 100 / 560 / 2);
+    // cObj.scaleToWidth(100);
+    // assert.equal(cObj.getScaledWidth(), 50, 'is influenced by zoom - width');
+    // assert.equal(cObj.get('scaleX'), 100 / 560 / 2);
```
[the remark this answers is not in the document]
asturur: this is because removing the duality of scene/viewport scaleToWidth, this test doesn't make sense anymore. This needs to be deleted

src/Collection.ts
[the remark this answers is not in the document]
asturur: i wanted to be sure the bounding box was in scene plane, and it is.

test/unit/object_geometry.js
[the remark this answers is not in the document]
asturur: those are just tests that are not valid anymore because of the removal of the extra param

src/canvas/SelectableCanvas.spec.ts
asturur: this test covers most of the point at the edge of selection, and cover cases with groups, padding, skew, strokeUniform.
I think is fine, and if it doesn't cover something that will bug out, we will extend the test.

asturur:
This is ready to go.

src/canvas/SelectableCanvas.ts
ShaMan123:
```ts
// expose as const, can be reused by other parts
    const CORNER_ORIGIN = {
      tl: new Point(-0.5, -0.5),
      tr: new Point(0.5, -0.5),
      br: new Point(0.5, 0.5),
      bl: new Point(-0.5, 0.5),
    };

// expose in ObjectGeometry
  containsViewportPoint(point: Point, padding = 0) {
    const vpt = this.getViewportTransform();
    const paddingVector = new Point().scalarMultiply(padding * 2);
    const [tl, tr, br, bl] = this.getCoords();
    const viewportCoords = (
      Object.entries({ tl, tr, br, bl }) as [
        keyof typeof CORNER_ORIGIN,
        Point
      ][]
    ).map(([key, point]) =>
      sendPointToPlane(point, vpt).add(
        paddingVector.multiply(CORNER_ORIGIN[key])
      )
    );

    return Intersection.isPointInPolygon(point, viewportCoords);
  }

```

src/Collection.ts
```diff
@@ -311,13 +311,16 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
      * Given a bounding box, return all the objects of the collection that are contained in the bounding box.
      * If `includeIntersecting` is true, return also the objects that intersect the bounding box as well.
      * This is meant to work with selection. Is not a generic method.
+     *     /**
+     * Moves an object to specified level in stack of drawn objects
+     * @param {TBBox} bbox a bounding box in scene coordinates
+     * @param {{ includeIntersecting?: boolean }} options an object with includeIntersecting
```
ShaMan123, line 317: this is not very useful info but the above it so that should be moved here IMO

src/Collection.ts
```diff
@@ -311,13 +311,16 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
      * Given a bounding box, return all the objects of the collection that are contained in the bounding box.
      * If `includeIntersecting` is true, return also the objects that intersect the bounding box as well.
      * This is meant to work with selection. Is not a generic method.
```
ShaMan123, line 313:
```suggestion
```

Why can't it be used in general? It is a good piece of logic

ShaMan123 (commented):
take a look at my suggestion for `containsViewportPoint`

src/Collection.ts
ShaMan123: Include objects that are contained in or intersect the bounding box

src/Collection.ts
asturur: it was written for intersection on the canvas with the drag rectangle.
We are not promoting it now, promoting is possible without breaking later.

src/canvas/SelectableCanvas.ts
asturur: There is a lot to say about this.

Is better explained in person but i can summarize that Object geometry is done with padding and viewport. We just remove the concept that padding is part of the object geometry, is just a selection concern, and saying 'containsViewportPoint' and adding back padding is rolling back the change.

Also in this example i don't understand who is rotating the padding vector. I see we are using a direction vector to correctly move the point in that direction, but where is the rotation?

src/Collection.ts
asturur: I think that i wanted to copy a parameter section from another function and i pulled in a piece of comment without noticing. removing it.

src/canvas/SelectableCanvas.ts
ShaMan123: > but where is the rotation?

I missed that

> We just remove the concept that padding is part of the object geometry, is just a selection concern, and saying 'containsViewportPoint' and adding back padding is rolling back the change.

I have much to say about that. I have said that geometry must occur in the viewport but you did not take that into account so I don't see as rolling back

asturur: I don't remember you saying that, or we wouldn't have remove the absolute from everywhere.
Anyway, next meeting then.
I l freeze everything meanwhile

asturur:
follow up to change the padding calculation to use more matrix and vector names

asturur:
@ShaMan123 approved in meeting

PR #9396, opened 2023-10-02
fix(Canvas): in/out event names were swapped

jiayihu:
Event names were swapped during the migration.

### DIFF
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1325,7 +1325,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
     const targetChanged = oldTarget !== target;
 
     if (oldTarget && targetChanged) {
-      const outOpt = {
+      const outOpt: CanvasEvents[typeof canvasOut] = {
         ...data,
         e,
         target: oldTarget,
@@ -1338,7 +1338,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       oldTarget.fire(targetOut, outOpt);
     }
     if (target && targetChanged) {
-      const inOpt: TPointerEventInfo = {
+      const inOpt: CanvasEvents[typeof canvasIn] = {
         ...data,
         e,
         target,
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1322,7 +1322,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         pointer: this.getPointer(e, true),
         absolutePointer: this.getPointer(e),
       };
-      fireCanvas && this.fire(canvasIn, outOpt);
+      fireCanvas && this.fire(canvasOut, outOpt);
       oldTarget.fire(targetOut, outOpt);
     }
     if (target && targetChanged) {
@@ -1335,7 +1335,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         pointer: this.getPointer(e, true),
         absolutePointer: this.getPointer(e),
       };
-      fireCanvas && this.fire(canvasOut, inOpt);
+      fireCanvas && this.fire(canvasIn, inOpt);
       target.fire(targetIn, inOpt);
     }
   }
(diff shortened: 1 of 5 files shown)


src/canvas/Canvas.ts
```diff
@@ -1334,7 +1334,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         pointer: this.getPointer(e, true),
         absolutePointer: this.getPointer(e),
       };
-      fireCanvas && this.fire(canvasIn, outOpt);
+      fireCanvas && this.fire(canvasOut, outOpt);
```
jiayihu, line 1337: The fact that TS didn't catch this means probably that somewhere the type is not precise, although I was not able to spot it with a quick check.
Also @ShaMan123 says I ruined TS for him, so I won't stress it 😄

jiayihu: Okay I found and fixed the issue on a second look, now it will raise a type error:

Screenshot 2023-10-03 at 00 25 25

ShaMan123: It is not you who ruined TS for me
TS managed by itself
I did look at this part and was confused about it

ShaMan123: b6f076c is to blame
https://github.com/fabricjs/fabric.js/blob/d55fabcd671b183aa40091c9ed6d4b7fa24c87fa/src/mixins/canvas_events.mixin.ts#L1220-L1227

ShaMan123: surprising tests did not pick this up, will add

ShaMan123 (dismissed):

src/canvas/Canvas.ts
ShaMan123: they did - the snapshot was wrong

ShaMan123 (approved):
fixed types and added tests

ShaMan123 (approved):

jiayihu:
Why was the type fix reverted? If you use the optional operator, it won't complain if the `nextTarget/previousTarget` is not passed at all, even if undefined. That's why it was not giving an error.

src/canvas/__tests__/eventData.test.ts
```diff
@@ -122,3 +122,20 @@ it('A selected subtarget should not fire an event twice', () => {
...
+
+  jest.spyOn(target, 'toJSON').mockReturnValue('target');
+
+  const targetSpy = jest.spyOn(target, 'fire');
+  const canvasSpy = jest.spyOn(canvas, 'fire');
+  const enter = new MouseEvent('mousemove', { clientX: 5, clientY: 5 });
+  const exit = new MouseEvent('mousemove', { clientX: 20, clientY: 20 });
+  canvas._onMouseMove(enter);
+  canvas._onMouseMove(exit);
+  expect(targetSpy.mock.calls).toMatchSnapshot();
```
jiayihu, line 139: @ShaMan123  This could have been `targetSpy.mock.calls.map(([event]) => event)`, then even use `toMatchInlineSnapshot` for a much smaller and nicer snapshot, while being convenient.

ShaMan123:
> Why was the type fix reverted? If you use the optional operator, it won't complain if the `nextTarget/previousTarget` is not passed at all, even if undefined. That's why it was not giving an error. Whereas `nextTarget: fabric.Object | undefined` means that it must be explicitly provided, even if undefined, which is what we do.

I didn't look into it but build was failing.
My bad

asturur:
@ShaMan123 @jiayihu  you are welcome to merge laser focused fixes like this but:
- The source of the commit that broke the code shouldn't be hidden in comments, needs to be in PR description
- as pointed out in comments snapshot tests are uneffective so please let's stop adding snapshot tests for those kinds of thing.

We can simply say:
theMock has been called with 'mousein' or whatever, a blanket cover all of the event data isn't a good idea imho

src/canvas/__tests__/eventData.test.ts
asturur: yes.
stop `toMatchSnapthot`.

PR #9400, opened 2023-10-04
chore(): export poly control utils

ShaMan123:
## Motivation

I need to use the poly controls (at work) with my own subclass of Control so I need the underlying  methods to be exposed

## Changes

exported and renamed methods for consumption

### DIFF
--- a/src/controls/polyControl.ts
+++ b/src/controls/polyControl.ts
@@ -21,7 +21,7 @@ const getSize = (poly: Polyline) => {
  * This function locates the controls.
  * It'll be used both for drawing and for interaction.
  */
-const factoryPolyPositionHandler = (pointIndex: number) => {
+export const createPolyPositionHandler = (pointIndex: number) => {
   return function (dim: Point, finalMatrix: TMat2D, polyObject: Polyline) {
     const x = polyObject.points[pointIndex].x - polyObject.pathOffset.x,
       y = polyObject.points[pointIndex].y - polyObject.pathOffset.y;
@@ -41,7 +41,7 @@ const factoryPolyPositionHandler = (pointIndex: number) => {
  * and the current position in canvas coordinate `transform.target` is a reference to the
  * current object being transformed.
  */
-const polyActionHandler = (
+export const polyActionHandler = (
   eventData: TPointerEvent,
   transform: TTransformAnchor,
   x: number,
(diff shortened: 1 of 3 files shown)


src/controls/polyControl.ts
```diff
@@ -104,6 +104,9 @@ const anchorWrapper = (
   };
 };
 
+export const createPolyActionHandler = (pointIndex: number) =>
+  factoryPolyActionHandler(pointIndex, polyActionHandler);
+
 export function createPolyControls(
```
ShaMan123, line 110: might be that this is redundant now but I do not mind it exported as is

src/controls/polyControl.ts
```diff
@@ -124,8 +127,8 @@ export function createPolyControls(
   ) {
     controls[`p${idx}`] = new Control({
       actionName: 'modifyPoly',
-      positionHandler: factoryPolyPositionHandler(idx),
-      actionHandler: anchorWrapper(idx, polyActionHandler),
+      positionHandler: createPolyPositionHandler(idx),
+      actionHandler: createPolyActionHandler(idx),
```
ShaMan123, line 131: these are the only things I need exported but why not export the rest?

ShaMan123 (commented):
Please merge ASAP

src/controls/polyControl.ts
```diff
@@ -69,7 +69,7 @@ const polyActionHandler = (
 /**
  * Keep the polygon in the same position when we change its `width`/`height`/`top`/`left`.
  */
-const anchorWrapper = (
+export const factoryPolyActionHandler = (
```
asturur, line 72: since this is an export now let's give it a name that is more understandable:
polygonAnchorPointWrapper? wrapWithPolygonAnchorPoint?

this is creating a new action handler rather than being a pure wrapper, but the name factory is a bit misleading because the action handler itself is passed as a parameter.

For me anchorPoint needs to be in the name

ShaMan123: It doesn't have to be exported

asturur:
i see where the issue is with names, we can merge like this, please use only the export createPolyPositionHandler and createPolyActionHandler, i ll redivide the wrapper so it can be a wrapper. Now is not important.

asturur (approved):

PR #9404, opened 2023-10-07
CI(): fix build caching

### LINKED ISSUE #9403: [Bug CI]: tests are failing when merging to master
The test run that runs on push to master fails which is very strange
Check out the [commit history](https://github.com/fabricjs/fabric.js/commits/master)

ShaMan123:
## Motivation

closes #9403 

## Description

The build caching was producing a cache for the push to master event.
That is not desired because each time the build changes.
The cache key was wrong (`npm-`). It was generic so the stale cache was used for a long time.

## Changes

Cache the build only for PRs

### DIFF
--- a/.github/actions/build-fabric-cached/action.yml
+++ b/.github/actions/build-fabric-cached/action.yml
@@ -4,11 +4,12 @@ runs:
   using: 'composite'
   steps:
     - name: Get cache build
+      if: github.event_name == 'pull_request'
       id: build-cache
       uses: actions/cache@v3
       with:
         path: ./dist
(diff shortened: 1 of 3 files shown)


asturur:
This was supposed to rebuild every commit, so i wonder why with master wasn't good.
The sha of head of pr changes every commit and either changes or not changes when merging to master.
In each case the last build should be correct, no?

asturur:
oh right the merge event probably doesn't have a `github.event.pull_request.head.sha` but should have something similar we could have used, to reuse the latest one.

asturur (approved):

PR #9405, opened 2023-10-08
fix: toJSON function param

xiaoxintang:
## Motivation
fix toJSON function bug

## Description
`toJSON` is call `toObject` & it's doc show we can have an array param but actually not

## Changes
add optional param like `toObject` do


asturur:
Sorry this can't be done. You have to use toObject for the extra parameters. To JSON receives already some parameters, i think it receives the key of the object you are serializing ans so we can't put there a different argument.

asturur:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior

xiaoxintang:
so,you should delete this doc? @asturur 

image

asturur:
if the bug has been fixed in 5.3.x yes

Issue #9407, opened 2023-10-11
[Bug]: export svgColor not comply with svg specification


tyutjohn:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.0-beta13

### In What environments are you experiencing the problem?

Firefox, Chrome, Safari, Node.js

### Node Version (if applicable)

16.20.2

### Link To Reproduction

https://jsfiddle.net/fu21Lbd8/

### Steps To Reproduce

1. Use the following properties likes fill and backgroundColor or opacity，the exported svg is wrong,

![image](https://github.com/fabricjs/fabric.js/assets/36732345/cf5a2031-b6c7-493d-9523-78caaf960ff3)

The correct output is 
![image](https://github.com/fabricjs/fabric.js/assets/36732345/f19e3d12-26ee-472c-a00a-aecec6014b76)

2. I confirm this is a bug introduced in version v6，v5 is correct, like this:  https://jsfiddle.net/w2h5muf9/

### Expected Behavior

<rect fill="red" x="0" y="0" width ="10" height="10"></rect>

### Actual Behavior

<rect fill:"red" fill-opacity: "0.5;" x="0" y="0" width ="10" height="10"></rect>

tyutjohn:
I found the wrong file and will provide a PR to fix it. https://github.com/fabricjs/fabric.js/blob/master/src/util/misc/svgParsing.ts#colorPropToSVG

PR #9408, opened 2023-10-11
fix: colorPropToSVG comply with svg specification

tyutjohn:
## Motivation
@closes https://github.com/fabricjs/fabric.js/issues/9407

## Description

Change the output content of the colorPropToSVG function

## Changes
Change ':' to '='

How to write tests correctly, I need a little help, old test component seem doesn't work. This change is very small. Can anyone help me add test? I'm a little busy recently, Sorry

### DIFF
--- a/src/util/misc/svgParsing.ts
+++ b/src/util/misc/svgParsing.ts
@@ -82,8 +82,6 @@ export type MeetOrSlice = 'meet' | 'slice';
 
 export type MinMidMax = 'Min' | 'Mid' | 'Max' | 'none';
 
-export type ColorSeparator = ':' | '=';
-
 export type TPreserveArParsed = {
   meetOrSlice: MeetOrSlice;
   alignX: MinMidMax;
@@ -139,24 +137,37 @@ export const matrixToSVG = (transform: TMat2D) =>
  * we work around it by "moving" alpha channel into opacity attribute and setting fill's alpha to 1
  * @param prop
  * @param value
+ * @param {boolean} inlineStyle The default is inline style, the separator used is ":", The other is "="
  * @returns
  */
-export const colorPropToSVG = (prop: string, value?: any, separator: ColorSeparator = ':') => {
+export const colorPropToSVG = (
+  prop: string,
+  value?: any,
+  inlineStyle = true
+) => {
+  let colorValue;
+  let opacityValue;
   if (!value) {
-    return separator === ':' ? `${prop}: none; ` : `${prop}="none" `;
+    colorValue = 'none';
   } else if (value.toLive) {
-    return separator === ':' ? `${prop}: url(#SVGID_${value.id}); ` : `${prop}="url(#SVGID_${value.id})" `;
+    colorValue = `url(#SVGID_${value.id})`;
   } else {
     const color = new Color(value),
       opacity = color.getAlpha();
 
-    let str = separator === ':' ? `${prop}: ${color.toRgb()}; ` : `${prop}="${color.toRgb()}" `;
-
+    colorValue = color.toRgb();
     if (opacity !== 1) {
-      //change the color in rgb + opacity
-      str += separator === ':' ? `${prop}-opacity: ${opacity.toString()}; ` : `${prop}-opacity="${opacity.toString()}" `;
+      opacityValue = opacity.toString();
     }
-    return str;
+  }
+  if (inlineStyle) {
+    return `${prop}: ${colorValue}; ${
+      opacityValue ? `${prop}-opacity: ${opacityValue}; ` : ''
+    }`;
+  } else {
+    return `${prop}="${colorValue}" ${
+      opacityValue ? `${prop}-opacity="${opacityValue}" ` : ''
+    }`;
   }
 };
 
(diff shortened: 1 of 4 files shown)


ShaMan123:
I am guessing this is a regression, how didn't our tests catch this?

ShaMan123 (changes requested):
This seems wrong to me
Looking at usages in fabric the only usage that looks wrong is 
image

ShaMan123:
and indeed that is what you reported in your issue

ShaMan123:
I suggest refactoring the util `colorPropToSVG` to accept a third arg `separator: ':' | '='`

tyutjohn:
> refactoring

I don't know if it's necessary to add a third arg bacause I only found the function is used only here

tyutjohn:
> I suggest refactoring the util `colorPropToSVG` to accept a third arg `separator: ':' | '='`

On the old code I only found a file using it, do you want to add the default value ':' to the third parameter?

ShaMan123:
Find all usages and add what is needed IMO, very little usage.

tyutjohn:
> I am guessing this is a regression, how didn't our tests catch this?

How to do regression testing? The old test component should not test whether the exported svg is correct.

asturur:
I m actually trying to figure out which part was broken and then i can help you writing a test.

src/util/misc/svgParsing.ts
```diff
@@ -139,20 +141,20 @@ export const matrixToSVG = (transform: TMat2D) =>
...
     const color = new Color(value),
       opacity = color.getAlpha();
 
-    let str = `${prop}: ${color.toRgb()}; `;
+    let str = separator === ':' ? `${prop}: ${color.toRgb()}; ` : `${prop}="${color.toRgb()}" `;
 
     if (opacity !== 1) {
       //change the color in rgb + opacity
-      str += `${prop}-opacity: ${opacity.toString()}; `;
+      str += separator === ':' ? `${prop}-opacity: ${opacity.toString()}; ` : `${prop}-opacity="${opacity.toString()}" `;
```
asturur, line 157: the code is fine as it is, and i not holding an approval on this.
But you you want to step up the game of cleaning up, you can remove at least 2 ternaries with something like this:

```ts
let colorValue;
let opacityValue;
if (!value) { 
  colorValue = 'none';
} else if(...) {
  colorValue = `#SVGID_${value.id}`;
} else {
 ....
 do color value and opacity value;
}
if (separator === ':') {
  return `${prop}: ${colorValue} ${opacityValue ? `${prop}-opacity: ${opacity.toString()}; ` : ''}`
} else {
 return `${prop}="${color.toRgb()}" ${opacityValue ? `${prop}-opacity="${opacity.toString()}" ` : ''}`
}

```

this maybe is easier to read? just a suggestion follow your gut feeling is not a big deal

asturur:
for the test, create a file

src/shapes/Text/TextSVGExportMixin.spec.ts

and inside test an svg export with background color:

```ts
describe('TextSvgExport', () => {
  it('exports text background color correctly', () => {
    const myText = new Text('This is a test', { backgroundColor: 'rgba(100, 0, 100, 0.5)' });
    const svgString = myText.toSVG();
    expect(svgString).toInclude('fill=....') // find the correct string
    expect(svgString).not.toInclude('fill: ....') // find the other string
  })
});

```

in case .toInclude is not an expect matcher, just use expect(svgString.includes(...string)).toBe(true) and toBe(false)

Repeat the same test with different color to test 'none' and the other cases with a gradient i suppose.

asturur:
i also argue if seaparator is a good variable name.
Is kind of indirect, because the separator is the ':' for styles and '=' for attributes, but what change is not just the separator but the quotes too.
A non expert dev could say, ok the function support a separator i can add a different one, while what the function does is just  forInlineStyle true/false or forAttributes false/true. Probably a boolean is nicer than the separator string.

Not an holding issue for me, just thought it could be pointed out

tyutjohn:
> i also argue if seaparator is a good variable name. Is kind of indirect, because the separator is the ':' for styles and '=' for attributes, but what change is not just the separator but the quotes too. A non expert dev could say, ok the function support a separator i can add a different one, while what the function does is just forInlineStyle true/false or forAttributes false/true. Probably a boolean is nicer than the separator string.
> 
> Not an holding issue for me, just thought it could be pointed out

I completely agree with you, I will improve it and add a new test file to complete the testing of the function, thanks for you help

tyutjohn:
> > i also argue if seaparator is a good variable name. Is kind of indirect, because the separator is the ':' for styles and '=' for attributes, but what change is not just the separator but the quotes too. A non expert dev could say, ok the function support a separator i can add a different one, while what the function does is just forInlineStyle true/false or forAttributes false/true. Probably a boolean is nicer than the separator string.
> > Not an holding issue for me, just thought it could be pointed out
> 
> I completely agree with you, I will improve it and add a new test file to complete the testing of the function, thanks for you help

Work done, everything looks fine

asturur:
We also need a new line in the changelog.md file in the root folder, sorry i forgot to ask you that

tyutjohn:
> We also need a new line in the changelog.md file in the root folder, sorry i forgot to ask you that

Already added, please try again

asturur (approved):

asturur:
Since @ShaMan123 requested some changes, let's give him time to review, but i think this is great.
Thank your for your contribution

ShaMan123 (approved):
Good job guys

PR #9419, opened 2023-10-17
patch(matrix): expose `calcPlaneRotation`

ShaMan123:
## Motivation

I need to calc angles of planes/matrices in certain cases.
I prefer to avoid decomposition because in most cases I need the radian value and not the degree.
In addition the rest of the calculation is not needed and is wasteful in terms of perf.
Also I believe fabric can and should reduce its dependancy on decomposition for all the known and unknown reasons, some I discovered in #8767 
This is a small step in that direction.

## Description

Expose `calcPlaneRotation` and use it across fabric.
There a number of usages I left in functions I dislike and that I have changed in several open PRs.
I do not want conflicts nor do I want to patch them because they need extensive work.


ShaMan123 (commented):
small and concise

src/controls/util.ts
```diff
@@ -85,10 +83,9 @@ export function findCornerQuadrant(
   control: Control
 ) {
   //  angle is relative to canvas plane
-  const angle = fabricObject.getTotalAngle(),
-    cornerAngle =
-      angle + radiansToDegrees(Math.atan2(control.y, control.x)) + 360;
-  return Math.round((cornerAngle % 360) / 45);
+  const rotation = calcPlaneRotation(fabricObject.calcTransformMatrix()),
+    cornerAngle = rotation + Math.atan2(control.y, control.x) + twoMathPi;
+  return Math.round((cornerAngle % twoMathPi) / quarterPI);
```
asturur, line 88: can we type this function saying it returns a number explicitly?
I was confused for a memoment thinking we moved from degrees to radians

src/controls/controlRendering.ts
```diff
@@ -124,9 +124,8 @@ export function renderSquareControl(
   // this is still wrong
   ctx.lineWidth = 1;
   ctx.translate(left, top);
-  //  angle is relative to canvas plane
-  const angle = fabricObject.getTotalAngle();
-  ctx.rotate(degreesToRadians(angle));
+  //  rotation is relative to canvas plane
+  ctx.rotate(calcPlaneRotation(fabricObject.calcTransformMatrix()));
```
ShaMan123, line 128: > So please keep using it where it was, since you removed the overhead of qrDecompose from it. The conversion from radians to degree is a multiplication and is not worth loosing the optimization for the common use case.

are you talking about this change??
I don't get your reason, what do you mean?
Why change to an angle and back to rad? It is purly redudant

asturur:
yes i want to keep getTotalAngle in use because it avoids using the calcTransformMatrix when there is no need for it.
if you need calcPlaneRotation to shortcut some computation cost of qrDecompose that is fine, if you think is useful to expose that is fine too,

It may be redundant to convert an angle to a degree, but is way less than creating the matrix from scratch when there is no need to it because we just need the angle.

By putting calcPlaneRotation in place where a simple angle would be ok you are wasting more than you are gaining.
So that is not ok.

We added getTotalAngle because we had to take in account the new edge case of nested objects, but that is not the norm. That would have been just 'this.angle' before, and that is what it is in most of the cases.

Your Motivation:

Motivation 1:
```
I need to calc angles of planes/matrices in certain cases.
I prefer to avoid decomposition because in most cases I need the radian value and not the degree.
In addition the rest of the calculation is not needed and is wasteful in terms of perf.
```
^ this are facts and are correct, i can't really say nothing about. An additive change and some calculation change, fine.

Motivation 2:
```
Also I believe fabric can and should reduce its dependancy on decomposition for all the known and unknown reasons
```
^ this is a belief to which i disagree with and so the path of not changing is the preferred one.

That is not the direction we will go if we don't find a framework to collaborate in advance on changes.

ShaMan123:
> this is a belief to which i disagree with and so the path of not changing is the preferred one.
That is not the direction we will go if we don't find a framework to collaborate in advance on changes.

That is why I wrote **I believe**

> It may be redundant to convert an angle to a degree, but is way less than creating the matrix from scratch when there is no need to it because we just need the angle.
By putting calcPlaneRotation in place where a simple angle would be ok you are wasting more than you are gaining.
So that is not ok.

I don't see a case where a matrix is not created, please enlighten me.
Also, please understand that the last PRs you were very picky about asoteric perf issues so this feels kind of off. Indeed I did not think of the matrix creation because I am sure it is created once rendering starts or setCoords is called, and that happens in every use case.

I do not want to defend so enlighten me instead.

asturur:
is not esoteric performance issues.
calcTransformMatrix is more costly than accessing this.angle.
The cache validation part of getTransformMatrix is probably as expensive as calculating the matrix itself and is not done when render start, is done every time.

So all the time you are running findCornerQuadrant on an object that has no group, calling getTotalAngle will let us pick the best way to get angle for that situation.
That's it.

asturur (approved):

ShaMan123:
I actually forgot to expose this

Issue #9109, opened 2023-07-19
[Bug]: Cannot read properties of undefined (reading 'cloneNode')


natuan62:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.0-beta10

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

18.0.0

### Link To Reproduction

https://codesandbox.io/s/gracious-davinci-qvdr82?file=/src/App.vue

### Steps To Reproduce

**fabric.loadSVGFromString** with svg content

I think this block cause error

`xlink:href="#ding"` access to ding id not exist

```
  <use
      id="use6112"
      style="stroke:#4949e3;stroke-width:10.63"
      xlink:href="#ding"
      transform="translate(-835.03 468.69)"
      height="1052.3622"
      width="744.09448"
      y="0"
      x="0"
  />
```

Here is svg content
```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->
<svg
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns:cc="http://creativecommons.org/ns#"
    xmlns:xlink="http://www.w3.org/1999/xlink"
    xmlns:dc="http://purl.org/dc/elements/1.1/"
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns:ns1="http://sozi.baierouge.fr"
    id="svg5319"
    sodipodi:docname="basbeima_Face_1.svg"
    viewBox="0 0 627.53 689.12"
    sodipodi:version="0.32"
    version="1.0"
[...]

PR #9427, opened 2023-10-21
Fix(SVGParser) ignore missing xlink target issue on svg parsing

gloriousjob:
## Motivation

@closes  https://github.com/fabricjs/fabric.js/issues/9109

## Description

Sometimes an svg has use tags with an xlink with no target, which most parsers ignore. We should ignore these use tags as well.

## Changes

Adds a null check and returns early, preventing an NPE.
Also removed elementById since it was needed for IE8 support, which we no longer support (see support for everything else here: https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById).


ShaMan123 (commented):
This looks good
Thanks!
Can this happen with other directives like clip paths? Should we igonre them as well?

gloriousjob:
I didn't realize that was being used by the use tag (elements_parser apparently handles this). But I found an example and tested it. It worked normally and when I changed the id to a bad one, it ignored the clipping so looks like the code is already healthy for that.
 Only question I have now is how to get the 2 failures to pass. I'm not sure if this is because I'm using my own repo since I don't have write access on the official repo.

ShaMan123:
> I didn't realize that was being used by the use tag (elements_parser apparently handles this). But I found an example and tested it. It worked normally and when I changed the id to a bad one, it ignored the clipping so looks like the code is already healthy for that.

Could you add a test?


>  Only question I have now is how to get the 2 check failures to pass. I'm not sure if this is because I'm using my own repo since I don't have write access on the official repo.

You are right, it is because of write access.

gloriousjob:
> Could you add a test?

Tacked on additional test.

test/unit/parser.js
```diff
@@ -804,6 +804,36 @@
...
+      `<svg viewBox="0 0 128 128" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+      <path id="heart" d="M10,30 A20,20,0,0,1,50,30 A20,20,0,0,1,90,30 Q90,60,50,90 Q10,60,10,30 Z" />
+      <use clip-path="url(#myClip)" href="#heart" fill="red" />
+      </svg>`;
+
+    fabric.loadSVGFromString(string).then(({ objects }) => {
+      assert.equal(objects[0].constructor.type, 'Path');
+      done();
+    });
+  });
```
asturur, line 835: please add those to jests instead of qunit.

asturur:
Do we have an svg with missing xlink and can we see how the browser renders it?

test/unit/parser.js
gloriousjob: done

gloriousjob:
> Do we have an svg with missing xlink and can we see how the browser renders it?

The svg from the unit test looks like:


Pretty simple but without the fix, nothing renders in fabric whereas it renders like this with the fix.

asturur:
i moved the read shape into the defs element so we can test that the use element with the wrong clip-path still work without the clippath.
If the path is outside defs, you never know if the parser ignored the wrong clip-path attribute or ignored the whole `<use>` node

asturur (approved):

asturur:
@ShaMan123 i m merging this since is a good fix and we can push it up with the beta right away, ok?

ShaMan123:
sure

PR #9429, opened 2023-10-23
update(findTarget): Select the active object that is covered by other objects

zhe-he:
When I set `preserveObjectStacking` to true, if I drag an active object behind an object with a higher stacking order, I am unable to continue selecting the active object by clicking on the overlapping area. I found that the issue lies in the `findTarget` function, so I removed the relevant code to make it support my requirement. The image below shows the before and after effects, as well as an example demo.

https://github.com/fabricjs/fabric.js/assets/11306583/db5bfa74-1f58-49c2-89bf-0a8ed917442f

[demo](https://codepen.io/zhe-he-the-vuer/pen/poGoGjM)

Alternative solution: If you want to keep the part that I removed but still support the requirement shown in the image, could it be possible to introduce a parameter that allows users to customize it?

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -729,24 +729,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
         activeObject === this.searchPossibleTargets([activeObject], pointer)
       ) {
         // active object is not an active selection
-        if (!this.preserveObjectStacking) {
-          return activeObject;
-        } else {
-          const subTargets = this.targets;
-          this.targets = [];
-          const target = this.searchPossibleTargets(this._objects, pointer);
-          if (
-            e[this.altSelectionKey as ModifierKey] &&
-            target &&
-            target !== activeObject
-          ) {
-            // alt selection: select active object even though it is not the top most target
-            // restore targets
-            this.targets = subTargets;
-            return activeObject;
-          }
-          return target;
-        }
+        return activeObject;
       }
     }
 


asturur:
hi @zhe-he how this was behaving on 5.3.0? we should compare to understand if we broke selection or if we are proposing a change.

zhe-he:
@asturur  Thank you for your attention!

I wrote a similar example, which is a demo for version 5.3.0. They have the same result. The implementation method is to block this.preserveObjectStacking check, as shown in the screenshot below

[5.3.0 demo](https://codepen.io/zhe-he-the-vuer/pen/mdveKMm)

image

LBZZYZ:
@zhe-he
It seems that the [altSelectionKey](http://fabricjs.com/docs/fabric.Canvas.html#altSelectionKey) can address your needs. However, there isn't a global switch; you must set an alternative key.

zhe-he:
@LBZZYZ  
    Thank you, I didn't notice the altSelectionKey parameter.
    When A and B have an overlapping area R, and A is below B. Without selecting A, holding down an auxiliary key and moving to R can select A, which is a good way. However, when A is selected and the mouse is moved to R, it still needs to hold down an auxiliary key to select A, which is a strange solution.

LBZZYZ:
@zhe-he 
You're right, and therefore, I believe a global switch is still necessary. This would allow users to move without having to hold down their keyboards constantly.

asturur:
We can't bake in all the possible UX configurations in fabricJS.
We will have to reason on how interactivity work and make it something that can be overwritten with easy boundaries.
For now you have to override the method where the altSelectionKey is used and make your own switch

PR #9433, opened 2023-10-23
fix(#9172): dep export `Object`, `Text`, `Image`

ShaMan123:
## Motivation

https://github.com/fabricjs/fabric.js/commit/23430b187d2bc17540eafce17936c80958c274a5#r130686412

Creating a subclass is wrong in order to depreacte `Object`
The deprecated class is not part of the proto chain so it is not a deprecating change, it is breaking and useless because it can't be used, e.g. `instanceof`, prototype mutation etc. are all broken

## Changes

I simply re-exported the class with a different name
I was sure that was mentioned in the comments of #9172 but I missed it

## In Action

From VSCODE:
The deprecation is noticable, but the comment is not

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [next]
 
+- fix(#9172): dep export `Object` [#9433](https://github.com/fabricjs/fabric.js/pull/9433)
+
 ## [6.0.0-beta13]
 
 - fix(Object): fixes centeredScaling prop type [#9401](https://github.com/fabricjs/fabric.js/pull/9401)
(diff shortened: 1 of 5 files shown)


ShaMan123:
@asturur please review this and bump to beta15, I can't use beta14 without changing all fabric.Object instances and I perfer not to

asturur:
but what about Image and Text then? did we handle them differently?

ShaMan123:
true, I will push

ShaMan123 (commented):
ready

asturur (approved):

ShaMan123:
can you bump beta?

asturur:
sure

PR #9434, opened 2023-10-23
patch(Canvas): move event mouse:up:before earlier in the logic for more control

ShaMan123:
## Motivation

I want to disable mouseup default behavior from the `mouse:up:before` event
In mouse down it is possible with `delete canvas._target`, see https://github.com/fabricjs/fabric.js/blob/9c8e102d93c04c274842b5ea2d1ea84e787b1886/src/canvas/Canvas.ts#L1061-L1063
In mouse up it isn't because of the order of execution.
So I changed the `const target` to be defined after the `mouse:up:before` event is called

## Description

Please add this to beta15


jiayihu:
LGTM. For a long-term solution, returning `false` from `handleEvent` sounds a better solution, or even checking `event. defaultPrevented` on the event.

jiayihu (approved):
LGTM

ShaMan123:
> LGTM. For a long-term solution, returning false from `handleEvent` sounds a better solution, or even checking event. `defaultPrevented` on the event.

definately, I agree

jiayihu (approved):

src/canvas/Canvas.ts
```diff
@@ -788,9 +788,9 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
   __onMouseUp(e: TPointerEvent) {
     const transform = this._currentTransform;
     this._cacheTransformEventData(e);
-    const target = this._target;
     const isClick = this._isClick;
     this._handleEvent(e, 'up:before');
+    const target = this._target;
```
asturur, line 793: at this point let's also move isClick down here?

ShaMan123: I also want isClick to be available only on mouseup event

src/canvas/Canvas.ts
```diff
@@ -105,10 +104,6 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
    */
   private declare _dropTarget: FabricObject<ObjectEvents> | undefined;
 
-  declare currentTarget?: FabricObject;
-
-  declare currentSubTargets?: FabricObject[];
```
ShaMan123, line 110: unused so it seems

src/canvas/Canvas.ts
```diff
@@ -786,23 +780,25 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
       }
       return;
     }
 
     if (checkClick(e, MIDDLE_CLICK)) {
       if (this.fireMiddleClick) {
-        this._handleEvent(e, 'up', MIDDLE_CLICK, isClick);
+        this._handleEvent(e, 'up', MIDDLE_CLICK);
       }
       this._resetTransformEventData();
```
ShaMan123, line 803: why is this called here and not in the RIGHT_CLICK condition?

src/canvas/Canvas.ts
```diff
@@ -917,34 +913,29 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
         subTargets: targets,
         button,
-        isClick,
         pointer: this.getPointer(e, true),
         absolutePointer: this.getPointer(e),
         transform: this._currentTransform,
-      };
-    if (eventType === 'up') {
-      options.currentTarget = this.findTarget(e);
-      options.currentSubTargets = this.targets;
```
ShaMan123, line 946: I have no idea why this was like this
Nothing is using it in fabric so I removed it
Any objections?
The dev can call `findTarget` from an event handler if they need

src/canvas/__tests__/__snapshots__/eventData.test.ts.snap
ShaMan123, line 1: This is why I like snapshots
Everything is visible here
But I understand why someone dislikes it for the same reason

ShaMan123 (commented):
Changed `isClick` to populate only on up events including `up:before`
Removed used `currentTarget` `currentSubTargets` from the event data

src/canvas/Canvas.ts
```diff
@@ -786,23 +780,25 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
+    const transform = this._currentTransform;
+    const isClick = this._isClick;
     const target = this._target;
+
     // if right/middle click just fire events and return
     // target undefined will make the _handleEvent search the target
     if (checkClick(e, RIGHT_CLICK)) {
       if (this.fireRightClick) {
-        this._handleEvent(e, 'up', RIGHT_CLICK, isClick);
+        this._handleEvent(e, 'up', RIGHT_CLICK);
```
asturur, line 794: While the doc stated that isClick was good only for left button, i do think it works also for right and middle.
Let's check if it works and let's keep it in case

src/canvas/Canvas.ts
asturur: i would rather keep this, is kind of unrelated to the PR and i don't have the tools or the time into digging why this is here.
There is for sure some connection with the current transform that return a different target from the one that actually is possibly under the mouse, and untill today we surfaced it and patched it in this way.

test/unit/canvas_events.js
```diff
@@ -346,24 +346,6 @@
...
-    var opt;
-    canvas.on('mouse:up', function(_opt) {
-      opt = _opt;
-    });
-    canvas.__onMouseDown(e1);
-    canvas.__onMouseMove(e2);
-    canvas.__onMouseUp(e2);
-    assert.equal(opt.target, rect1, 'options match model - target');
-    assert.equal(opt.currentTarget, rect2, 'options match model - currentTarget');
-  });
```
asturur, line 365: yes we had a specific test that say, should return. So it isn't casual, is something we added for a reason and if we want to remove it it should be at least its individual revertable PR.

src/canvas/Canvas.ts
asturur: probably a bug? each event end has to delete the old data or getPointer will be broken until the next mouse move at minimum.

ShaMan123: https://github.com/fabricjs/fabric.js/pull/9449/files#r1374191785

asturur (approved):

PR #9449, opened 2023-10-27
refactor(Canvas): BREAKING: remove button from mouse events, delegate to event.button property

ShaMan123:
## Motivation

#9434 made me look at the unused `button` property

## Description

first merge #9434 

Remove the `button` property of `TPointerEventInfo` because of the following:

- it is unused 
- the dev can/should access it from `e` directly
- the values do not match `e.button` and cause confusion

## Changes

- removed `button` from event
- fixed `notALeftClick` in text click


src/shapes/IText/ITextClickBehavior.ts
```diff
@@ -10,12 +10,10 @@ import { ITextKeyBehavior } from './ITextKeyBehavior';
 import type { TOptions } from '../../typedefs';
 import type { TextProps, SerializedTextProps } from '../Text/Text';
 
-// TODO: this code seems wrong.
-// e.button for a left click is `0` and so different than `1` is more
-// not a right click. PR 3888 introduced this code and was about left clicks.
-function notALeftClick(e: MouseEvent) {
-  return e.button && e.button !== 1;
```
ShaMan123, line 17: because `e.button === 0` for a left click it would return `false`:
`notALeftClick(LEFT_CLICK_EV)` => `false` => is a left click 

so it was correct for left and right (`e.button === 2`) but not for middle click (`e.button === 1`) which would return false. Because it is rarly used we didn't catch it I think

It will account for event with no `button` property as before

src/constants.ts
```diff
@@ -15,9 +15,9 @@ export const ALIASING_LIMIT = 2;
 /* "magic number" for bezier approximations of arcs (http://itc.ktu.lt/itc354/Riskus354.pdf) */
 export const kRect = 1 - 0.5522847498;
 
-export const LEFT_CLICK = 1;
-export const MIDDLE_CLICK = 2;
-export const RIGHT_CLICK = 3;
+export const LEFT_CLICK = 0;
+export const MIDDLE_CLICK = 1;
+export const RIGHT_CLICK = 2;
```
ShaMan123, line 20: we can remove these values or not, they are used in a single place now
now they are the same values as `MouseEvent#button`

src/canvas/Canvas.ts
```diff
@@ -1060,19 +1043,15 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-    if (checkClick(e, MIDDLE_CLICK)) {
-      if (this.fireMiddleClick) {
-        this._handleEvent(e, 'down', MIDDLE_CLICK);
-      }
+    // if right/middle click just fire events
+    if ((e as MouseEvent).button) {
+      this[FIRE_CLICK_EVENT_MAP[(e as MouseEvent).button as 1 | 2]] &&
+        this._handleEvent(e, 'down');
+
+      this._resetTransformEventData();
```
ShaMan123, line 1054: added `_resetTransformEventData`

src/canvas/Canvas.ts
```diff
@@ -1055,19 +1042,15 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-      return;
-    }
 
-    if (checkClick(e, MIDDLE_CLICK)) {
-      if (this.fireMiddleClick) {
-        this._handleEvent(e, 'down', MIDDLE_CLICK);
-      }
+    // if right/middle click just fire events
+    if ((e as MouseEvent).button) {
+      this[FIRE_CLICK_EVENT_MAP[(e as MouseEvent).button as 1 | 2]] &&
```
asturur, line 1050:
```suggestion
      if (this.fireMiddleClick && e.button === 1 || this.fireRightClick && e.button === 2) {
        .. handle event
        .. reset transform data
        return
      }
```

i find this way more readable, maybe is slight more code, but the map to reference a name of a property is so much indirect i can't read it.

this should also make not necessary the outer if for the e.button

I understand that we are not considering the buttons from 3 to 5, nor when multiple buttons are down, but that is a separate issue i m also not sure which is the correct way to handle it.

ShaMan123: will do

ShaMan123: done

ShaMan123 (commented):
a

src/constants.ts
```diff
@@ -15,10 +15,6 @@ export const ALIASING_LIMIT = 2;
 /* "magic number" for bezier approximations of arcs (http://itc.ktu.lt/itc354/Riskus354.pdf) */
 export const kRect = 1 - 0.5522847498;
 
-export const LEFT_CLICK = 1;
-export const MIDDLE_CLICK = 2;
-export const RIGHT_CLICK = 3;
```
asturur, line 20: this is the perfect case for a const enum once in a while but we don't need to do it now.

asturur (approved):

PR #9450, opened 2023-10-27
dep(): up event `currentTarget`, `currentSubTargets`

ShaMan123:
## Description

First  merge #9434 
Remove up event unused `currentTarget`, `currentSubTargets`


asturur:
i don't think this should be removed.
Is there, is usable, is ok.

ShaMan123:
I saw the same about other things.
Sometimes you disagree and point out we want a lean codebase.
There is no usage and it is extremely simple to call `findTarget` from a mouseup handler.
Anyways I don't mind, feel free to close or merge

Issue #8336, opened 2022-10-01
ci(): action permissions for PR from fork


ShaMan123:
I read somewhere that PRs from forks receive a limited access token to avoid the fork accessing the repo.
This is why actions that do stuff like post a comment fail (https://github.com/fabricjs/fabric.js/actions/runs/3165603032/jobs/5154756609) but there is a work around with tokens or permissions.

see https://github.com/romeovs/lcov-reporter-action/issues/25

ShaMan123:
`pull-request: write` permission

Or a generating an access token for PRs with the required permissions

ShaMan123:
Needed for all actions that do something on the repo such as add a comment:
- build stats

asturur:
Ok i think i found something, i have to generate a new token and for build stats instead we need to actually do some pass through of the token.

PR #9455, opened 2023-10-27
ci(): try pull_request_target event to give access to fork prs

asturur:
## Motivation

This is trying to solve the issue in #8336 

The pull_request_target should allow the PR to run on the workflow of the base branch of the pr, so not allowing forks to commit changes to workflows, making the workflows they run trusted and so writable.

I have no idea how it works in combination with the pull_request event, i guess we can find out

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: ./.github/actions/cached-install
         with:
-          node-version: 18.x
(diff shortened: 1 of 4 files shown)


asturur:
no this is not what we want.

asturur:
https://github.com/actions/checkout/issues/518

need to try this

asturur:
for build stats we can revert, we compare the master branch with the pull request head instead of vice versa.

PR #9460, opened 2023-10-29
fix(Polyline, Polygon): Fix wrong pathOffset for polyline with the normal bounding box calculation.

### LINKED ISSUE #9459: [Bug]: Regression: creating a polyline/polygon with skewX at initialization time creates wrong bounding box
## Description

adding to a canvas a polyline with this code in lastest master

```js
var pts = [
  {
    x: 300,
    y: 200,
  },
  {
    x: 440,
    y: 340,
  },
  {
    x: 520,
    y: 220,
  },
  {
    x: 580,
    y: 280,
  },
];

var polyObj = new fabric.Polyline(pts, {
  fill: 'transparent',
  stroke: 'black',
  strokeWidth: 10,
  originX: 'center',
  originY: 'center',
  scaleX: 1.4,
  scaleY: 1.1,
  skewX: 10,
  strokeUniform: false, // doesn't change
  objectCaching: false,
});

canvas.add(polyObj)
```

Produces this polyline

image

while executing the same code on the execute tab of http://fabricjs.com/kitchensink produces this other polyline

image

## Notes

strokeUniform value does not seems to affect anything
Adjusting skew with the controls in latest master produces the correct bounding box
Adding skewX after the polyline has been initialized produces the correct bounding box

asturur:
## Motivation

The bounding box is wrong when skewX or skewY is in play.
This because the pathOffset that we need is different if we are using exactBoudningBox or the normal simple calculation

close #9459


ShaMan123 (commented):
This looks OK, I would prefer a playwright test to see it all but maybe that is too much (just lazy to test it locally)

I uncovered a related bug that should be addressed as well. Without it it will be hard to determine if this is a complete fix.
`setBoundingBox` uses `pathOffset` or the Path2D bbox values to reposition the Path/Poly  instance.
`pathOffset` is a vector in the instance plane, right? I am extremely confused and wasted 2 days on this.
`setPositionByOrigin` accepts a point in the parent plane so this seems to be wrong.
Please confirm

https://github.com/fabricjs/fabric.js/blob/4aeb8f07d071d922808fbbccec07d22702fac625/src/shapes/Polyline.ts#L196-L206

https://github.com/fabricjs/fabric.js/blob/4aeb8f07d071d922808fbbccec07d22702fac625/src/shapes/Path.ts#L299-L305

asturur:
> This looks OK, I would prefer a playwright test to see it all but maybe that is too much (just lazy to test it locally)
> 
> I uncovered a related bug that should be addressed as well. Without it it will be hard to determine if this is a complete fix. `setBoundingBox` uses `pathOffset` or the Path2D bbox values to reposition the Path/Poly instance. `pathOffset` is a vector in the instance plane, right? I am extremely confused and wasted 2 days on this. `setPositionByOrigin` accepts a point in the parent plane so this seems to be wrong. Please confirm
> 
> https://github.com/fabricjs/fabric.js/blob/4aeb8f07d071d922808fbbccec07d22702fac625/src/shapes/Polyline.ts#L196-L206
> 
> https://github.com/fabricjs/fabric.js/blob/4aeb8f07d071d922808fbbccec07d22702fac625/src/shapes/Path.ts#L299-L305

i m not sure they relate.

pathOffset is just a shift used at rendering time to reorder the points of the path around 0,0.
It has nothing to do with setPositionByOrigin.

left and top, width and height are in the scene plane and are calculated by the points positions.

pathOffset is in the  object plane and just offsets the commands to render the lines of the polygon around the center of the bounding box.

in theory they never mix one with another in a direct way

asturur:
If we want to add a test we can add a visual one.
I wish i didn't have to spin up a browser for each test.

ShaMan123:
> left and top are in the scene plane and are calculated by the points positions.

left and top are in the parent plane

> in theory they never mix one with another in a direct way

then it is a bug

ShaMan123:
The left/top values return from the bbox calculation of the Path/Poly are indeed in the object plane.
So they have nothing to do with left/top of the transform
Also width/height exist in the object plane and can't be used to position the object in the parent plane beause they do not account for the object transform

asturur:
> > left and top are in the scene plane and are calculated by the points positions.
> 
> left and top are in the parent plane
> 
> > in theory they never mix one with another in a direct way
> 
> then it is a bug

no @ShaMan123 the pathOffset is like that since forever and always worked.
It is an offset to help centering the drawing inside the bounding box.
pathOffset and left/top never mix.

asturur:
> The left/top values return from the bbox calculation of the Path/Poly are indeed in the object plane. So they have nothing to do with left/top of the transform Also width/height exist in the object plane and can't be used to position the object in the parent plane beause they do not account for the object transform

This is totally unrelated to this single bug here.

asturur:
> The left/top values return from the bbox calculation of the Path/Poly are indeed in the object plane. So they have nothing to do with left/top of the transform Also width/height exist in the object plane and can't be used to position the object in the parent plane beause they do not account for the object transform

- width and height aren't used for position, are used just for the offset of rendering inside the box. that should be fine
- top and left are used to position the object when you create a polygon/path in the canvas and probably calling functions based on _calcDimensions from inside a group will result in correct size but wrong position. We should open an issue an think how that could be solved.

ShaMan123:
Yes it is unrelated for sure but indeed a bug

asturur:
i tried to add a visual test as i wanted and then i made a new branch for the test becasue i m trying to move all the visual tests to playwright. so i ll merge this, follow up with test pr, and open an issue for setDimensions inside a group

ShaMan123:
Please take inyo account PRs I wrote that migrated tests so not to duplicate work
They are tagged with CI/CD

asturur:
This pr was waiting for more interaction tests but there is a lot of stuff on the CI right now.
I would like to merge the fix and  follow up with tests when possible

src/shapes/Polyline.ts
```diff
@@ -163,15 +163,20 @@ export class Polyline<
...
-      offsetY = bbox.top + bbox.height / 2,
-      pathOffsetX = offsetX - offsetY * Math.tan(degreesToRadians(this.skewX)),
-      pathOffsetY =
-        offsetY - pathOffsetX * Math.tan(degreesToRadians(this.skewY)),
       scale = new Point(this.scaleX, this.scaleY);
+    let offsetX = bbox.left + bbox.width / 2,
+      offsetY = bbox.top + bbox.height / 2;
+    if (this.exactBoundingBox) {
+      // Order of those assignments is important.
+      // The second assignment relies on offsetX being already changed
```
ShaMan123, line 171: I think it is better to put the comment above offsetY stating that it relies on the new offsetX or introduce more variables to make this clear

ShaMan123 (approved):

asturur:
ok i want to test the open PR about polycontrols with this fix merged in, i don't want something to be broken after this

PR #9461, opened 2023-10-30
fix(Text) Fix style transfer issue on a line that is not empty

### LINKED ISSUE #9028: [BUG]: text style onInput

asturur:
## Motivation

Fix las bug mentioned in #9028 that adds unwanted style on previous text when we insert a new line.
This bug is also present in 5.x

image

bug after inserting a new line

image

 
close #9028

## Changes

- when insert a new line and some text is available after the newLine we won't prestyle the char 0 of the new created line.


asturur:
This in theory removes the last known style text bug with typing.
I'm sure that with some combination of copy paste we can trigger some uncovered case

e2e/tests/text/adding-text/index.spec.ts-snapshots/5-after-pasting-splitByGrapheme-true.png
ShaMan123, line 1: there is another major bug here
fabric.js should be in the first line and not break, this is a regression I am seeing at work as well

ShaMan123 (commented):
I really do not like these patches
As I said and PRed the problem is deeper so you will keep finding these annoying edge cases
Reviewing this will take time

e2e/tests/text/adding-text/index.spec.ts-snapshots/5-after-pasting-splitByGrapheme-true.png
asturur: this is splitByGrapheme true so this will not take care of words, will just wraps whenever it feels it has to wrap.

asturur:
> I really do not like these patches As I said and PRed the problem is deeper so you will keep finding these annoying edge cases Reviewing this will take time

As i said, we don't refactor to fix bugs, and moreover we are not refactoring text anytime soon.
If we want to improve text code we need to spin up a new class that will slowly replace the older, so small dumb fixes to the current text code is all we can do.

asturur:
Anyway there is no rush here to merge this, this particular bug is here since before 5.

src/shapes/IText/ITextBehavior.ts
```diff
@@ -805,9 +805,10 @@ export abstract class ITextBehavior<
     copiedStyle?: { [index: number]: TextStyleDeclaration }
   ) {
     const newLineStyles: { [index: number]: TextStyleDeclaration } = {};
-    const isEndOfLine =
-      this._unwrappedTextLines[lineIndex].length === charIndex;
-    let somethingAdded = false;
+    const originalLineLength = this._unwrappedTextLines[lineIndex].length;
+    const isEndOfLine = originalLineLength === charIndex;
+
+    let someStyleIsCarringOver = false;
```
gloriousjob, line 811: Do you mean someStyleIsCarryingOver?

src/shapes/IText/ITextBehavior.ts
```diff
@@ -828,14 +829,16 @@ export abstract class ITextBehavior<
...
       // if is end of line, the extra style we copied
       // is probably not something we want
       this.styles[lineIndex + qty] = newLineStyles;
       styleCarriedOver = true;
     }
-    if (styleCarriedOver) {
+    if (styleCarriedOver || originalLineLength > charIndex) {
       // skip the last line of since we already prepared it.
+      // or contains text without style that we don't want to style
+      // just becuase it changed line
```
gloriousjob, line 841: because

gloriousjob:
Just checking but the example shows splitting a styled, then unstyled into two lines. Is the reverse handled as well (i.e., foo is unstyled, bar is red, and splitting doesn't make the b unstyled)?

src/shapes/IText/ITextBehavior.ts
ShaMan123: @gloriousjob you can use the suggestion markdown
image

```suggestion
    let someStyleIsCarryingOver = false;
```

e2e/tests/text/adding-text/index.spec.ts-snapshots/5-after-pasting-splitByGrapheme-true.png
ShaMan123: If you look at v5 that is not the same I think

e2e/tests/text/adding-text/index.spec.ts
```diff
@@ -80,5 +81,29 @@ for (const splitByGrapheme of [true, false]) {
...
+    await canvasUtil.click({
+      position: clickPointEnd,
+      delay: 200,
+    });
+    await page.mouse.down();
+    await page.mouse.move(1, clickPointEnd.y + 150, { steps: 15 });
+    await page.mouse.up();
+    // we remove them because space is finishing
+    await canvasUtil.press('Backspace');
+    // lets click where style end to show that we can add new line without carrying over
```
ShaMan123, line 95: I suggest usign `test.step` to organize the test instead of leaving comments
It is more clear when reading and when looking at test results

ShaMan123 (changes requested):
Playwright is a very very good tool that makes reviewing this PR simple
However the test passes on current master even though it is broken, please make it break

e2e/tests/text/adding-text/index.spec.ts
ShaMan123: screenshot
[85800a532aaae57a102d53213c2b8114a0cadb8e.webm](https://github.com/fabricjs/fabric.js/assets/34343793/dfb8a1cc-cd8a-4f35-9db6-db0b9a93557a)

ShaMan123: image

e2e/tests/text/adding-text/index.spec.ts-snapshots/5-after-pasting-splitByGrapheme-true.png
gloriousjob: Setting split should split the word up (as in the other split true tests). I wondered about the word grapheme since this is really a Unicode character split but apparently there's different takes on what's considered a character vs. grapheme. I think the question here is what was broken in v5 that was considered "working" and what is actually a regression.

src/shapes/IText/ITextBehavior.ts
```diff
@@ -828,14 +829,16 @@ export abstract class ITextBehavior<
       }
     }
     let styleCarriedOver = false;
-    if (somethingAdded && !isEndOfLine) {
+    if (someStyleIsCarringOver && !isEndOfLine) {
```
gloriousjob, line 832:
```suggestion
    if (someStyleIsCarryingOver && !isEndOfLine) {
```

src/shapes/IText/ITextBehavior.ts
gloriousjob:
```suggestion
      // just because it changed lines
```

e2e/tests/text/adding-text/index.spec.ts-snapshots/5-after-pasting-splitByGrapheme-true.png
asturur: splitByGrapheme is an option that was introduced for languages that do not use spaces to break text.
It doesn't split by unicode character but by grapheme clusters, if the appropriate splitter is provided.
Otherwise it splits with the rudimental grapheme splitter of fabricJS.
a grapheme can be made of more unicode codepoints/characters glued together. In our example it doesn't happen.

asturur: regarding ' I think the question here is what was broken in v5 that was considered "working" and what is actually a regression.' i think nothing broken in v5 was considered working for styles.
This bug wasn't listed anywhere and i didn't know or remember of it.

asturur:
i will add the test in master first, with steps as commented.
https://github.com/fabricjs/fabric.js/pull/9531/files?diff=unified&w=1

We are adding an extra test step to cover the bug, and we are adding a step construct between the blocks of the test.
The screenshots are slightly smaller to avoid dealing with too small percentages

asturur:
hopefully test fail, but i wouldn't count on it.

asturur:
so at commit [29dcc2e](https://github.com/fabricjs/fabric.js/pull/9461/commits/29dcc2e88261cc059d45f2691f510de48b857a3c) we have failure

asturur:
and at commit [b84e6f8](https://github.com/fabricjs/fabric.js/pull/9461/commits/b84e6f87694a1abfdee671cd7810e2b55e62e87f) we have good test.

Locally it passes anyway, but i can't do miracles, i should redo a different test that show this particular but bigger or find a way to only have a zone of the screenshot, but i can't find an option for it.

But also this is good enough

e2e/tests/text/adding-text/index.spec.ts
```diff
@@ -112,7 +112,7 @@ for (const splitByGrapheme of [true, false]) {
       // and there is no style on the part of text that follows, but there is visible text.
       await expect(await canvasUtil.screenshot()).toMatchSnapshot({
         name: `6-after-adding-a-newline-splitByGrapheme-${splitByGrapheme}.png`,
-        maxDiffPixelRatio: 0.01,
+        maxDiffPixelRatio: 0.009,
```
ShaMan123, line 115:
```
await expect(page).toHaveScreenshot(`${name}.png`, {
          clip: {
            x,
            y,
            width,
            height,
          },
          maxDiffPixels, // insteadof percentage
        });
```

ShaMan123 (commented):
There a typos that are commented by @gloriousjob 
And I cpmmented a way tp clip the snapshot and perform a more reliable diff using number of pixels instead of percentage

e2e/tests/text/adding-text/index.spec.ts
asturur: where is this clip in the documentation? i looked for something like this.

asturur: oh ok is an entirely different way to do visual comparison. the locator can't be clipped.

ShaMan123: but the image can, right?

e2e/tests/text/adding-text/index.spec.ts
```diff
@@ -110,10 +110,18 @@ for (const splitByGrapheme of [true, false]) {
...
+      await expect(page).toHaveScreenshot(
+        `6-after-adding-a-newline-splitByGrapheme-${splitByGrapheme}.png`,
+        {
+          clip: {
+            x: 0,
+            y: clickPointYellow.y - 20,
+            width: 120,
+            height: 120,
+          },
+          maxDiffPixels: 0.029,
```
ShaMan123, line 122: I thought this was an interger

ShaMan123 (commented):
The playwright test is failing

ShaMan123 (approved):

PR #9463, opened 2023-10-30
fix(Canvas): control down/up pointer

ShaMan123:
## Description

The pointer for mouseDownHandler/mouseUpHandler should be from the viewport

https://github.com/fabricjs/fabric.js/pull/8519/files#diff-72332330d8e00b2fda6e57d000a93e4e8dd8a46fdc8bcbdb3320c23b1509f8b5L748-L752

the bug exists on v5:
https://github.com/fabricjs/fabric.js/blob/5.x/src/mixins/canvas_events.mixin.js#L468-L472

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(Canvas): control down/up pointer [#9463](https://github.com/fabricjs/fabric.js/pull/9463)
 - feat(util): expose `calcPlaneRotation` [#9419](https://github.com/fabricjs/fabric.js/pull/9419)
 - refactor(Canvas): BREAKING remove button from mouse events, delegate to event.button property [#9449](https://github.com/fabricjs/fabric.js/pull/9449)
 - patch(Canvas): move event mouse:up:before earlier in the logic for more control [#9434](https://github.com/fabricjs/fabric.js/pull/9434)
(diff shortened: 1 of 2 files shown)


ShaMan123:
It is another control confusion
Controls work with oCoords, but then the down/up handler receive a pointer in the scene, why?


> But then if we are in a group we transform it in the group plane that from the object that are inside it still acts as a normal point since it maps 1:1 with their aCoord.

Also a confusion. Actually a patch/hack I did to fix controls under group without a lot of refactoring. But it makes no sense. All should be in the viewport as controls are

asturur:
yeah i don't know.
Don't we want to go away from the duality of coordinates?

your patch/hack (as you call it), could be less hacky by moving that plane shift inside the action handler with a wrapper, called 'support nested object'
That would make the api less confusing and leaving the developer the choice to use it or not during a custom action.
That is probably a good thing to think about

So for you since those works with oCoords they should work viewport coords.

Since we don't have any custom existing mousup/down coordinates why we don't offer a wrapper for actions like the withFixedPosition, and we call it withViewportCoordinates? so that we can change what the custom action see and who writes the custom action can pick up a wrapper or not.
I think what are you doing with your control dictates what in that moment is considered normal and correct.

One example is that i want to check where i m dropping a control compared to other objects and since all geometry function are getting the viewport version removed, i would then need the version with the scene pointer.

if i want to know if i m overing another control, for example to connect an arrow between 2 controls of 2 objects, the viewport one seems more usable.

ShaMan123:
That is why I promote `sendPointToPlane` and that is why I think we should send viewport coords and let the dev use sendPointToPlane.
Simple and easy and standardized

ShaMan123:
> Don't we want to go away from the duality of coordinates?

Indeed go away

ShaMan123:
In fact this the control API has ambiguous values for x, y passed args.
In actionHandler x, y are in the parent plane, in up/down handlers they are in the scene plane
That is clearly a bug but this PR is stale

PR #9469, opened 2023-11-01
fix(Object): Fix detection of falsy shadows in Object.needsItsOwnCache method

### LINKED ISSUE #9390: [Bug]: blurred text in zoom if use paintFirst: "stroke"
### CheckList

### Version

6.0.0-beta9

Chrome

None

### Link To Reproduction

https://codepen.io/jonnas123/pen/poqVQoG

### Steps To Reproduce

Hello, I'm new here, not sure if it's a bug or new feature request. Please go to the link and zoom the text. You will see the text is blurred. After I remove the paintFirst: "stroke" the blurred is gone. How to go with paintFirst: "stroke" without blurred? I need to use paintFirst: "stroke".

### Expected Behavior

After removing paintFirst: "stroke"

sctwo

### Actual Behavior

Before removing paintFirst: "stroke"

scone

_No response_

gloriousjob:
## Motivation
@closes #9390

## Description

Fix blur that can happen when zooming in on text. This is due to incorrect caching that should refresh instead.

## Changes

Update shadow on Objects to have undefined instead of null. Also fix cache check to check for not undefined rather than typeof.
Changing to undefined required having to change several tests which were expecting null and also to use an actual JSON object rather than a parsed string (string parse does not support undefined).
Utilized playwright to have an image compare for the test. I tried reverting the code and saw that it had a 0.02 diff so I used 0.01 to ensure it failed if the old code was present.

### DIFF
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -845,7 +845,7 @@ export class FabricObject<
       this.paintFirst === 'stroke' &&
       this.hasFill() &&
       this.hasStroke() &&
-      typeof this.shadow === 'object'
+      !!this.shadow
     ) {
       return true;
     }
--- a/test/unit/object.js
+++ b/test/unit/object.js
@@ -1305,34 +1305,6 @@
     assert.equal(object.shouldCache(), true, 'if objectCaching is false, but we have a clipPath, group not cached, we cache anyway');
 
   });
-  QUnit.test('needsItsOwnCache', function(assert) {
-    var object = new fabric.Object();
-    assert.equal(object.needsItsOwnCache(), false, 'default needsItsOwnCache is false');
-    object.clipPath = {};
-    assert.equal(object.needsItsOwnCache(), true, 'with a clipPath is true');
-    delete object.clipPath;
-
-    object.paintFirst = 'stroke';
-    object.stroke = 'black';
-    object.shadow = {};
-    assert.equal(object.needsItsOwnCache(), true, 'if stroke first will return true');
-
-    object.paintFirst = 'stroke';
-    object.stroke = 'black';
-    object.shadow = null;
-    assert.equal(object.needsItsOwnCache(), true, 'if stroke first will return false if no shadow');
-
-    object.paintFirst = 'stroke';
-    object.stroke = '';
-    object.shadow = {};
-    assert.equal(object.needsItsOwnCache(), false, 'if stroke first will return false if no stroke');
-
-    object.paintFirst = 'stroke';
-    object.stroke = 'black';
-    object.fill = '';
-    object.shadow = {};
-    assert.equal(object.needsItsOwnCache(), false, 'if stroke first will return false if no fill');
-  });
   QUnit.test('hasStroke', function(assert) {
     var object = new fabric.Object({ fill: 'blue', width: 100, height: 100, strokeWidth: 3, stroke: 'black' });
     assert.equal(object.hasStroke(), true, 'if strokeWidth is present and stroke is black hasStroke is true');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(Object): Fix detection of falsy shadows in Object.needsItsOwnCache method [#9469](https://github.com/fabricjs/fabric.js/pull/9469)
 - feat(util): expose `calcPlaneRotation` [#9419](https://github.com/fabricjs/fabric.js/pull/9419)
 - refactor(Canvas): BREAKING remove button from mouse events, delegate to event.button property [#9449](https://github.com/fabricjs/fabric.js/pull/9449)
 - patch(Canvas): move event mouse:up:before earlier in the logic for more control [#9434](https://github.com/fabricjs/fabric.js/pull/9434)
--- a/src/shapes/Object/Object.spec.ts
+++ b/src/shapes/Object/Object.spec.ts
@@ -1,3 +1,5 @@
+import { Shadow } from '../../Shadow';
+import { Rect } from '../Rect';
 import { FabricObject } from './Object';
 
 describe('Object', () => {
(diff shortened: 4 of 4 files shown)


src/shapes/Object/Object.ts
```diff
@@ -845,7 +842,7 @@ export class FabricObject<
       this.paintFirst === 'stroke' &&
       this.hasFill() &&
       this.hasStroke() &&
-      typeof this.shadow === 'object'
+      this.shadow !== undefined
```
ShaMan123, line 845: This is the fix if I understand correctly

ShaMan123: To repsect undefined, null, fabric.Object:
```suggestion
      !!this.shadow
```

ShaMan123 (changes requested):
Just gave a quick glance
Good catch!

Regarding the fix, I suggest extracting all the shadow undefined to a standalone PR, and rebasing this PR ontop of that or the other way around.
I am speaking on behalf of @asturur I am sure.
The fix itself can be simplified regardless to accomadate both undefined/null

Regarding the test, I don't see it runs in node , follow the template `e2e/tests/template-with-node`

ShaMan123:
regardless I agree shadow should be undefined and not null

src/shapes/Object/Object.ts
gloriousjob: So you're thinking even though we typed it to object or undefined, null might sneak in somehow?

ShaMan123: no
I am pointing this out to stress that the fix is not blocked by the undefined change

ShaMan123: and safer anyways

asturur:
yes shadow null is weird, let's fix it counting that one day the shadow type will be just an optional, so let's use the `!!shadow` that today covers both null and undefined.

But i can't see why this PR balooned up to 30 changed files, let me read everything slowly

src/shapes/Object/Object.ts
asturur: yes i agree, !!this.shadow accomodate everything for now and does not require us to do a large change like this.
I would prefer to change the types in a different PR.

asturur:
There is a reason why i don't want the visual test and i ll try to be exhaustive.

There are testing related reasons first.
Being blurry is not a bug in general, and being blurry is the side effect of the bug you experienced in your particular case.
The bug here is that you are asking for objectCaching false and you are forced to use the cache without a valid reason.

The bug is that the check on shadow was done on typeof shadow === 'object' and null is an object, so shadow or not shadow was always true.

The bug is that the function 
```
  needsItsOwnCache() {
    if (
      this.paintFirst === 'stroke' &&
      this.hasFill() &&
      this.hasStroke() &&
      typeof this.shadow === 'object'
    ) {
      return true;
    }
    if (this.clipPath) {
      return true;
    }
    return false;
  }
```

returns true when it shouldn't and so i would like to test this.

so we need to make an object, any, it doesn't have to be text, and test the output of this function, possibly with a description.

```ts
describe('needsItsOwnCache', () => {
  it('return false for default values', () => {
    const rect = new Rect({ width: 100, height: 100 }):
    epxect(rect.needsItsOwnCache()).toBe(true);
  });
  it('return true when a clipPath is present', () => {
     ...
  });
  it('return true when paintFirst is stroke and there is a shadow', () => {
     ...
  });
  it('return false when paintFirst is stroke and there is no shadow', () => {
     ...
  });
  it('any other relevant cases', () => {
     ...
  });
})
```

On top of this, there are other reasons why i don't think an e2e test is a good idea for this:
- being blurry is hard to detect, font rendering differences between OS and machine config are possibly stronger than the blurriness
- **being blurry depends on caching implementation, if we do a better cache that is not blurry the test is not covering the bug anymore**
- start a browser window to run this test is no joke for a ci, we need to use the e2e tests for flows and the fact that we use it also for rendering tests  is just because we want to get away from qunit  at some point, but i would rather not add them unless we are testing a rendering property itself.

I m sorry you made a 30 files PR and i m asking you to go back to a one line PR + a jest test.

Then if you want to open a following PR for the type change across files that is fine by me,

asturur (changes requested):
i left in a comment the reason for requesting changes

src/shapes/Object/Object.ts
gloriousjob: For the types change, are you suggesting changing more types from null? Just wanted to see the idea for the other PR

gloriousjob:
@asturur Understood about the reasons and I appreciate the insight.
Thanks for providing some test cases, though I found they were pretty much in the QUnit test so I just moved them over into jest, except using Rect like your example instead of Object. Oddly enough, the no shadow test passed in QUnit but failed in jest with the original code... I don't really understand that but sounds like jest correctly tested.

ShaMan123 (approved):
After reading comments by @asturur I also agree
Also I would add a test case of `paintFirst = fill` 
You could/should use `it.each` since all cases do the same and vary in params

src/shapes/Object/Object.ts
gloriousjob: Understood. We just don't want to touch as many times.

test/unit/object.js
```diff
@@ -1305,34 +1305,6 @@
...
-
-    object.paintFirst = 'stroke';
-    object.stroke = 'black';
-    object.shadow = {};
-    assert.equal(object.needsItsOwnCache(), true, 'if stroke first will return true');
-
-    object.paintFirst = 'stroke';
-    object.stroke = 'black';
-    object.shadow = null;
-    assert.equal(object.needsItsOwnCache(), true, 'if stroke first will return false if no shadow');
```
asturur, line 1323: lol look at this.

asturur (approved):

asturur (approved):

test/unit/object.js
ShaMan123: lol indeed

ShaMan123:
@gloriousjob good job! thanks

test/unit/object.js
gloriousjob: Yeah, for some reason, this was passing. I don't get why but the jest test failed with the old code so at least jest is working right.

ShaMan123: the test was wrong to begin with, you fixed it, that why we lol
  ```js
 it('returns false when paintFirst is stroke and there is no shadow', () => {
      const rect = new Rect({ width: 100, height: 100 });
      rect.paintFirst = 'stroke';
      rect.stroke = 'black';
      rect.shadow = null;
      expect(rect.needsItsOwnCache()).toBe(false);
    });
```

gloriousjob: D'oh, I see. I corrected based on the name and didn't notice the original returned the wrong thing.

PR #9470, opened 2023-11-01
fix(BaseConfiguration): set `devicePixelRatio` from window

### LINKED ISSUE #9464: [Bug]: Single or first StaticCanvas on page is blurry / doesn't respect enableRetinaScaling
### CheckList

### Version

6.0.0-beta14

Firefox, Chrome, Safari

None

### Link To Reproduction

https://jsfiddle.net/klevendev/quywt9ev

### Steps To Reproduce

1. Initialize two instances of StaticCanvas with enableRetinaScaling = true
2. Add some text to them
3. Display on a Retina screen

### Expected Behavior

I expected to see crisp text on both canvas elements.

### Actual Behavior

I saw blurry text on the first

image

---

The first StaticCanvas is always blurry on Retina. 

You can uglyhack around it by writing: 

`const foo = new StaticCanvas();`

Before your "proper" canvas. But that probably needs to be disposed and managed in some way.

_No response_

AdamMerrifield:
## Motivation
closes #9464 

## Description
In the browser `window.devicePixelRatio` is not set to the `config` object until the first run of `getEnv`. This needs to run only on the browser instance of fabric.

## Changes
Add a `getEnv()` call to index.ts so it runs a single time in the browser to set `config.devicePixelRatio`.

## In Action
https://jsfiddle.net/35Lxfq7v/1/


ShaMan123 (changes requested):
This will break SSR I believe

ShaMan123:
the question is why getEnv isn't called when the app starts

AdamMerrifield:
It looks like you're right, this does break SSR. I am not sure how to call this at app start while allowing SSR to continue to function other than wrapping this in a conditional of `typeof window !== 'undefined'`.

Calling getEnv inside the constructor of StaticCanvas seems to work with the SSR codesandbox example, but is not eloquent.

ShaMan123:
> It looks like you're right, this does break SSR. I am not sure how to call this at app start while allowing SSR to continue to function other than wrapping this in a conditional of `typeof window !== 'undefined'`.
> 


that is also breaking SSR
please find why it is not called, or rather what calls it for the second StaticCanvas or for the first Canvas

AdamMerrifield:
In the example above this is the trace for getEnv's first run which is during `canvas1.setDimensions`
Screenshot 2023-11-01 at 1 40 10 PM

The first use of `config.devicePixelRatio` is from `_setDimensionsImpl` in the constructor for `StaticCanvas`.
If you do not run `setDimensions` on either canvas, both are blurry https://jsfiddle.net/rcoj3pz6/

asturur:
There are 2 issues here:
1) calling getEnv calls a config.configure that is unexpected, because i m getting something and i execute a side effect that could potentially override my config?
2) somehow when a canvas start we need, or when a dom manager is created, we need to grab the config and understand if a custom value for device pixel ratio has been added, if not, we need to add the default.

Alternatively we need a custom node config class that has retinaScaling set to 1 and the browser version that default to window.devicepixelRatio.

ShaMan123 (dismissed):
Not the correct approach

ShaMan123:
No value can be set before loading or else SSR will break

ShaMan123:
fabric logic can't execute until the page loads in the client for SSR to work
This means that calculating stuff as const in the global scope when importing is not possible and that is why we exposed the config object and that is why `getEnv` defaults to the browser env when it is not set. It can't be set at import time because of SSR.
I did not investigate why StaticCanvas does not call `setEnv`. That is the bug. The solution is not to call it, but to understand first why it is not called.
Also the fact that the second canvas is correct points to that direction or else I would expect all canvases to be blurry because of not initializing env

ShaMan123:
I think I should add a e2e SSR test, @asturur thoughts?

ShaMan123:
> I see 2 issues here:
> 
> 1. calling getEnv calls a config.configure that is unexpected, because i m getting something and i do not know i m about to execute a side effect that could potentially override my config.
> 2. somehow when a canvas starts or when a dom manager starts we need to grab the config and understand if a custom value for device pixel ratio has been added, if not, we need to add the default.
> 
> Alternatively we need a custom node config class that has retinaScaling set to 1 and the browser version that default to window.devicepixelRatio.
> 
> The config module could also just have a `typeof window !== 'undefined' ? window.devicePixelRatio : 1` as the default value for the retina scaling?


This is how we solved the SSR issue (`typeof window !== 'undefined' ? window.devicePixelRatio : 1` will always equal 1 in SSR)
If we need to support configuring dpr we can simply add a check to see if it is defined and if not define it.

asturur:
@ShaMan123  for me the code as i see it at commit 61ba5539ab41a0c1bbbed46d37744971bed51ce3 is the way i would have solved it.
Solve the side effect and in ssr gives you 1.

I agree the only issue we have to avoid is that importing fabric breaks the code in node.
I do think is still on developer knowing that it can't run fabricJS code outside event handlers or useEffect(s) otherwise SSR will break.

If the developer needs to use SSR to produce an SVG to add to the dom at rendering time ( i don't know to send a placeholder of the canvas that looks like the canvas and then swap it to interactive ) that can be done, importing the  node version of fabric.

ShaMan123:
If you do that you can't import fabric outside of ueEffect and you have to use dynamic imports
Please let me take a look at this before moving forward.
I still did not hear why the method is not called as it should, we are debating the wrong thing

asturur:
@ShaMan123 why so?

having a typeof window !== undefined in node code is totally fine imho. It doesn't break anything.

On top of that the fact that we are calling a setConfig without dev consent, is a bit strange.
We could be overriding something already set.

ShaMan123:
can you explain why we should support changing the dpr apart from the dpr change event that we should support out of the box?

ShaMan123:
Sticking to the bug report, since changing `StaticCanvas` to `Canvas` fixes the bug it is signals that the bug is not hear.
Yes, it can be patched by endless changes but they will destroy SSR support that I worked hard to achieve

ShaMan123:
> having a typeof window !== undefined in node code is totally fine imho. It doesn't break anything.

That is wrong. It is breaking. Will SSR run this line on import? If so it is breaking

ShaMan123:
I asked to trace the call for a reason
I found the bug and it was not hard to find so adhearing my advice would have resoled this long before.
`_isRetinaScaling` accesses config.devicePixelRatio before getEnv is called.
Simple as that.
I will PR a fix

ShaMan123:
I still want to undersatnd why dpr is configurable
I see no reason for that apart for v5 no imports

ShaMan123:
To patch this you can call `getEnv()` from `_isRetinaScaling` before it accesses the config

asturur:
what do you mean by dpr?

ShaMan123:
DevicePixelRatio

asturur:
@AdamMerrifield  Sorry we got in a convoluted discussion over this PR, this PR was OK and we are going to merge it and then reiterate on additional improvements.
But this fixes the bug you found and should be merged

PR #9474, opened 2023-11-03
chore(): Reorder defaults for fabric object + patch `setRelativeXY`

gloriousjob:
## Motivation
Original idea was move fabric object constructor tests into jest. This revealed that the test was testing for a property that FabricObject doesn't even have and led to moving properties for that into InteractiveFabricObject, which did have them. Also created brief constructor test for InteractiveFabricObject for property in that class. 

## Description

Move InteractiveFabricObject defaults to its proper file so as not to have defaults not typed to FabricObject assigned. Re-arrange the remaining properties by class name and order in the class.

## Changes
- Move InteractiveObject defaults to its file
- Reorder defaults for fabric object by the class name and how the class has the properties ordered
- Move object constructor qunit tests to jest
- Move FX_DURATION to straighten since that's the only thing using it and it's in the parking lot
- Move `colorProperties` to static


gloriousjob:
I was curious what migrating some of the tests from qunit to jest might reveal because of my previous PR. Let me know if this makes sense or is a bit much (i.e., you don't want tests moved over).

asturur:
This is not wrong imho, but since i don't want to scroll the page up and down and check properties one by one, before adding this i will add a test in master that checks that the end result will be the same.

In general there is an issue with reordering things:
- useless diffs
- tainted history ( when did we change this file last time? 3 days ago? oh no it was just an alphabetical sort... )

so in general i m against diffs in pr that are done for example to reorder imports.

gloriousjob:
> This is not wrong imho, but since i don't want to scroll the page up and down and check properties one by one, before adding this i will add a test in master that checks that the end result will be the same.
> 
> In general there is an issue with reordering things:
> 
> * useless diffs
> * tainted history ( when did we change this file last time? 3 days ago? oh no it was just an alphabetical sort... )
> 
> so in general i m against diffs in pr that are done for example to reorder imports.

I understand not wanting to reorder so you don't add history you don't want (and you have to go somewhere else anyway to find the default so may not matter that the order is different). What do you think about moving the defaults that aren't in fabric object? Also about migrating tests? The migration revealed that selectable isn't a part of fabric object but the test was passing because the default is being crammed into there, even enough it's a property of a subclass. I can redo the branch and keep the original order but I think the owndefaults should represent what they are.

src/shapes/Object/defaultValues.ts
```diff
@@ -36,75 +36,48 @@ export const cacheProperties = [
...
-  strokeUniform: false,
-  dirty: true,
-  paintFirst: 'fill',
-  activeOn: 'down',
-  colorProperties: ['fill', 'stroke', 'backgroundColor'],
   clipPath: undefined,
   inverted: false,
   absolutePositioned: false,
-  FX_DURATION: 500,
+  centeredRotation: true,
```
ShaMan123, line 81: interactive, related to centeredScaling

src/shapes/Object/defaultValues.ts
```diff
@@ -36,75 +36,48 @@ export const cacheProperties = [
...
   skewX: 0,
   skewY: 0,
-  cornerSize: 13,
-  touchCornerSize: 24,
-  transparentCorners: true,
-  hoverCursor: null,
-  moveCursor: null,
+  originX: LEFT,
+  originY: TOP,
+  angle: 0,
```
ShaMan123, line 52: IMO opinion it makes sense to have the transform listed in the order iti s applied


translate
rotation
scaling, flipping
skewing

src/shapes/Object/InteractiveObject.ts
ShaMan123, line 1: I think it should reside next to `src/shapes/Object/defaultValues.ts` or even in the same file as another export

src/shapes/Object/defaultValues.ts
```diff
@@ -36,75 +36,48 @@ export const cacheProperties = [
...
-  borderColor: 'rgb(178,204,255)',
-  borderDashArray: null,
-  cornerColor: 'rgb(178,204,255)',
-  cornerStrokeColor: '',
-  cornerStyle: 'rect',
-  cornerDashArray: null,
-  centeredScaling: false,
-  centeredRotation: true,
+  // coming from AnimatableObject
+  colorProperties: ['fill', 'stroke', 'backgroundColor'],
```
ShaMan123, line 58: this should be a static prop or even not a prop

src/shapes/Object/defaultValues.ts
```diff
@@ -36,75 +36,48 @@ export const cacheProperties = [
...
+  // FabricObject itself
+  minScaleLimit: 0,
```
ShaMan123, line 60: this is related to scale

ShaMan123 (changes requested):
I also understand the diff point
However if we keep it contained to a standalone PR is it so bad
The other side of this point is we don't want to keep the code nt tidy just because of the diff.
The import order for me is a pain. But I let go of it.
Reordering makes sense because:
1. The migration was incremental so cleanup is needed, all development is actually
2. For accessibility of the code. We know how it is built, but the community doesn't and it makes it harder to contribute

ShaMan123:
migrating tests is welcome

asturur:
i m adding the test in master so we can merge this if the test pass

asturur:
@gloriousjob there is some nasty merge conflict. can you fix it?

asturur:
ok maybe i solved the conflict

asturur:
I added the test and merged master in, i let the discussion on the best ordering to you guys.
Choose one and don't touch it for long time after.

src/shapes/Object/defaultValues.ts
gloriousjob: So this one is actually used by FabricObject in the rotate method so it wouldn't make sense to move it to the subclass... unless we think the rotate method should move to InteractiveObject as well. To be honest, I'm a bit confused about the FabricObject class, because there's a FabricObject in Object.ts and a FabricObject in FabricObject.ts. The inheritance is Object > InteractiveObject > FabricObject but the class names look like FabricObject > InteractiveObject > FabricObject.

Thoughts?

src/shapes/Object/defaultValues.ts
gloriousjob: so sounds like angle should be below hieght, scale's above flip's?
Should I also add a comment like // transforms apply translation, then rotation, scaling, flipping, skewing?
Would you like the comments about the super classes removed to allow for where transforms should list?

src/shapes/Object/defaultValues.ts
gloriousjob: So you think this should move up with scaleY?

gloriousjob:
> I added the test and merged master in, i let the discussion on the best ordering to you guys.
> Choose one and don't touch it for long time after.

Thanks! I'll work with @ShaMan123 for now. Not saying we'll get to the perfect place but hopefully better.

src/shapes/Object/defaultValues.ts
ShaMan123: I see
You are right
I don't know what to say
Maybe we extract all transform props including centeredScaling to a standalone interface?

The rotate method should be exposed regardless of interactivity, along side scale, skew etc. but that is a different topic. I can point out where I did that.
The prototype chain of FabricObject is just a compromise we made to make an incremental TS migration from mixins to  classes.
It is not good enough and will require a major redesign of the code to change to a better composition based architecture.
So the division is actually a legacy design, a left over from the incremental TS migration. Mixins disguised as classes.

We discussed extracting stuff to managers and already started with simple ones (TextEditingManager) but the major managers have yet to be planned, e.g. caching, rendering etc.
Add to that the fact that we wanted to publish v6 a year ago and that there are 100 open prs and many more things we want to do.
I don't like this as well but since it is more internal design I guess we can live with it until we decide not to.

Hope that gave you some context.

src/shapes/Object/defaultValues.ts
gloriousjob: Made static

src/shapes/Object/InteractiveObject.ts
gloriousjob: Moved to default values.ts

src/shapes/Object/defaultValues.ts
gloriousjob: Thanks, it does. So you think we should have a TransformableObjectProps that ObjectProps extends and move centeredScaling to Object.ts?

ShaMan123: I would call it  ObjectTransformProps, yes
Then all related props reside there not needing to follow the interactie/non onteractive object legacy distinguish

src/shapes/Object/defaultValues.ts
ShaMan123: > Should I also add a comment like // transforms apply translation, then rotation, scaling, flipping, skewing?

No because we have a comment about it I think `composeMatrix` or something. You could point to that method.

> Would you like the comments about the super classes removed to allow for where transforms should list?

Which?

src/shapes/Object/defaultValues.ts
ShaMan123: answered by ObjectTransformProps thread

src/shapes/Object/AnimatableObject.ts
```diff
@@ -17,7 +17,7 @@ export abstract class AnimatableObject<
    * List of properties to consider for animating colors.
    * @type String[]
    */
-  declare colorProperties: string[];
+  static colorProperties: string[] = ['fill', 'stroke', 'backgroundColor'];
```
ShaMan123, line 20: I didn't see changes to the animation logic that depends on this

src/shapes/Object/defaultValues.ts
ShaMan123: > Would you like the comments about the super classes removed to allow for where transforms should list?

I see know what you mean. Yes I think they do not belong here and keep the legacy design we don't want stuck in our code and mind

gloriousjob: changed comment to reflect composeMatrix, pulled out comments about superclasses and minor reorder.

src/shapes/Object/defaultValues.ts
gloriousjob: Moved centered props to ObjectTransformProps and extended base Object class to use it.

src/shapes/Object/defaultValues.ts
ShaMan123, line 1: We can have a defaults in each file and export from here a spread of all
```
export const defualtValues = {
  ...fillDefaults,
  ...strokeDefaults,
 ...transformDefaults
}
```

what do you think?

src/shapes/Object/Object.spec.ts
ShaMan123, line 1: I don't see you removed old tests that this covers

src/shapes/Object/ObjectGeometry.ts
```diff
@@ -170,12 +170,12 @@ export class ObjectGeometry<EventSpec extends ObjectEvents = ObjectEvents>
...
-      point,
-      originX || this.originX,
-      originY || this.originY
-    );
+  setRelativeXY(
+    point: Point,
+    originX: TOriginX = this.originX,
+    originY: TOriginY = this.originY
+  ) {
+    this.setPositionByOrigin(point, originX, originY);
```
ShaMan123, line 178: this is not only nicer but is actually a fix for `setRelativeXY(point, 0, 0)` when originX/Y are not left/top

ShaMan123 (changes requested):
I pushed a few commits instead of explaining myself (that would take longer and would be less effiecient in all manners)
Hope you are OK with that
Should we export all these types to the consumer or just the final aggragated types?

src/shapes/Object/defaultValues.ts
ShaMan123: I prefer it because it seems more orderly and easy to maintain/find/navigate

ShaMan123:
Can you rebase/update from master?

src/shapes/Object/AnimatableObject.ts
```diff
@@ -53,7 +53,9 @@ export abstract class AnimatableObject<
     options: Partial<AnimationOptions<T>> = {}
   ): TAnimation<T> {
     const path = key.split('.');
-    const propIsColor = this.colorProperties.includes(path[path.length - 1]);
+    const propIsColor = AnimatableObject.colorProperties.includes(
```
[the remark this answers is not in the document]
gloriousjob: This one is odd to me but probably because other OOP languages wouldn't really support referencing a subclass like this. Is the idea that we keep AnimatableObject as is but can have a subclass with different props? Sounds good, just not something I'm used to.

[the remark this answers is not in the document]
ShaMan123: This is the way to access the static side of the class so indeed it allows you to use the static interface without limiting subclassing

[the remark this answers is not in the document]
gloriousjob: Makes sense. Just not something other languages would do. Applied as asked, although with typeof.

src/shapes/Object/Object.spec.ts
gloriousjob: Removed.

src/shapes/Object/defaultValues.ts
gloriousjob: My concern there would be that the PR was opened so that we could separate the properties of the base Object from the InteractiveObject so that unit tests would work correctly because of correct class encapsulation. Should transformDefaults include the interactive defaults or just be the ones in the base Object?
Also, what about other properties (e.g., is paintFirst a fill or a stroke property? Does objectCaching fall into any of these categories)?

I also got the impression that @asturur liked everything together so orderly may depend on the beholder (although I lean towards your view, @ShaMan123).

ShaMan123: It is a matter a taste, let's try to wrap this up.
`paintFirst` belongs to rendering props if we have something like that.
Regarding transform props, good point.

ShaMan123: > `paintFirst` belongs to rendering props if we have something like that.


Maybe next to objectCaching

ShaMan123 (approved):
This is already a lot better

src/shapes/Object/defaultValues.ts
gloriousjob: I think I’ll leave this alone due to the separation, but you’re welcome to provide a commit as long as it keeps the separation (it kind of would make more sense being in their files since that’s supposed to be the pattern).

The only other thing I’d like in this PR would be to rename the FabricObject in object.ts to BaseObject. This would reduce some confusion on which class is being referenced (the base object versus the one people use from the library). I understand why it was renamed since it’s not good practice to name a class something that JavaScript has as a keyword but duplicating another name seems just as bad. I’ll let you guys decide how to resolve this comment.

ShaMan123: That should be done in a separate PR
I am merging
Good job, glorious ;)

ShaMan123:
Can you add a changelog entry? Then I can merge.
I changed computer and I am not sure how to push to the PR anymore... previously I used GitHub desktop for that.

gloriousjob:
Not sure if I did it right but just checked in an additional line. I tried to make it match the lines around it.

ShaMan123 (approved):

PR #9478, opened 2023-11-03
chore(): rm `isClick` artifacts leftovers from #9434

ShaMan123:
## Description

I missed `isClick` props in #9434

## Changes

- rm remaining `isClick` props from events that are not mouseup
- ~rm `isClick` check inside `handleSelection` because it runs inside a condition that checks that already in mouse up~

### DIFF
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -242,7 +242,6 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
     const target = this._hoveredTarget;
     const shared = {
       e,
-      isClick: false,
       ...getEventPoints(this, e),
     };
     this.fire('mouse:out', { ...shared, target });
@@ -1296,7 +1295,6 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         e,
         target: oldTarget,
         nextTarget: target,
-        isClick: false,
         ...getEventPoints(this, e),
       };
       fireCanvas && this.fire(canvasOut, outOpt);
@@ -1308,7 +1306,6 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         e,
         target,
         previousTarget: oldTarget,
-        isClick: false,
         ...getEventPoints(this, e),
       };
       fireCanvas && this.fire(canvasIn, inOpt);
@@ -1511,7 +1508,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
    * - selects objects that are contained in (and possibly intersecting) the selection bounding box
    * - sets the active object
    * ---
-   * runs on mouse up
+   * runs on mouse up after a mouse move
    */
   protected handleSelection(e: TPointerEvent) {
     if (!this.selection || !this._groupSelector) {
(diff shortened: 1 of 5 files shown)


src/canvas/Canvas.ts
```diff
@@ -1546,13 +1542,10 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-      ? collectedObjects[0]
-        ? [collectedObjects[0]]
-        : []
-      : collectedObjects.length > 1
-      ? collectedObjects.filter((object) => !object.onSelect({ e })).reverse()
-      : collectedObjects;
+    const objects =
+      collectedObjects.length > 1
+        ? collectedObjects.filter((object) => !object.onSelect({ e })).reverse()
+        : collectedObjects;
```
ShaMan123, line 1548: why don't we call onSelect in this case?

ShaMan123 (commented):
we should merge this

src/canvas/Canvas.ts
asturur: because it gets solved by setActiveObject shortly below, while the object.length > 1 uses an collection.add method

src/canvas/Canvas.ts
```diff
@@ -1533,8 +1530,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       point2 = point1.add(new Point(deltaX, deltaY)),
       tl = point1.min(point2),
       br = point1.max(point2),
-      size = br.subtract(tl),
-      isClick = point1.eq(point2);
```
asturur, line 1537: this isClick here is not like the ones we removed from events. i prefer to leave it in.

src/canvas/Canvas.ts
ShaMan123: I see
This requires a comment

src/canvas/Canvas.ts
ShaMan123: Did you read my comment about this?

ShaMan123: The method is called only when the browser determines it is not a click
So there is no need and we can test that this line is always false to be sure

asturur: Yes i read it.
There are 2 points.
One is that this pr is open to clear something that was left over from the isClick in the mouse event object, this isn't. This is a different part of code, is not a leftover from the previous PR.

The other is that you can mouse move  as much as you want, but then if you go back an leave the mouse in the same point, this will still return true. You can also trigger a mouse move and end up with the mouse in the same exact position, because pixels are """large""" compared to the sensibility of mouse move, since there are no decimal pixels in mouse events.

For example, when i try to click with the trackpad and my finger moves a little bit, here is what i get

image

i added this in the sandbox:

canvas.on('mouse:move', ({ e }) => {
  console.log(e.clientX, e.clientY);
});

and as you see i fire a bunch of mouse move that are always on the same point

ShaMan123: restored though I do not understand the use case - the test might be incorrect

ShaMan123 (commented):
ready to merge

src/canvas/Canvas.ts
asturur: There is no use case as feature.
Is there to prevent misclicks or mis-selections.

asturur (approved):

asturur (approved):

src/canvas/Canvas.ts
ShaMan123: So i do not understand why the click condition does not cover it
For a chat

PR #9480, opened 2023-11-04
refactor(env): Change the way the environment and retina are initialized

### LINKED ISSUE #9464: [Bug]: Single or first StaticCanvas on page is blurry / doesn't respect enableRetinaScaling

ShaMan123:
## Motivation

closes #9464

## Description

StaticCanvas would call `_isRetinaScaling` before fabric would call `getEnv`.
In the browser env the devicePixelRatio config value was set during the first `getEnv` call (a contrain enforced to support SSR).
This resulted in a StaticCanvas that was initialized with a wrong devicePixelRatio value (defaulting to 1 in config)

## Changes

- rm `Canvas#_isRetinaScaling` - it did a useless check that is done in `getRetinaScaling`
- moved the node `setEnv` call the `index.node.ts` to keep the side effects in the index file.
- exposed `getDevicePixelRatio` that checks the `config.devicePixelRatio` and fallbacks to the env value. Having it evaluated each call is another fix bacause the value can change when moving the browser between screens
- `config.devicePixelRatio` has been made optional.

## TODO

I think we need to suport devicePixelRatio change event

## In Action

https://codesandbox.io/s/fabric-vanillajs-sandbox-forked-pc9hvk


asturur:
Before you use a lot of time on this plan, devicePixelRatio needs to be configurable, so its place is in the config.
The rest of the cleanup could be good.

asturur:
I also double checke that typeof window is totally fine. I was sure, but i double checked in case there was something with our code that made it unsafe in other ways.

i tried SSR with the next template from codesanbox and i set it to 3 in node, just to be sure i wasn't running the old code, and SSR works just fine.

image

ShaMan123:
What did you set to 3? devicePixelRatio?
What use case is that?
The use case I am talking about is using the window value and that can not be done on import:

image

ShaMan123:
> Before you use a lot of time on this plan, devicePixelRatio needs to be configurable, so its place is in the config. The rest of the cleanup could be good.

Why? Please explain. I do not understand what is the use case

src/env/index.ts
```diff
@@ -30,9 +30,18 @@ export const setEnv = (value: TFabricEnv) => {
...
 export const getFabricDocument = (): Document => getEnv().document;
 
 export const getFabricWindow = (): (Window & typeof globalThis) | DOMWindow =>
   getEnv().window;
+
+/**
+ * @returns the config value if defined, fallbacks to the environment value
+ */
+export const getDevicePixelRatio = () =>
+  Math.max(config.devicePixelRatio ?? getEnv().devicePixelRatio, 1);
```
ShaMan123, line 47: I do not like this but it is not a big deal if you think dpr belongs to config

ShaMan123 (commented):
updated description
this PR is ready

asturur:
> What did you set to 3? devicePixelRatio? What use case is that? The use case I am talking about is using the window value and that can not be done on import:
> 
> image

Yeah this you do will error. if you put a typeof in front of window and check for it to be not undefined is good.
you can't access window, you can check if is defined and the access it.

In my comment i specified i used 3, because i wanted to be sure i wasn't running the old code that was returning 1 on node, so in the config i added:
`typeof window !== 'undefined' ? window.devicePixelRatio : 3` just to be sure i was running my fresh built code and not some stale old one i had somewhere in the file system.

devicePixelRatio was configurable by choice, because sometimes you may want to limit it ( in some devices is 3.5 or 4) or simply you want to set it to 2 and you want to ignore extra value that is there because of the user zooming the window.

ShaMan123:
> you want to ignore extra value that is there because of the user zooming the window.

How does zooming affect dpr?

> Yeah this you do will error. if you put a typeof in front of window and check for it to be not undefined is good.
you can't access window, you can check if is defined and the access it.

This is my point exactly. You can't do that if you want to reliably access a window value in SSR because there will be no window at the time of the call. And then it means that fabric will use 1 as the dpr value instead of the window dpr value.

ShaMan123 (commented):
@asturur I applied a change to align better to what you suggested
Please review it again

asturur (approved):

PR #9481, opened 2023-11-04
tests(): Start moving visual tests to playwright

asturur:
## Motivation

This is an idea / proposal to have something simpler to setup when doing visual tests.
As the old QUNIT-VISUAL we can just prepare a list of functions with some metadata and have them run in sequence.
Those will run opening a single browser window and a single ci action, deprecating the old qunit-visual that runs on 3 node versions and 2 browsers.

The code is a bit messy and is not integrated ( on purpose ) with our generic pattern for single interactive e2e tests.
Eventually also my before function would be moved away from where i added it and just be specific of this kind of test.

I didn't convert more tests because i wanted to talk about it first.
Also is not urgent.


ShaMan123:
please merge #9110 #9333 before you migrate the tests

ShaMan123:
#9110 does what you suggest
use a single browser to run all the snapshots. It is very fast.
I remeber you had some things you wanted to change there.

e2e/tests/rendering/index.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+        ).toMatchSnapshot({
+          name: `${testCase.title}.png`,
+          maxDiffPixelRatio: testCase.percentage,
+        });
+      });
+    }
+    if (testCase.disabled !== 'node') {
+      await test.step(`node - ${testCase.title}`, async () => {
+        // we want the browser snapshot of a test to be committed and not the node snapshot
+        config.config.updateSnapshots = 'none';
```
ShaMan123, line 32: this is wrong because it will override the `-u` flag passed in argv

e2e/tests/test.ts
```diff
@@ -53,6 +56,26 @@ export function before(
...
+  renderingTests.forEach((renderTest) => {
+    if (renderingTestMap.has(renderTest.title)) {
+      throw new Error(
+        `test identifiers must be unique: ${renderTest.title} is already defined`
+      );
+    }
+    renderingTestMap.set(renderTest.title, renderTest.boundFunction);
+  });
+}
+
```
ShaMan123, line 78: I am not sure I under this.
Do you want to create a new canvas for every test?
Why? What is the gain?
What about simply resizing the canvas before the test and adding a `canvas.clear()` after the snapshot call?
Seems more straight forward, simpler and faster and can be exposed on the CanvasUtil.

ShaMan123 (commented):
Seems that Polyline code hs creeped in from another branch
I would do this differently as explained in the comments

PR #9483, opened 2023-11-05
refactor() BREAKING Remove calculate option from geometry methods

asturur:
## Motivation

All the geometry methods had a calculate optional parameter that was used to get the values without running setCoords.
setCoords is one of the learning step of fabricJS.
When you change a geometry property of an object, you have to call setCoords for the basic standard coordinates to be cached correctly.

FabricJS does that for every user interaction point, but for the code written by developer that is on the developer.
`calculate: boolean` defaulted to false and created more harm than good.

You could either:
- call setCoords, update the object, and then call obj.isOnScreen
OR
- call obj.isOnScreen(true) and get the same result but with wrong cached coords

this wasn't true for all the parts of fabricJS that work on the object coordinates, target check for example is one.
This led to situation where you could use calculate true, but then have to call setCoords anyway to get things working correctly.

At this point is easier to learn that if you change object position or size with your code, you have also to call setCoords for things to just works.

### Affected methods


asturur:
Needs rebase when the oher one is merged.

asturur:
this is reviewable now

src/shapes/Object/ObjectGeometry.ts
```diff
@@ -360,50 +334,41 @@ export class ObjectGeometry<EventSpec extends ObjectEvents = ObjectEvents>
...
    * @param {Point} pointBR Top Right point
-   * @param {Boolean} calculate use coordinates of current position instead of stored ones
    * @return {Boolean} true if the object contains the point
    */
-  private _containsCenterOfCanvas(
-    pointTL: Point,
-    pointBR: Point,
-    calculate?: boolean
-  ): boolean {
+  private _containsCenterOfCanvas(pointTL: Point, pointBR: Point): boolean {
```
ShaMan123, line 339: I would like to remove this method
Used twice in this module only

ShaMan123 (dismissed):
I would like to rename aCoords, oCoords to something like cornerCoords and controlCoords

ShaMan123:
I also want to remove `_getCoords` and maybe `calcACoords` or at least rename it

src/shapes/Object/ObjectGeometry.ts
ShaMan123: rm

ShaMan123 (dismissed):
looks good

ShaMan123 (approved):

PR #9484, opened 2023-11-05
DRAFT/EXPERIMENT building with esbuild

asturur:
## Motivation

The only motivation is speed.
image
I don't have really will to go over all the choices again about supported features, minification and all the rest, but the build speed with esbuild is very quick.

Leaving it here, maybe one day we have time to tackle it


asturur:
I want to try also 
https://swc.rs/ , https://www.npmjs.com/package/estrella, and https://tsup.egoist.dev/

I wanted to try those from the start, but at the end i preferred something i already knew a bit and knew how to use.

asturur:
Keeping the PR closed for now to avoid noise in the list

PR #9488, opened 2023-11-08
DRAFT: SVG export and isPath removal

asturur:
## Motivation

We discussed the type assertion and the removal of isPath, and its usage that excluded polygon and polyline in svg gradient export.
The test added reveals that indeed path export for gradient is slightly different, and so this PR would break it.
Better investigation is needed.

## Changes

remove internal utility isPath


asturur:
Ok the difference is that at time of svg export:
- path represent the pathOffset with a translation
- polygon absorbs the translation by subtracting it to each point value

so it make sense that the gradients with userSpaceOnUse behave differently.

Not sure is worth touching, probably it is, with the right priority.

An idea would be that the function:
`_getOffsetTransform` from Path gets duplicate to Polyline together with the toSVG one that adds the extra transform.

it would be made only to unify all the handling of objects that use a pathOffset object, that is not bad, but then we have to test for SVG exports regressions

PR #9490, opened 2023-11-10
test(): add SSR test

ShaMan123:
## Motivation

The recent SSR discussions and bugs related to env convinced me we need a test that shows SSR works so we can change env stuff with confident

## Description

I created a test using the next template

## Changes

I had to make the next template emit a signal so the test knows when the server/app is ready so I added logic to the template (instrumentation hook)
I adapted the cli start cmd by adding flags so that the browser and vscode launch only if flagged + added a port flag.
Added `tsconfig` to the vanilla app to fix a [parcel ts bug](https://github.com/parcel-bundler/parcel/issues/839#issuecomment-631801308)
Fixed node template so that after deploying it works (node version mismatch)


e2e/tests/SSR/index.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    '.codesandbox',
+    'templates',
+    'next'
+  );
+
+  let task: ChildProcessWithoutNullStreams;
+
+  test.beforeAll(({}, testInfo) => {
+    testInfo.setTimeout(60 * 1000);
+    task = exec(`npm start next -- -p ${PORT} --no-watch`, {
```
ShaMan123, line 22: once playwright/fabric are mjs then we can import `startSandbox` directly

ShaMan123 (commented):
Ready to merge, might seem a lot of changed files but most are minor patches

e2e/tests/SSR/index.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        .dblclick({
+          position: {
+            x: 625,
+            y: 50,
+          },
+        });
+      await page.getByRole('textbox').press('Control+a');
+      await page
+        .getByRole('textbox')
+        .fill("fabric supports SSR!\nIsn't that amazing?!");
```
asturur, line 55: This message is confusing.
What fabric support is only not blowing up node with an import that run browser code.
How do we ensure that the page was actually rendered by server?
Imho the test can just curl the output of  http://localhost:${PORT}/ and verify it contains the snapshot of html.
The fact that the page load could also simply mean that SSR isn't active, no?

I would also change `fabric supports SSR!\nIsn't that amazing?!` in a softer `a page that uses fabricJS will render ok with framework that use SSR`

ShaMan123: > How do we ensure that the page was actually rendered by server?

It doesn't, does it?
That is truly supporting SSR.
It is confusing saying the current state is SSR support.
I mean you can import fabric in SSR and it will not break. It will run on the browser after dehydration.
It is interesting to support SSR by rendering on node and sending it as an image or something.

ShaMan123 (commented):
applied minors

e2e/tests/SSR/index.spec.ts
ShaMan123: Regarding the test I just want to prove fabric works after window loads

.codesandbox/templates/vanilla/tsconfig.json
ShaMan123, line 1: fixes parcel class bug
https://github.com/parcel-bundler/parcel/issues/839#issuecomment-631801308

ShaMan123 (commented):
fixed a parcel bug and rebased
this is good to go and should merge ASAP

ShaMan123 (commented):
a few more crucial CD fixes

.codesandbox/templates/node/.codesandbox/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM node:18
```
ShaMan123, line 1: fixes deployed sandbox node version
our package.json require >16.20 I think so this fixes that
https://github.com/codesandbox/codesandbox-client/issues/1469#issuecomment-1817786803

.codesandbox/start.mjs
```diff
@@ -7,13 +7,29 @@ import { build } from '../scripts/build.mjs';
 import { subscribe } from '../scripts/buildLock.mjs';
 import { wd } from '../scripts/dirname.mjs';
 
+const TEMPLATE_PORTS = {
+  vanilla: 1234,
+  next: 3000,
+  node: 8080,
```
ShaMan123, line 13: a bad idea

ShaMan123 (commented):
finished

.codesandbox/start.mjs
ShaMan123: removed in favor of getPort

PR #9495, opened 2023-11-16
Update package.json

bigyifeng:
Don't need a canvas because it's useless


ShaMan123:
useless for you, but not for all the projects running in node

PR #9496, opened 2023-11-17
Issue 9475 bad divide breaks scaling in 0 height polyline

### LINKED ISSUE #9475: [Bug]: The bounding box disappear if scaling Y for a Rectangle whose strokeWidth = 0 and initial height = 0
### CheckList

### Version

6.0.0-beta15

Microsoft Edge

None

### Link To Reproduction

https://codepen.io/ritali4912/pen/PoVogxQ

### Steps To Reproduce

In my case, I create a rectangle with strokeWidth = 0 and initial its height = 0.

1. Create the rectangle with strokeWidth = 0 and initial its height = 0.
2. Active select it.
3. Drag the 'mt' or 'mb' to scaling Y.

P.S. In above CodePen, you need to select the top-left region through mouse to active the transparent rectangle.

### Expected Behavior

The rectangle successfully scale the Y.

### Actual Behavior

The bounding box disappear because the scaleY change to Infinity.

**In the beginning**
image

**Drag the control mt or mb**
image

You can see the scaleY got Infinity because the dim.y is 0.

I got correct behavior if I adjust the target.height in the beginning logic of scaleObject(): 
target.height = target.height === 0 ? 1/target.scaleY : target.height;
Same for target.width.

But I believe there is more reasonable place to adjust it.
Please check it.
Thanks.

_No response_

gloriousjob:
## Motivation

@closes #9475

## Description

0 height polygons have issues where they disappear when you scale them.

## Changes

Adds guards to ensure divide by 0's do not happen to create these problems. This is done by preventing the scale when value is 0.


src/controls/polyControl.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    exactBoundingBox: false,
+  });
+  const controls = createPolyControls(target);
+  target.controls = controls;
+  const corner = 'p1';
+  // create mouse pointer that moves the second point slightly
+  const pointer = new Point(target.points[1]).subtract(pointDiff);
+
+  // create transform event based on polyline
+  const transform: Transform & { pointIndex: number } = {
```
ShaMan123, line 30: does `canvas._setupCurrentTransform` instead do the job?
Maybe then assigning some values to it or using jest spies.

src/controls/polyControl.ts
ShaMan123, line 1: I don't understand the logic
Why don't we block the action instead?
Can we know in advance if a point causes 0 size? if not I guess we can revert the change in the action handler

src/controls/scale.ts
ShaMan123, line 1: same comment I guess
I would extend forbidScaling method or the forbidScaling flag and return
https://github.com/fabricjs/fabric.js/blob/44a75a3a7d9acd7d09d3ce4f5b0a92ec49d0680b/src/controls/scale.ts#L129-L131

src/controls/scale.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    rectOptions.height / 2,
+    rectOptions.width / 2
+  ).subtract(pointDiff);
+  // add rect to canvas or we get undefined issues
+  const canvas = new Canvas('canvas', { width: 100, height: 100 });
+  canvas.add(target);
+
+  // create transform event based on polyline
+  const corner = 'mt';
+  const transform: Transform = {
```
ShaMan123, line 29: same `_setupCurrentTransform`

ShaMan123 (commented):
See my comments
This is a good fix but I have a suggestion to make it more straight forward

src/controls/polyControl.ts
```diff
@@ -52,7 +52,15 @@ export const polyActionHandler = (
...
     size = poly._getTransformedDimensions(),
-    sizeFactor = polygonBaseSize.divide(size),
+    // if poly has 0 width or 0 height, NaN's occur so guard for that case
+    sizeFactor = new Point(
+      !size.x || Number.isNaN(size.x)
+        ? 1 / poly.scaleX
+        : polygonBaseSize.x / size.x,
+      !size.y || Number.isNaN(size.y)
+        ? 1 / poly.scaleY
+        : polygonBaseSize.y / size.y
```
asturur, line 62: i have the suspect this won't work with skewing.
for me `sizeFactor = polygonBaseSize.divide(size),` is basically a way to counter transform the skew + scale transformation of the object.
Instead of doing this we could try to calculate the scale/skew/flip matrix, invert it and use transform point on mouseLocalPosition? would that be equivalent?

I m not sure it would be for both the cases of exactBoundingBox and not.

src/controls/polyControl.ts
gloriousjob: All I'm doing is making sure the values do not become NaN. That said, there might be an alternative to trying to handle this.

The use case is two or more points (the issue's case had three but I'm using two in the tests) are in a flat line but then you move one of them and the line disappears due to the NaN. Should we not allow polygon controls in this state? I'd think we should handle it but would like input.

src/controls/polyControl.ts
gloriousjob: I'm not sure how to implement this (it's been a while since I did Matrices and linear algebra). We could separate the MR so that scaling gets fixed if we don't want to try to figure this out right now. The code is basically an expansion of the original code, just providing a value in the 0/NaN case and using the old value otherwise. Are you thinking polygonBaseSize.divide(size) wasn't handling skewing?

src/controls/scale.ts
```diff
@@ -70,6 +70,13 @@ export function scalingIsForbidden(
   if (lockY && by === 'y') {
     return true;
   }
+  // object disappears if scaling 0 sized object so forbid
+  if (fabricObject.width === 0 && by == 'x') {
+    return true;
+  }
+  if (fabricObject.height === 0 && by === 'y') {
+    return true;
+  }
```
ShaMan123, line 79: this is better, but there is the case that `by` is both and should be blocked as well
I would just block it all:
```
return fabricObject.width === 0 || fabricObject.height === 0
```

src/controls/scale.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  const canvasOffset = canvas.calcOffset();
+  const moveEvent = new MouseEvent('mousemove', {
+    clientX: canvasOffset.left + target.left + target.width / 2 - pointDiff.x,
+    clientY: canvasOffset.top + target.top + target.height / 2 - pointDiff.y,
+  });
+  canvas.setActiveObject(target);
+  target.__corner = target._findTargetCorner(
+    canvas.getViewportPoint(moveEvent)
+  );
+  canvas._setupCurrentTransform(moveEvent, target, false);
```
ShaMan123, line 31:
```suggestion
  canvas._setupCurrentTransform(moveEvent, target, true);
```

Don't you want to control in the transform?

src/controls/polyControl.ts
ShaMan123: I think a playwright test can prove you right.
I have an app that worksaround this issue by checking the changed point is not equal to its neighbors and if is it is shifted by `Number.EPSILON` or something

src/controls/scale.test.ts
gloriousjob: That's what I get for copying the first test from setupCurentTransform. Thanks!

src/controls/scale.ts
gloriousjob: I'll keep the if just to make clear that the case is particular but thanks for pointing out the other by case.

src/controls/polyControl.test.ts
gloriousjob: Pulling out code to isolate problem. Separate PR can be done for polycontrol issue.

src/controls/polyControl.ts
gloriousjob: Pulling out code to isolate problem. Separate PR can be done for polycontrol issue.

src/controls/scale.test.ts
gloriousjob: Updated as requested.

src/controls/polyControl.ts
gloriousjob: Pulling out code to isolate problem. Separate PR can be done for polycontrol issue.

gloriousjob:
I meant to only rename the branch but it closed the PR. I opened a new PR for the renamed branch.

PR #9500, opened 2023-11-19
refactor(#9498): refactor endCurrentTransform

### LINKED ISSUE #9498: `endCurrentTransform` severe bug
https://github.com/fabricjs/fabric.js/blob/44a75a3a7d9acd7d09d3ce4f5b0a92ec49d0680b/src/canvas/SelectableCanvas.ts#L1182-L1183

This is a bug because e is optional but `endCurrentTransform` does not accept an optional param causing a exception

ShaMan123:
## Motivation

closes #9498 

## Description

I encountered an exception thrown by fabric when using `_discardActiveObject => endCurrentTransform` w/o passing the `e` param which was typed as optional in `_discardActiveObject` but was not in `endCurrentTransform`/`modified` event.
Types were wrong, that is why we missed it.
I had to refactor some code and that made some cleanup possible.

## Changes

- squashed `_finalizeCurrentTransform` => `endCurrentTransform` and moved from `Canvas` to `SelectableCanvas`
- exposed `FabricObject#isTransforming(action)`
- rm `_scaling` flag: blame point to https://github.com/fabricjs/fabric.js/commit/20a87c66098e2328e7f60b94a33f52f1055bc451#diff-8eca4d2c9c7324551d9e6a7ce3817a987c9d598332b3f2f6ab02d2f1c98cde4fR1033 - not used by fabric => `isTransforming('scale')` can be used instead
- rm `isMoving`  => `isTransforming('drag')` is used instead
- refactor `_handleEvent` to support passing a custom object needed for mouse up after `_currentTransform` has been nullified


src/canvas/SelectableCanvas.ts
```diff
@@ -647,6 +647,33 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+    const { target, action, actionPerformed } = transform;
+    const options = {
+      e,
+      target,
+      transform,
+      action,
+    };
+
+    target.setCoords();
+    this._currentTransform = null;
```
ShaMan123, line 669: this is a change done when sqaushing `_finalizeCurrentTransform` requiring a bit more work to adapt mouse up logic

ShaMan123 (commented):
This looks good to me, let's see what tests have to say

asturur:
Somehow it was working in javascript before.
Can we make this optional without refactoring and renaming?

asturur:
Where is the exception thrown?
My understanding is that e will be undefined in the event handler, and so developer code will eventually thrown.
Or is fabric throwing by itself?

ShaMan123:
You understand correctly

ShaMan123:
stack call drill down:
`_discardActiveObject`
`endCurrentTransform` 
`_finalizeCurrentTransform`
fire `modified` event

I can do a patch that is not as clean as this and PR this on top. Should I?

ShaMan123:
Actually I think not because the only change needed is in types making e optional in the modified event

asturur:
Sorry i completely forgot about this on.
I wanted exactly to ask if we can just make the type optional and just write over it that the only case we know in which the event is missing is when we call endTransform by code and not by user actions?

ShaMan123:
> Sorry i completely forgot about this on. I wanted exactly to ask if we can just make the type optional and just write over it that the only case we know in which the event is missing is when we call endTransform by code and not by user actions?

I don't understand what you suggest. Isn't that this PR? Are you missing a comment in the code?

asturur:
we fix the type issue with this: https://github.com/fabricjs/fabric.js/pull/9596
Then when you can expand on why the other refactors are necessary or better or needed for some other bug we can evaluate the other changes as well

PR #9504, opened 2023-11-20
fix(): fire Poly control events

ShaMan123:
## Motivation 

#9503 
PolyControl do not have a wrapWithFireEvent

## Description

Fire an event when modifiying a poly via a poly control.

TODO:
- I have found while testing that `wrapWithFireEvent` will fire a stale `transform` object on the first call because `transform.actionPerformed` is not yet set on `transform` but is `true`. It is set after `wrapWithFireEvent` is called. That needs fixing.

## Changes

- add `wrapWithFireEvent`
- adapt poly control test

### DIFF
--- a/e2e/tests/controls/polygon-controls/index.spec.ts
+++ b/e2e/tests/controls/polygon-controls/index.spec.ts
@@ -30,58 +30,30 @@ for (const exact of [true, false]) {
       name: toSnapshotName('initial', exact),
     });
 
-    const modifyPoly = (
-      title: string,
-      controlName: string,
-      delta: { x: number; y: number }
-    ) =>
-      test.step(title, async () => {
-        const eventStreamHandle =
-          await test.step('subscribe to poly events', () =>
-            starUtil.evaluateHandle((object) => {
-              const events: { action: string; corner: string }[] = [];
-              const disposer = object.on(
-                'modifyPoly',
-                ({ transform: { action, corner } }) =>
-                  events.push({ action, corner })
-              );
-              object.once('modified', ({ transform: { action, corner } }) => {
-                disposer();
-                events.push({ action, corner });
-              });
-              return events;
-            }));
-
-        const controlPoint = await starUtil.getObjectControlPoint(controlName);
-        await page.mouse.move(controlPoint.x, controlPoint.y);
-        await page.mouse.down();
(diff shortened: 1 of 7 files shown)


ShaMan123 (commented):
ready

src/controls/polyControl.ts
```diff
@@ -126,7 +132,7 @@ export function createPolyControls(
     idx++
   ) {
     controls[`p${idx}`] = new Control({
-      actionName: 'modifyPoly',
+      actionName: ACTION_NAME,
```
asturur, line 135: Did we call it modifyPoly?
This is a terrible name.
Shouldn't be this just resizing? or modifying? it doesn't matter for this pr of course.

e2e/tests/controls/polygon-controls/index.spec.ts
```diff
@@ -1,83 +1,87 @@
...
-  expect(await canvasUtil.screenshot()).toMatchSnapshot({
-    name: 'moved_controls_p4_exact.png',
+    await modifyPoly('drag the p2 control', 'p2', { x: 300, y: 100 });
+    await modifyPoly('drag in the p4 control in the opposite direction', 'p4', {
+      x: -300,
+      y: -100,
+    });
   });
-});
+}
```
asturur, line 87: this test refactor blocks me.
Can we leave the tests as a simple sequence of steps?
In the moment we want to do something different this change becomes a difficulty more than an helper.

We can verify the event firing with a simple unit test i believe, we don't need to add it here.

src/controls/polyControl.ts
ShaMan123: I agree. BAD name. resizing... not percise. modifying  too general.

asturur: i have to write an issue to cover the fact that actionName of wrapWithFireEvent will not be in sync with control.actionName.
Not sure if is worth fixing or just documenting
control.actionName is used for end transform event, while this one is used for in transform event.
The current situation must be at least explained somewhere

asturur (approved):

src/controls/polyControl.ts
ShaMan123: I think it ought to be the same value. Why else?

PR #9507, opened 2023-11-21
refactor(): rename `_pointIsInObjectSelectionArea` => `isPointInObjectSelectionArea`

ShaMan123:
## Motivation

A bug in `TextEditingManager`
`_pointIsInObjectSelectionArea` is proving itself as a useful method, overriding it makes a lot of sense in many use cases (custom padding logic etc.)

## Description

Being able to override `_pointIsInObjectSelectionArea` is a must so it should at least be protected
I made it public because I don't see any reason why not to. It does not change internal state which is the only reason I see to have a method restricted in access.

## Changes

- exopse and rename `_pointIsInObjectSelectionArea` => `isPointInObjectSelectionArea`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - refactor(): Layout Manager [#9152](https://github.com/fabricjs/fabric.js/pull/9152)
 - refactor(): transferring object between active selections, expose `FabricObject#parent`, rm `isActiveSelection` [#8951](https://github.com/fabricjs/fabric.js/pull/8951)
 - refactor(): **BREAKING beta** `getParent` => `FabricObject#parent` [#8951](https://github.com/fabricjs/fabric.js/pull/8951)
+- fix(): `TextEditingManager` null exception + rename `_pointIsInObjectSelectionArea` [#9507](https://github.com/fabricjs/fabric.js/pull/9507)
 - fix(): fire Poly control events [#9504](https://github.com/fabricjs/fabric.js/pull/9504)
 - test(FabricObject): add a snapshot of the default values so that reordering and shuffling is verified. [#9492](https://github.com/fabricjs/fabric.js/pull/9492)
 - feat(FabricObject, Canvas) BREAKING: remove calculate true/false from the api. [#9483](https://github.com/fabricjs/fabric.js/pull/9483)
--- a/src/canvas/SelectableCanvas.spec.ts
+++ b/src/canvas/SelectableCanvas.spec.ts
@@ -4,7 +4,7 @@ import { Canvas } from './Canvas';
 import { Group } from '../shapes/Group';
 
 describe('Selectable Canvas', () => {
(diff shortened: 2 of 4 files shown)


ShaMan123 (commented):
done

src/canvas/TextEditingManager.ts
```diff
@@ -13,7 +13,8 @@ export class TextEditingManager {
 
   constructor(canvas: Canvas) {
     const cb = () => {
-      const { hiddenTextarea } = canvas.getActiveObject() as ITextBehavior;
+      const { hiddenTextarea } =
+        (canvas.getActiveObject() as ITextBehavior) || {};
       hiddenTextarea && hiddenTextarea.focus();
```
gloriousjob, line 18: Would it make sense to replace the two lines with this? Or do we foresee using grabbing more properties in the future?
```suggestion
      (canvas.getActiveObject() as ITextBehavior)?.hiddenTextarea?.focus();
```

ShaMan123: @asturur doesn't like how it transpiles.
I prefer it/don't care about transpiling

gloriousjob: If it makes a minor performance diff, I get it. Good to know.

gloriousjob (approved):
Only other thing I could think of is a test case showing an override of isPointInObjectSelectionArea but it doesn't make sense to test an override of every protected method so no need.

src/canvas/TextEditingManager.ts
ShaMan123: We might change a build tool and/or target more recent browsers making these considerations redundant

ShaMan123:
Just to say that this week I had to impl something regarding selection and this method makes it super easy.
I impl superb math functions using [`bezier-js`](https://github.com/Pomax/bezierjs), so things like selection an object by its shape and not by its bbox are quite straight forward using it and performance is outstanding.
It itches me to PR something into fabric (`isPointInStroke`, `isPointInShape`) but I won't for now.
Another very good point is that all fabric's path utils can be removed entirely in favour of `bezier-js`. It has it all and in better shape so I am for to do that at least. The path util is a complete mess to a point it took me hours to try to use it for these purposes.

src/canvas/TextEditingManager.ts
asturur: is way larger more than a performance diff. Here the performance is not meaningful.

it becomes something like this:

```
(_canvas$getActiveObje = canvas.getActiveObject()) === null || _canvas$getActiveObje === void 0 || (_canvas$getActiveObje = _canvas$getActiveObje.hiddenTextarea) === null || _canvas$getActiveObje === void 0 ? void 0 : _canvas$getActiveObje.focus();
```

I can't really like it.

ShaMan123:
closing => use `_checkTarget` overrides instead

PR #9510, opened 2023-11-24
refactor(Group) Layout manager, make the layout owned by the group.

asturur:
## Motivation

The current version of Layout manager allows for a developer to either pass or not pass a layout manager when creating a group.
So the group can have it one or not.
It will create one on the fly when creating the group, and accepting one when using trigger layout.
This bring to some added up difficulty for the api:
- you do not know if a group has its own layout manager or not
- If you don't have a layout manager on a group you do not know with which strategy the group has been sized
- when using triggerLayout you could pass a layoutManager with strategy A but also pass a strategy B in the options

Now all of this freedom maybe has its own usefulness, but for a first release i don't want it.

This PR changes the following

- when you create a group, if you don't pass a LayoutManager, you get the default fit content and it gets attached to the group
- when calling triggerLayout you can pass a strategy but the LayoutManager used will be the one that is owned in the group

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,7 +12,7 @@
   "editor.formatOnSave": true,
   "editor.formatOnPaste": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": false,
-    "source.removeUnusedImports": true
+    "source.organizeImports": "never",
+    "source.removeUnusedImports": "explicit"
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,7 +12,7 @@
   "editor.formatOnSave": true,
   "editor.formatOnPaste": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": "never",
-    "source.removeUnusedImports": "explicit"
+    "source.organizeImports": false,
+    "source.removeUnusedImports": true
   }
 }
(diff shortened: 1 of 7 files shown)


asturur:
@ShaMan123 @jiayihu

asturur:
## Open question:
if i have a group with a manager strategy A, and i ask a triggerLayout with strategy B, in my opinion the current layout manger ( and the only one that the group will see ) should be updated with that strategy that you just requested.

Opinions/Thoughts?

ShaMan123:
Agreed

ShaMan123:
On the ither hand this is an imperative call and the rest or declarative. Maybe that is why I made the difference in the first place.

asturur:
Which are the other ones?
We have group creation and trigger layout.
triggerLayout is imperative yes.
I don't see any another method aside this one, and group constructor.

There is an alternative, that you can't pass a strategy, and you first change the strategy manually and then you call the layout triggering.

ShaMan123:
Sorry
I meant 2 cases calling this the imperative by the dev and declarative by the events.
Simple is good. Let's move on with your suggestion

asturur:
You mean i remove the strategy from triggeLayout and we  do set before then call it?

asturur:
we can figure out that bit in the main PR.

PR #9513, opened 2023-11-27
fix(): `enterEditing` after `endCurrentTransform`

ShaMan123:
## Description

1. transform object
2. call `endCurrentTransform`
3. mouse up inside the object
4. object enters editing => this is not expected

### DIFF
--- a/src/shapes/IText/ITextClickBehavior.ts
+++ b/src/shapes/IText/ITextClickBehavior.ts
@@ -84,7 +84,7 @@ export abstract class ITextClickBehavior<
     this.__lastLastClickTime = this.__lastClickTime;
     this.__lastClickTime = this.__newClickTime;
     this.__lastPointer = newPointer;
-    this.__lastSelected = this.selected;
+    this.__lastSelected = this.selected && !this.getActiveControl();
   }
 
   isTripleClick(newPointer: XY) {


asturur:
@ShaMan123 i need to understand why you call endCurrentTransform manually.
This is an issue with the transition of the api, and is important to understand which is the use case for that.
Is not the first misfunction of endCurrentTransform that wasn't a thing indeed in 5.x

How do you use it?

src/shapes/IText/ITextClickBehavior.ts
```diff
@@ -84,7 +84,7 @@ export abstract class ITextClickBehavior<
     this.__lastLastClickTime = this.__lastClickTime;
     this.__lastClickTime = this.__newClickTime;
     this.__lastPointer = newPointer;
-    this.__lastSelected = this.selected;
+    this.__lastSelected = this.selected && !this.getActiveControl();
```
asturur, line 87: __lastSelected is at this point a wrong name.
It is used exclusively to stop the mouse up default action and does nothing else, is already private.

I would rename it to `_willEnterEditOnMouseUp' and add a JSDOC block on the top where is declared.

What do you think?

I would discourage anyone to build custom logic on those internals state flip flops

asturur: Doesn't even need to be done now. We can do an extra PR with that.

asturur (approved):

asturur (approved):

src/shapes/IText/ITextClickBehavior.ts
ShaMan123: I agree with the name
Since it is a very crucial part of state there are use cases that you need to block the enter editing somehow.
For the majority of cases this small fix should do

ShaMan123:
The use case:
A user is transforming, during which endCurrentTransform is called (reasons could be a user action like keyboard input, a db update changing something in the object etc.).
The user's mouse is still down and moving.
If mouseup occurs over the transform target it would enter editing and that is not expected.

PR #9515, opened 2023-11-28
refactor(): `getActiveControl` return value

ShaMan123:
## Motivation

`getActiveControl` is a convenient method we exposed in #9102 but not enough.
Using it made me realize the need to expose both the control and the coord of the control with it.
For us it is easy to get that but for a dev it means they need to understand the cood system.
Also it is not convenient trying to access the control/coord if the control key is undefined

## Changes

- **BREAKING beta** refactor(): `getActiveControl(): string` => `getActiveControl(): {
    key: string;
    control: Control;
    coord: TOCoord;
} | undefined`
- Made `Transform#action` optional so it is easier to manipulate transform. I want to do the same with `corner` but this is a big diff so i will in a dedicated PR.
- squash `_beforeTransform` into `_setupCurrentTransform`
- rm unused `Transform#reset`


ShaMan123 (commented):
- Made `Transform#action` optional so it is easier to manipulate transform. I want to do the same with `corner` but this is a big diff so i will in a dedicated PR.
- squash `_beforeTransform` into `_setupCurrentTransform`
- `_setupCurrentTransform` return value void => transform
- rm unused `Transform#reset`

src/canvas/SelectableCanvas.spec.ts
ShaMan123, line 1: migrated test from `test/unit/canvas_events.js`

ShaMan123:
Also I have found something I should PR to fix
If `alreadySelected` is false the control is not set in current transform but its mouseDownhandler is called
image

ShaMan123:
@asturur I want this in before we publish
Should I split this up into chunks?

ShaMan123 (commented):
I just reviewed again
This is a focused PR IMO

asturur:
next in line for possibly merging

src/canvas/Canvas.ts
```diff
@@ -1101,15 +1101,14 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-        if (mouseDownHandler) {
+        mouseDownHandler &&
           mouseDownHandler.call(
             control,
             e,
             this._currentTransform!,
             pointer.x,
             pointer.y
           );
-        }
```
asturur, line 1112: those changes here are applied by the minifier later on, but imho they take away on readibility. In the default safe lint or prettier profile are not considered, so i won't hold anything on it since may fall in the opinion realm

src/controls/util.ts
```diff
@@ -25,7 +25,7 @@ export const NOT_ALLOWED_CURSOR = 'not-allowed';
  */
 export const getActionFromCorner = (
   alreadySelected: boolean,
-  corner: string,
+  corner: string | undefined,
```
asturur, line 28: why not using the '?'

Smrtnyk: its usually better to init all props even if undefined, rather than having some missing and some not
that way browser engine can do smarter optimizations when objects have similar shapes

ShaMan123: I don't remember
I agree with '?'

ShaMan123: Ahh because it can't be optional because it is not the last arg

asturur: oh i thought it was an object.

asturur:
had a rough end of week, back on this now.

asturur (approved):

PR #9518, opened 2023-11-30
fix(Animation): safeguard zero values, fix rounding  issue, expose `promise` property

ShaMan123:
## Motivation

Making animation more robust to different use cases:

- `startValue === endValue`
- `duration <= 0` 
- rounding issues caused by `endValue` being recalculated in BaseAnimation instead of using  the passed value (small `byValue` and long `duration` maybe cause floating point imprecision)

## Description

- pass `endValue` to fix rounding issue
- expose `promise` that resolves/rejects on complete/abort
- handled the zero cases in `start` and complete the animation accordingly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(Animation): safeguard zero values, fix rounding issue, expose `promise` property [#9518](https://github.com/fabricjs/fabric.js/pull/9518)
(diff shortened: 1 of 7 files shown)


ShaMan123:
There is a bug in 
```ts
object.animate({ ...values }, { onComplete })
```

`onComplete` will fire on each of the changed values meaning that if my state counts on `onComplete` firing my state will be broken because I don't expect it to fire for each value, rather at the end of all animations.
Same goes for all callbacks.

ShaMan123:
I have moved to using a render loop so this is useless for me and seems is not needed/wanted in fabric

PR #9522, opened 2023-12-03
refactor(): Change how LayoutManager handles restoring groups

asturur:
## Motivation

Trying to remove the special boolean for the constructor for objects relative to center.
The issue with restoring an object from data is that the data is correct, and so the layout operation is not needed at all.
Instead of having logic inside the layout manager to do calculation but don't mutate objects or the position, we can just skip the layout manager all together.

I wish there was a way make that option not usable by a dev but i couldn't think of a simple solution.

TODO:
- document what to expect from group layout manager
- add a jest test that shows that the layout manager isn't called when requested

FOLLOWUP:
proper layout manager serialization and restore


ShaMan123:
However it doesn't respect overrides if someone want to override FitContentLayout and use it in fromObject... but that might be solved in that scope

ShaMan123:
so actually all we need is to pass an option to FitContentLayout in fromObject
```ts
export class FitContentLayout {
  performInitializationLayout = true
  shouldPerformLayout(context: StrictLayoutContext) {
    return context.type !== LAYOUT_TYPE_INITIALIZATION || this.performInitializationLayout;
  }
}
```

And then in `fromObject` we flag it

asturur:
i would like the layout manager to no have to deal with fromObject quirks.
What about the idea of passing for init a layout manager that is no-op and then we add the real one?

asturur:
something like this example commit here:
0d3474cd83d44d7497fed558092be146cf8c1e26

asturur:
> something like that can work but bear in mind overriding and customizing possibilities

I want to see how everything looks like when group can be properly serialized with the constructor and the layout managers in the serialization process.

Then if someone really need anyhing more than fit-content, fixed size and clip-path layouts will eventually complain or silently find a solution

asturur:
The change with the class has been done, i don't plan to touch anything else if not requested changes. @ShaMan123

asturur:
@ShaMan123 do you have time to review?

ShaMan123:
Not today
Tomorrow

asturur:
@ShaMan123 any chance we can move forward with this and proceed to finish the layout manager?

src/shapes/Group.spec.ts
```diff
@@ -15,6 +16,101 @@ describe('Group', () => {
...
+            height: 100,
+            top: 0,
+            left: 0,
+            strokeWidth: 0,
+          }),
+        ],
+        objectOptions
+      );
+      expect(group.left).toBe(6);
+      expect(group.top).toBe(4);
```
ShaMan123, line 88: seems off to me

src/shapes/Group.spec.ts
```diff
@@ -15,6 +16,101 @@ describe('Group', () => {
...
+
+    const group = await Group.fromObject(objectData);
+    expect(group.width).toBe(objectData.width);
+    expect(group.height).toBe(objectData.height);
+    expect(group.left).toBe(objectData.left);
+    expect(group.top).toBe(objectData.top);
+    group.triggerLayout();
+    expect(group.width).toBe(100);
+    expect(group.height).toBe(100);
+  });
```
ShaMan123, line 39: great test

src/shapes/Group.ts
```diff
@@ -644,21 +650,14 @@ export class Group
...
-        restoredGroup.height = options.height;
-      }
-      return restoredGroup;
+      const group = new this(objects, {
+        ...options,
+        ...hydratedOptions,
+        layoutManager: new RespectfulLayoutManager(),
+      });
+      group.layoutManager = new LayoutManager();
+      group.setCoords();
```
ShaMan123, line 659: do we need this?

src/shapes/Group.ts
```diff
@@ -28,6 +28,18 @@ import {
...
 } from '../LayoutManager/constants';
 
+/**
+ * This class is created to handle the specific case of the group fromObject.
+ * Is not meant to be used anywhere outside the fromObject function.
+ * We could have used a boolean in the constructor but we think the boolean
+ * would stay in the group's constructor interface and create confusion.
+ * The point of this LayoutManager is really just to don't touch anything.
+ */
+class RespectfulLayoutManager extends LayoutManager {
```
ShaMan123, line 38: Can we rename this?
DehydrationLayoutManager/NoopLayoutManager/FromObjectLayoutManager?

src/shapes/Group.ts
```diff
@@ -28,6 +28,18 @@ import {
   LAYOUT_TYPE_REMOVED,
 } from '../LayoutManager/constants';
 
+/**
+ * This class is created to handle the specific case of the group fromObject.
+ * Is not meant to be used anywhere outside the fromObject function.
+ * We could have used a boolean in the constructor but we think the boolean
+ * would stay in the group's constructor interface and create confusion.
+ * The point of this LayoutManager is really just to don't touch anything.
+ */
```
ShaMan123, line 37:
```suggestion
/**
 * This class handles the specific case of creating a group using {@link Group#fromObject} and is not meant to be used in any other case.
 * We could have used a boolean in the constructor, as we did previously, but we think the boolean
 * would stay in the group's constructor interface and create confusion, therefore it was removed.
 * This layout manager doesn't do anything and therefore keeps the exact layout the group had when {@link Group#toObject} was called.
 */
```

ShaMan123 (commented):
Looking good apart from a test I don't understand

src/shapes/Group.ts
asturur: is an internal name is not moving from this file, it doesn't really matter. i can call it Noop yes

src/shapes/Group.ts
asturur: yes because is the layoutManager that is doing setCoords for the group initialization.
It doesn't have to, but is doing it, it changes width and height or top and left, so it make sense it also calls setCoords.
If the the layoutManager is a no-op then a manual call to setCoords is necessary

src/shapes/Group.spec.ts
asturur: this is correct, the layout manager has been built like this:
```js
    if (context.type === LAYOUT_TYPE_INITIALIZATION) {
      // TODO: what about strokeWidth?
      target.set({
        left:
          context.x ?? nextCenter.x + size.x * resolveOrigin(target.originX),
        top: context.y ?? nextCenter.y + size.y * resolveOrigin(target.originY),
      });
    }
```
initialization will take position from context if available,

and in constructor:

```js
    layoutManager.performLayout({
      type: LAYOUT_TYPE_INITIALIZATION,
      target: this,
      targets: [...objects],
      x: options.left,
      y: options.top,
    });
```

we pass top and left as context.

src/shapes/Group.ts
asturur: renamed to NoopLayoutManager

asturur:
Ok off to the second part then, restoring fromObject.

src/shapes/Group.ts
ShaMan123: The layout manager did not call `setCoords` in the intialization layout.
  
Screenshot 2023-12-29 at 8 02 36

PR #9531, opened 2023-12-07
test() Backport test add text from the text fix to capture the failing situation first

asturur:
## Motivation

Modify the existing text test to capture the issue, so that the PR can just fix the bug and modify the snapshots.


asturur:
Not sure the pixel difference will work.

PR #9537, opened 2023-12-09
fix(): perform layout on poly change + initialization object subscription

### LINKED ISSUE #9534: [Bug]: What is the situation where a group composed of two polygons in fabricjs is restricted to a rectangular area when dragging points, and graphics are not displayed outside the rectangular area?
### CheckList

- [ ] I agree to follow this project's Code of Conduct
- [ ] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [ ] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [ ] I have searched and referenced existing issues and discussions
- [ ] I am filing a **BUG** report.
- [ ] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [ ] I have created an accurate and minimal reproduction

### Version

6.0.0-beta9

Chrome

16.13.1

### Link To Reproduction

The company's products are currently unavailable

### Steps To Reproduce

1. What is the situation where a group composed of two polygons in fabricjs is restricted to a rectangular area when dragging points, and graphics are not displayed outside the rectangular area?
2. Some of the codes are as follows：
` getPolyBox(points, ishole) {
        const boxAttr = this.configuration.boxAttr[BoxTypes.PolyBox] || {};
        const { default_line_color, line_width, point_size, default_point_color } = boxAttr || {};
        const polyboxes = [];
        let params;

        if (ishole) {
            params = {
                stroke: default_line_color,
                strokeWidth: Number(line_width) / this.canvas.getZoom(),
                fill: "rgba(0, 0, 0, 0)",
                name: BoxTypes.PolyBox,
                hasBorders: false,
                hasControls: true,
                objectCaching: false,
                originX: 'center',
                originY: 'center',
                cornerStyle: 'circle',
                cornerSize: point_size,
                cornerColor: default_point_color,
            }
            const polyBox1 = new fabric.Polygon(points, params);
            const params2 = { ...params }
            params2.globalCompositeOperation = 'destination-out'
            params2.fill = "rgb(22, 222, 0)"
            params2.isData = true
            const polyBox2 = new fabric.Polygon(points, params2);
            polyboxes.push(polyBox2)
            polyboxes.push(polyBox1)
        } else {
            params = {
                stroke: default_line_color,
                strokeWidth: Number(line_width) / this.canvas.getZoom(),
                fill: "rgba(0, 0, 0, 0)",
                name: BoxTypes.PolyBox,
                hasBorders: false,
                hasControls: true,
                originX: 'center',
                originY: 'center',
                cornerStyle: 'circle',
                objectCaching: false,
                cornerSize: point_size,
                cornerColor: default_point_color,
                transparentCorners: false,
                isData: true
            }
            const polyBox = new fabric.Polygon(points, params);
            polyboxes.push(polyBox)
        }
        return polyboxes
    }

    createGroup(polyboxes) {
        const boxAttr = this.configuration.boxAttr[BoxTypes.PolyBox] || {}
        const { default_line_color, line_width } = boxAttr || {}
        const group = new fabric.Group(polyboxes, {
            stroke: default_line_color,
            statusColor: default_line_color,
            strokeWidth: Number(line_width) / this.canvas.getZoom(),
            name: BoxTypes.PolyHole,
            noScaleCache: false,
            originX: 'center',
            originY: 'center',
            hasBorders: false,
            hasControls: false,
            // objectCaching: false,
            strokeUniform: true,
            centeredScaling: true,
            padding: 0,
            lockScalingFlip: true,
        });
        return group;
    }`
3.
CA3AB690-1FC5-45d8-B490-C00FC6C5B77B
74805A2E-606B-4c61-9CEA-640D3455EAD9
B5972982-D165-4fb5-9961-4EE6F3249D8B
2AC86103-1E4A-4681-B0EE-902BA8B43BAA

### Expected Behavior

Expect polygons to render properly

### Actual Behavior

As shown in the figure

_No response_

ShaMan123:
## Motivation

closes #9534 

## Description

Subscribe to the poly change event to layout group

## Changes

- fix: assign the layoutManager to group before initial layout so it subscribes objects properly. We should reconsider binding a group to a layout manager
- fix accounting for stroke width twice for Poly with `exactBoundingBox`

## In Action

https://codesandbox.io/p/sandbox/fabric-vanillajs-sandbox-forked-9fgf6r?file=%2Fsrc%2Findex.ts%3A28%2C69


ShaMan123 (commented):
DONE

src/shapes/Object/InteractiveObject.ts
```diff
@@ -436,13 +436,14 @@ export class InteractiveFabricObject<
...
-              new Point(options.scaleX, options.scaleY)
-        ).scalarMultiply(this.strokeWidth);
+        stroke = !this.isStrokeAccountedForInDimensions()
+          ? (this.strokeUniform
+              ? new Point().scalarAdd(this.canvas ? this.canvas.getZoom() : 1)
+              : // this is extremely confusing. options comes from the upper function
+                // and is the qrDecompose of a matrix that takes in account zoom too
+                new Point(options.scaleX, options.scaleY)
+            ).scalarMultiply(this.strokeWidth)
+          : ZERO;
```
ShaMan123, line 446: this is bad practice, having custom methods that repeatedly calculate the same thing differently.
We should IMO use aCoords and render in the scene plane. That way we do not need to handle so many cases and we can stick to fixing aCoords once instead f finding these issues again and again

src/shapes/Object/ObjectGeometry.ts
```diff
@@ -178,6 +178,10 @@ export class ObjectGeometry<EventSpec extends ObjectEvents = ObjectEvents>
     );
   }
 
+  isStrokeAccountedForInDimensions() {
+    return false;
+  }
```
ShaMan123, line 183: this isn't great as well but since @asturur yu do not want big refactors this is what I can offer

ShaMan123 (commented):
Fixed group layout edge case + border issue when `Poly#exactBoundingBox`

src/shapes/Object/ObjectGeometry.ts
asturur: The best thig i can think of is to kick the exactBoudingbox in its own polyline object.
I think long term makes more sense.
This seem unrelated from the PR entirely, can we remove it for now?

src/shapes/Group.ts
```diff
@@ -136,9 +139,7 @@ export class Group
     });
 
     // perform initial layout
-    const layoutManager =
-      // not destructured on purpose here.
-      options.layoutManager || new LayoutManager();
+    this.layoutManager = layoutManager;
```
asturur, line 142: i removed destructuring because is extremely noisy in the constructor, please don't change this since the change was fresh and commented. the rest operator on the function parameter becomes larger code and also slower. We do not get any benefit from extracting the layoutManager and assign it as a default in the constructor, since the layout manager is not used in the super class of the group.

src/LayoutManager/LayoutManager.spec.ts
```diff
@@ -772,6 +773,19 @@ describe('Layout Manager', () => {
...
+        width: 200,
+        height: 200,
+        strokeWidth: 0,
+      });
+      jest.spyOn(child, 'toJSON').mockReturnValue('child');
+      const group = new Group([child]);
+      expect(
+        Array.from(group.layoutManager['_subscriptions'].keys())
+      ).toMatchObject([child]);
+    });
```
asturur, line 787: i do not understand this test and why has to involve toJSON of fabricJS objects.

asturur: I mean i understand you want  to test the _subscription part but why do you have to mock toJSON. When we run code toJSON is not mockable.

asturur (changes requested):
Can we have PR with only the part 'fix(): perform layout on poly change + initialization object subscription' ?
The issues with exact bounding box are there, but i don't think excluding the stroke calculation for all the methods that usually account for it is a good without evaluating other options first

src/LayoutManager/LayoutManager.spec.ts
ShaMan123: That is exactly why I mock it, to avoid huge useless dumps
Make the test fail and see how elegant the dump is. It uses the mocked value in the error message.
It should be a test util we use to register all fabric instances in a test. I used it in one of my CI pending PRs but did not expose it as a util.
```
registerTestItem({ a, b, c, d})
```
Then the var name is used in test output

ShaMan123: So to explain.
The test was failing and the dump was unreadable so I added it for the future in case it does

src/shapes/Group.ts
ShaMan123: I don't mind about destruction, though having super setting values has become something I hate. I have encountered a number of hard to debug issues caused by fabric's usage of it. But that is a different topic.
I don't care about destructing or not but your change not to destruct the object introduced a regression and confusion.
It has to be set on the instance before calling performLayout or else it will not subscribe the objects.
Now, after you made group bind a layout manager I think we should remove that check all together but we need to consider how it will affect sharing a layout manager between groups and if we allow it or not.
So yes I prefer destructing because there is less ambiguity. It is just that a naive syntax introduced bugs.
You did not write why it was not destructed on purpose.

ShaMan123: To conclude, the syntax is 99% of the times something I can live with and bugs suck

src/shapes/Object/ObjectGeometry.ts
ShaMan123: I like this suggestion
Then we can mark the old one a legacy and set it in queue for deprecation or not.
It is related to the layout manager but indeed not to this PR.

ShaMan123:
@asturur I know how much you hate strokeUniform.
I have spent the past week and a half, maybe even 2 weeks on bug fixes all related to fabric's bugs when handling strokeUniform.
The amount of trouble was so big that at some point I decided to try your suggestion to listen to `object:scaling` and adapt objects' strokeWidth accordingly.
I regret I didn't write then when it was fresh in my mind beause I don't remember why I dumped that approach but I do remember it was extremely hard to manage and required me to override too many methods all across fabric. 
In other words it made me implement strokeUniform just in a different way and it was not easier.
Until that tryout I thought the same.

It all comes down to math. I fixed the math and everything worked.
The bug in fabric is a real issue and should be fixed.
Once fixed I don't think it will bother us any longer but since it is not it keeps troubling.
Just as a spoiler: I could not use ObjectOrigin methods because they all depend on the broken `_getTransformedDimensions` so even finding a center point was not trivial. This method is used everywhere in fabric and is the source of all problems (including scale control jumps when you start scaling and many more)
That and moving away from decomposition will fix a vast amount of bugs that might have been considered `won't fix`

asturur:
we need to make a good step forward.
You have to stop calling `_getTransformedDimensions` broken. Is the wrong approach to this codebase.
`_getTransformedDimensions` is not broken, it was doing its job, then stuff got added around it that weren't compatible, and now we point the finger to it. The finger needs to be pointed to strokeUniform.

Once that is clarified we can move on.

It doesn't look good to have '_getTransformedDimensions' to have to know if someone else will have to take in account for stroke. It doesn't even look good the boolean we have right now in getNonTransformedDimensions for polygon/polyline.

The exact bounding box has so much different math from text and rect that is probably fine as its own class with its own geometry methods. Who needs exact bounding box can initiate a different kind of polygon or polyline.

We can also merge this fix but my intention is to wipe the exact bounding box as they are today because mixed with the other logic they look as bad as the bug we had before having the option to use them.

asturur:
Please improve with PR description.
This pr has a title that cover 2 of the 3 changes, and the changes are listed one in description and 2 in the changes.
There is a suggestion with no explanations in the description of a bug.

ShaMan123:
> we need to make a good step forward. You have to stop calling `_getTransformedDimensions` broken. Is the wrong approach to this codebase. `_getTransformedDimensions` is not broken, it was doing its job, then stuff got added around it that weren't compatible, and now we point the finger to it. The finger needs to be pointed to strokeUniform.
> 
> Once that is clarified we can move on.
> 
> It doesn't look good to have '_getTransformedDimensions' to have to know if someone else will have to take in account for stroke. It doesn't even look good the boolean we have right now in getNonTransformedDimensions for polygon/polyline.
> 
> The exact bounding box has so much different math from text and rect that is probably fine as its own class with its own geometry methods. Who needs exact bounding box can initiate a different kind of polygon or polyline.
> 
> We can also merge this fix but my intention is to wipe the exact bounding box as they are today because mixed with the other logic they look as bad as the bug we had before having the option to use them.

Until we discuss this together I don't see how to explain it to you more than I have.
There are more things affected by this call such as `drawBorders`.
All of them can be fixed.
All require a refactor.
Any approach must break something to fix a bug.
That is why we need to discuss this face to face and plan it.
Saying it is not broken is saying to me I don't understand how to use fabric and that I spent almost 2 weeks of bug fixing because of what then? I don't think that is the case. True, all these objects are `strokeUniform`. 
Why not fix the math and make it work? 
You can remove it, you can remove many things that aren't fixed. But it has a cost that should be discussed as well.
If you made up your mind and I have no input that is a different matter.

src/shapes/Object/ObjectGeometry.ts
ShaMan123: I will extract this commit to a dedicated PR

src/shapes/Group.ts
ShaMan123: > Is there any case in which we end up executing that line with a layout manager and a group that are not bound together?

That is what I mentioned

> Now, after you made group bind a layout manager I think we should remove that check all together but we need to consider how it will affect sharing a layout manager between groups and if we allow it or not.

src/shapes/Object/ObjectGeometry.ts
```diff
@@ -178,6 +178,10 @@ export class ObjectGeometry<EventSpec extends ObjectEvents = ObjectEvents>
     );
   }
 
+  isStrokeAccountedForInDimensions() {
```
ShaMan123, line 181: deprecation notice

ShaMan123 (commented):
What about this? Should I extract the exactBoundingBox logic?

ShaMan123 (commented):
I dislike this as well, and it is due to not destructing
Screenshot 2024-01-06 at 12 15 56

ShaMan123 (commented):
reverted the exactBBox fix to a dedicated Pr

asturur (approved):

asturur:
So what is left for the layout manager? check that optional binding with the group?

ShaMan123:
I have another idea that might make all the hard to reason about correction logic void.
But that is a change that I am not sure you will like before v6.
It borrows the `pathOffset` concept and adds a transformation layer between the group and the objects, same as `pathOffset` for the path/poly points.
I might mean that correction will not be needed and that objects can have any origin and not center only as it is now.

Issue #9527, opened 2023-12-06
[Bug]:  Trimming a graphic with shadows will result in the loss of the canvas background.


zhe-he:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.0-beta16

### Node Version (if applicable)

None

### Link To Reproduction

[...]

PR #9540, opened 2023-12-11
fix(): bubble dirty flag

ShaMan123:
## Motivation

#9527

Safeguard flagging dirty from an infinite loop
This can happen when customizing group caching

## Description

Bubble the dirty flag only if it is true.
This is a fix not only to safeguard.
`child.set({ dirty: false })` should never bubble that to its group


asturur:
i can add the test and merge quickly

asturur (approved):

PR #9547, opened 2023-12-13
fix(): `calcOCoords`

ShaMan123:
## Motivation

Was experiencing severe performance hits caused by calcOCoords so I gave it a critical look

## Description

You can see in the video how the fix looks like
Keep to the bbox and not to the fact that control logic is broken.
This bug was a wrong multilication order.
https://github.com/fabricjs/fabric.js/blob/e4ac8b5284b2c1be80b6174d9b8d31e41136e8ce/src/shapes/Object/InteractiveObject.ts#L223-L230
One step towards #8767 

Performance: 
Reduce the amount of calculations by relaying on the calculated aCoords
Use a simpler transform builder

This also removes comlexity of logic/maintainance because it reuses aCoords instead of generating more code to do the same thing


ShaMan123 (commented):
fixed

ShaMan123:
isn't it bad that no test failed?

ShaMan123 (commented):
not perfect but better I think

src/shapes/Object/InteractiveObject.ts
```diff
@@ -212,81 +218,81 @@ export class InteractiveFabricObject<
...
-        position,
-        this._calcCornerCoords(control, position)
-      );
+      const position = control.positionHandler(size, t, this, control);
+
+      /**
+       * {@link coords} are sometimes used as points.
+       * Those are points to which we assign corner and touchCorner.
+       * Therefore, for now don't remove this assign for an object spread.
+       * §corner§/§touchCorner§ are used to determine the interaction area of a control.
```
ShaMan123, line 256:
```suggestion
       * `corner`/`touchCorner` are used to determine the interaction area of a control.
```

src/shapes/Object/InteractiveObject.ts
```diff
@@ -212,38 +215,37 @@ export class InteractiveFabricObject<
...
-        0,
-      ]),
-      transformOptions = this.group
-        ? qrDecompose(this.calcTransformMatrix())
-        : undefined,
-      dim = this._calculateCurrentDimensions(transformOptions),
-      coords: Record<string, TOCoord> = {};
+    const vpt = this.getViewportTransform();
+    const center = this.getCenterPoint();
+    const rotation = calcPlaneRotation(this.calcTransformMatrix());
```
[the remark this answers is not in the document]
ShaMan123: not a good idea if we want to support vpt transformations apart from zooming/panning

asturur:
> isn't it bad that no test failed?

No because there was no bug, the case you mention under bug is actually a non supported vieportTransform, so it wasn't tested.

src/util/misc/matrix.ts
```diff
@@ -166,11 +166,17 @@ export const createTranslateMatrix = (x: number, y = 0): TMat2D => [
  */
 export function createRotateMatrix(
   { angle = 0 }: TRotateMatrixArgs = {},
+  origin: Partial<XY> = {}
+): TMat2D {
+  return createRotationMatrix(degreesToRadians(angle), origin);
+}
+
+export function createRotationMatrix(
```
asturur, line 174: can we remove this double function? we talked about it in some meeting, to keep just the one with angle

src/shapes/Object/InteractiveObject.ts
```diff
@@ -212,81 +218,81 @@ export class InteractiveFabricObject<
...
+    const rotation = calcPlaneRotation(
+      multiplyTransformMatrices(vpt, this.calcTransformMatrix())
+    );
+    const angle = radiansToDegrees(rotation);
+
+    // calculate the viewport bbox size of the object by inverting the rotation
+    // in order to measure the bbox according to the object's axis
+    const bboxTransform = multiplyTransformMatrixArray([
+      vpt,
+      createTranslateMatrix(center.x, center.y),
```
asturur, line 232: why a translation is important if all you want to use at the end is width/height?

src/util/misc/matrix.ts
ShaMan123: Do you mind if I change the signature to accept angle as it does or rotation for the rad value?

src/shapes/Object/InteractiveObject.ts
ShaMan123: Because we apply the transform on the points with respect to the object center
If we do not translate the rotation will be applied at 0,0

src/shapes/Object/InteractiveObject.ts
```diff
@@ -212,81 +218,81 @@ export class InteractiveFabricObject<
...
+    const translation = center.transform(vpt);
+    const t = multiplyTransformMatrices(
+      createTranslateMatrix(translation.x, translation.y),
+      createRotationMatrix(rotation)
+    );
+
+    // calculate the controls' center point and touch region
+    const coords: Record<string, TOCoord> = {};
     this.forEachControl((control, key) => {
-      const position = control.positionHandler(dim, finalMatrix, this, control);
```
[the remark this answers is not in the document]
ShaMan123: They should be apart from the inberted vpt zokm that is applied at the correct order.

src/util/misc/matrix.ts
asturur: I think we have the api in angle, so no api with radiants. calcRotateMatrix is good enough.

src/shapes/Object/InteractiveObject.ts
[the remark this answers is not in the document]
asturur: so this would displace all controls, those needs to be the same for when the zoom is a standard one.

ShaMan123:
Handled by redo coords, ported to https://github.com/ShaMan123/fabric.js/pull/4

PR #9550, opened 2023-12-17
perf(): reduce setCoords calls

ShaMan123:
## Description

After removing lineCoords we have 2 sets of coords
Corner coords (`aCoords`) used for object geometry (`isOnScreen`, intersection etc.) that belong to the parent plane, `getCoords` returning them in the scene plane.
Control coords (`oCoords`) used for control rendering/hit detection that belong to the viewport plane.

This distinction is very important and useful.
It means that viewport changes do not affect corner coords. 
Also since they belong to the parent plane they need to be recalculated only if the object itself changes its own transform/size. This means that if an ancestor changes transform we do not need to recalculate the corner coords.
Since control coords exist only on the active object it seems important to separate use cases of `setCoords` to reduce computation.

I have found that fabric used to call `setCoords` before drawing the controls (at the end of the rendering cycle) instead of after a transform action. This led to a stale coords state between the 2 calls affecting apps using corner controls in the rendering cycle. It was always a frame behind. This is fixed now.

Looking into `setCoords` in general it seems we should decide when they are first set. It is ambiguous.
Does it happen when an object mounts the canvas? What about an object added to a group? We can check if the group has a canvas set, therefore it is mounted , therefore `setCoords` should be called.
It also means that object geometry should not be used until an object is mounted.

This PR is part of a performance effort I am looking into.
I have a ground breaking idea about a binary index that will reduce computation cost of `isOnScreen` exponentially.
I wish to contribute it to the repo but can't wait with it more than a number of days so if it can't merge soon I will depart from fabric (at least with this) or try to expose it as a standalone but that will be a shame.

Another change I have introduced is `interactive` including `subTargetCheck` when true.
To be able to select sub targets both need to be flagged as true.
`subTargetCheck` flags `searchPossibleTargets` to traverse the children of a group.
`interactive` flags to return the top most sub target found within `searchPossibleTargets`.
So essentially `interactive = true` contains `subTargetCheck = true`. 
It is more straight forward as well.

### DIFF
--- a/src/LayoutManager/LayoutManager.ts
+++ b/src/LayoutManager/LayoutManager.ts
@@ -251,6 +251,8 @@ export class LayoutManager {
       target.setPositionByOrigin(nextCenter, CENTER, CENTER);
       // invalidate
       target.setCoords();
+      (target.subTargetCheck || target.interactive) &&
+        target.forEachObject((object) => object.setCoords());
       target.set('dirty', true);
     }
   }
@@ -286,6 +288,8 @@ export class LayoutManager {
       left: object.left + offset.x,
       top: object.top + offset.y,
     });
+    // TODO: should we delete aCoords/oCoords here so they are not stale if subTargetCheck is false?
+    //  or should we apply offset instead of recalculating aCoords?
   }
 
   protected onAfterLayout(
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1332,7 +1332,6 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       this.setCursor(this.defaultCursor);
       return;
     }
-    let hoverCursor = target.hoverCursor || this.hoverCursor;
     const activeSelection = isActiveSelection(this._activeObject)
         ? this._activeObject
         : null,
@@ -1345,17 +1344,11 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         target.findControl(this.getViewportPoint(e));
 
     if (!corner) {
-      if ((target as Group).subTargetCheck) {
-        // hoverCursor should come from top-most subTarget,
-        // so we walk the array backwards
-        this.targets
-          .concat()
-          .reverse()
-          .map((_target) => {
-            hoverCursor = _target.hoverCursor || hoverCursor;
-          });
-      }
-      this.setCursor(hoverCursor);
+      // hoverCursor should come from top-most subTarget
+      const hoverCursor =
+        (target as Group).subTargetCheck &&
+        this.targets.find((target) => target.hoverCursor)?.hoverCursor;
+      this.setCursor(hoverCursor || target.hoverCursor || this.hoverCursor);
     } else {
       const control = corner.control;
       this.setCursor(control.cursorStyleHandler(e, control, target));
--- a/src/canvas/Canvas.ts
+++ b/src/canvas/Canvas.ts
@@ -1304,19 +1304,16 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
     transform: Transform,
     pointer: Point
   ) {
-    const x = pointer.x,
-      y = pointer.y,
-      action = transform.action,
-      actionHandler = transform.actionHandler;
-    let actionPerformed = false;
-    // this object could be created from the function in the control handlers
+    const { action, actionHandler, target } = transform;
 
-    if (actionHandler) {
-      actionPerformed = actionHandler(e, transform, x, y);
-    }
+    const actionPerformed =
+      !!actionHandler && actionHandler(e, transform, pointer.x, pointer.y);
+    // This call needs to recurse down
+    actionPerformed && target.setCoords();
+
     if (action === 'drag' && actionPerformed) {
-      transform.target.isMoving = true;
-      this.setCursor(transform.target.moveCursor || this.moveCursor);
+      target.isMoving = true;
+      this.setCursor(target.moveCursor || this.moveCursor);
     }
     transform.actionPerformed = transform.actionPerformed || actionPerformed;
   }
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -188,8 +188,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
   declare fireMiddleClick: boolean;
 
   /**
-   * Keep track of the subTargets for Mouse Events
-   * Sorted in descending z index order, top most object first
+   * Keep track of the subTargets for Mouse Events, ordered bottom up from innermost nested subTarget
    * @type FabricObject[]
    */
   targets: FabricObject[] = [];
@@ -864,15 +863,31 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     pointer: Point
   ): FabricObject | undefined {
     const target = this._searchPossibleTargets(objects, pointer);
-    // if we found something in this.targets, and the group is interactive, return that subTarget
+
+    // if we found something in this.targets, and the group is interactive, return the innermost subTarget
+    // that is still interactive
     // TODO: reverify why interactive. the target should be returned always, but selected only
     // if interactive.
-    return target &&
+    if (
+      target &&
       isCollection(target) &&
       target.interactive &&
       this.targets[0]
-      ? this.targets[0]
-      : target;
+    ) {
+      /** targets[0] is the innermost nested target, but it could be inside non interactive groups and so not a selection target */
+      const targets = this.targets;
+      for (let i = targets.length - 1; i > 0; i--) {
+        const t = targets[i];
+        if (!(isCollection(t) && t.interactive)) {
+          // one of the subtargets was not interactive. that is the last subtarget we can return.
+          // we can't dig more deep;
+          return t;
+        }
+      }
+      return targets[0];
+    }
+
+    return target;
   }
 
   /**
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -189,6 +189,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
 
   /**
    * Keep track of the subTargets for Mouse Events
+   * Sorted in descending z index order, top most object first
    * @type FabricObject[]
    */
   targets: FabricObject[] = [];
@@ -836,7 +837,10 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     while (i--) {
       const target = objects[i];
       if (this._checkTarget(target, pointer)) {
-        if (isCollection(target) && target.subTargetCheck) {
+        if (
+          isCollection(target) &&
+          (target.subTargetCheck || target.interactive)
+        ) {
           const subTarget = this._searchPossibleTargets(
             target._objects as FabricObject[],
             pointer
@@ -1132,8 +1136,9 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
 
     if (isActiveSelection(object) && prevActiveObject !== object) {
       object.set('canvas', this);
-      object.setCoords();
     }
+
+    object.setCoords();
 
     return true;
   }
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -379,21 +379,7 @@ export class StaticCanvas<
    * @param {Array} vpt a Canvas 2D API transform matrix
    */
   setViewportTransform(vpt: TMat2D) {
-    const backgroundObject = this.backgroundImage,
-      overlayObject = this.overlayImage,
-      len = this._objects.length;
-
-    this.viewportTransform = vpt;
-    for (let i = 0; i < len; i++) {
-      const object = this._objects[i];
-      object.group || object.setCoords();
-    }
-    if (backgroundObject) {
-      backgroundObject.setCoords();
-    }
-    if (overlayObject) {
-      overlayObject.setCoords();
-    }
+    this.viewportTransform = [...vpt];
     this.calcViewportBoundaries();
     this.renderOnAddRemove && this.requestRenderAll();
   }
--- a/src/controls/polyControl.ts
+++ b/src/controls/polyControl.ts
@@ -91,8 +91,6 @@ export const factoryPolyActionHandler = (
     poly.left -= diff.x;
     poly.top -= diff.y;
 
-    actionPerformed && poly.setCoords();
-
     return actionPerformed;
   };
 };
--- a/src/controls/polyControl.ts
+++ b/src/controls/polyControl.ts
@@ -91,6 +91,8 @@ export const factoryPolyActionHandler = (
     poly.left -= diff.x;
     poly.top -= diff.y;
 
+    actionPerformed && poly.setCoords();
+
     return actionPerformed;
   };
 };
--- a/src/controls/wrapWithFixedAnchor.ts
+++ b/src/controls/wrapWithFixedAnchor.ts
@@ -21,6 +21,7 @@ export function wrapWithFixedAnchor<T extends Transform>(
       transform.originX,
       transform.originY
     );
+    actionPerformed && target.setCoords();
     return actionPerformed;
   }) as TransformActionHandler<T>;
 }
--- a/src/controls/wrapWithFixedAnchor.ts
+++ b/src/controls/wrapWithFixedAnchor.ts
@@ -21,7 +21,6 @@ export function wrapWithFixedAnchor<T extends Transform>(
       transform.originX,
       transform.originY
     );
-    actionPerformed && target.setCoords();
     return actionPerformed;
   }) as TransformActionHandler<T>;
 }
--- a/src/shapes/ActiveSelection.ts
+++ b/src/shapes/ActiveSelection.ts
@@ -20,6 +20,8 @@ export interface ActiveSelectionOptions extends GroupProps {
 const activeSelectionDefaultValues: Partial<TClassProperties<ActiveSelection>> =
   {
     multiSelectionStacking: 'canvas-stacking',
+    subTargetCheck: true,
+    interactive: false,
   };
 
 /**
@@ -66,13 +68,6 @@ export class ActiveSelection extends Group {
       layoutManager:
         options.layoutManager ?? new ActiveSelectionLayoutManager(),
     });
-  }
-
-  /**
-   * @private
-   */
-  _shouldSetNestedCoords() {
-    return true;
   }
 
   /**
@@ -149,6 +144,7 @@ export class ActiveSelection extends Group {
     // the object will be in the objects array of both the ActiveSelection and the Group
     // but referenced in the group's _activeObjects so that it won't be rendered twice.
     this._enterGroup(object, removeParentTransform);
+    object.setCoords();
   }
 
   /**
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -91,7 +91,6 @@ export class Group
   /**
    * Used to allow targeting of object inside groups.
    * set to true if you want to select an object inside a group.\
-   * **REQUIRES** `subTargetCheck` set to true
    * This will be not removed but slowly replaced with a method setInteractive
    * that will take care of enabling subTargetCheck and necessary object events.
    * There is too much attached to group interactivity to just be evaluated by a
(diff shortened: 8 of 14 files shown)


src/canvas/Canvas.ts
```diff
@@ -1400,17 +1399,11 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
         target._findTargetCorner(this.getViewportPoint(e));
 
     if (!corner) {
-      if ((target as Group).subTargetCheck) {
-        // hoverCursor should come from top-most subTarget,
-        // so we walk the array backwards
```
ShaMan123, line 1405: wrong

src/canvas/SelectableCanvas.ts
```diff
@@ -190,6 +190,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
 
   /**
    * Keep track of the subTargets for Mouse Events
+   * Sorted in descending z index order, top most object first
```
ShaMan123, line 193: right

src/shapes/ActiveSelection.ts
```diff
@@ -101,6 +106,7 @@ export class ActiveSelection extends Group {
     // the object will be in the objects array of both the ActiveSelection and the Group
     // but referenced in the group's _activeObjects so that it won't be rendered twice.
     this._enterGroup(object, removeParentTransform);
+    object.setCoords();
```
ShaMan123, line 109: needs a mounted decision

src/shapes/Object/InteractiveObject.ts
```diff
@@ -511,7 +511,6 @@ export class InteractiveFabricObject<
       ctx.strokeStyle = options.cornerStrokeColor;
     }
     this._setLineDash(ctx, options.cornerDashArray);
-    this.setCoords();
```
ShaMan123, line 514: bad approach, leading to unnecessary computation and a stale state

ShaMan123 (commented):
I also made `Group#interactive` include `subTargetCheck` logically so now it is enough to 
```
new Group(objects, { interactive: true })
``` 

This will not change anything else.


There is a decision to make and then I can complete this PR

.vscode/settings.json
```diff
@@ -12,7 +12,7 @@
   "editor.formatOnSave": true,
   "editor.formatOnPaste": true,
   "editor.codeActionsOnSave": {
-    "source.organizeImports": false,
-    "source.removeUnusedImports": true
+    "source.organizeImports": "never",
+    "source.removeUnusedImports": "explicit"
```
asturur, line 16: i think we should update this file, my vscode started to change it too, but i m not sure if it was mine or is a generic change.

src/brushes/SprayBrush.ts
```diff
@@ -143,6 +144,7 @@ export class SprayBrush extends BaseBrush {
         objectCaching: true,
         subTargetCheck: false,
         interactive: false,
+        layoutManager: new LayoutManager(new FixedLayout()),
```
asturur, line 147: can you explain this change? doesn't seem related to setCoords

src/canvas/Canvas.ts
asturur: What is wrong exactly?

src/canvas/SelectableCanvas.ts
```diff
@@ -845,7 +846,10 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     while (i--) {
       const target = objects[i];
       if (this._checkTarget(target, pointer)) {
-        if (isCollection(target) && target.subTargetCheck) {
+        if (
+          isCollection(target) &&
+          (target.subTargetCheck || target.interactive)
```
asturur, line 851: target interactive requires subtargetcheck, so we shouldn't be checking for both.

src/canvas/StaticCanvas.ts
```diff
@@ -353,21 +353,7 @@ export class StaticCanvas<
...
-    for (let i = 0; i < len; i++) {
-      const object = this._objects[i];
-      object.group || object.setCoords();
-    }
-    if (backgroundObject) {
-      backgroundObject.setCoords();
-    }
-    if (overlayObject) {
-      overlayObject.setCoords();
-    }
```
asturur, line 370: why deleting those? control coords need to be set  when we change the vpt.

src/shapes/Text/Text.ts
```diff
@@ -1692,6 +1691,7 @@ export class FabricText<
     }
     if (this._forceClearCache) {
       this.initDimensions();
+      this.setCoords();
     }
```
asturur, line 1695: this whole thing the text has that can change the size just before rendering is wrong, is probably a patch that was done fore some reason, but this shouldn't exist. Render renders, period.

src/LayoutManager/LayoutManager.ts
```diff
@@ -248,6 +250,8 @@ export class LayoutManager {
       left: object.left + offset.x,
       top: object.top + offset.y,
     });
+    // TODO: should we delete aCoords/oCoords here so they are not stale if subTargetCheck is false?
+    //  or should we apply offset instead of recalculating aCoords?
```
asturur, line 254: an object without coords could create more damages than one with stale, different typings, always have to check if the coords are there.
We should document that if you have a group without subtargetCheck and you plan on using the group inner object coordinates, you have to update them when you use them.

.vscode/settings.json
ShaMan123: generic

src/brushes/SprayBrush.ts
ShaMan123: Not very, true
No need for the spray brush to have a fit content layout

src/canvas/Canvas.ts
ShaMan123: the order of targets is `[topMost, ..., bottomMost]`?

ShaMan123: > Sorted in descending z index order, top most object first

src/canvas/SelectableCanvas.ts
ShaMan123: I thought I wrote about this in the description, sorry, updated now.

> Another change I have introduced is `interactive` including `subTargetCheck`
To able to select sub targets both need to be flagged as true.
`subTargetCheck` flag `searchPossibleTargets` to traverse the children of a group.
If `interactive` is flagged it means we want that to happen so essentially `interactive = true` contains `subTargetCheck = true`. it is more straight forward as well.

src/canvas/StaticCanvas.ts
ShaMan123: We need control coords only for the active object

ShaMan123: And regarding corner coords, we need them on these objects only if they do not ignore the vpt

src/shapes/Text/Text.ts
ShaMan123: suggestions?

src/LayoutManager/LayoutManager.ts
```diff
@@ -216,6 +216,8 @@ export class LayoutManager {
       target.setPositionByOrigin(nextCenter, CENTER, CENTER);
       // invalidate
       target.setCoords();
+      (target.subTargetCheck || target.interactive) &&
+        target.forEachObject((object) => object.setCoords());
```
ShaMan123, line 220: here is a good example of the benefit of separating `setCoords` to `setCornerCoords` and `setControlCoords` because here we need only `setCornerCoords`

ShaMan123: Do you objections doing that?

ShaMan123 (commented):
rebased

src/canvas/StaticCanvas.ts
ShaMan123: Looking at the code we do not even check if these objects are on screen and these objects are not selectable apart from imperatively calling `setActiveObject` (which I believe should call `setCoords` if it doesn't) thus making the `setCoords` call redundant

src/controls/wrapWithFixedAnchor.ts
```diff
@@ -21,6 +21,7 @@ export function wrapWithFixedAnchor<T extends Transform>(
       transform.originX,
       transform.originY
     );
+    actionPerformed && target.setCoords();
```
ShaMan123, line 24: this should belong to `_performTransformAction`

src/shapes/Group.ts
```diff
@@ -479,16 +470,6 @@ export class Group
...
 
-  /**
-   * @override
-   * @return {Boolean}
-   */
-  setCoords() {
-    super.setCoords();
-    this._shouldSetNestedCoords() &&
-      this.forEachObject((object) => object.setCoords());
-  }
```
ShaMan123, line 490: @jiayihu look here, this removes the perf hit you described

src/controls/polyControl.ts
```diff
@@ -91,6 +91,8 @@ export const factoryPolyActionHandler = (
     poly.left -= diff.x;
     poly.top -= diff.y;
 
+    actionPerformed && poly.setCoords();
```
ShaMan123, line 94: once moved to `_performTransformAction` can be removed from here

ShaMan123 (commented):
All in all this looks like an important PR

src/LayoutManager/LayoutManager.ts
ShaMan123: Also this line is incorrect because it goes down 1 level but should recurse down

src/canvas/Canvas.ts
```diff
@@ -1345,17 +1344,11 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-          .concat()
-          .reverse()
-          .map((_target) => {
-            hoverCursor = _target.hoverCursor || hoverCursor;
-          });
-      }
-      this.setCursor(hoverCursor);
+      // hoverCursor should come from top-most subTarget
+      const hoverCursor =
+        (target as Group).subTargetCheck &&
```
ShaMan123, line 1349: should it be:
```suggestion
        (target as Group).interactive &&
```

?

src/canvas/Canvas.ts
```diff
@@ -1304,19 +1304,16 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
+    const actionPerformed =
+      !!actionHandler && actionHandler(e, transform, pointer.x, pointer.y);
+    // This call needs to recurse down
+    actionPerformed && target.setCoords();
+
     if (action === 'drag' && actionPerformed) {
-      transform.target.isMoving = true;
-      this.setCursor(transform.target.moveCursor || this.moveCursor);
+      target.isMoving = true;
+      this.setCursor(target.moveCursor || this.moveCursor);
```
ShaMan123, line 1316: doesn't belong here

ShaMan123 (commented):
cleaned up a small bit

jiayihu:
Chipping in as request to report that current `setCoords` performance is terrible for a few reasons:
1. `setCoords` always set the control coords even though most of the time you don't need to show the controls. In fabric by default actually only the activeObject renders the controls
2. `obj.drawControls` calls `obj.setCoords` internally, so whenever controls are renderered, their position is recalculated. With groups, this is aggravated by point 3
3. Group `setCoords` calls it recursively for the children. Due to point 2, calling `group.drawControls` will call `group.setCoords()`, which in turn calls the children `setCoords`. But then when `child.drawControls` is called, this gets repeated. This results probably in factorial `O(n!)` time complexity.

As just an example, we recently implemented a Table, which is a Group of cells. Each cell is in turn a group of content. Each cell has 4 controls. Re-rendering a table with only 150 cells changes from 40s to 2.5 once setCoords is removed from drawControls:

Screenshot 2024-04-08 at 10 56 03
Screenshot 2024-04-08 at 10 56 58

asturur:
I also want to do some retrospect on how we got here.
SetCords was basically never called and was always left to the developer to don't mess up with it, how we end up that is called at draw control time i do not know.

src/canvas/StaticCanvas.ts
```diff
@@ -379,21 +379,7 @@ export class StaticCanvas<
...
+    this.viewportTransform = [...vpt];
```
asturur, line 382: this wasn't spread before, it shouldn't be spread now

jiayihu:
> never gave a figure of the performance it was improving.

I suggested @ShaMan123 we should add a benchmark test that simply renders a Group with 100 Groups, with 2 children per group. Then measure how much times it takes and how much this PR affects it. Also it will avoid future regressions.

src/canvas/StaticCanvas.ts
ShaMan123: Are you sure?
When we init the canvas we spread to avoid mutation, why not here?

ShaMan123:
@asturur this discussion can lead to how fabric uses `isOnScreen` and should one day. It is very wasteful. We are starting to look into it if you want in.

src/canvas/StaticCanvas.ts
ShaMan123: Exactly, the active object calls setCoords, the rest should not. Mind that `setActiveObject` call setCoords so it seals the gap.

asturur:
> @asturur this discussion can lead to how fabric uses `isOnScreen` and should one day. It is very wasteful. We are starting to look into it if you want in.



> > never gave a figure of the performance it was improving.
> 
> I suggested @ShaMan123 we should add a benchmark test that simply renders a Group with 100 Groups, with 2 children per group. Then measure how much times it takes and how much this PR affects it. Also it will avoid future regressions.
> 
> > SetCords was basically never called and was always left to the developer to don't mess up with it, how we end up that is called at draw control time i do not know.
> 
> Screenshot 2024-04-08 at 17 19 09
> 
> It was there since v5.3 but not on v3.6 (which we were using at my company before migrating to v6) but it's definitely suspicious to call setCoords during drawing. Maybe it was added to workaround stale coords without minding about performance.

Was added in 4.0 when the ability to call render controls with styleoverrides was added.
I think because you could then run multiple controls renders and we didn't want the controls to be offplace.
In that moment probably the same oCoords for interactivity where used for renders ( and maybe now too ) and we thought it was nicer if you didn't have to call that manually.
There weren't nested ocoords tho.
So i m sure that is the reason why that was there, but for sure now it can't be there at all.

asturur:
> @asturur this discussion can lead to how fabric uses `isOnScreen` and should one day. It is very wasteful. We are starting to look into it if you want in.

Just keep in mind isOnScreen was used to skip some rendering for first level objects. That was all, there aren't more intention behind it.
I don't see how is wasteful in theory it should check at worst 5 points in a rectangle i don't see how that is so bad. But please open a discussion on the same example as i m doing for the other changes. Keep it in draft until you want other to ask/comment

jiayihu:
> There weren't nested ocoords tho.

Bear in mind that it's not only about nested oCoords. It still does it for all the canvas objects, which is very expensive. Internally we have changed it to be done only for the needed objects, typically the activeObject.
So I kept the method call in `obj.drawControls`, which is not optimal during rendering of course, for exactly the reason of not having to call it manually. It'd be great if we can find a way to have both, i.e. call it not during rendering but not having to remember it.

asturur:
If we are doing all objects 'setCoords' every render then we introduced a bug. That was never the intention, again setCoords was something that was called on setViewport and at the end of a transform.
At some point it was added during a drag operation but always on a single object.

Anyway this PR won't be breaking, so we can move forward with a release.
My request is to keep it strict to removing setCoords call and removing all the accessory changes ( like overlapping interactive with subtargetcheck and whatever is not a setCoord call )

asturur:
i also see that calcOcoords with the nested cords for groups interactivity became way slower in the case of groups. so that is also probaly a cause of general slowness
We don't have a setCoords for the specific situation and a setCoords generic that updates all.

When doing setCoords on a group that isn't interactive, setting the oCoords of all the nested objects is a waste, and if the group is interactive but selected, is a waste anyway.

But setcoords is imperative, maybe having it going deeply nested was an error because most of the time aCoords do not change and that is all we need for targeting check.

But that is topic for a different discussion/pr

ShaMan123:
> If we are doing all objects 'setCoords' every render then we introduced a bug. That was never the intention, again setCoords was something that was called on setViewport and at the end of a transform.
> At some point it was added during a drag operation but always on a single object.

I was surprised to see that all the above are not as expected. Calling setCoords when drawing controls, not calling them otherwise

asturur:
> > If we are doing all objects 'setCoords' every render then we introduced a bug. That was never the intention, again setCoords was something that was called on setViewport and at the end of a transform.
> > At some point it was added during a drag operation but always on a single object.
> 
> I was surprised to see that all the above are not as expected. Calling setCoords when drawing controls, not calling them otherwise

The gotcha always said, if something is misaligned doesn't work call setCoords.
Exactly because we never called if not when strictly necessary

ShaMan123:
> > > If we are doing all objects 'setCoords' every render then we introduced a bug. That was never the intention, again setCoords was something that was called on setViewport and at the end of a transform.
> > > At some point it was added during a drag operation but always on a single object.
> > 
> > 
> > I was surprised to see that all the above are not as expected. Calling setCoords when drawing controls, not calling them otherwise
> 
> The gotcha always said, if something is misaligned doesn't work call setCoords. Exactly because we never called if not when strictly necessary

That is what the gotcha says but the code does something entirely different so the gotcha is unrelated.
Fabric doesn't call setCoords after a transform action, which is the only place it 100% should.

asturur:
@ShaMan123 is not unrelated, things changed pr after pr probably.
on 5.x setCoords was on finalizeCurrentTransform ( correct ), on scale method ( correct ), on adjustPosition ( correct ), on onComplete of animate ( maybe not correct ) , text editing (  correct ) and group layouting  ( correct ) and on  renderControls ( maybe not correct )

Also today is still on '_finalizeCurrentTransform' so that aspect is still there, not sure why you say it doesn't

src/shapes/Text/Text.ts
asturur: i don't have a suggestion but is also evident that if there is a dependency between having the correct dimensions and being detected on screen, so for correctness this check and setCoords shoud be called before the onScreen early return, and also Object will do the isOnscreen check again.
This render method should take in account that if _forceClearCache is true, should execute the initDimension and then move to super.render. the precheck is not useful, is a double cost 99.9% of the time

asturur:
A red flag i see in this PR is that we remove the nested call of group setCoords done with the subclass and then we comment in 2 places of added code that the call should be recursive, but we just removed the recursiveness from the code.
I don't think that change is a good one.

The only changes that are 100% correct here is the one from viewport and then moving the setCoord from the drawControl call earlier to the performAction could be correct without much doubts too.
But at that point you should remove the one from 'finalizeCurrentTransform'.

Where are the performance improvement coming from?
drawControl is removed but performAction compensate it
The only thing i see is removing the nesting from the group setCoords, but also then realizing is sitll necessary in many cases

ShaMan123:
> drawControl is removed but performAction compensate it

Only during a transform, if for other reasons you need to render the canvas this is better.

> But at that point you should remove the one from 'finalizeCurrentTransform'.

agreed



Regarding the recursive setCoords call of group. Do we need it?
For geometry we need it.
For isOnScreen we do not because it doesn't check nested objects.
For controls we do not because the setActiveObject will handle that. 
Thoughts?

ShaMan123:
> Regarding the recursive setCoords call of group. Do we need it?
> For geometry we need it.


This makes my suggestion to delete aCoords/oCoords relevant. If they do not exist they could be recalculated on the fly from `getCoords` etc.

jiayihu:
> If they do not exist they could be recalculated on the fly from getCoords

Then you'd get worse performance having to recompute them everytime, for instance for `_checkTarget->_pointIsInObjectSelectionArea`

ShaMan123:
Why? We delete them only if an ancestors invalidates geometry

asturur:
my suggestion is to take the no brainers from this PR and to think better what to do for the rest.

Do we need to have oCoords cached? at this point i do not know.
You need to have them cached only if you are running findTargetControl on non selected objects.
Fabric by default doesn't change the controls anymore if the object is not already selected, and if you select it you can call setCoords once and then keep updated during transforms.

But of course if you know you are using an override where you can click a control also when the object is not selected, then things change.

ShaMan123:
> Do we need to have oCoords cached? at this point i do not know.

Great point!
Ok then, so the remaining question is about the nested setCoords. I think it is an important perf hit we should resolve so I shall reiterate on it after the non brainers.
I have an idea for hashing both the matrices and the coords.

ShaMan123:
closing in favor of #9793

PR #9561, opened 2023-12-29
revert(): rm active selection ref

ShaMan123:
## Description

This PR reverts the active selection ref on the canvas.
It was a bad decision that opened odd bugs and bad design
Voids the need for #9525


ShaMan123:
@asturur please review this so we can decide on the right way forward

src/canvas/Canvas.ts
```diff
@@ -8,6 +8,7 @@ import type {
   Transform,
 } from '../EventTypeDefs';
 import { Point } from '../Point';
+import { ActiveSelection } from '../shapes/ActiveSelection';
```
asturur, line 11: i want to avoid the library to be a monoblock unsplittable, so please we need to do a function that identify this being an active selection without importing it.

src/canvas/Canvas.ts
```diff
@@ -1470,32 +1469,34 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
           this._hoveredTargets = [...this.targets];
         }
         this._fireSelectionEvents(prevActiveObjects, e);
       } else {
         (activeObject as IText).exitEditing &&
           (activeObject as IText).exitEditing();
         // add the active object and the target to the active selection and set it as the active object
-        activeSelection.multiSelectAdd(activeObject, target);
-        this._hoveredTarget = activeSelection;
+        const newActiveSelection = new ActiveSelection();
```
asturur, line 1493: this bothers me. How can we avoid importing the active selection on the canvas?

asturur: i would use the class registry if it wasn't that then you have to import the activeSelection in order to trigger the registration.

asturur: No i think that the classRegistry is the correct thing to do here:
- avoid forced imports
- allow for eventual subclassing
I was wrong in saying that you have to import it, the fact that we actually exports it will grant its presence in the classRegistry

asturur:
i am ok reverting this feature, i was neutral about it, if the code in this way is simpler i prefer to roll it back

asturur:
I think we should use the classRegistry to load the ActiveSelection class and we  should avoid to use instanceOf for it.
I can do those changes if you are in rush of merging and you don't have time today, otherwise you can do them

src/canvas/Canvas.ts
ShaMan123: > I was wrong in saying that you have to import it, the fact that we actually exports it will grant its presence in the classRegistry

This is not precise. We want tree shaking and therefore we want the classRegistry to not account as a side effect so we do want an explicit import to trigger the ActiveSelection import

I agree that we should use the class registry to allow subclassing

ShaMan123:
> I think we should use the classRegistry to load the ActiveSelection class and we should avoid to use instanceOf for it. I can do those changes if you are in rush of merging and you don't have time today, otherwise you can do them

I will do it now.
What else apart from `instanceof`? Back to `isActiveSelection`?

src/canvas/Canvas.ts
asturur: i meant we don't want to break the splittability, and with current setup ActiveSelection will work anyway

at least we can do a script
for custom builds and the active selection would just work.

asturur:
> > I think we should use the classRegistry to load the ActiveSelection class and we should avoid to use instanceOf for it. I can do those changes if you are in rush of merging and you don't have time today, otherwise you can do them
> 
> I will do it now. What else apart from `instanceof`? Back to `isActiveSelection`?

yes something that can detect withou using the class.

if we didn't have selections of objects inside a group we could just say that the activeObject is not in canvas, that would grant is an active selection, but now we have to flag it maybe recognizing the stac property we added recently or something else

ShaMan123 (commented):
All tests should pass now
Remaining is to decide about how to handle `instanceof ActiveSelection`
Since there is no issue with import cycles I don't mind leaving it as is.
I could re-expose `isActiveSelection` using `classRegistry`  
```
const isActiveSelection = (object: FabricObject) =>
  object instanceof
  classRegistry.getClass<typeof ActiveSelection>('ActiveSelection');
```

Is that better?

asturur:
> All tests should pass now Remaining is to decide about how to handle `instanceof ActiveSelection` Since there is no issue with import cycles I don't mind leaving it as is. I could re-expose `isActiveSelection` using `classRegistry`
> 
> ```
> const isActiveSelection = (object: FabricObject) =>
>   object instanceof
>   classRegistry.getClass<typeof ActiveSelection>('ActiveSelection');
> ```
> 
> Is that better?

That would be fine imho, but is slightly different from the others, but does the job.
Wouldn't be applicable to the other functions we have there, but for the specific case of activeSelection i think is fine.

asturur:
not sure typing getClass in that way is a good move, but we can still default to any as it was before and patch where it creates issues

ShaMan123:
> not sure typing getClass in that way is a good move, but we can still default to any as it was before and patch where it creates issues

Can you elaborate?
I don't really like passing `typeof Class` and prefer passing `Class` but TS can't easily support that
Also I thought of adding a default any

ShaMan123:
Mu thoughts about `isActiveSelection` still stand, please consider.
It is used only by Canvas/SelectableCanvas. These classes depend on ActiveSelection. So there is no reason not to import ActiveSelection and use instanceof, de facto no import cycles were created.
I think `isActiveSelection` is messy and weird and opens up strange issues with the class registry and it means the ActiveSelection was imported anyways so why not just do it without this workaround?
Anyways I will expose the util, it is easy to expel it later on

ShaMan123 (commented):
Looks ready

ShaMan123:
jest just proved that.
waiting for your review on this

> My thoughts about `isActiveSelection` still stand, please consider. It is used only by Canvas/SelectableCanvas. These classes depend on ActiveSelection. So there is no reason not to import ActiveSelection and use instanceof, de facto no import cycles were created. I think `isActiveSelection` is messy and weird and opens up strange issues with the class registry and it means the ActiveSelection was imported anyways so why not just do it without this workaround? Anyways I will expose the util, it is easy to expel it later on

asturur:
> > not sure typing getClass in that way is a good move, but we can still default to any as it was before and patch where it creates issues
> 
> Can you elaborate? I don't really like passing `typeof Class` and prefer passing `Class` but TS can't easily support that Also I thought of adding a default any

> My thoughts about `isActiveSelection` still stand, please consider. It is used only by Canvas/SelectableCanvas. These classes depend on ActiveSelection. So there is no reason not to import ActiveSelection and use instanceof, de facto no import cycles were created. I think `isActiveSelection` is messy and weird and opens up strange issues with the class registry and it means the ActiveSelection was imported anyways so why not just do it without this workaround? Anyways I will expose the util, it is easy to expel it later on

Doing isActiveSelection or doing  instanceOf ActiveSelection is the same exact thing, at the net of the forced import.
Canvas do not depend on activeSelection if you don't plan to use it and disable the multi selection entirely and anyway being splittable is a target goal.
So if we have to check for a switch in a piece of code is better we do it in a way it doesn't bind the code together.

src/util/misc/objectEnlive.ts
```diff
@@ -88,13 +94,14 @@ export const enlivenObjects = <
     Promise.all(
       objects.map((obj) =>
         classRegistry
-          .getClass(obj.type)
-          .fromObject(obj, {
-            signal,
-            reviver,
```
asturur, line 94: ok i wonder how reviver ended up here in the refactors

src/canvas/SelectableCanvas.ts
```diff
@@ -1296,11 +1272,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
   _realizeGroupTransformOnObject(
     instance: FabricObject
   ): Partial<typeof instance> {
-    if (
-      instance.group &&
-      instance.group === this._activeSelection &&
-      this._activeObject === instance.group
-    ) {
+    if (instance.group && isActiveSelection(instance.group)) {
```
asturur, line 1275: this check is different now, we still have to check that instance.group is activeObject
Being an activeSelection, doesn't mean you are the active one.

src/util/misc/objectEnlive.ts
ShaMan123: just because of the type change

asturur: yes but it never was part of fromObject, why did we put it there

asturur:
using the classRegistry check with Jest is broken because jest imports only what it needs, so i had to revert it to checking a random exclusive property

src/util/misc/objectEnlive.ts
ShaMan123: a good question for the blame and spare time to answer

asturur:
i m good to go

asturur (approved):

ShaMan123:
we should follow up with removing the code in the layout manager `shouldResetTransform`

asturur:
be mindful that this is not a revert.
Revert are safe by default they remove a PR as it was, usually from git revert command.
This should have been called refactor.

PR #9563, opened 2023-12-29
Issue 9475 bad divide breaks scaling in polygon

### LINKED ISSUE #9475: [Bug]: The bounding box disappear if scaling Y for a Rectangle whose strokeWidth = 0 and initial height = 0

gloriousjob:
## Motivation

@closes #9475

## Description

0 height polygons have issues where they disappear when you scale them.

## Changes

Adds guards to ensure divide by 0's do not happen to create these problems. This is done by preventing the scale when value is 0.


gloriousjob:
Original PR here: https://github.com/fabricjs/fabric.js/pull/9496
I renamed the branch since I'm isolating to the issue number that the branch was not named to.

src/controls/scale.ts
```diff
@@ -70,6 +70,13 @@ export function scalingIsForbidden(
   if (lockY && by === 'y') {
     return true;
   }
+  // object disappears if scaling 0 sized object so forbid
+  if (fabricObject.width === 0 && by == 'x') {
+    return true;
+  }
+  if (fabricObject.height === 0 && by === 'y') {
+    return true;
+  }
```
ShaMan123, line 79:
```suggestion
  if (fabricObject.width === 0 && by !== 'y') {
    return true;
  }
  if (fabricObject.height === 0 && by !== 'x') {
    return true;
  }
```

ShaMan123 (commented):
what do you think? What about the scaling from the corner case?
But still I am not sure this is good enough
I think we should expect calculations to be safe in all cases.
However, that can be done in the future and doesn't have to be part of this PR

asturur:
i m ok with this, regarding the best way to do this, this was an issue we also had with line, and somehow we solved it by hiding the controls.
I do not have strong opinions about calculations being safe all the time, important is to not break.

src/controls/scale.ts
gloriousjob: I tried this with a 0 height rectangle. It turned out that the bottom right corner is what was showing on the right side, so I couldn't extend the rectangle in the x direction. That said, I found out the diagonal was working because it used uniform scaling and the uniform scaling calculation uses a different calculation based on adding the scales and doesn't break as a result. But if the uniform scaling was off, the diagonal ends up with the issue (i.e., NaN and the polygon disappears).

So this is a good thought since non-uniform scaling will break without taking the 'both' case into consideration. But it almost begs for the diagonals to be hidden. Is there an easy way to do this (or would that be a bad idea since it could add confusion on how corners show)?

ShaMan123: Not sure I understand

gloriousjob: Basically, because it’s a line, the diagonal scalers stack on top of the edge scaler and making the both case forbidden means you can’t extend the line.

To make things confusing, when uniform scaling is on and both is allowed, the diagonals work but when uniform scaling is off, this diagonals are broken.

asturur: i m pretty much sure there was code for the simple Line object that was hiding the corners when one dimension was 0.

asturur: Regarding @ShaMan123 comment let's take the most restrictive fix, so if one case of diagonals would work, we don't care we lock both.
As a side effect a rectangle created with zero width won't even stretch, it won't error, but in general developer ahve to be driven away from such cases.

ShaMan123: we could even argue that during the render cycle an object with width/height of 0 is not on screen and should therefore be skipped and not rendered at all

ShaMan123: seems I have found the source of these bugs


Screenshot 2024-01-06 at 15 07 26

Should be `!width || !height`

ShaMan123:
```
  isNotVisible() {
    return !this.opacity || !this.width || !this.height || !this.visible;
  }

```

ShaMan123: IMO it is weird the an object with height of 0 and strokeWidth >0 draws the stroke

asturur: that only skips rendering, it doesn't makes it unselectable. If you have padding the object is still selectable.

asturur: > seems I have found the source of these bugs
> 
> Screenshot 2024-01-06 at 15 07 26
> Should be `!width || !height`

This is unrelated.
the boolean visible is for the developer, and determines rendering and selectability.
The 'isNotVisible' is an internal check to skip rendering code when will produce no output. It has nothing to do with our calculation blowing away because of a zero division.

asturur:
I think the test is a bit over complicated i think we can slim it down a bit

src/controls/scale.ts
gloriousjob, line 1: @asturur What would you think of this change in the place of old lines 177/178? Basically it just verifies the non-zero value and defaults the scale to 1 if it is 0:
      scaleX = dim.x ? Math.abs((newPoint.x * target.scaleX) / dim.x) : 1;
      scaleY = dim.y ? Math.abs((newPoint.y * target.scaleY) / dim.y) : 1;

This seems closer to the bugfix's original idea and might be a better fix than blocking the scale considering some of the concerns brought up.

asturur: It doesn't make sense to go from 0 to 2 with scaling, especially in a world in which you can just slap on a rectangle a change height control.
FabricJS is not a piece of JS to build a foolproof ux for editors covering all the possible programmers mistakes.
It supports interactivity but you have to know what are you doing.
Another example is string in place to numbers, i do not think we should check that a value is a string and in case parse an integer from it because some developers do not remember that any input form return a string even if is of a numeric kind.

setting scale to 1 is a bit dangerous, it would break the uniform scaling.

Imagine you have a poling scaled to 4x on both sizes.
then you move the points in a line, you obtain size 0 on height, and then you try to scale, you reset scale to 1.
then you go back to editing points and you move back a point, now your height is again different from 0 but your polygon is now scaled with a non uniform value.

Again this is a very weird case, we have just to avoid to fillup the console in red, so imho blocking it is good.
We do not have to trigger the forbid sign, we can also just return false from the action so scaling won't happen at all.

The part of showing the disallow cursor is optional if we think it makes more sense to warn the dev/user that something is being blocked rather than thinking their left mouse button is being defective

asturur:
> I think we can use the logic concept @asturur implemented in #9575 to fix this for good If we calculate the ratio between the 2 vectors that define the scale axis it should be safer/easier We have the anchor point and the pointer. Calculate the ref vector: `A = controlPosition - anchorPoint` Calculate the current vector: `A' = pointer - anchorPoint` Calculate the ratio: `ratio = magnitude(A) > 0 ? A' / A : 0` Calculate new scale: `scale *= ratio` We should use a dot product calculation to obtain the magnitude with respect to the direction so it supports flipping

but any scale you give to a zero size item will still give you zero.
the fix we implemented here imho is just fine, 0 scaling doesn't exist.
Just i would like to slim down the test that is too confusing

asturur:
Merging this next, but i will try to make the test as simple as possible.

asturur:
At the end i extended the test because i couldn't simplify it.

asturur (approved):

PR #9564, opened 2024-01-01
feat() LayoutManager save/restore from object

asturur:
## Description

Adding save and restore to group with a layout manager that isn't the default one.

### TODO
- [x] would be nice to reduce toObject output at least for the default value case
- [x] add new tests in jest that cover all the cases
- [ ] ~add documentation on what to expect from every single strategy~ i do not have a format for that right now

### DIFF
--- a/src/shapes/Group.spec.ts
+++ b/src/shapes/Group.spec.ts
@@ -1,4 +1,4 @@
-import { FixedLayout, LayoutManager } from '../LayoutManager';
+import { FixedLayout, LayoutManager, ClipPathLayout } from '../LayoutManager';
 import { Canvas } from '../canvas/Canvas';
 import { Group } from './Group';
 import type { GroupProps } from './Group';
@@ -151,7 +151,6 @@ describe('Group', () => {
         height: 50,
         layoutManager: new LayoutManager(new FixedLayout()),
       });
-      group.layoutManager = new LayoutManager(new FixedLayout());
       const serialized = group.toObject();
       expect(serialized.layoutManager).toMatchObject({
         type: 'layoutManager',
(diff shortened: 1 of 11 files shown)


src/shapes/Group.spec.ts
```diff
@@ -115,6 +138,125 @@ describe('Group', () => {
...
+      const { group } = makeGenericGroup({
+        top: 20,
+        left: 40,
+        clipPath: new Rect({ width: 30, height: 10 }),
+        layoutManager: new LayoutManager(new ClipPathLayout()),
+      });
+      expect(group.top).toBe(20);
+      expect(group.left).toBe(40);
+      // TO DO BUG: this should be 30
+      expect(group.width).toBe(31);
```
ShaMan123, line 250: why?

ShaMan123 (approved):

src/shapes/Group.spec.ts
asturur: clippaths do not render border by our spec ( and svg spec ) so if you have a clipPath object with a huge border, that won't do anything and will look like extra padding in the group around the are a you wanted to clip to.

Look here when we render clipPath:
```ts
  /**
   * @param {boolean} forClipping apply clipping styles
   */
  drawObject(ctx: CanvasRenderingContext2D, forClipping?: boolean) {
    const originalFill = this.fill,
      originalStroke = this.stroke;
    if (forClipping) {
      this.fill = 'black';
      this.stroke = '';
      this._setClippingProperties(ctx);
    } else {
      this._renderBackground(ctx);
    }
    this._render(ctx);
    this._drawClipPath(ctx, this.clipPath);
    this.fill = originalFill;
    this.stroke = originalStroke;
  }
```

ShaMan123: I didn't know that

asturur: i can't find the written spec w3g anymore https://www.w3.org/TR/2000/WD-SVG-20000629/masking.html#ClippingPaths but i m sure we can do an SVG and verify.

PR #9569, opened 2024-01-06
fix(FabricObject): Remove unwanted set type warning

asturur:
## Description

Warning when trying to write the type property is useful only to warn the developer intentionally messing with it.
restoring fromObject wasn't fully guarded from triggering it, now all the standard cases should be covered.

### DIFF
--- a/src/Pattern/types.ts
+++ b/src/Pattern/types.ts
@@ -13,5 +13,6 @@ type ExportedKeys =
 export type PatternOptions = Partial<Pick<Pattern, ExportedKeys>>;
 
 export type SerializedPatternOptions = PatternOptions & {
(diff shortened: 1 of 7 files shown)


ShaMan123 (approved):

src/Pattern/types.ts
```diff
@@ -13,5 +13,6 @@ type ExportedKeys =
 export type PatternOptions = Partial<Pick<Pattern, ExportedKeys>>;
 
 export type SerializedPatternOptions = PatternOptions & {
+  type: 'pattern';
```
asturur, line 16: i m not sure why on toObject we still emit 'pattern' instead of 'Pattern', but that's it.

PR #9570, opened 2024-01-06
refactor(): Rewrite how typeAssertion works to avoid isType and add tests for subclasses

asturur:
## Description

Rewrite typeAssertions so they do not need to use isType.
The reason for this was the discussion about isType not being safe in the case of subclasses and also the general will of moving away from isType.

IsType hasn't been removed yet

replace https://github.com/fabricjs/fabric.js/pull/8952


src/util/typeAssertions.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+      expect(isTextObject(text)).toBe(true);
+    });
+    test('can detect Textbox', () => {
+      const text = new Textbox('hello');
+      expect(isTextObject(text)).toBe(true);
+    });
+    test('can detect other subclasses', () => {
+      class NewText extends Textbox {
+        status = 'new';
+      }
```
ShaMan123, line 33: to make it more complete you should add a 
```
static type = 'NextText'
```

src/util/typeAssertions.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+  });
+  describe('isPath', () => {
+    test('can detect Path', () => {
+      const path = new Path('M 0 0 L 1 1');
+      expect(isPath(path)).toBe(true);
+    });
+    test('can detect other subclasses', () => {
+      class NewPath extends Path {
+        status = 'new';
+      }
```
ShaMan123, line 50: same

src/util/typeAssertions.ts
ShaMan123, line 1: As I mentioned in the past I dislike this pattern

ShaMan123 (approved):

src/util/typeAssertions.spec.ts
asturur: oh rigth

src/util/typeAssertions.ts
asturur: What is the pattern here? what pattern are you disliking?

ShaMan123: Type assertions: the need to run code in the global/canvas/not specific class scope of class specific logic.
It is bad design IMO that we are gradually and hopefully moving away from.
That is why I prefer to refactor logic and remove all of this instead of providing a different way to check `isTextObject`.
I think that if we use this pattern it means the class that needs it should expose a method on its base class and override it to allow a robust API

asturur: Yes but you have to learn to divide things.
There are 2 issues here. isType will break on subclasses with different names. This PR is addressing this and is easy to address.
Dividing the code that check specific classes in global spaces to newer code in classes is far more involved that doing this fix, and also is far more risky to do the refactor with the scope of removing isTextObject, rathern than sitting down and replanning the code with the scope of replanning the code.

This PR is about isType and it doesn't condone or support the use of object type checkings in canvas or in the generic object class, is just a different piece of work.

ShaMan123: For future PRs it would be good if you wrote in the description what you just described.
Otherwise I can't tell your intentions

ShaMan123: I mean, saying something like: This coding pattern is not good enough and we plan to move away from it but for now we will use it as is

PR #9572, opened 2024-01-06
fix(): `exactBoundingBox` stroke calculations

ShaMan123:
mark method

## Description

cherry picked out of #9537 
`exactBoundingBox` should IMO become the standard of object geometry API, meaning, an object should expose it final width and height values including stroke and whatever. Fabric should not , therefore, attempt to perform any other calculation besides sending the size vector to the relevant plane.
This will make geometry easier, straight forward, same for subclassing and will confine the logic to a single place instead of having a number of placing messing around with stroke calculations, all of them not compatible with stroke uniform etc.

I do not think this PR is good. It is an ugly patch.
A good solution would be to stick to what I have outlined above and remove all stroke calculations from fabric apart from the method that determines the size of an object on the object prototype

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [next]
 
 - fix() Addressing path cloning slowness ( partially ) [#9573](https://github.com/fabricjs/fabric.js/pull/9573)
+- fix(): `exactBoundingBox` stroke calculations [#9572](https://github.com/fabricjs/fabric.js/pull/9572)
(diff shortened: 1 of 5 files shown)


asturur:
We talked about the next step after this fix will be to introduce a new Polyline and Polygon object that work in that way. We deprecate the exactBoudningBox from the standard one and ( maybe we should deprecate right away ), and if the new one is successfull and easier to manage we can migrate other shapes to that logic.

asturur (changes requested):
Please add deprecation notice on the exactBoundingBox boolean

ShaMan123 (commented):
done

asturur (approved):

PR #9573, opened 2024-01-06
fix() Addressing path cloning slowness ( partially )

asturur:
## Description

While working on a fun pet project with fabricJS i noticed that cloning many groups slowed down the app very much.
The flameGraph showed clearly i was running many times one cloneDeep, so i decided to look into it.

The path i added in this benchmark is the large i have and i m not using yet in the project, while the normal one is a real one.

It turns out that the JSON.stringify/parse approach for cloning is very unefficient for things you can easily customize for specific loop.

Those are the results i got:

```
{ pathCloneDeep: 650, pathOldCode: 563 }
{ cloningSimple: 770, cloningCustom: 25 }
{ cloningSimpleNormal: 141, cloningCustomNormal: 6 }
```

I can read as follow:
Creating and cloning a single large path is extremely slow in any case.
This process include 2 time parsing the path, 2 bounding box calculation, and the data structure clone.
But the only change in serialization of the large path account for around 80+ ms, that seems a LOT.

If i repeat the test for the cloning part only, on 100 iterations the path data cloning seems 30 times faster anyway,

If i take a simpler path, the difference is in the 20 times region.

So i m proposing to just bring back the small old code.

I know this is a synthetic benchmark, but when working with the app i did got slow, and the flamegraph pointed me exactly here, so is not that much of a stretch, is tangible.


asturur:
Added note: the result of this change in my particular case of the app may be exacerbated or the improvements can be less because of the frequent GC call i was getting during the clone process. if JSON.stringify was trashing more memory i will have less GC calls and so more improvements or the same number of GC calls and so less improvements. ( i ll try to give it a stab )

src/benchmarks/pathCloning.mjs
ShaMan123, line 1: can you extract all the data to a standalone file, maybe even a json so it is easier to handle?

ShaMan123 (dismissed):
So doesn't this mean we should introduce a better cloner in general?
I wanted to use [lodash.cloneDeep](https://www.npmjs.com/package/lodash.clonedeep) back when we touched it.
Are you against? And if so why? 
Screenshot 2024-01-06 at 12 58 29

src/benchmarks/pathCloning.mjs
ShaMan123: even scrolling throught it while reviewing is not very friendly

asturur:
If i try it in the real test scenario of an app what i get is this, around 10% reduction ( on an average of 3 repetitions for each case ). This change also include the console not spamming warning, i have no idea how much that influences on the JS execution time
Current fabricJS
image
This branch
image

src/benchmarks/pathCloning.mjs
asturur: what data?

ShaMan123: all the path data that can be extracted to a json

ShaMan123:
I would ad a test from cloning objects and then introduce lodash to see what is faster

asturur:
> So doesn't this mean we should introduce a better cloner in general? I wanted to use [lodash.cloneDeep](https://www.npmjs.com/package/lodash.clonedeep) back when we touched it. Are you against? And if so why? Screenshot 2024-01-06 at 12 58 29

I don't to add dependencies and then worry for patches. I would like to stay dependency free for browser.
We have 2 cloneDeep cases remaining, one is points and should probably be handled like this, the other is styles and i have no idea how ugly it gets.

ShaMan123:
I see
`stylesToArray` doesn't really need that and can be done in a better way.
Maybe the same applies to `_fromObject`
In `setDragImage` it can be dropped in favor of a spread IMO
And the polyline points as you mentioned

asturur:
i added a case for lodash to rule out lazyness in exploring, but is slower anyway
i tested both the 4.5 single package and the version we have in our deps because of tests scripts.

```
{ pathCloneDeep: 641, pathOldCode: 492 }
{ cloningLodash: 366, cloningSimple: 781, cloningCustom: 24 }
{
  cloningLodashNormal: 54,
  cloningSimpleNormal: 142,
  cloningCustomNormal: 6
}
```

src/benchmarks/pathCloning.mjs
asturur: i made a mjs file with 2 consts. because i understood what you meant later but i didn't read the comment

ShaMan123:
what about the rest?

jiayihu:
> This change also include the console not spamming warning, i have no idea how much that influences on the JS execution time

That can affect the performance significantly depending on the browser because writing on the console is an I/O operation basically. So it depends if the browser accounts for that time and removes it. It's better to run benchmarks without console logs.

Regarding deep cloning, there is `structuredClone` now natively: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

ShaMan123:
> Regarding deep cloning, there is `structuredClone` now natively: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

Very useful! We should use that

asturur:
image

is out of range for now, requires polyfills.
We can benchmark to get an idea for the future

asturur:
> what about the rest?

Sorry didn't see this.
At the first occasion we can update.
This was a specific issue that was hitting me hard and was a quick fix. But i need to prioritize the release of 6.0 this is an improvement that can go out anytime. Anyone can open a PR for it. One per points, one per text style

PR #9575, opened 2024-01-06
fix(): fix modify polygon points with zero sized polygons ( particular case of axis oriented lines )

### LINKED ISSUE #9556: [Bug]: createPolyControls with two points (Line) has locked axis
### CheckList

### Version

6.0.0-beta16

Chrome

None

### Link To Reproduction

https://codesandbox.io/p/devbox/fabric-vanillajs-sandbox-forked-nfc5j4

### Steps To Reproduce

Try to move control points of red and green polygons (Lines)

### Expected Behavior

These points should not be fixed on the axis when approaching 0 values

### Actual Behavior

https://github.com/fabricjs/fabric.js/assets/14239290/abde23bf-4701-40af-860c-979d37122de3

_No response_

asturur:
## Description

Since the polygon points are by design ignoring padding, we do not need to use the standard getLocalPoint and then deal with calculations that involve width/height directly.

We can use a plain matrix that should take care of placing the new point correctly.

This PR fixes the bug mentioned in #9556 and also fixes the displacement of the polygon when skew is involved

closes #9556


src/controls/polyControl.ts
```diff
@@ -50,20 +46,15 @@ export const polyActionHandler = (
...
-  const finalPointPosition = mouseLocalPosition
-    .multiply(adjustFlip)
-    .multiply(sizeFactor)
-    .add(poly.pathOffset);
+  const { target, pointIndex } = transform;
+  const poly = target as Polyline;
+  const mouseLocalPosition = sendPointToPlane(
+    new Point(x, y),
+    undefined,
+    poly.calcTransformMatrix()
```
asturur, line 54: @ShaMan123  i do remember that x,y are always in the scene coordinates, but i m not sure that this is true also for when we run an action inside an interactive group. Do you remember if that was changed to accomodate interactive groups?

If those are in parent coordinates i have to change this with calcOwnTransform()

ShaMan123: I am very surprised to see x,y in the parent plane and not in the scene plane
bad decision I made at that point to limit the amount of changes

Screenshot 2024-01-06 at 20 04 17

asturur: yeah we talked about this, but i wanted to be sure. So i ll use calcOwnMatrix

src/controls/polyControl.ts
```diff
@@ -90,19 +81,15 @@ export const factoryPolyActionHandler = (
...
+      .transform(poly.calcOwnMatrix());
 
-    poly.setPositionByOrigin(
-      anchorPointInParentPlane,
-      newPositionNormalized.x + 0.5,
-      newPositionNormalized.y + 0.5
-    );
+    const diff = newAnchorPointInParentPlane.subtract(anchorPointInParentPlane);
+    poly.left -= diff.x;
+    poly.top -= diff.y;
```
asturur, line 92: @ShaMan123 this approach with diffing the position of the anchor point in a reference plance, before and after the action works. Now also skew works fine.

ShaMan123: nice!

ShaMan123 (approved):
This looks great
I would do a small syntax change in `createPolyPositionHandler` to match the rest
Screenshot 2024-01-07 at 8 08 32

asturur:
I made the test case larger and included all variation of combination of settings of polygons ( 4 ) all with skewX/skewY/angle/scale and it works fine now.

ShaMan123:
I agree but we should stick to playwright and use what you did in the vabilla app

PR #9577, opened 2024-01-07
feat(Next.js sandbox): simpler canvas hook

ShaMan123:
## Description

After discussions about async disposing raised some concerns and after @asturur showed me there should not be any,
 I decided to update the next.js app that uses react18 to prove there is no problem using fabric and async disposing.
I decided to make the hook a very simple one so devs could use it and we could use it as a reference point.

I intentionally added an infinite render loop using `animate` to see how async disposing plays with react.
I discovered that the initialization callback `onLoad` has to be a dependency of the canvas main hook (`useEffect`) to cover the case of changing the `onLoad` hook in dev mode so that the canvas is disposed and recreated.


ShaMan123 (commented):
This is good and covers all the active discussions about async disposing.
I would like to create a test for this using the next app but not now since we have CI PRs that make that possible in the pipeline

asturur:
What i will do at some point, is an initialization method for the canvas in which you pass a div and 2 canvases so that you don't have to dispose ever, but you can have react change the items under the hood.
As soon as i have a free 2 hours i ll do that too.

asturur:
@ShaMan123 in general the angular standard commit types are:

```
build|ci|docs|feat|fix|perf|refactor|test
```
patch is generic and i avoid using it on purpose

asturur (approved):

ShaMan123:
> What i will do at some point, is an initialization method for the canvas in which you pass a div and 2 canvases so that you don't have to dispose ever, but you can have react change the items under the hood.
> As soon as i have a free 2 hours i ll do that too.

I was thinking the same
I would argue there is no need to pass the canvas refs at all. Only a div.
Think of the number of canvases being configurable in the future

ShaMan123:
Thinking of it we actually allow now to pass nothing to the Canvas constructor so we are really close

PR #9581, opened 2024-01-09
BREAKING BETA(LayoutManager): rm `shouldResetTransform`

ShaMan123:
## Description

Thanks to #9561 it is now possible to remove the badly designed `shouldResetTransform` hook from the layout manager, me to blame.
Thanks for the patience!

### DIFF
--- a/src/LayoutManager/LayoutManager.spec.ts
+++ b/src/LayoutManager/LayoutManager.spec.ts
@@ -305,19 +305,27 @@ describe('Layout Manager', () => {
         relativeCorrection: new Point(-30, -40),
       });
 
-      const target = new Group([], { scaleX: 2, scaleY: 0.5, angle: 30 });
+      const rect = new FabricObject({ width: 50, height: 50 });
+      const target = new Group([rect], { scaleX: 2, scaleY: 0.5, angle: 30 });
 
       const context: StrictLayoutContext = {
         bubbles: true,
         strategy: manager.strategy,
         target,
+        targets: [rect],
         ...options,
         stopPropagation() {
           this.bubbles = false;
         },
       };
 
-      expect(manager['getLayoutResult'](context)).toMatchSnapshot();
+      expect(manager['getLayoutResult'](context)).toMatchSnapshot({
+        cloneDeepWith: (value: any) => {
+          if (value instanceof Point) {
+            return new Point(Math.round(value.x), Math.round(value.y));
+          }
+        },
+      });
     });
   });
 
--- a/src/LayoutManager/LayoutManager.spec.ts
+++ b/src/LayoutManager/LayoutManager.spec.ts
@@ -475,12 +475,6 @@ describe('Layout Manager', () => {
         const canvasFire = jest.fn();
         target.canvas = { fire: canvasFire };
 
-        const shouldResetTransform = jest
-          .spyOn(manager.strategy, 'shouldResetTransform')
-          .mockImplementation(() => {
-            lifecycle.push(shouldResetTransform);
-          });
-
         const context: StrictLayoutContext = {
           bubbles,
           strategy: manager.strategy,
@@ -501,11 +495,9 @@ describe('Layout Manager', () => {
         manager['onAfterLayout'](context, layoutResult);
 
         expect(lifecycle).toEqual([
-          shouldResetTransform,
           targetFire,
           ...(bubbles ? [parentPerformLayout] : []),
         ]);
-        expect(shouldResetTransform).toBeCalledWith(context);
         expect(targetFire).toBeCalledWith('layout:after', {
           context,
           result: layoutResult,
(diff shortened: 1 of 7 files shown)


src/shapes/ActiveSelection.ts
```diff
@@ -146,12 +147,11 @@ export class ActiveSelection extends Group {
...
 
   /**
-   * If returns true, deselection is cancelled.
-   * @since 2.0.0
-   * @return {Boolean} [cancel]
+   * @override remove all objects and reset transform
    */
   onDeselect() {
     this.removeAll();
+    resetObjectTransform(this);
```
ShaMan123, line 154: do we want this?

ShaMan123 (commented):
cleanup is done
now decisions

ShaMan123:
I have noticed that we introduced a regression with polys flagged with exactBoundingBox

ShaMan123 (commented):
removed resetting transform, ready to go

ShaMan123:
We can bind, sure. You want it in this PR?

> My understanding is that it takes care of the the size, of the group and the position of the group.

position of the children, size of the group so doing stuff like table or flex layout is part of the scope.
We (@ebigtm and I) did a flex one already and we might publish it.
A table layout is part of what a project I am involved with in doing now.
I also saw @iocoker do a calendar, which is a table layout.

So it is possible and the API is pretty solid so hooking into should be relatively straight forward.
The only pain is handling the correction, that makes it just hard for "no reason".
That is why I am thinking, what if we use left top origin instead of center. Maybe that will make it easier.
And I will try this small hack I have in mind that is equiv to `pathOffset` that might release us from the need to correct the position

src/shapes/ActiveSelection.ts
```diff
@@ -146,9 +146,7 @@ export class ActiveSelection extends Group {
   }
 
   /**
-   * If returns true, deselection is cancelled.
-   * @since 2.0.0
-   * @return {Boolean} [cancel]
+   * @override remove all objects
```
asturur, line 149: why removing the comment?

asturur:
We can make another pr for the binding issue

src/shapes/ActiveSelection.ts
ShaMan123: I think the jsdoc concats the comments of the super method so there is no need, it is spam

ShaMan123:
> > I have noticed that we introduced a regression with polys flagged with exactBoundingBox
> 
> What is the regression?

I should open a ticket.

1. stroke uniform poly
2. perform multiselection, select and deselect the poly
3. poly with jump a few pixels, maybe related to the stroke width value

ShaMan123:
So can we merge?

src/shapes/ActiveSelection.ts
asturur: that is bad, because if i m just reading the codebase i can't concat anything

src/LayoutManager/LayoutManager.spec.ts
```diff
@@ -305,19 +305,27 @@ describe('Layout Manager', () => {
...
 
-      const target = new Group([], { scaleX: 2, scaleY: 0.5, angle: 30 });
+      const rect = new FabricObject({ width: 50, height: 50 });
+      const target = new Group([rect], { scaleX: 2, scaleY: 0.5, angle: 30 });
 
       const context: StrictLayoutContext = {
         bubbles: true,
         strategy: manager.strategy,
         target,
+        targets: [rect],
```
asturur, line 315: The layoutManager.getLayoutResult is not using targets at all, it is using target.getObjects().

ShaMan123: hmmm
that is maybe something to discuss
targets are objects that changed (I think it is used only for subscribing)

src/shapes/ActiveSelection.ts
ShaMan123: And if you hover with your mouse?

ShaMan123: Screenshot 2024-01-11 at 10 32 50
 I can restore it but my point is that then you have to update a signle comment everywhere 
That is not manageable

src/LayoutManager/LayoutManager.spec.ts
asturur: Adding or removing this line doesn't do anything else than helping TS, is also a bit confusing in the test definition that one test 'LAYOUT_TYPE_INITIALIZATION ' pass empty targets to override this one here while 'LAYOUT_TYPE_IMPERATIVE' doesn't.

Also on my vscode both '[]' and '[rect]' give a red error in typescript.

I will investigate this typing issue right after merging, i don't think a function that doesn't use 'targets' should requires it in the typings

src/shapes/ActiveSelection.ts
asturur: yeah VSCODE does this, i m just not sure how it is supposed to be by the JSDOCS logic

asturur (approved):

PR #9591, opened 2024-01-13
fix(LayoutManager): remove unnecessary check

asturur:
## Description

Remove an instance check that shouldn't be needed anymore since we fixed that each group has its own layout manager and every layout manager has a target that is the group.


ShaMan123 (approved):
Do we want to stop passing target since it is bound?
I think it is fine as is but it can be removed.
Note that this pr means the layout manager can not be shared.
I prefer this design.

asturur:
Yes i think we landed on one group one layout manager, since the layout manager has state and is not entirely usable ( subscriptions and prevStrategy )
Also since is new i prefer the surface of features is restricted

asturur:
> Do we want to stop passing target since it is bound?

I don't want to change too much.
Target is passed by the group that uses it.
Documentation has to play a role in describing how to use it anyway

PR #9593, opened 2024-01-13
chore(): use deconstruction and constants in place of strings to save some bytes of code

asturur:
## Description
This is no functional changes a tentative to recover some bundle size


src/canvas/canvas_gestures.mixin.ts
ShaMan123, line 1: What is the fate of this file?

src/constants.ts
```diff
@@ -25,3 +25,15 @@ export const RIGHT = 'right';
...
 
 export const reNewline = /\r?\n/;
+
+export const MOVING = 'moving';
+export const SCALING = 'scaling';
+export const ROTATING = 'rotating';
+export const ROTATE = 'rotate';
+export const SKEWING = 'skewing';
+export const RESIZING = 'resizing';
+export const MODIFYPOLY = 'modifypoly';
```
ShaMan123: can we rename this?

src/controls/scaleSkew.ts
```diff
@@ -26,11 +27,11 @@ export const scaleOrSkewActionName: ControlCallback<
   const isAlternative = isAltAction(eventData, fabricObject);
   if (control.x === 0) {
     // then is scaleY or skewX
-    return isAlternative ? 'skewX' : 'scaleY';
+    return isAlternative ? 'skewX' : SCALE_Y;
   }
   if (control.y === 0) {
     // then is scaleY or skewX
-    return isAlternative ? 'skewY' : 'scaleX';
```
ShaMan123: skewX/Y?

ShaMan123 (approved):
In an ideal condition I would ask this to be merged once our queue is empty to avoid annoying merge conflicts with open PRs because it is a lot of diff ink

But we are not there yet so it is what it is

asturur:
this can wait a lot is not important

src/canvas/canvas_gestures.mixin.ts
asturur: at some point we need to find a solution for gestures

ShaMan123: I gave some thought.
We can introduce a control that fits the entire object for a 2 finger gesture that scales and rotates the object.
The math sounds straight forward.
Measuring the diff of the vector created between the 2 touches.
Then the dev can do for 3 fingers whatever based on the 2 finger control.
Thoughts?

ShaMan123: I would then argue that a mobile/touch app will not want the default controls

ShaMan123: Then touch become a first class citizen in fabric

PR #9597, opened 2024-01-14
fix(StyledText): add ability to unset style (issue #9578)

### LINKED ISSUE #9578: [Bug]: SetSelectionStyles can't delete style properties
## Problem

Currently setSelectionStyles only extends an object with new properties.
If we want to remove the fontSize property from a piece of selection we can't set it to undefined or that will cause the renderer to truncate the rendering.
In order to remove fontSize information for example we would need to set it to the default object fontSize and then run the function to normalize the style (text.cleanStyle).

This seems a bit too much, it should be possible to eventually directly remove some style property from the selection and not just add them.

I discovered this working on the sub/super script demo code that works on 5.x but doesn't work on 6.x

This was working code before, now is no more.
```js
function remove() {
	var active = canvas.getActiveObject();
	if (!active) return;
	active.setSelectionStyles({
		fontSize: undefined,
		deltaY: undefined,
	});
	canvas.requestRenderAll();
}
```

There is also another not nice behaviour.

```ts
  /**
   * Gets style of a current selection/cursor (at the start position)
   * @param {Number} startIndex Start index to get styles at
   * @param {Number} endIndex End index to get styles at, if not specified startIndex + 1
   * @param {Boolean} [complete] get full style or not
   * @return {Array} styles an array with one, zero or more Style objects
   */
  getSelectionStyles(
    startIndex: number,
    endIndex?: number,
    complete?: boolean
  ): TextStyleDeclaration[] {
    const styles: TextStyleDeclaration[] = [];
    for (let i = startIndex; i < (endIndex || startIndex); i++) {
      styles.push(this.getStyleAtPosition(i, complete));
    }
    return styles;
  }
```
getSelectionStyles with complete false will return an array of references to the style.
getSelectionStyles with complete true will return a clone of the style.

This is sneaky and should be made explicit

## Possible solutions:
- when calling setSelectionStyle change the extend code to actually delete from the object properties that are set as undefined.
- when rendering do not break if fontSize or other values are undefined
- expose a method deleteStyleFromSelection in which we can pass an object with keys we want to delete

for the other issue:
at developer facing level always expose a copy of the style and in case make a specific method if we want to expose a reference to the style object of a specific position

asturur:
## Description

closes #9578

_setSelectionStyles is behaving in a very confusing way.

`_getSelectionStyle` is the culprit but i m not touching it right now.
`_getSelectionStyle` returns a reference to the style object or a new empty one. The reference can also be to an empty object as well. So when you get an empty object you don't really know if is an actual reference or the original one.

Because of this we were checking if the returned object was empty, if it was we were anyway setting there a new empty object once again, so we would know that the text call would return a reference to the one we just created ( really strange ).

At that point we would use Object.assign to modify it.

This wasn't the bug actually, the bug is that if i pass `fontSize: undefined` i want to remove fontSize from the  current style section and not merge an undefined value that breaks rendering.


ShaMan123:
> #### Review or Edit in CodeSandbox
> Open the branch in [Web Editor](https://codesandbox.io/p/github/fabricjs/fabric.js/set-selection-styles?mode=review&utm_source=gh_app) • [VS Code](https://codesandbox.io/p/vscode?owner=fabricjs&repo=fabric.js&branch=set-selection-styles&utm_source=gh_app) • [Insiders](https://codesandbox.io/p/vscode?owner=fabricjs&repo=fabric.js&branch=set-selection-styles&insiders=true&utm_source=gh_app)
> 
> Open Preview

This WORKS! I think I once the PR closes the links die, that is why it didn't work when I tried

src/shapes/Text/StyledText.ts
```diff
@@ -188,11 +188,21 @@ export abstract class StyledText<
...
+      ...this._getStyleDeclaration(lineIndex, charIndex),
+      ...styles,
+    };
+
+    // then delete what is undefined in styles from newStyle
+    Object.keys(styles).forEach((key) => {
+      if (styles[key as keyof TextStyleDeclaration] === undefined) {
+        delete newStyle[key as keyof TextStyleDeclaration];
+      }
+    });
```
ShaMan123, line 202: Since you are creating a new object I think it is better to filter out undefined beforehand instead of creating it and then deleting stuff

```suggestion
    // first create a new object that is a merge of existing and new
    const newStyle: TextStyleDeclaration = {
      ...this._getStyleDeclaration(lineIndex, charIndex),
      ...Object.fromEntries(Object.entries(styles).filter(([key, value]) => value!==undefined),
    };

```

ShaMan123: can we also rename styles to style
very confusing, it is not an array, right?
Why don't we support an array?

src/shapes/Text/StyledText.ts
```diff
@@ -244,10 +254,14 @@ export abstract class StyledText<
 
   /**
    * get the reference, not a clone, of the style object for a given character,
-   * if not style is set for a pre det
+   * if not style is set for a pre determined line or char, return a new object.
+   * this is tricky and confusing because when you get an empty object you can't
+   * determine if is a reference or a new one.
+   * @TODO this should return always a reference or always a clone, and if necessary undefined.
```
ShaMan123, line 260: typos

src/shapes/Textbox.ts
```diff
@@ -225,9 +225,10 @@ export class Textbox<
   }
 
   /**
+   * @protected
    * @param {Number} lineIndex
    * @param {Number} charIndex
-   * @private
+   * @return {TextStyleDeclaration} style object a REFERENCE to the existing one or a new empty object
```
ShaMan123, line 231: rephrase

ShaMan123 (changes requested):
Minor changes

ShaMan123:
I would say that the name `setselectionStyles` is more than confusing because it is actually accepting a single style.
I used the codesandbox app to test the bug fix. So fun!

src/shapes/Text/StyledText.ts
asturur: yes is a single style.
We don't support an array becuase this is a single style object that you apply to a range of text.

asturur: This code is not equivalent. We need to delete what is in the old style as well, we are not cleaning up styles.

The equivalent would be something in the lines of:
- mixing the entries of styles and the previousStyle
- loop with a filter
- do the fromEntries

ShaMan123: I disagree, didn't you delete only the incoming `stylez arg?
Let's change this, see how confusing it is, this thread is evidence

asturur:
@ShaMan123 could you re-review this one? i would not rename the method at this point, but let me know how you feel strongly about that object code.

src/shapes/Text/StyledText.ts
```diff
@@ -243,11 +253,15 @@ export abstract class StyledText<
   }
 
   /**
-   * get the reference, not a clone, of the style object for a given character,
-   * if not style is set for a pre det
+   * Get a reference, not a clone, to the style object for a given character,
+   * if no style is set for a line or char, return a new empty object.
+   * This is tricky and confusing because when you get an empty object you can't
+   * determine if it is a reference or a new one.
+   * @TODO this should always return a reference or always a clone or undefined when necessary.
```
ShaMan123, line 260: good comment
bad design

ShaMan123 (approved):
The PR name is confusing 
I would say:
add ability to unset style

src/shapes/Text/StyledText.ts
asturur: yes we have a TODO to fix it. Fixing it now has a lot of subtle consequences i don't want to think of

src/shapes/Text/StyledText.ts
```diff
@@ -181,18 +181,28 @@ export abstract class StyledText<
...
+      ...this._getStyleDeclaration(lineIndex, charIndex),
+      ...style,
+    };
+
+    // then delete what is undefined in styles from newStyle
+    Object.keys(style).forEach((key) => {
+      if (style[key as keyof TextStyleDeclaration] === undefined) {
+        delete newStyle[key as keyof TextStyleDeclaration];
+      }
+    });
```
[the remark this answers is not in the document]
asturur: so newStyle is created correctly right now, you don't like the shape of the code that creates it.

if i use filter, i have to:
- mix the 2 object together, creating a new one 
- use Entries, that is a loop + creation of an array
- filter that returns the new list of entries we want to keep
- fromEntries that create a new object from the filtered list

this is more compact, but i can't call it more straight  forward at all:
```js
 Object.fromEntries(Object.entries({
    ...this._getStyleDeclaration(lineIndex, charIndex),
    ...style,
  }).filter(([key, value]) => value !== undefined));
```

I can change it but you didn't mention what is unpreferable about the actual code.

The original comment was:

```
Since you are creating a new object I think it is better to filter out undefined beforehand instead of creating it and then deleting stuff
```

We are still creating a new object with all the stuff ( the spread statement ), then we are creating an array of arrays with all entries, a new one filtered, then we are creating a new object once again.

Is that you don't want to carry over an object on which has been used the delete keyword?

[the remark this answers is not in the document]
asturur: With pick is overly complicated

asturur:
Addressed the comments that were added post approval

src/shapes/Text/StyledText.ts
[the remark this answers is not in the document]
asturur: pickby gives us a negative code size and probably the same speed, while the fromEntries <-> toEntries is definetely worse on both aspects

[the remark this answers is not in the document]
ShaMan123: pickBy,yes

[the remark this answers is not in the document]
ShaMan123: nice!

PR #9602, opened 2024-01-14
fix(SVGParser): avoid crashing on SVG that use @import css feature

asturur:
## Description

initially found in  #7434 by [k88manish](https://github.com/k88manish), credits to hi,

This won't parse the style that the svg is trying to import, but will stop the parser from crashing.
The original pr can't be merged because based on the old JS code.

The SVG could have an import statement inside a style tag.
FabricJS expect the content of style tags to be css rules and its rudimental parser will break if inside there is just an import.
This is easily detectable because it will rely on the detection of '{' and '}', if those are not found the split command will return an array of length 1, at that point we exit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(SVGParser): avoid crashing on SVG that use @import css feature [#9602](https://github.com/fabricjs/fabric.js/pull/9602)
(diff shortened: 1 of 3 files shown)


asturur:
@ShaMan123 i m merging this one, is literally and extra condition on a find, with a test, let's get it out of the way.

PR #9605, opened 2024-01-17
fix: Support user-defined layout managers during initialization

zhe-he:
I defined a `Frame` that inherits from a `Group` and customized the `LayoutManager`. When I load the exported data from the last time using the canvas.loadFromJSON method, I found that the getLayoutResult method of the internal `NoopLayoutManager` is executed first for the initially loaded `Frame`. However, I have overridden this method in my `Frame` and it should not be executed. So, I identified the issue at this point.

### DIFF
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -30,17 +30,6 @@ import {
 import type { SerializedLayoutManager } from '../LayoutManager/LayoutManager';
 import type { FitContentLayout } from '../LayoutManager';
 
-/**
- * This class handles the specific case of creating a group using {@link Group#fromObject} and is not meant to be used in any other case.
- * We could have used a boolean in the constructor, as we did previously, but we think the boolean
- * would stay in the group's constructor interface and create confusion, therefore it was removed.
- * This layout manager doesn't do anything and therefore keeps the exact layout the group had when {@link Group#toObject} was called.
- */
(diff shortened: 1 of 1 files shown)


ShaMan123:
Hiw do you use it? What data do you have to restore?
Don't you have the bbox data of your objects?
We skip the layout because we expect the layout data to be available and known

asturur:
My opinion is that if the 'frame' needs to execute the layoutManager and not trust the data that has been serialized, the 'frame' should have its own 'fromObject' method.

asturur:
Let me explain the reasoning around it.
The Group as of now is trusting its serialized representation when restoring.
That works well and a lot of tests and code rely on that. The LayoutManager itself won't work if we don't follow this right now.
Recalculating the Layout is expensive, if a group is correctly serialized it doesn't need to do that, and this is the premise on which the actual code has been built.

If you write your own fromObject method you keep the freedom of creating your own class while the Group can continue to work as it is working currently.

asturur:
Of course if you want help in understanding why the current 'frame' needs this different approach, ask!

ShaMan123:
I think this is reasonable and more important for me is that it is easy to customize, as you suggested.
So it answers my needs.
If one day we are convinced differently we could look into a different approach.
Closing, voice your opinions otherwise.

PR #9610, opened 2024-01-19
fix(): compositionEnd event handler is not registered correctly. (regression from f91362c )

xiaofeiniuuu:
Textbox Chinese input with ctrl+a select all question

The inCompositionMode variable was not set to false after input was completed

This is because the input of the event name compositionend binding is incorrect

The current binding method in the code

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): compositionEnd event handler is not registered correctly. (regression from f91362c ) [#9610](https://github.com/fabricjs/fabric.js/pull/9610)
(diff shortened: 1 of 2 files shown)


ShaMan123 (commented):
Good catch!
Please run
npm run prettier:write 
to fix the added whitespace and commit the changes with a change log entry
Then we can merge

asturur:
seems a regression we introduced here:
https://github.com/fabricjs/fabric.js/commit/f91362cd0336cb205d92f6e9c4bdb555fd5d064c
Please fix the prettier warnings and we can merge

asturur:
ok i managed to push prettier:write on the remote branch

asturur:
Thank you good catch!

PR #9614, opened 2024-01-21
fix(Group): layout manager `fromObject`

ShaMan123:
## Description

Motivation #9605

A different approach to the `NoopLayoutManager`
This approach uses the standard API to skip layout: `shouldPerformLayout` and makes it an edge case handled within the API and not externally, making it even easier to use because now `shouldPerformLayout` handles all cases

During this PR I found a bug:
Since the `NoopLayoutManager` skipped `performLayout` children were not subscribed on initialization. This would not have happened if we noop `calcLayoutResult` which is what this PR does using the API

I also removed the `setCoords` call because it was introduced in https://github.com/fabricjs/fabric.js/pull/9522#discussion_r1438003603 wrongly

### DIFF
--- a/src/LayoutManager/LayoutManager.spec.ts
+++ b/src/LayoutManager/LayoutManager.spec.ts
@@ -5,6 +5,7 @@ import { Group } from '../shapes/Group';
 import { FabricObject } from '../shapes/Object/FabricObject';
 import { Rect } from '../shapes/Rect';
 import { LayoutManager } from './LayoutManager';
+import { ClipPathLayout } from './LayoutStrategies/ClipPathLayout';
 import { FitContentLayout } from './LayoutStrategies/FitContentLayout';
 import { FixedLayout } from './LayoutStrategies/FixedLayout';
 import {
@@ -740,7 +741,12 @@ describe('Layout Manager', () => {
       ).toMatchObject([child]);
     });
 
-    it('should subscribe objects when created `fromObject`', async () => {
+    describe.each([
+      [FitContentLayout.type, FitContentLayout],
+      [FixedLayout.type, FixedLayout],
+      [ClipPathLayout.type, ClipPathLayout],
+      [undefined, FitContentLayout],
+    ] as const)('fromObject %s', (type, strategy) => {
       const objectData = {
         width: 2,
         height: 3,
@@ -756,13 +762,32 @@ describe('Layout Manager', () => {
             strokeWidth: 0,
           }).toObject(),
         ],
+        clipPath: new Rect({
+          width: 50,
+          height: 50,
+          top: 0,
+          left: 0,
+          strokeWidth: 0,
+        }).toObject(),
+        layoutManager: {
+          strategy: type,
+        },
       };
+      it('should subscribe objects', async () => {
+        const group = await Group.fromObject(objectData);
+        const rect = group.item(0);
+        expect(
+          Array.from(group.layoutManager['_subscriptions'].keys())
+        ).toMatchObject([rect]);
+      });
 
-      const group = await Group.fromObject(objectData);
-      const rect = group.item(0);
-      expect(
-        Array.from(group.layoutManager['_subscriptions'].keys())
-      ).toMatchObject([rect]);
+      it('should skip initialization layout', async () => {
+        const spy = jest.spyOn(strategy.prototype, 'shouldPerformLayout');
+        const group = await Group.fromObject(objectData);
+        expect(group.layoutManager.strategy['_fromObject']).toBeTruthy();
+        expect(spy).toHaveBeenCalledTimes(1);
+        expect(spy).toHaveReturnedWith(false);
+      });
     });
 
     test.each([true, false])(
--- a/src/LayoutManager/LayoutManager.spec.ts
+++ b/src/LayoutManager/LayoutManager.spec.ts
@@ -3,6 +3,7 @@ import { Point } from '../Point';
 import { StaticCanvas } from '../canvas/StaticCanvas';
 import { Group } from '../shapes/Group';
 import { FabricObject } from '../shapes/Object/FabricObject';
+import { Rect } from '../shapes/Rect';
 import { LayoutManager } from './LayoutManager';
 import { FitContentLayout } from './LayoutStrategies/FitContentLayout';
 import { FixedLayout } from './LayoutStrategies/FixedLayout';
@@ -737,6 +738,31 @@ describe('Layout Manager', () => {
       expect(
         Array.from(group.layoutManager['_subscriptions'].keys())
       ).toMatchObject([child]);
+    });
+
+    it('should subscribe objects when created `fromObject`', async () => {
+      const objectData = {
+        width: 2,
+        height: 3,
+        left: 6,
+        top: 4,
+        strokeWidth: 0,
+        objects: [
+          new Rect({
+            width: 100,
+            height: 100,
+            top: 0,
+            left: 0,
+            strokeWidth: 0,
+          }).toObject(),
+        ],
+      };
+
+      const group = await Group.fromObject(objectData);
+      const rect = group.item(0);
+      expect(
+        Array.from(group.layoutManager['_subscriptions'].keys())
+      ).toMatchObject([rect]);
     });
 
     test.each([true, false])(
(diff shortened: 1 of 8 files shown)


src/LayoutManager/LayoutManager.spec.ts
```diff
@@ -739,6 +740,31 @@ describe('Layout Manager', () => {
...
+          }).toObject(),
+        ],
+      };
+
+      const group = await Group.fromObject(objectData);
+      const rect = group.item(0);
+      expect(
+        Array.from(group.layoutManager['_subscriptions'].keys())
+      ).toMatchObject([rect]);
+    });
```
ShaMan123: will fail on master

ShaMan123 (commented):
This is better IMO

ShaMan123:
@zhe-he take a look 
This should provide you for what you asked for
You will need to override `shouldPerformLayout` and return true from the specific case

asturur:
Please remove the bug in its own PR, so that can go in in any moment.

I don't want to discuss the api change before we know why this would be necessary at all.

ShaMan123:
will do

ShaMan123:
Wait, not good.
Since the subscriptions will be present of NoopLayoutManager it will cause a bug when unsubscribing.

ShaMan123:
shows that it is better to stick to the API

zhe-he:
Hmm... I'll explain why I customized this.

Similar to Figma's `Frame`, `Frame` can have auto-layout properties or not. When it doesn't have auto-layout properties, you set the width and height yourself. When it does have auto-layout properties and its width is adaptive, then during initialization, I need to calculate the widths of all its subsets to determine its width.

Now I'm solving this issue by setting it again after initialization.

I think this approach (LayoutStrategy#shouldPerformLayout) should also work. I'll look into it first. Thank you for your suggestion.

asturur:
@zhe-he but what is wrong with the suggestion of just having your fromObject method in your class?
We don't have to solve problems with new apis that can be solved already with the current api.

This proposal of fromObject is a step back in the world of 'initialized: true/false' that we always said we don't like.

asturur:
> Wait, not good. Since the subscriptions will be present of NoopLayoutManager it will cause a bug when unsubscribing.

can you explain this comment here?

asturur:
the fromObject with the flag is a big no go for me.
Since there is that bug of the subscriptions that needs to be fixed, i will revert the pr of the noop first of all,
then we can reason of how to go back there without the bug.

I rather prefer an extra type than a flag.

asturur:
or make a direct call to the object registration after assigning the correct layout manager

zhe-he:
> The Group as of now is trusting its serialized representation when restoring.
> That works well and a lot of tests and code rely on that. The LayoutManager itself won't work if we don't follow this right now.
> Recalculating the Layout is expensive, if a group is correctly serialized it doesn't need to do that, and this is the premise on which the actual code has been built.

I may not modify the properties of child objects on the canvas, but instead directly modify the properties of child objects in the exported serialized data. In this case, I have two options in front of me: one is to modify the parent object's properties in the exported serialized data through computation, and the other is to recompute the parent object's properties at initialization. If I choose the second approach, I cannot trust its serialized representation. So I would like to execute a custom performantLayout instead of an empty performantLayout during initialization. 

---
I see that the fromObject method has been added here, and by modifying the value of _fromObject, the performLayout is no longer executed when loading serialized data. If I don't trust the serialized data, I can simply modify the return value of shouldPerformLayout in the custom object to execute performLayout during initialization. I think this method is great.

---
> the fromObject with the flag is a big no go for me.

Can we add an additional property 'dirty' to the LayoutManager, with a default value of false or undefined? If the user manually changes it to true, performLayout will be executed during initialization; otherwise, it won't be executed. 
Alternatively, we can split performLayout into two functions: the original performLayout and a new performLayout2, which is only executed after loading serialized objects. performLayout2 is initially an empty function for users to customize, and its invocation can be placed inside Group.fromObject.

ShaMan123:
I like both approaches
The first approach is very simple and is a fabric convention, therefore a great approach.
For the second, how would you know to skip the initialization layout if performLayout2 will be called? We will need a flag in the constructor so the first approach is better
The dev could then flag the layout as dirty when constructing the layout classes

zhe-he:
The second solution I want to propose is probably like this.
```javascript
class Group {
// noop
fromObjectPerformLayout() {},
fromObject() {
      // All layouts skip performLayout during initialization.
      const group = new this(objects, ...);
      // Custom layouts can override the performLayout2 method to recalculate.
      group.fromObjectPerformLayout();
}
}
```
```javascript
class Frame extends Group {
    fromObjectPerformLayout() {
          this.layoutManager.performLayout({ type: "xxx" });
    }
}

```

asturur:
The old pr can't be easily reverted.
There are test changes and a lot of stuff that need to be re-done again.

asturur:
@zhe-he is still unclear why you can't write your own fromObject method  in the Frame class, that just adds the layout manager, exactly as you did with your initial pr.

zhe-he:
> @zhe-he is still unclear why you can't write your own fromObject method in the Frame class, that just adds the layout manager, exactly as you did with your initial pr.

Hmm... I can modify the fromObject method in Frame to solve my issue.
It may not be necessary to modify the fromObject method in Group.
 Perhaps the pull request I submitted could be a mistake? 
I didn't consider that Fabric relies on the imported serializable objects when I submitted the pull request.

zhe-he:
If not modified, all custom layout objects will not be initialized during data import. If they need to be, they must override the fromObject method.

asturur:
@zhe-he i still think i have very little informations to give you suggestions.
How many custom classes you made?
I understand one, the Frame.
The Frame wants to run the performLayout at the point of restore, contrary to what group does.
If this is the only issue it can be solved with Frame.fromObject being custom. And this we said.
If there is more maybe this is not enough.

zhe-he:
@asturur 
Currently, there are three custom classes. One is "Frame," which inherits from "Group." The second one is "Component," which inherits from "Frame." The third one is "SubComponent," which inherits from "Component." They are in a chain relationship. Therefore, I only need to modify the "frameObject" of "Frame" to solve this problem.

asturur:
The registration bug needs to be fixed, and i don't want to do internal flags for fromObject.
I have a different proposal to disconnect the registration of objects from the performLayout, let me see if i can make a different example

asturur:
well what i wanted to do can't be really done as good as i wish because i don't like the NoopClassManager hack i have to add. `{} as Group`
But this is a tentative of separating registration from layouting.
https://github.com/fabricjs/fabric.js/compare/registering-objects-group-fromObject?expand=1

It would be also ok to make subscribe public, and just call it on every object in the Group FromObject.

I didn't even test it for now, i wanted to put on the table first.

I have some notes on the context, one is that i can't see where we use 'stopPropagaton' at all, and i wonder why is there.

asturur:
I think after much thinking my preferred solution is this one:
https://github.com/fabricjs/fabric.js/pull/9661

Just expose the subscription and call it when necessary.

Please let me know, i see you are maybe all busy, but if in 1 or 2 weeks i don't get a feedback i ll just move on.

ShaMan123:
What do about adding a prop/flag/indication to the options passed by `fromObject`?
```
new FabricObject({ ...options, deserialized:true })
```

I think this is so much easier and straight forward and provides useful context.



We can protect it by wrapping in with a private class and checking instanceof if you want to be protective but I am not sure that is needed

```typescript
class DeserializedOptions {
   constructor(options) { 
      Object.assign(this, options)
  }
}



class FabricObject {
   static fromObject(options) {
      return new FabricObject(new DeserializedOptions(options);
  }
  constructor(options) {
    const isDeserialized = options instanceof DeserializedOptions
  }

}

```

ShaMan123:
I dislike both the suggestions, I commented in the PR

asturur:
@ShaMan123 'dislike' is a sensation that does not really matter.
Just explain why in the opinion the solution isn't good.

You pointed out the restriction in the api context and that has been restored, what else isn't working in your opinion?

We both said in the past that internal flags to say 'initialized: true/false' or specific flags to initialize the object differently were something that long term didn't work out.
Is not that i don't like them, we excluded them in the past.
Indeed we removed the flag from the group that was taking care of doing a dividing the codepath for objects that were restored from json, i don't see why we would like to put it back now. What changed?

That is why i don't wan't to add back a flag like fromObject() or '{ fromObject: true }'.

PR #9502, opened 2023-11-19
CD(): vue template

ShaMan123:
## Motivation

https://github.com/vuejs/core/issues/2370
https://github.com/vuejs/core/issues/3236

## Description

Add a vue template to make vue quick starting easier
Merge #9490 first

PR #9615, opened 2024-01-21
CD(): expose vue deployed app

ShaMan123:
## Description

When I created the vue app #9502  I forgot to update some config files so I updated them and the README


ShaMan123 (commented):
It works, see the codesandbox comment

asturur (approved):

PR #9618, opened 2024-01-21
Fix(StaticCanvas): Restrict and correct setDimensions typings.

asturur:
## Description

setDimensions can be used as:
```
canvas.setDimensions({
  width: '100%',
  height: 'auto',
}, { cssOnly: true });
```

When we want a responsive static canvas.
Current types have been restricted incorrectly to number only here: 
https://github.com/fabricjs/fabric.js/commit/41077213090b81c255d898a5dfbf94e2428e2317#diff-75d92bd48875f2c27e21818dc1794f3cfae0b91452086e34ce64fc09dc707e71L441


src/canvas/StaticCanvas.ts
```diff
@@ -45,10 +45,19 @@ import { staticCanvasDefaults } from './StaticCanvasOptions';
...
+ * Better try to restrict with types to avoid confusion.
+ */
+export type TCanvasSizeOptions =
+  | {
+      backstoreOnly?: true;
+      cssOnly?: false;
+    }
+  | {
+      backstoreOnly?: false;
+      cssOnly?: true;
```
ShaMan123, line 59:
```suggestion
  | {
      backstoreOnly?: true;
      cssOnly?: never;
    }
  | {
      backstoreOnly?: never;
      cssOnly?: true;
```

should be never or undefined, not sure

src/canvas/StaticCanvas.ts
```diff
@@ -323,14 +332,11 @@ export class StaticCanvas<
    * @param {Boolean}       [options.cssOnly=false]       Set the given dimensions only as css dimensions
    */
   setDimensions(
-    dimensions: Partial<TSize>,
-    { cssOnly = false, backstoreOnly = false }: TCanvasSizeOptions = {}
+    dimensions: Partial<TSize | CSSDimensions>,
+    options: TCanvasSizeOptions = {}
```
ShaMan123, line 336: question about this
What happens if I pass a css dimension string with option cssOnly = false?

I think we should change the second arg to be a string option or merge it into the first arg.
`setOnly?: 'css' | 'backstore'`

ShaMan123: Or better off 
```
css?: boolean = true,
backstore?: boolean = true,
```

OR add
```
cssWidth?: whatever,
cssHeight?: whatever,
```

ShaMan123 (commented):
I never knew this was an option
Curious to see how it works with window resizing

src/canvas/StaticCanvas.ts
ShaMan123: I prefer the last, explicit and clear

```
width: number,
height: number,
cssWidth?: whatever,
cssHeight?: whatever,
```

src/canvas/StaticCanvas.ts
asturur: i tried this, it gets angry.  false is still a valid option you could pass, the important is that you don't do the double true.

src/canvas/StaticCanvas.ts
asturur: yes with 4 width/height would be nice, but would be another breaking change.
Let's have this as a fix of the current situation and i ll open a different PR with the breaking change

asturur:
> I never knew this was an option Curious to see how it works with window resizing

i was doing this https://tapto-designer.netlify.app/ and with the window resize event and setDimensions it was too laggy and also too many react issue with the resize observer, so i decided to go with the CSS and i noticed i wasn't possible anymore because of TS ( but worked if forced ).

If you have a bunch of jpegs on your pc ( like 10 or more ) you can load them with add files ( is all local, data goes nowhere ) and you ll get a canvas per jpeg. Then you can resize the window and you will see is pretty smooth

src/canvas/StaticCanvas.ts
asturur: maybe i can just add cssWidth and cssHeight and deprecate the rest.

asturur: > What happens if I pass a css dimension string with option cssOnly = false?

This has been solved now with stricter typings

asturur:
cssWidth and cssHeight are probably the way to go, and a specific PR should refactor the code around to make that possible

src/canvas/StaticCanvas.ts
```diff
@@ -322,15 +347,21 @@ export class StaticCanvas<
...
+    dimensions: Partial<TSize | CSSDimensions>,
+    options?: TCanvasSizeOptions
   ) {
-    this._setDimensionsImpl(dimensions, {
-      cssOnly,
-      backstoreOnly,
-    });
-    if (!cssOnly) {
+    this._setDimensionsImpl(dimensions, options);
+    if (options && !options.cssOnly) {
```
ShaMan123, line 364: @asturur  this is a regressions
if no options are passed the canvas does not render

ShaMan123: Quoting: A typings PR should be only a typings PR

asturur: @ShaMan123 you are using snarky comments since yesterday.
I don't like that.
This PR has been extensively looked by me and you and a bug slipped in.
It was a typing pr indeed, no logic was meant to change

asturur: And the PR is titled as fix() and includes the word typings is not that i wrote `chore(TS)` and changed the code

PR #9619, opened 2024-01-21
chore(): Path remove unused code from path render function

asturur:
## Description

Those let variables were a leftover from the old rendering code when we supported in render all type of path commands.
After the path semplification logic ( moving all commands to absolutes L, M, C, Q, Z ) the carry over of those values is not needed.


ShaMan123 (approved):
I was thinking
Why don't we have a ref of Path2D to save on iterations?
Currently this runs every render cycles if caching is invalidated

asturur:
yes Path2D was an optimization we skipped because not supported by all browsers, but it was planned long ago, like when issues had 3 numbers.
It should be done, with an eye on how to invalidate/refresh that smartly, but definetely can be done

PR #9620, opened 2024-01-21
chore(Object) removed code for options we dont want to use in _getTransformedDimensions

asturur:
## Description

It happened already that we tried to use _getTransformedDimensions with generic overriding options to work around issues, and then we disagreed on that being a good practice.
This PR reduces the options that are accepted by _getTransformedDimensions to the one we really use.

### DIFF
--- a/src/shapes/Polyline.ts
+++ b/src/shapes/Polyline.ts
@@ -18,6 +18,7 @@ import type { ObjectEvents } from '../EventTypeDefs';
 import { cloneDeep } from '../util/internals/cloneDeep';
 import { CENTER, LEFT, TOP } from '../constants';
 import type { CSSRules } from '../parser/typedefs';
+import type { _getTransformedDimensionsParams } from './Object/ObjectOrigin';
 
 export const polylineDefaultValues: Partial<TClassProperties<Polyline>> = {
   /**
@@ -163,10 +164,14 @@ export class Polyline<
       bboxNoStroke = makeBoundingBoxFromPoints(
         this.points.map((p) => transformPoint(p, matrix, true))
       ),
+      // @TODO it is very strange that we allow for scaleX and scaleY in the options
+      // but then here we don't use them
       scale = new Point(this.scaleX, this.scaleY);
     let offsetX = bbox.left + bbox.width / 2,
       offsetY = bbox.top + bbox.height / 2;
     if (this.exactBoundingBox) {
+      // @TODO it is very strange that we allow for skewX and skewY in the options
+      // but then here we don't use them
       offsetX = offsetX - offsetY * Math.tan(degreesToRadians(this.skewX));
       // Order of those assignments is important.
       // offsetY relies on offsetX being already changed by the line above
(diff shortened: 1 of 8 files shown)


src/shapes/Polyline.ts
```diff
@@ -232,33 +237,21 @@ export class Polyline<
...
-      if (
-        Object.keys(options).some(
-          (key) =>
-            this.strokeUniform ||
-            (this.constructor as typeof Polyline).layoutProperties.includes(
-              key as keyof TProjectStrokeOnPointsOptions
-            )
-        )
-      ) {
+      if (this.strokeUniform || 'skewX' in options || 'skewY' in options) {
```
asturur, line 250: there are 4 options and only skewX and skewY are  in layoutProperties, so we can check specifically.

jiayihu:
At company where I work, we actually use this hack to get the transformed dimensions of an object when only some properties (delta) have been changed by another user. So we call the method with the changed methods to calculate what the dims are going to be.

Of course we can just define a function that does as before. I'm just wondering what would be the rational of reducing the amount of accepted props while not offering a good alternative. Nobody seems to get benefits here.

asturur:
it wasn't about benefits but just that his shouldn't be done.
This is just a clean up.
getTransformedDimensions supports scaleX scaleY and skewX, skewY because of how the skew mouse handler is written.
Then at some point we extended it to support more stuff and we realized it won't work, once i used with strokeWidth and we agreed was a bad idea and we changed the PR.

I don't need to merge this now if you don't have an alternative, but in general a method that says:
'tell me how big this instance is, but pretending some values are different' doesn't make much sense.
The biggest issue i have with this functionality is also that you have to pass down the values to other utils and support it everywhere down the chain.

You can create a clone of the instance and measure it with different props, or use the  Function.call method passing an ad hoc context.

jiayihu:
> The biggest issue i have with this functionality is also that you have to pass down the values to other utils and support it everywhere down the chain.

From a consumer POV this is actually better than the options you suggested, as it doesn't involve doing mutations (unreliable as you depend on the exact implementation of each underlying method) or creating clones (unreliable as well since some internal states may not be cloned correctly since we're dealing with non-trivial classes, not plain objects).

Passing down values is more unconvenient for the maintainer, but far more predictable and reliable for consumer. I wonder if we could borrow some concepts from functional-programming, such as...the Reader Monad? https://dev.to/gcanti/getting-started-with-fp-ts-reader-1ie5. 

Just kidding. But it could be an idea for instance starting to extract some props under `obj.state`, e.g.:

```
transformState: { x: number, y: number, width: number, height: number, scaleX: number, scaleY: number, angle: number, skewX: number, skewY: number }
```

Then having methods using `this.transformState`, so one would' only need to override `transformState` following your 3rd suggestion. Or the utils just need to accept the transform state as single extra param, if you want to keep them pure.

Could potentially also play nicer with reactive states. I've suggested multiple times to @ShaMan123 the need for some sort of higher-level reactive state like https://preactjs.com/guide/v10/signals/, which avoids the need for `getters` when you want some properties to be derived from others. For instance, transformed dimensions is a derived state of `transformState`.

asturur:
i think the solution is very simple, methods work on the instance state, function can work on any passed parameters.

So if this kind of predicting size is really useful, it should be come an util, then the method can call the util with its own state, 'this' while the utility itself can be called with any state object you want.

But again, this shouldn't really have existed in my opinion, i should have never added scaleX, scaleY, skewX and skewY ages ago, but in that moment supporting skew by mouse seemed so cool to me. I do not know why. I think skewing by mouse is mostly useless.

jiayihu:
> i think the solution is very simple, methods work on the instance state, function can work on any passed parameters.

True and fabric should limit usage of methods IMO, as it makes overriding harder. You have to ensure the instance state is exactly as expected at each internal method. It's better instead to prepare all the state mutations at once, then call functions passing the state as params.
This makes the code also more easier to reason. Mutations typically happen at event start/end, core logic is pure. And for overriding, you just need to do the mutations accurately at the event start/end indeed.

> Is your use case measuring changes to the instance size from changing properties? Is that all?

Yes. I'm just getting the "future" dims passing the change delta, then I'll use the dims to compute the transform matrix of where the object is gonna be. The transform matrix will then be used as "invariant" thoughout the syncing logic, as it tells where the object MUST be in the end, no matter what's the current transform. 
Useful for different reasons, even just to debug current object transform vs expected final transform.

PR #9621, opened 2024-01-21
chore(): Remove over-protective cloneDeep from fromObject

asturur:
## Description

We need to protect from unwanted references when restoring from data.
Today we deeply clone every object, and that is expensive ( measured on an app with 91 canvases  it went from 480 to around 420 ms ) and overly protective.

In the case of Object._fromObject `options` ( renamed in `serializedOptions` ) is passing through the enliving process that already a shallow clone

Other deep level cloning are left to the single objects duty or single properties ( gradients, pattern, array ) but they need to be carefully inspected

This PR is a wip and will add more checks on data cloning later


src/gradient/Gradient.ts
```diff
@@ -112,14 +112,19 @@ export class Gradient<
...
     this.coords = {
       ...(this.type === 'radial' ? radialDefaultCoords : linearDefaultCoords),
       ...coords,
     } as GradientCoords<T>;
-    this.colorStops = colorStops.slice();
+    this.colorStops = colorStops.map((colorStop) => ({
+      color: colorStop.color,
+      offset: colorStop.offset,
+      opacity: colorStop.opacity,
+    }));
```
asturur, line 127: Gradient should have a fromObject like Pattern

src/shapes/Object/Object.ts
```diff
@@ -1564,20 +1563,22 @@ export class FabricObject<
...
   static _fromObject<S extends FabricObject>(
-    { type, ...object }: Record<string, unknown>,
+    { type, ...serializedObjectOptions }: Record<string, unknown>,
     { extraParam, ...options }: Abortable & { extraParam?: string } = {}
   ): Promise<S> {
-    return enlivenObjectEnlivables<any>(cloneDeep(object), options).then(
-      (enlivedMap) => {
-        const allOptions = { ...options, ...enlivedMap };
+    return enlivenObjectEnlivables<any>(serializedObjectOptions, options).then(
+      (enlivedObjectOptions) => {
```
asturur, line 1570: enlivedObjectOptions is a shallow clone of object so we can delete or reassign properties freely.

src/shapes/Object/Object.ts
```diff
@@ -1564,20 +1563,22 @@ export class FabricObject<
    * @returns {Promise<FabricObject>}
    */
   static _fromObject<S extends FabricObject>(
-    { type, ...object }: Record<string, unknown>,
+    { type, ...serializedObjectOptions }: Record<string, unknown>,
     { extraParam, ...options }: Abortable & { extraParam?: string } = {}
   ): Promise<S> {
-    return enlivenObjectEnlivables<any>(cloneDeep(object), options).then(
-      (enlivedMap) => {
-        const allOptions = { ...options, ...enlivedMap };
```
asturur, line 1572: the only other valuable here should be signal. not worth a full spread.

src/gradient/Gradient.ts
ShaMan123: #9144

PR #9622, opened 2024-01-21
feat(Gradient): BREAKING: Remove opacity from colorstops in live Gradient class

asturur:
## Description

Another item that comes up in my flame graphs often is the gradien 'toLive' function.
Half of the time ( 0.13 of 0.25 in most samples ) is spent parsing a color string and outputting a new rgba value.

That is wasteful since the couple 'color' and 'opacity' on separate values is something that is inherited from SVG and doesn't have to bleed into fabric internals.

This PR propose to remove opacity from the gradient colorStop type, merging it in the color string itself.

This PR would be a no brainer if it wasn't that compatibility with old saved objects needs to be maintained and so a compatibility function needs to be created and maybe removed later down in next versions of fabric.


ShaMan123:
You can/should optimize the calls.
If nothing changed `toLive` should return a stored ref.
I agree with all the above

ShaMan123:
We can always expose a migration function for data and then devs can handle their `fromObject` instead of polluting fabric.

asturur:
> We can always expose a migration function for data and then devs can handle their `fromObject` instead of polluting fabric.

Yes i will make a migration function, called exactly migrationForBreakingChangeXYZ that then will get removed when grace time passes.

asturur:
> You can/should optimize the calls. If nothing changed `toLive` should return a stored ref. I agree with all the above

I do not know if i can create a toLive with ctx of value X and then reuse it on a different ctx somewhere else.

We can measure if that has a different impact once this part is cleaned up

asturur:
I may draft a spec for this, evaluate issue and try to fit into 6.0.
I m writing specs today

src/gradient/Gradient.ts
```diff
@@ -133,11 +132,9 @@ export class Gradient<
    */
   addColorStop(colorStops: Record<string, string>) {
     for (const position in colorStops) {
-      const color = new Color(colorStops[position]);
       this.colorStops.push({
         offset: parseFloat(position),
-        color: color.toRgb(),
-        opacity: color.getAlpha(),
+        color: colorStops[position],
```
asturur, line 137: maybe those can be sort correctly every time we add new.
Unsorted colorstops are often cause of strange bugs.
Canvas doesn't care much but SVGs or PDFs do.

asturur:
I wanted to see if i could complete this for 6.0 but requires AT LEAST making the json importer updater, so too much work pushed in rush

PR #9624, opened 2024-01-22
cd(): Surface the minified build as standard when importing and also provide a non bundled build

asturur:
## Description

I was looking at the bundle analyzer tool of one app and realized we are always importing the 900kb file
Also enabled source maps for everything.

Then i understood there is another option to build the files and that is by preserving the module import structure and renaming the outputs to .mjs so that the basic standard import commands can work


asturur:
Need to hook it up to a real app to test differences, but seems the part i was missing in my head.

asturur:
@ShaMan123 i have to test this, but this would allow to do something like:

```ts
import { Canvas, Text } from 'fabric/es'
```

And then your bundler would need to bundle only the stuff that is really necessary for the 2.

It wouldn't be as byte optimized as a tree shaking software that goes around the bundled build and chunks off pieces of code that are not needed, but in general would work

I have a small app in development so maybe tomorrow i m able to test this

asturur:
This is fabricJS with the minified es build
image

While if i declare sideEffects false and load the /es/ imports:

image

i get an entirely different  import size

asturur:
Now this can't be the default of course because if i m using either svgImport or loadCanvasFromJson with an app that requires a shape, like Path that i m not requiring in my code, that is going to break the app.

So this this /es/ folder should be kept for experimentation or 'i know what i m doing' for now and node should be left to a bundled output

ShaMan123 (approved):
Don't we want es to become the default export instead of the bundled one?

asturur:
> Don't we want es to become the default export instead of the bundled one?

Yes.
But we don't have a solution/guideline to explain how to work with generic data that you may import.

Probably the solution will be that utils like loadSVG will force import all classes, so that when you load them all the imports are resolved, and loadFromJSON instead of being a canvas method will be a util as well, so that when you trigger generic code you are forced to import all the dependencies.

This could leave freedom to people to not do that anyway by provide a clear UNSAFE_loadSVGFromString import if they know what they want to do.

The approximate goal is to be able to discard large chunks:
- all brushes
- all filters
- text
- canvas with interactivity

ShaMan123:
I would even set a goal to make them subpackages `@fabric/brushes` etc.

package.json
```diff
@@ -145,11 +145,19 @@
       ]
     }
   },
+  "sideEffects": false,
   "exports": {
     ".": {
-      "import": "./dist/index.mjs",
-      "require": "./dist/index.js",
-      "default": "./dist/index.js",
+      "import": "./dist/index.min.mjs",
```
ShaMan123, line 151: should not be minified so it seems

asturur: If there is something wrong with this change open a visible issue, comments on old PR have basically no visibility unless i m combing the emails one by one.

jiayihu: @asturur  The non-minified build is basically never picked up. This should have been done like React, where it picks the correct one based on the ENV:

Screenshot 2024-04-18 at 14 03 02

Forcing the minified build during development is just more painful for debugging. You can also reuse the same strategy to strip fabric's logs for the production build.

asturur: important is that production has the minified build, if you want top open a change to surface the non minified build in dev you can do it.
Surfacing non minified code in production seems as bad as minified in dev.

Regarding the affected test, what snapshot is it about?

jiayihu: Classes get renamed in the minified build so that affects Jest snapshots that included those classes, like Point. 

I’ll open a PR for the production build then.

jiayihu: Anyway if the non production build ends in production I wonder if it means that you have an issue with the app bundler, as I think node modules are typically distributed unminified, then the bundler can transpile and minify them in production. For instance one may want to ship as ES5 and there are as well different ways to minify.
But since React adopts the minified build directly on production I guess that’s the standard.

asturur: lodash, redux, ships unminified, lottie player ships unminified. Those are random picks i made.
Important is that there is something for the browser inclusion that is still minified.

jiayihu: On a look at `rollup.config.mjs`, I don't know exactly who the fabric build system works so I'll leave the PR to you guys if you want to do it. Otherwise no problem, we can wait if someone else has issues. Internally at my workplace we've just patched `package.json` to continue using the unminified build.

asturur: But do sourcemaps works for you or not?

asturur: I opened this PR because my VITE build was picking up the unminified code, so vite wasn't minifying anything for me

PR #9649, opened 2024-02-04
fix(Group, ActiveSelection): regression, wrong bounding box position when activeSelection or Group has originX/originY that are not default left/top

### LINKED ISSUE #9640: [Bug]: group select 2 rects as ActiveSelection which custom the originX and originY as 'center', but the ActiveSelection shift bboxLeftTop as center.
### CheckList

### Version

6.0.0-beta18

Microsoft Edge

None

### Link To Reproduction

https://codepen.io/ritali4912/pen/QWYKVpo

### Steps To Reproduce

I custom the **ActiveSelection default originX and originY as 'center'** through update classRegistry.
There are 2 rects with originX and originY as 'center'.
rect1: {left: 150, top: 150, width: 60, height: 60, originX: 'center, originY: 'center'}
rect2: {left: 250, top: 250, width: 60, height: 60, originX: 'center, originY: 'center'}

In the begin, 
image

After I group select them though press left mouse and move, **the position changed.**
image

But if I rollback the ActiveSelection originX and originY as **'left' and 'top'**
It is correct.
image

BTW, with custom ActiveSelection originX and originY to 'center'.
If I generate the ActiveSelection with step:
1. Click rect1
2. press 'Shift' and click rect2
3. I got correct position of the ActiveSelection.

### Expected Behavior

The ActiveSelection has correct position even I custom originX and originY to '**center**'

### Actual Behavior

The ActiveSelection has wrong location if I custom originX and originY to '**center**'

When I debugging I found in the LayoutStrategy.cs->calcBoundingBox, it got correct bboxCenter, but then after translate with originFactor, it got wrong center.

image

_No response_

asturur:
## Description

close #9640

I can't find the exact commit because testing the old code is complicated especially when we go back to the first group patch. 

It seems to me that the target origin is already taken in consideration in the  method 'commitLayout' when setting top and left of the active selection, and that his one internal here is a duplicate of that operation.

When we do:

```ts
const center = bboxLeftTop.add(bboxSize.multiply(originFactor));
```
We need to consider that:
bboxLeftTop is already top/left independently of the origin of the objects, that has been already ammortized in the previous function. originFactor is 0,0 when the the target origin is set to 'center', so this is moving the center to bboxLeftTop.

The center of the bounding box, regardless of objects origin seems to be 'bboxCenter' as calculated in this same function, and so we can trust that.

I built a test case in the vanila template to quickly test different combinations, but i would need your help @ShaMan123 because i m not 100% confident of why every single operation is there


src/LayoutManager/LayoutStrategies/LayoutStrategy.ts
```diff
@@ -89,19 +88,13 @@ export abstract class LayoutStrategy {
...
       });
-      const originFactor = new Point(
-        -resolveOrigin(target.originX),
-        -resolveOrigin(target.originY)
-      );
-      const sizeCorrection = actualSize
-        .subtract(bboxSize)
-        .multiply(originFactor);
-      // translate the layout origin from left top to target's origin
-      const center = bboxLeftTop.add(bboxSize.multiply(originFactor));
```
asturur, line 100: Seems to me this is accounted for here:
https://github.com/fabricjs/fabric.js/blob/0547c05dbb68ebe8437f6bb9783f393d463d0721/src/LayoutManager/LayoutManager.ts#L207
and some line lower for the case of non initialization

asturur:
Ok all layout visual tests are failing at least i have something to check why is failing

asturur:
So far i think the visual snapshot are bad indeed:

This is a top,left no rotation
image

this a bottom right no rotation
image

When the group is created by the test code:

```ts
    function createObjectsForOriginTests(originX, originY, options) {
        var rect1 = new fabric.Rect({ left: 150, top: 100, width: 30, height: 10, strokeWidth: 0 }),
            rect2 = new fabric.Rect({ left: 200, top: 120, width: 10, height: 40, strokeWidth: 0 }),
            controlPoint = new fabric.Circle({ radius: 5, fill: 'blue', left: 150, top: 100, originX: 'center', originY: 'center' }),
            tlControlPoint = new fabric.Circle({ radius: 5, fill: 'red', left: 150, top: 100, strokeWidth: 0 });

        var g = new fabric.Group([rect1, rect2, tlControlPoint], Object.assign({}, options, {
            originX, originY, backgroundColor: 'pink'
        }));
        return [g, controlPoint];
    }
```
It applies the originX/originY to the group at creation.
When i create a group around 2 objects positioned in the canvas i expect the group to wrap them correctly, in the same position, and then have top and left adjust to respect the originX/Y that the group has.

I didn't check the cases with strategy different from fit-content yet.

asturur:
I understand how this is not consistent with the other properties like angle or scasling,  since if i create a group with those already set they taken in consideration during layout and so the group will be bigger than the objects or the objects will appear rotated.

But originX/originY are kind of worse because if you don't adapt for those every group that is not top/left will be always moved by an offset compared to the original objects.

ShaMan123:
I can try to help
The initial description sounds possible
The following comments are hard to understand quickly

ShaMan123:
I don't understand why the test snapshot you provided are wrong

asturur:
> I don't understand why the test snapshot you provided are wrong

@shaman in the snapshot the  end result of the group is moved on the canvas depending the group orginX/originY.
But when i create a group around N objects, the group shouldn't move the objects around. The initial position of the group should match the one of the objects

asturur:
simplifed example, if i have 2 rects at 0,0 and 100,100 both with a 100 size,
I have a 2 rects starting from 0,0 and ending at 200,200 in the top left corner of my canvas.
When i wrap them with a group i expect the group to be in the top left corner of my canvas leaving the objects unmoved.
If i create the group with origin bottom right, the canvas group be outside view pushed out the top left corner of the canvas into the non visible area of the screen.
I don't think that was the behaviour before the layout manager ( for sure not in version 5 ).
I can run the example with a commit before the layout manager

asturur:
If i run this code in the old fabricJS:

```
canvas.clear();

const rect1 = new fabric.Rect({ width: 100, height: 100, originX:'left', originY: 'top' });

const rect2 = new fabric.Rect({ width: 100, height: 100, originX:'right', originY: 'center', left: 200, top: 150 });


canvas.add(rect1, rect2)

const g = new fabric.Group([rect1, rect2], { originX: -3, originY: 3 });

canvas.add(g);
```

The group stays put when i add it

image

And this wasn't supposed to change imho.

asturur:
While if i specify a position that overrides the result of my group and is calculated with the origin

```
canvas.clear();

const rect1 = new fabric.Rect({ width: 100, height: 100, originX:'left', originY: 'top' });

const rect2 = new fabric.Rect({ width: 100, height: 100, originX:'right', originY: 'center', left: 200, top: 150 });


const g = new fabric.Group([rect1, rect2], { originX: 0.5, originY: 0.5, top: 10, left: 10 });

canvas.add(g);
```

image

ShaMan123:
This is a refinement of the LayoutStrategies.
And IMO needs a face to face.

asturur:
@ShaMan123 is useful to have some comment on this before tomorrow

asturur:
@ShaMan123 remember to keep some time to gets your thougts on this. Mine so far are:
- apply the fix
- update the snapshots

src/LayoutManager/LayoutManager.ts
```diff
@@ -259,10 +259,8 @@ export class LayoutManager {
     { offset }: Required<LayoutResult>,
     object: FabricObject
   ) {
-    object.set({
-      left: object.left + offset.x,
-      top: object.top + offset.y,
-    });
+    object.left += offset.x;
+    object.top += offset.y;
```
ShaMan123, line 263: This may seem to you the same but it may not be
`set` invalidates the cache
If the group is nested and the layout doesn't bubble I think you will hit the case you need it
Am I right?

ShaMan123:
Did you manage to reproduce the bug? Can you make an e2e?
The test has become worthless so or you fix it by passing left and top 0 or remove it

ShaMan123:
Also since there is no repro I am not entirely convinced...
The fact is that the without passing left and top you can pass originX/Y
How does that behave? 
Since you are planning the removal of originX/Y maybe it is better to wait with this and do that first.
Then perhaps all this effort is futile

ShaMan123:
So if I have veto I would like to use it here

ShaMan123:
Meanwhile I fixed the test so it works as before by passing left/top

asturur:
@ShaMan123 can you explain what you want to do with the test?
Screenshot changes and test remains the same.
If we want to test more i can add another test.

What is the veto?

asturur:
> Did you manage to reproduce the bug? Can you make an e2e? The test has become worthless so or you fix it by passing left and top or remove it

The e2e is not necessary, because the bug is in the group creation, the one that the visual tests are catching.
I found other bugs in the clipPath layout present in this PR and identical in master that i m writing down and documenting but i m not going deep into fixing now

src/LayoutManager/LayoutManager.ts
asturur: So why do we have specific set dirty in the layout manager? Doesn't that cover all?

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
...
+
+### Clippath layout strategy
+
+The Clippath layout keeps the group around the clipath rather than the objects. The group will be as large as the clipPath, the center of the group will be the center of the clippath.
+This layout is clippath dependentent
+
+WARNING needs fix:
+The behaviour of this layout is currently broken if we want.
+The center of the group is the center of the clipPath visually but the position of the clipPath compared to the object on the canvas is hard to predict as it is now
+
```
asturur, line 37: @ShaMan123  I verified those behaviors in master and in the PR.

The differences are that now originX and originY do not count anymore for group layouting.
So for example in master with the fixed strategy the final group bounding box would be aligned with the top,left  corner or the center or the bottom,right corner depending of the group origin.
Object displacement would happen in any case apart the topleft.

The clipPath has a similar behaviour in master with the clipPath ending not in the center if is not using top,left and object being displaced.

Now the final viewport of the group is always aligned with the center of the initial bounding box, but never displaces anything.
The Clippath has an added behaviour that i do not understand, in particular that if you specify a position for the clippath is not clear how that relates with the objects. but this is out of scope for me to fix now, i m fine to know is known bug and move to other things

asturur:
There isn't more to do in this PR from my sides.
- bug is fixed
- old tests that would fail because of this change have snapshot updated
- new tests that behave differently ( old tests + fixed positions ) have been added
- some readme has been written to put down what is the actual expected functionality of the feature
- behaviour compared to beta is slightly changed for fixed layout
- behaviour compared to beta is slightly changed for clipPath layout and both are a bit of an issue

ShaMan123:
The veto is NOT to merge this PR until you remove originX/Y because this PR is clearly related.
Why did you revert the change I made to `group_layout.js`? What value does that test have know? None.
So if you think it is useless (I think it is useful) remove it. What do you learn or understand from all those misplaced objects?

src/LayoutManager/LayoutManager.spec.ts
```diff
@@ -858,7 +858,7 @@ describe('Layout Manager', () => {
       });
       expect(group).toMatchObject({ width: 100, height: 300 });
       expect(child.getCenterPoint()).toMatchObject({ x: 100, y: 100 });
-      expect(group.getCenterPoint()).toMatchObject({ x: 50, y: 150 });
+      expect(group.getCenterPoint()).toMatchObject({ x: 100, y: 100 });
```
ShaMan123, line 861: this is wrong

ShaMan123: You pass width and height and a fixed layout and no left and top
So the group's tl should be positioned at 0, 0
Anything is is not fixed layout

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
 
 Layout manager exposes a main public method that performs group layout given a target group and target objects.
 
+## Expectations for non interactive groups
+
+The layout manager will calculate the size and position of the group when is created or updated
+The FitContent strategy is the classic fabric 5.x group and will grow or shrink when objects are added or removed.
+In general the expectation is that creating or modifying an object will create a group with a postion and once that group is added to canvas, the position of the objects will be the same visually that those object would have had on the canvas if they weren't on the group.
```
ShaMan123, line 9: This is correct only if left and top are not passed
Could it be you missed that?

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
...
+This is a legacy behaviour that carries over this new code.
+The behaviour is not important if you are building a group from scratch and you are adding it in a specific position on the canvas, in that case the original position is meaningless and you are overriding it by specifying your postion.
+
+Since inside the group all objects are organized around the center is anyway easier if you can think of your objects position and alignment with scene coordinates and know that they will be maintained in the same position when grouped.
```
ShaMan123, line 13:
```suggestion
Since inside the group all objects are organized around the center is anyway easier if you can think of your objects position and alignment with scene coordinates and know that they will be maintained in the same position when grouped only if you did not pass transform options.
```

ShaMan123: This is very misleading

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
...
+
+### FitContent strategy
+
+When a group is created, FitContent strategy will consider all the objects in the group, find their bounding box, find the center of the bounding box, and assign those to the group.
+
+When a new object is added or removed, a new bounding box is calculated, a new center is found, all the remaining objects are moved by an offset so that are re-centered around the new center
```
ShaMan123, line 19:
```suggestion
When a new object is added or removed, a new bounding box is calculated, a new center is found, all the remaining objects are moved by an offset so that are re-positioned around the new center
```

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
...
+
+### Fixed layout strategy
+
+With the fixed layout a Developer can specify either width or height and have the other calculated.
+If the developer specifies both dimensions of the box, the bounding box is still calculated to obtain the center.
+Once the group is initialized with a fixed layout, adding new objects is not changing the center anymore.
```
ShaMan123, line 25:
```suggestion
Once the group is initialized with a fixed layout, adding new objects will not change the group's size and therefore also its position.
```

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
...
+The layout is fixed, is cropped by the bounding box itself.
```
ShaMan123, line 26: partially correct, cropping is not precise from what I recall

src/LayoutManager/README.md
```diff
@@ -2,6 +2,39 @@
...
+When larger or smaller than the actual boundingbox the group is centered on the initial center of the initial set of objects and the fixed bounding box extends from the center.
```
ShaMan123, line 27: wrong
it used to position by the passed origin and that is what you missed

src/LayoutManager/README.md
ShaMan123: What you describe is correct and proves to me that this PR is wrong
And that is why I want to block it until you remove originX/Y

ShaMan123 (changes requested):
not approving this change

src/LayoutManager/LayoutManager.spec.ts
asturur: i updated the readme since there was no explanation.
The group should not move the object that is wrapping unless a position is passed.
This is how it was before even the group rewrite.
If you artificially crafted a group that had smaller bbox than the objects, the bounding box would be centered around the group center cutting equally from all the sides.

The same issue here with origin, if you specify for this group center/center, it won't be positioned at 0,0 and also the object will be misplaced.

src/LayoutManager/README.md
asturur: is written below indeed in the paragraph about position.

src/LayoutManager/README.md
asturur: that is not a feature. we don't do that.
originX and originY represent the anchor point of the object that is located at top/left. They don't represent anything else. They can't also be used to say that since a bounding box is smaller the bbox left/top corner will be aligned wit the top/left corner of the elements, or in another case bottom/right. That is not originX/originY.

asturur:
> What you describe is correct and proves to me that this PR is wrong
And that is why I want to block it until you remove originX/Y

That behaviour is broken in master and in this PR in the same way.

asturur:
Your test has been moved and not reverted because is not correct to change the code and the test together.
The objects are misplaced, the snapshots show misplaced objects.
The code gets fixed the snapshot gets updated.
Then i took your will of having also a test with fixed position and i replicated the test few lines below, adding new snapshots.
So nothing has been deleted.

src/LayoutManager/README.md
ShaMan123: confusing

src/LayoutManager/README.md
ShaMan123: I wrote group layout to fix v5 behavior not to organize it
Centering it makes no sense and again this is why this PR should be blocked until you remove originX/Y
If you do not agree with me go ahead
I don't approve this direction.

asturur:
The proposal of removing originX and originY is still a proposal and won't be pushed in the middle of a process already started of closing a release and doing the website, it will have to wait and is not as related as you think.
If i just removed originX and originY from the objects and i would applied the result of a new default center/center to this code i will have indeed all misplaced objects as i have today when choosing center/center.
What this pr originated is the sensation that maybe originX and originY are too complicated and of little value given that all the functionalities can stay in place without those. But i m not defocusing to remove them now.

ShaMan123:
You say let's plan.
Now I look at this PR and think to myself "It is not planned".
I won't say more than I did only consider that you get to block any of my PRs.
I don't have such power or right. Now understand my frustration on both ends.
How it feels someone saying to you stop and how it feels not being able to say that.

asturur:
> You say let's plan. Now I look at this PR and think to myself "It is not planned". I won't say more than I did only consider that you get to block any of my PRs. I don't have such power or right. Now understand my frustration on both ends. How it feels someone saying to you stop and how it feels not being able to say that.

This is not planned no.
No one planned to break groups.
Groups as of now are broken if used without origin top/left and i attached the bug label because is a bug and not to circumvent any planning.
This is a regression from 5.x if you want.

Take your time to understand that or to prove the contrary, is not about frustration or anything.
The layout manager is broken for positioning and for nested active selections.
Nested active selections can be turned off but this no, this needs to be fixed.

src/LayoutManager/README.md
asturur: Well you didn't comunicate that clearly when you organize the group rewrite, at all.
Please try to explain what issue the centering causes apart solving the issue of shifting objects.

asturur:
Keep in mind my time for fabric today is probably finished, i used it to try to refine those PR.
Those 2 bugs of the groups are what stop us from moving forward post 6.0
If you have a different idea on how to solve this explain it.

The bug that is discovered about multi selection in interactive groups and that can be considered an edge case and that is monitored by a jest test in the other pr for me can be fixed later, is not immediate to trigger.

ritali4912:
@ShaMan123 @asturur 
Sorry to put pressure on this PR. 
We have based on beta 13 to develop our product and after upgrade to beta 18, we met this issue. 
We will soon develop Group feature which in our context also with (orignX/orignY: 'center') in 2 weeks, which fabric version we should go ahead? Can we get the fix of this issue?

asturur:
@ritali4912 this PR and the other marked as bug v6 are in theory the last 2 prs to draw a line and call master 6.0 ( or at least release candidate ).
@ShaMan123 has some concern about this change and while i don't understand why i asked him to try to make it explicit since there is a firmly request block on the PR both with the button and with the message.
Let's wait some more time but in a way or another the shifting behaviour of the objects should be fixed as our top priority.

asturur:
@shaman123 Attaching to this [thread](https://github.com/fabricjs/fabric.js/pull/9649#discussion_r1510639541) again trying to clarify this.
The release of 6.0 is holding for those 2 bugs.
I want to be clear about some points that for me are very important.
Group rewrite has been mishandled.
Pr have been accepted without descriptions of intent and without proper understanding.
You didn't say what you wanted to achieve in a clear way or some parts didn't say at all, and i didn't understand them just by reading the code.
There was nothing in particular to fix in the group, there are just features you wanted to add.
Now we are here, group creation is broken and more importantly in your opinion originX and originY should represent both the anchor point of the group and its relative placement over the objects. That is not possible and we need to move forward.

If you don't have a better way to handle this bug or if you have been too buys to handle it in the last 2 weeks i will just have to move forward.

I do not think anyone will miss a feature that was never announced or described anywhere outside that comment i linked.
Are you using this particular feature in any way?

ShaMan123:
This is not a feature.
It was meant to mimic v5 but I got it wrong. I don't remember anyways.
It wasn't easy writing the layout manager and indeed it is not designed in a good manner. I regret a lot of it.
I don't care about originX/Y especially since you plan to remove them.
I wanted to reduce confusion because using fabric is too confusing to begin with and resulted in additional confusion.


> Now we are here, group creation is broken and more importantly in your opinion originX and originY should represent both the anchor point of the group and its relative placement over the objects. That is not possible and we need to move forward.

Incorrect.


This PR seems fine but I do not understand why you are keeping the old tests, they are wrong as you pointed out and are useless. Keep them if that is important for you.

.codesandbox/templates/vanilla/src/testcases/activeSelectionOrigin.ts
ShaMan123, line 1: shouldn't this be turned into a playwright test?

src/LayoutManager/LayoutManager.ts
ShaMan123: Correct

asturur:
> This is not a feature. It was meant to mimic v5 but I got it wrong. I don't remember anyways. It wasn't easy writing the layout manager and indeed it is not designed in a good manner. I regret a lot of it. I don't care about originX/Y especially since you plan to remove them. I wanted to reduce confusion because using fabric is too confusing to begin with and resulted in additional confusion.
> 
> > Now we are here, group creation is broken and more importantly in your opinion originX and originY should represent both the anchor point of the group and its relative placement over the objects. That is not possible and we need to move forward.
> 
> Incorrect.
> 
> This PR seems fine but I do not understand why you are keeping the old tests, they are wrong as you pointed out and are useless. Keep them if that is important for you.

I m keeping old and new tests because in this situation only the combination of both cover the all possible combinations of properties of the group, only with both we an successfully catch an unwanted regression from a code change that seems safe.
I'm keeping originX and originY, while the only thing i planned is to present a proposal to remove them, because we said that the scope of the migration was TS and group and if i put that in the plans now another undefined effort starts. 
Now 2 years are passed we may like or not like the Group code as it is today, this is what we were able to accomplish in 2 years and that's it, it will work in some way, and for future changes we will have to do better.

Issue #9600, opened 2024-01-14
[bug/feature?] nested multi selections


asturur:
Looking at the nestedSelection use case

![image](https://github.com/fabricjs/fabric.js/assets/1194048/09bf0757-423a-4b6d-b2fe-0a9c9d7d760b)

as of now an active selection that span across groups can't update the groups themselves.
The update works only when the multi selection is dissolved.

One of the reasons is that the events that the layoutManager subscribe the objects to moving / scaling events but those do not fire where the activeSelection is being manipulated.

objectCaching or not doesn't change.

asturur:
In this screenshot i moved the 2 black circles but they are not updated

<img width="485" alt="image" src="https://github.com/fabricjs/fabric.js/assets/1194048/ed12c170-915b-4022-a4e9-9f5258ab3c40">

ShaMan123:
So it is caching,
The only thing needed is to update the dirty flagging in `FabricObject#_set`
It checks for `group` but should check for `parent` also

asturur:
i m not sure i quickly checked with objectCaching false and did not work either.
I will look on it later to undertand better, right now i m trying to preserve old code pr/requests that are based on old js code

ShaMan123:
So maybe it is because group rendering skips selected objects

asturur:
Also with nested feature look at correctness of https://github.com/fabricjs/fabric.js/pull/7893 toObject export

asturur:
that is why i tried right away ( is buried in a commit i did by error while capturing the test case )

```
One of the reasons is that the events that the layoutManager subscribe the objects to moving / scaling events but those do not fire where the activeSelection is being manipulated.
```

This was my opening consideration. active selection scaling does not trigger an object scaling.
But anyway i ll get to it, i want to use this issue as a way to understand better the mechanism and relationship

PR #9651, opened 2024-02-05
feat(LayoutManager): Handle the case of activeSelection with objects inside different groups

asturur:
## Description

This PR is supposed to help with the issue described in #9600

The strategy tried here is to:
- when the active selection is created, it subscribes its own objects.
- when this subscription happens detect if those objects have a parent that is not the active selection
- when that happens subscribe the group ( so the active selection ) on the parent layout manager ( the original group ) to perform a performLayout with target the the parent.

This should allow for N groups to subscribe the same active selection to modify perform layout on the groups.

The test case in the vanilla template is a good way to play around with those changes


asturur:
Ok this reproduction case uses clone, so it shared the bug with the objects not being registered.
So we need to solve that first.

asturur:
The bug was split around 2 locations:
- missing registration of active selection events
- missing dirty flag for layout strategies that do not require a perform layout

I have to check the jest failures, but this fixes the group interactivity

asturur:
This PR requires an end to end test

ShaMan123:
I would like to review this thoroughly when I find the time

asturur:
This could be probably done differently by defining an active selection strategy, but the registration part is in the layout manager and i don't want to flip more things right now.
But this is not the only way to fix this bug or the best one

asturur:
> I would like to review this thoroughly when I find the time

Yes but please when you have time for fabricJS repo, do this first, we have still 2 bugs in the layout manager that block us from drawing a line and cutting 6.0

src/LayoutManager/LayoutManager.ts
```diff
@@ -130,11 +142,27 @@ export class LayoutManager {
...
+        if (parent) {
+          // we may subscribe an active selection more than once,
+          // on each parent's layout manager.
+          // each parent's can subscribe itself just once because of unsubscribe
+          parent.layoutManager.subscribe(group!, {
+            ...context,
+            target: parent,
+          });
+        }
+      });
```
asturur, line 165: @shaman this is fix 1.

src/LayoutManager/LayoutManager.ts
```diff
@@ -55,7 +65,18 @@ export class LayoutManager {
...
+    } else {
+      // if there is no layout result, we may still need to flag the group as dirty
+      // if we are moving it through an active selection
+      if (
+        context.type === LAYOUT_TYPE_OBJECT_MODIFIED ||
+        context.type === LAYOUT_TYPE_OBJECT_MODIFYING
+      ) {
+        context.target.set('dirty', true);
+      }
+    }
```
asturur, line 79: @ShaMan123 this is fix 2, for the group that weren't updating visually.

src/LayoutManager/LayoutManager.ts
```diff
@@ -72,35 +93,26 @@ export class LayoutManager {
...
-          'modifyPoly',
-        ] as TModificationEvents[]
-      ).map((key) =>
-        object.on(key, (e) =>
-          this.performLayout({
+      ...layoutingEvents.map((key) =>
+        object.on(key, (e) => {
+          target.layoutManager.performLayout({
             trigger: key,
             e: { ...e, target: object },
```
asturur, line 107: target here should be outside of e and called something like 'triggerSource' or something on that way.
Usually what we call e is the event that generated the call to the callback, and in this case target isn't part of it.
So adding it there is a bit confusing. I understand we want to comunicate to the dev which object moved and triggered a layout, but that can be done with the specific name.

out of scope for this pr

asturur:
> This could be probably done differently by defining an active selection strategy, but the registration part is in the layout manager and i don't want to flip more things right now. But this is not the only way to fix this bug or the best one

On top of that it would be just smarter if when creating an active selection we would understand is pointless to register for events its own children and just register the activeSelection itself.
But that would require a different layoutManager or moving the registering part to the stratetgy. ( that could be done easily ).

e2e/tests/selection/multi-selection-in-groups/index.ts
```diff
@@ -0,0 +1,76 @@
+/**
+ * Runs in the **BROWSER**
+ * Imports are defined in 'e2e/imports.ts'
+ */
+
+import { Rect, Circle, Group, ClipPathLayout } from 'fabric';
+import { beforeAll } from '../../test';
+
+beforeAll(async (canvas) => {
+  canvas.setDimensions({ width: 700, height: 700 });
```
ShaMan123, line 10: I think you should reduce the size so the test has less whitespace

e2e/tests/selection/multi-selection-in-groups/index.ts
ShaMan123, line 1: Great test

src/LayoutManager/LayoutManager.ts
ShaMan123: Sounds correct

src/LayoutManager/LayoutManager.ts
```diff
@@ -130,11 +142,27 @@ export class LayoutManager {
     const { target } = context;
     const { canvas } = target;
     // handle layout triggers subscription
+    // @TODO: gate the registration when the group is interactive
```
ShaMan123, line 145: This is redundant IMO since if the group is not interactive nothing will trigger imperative layout because the user will not be able to select and transform a nested object.
It does subscribe the objects with no need but if the group becomes interactive we save the dev from a gotcha needing to subscribe the targets when toggling interactivity

src/LayoutManager/LayoutManager.ts
ShaMan123: I think we may want to subclass an ActiveSelectionLayoutManager for this, what do you think?

ShaMan123: OR we can subscribe all groups to the canvas level events and check if the target is the active selection or a child.
However it means we need the canvas ref to begin with and we don't in the initialization trigger

ShaMan123: So a different approach would be to have an ActiveSelectionLayoutManager and have it trigger layout on the parents when it triggers its own layout.

src/LayoutManager/LayoutManager.ts
```diff
@@ -130,11 +142,27 @@ export class LayoutManager {
...
     ) {
-      this.subscribeTargets(context);
+      const withDifferentParent = context.targets.filter((object) =>
+        this.subscribe(object, context)
+      );
+      // if objects have a different parent from group, they are in an active seleciton
+      // so we need to subscribe the active selection event to trigger the parent performLayout
+      withDifferentParent.forEach(({ group, parent }) => {
+        if (parent) {
+          // we may subscribe an active selection more than once,
```
ShaMan123, line 157: this doesn't sound good, bad for perf and for logic

src/LayoutManager/LayoutManager.ts
ShaMan123: I am not sure I understand why this is needed
What is the case in which there is no layout result?
The subscription will trigger the layout and it should have a result because the child transformed.

src/LayoutManager/LayoutStrategies/LayoutStrategy.ts
```diff
@@ -27,6 +27,8 @@ export abstract class LayoutStrategy {
 
   /**
    * Used by the `LayoutManager` to perform layout
+   * @TODO/fix: if this method is calcResult, should calc unconditionally.
```
ShaMan123, line 30: interesting
please elaborate

src/shapes/Group.ts
```diff
@@ -92,7 +92,12 @@ export class Group
    * Used to allow targeting of object inside groups.
    * set to true if you want to select an object inside a group.\
    * **REQUIRES** `subTargetCheck` set to true
+   * This will be not removed but slowly replaced with a method setInteractive
+   * that will take care of enabling subTargetCheck and necessary object events.
+   * There is too much attached to group interactivity to just be evaluated by a
```
ShaMan123, line 97: I disagree
IMO there is no need
In some open PR I made `interactive` include logically `subTargetCheck` because that is how we use it so the code should reflect it.
It makes it easier to deal with and restores `subTargetCheck` to what it originally was.
So the dev can flag `interactive` and that is enough

src/shapes/Group.ts
```diff
@@ -300,13 +308,14 @@ export class Group
...
-      this._activeObjects.push(object);
+      activeObjects.push(object);
       this._set('dirty', true);
-    } else if (this._activeObjects.length > 0) {
-      const index = this._activeObjects.indexOf(object);
+    } else if (activeObjects.length > 0) {
+      const index = activeObjects.indexOf(object);
       if (index > -1) {
-        this._activeObjects.splice(index, 1);
+        activeObjects.splice(index, 1);
```
ShaMan123, line 318: I am not sure this even used

ShaMan123 (commented):
See comment
The biggest question to answer is: Do you want to create a subclass `ActiveSelectionLayoutManger`?
It is a design question.
I think it makes it easier to understand
What do you think?

src/LayoutManager/LayoutManager.ts
asturur: I can check if it creates overall less code.
If it creates more code, no.

asturur: I ll try to test this before tomorrow so we can check the end result together

src/LayoutManager/LayoutManager.ts
asturur: not necessarly, indeed it doesn't have.
The clipPath strategy for example rarely does the layout.
And also the clipPath strategy does not need to redo a layout when you move an object around

e2e/tests/selection/multi-selection-in-groups/index.ts
asturur: i think it covers most of the surface, but yes i can shrink to the exact needed

ShaMan123:
> > This could be probably done differently by defining an active selection strategy, but the registration part is in the layout manager and i don't want to flip more things right now. But this is not the only way to fix this bug or the best one
> 
> On top of that it would be just smarter if when creating an active selection we would understand is pointless to register for events its own children and just register the activeSelection itself. But that would require a different layoutManager or moving the registering part to the stratetgy. ( that could be done easily ).

I am not against

src/LayoutManager/LayoutManager.ts
asturur: This is in contrast with wanting to register all the objects all the time.
Right now if an active selection has 2 objects in the same group i think will try to register twice, but trying a different approach my fix that.

src/shapes/Group.ts
asturur: There is no need to register events for objects that will never move or be target of actions.
Group interactivity was put in as a part of the layout manager while it has nothing to do with layouting, nor layouting should take care of attaching event for detecting movements.
Those 2 parts now are strongly bound together and i can't think of my layout logic without having to think of the interaction and registration part.

Registering everything all the time doesn't bring performance issues, but is also completely unnecessary.

src/shapes/Group.ts
asturur: i just wanted to save some bytes with minification. I think is used for rendering and moving active selections

asturur:
@ShaMan123 this is the version with the subclassed ActiveSelectionLayoutManager.
https://github.com/fabricjs/fabric.js/pull/9687

Seems leaner but of course it isn't.
But is an option

asturur:
Oh well is an option only at the cost of making subscribe public, because we need to register the event on the parent layout manager while the activeSelection won't have any real registration.
Not sure how cleanup will work then.

asturur:
There is still more work to do to have something that looks readable and is robust.

asturur:
Ok this should be ready to go now.
I managed to condense the logic to a single override that can't be broken by making unrelated changes to the LayoutManager, and tests are covering its functionality in detail.

Code size is larger than the initial approach but there is no workaround for that.
There is still a possibility to reduce the size of the object registration by merging modified and modifying, but it is out of scope for this that is a fix of a bug.
I tried and requires some clever typescripting

e2e/tests/selection/multi-selection-in-groups/index.ts
asturur: reduced

src/LayoutManager/LayoutManager.ts
ShaMan123: I still think it is not the LayoutManager's responsibility to do this part...

src/LayoutManager/LayoutManager.ts
ShaMan123: > As a counter argument interactive group is an edge case, for now we don't need to go to the extreme that objects get registered for handlers only when they get selected or when a trasform action starts, but at least when the group is interactive yes.

This is not bad

src/LayoutManager/LayoutManager.ts
ShaMan123: and anyways it belongs to ActiveSelectionLayoutManager

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -0,0 +1,27 @@
...
+    childObject: FabricObject,
+    context: RegistrationContext & Partial<StrictLayoutContext>
+  ): (() => void)[] {
+    const { parent, group: activeSelection } = childObject;
+    if (!parent || !activeSelection || activeSelection === parent) {
+      // nothing to do here
+      return [];
+    }
+    return buildStandardEvents(activeSelection, parent);
+  }
```
ShaMan123, line 26: this is great!

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -0,0 +1,27 @@
+import { LayoutManager, buildStandardEvents } from './LayoutManager';
+import type { RegistrationContext, StrictLayoutContext } from './types';
+import type { FabricObject } from '../shapes/Object/FabricObject';
+
+/**
+ * The ActiveSelectionLayoutManager is exactly as the LayoutManager
+ * But what it does differently is that it subscribes the activeSelection
+ * modification events to modify the layout of the original group of the
+ * children. This is done by swapping target and object in the subscribeTargets
```
ShaMan123, line 9: > This is done by swapping target and object in the subscribeTargets

IMO this sentence is more confusing than helpful, especially for the dev

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -0,0 +1,27 @@
...
+ * method. All the rest is identical
+ */
+export class ActiveSelectionLayoutManager extends LayoutManager {
+  /**
+   * subscribe to object layout triggers
```
ShaMan123, line 14: I would write here a better explanation

src/LayoutManager/LayoutManager.ts
```diff
@@ -30,6 +30,40 @@ export type SerializedLayoutManager = {
   strategy: string;
 };
 
+export const buildStandardEvents = (object: FabricObject, target: Group) => {
```
ShaMan123, line 33: shouldn't this method be part of the class??
why did you decide not to have there?

src/LayoutManager/LayoutManager.ts
ShaMan123: > The clipPath strategy for example rarely does the layout.
> And also the clipPath strategy does not need to redo a layout when you move an object around

Are you sure about this? An object moving changes the size, hence changes the center, hence changes the clip path position

I am not sure anymore.

IMO the correct approach is to delegate this responsibility to the ActiveSelection.
Consider the following. I have a multi nested active selection. I use `activeSelection.set` to change some transform prop (angle, scale).
The Layout API expects the dev to call performLayout at this point (a different discussion if it should trigger by itself).
If the dev doesn't call `triggerLayout` and groups are caching visuals will be wrong regardless of layout. Consider a object that was scaled now covers its sibling in the unchanged group area.

src/LayoutManager/LayoutManager.ts
ShaMan123: However toggling interactivity... I am not sure. There is another PR I did which I mentioned in this PR regarding it
I will find it

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        const manager = new ActiveSelectionLayoutManager();
+        const object = new FabricObject();
+        const group = new Group([object], {
+          interactive: true,
+          subTargetCheck: true,
+        });
+        const as = new ActiveSelection([object], { layoutManager: manager });
+        const asPerformLayout = jest.spyOn(manager, 'performLayout');
+        const groupPerformLayout = jest.spyOn(
+          object.parent!.layoutManager,
```
ShaMan123, line 52:
```suggestion
          group,
```

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        const asOn = jest.spyOn(as, 'on');
+        manager.subscribeTargets({
+          targets: [object],
+          target: as,
+        });
+        expect(objectOn).not.toHaveBeenCalled();
+        expect(objectOff).not.toHaveBeenCalled();
+        expect(asOn).toHaveBeenCalledTimes(triggers.length);
+        expect(objectOff).not.toHaveBeenCalled();
+      });
```
ShaMan123, line 40: This test is missing assertion checking double registration of objects with the same parent not registering the AS twice

ShaMan123 (commented):
I reviewed the code.
The only thing that I think needs to change is the dirty flagging.

src/LayoutManager/ActiveSelectionLayoutManager.ts
asturur: What would you write?

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
asturur: I m not sure with this new code structure they will register once only

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
asturur: is not group.performLayout that is called, but the layoutManager method is called directly

asturur:
action items for me:
- verify double registration
- verify why build standard events is not part of the class ( maybe minification issues ? )
- see if i can delegate dirty flag to something else

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -1,27 +1,57 @@
...
+ * The standard LayoutManager would subscribe the children of the activeSelection to
+ * perform layout changes to the activeselection itself, what we need instead is that
+ * the transformation applied to the active selection will trigger changes to the
+ * original group of the children ( the one referenced under the parent property )
+ * This subclass of the LayoutManager has a single duty to fill the gap of this difference.
  */
 export class ActiveSelectionLayoutManager extends LayoutManager {
   /**
-   * subscribe to object layout triggers
+   * Subscribe an object to transforms events that will trigger a layout change in the group
```
ShaMan123, line 19:
```suggestion
   * Subscribe an object to transform events that will trigger a layout change on the parent
```

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -1,27 +1,57 @@
...
+    const { parent, group: activeSelection } = object;
     if (!parent || !activeSelection || activeSelection === parent) {
       // nothing to do here
-      return [];
+      return;
+    }
+    this.unsubscribe(object, context);
+    const disposers = this.attachHandlers(activeSelection, {
+      target: parent,
+      targets: [object],
```
ShaMan123, line 36: I don't like this part.
It means the context of the trigger is incomplete. There might be more objects from this parent

src/LayoutManager/LayoutManager.ts
```diff
@@ -106,16 +72,56 @@ export class LayoutManager {
...
+  /**
+   * Attach event handlers to actions we know are layout invalidating when
+   * performed on child objects. returns the disposers for later unsubscribing
+   * @param {FabricObject} childObject
+   * @param {RegistrationContext & Partial<StrictLayoutContext>} context
+   * @returns {VoidFunction[]} disposers remove the handlers
+   */
   protected attachHandlers(
     childObject: FabricObject,
     context: RegistrationContext & Partial<StrictLayoutContext>
```
ShaMan123, line 84: This type doesn't make sense
The point of StrictLayoutContext is that it is not partial IIRC

src/LayoutManager/LayoutManager.ts
```diff
@@ -106,16 +72,56 @@ export class LayoutManager {
...
+            target,
+          })
+        )
+      ),
+    ];
   }
 
   /**
-   * subscribe to object layout triggers
+   * Subscribe an object to transforms events that will trigger a layout change in the group
```
ShaMan123, line 121:
```suggestion
   * Subscribe an object to transform events that will trigger a layout change on the group
```

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -1,27 +1,57 @@
...
+    });
+    // subscriptions are on the parent, so the active seleciton
+    // will keep has many copies of events has many unique groups
+    // is interfering with
```
ShaMan123, line 40: Really hard to understand
Please rephrase or remove

src/LayoutManager/ActiveSelectionLayoutManager.ts
ShaMan123: Looking at `attachHandlers` I don't see it requires `targets`
It used to but you changed it, right?
I do think it is good to have all the context available for all calls but it must be the correct one. Better none than a wrong one

src/LayoutManager/LayoutManager.ts
```diff
@@ -106,16 +72,56 @@ export class LayoutManager {
     this._prevLayoutStrategy = strictContext.strategy;
   }
 
+  /**
+   * Attach event handlers to actions we know are layout invalidating when
+   * performed on child objects. returns the disposers for later unsubscribing
+   * @param {FabricObject} childObject
+   * @param {RegistrationContext & Partial<StrictLayoutContext>} context
+   * @returns {VoidFunction[]} disposers remove the handlers
```
ShaMan123, line 80:
```suggestion
   * Attach event handlers on child object triggers that invalidate layout
   * @param {FabricObject} childObject
   * @param {RegistrationContext & Partial<StrictLayoutContext>} context
   * @returns {VoidFunction[]} disposers 
```

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -1,27 +1,57 @@
...
-  protected attachHandlers(
-    childObject: FabricObject,
+  protected subscribe(
+    object: FabricObject,
     context: RegistrationContext & Partial<StrictLayoutContext>
-  ): (() => void)[] {
-    const { parent, group: activeSelection } = childObject;
+  ) {
+    const { parent, group: activeSelection } = object;
     if (!parent || !activeSelection || activeSelection === parent) {
```
ShaMan123, line 29: in fact you could skip re-subscribining

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
```diff
@@ -39,17 +39,30 @@ describe('ActiveSelectionLayoutManager', () => {
         expect(objectOff).not.toHaveBeenCalled();
       });
 
-      it('a subscribed activeSelection should trigger layout on the object parent', () => {
+      it('a subscribed activeSelection should trigger layout on the object parent once per parent', () => {
```
ShaMan123, line 42: How does this test really test that?
Seems incomplete to me, am I missing something?

src/LayoutManager/LayoutManager.ts
ShaMan123: Should this happen regardless of ActiveSelection?
In clip path layout?

You said that interactive group and layout are deeply coupled.
What about caching and everything?

ShaMan123 (changes requested):

src/LayoutManager/ActiveSelectionLayoutManager.ts
asturur: i kept it there because you asked for the context to not be removed.

src/LayoutManager/LayoutManager.ts
asturur: `What about caching and everything` does not make sense
@ShaMan123 is not worth.
I think is also better to just do blunt cache invalidation in the method always rather than even using the if since it will be needed anyway.

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
asturur: the _subscription map is a 'new Map()` and so supports one value per key.
Once we test the key is per group and not per object, we know there is only one registration for that group.
It comes intrinsic from the type of key you use.

src/LayoutManager/LayoutManager.ts
ShaMan123: That sounds better to me

ShaMan123: So then we should remove the if and just do it

ShaMan123: Or maybe even do it on `Group#onAfterLayout`

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
ShaMan123: Yes, but the test doesn't assert there aren't floating subscriptions
Or maybe it does actually, in the subscribe tests

src/LayoutManager/ActiveSelectionLayoutManager.ts
asturur: i could but this is not we do in the other group.
This surface another issue anyway.
if i later remove a single object of many in the same parent are remove from the active selection, this will break the group layout.
Meaning that we need to keep multiple events and invalidate more than once.
Not fixing now, at least this works most of the time.

asturur: Added a test to certify the bug exists

src/LayoutManager/LayoutManager.ts
asturur: There is a story for this i need to check

asturur:
@ShaMan123 Please remember of this one too i look forward merging it next week.
If we can't merge this we can disable interactive groups and release anyway.

ShaMan123:
> action items for me:
> 
> * verify double registration
> * verify why build standard events is not part of the class ( maybe minification issues ? )
> * see if i can delegate dirty flag to something else

What is the status?

> see if i can delegate dirty flag to something else

#9741

asturur:
All that stuff has been resolved since the relevant conversation have been Resolved
The dirty flag is now on the afterLayout, there is a test for double registration and the function with the events went back to what was before

src/LayoutManager/LayoutManager.ts
```diff
@@ -231,7 +252,7 @@ export class LayoutManager {
       target.setPositionByOrigin(nextCenter, CENTER, CENTER);
       // invalidate
       target.setCoords();
-      target.set({ dirty: true });
+      context.target.set('dirty', true);
```
ShaMan123, line 255:
```suggestion
     target.set('dirty', true);
```

src/LayoutManager/LayoutManager.ts
```diff
@@ -55,24 +55,31 @@ export class LayoutManager {
...
+   * performed on child objects ( general transforms ).
+   * Returns the disposers for later unsubscribing and cleanup
+   * @param {FabricObject} childObject
+   * @param {RegistrationContext & Partial<StrictLayoutContext>} context
+   * @returns {VoidFunction[]} disposers remove the handlers
    */
-  protected subscribe(
-    object: FabricObject,
+  protected attachHandlers(
+    childObject: FabricObject,
```
ShaMan123, line 75: I think childObject is misleading because it can be the active selection, right? eventTarget perhaps?

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -0,0 +1,56 @@
...
+   */
+  protected unsubscribe(
+    object: FabricObject,
+    context?: RegistrationContext & Partial<StrictLayoutContext>
+  ) {
+    const { parent } = object;
+    if (parent) {
+      (this._subscriptions.get(parent) || []).forEach((d) => d());
+      this._subscriptions.delete(parent);
+    }
```
ShaMan123, line 54:
```suggestion
    parent && super.unsubscribe(parent);
```

src/LayoutManager/ActiveSelectionLayoutManager.ts
ShaMan123, line 1: So this manager is in charge of triggering layout of other managers, right?

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+        asPerformLayout.mockClear();
+        // we don't keep record of subscriptions on objects
+        expect(manager['_subscriptions'].get(object)).toBeUndefined();
+        expect(manager['_subscriptions'].get(object2)).toBeUndefined();
+        expect(manager['_subscriptions'].get(object3)).toBeUndefined();
+        expect(manager['_subscriptions'].get(group)).toBeDefined();
+        as.remove(object3);
+        groupPerformLayout.mockClear();
+        asPerformLayout.mockClear();
+        // BUG! i removed an object only and group layour is not called anymore
```
ShaMan123, line 161: that is because the ActiveSelectionLayoutManager calls unsubscribe, it should check if there are no more children of that parent left

src/LayoutManager/ActiveSelectionLayoutManager.ts
ShaMan123: I am not sure it is the best because a LayoutManager can't control what triggers it anymore by calling unsubscribe.
What if you subscribe the active selection on the group layout manager?

src/LayoutManager/ActiveSelectionLayoutManager.ts
ShaMan123: This is causing the bug
You need to check that there are no more objects left for this parent before unsubscribing it

ShaMan123 (changes requested):
See comments

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
asturur: that is a separate bug will stay there till someone has time to fix it

src/LayoutManager/ActiveSelectionLayoutManager.ts
asturur: This pr is here to fix the nested multiseleciton in groups.
To do that we need to trigger performLayout on groups and you asked to subclass the LayoutManager to do so, so i would argue that yes, as a consequence this LayoutManager needs to trigger other LayoutManagers.

I m not sure why a LayoutManager should be in control to trigger who can call it.
Whoever needs to call the layout manager will call it, the event registration is a way to call it but is not that the layout manager needs to have special powers over it.

```
What if you subscribe the active selection on the group layout manager?
```
I do not understand if this sentence points out a possible problem or is suggesting an alternative.

ShaMan123:
The alternative I suggest is to subscribe the active selection events on the group layout manager from the active selection layout manager subscribe method.
The active selection layout manager will not have any subscriptions of its own.
When unsubscribing the active selection from a group (calling `activeSelectionLayoutManager.unsubscribe(nestedObject)`) we should check there are no more children of that group left in it.
Since unsubscribe is called from subscribe we need to handle that as well to avoid multiple registration in this case. If we check if an object is already subscribed instead of unsubscribing it, that should do the job and is actually better.
It keeps all the triggers of a group's layout on its manager, keeping it in control, easy to unsubscribe at will, no surprises.
Also it resembles how active selection behaves in terms of rendering. It delegates the task to others.

What do you think?

src/LayoutManager/ActiveSelectionLayoutManager.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+    this.unsubscribe(object, context);
+    const disposers = this.attachHandlers(activeSelection, {
+      target: parent,
+      targets: context.targets,
+    });
+    // the _subscriptions Map is using the parent as a key.
+    // This will ensure that a single set of events is kept for each unique parent
+    this._subscriptions.set(parent, disposers);
```
ShaMan123, line 40:
```suggestion
    parent.layoutManager.subscribe(group, context); // context as is?
```

src/LayoutManager/ActiveSelectionLayoutManager.ts
ShaMan123, line 54:
```suggestion
    const { parent, group: activeSelection } = object;
    if (parent && activeSelection && activeSelection.getObjects().every(object => object.parent !== parent)) {
      parent.unsubscribe(activeSelection, context);
    }
```

src/LayoutManager/LayoutManager.ts
```diff
@@ -90,16 +97,30 @@ export class LayoutManager {
...
+   * Subscribe an object to transform events that will trigger a layout change on the parent
+   * This is important only for interactive groups.
+   * @param object
+   * @param context
+   */
+  protected subscribe(
+    object: FabricObject,
+    context: RegistrationContext & Partial<StrictLayoutContext>
+  ) {
+    this.unsubscribe(object, context);
```
ShaMan123, line 122:
```suggestion
    if (this._subscriptions.get(object) {
      return;
    }
```

src/LayoutManager/LayoutManager.ts
```diff
@@ -55,24 +55,31 @@ export class LayoutManager {
...
+  ): (() => void)[] {
     const { target } = context;
-    this.unsubscribe(object, context);
-    const disposers = [
-      object.on('modified', (e) =>
-        this.performLayout({
+    // TODO fix typescript that block us from condesing this to a single call per key.
+    return [
+      childObject.on('modified', (e) =>
+        target.layoutManager.performLayout({
```
ShaMan123, line 82: can revert back to this

src/LayoutManager/LayoutManager.ts
```diff
@@ -90,16 +97,30 @@ export class LayoutManager {
           'modifyPoly',
         ] as TModificationEvents[]
       ).map((key) =>
-        object.on(key, (e) =>
-          this.performLayout({
+        childObject.on(key, (e) =>
+          target.layoutManager.performLayout({
```
ShaMan123, line 101: can revert back to this

ShaMan123 (commented):
This is how I would try it

asturur:
@ShaMan123 change the code as you want.
 i see i don't have any will to move this code further.
For me the first one with the if and the parent//group check was already great, this solution is already a bit more than i want,
This code will deleted at some point anyway.

Push at will and when you are done i ll just check the current test pass

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+        groupPerformLayout2.mockClear();
+        asPerformLayout.mockClear();
+        expect(group.layoutManager['_subscriptions'].get(as)).toBeDefined();
+        expect(group2.layoutManager['_subscriptions'].get(as)).toBeDefined();
+        expect(manager['_subscriptions'].size).toBe(0);
+
+        const event = { foo: 'bar' };
+        triggers.forEach((trigger) => as.fire(trigger, event));
+        expect(asPerformLayout).not.toHaveBeenCalled();
+        expect(groupPerformLayout.mock.calls).toMatchObject([
```
ShaMan123, line 78: I wanted to do the same for `groupPerformLayout2` but it throws a weird error or max call stack exceeded

src/LayoutManager/LayoutManager.ts
```diff
@@ -55,51 +55,71 @@ export class LayoutManager {
...
+
+  /**
+   * Subscribe an object to transform events that will trigger a layout change on the parent
+   * This is important only for interactive groups.
+   * @param object
+   * @param context
    */
   protected subscribe(
     object: FabricObject,
     context: RegistrationContext & Partial<StrictLayoutContext>
```
ShaMan123, line 119: I think we should change the type for all subscription methods
```suggestion
    context: RegistrationContext
```

ShaMan123 (approved):
@asturur please review
Should we extract the subscription logic before the release? It will be breaking. Up to you.

asturur:
> @asturur please review Should we extract the subscription logic before the release? It will be breaking. Up to you.

No. We just declare the LayoutManager as beta and amen. Do not extend it, do not touch it, it will change.

asturur:
How can i see the changes from the latest commit? it gives me that everything was done 50 minutes ago

ShaMan123:
view commits 6aa468c237b0a7b8e1c104861b4c3659ef627da4...HEAD

ShaMan123:
The e2e test is failing because dirty flag is not being set for some reason

asturur:
> The e2e test is failing because dirty flag is not being set for some reason

Because you deleted the catch'all dirty flag. Indeed when reviewing i was thinking why do that

asturur:
I restored it, let's see if is that one.

ShaMan123:
because it didn't make a difference from what I could see

asturur:
> because it didn't make a difference from what I could see

We talked about this, who is going to clear the cache for the groups that do not have a layout result like clipPath or fixed?

Indeed the test pass now, and that cache invalidation was the part 2 of the 2 part fixes i discussed during the PR all the time, active selection events registration was just half of the fix.
Do you remember?

ShaMan123:
yes, weird. I restored it locally and re-ran. Maybe I forgot to build.
In that case we can restore your change to layoutObject if you wish, not using set.
I am good with this anyways.

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
asturur: It happened to me too, and there was a bug in the code in that moment. Maybe is worth double checking

ShaMan123:
Confirmed, all good

src/LayoutManager/LayoutManager.ts
```diff
@@ -231,7 +251,7 @@ export class LayoutManager {
       target.setPositionByOrigin(nextCenter, CENTER, CENTER);
       // invalidate
       target.setCoords();
-      target.set({ dirty: true });
+      target.set('dirty', true);
```
ShaMan123, line 254: this can be removed now because of the one in onAfterLayout

ShaMan123 (approved):

asturur:
Monday i will add some comments on the test because is a bit hard to remember why mock should be called or not and then  i will merge.

src/LayoutManager/ActiveSelectionLayoutManager.spec.ts
ShaMan123: still happening

asturur: Ok i ll look with fresh eyes, maybe there is a situation in which you get an infinite loop.

asturur: no issues for me, it works fine

ShaMan123: I think you should have committed it

asturur: i did i pushed up commits with the checks for groupPerformLayout2 mirrored to groupPerformLayout

asturur: https://github.com/fabricjs/fabric.js/pull/9651/commits/a22cc7c17ee52a35f1d0f98a14d763ada235daf0

ShaMan123: Missed that! Great!

PR #9661, opened 2024-02-10
feat(LayoutManager): Expose objects registration

asturur:
## Description

When we stopped running the layout manager from group that comes from a json object, we removed also the object registration part and that broke interactive groups that won't react when the objects are transformed.

This PR fixes the issue by explicitly registering the objects.

This is a temporary fix, the final goal will probably be to deprecate the interactivity property of the group and add a setInteractive method that will also take care of event registration, since it doesn't seems right to have event handler registered anyway for objects that are non interactive.

There are probably other parts in the transform lifecycle where we can handle this but for now we are registering everything every time.

### DIFF
--- a/src/LayoutManager/LayoutManager.spec.ts
+++ b/src/LayoutManager/LayoutManager.spec.ts
@@ -12,12 +12,7 @@ import {
   LAYOUT_TYPE_INITIALIZATION,
   LAYOUT_TYPE_REMOVED,
 } from './constants';
-import type {
-  LayoutContext,
-  LayoutResult,
-  LayoutTrigger,
-  StrictLayoutContext,
-} from './types';
+import type { LayoutContext, LayoutResult, StrictLayoutContext } from './types';
 
 describe('Layout Manager', () => {
   it('should set fit content strategy by default', () => {
@@ -158,7 +153,7 @@ describe('Layout Manager', () => {
         expect(performLayout.mock.calls).toMatchObject([
           [
             {
-              e: { target: object, ...event },
+              e: { ...event },
               target,
               trigger: 'modified',
               type: 'object_modified',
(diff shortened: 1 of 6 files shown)


src/LayoutManager/LayoutManager.spec.ts
```diff
@@ -182,39 +179,24 @@ describe('Layout Manager', () => {
...
-      { trigger: LAYOUT_TYPE_INITIALIZATION, action: 'subscribe' },
-      { trigger: LAYOUT_TYPE_ADDED, action: 'subscribe' },
-      { trigger: LAYOUT_TYPE_REMOVED, action: 'unsubscribe' },
-    ] as {
-      trigger: LayoutTrigger;
-      action: 'subscribe' | 'unsubscribe';
-    }[])(
-      '$trigger trigger should $action targets and call target hooks',
-      ({ action }) => {
-        const lifecycle: jest.SpyInstance[] = [];
```
asturur, line 196: this 3 tests were failing because i wrote bad code. And it was immediate to me to understand what they were testing for 2 reasons:
The connection between the array of data and the test variables force me to look up and down every time to understandand what is action and what is trigger, and the lifecycle array was a convoluted way to spell 3 times what has been calle, but without clearly saying with what has been called.

In general we should keep those array method for tests either for simple cases [true, false] or for large cases in which we need to use points or something else.

In the other cases i really prefer to spell clearly what the test is doing

src/LayoutManager/LayoutManager.ts
```diff
@@ -59,14 +64,14 @@ export class LayoutManager {
...
-  protected subscribe(object: FabricObject, context: StrictLayoutContext) {
+  protected subscribe(object: FabricObject, context: { target: Group }) {
     const { target } = context;
     this.unsubscribe(object, context);
     const disposers = [
       object.on('modified', (e) =>
         this.performLayout({
           trigger: 'modified',
-          e: { ...e, target: object },
+          e,
```
asturur, line 74: in the object modified, target is already available in the event, we don't need to repeat it. At leat this says typescript

asturur: on line 88 ( now 93 ) we do the same for the other event, but in my opinion we shouldn't. If we want to forward the original event that is fine, but we are adding the target itself in the object event, we don't need to copy it in the forwarded event.

ShaMan123: I remember something about this.
If you wish to do this you should add tests to verify this.
Notice that we add a `target` prop to this object.

ShaMan123 (commented):
What I don't like about this is that is makes the API missing
Who knows what context the dev needs to decide if to subscribe/unsubscribe targets?
I do not want to reduce the amount of context available.
I suffer from it when I work with fabric and then need to do complex overriding to provide a missing bit of context.

asturur:
We can add more context, but the logic should be, we write the minimal code, and we extend if a use case will surface.
Said so, what context are you using right now that you think should be also in subscribe and unsubscribe?

also stopPropagation should have the same treatment. I don't see where are we using it. Either provide an example in some doc/readme of why is there or we need to remove it.

src/LayoutManager/LayoutManager.ts
asturur: I see that there is a target prop to this object, that is why i don't see any reason to duplicate it on the event.
Probably also e should be called 'triggerEventData' or something that can let us know what 'e' is because is not clear. The only purpose of this 'e' is to be found later inside 'layout:after' event, correct? What else?

asturur:
I restored the full context, missing there was just the `type: 'LAYOUT_INITIALIZATION'`

ShaMan123:
> also stopPropagation should have the same treatment. I don't see where are we using it. Either provide an example in some doc/readme of why is there or we need to remove it.

Simple case
If you have a tree and a layout starts bubbling you might want an event handler to be able to stop the bubbling.
Like any DOM event

ShaMan123:
We are discussing a single flag. IMO it is too much time and resources spent on it.

asturur:
> We are discussing a single flag. IMO it is too much time and resources spent on it.

If you don't have time you don't have time. I'll see what i can do to finish the LayoutManager

PR #9668, opened 2024-02-15
refactor(): `_findTargetCorner` now returns the key and the control

jiayihu:
Just like `getActiveControl`, this PR refactors `_findTargetCorner` to return the Control as well. This allows to override the returned control, for instance allowing a Group to return children controls to allow "pass-through" interactions.

Honestly, I believe `Control` should have a `key/name` property. It's very common to need that information and having `getActiveControl` and `_findTargetCorner` return `{ key: string, control: Control }` instead of just `Control` is a clear sign of that.
For the Control itself it's very convenient to know the key as `this.key`, for instance you may have several instances of the same Control class, but with different keys, e.g. resize controls.


asturur:
Control had a name property, at least all my fabric 5.x apps have one.

asturur:
The constructor of Control uses Object.assign, so you can probably just pass in the name and get it in the control.
I wonder if with the interfaces those properties can be added easily without the need of subclassing.

jiayihu:
> The constructor of Control uses Object.assign, so you can probably just pass in the name and get it in the control.

That’s true. Didn’t think about it. 
Anyway if we make it built-in in fabric it means we can drop the object return type in getActiveControl and _findTargerCorner.

ShaMan123:
Question:
Does it make sense for the controls object to be an array?
Then we force the control class to accept a key/name?
IMO it makes sense because control hit detection depends on the control stack order and an array makes that very clear.

jiayihu:
Might be a good idea but out of the scope of this PR. I guess it'd be similar to `findTarget` returning an array of targets in the stack order, something that you tried already IIRC

asturur:
im sure the name was in at some point. I think i removed it because i realized it was duplicate of the key.
If the issue is a name i m glad to add a name at this point rather than changing another function return style.

Changing the return of the function of the internal function is not terrible anyway.
Doing both maybe is unnecessary.

src/shapes/Object/InteractiveObject.ts
```diff
@@ -189,29 +189,36 @@ export class InteractiveFabricObject<
    * @param {boolean} forTouch indicates if we are looking for interaction area with a touch action
    * @return {String|Boolean} corner code (tl, tr, bl, br, etc.), or 0 if nothing is found.
    */
-  _findTargetCorner(pointer: Point, forTouch = false): string {
+  _findTargetCorner(
+    pointer: Point,
+    forTouch = false
+  ): { key: string; control: Control } | undefined {
```
asturur, line 195: this return type changed like 4 times in 2023

src/canvas/Canvas.ts
```diff
@@ -812,19 +812,20 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
       if (
         target.selectable &&
         target !== this._activeObject &&
         target.activeOn === 'up'
       ) {
         this.setActiveObject(target, e);
         shouldRender = true;
-      } else {
-        const control = target.controls[corner];
+      } else if (control) {
```
asturur, line 828: if you add `if control` here, the control check before the getMouseUpHandler is not necessary anymore.

src/canvas/Canvas.ts
```diff
@@ -1055,7 +1056,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       );
       if (target === this._activeObject && (corner || !grouped)) {
         this._setupCurrentTransform(e, target, alreadySelected);
-        const control = target.controls[corner],
+        const control = corner?.control,
```
asturur, line 1059: i hate the question mark because the need of transpiling it. in case can you do `corner && corner.control` ?

src/shapes/Object/InteractiveObject.ts
jiayihu: Sign of bad API. If `name` is re-introduced, the return type could just be `: Control`, which is much better IMO.

jiayihu:
> im sure the name was in at some point. I think i removed it because i realized it was duplicate of the key.
If the issue is a name i m glad to add a name at this point rather than changing another function return style.

The name is not the issue but I'd welcome the change.

> Changing the return of the function of the internal function is not terrible anyway.
Doing both maybe is unnecessary.

The change of returned value is still necessary for me to allow overriding the method to return a Group's subTarget control. There is no reason why fabric should first call `_findTargetCorner` then do `target.controls[corner]`. By returning the control directly, the fabric code is less verbose while also giving more flexibility to devs.

I'm not suggesting fabric should start caring about devs overriding internal methods, it's just that this seems a win-win for both. Fabric gets a simpler code, I get an overridable method "by chance" 😄 All thanks to improve return value

src/shapes/Object/InteractiveObject.ts
asturur: Or add your personal use case whichever it is if is not secret

jiayihu: > But it should also be called findTargetControl

I thought about that as well, I'd honestly prefer renaming but did not want to break the API, although it's internal. But if you're okay with it I think it would make more sense.

asturur:
I fixed the action for the changelog, first try!

https://github.com/fabricjs/fabric.js/assets/1194048/1e3db937-14f3-4d92-955c-12aa553239e0

asturur:
@jiayihu  and @ShaMan123 if you want to merge this PR fix the unit tests or tell me you can't but you would and i ll fix the unit tests.

asturur:
@ShaMan123  do you have opinions here?

asturur (approved):
pending @ShaMan123 review

src/shapes/Object/InteractiveObject.ts
```diff
@@ -189,29 +189,36 @@ export class InteractiveFabricObject<
...
           this,
           pointer,
           forTouch ? corner.touchCorner : corner.corner
         )
       ) {
         // this.canvas.contextTop.fillRect(pointer.x - 1, pointer.y - 1, 2, 2);
-        return (this.__corner = key);
+        this.__corner = key;
+
+        return { key, control };
```
ShaMan123, line 217: I'd say we should return the same value as getActiveControl, meaning the coord as well

src/canvas/Canvas.ts
```diff
@@ -1055,7 +1055,7 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
       );
       if (target === this._activeObject && (corner || !grouped)) {
         this._setupCurrentTransform(e, target, alreadySelected);
-        const control = target.controls[corner],
+        const control = corner ? corner.control : undefined,
```
ShaMan123, line 1058: I prefer `corner` being renamed.

ShaMan123 (approved):
I also am for renaming the method

src/canvas/Canvas.ts
asturur: corner is an internal variable right? if yes it can be renamed of course.

src/shapes/Object/InteractiveObject.ts
jiayihu: Done

src/canvas/Canvas.ts
jiayihu: Done, although `handle` sounds weird and never used before

jiayihu:
> I also am for renaming the method

Yeah, maybe we should call it `_findTargetControlHandle` and keep `_findTargetCorner: (pointer) => _findTargetControlHandle(pointer)?.key` basically

jiayihu:
`findControl` is okay for me, @asturur ?

asturur:
At this point changing what the method returns or what returns and how is called is the same.

asturur:
i m good with findControl but probably i don't want to touch anything else.
I ll update the PR and if we have second thoughts tomorrow we just can revert and fix it

ShaMan123:
I am now migrating this breaking change and I have noticed that the only usage left of `FabricObject#__corner` is by `getActiveControl`.
So I suggest removing `FabricObject#__corner` and `getActiveControl` and not exposing it as part of v6

ShaMan123:
updated comment

asturur:
The point of __corner, name apart, is that is precalculated by your last mouse movement and is used by developers too.
so is a good way to find it right away when you don't have a transform going on. `__corner` is a long standing feature, we could remove `getActiveControl` but not __corner.

asturur:
it is like that, funny or not is exposed through the object during events
It can be upgraded to official event property but removed seems removing something useful.

Issue #9653, opened 2024-02-06
[Feature]: Add orientation to circle


lbordowitz:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have searched and referenced existing issues, feature requests and discussions
- [X] I am filing a **FEATURE** request.

### Description

In some cases, it's more natural to program in a way where the start angle would be greater than the end angle. An additional property to the circle, `clockwiseOrientation`, would aid in making code more readable and straightforward.

### Current State

Currently, the circle will be drawn in a clockwise orientation, from `startAngle` to `endAngle`. If someone wants to use a circle in a counterclockwise orientation, they would need to set `endAngle` to their arc's beginning angle, and `startAngle` to the end of their arc's angle. The code, therefore, would have to separate "application logic" start/end angle and "drawing logic" start/end angle. This is possible, but has already tripped me up more times than I care to admit.

### Additional Context

`startAngle` and `endAngle` are not used in very many places internally. The switching logic would be pretty simple to implement.

ShaMan123:
PR is welcome to introduce clockwise/counterclockwise

lbordowitz:
@asturur I have added a screenshot in my pull request, #9670

asturur:
close as completed

Issue #9654, opened 2024-02-06
[Feature]: Use radians for angles in Circle


lbordowitz:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have searched and referenced existing issues, feature requests and discussions
- [X] I am filing a **FEATURE** request.

### Description

With a clean break between Fabric 5.0.0 and Fabric 6.0.0, it would be nice to have the angle provided in radians in the Circle class, maybe any other places that are still using degrees.

### Current State

The current code uses degrees instead of radians. That's a bit awkward in some spots.

### Additional Context

The `endAngle` documentation mentions " End angle of the circle probably will change to degrees in next major version ". Was it previously radians, and then changed to degrees? If that's the case, maybe just the documentation needs to change?

ShaMan123:
An effort was made to move all to deg
Why do you need rads?
It is annoying working in rads and needing to convert only for it to be converted back.
I would say that a robust soultion would be to introduce a class
```
class Angle {
  static deg(value: TDegree) 
  static rad(value: TDegree) 

  // impl lazy getters
  get/set deg() 
  get/set rad() 
  
}

```

and have the codebase use this class all over 
with lazy getters it will save compution and make all of us happy

asturur:
Yes it was in radians before and then moved to angle and the documentation should be fixed.

PR #9670, opened 2024-02-16
feat(Circle): Add counterclockwise parameter to Circle class

lbordowitz:
## Description

This resolves issue #9653 to implement orientation to the circle. I have also taken the liberty of tidying the documentation, per #9654 , to emphasize that the angle is in degrees and will stay that way.

## In Action

I used `npm start node` and the following code snippet for `index.mjs`:

```js
const canvas = new fabric.StaticCanvas(null, { width: 500, height: 500 });
    const circle1 = new fabric.Circle({
      left: 100,
      top: 100,
      radius: 20,
      originX: 'center',
      originY: 'center',
      fill: 'rgba(0,0,0,0)',
      noScaleCache: false,
      strokeUniform: true,
      strokeWidth: 1,
      stroke: '#00ff00',
      selectable: false,
      hasControls: false,
      lockScalingX: true,
      lockScalingY: true,
      startAngle: 45,
      endAngle: 270,
    })
    const circle2 = new fabric.Circle({
      left: 200,
      top: 100,
      radius: 20,
      originX: 'center',
      originY: 'center',
      fill: 'rgba(0,0,0,0)',
      noScaleCache: false,
      strokeUniform: true,
      strokeWidth: 1,
      stroke: '#00ff00',
      selectable: false,
      hasControls: false,
      lockScalingX: true,
      lockScalingY: true,
      startAngle: 45,
      endAngle: 270,
      clockwise: false,
    })
    canvas.add(circle1, circle2);
    canvas.renderAll();
```

This is the resulting snapshot:

Note that they sweep out complementary angles.

### DIFF
--- a/src/shapes/Circle.ts
+++ b/src/shapes/Circle.ts
@@ -19,20 +19,26 @@ interface UniqueCircleProps {
   radius: number;
 
   /**
-   * degrees of start of the circle.
-   * probably will change to degrees in next major version
-   * @type Number 0 - 359
+   * Angle for the start of the circle, in degrees.
+   * @type TDegree 0 - 359
    * @default 0
    */
   startAngle: number;
 
   /**
-   * End angle of the circle
-   * probably will change to degrees in next major version
-   * @type Number 1 - 360
+   * Angle for the end of the circle, in degrees
+   * @type TDegree 1 - 360
    * @default 360
    */
   endAngle: number;
+
+  /**
+   * Orientation for the direction of the circle.
+   * Setting to true will switch the arc of the circle to traverse from startAngle to endAngle in a counter-clockwise direction.
+   * Note: this will only change how the circle is drawn, and does not affect rotational transformation.
+   * @default false
+   */
+  counterClockwise: boolean;
 }
 
 export interface SerializedCircleProps
@@ -41,12 +47,18 @@ export interface SerializedCircleProps
 
 export interface CircleProps extends FabricObjectProps, UniqueCircleProps {}
 
-const CIRCLE_PROPS = ['radius', 'startAngle', 'endAngle'] as const;
+const CIRCLE_PROPS = [
+  'radius',
+  'startAngle',
+  'endAngle',
+  'counterClockwise',
+] as const;
 
 export const circleDefaultValues: UniqueCircleProps = {
   radius: 0,
   startAngle: 0,
   endAngle: 360,
+  counterClockwise: false,
 };
 
 export class Circle<
@@ -60,6 +72,7 @@ export class Circle<
   declare radius: number;
   declare startAngle: number;
   declare endAngle: number;
+  declare counterClockwise: boolean;
 
   static type = 'Circle';
 
@@ -101,7 +114,7 @@ export class Circle<
       this.radius,
       degreesToRadians(this.startAngle),
       degreesToRadians(this.endAngle),
-      false
+      this.counterClockwise
     );
     this._renderPaintInOrder(ctx);
   }
@@ -169,14 +182,10 @@ export class Circle<
         startY = sin(start) * radius,
         endX = cos(end) * radius,
         endY = sin(end) * radius,
-        largeFlag = angle > 180 ? '1' : '0';
+        largeFlag = angle > 180 ? 1 : 0,
+        sweepFlag = this.counterClockwise ? 0 : 1;
       return [
-        `<path d="M ${startX} ${startY}`,
-        ` A ${radius} ${radius}`,
-        ' 0 ',
-        `${largeFlag} 1`,
-        ` ${endX} ${endY}`,
-        '" ',
+        `<path d="M ${startX} ${startY} A ${radius} ${radius} 0 ${largeFlag} ${sweepFlag} ${endX} ${endY}" `,
         'COMMON_PARTS',
         ' />\n',
       ];
--- a/test/unit/circle.js
+++ b/test/unit/circle.js
@@ -117,6 +117,7 @@
       radius:                   0,
       startAngle:               0,
       endAngle:                 360,
+      counterClockwise:         false,
       skewX:                    0,
       skewY:                    0,
       strokeUniform:            false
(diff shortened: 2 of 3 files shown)


src/shapes/Circle.ts
```diff
@@ -19,20 +19,25 @@ interface UniqueCircleProps {
   radius: number;
 
   /**
-   * degrees of start of the circle.
-   * probably will change to degrees in next major version
+   * Angle for the start of the circle, in degrees.
    * @type Number 0 - 359
```
ShaMan123, line 23:
```suggestion
   * @type TDegree 0 - 359
```

src/shapes/Circle.ts
```diff
@@ -19,20 +19,25 @@ interface UniqueCircleProps {
...
    * @default 0
    */
   startAngle: number;
 
   /**
-   * End angle of the circle
-   * probably will change to degrees in next major version
+   * Angle for the end of the circle, in degrees
    * @type Number 1 - 360
```
ShaMan123, line 30:
```suggestion
   * @type TDegree 1 - 360
```

src/shapes/Circle.ts
```diff
@@ -19,20 +19,25 @@ interface UniqueCircleProps {
...
    * @default 360
    */
   endAngle: number;
+
+  /**
+   * Orientation for the direction of the circle.
+   * Setting to false will switch the arc of the circle to traverse from startAngle to endAngle in a counterclockwise direction,
+   * @default true
+   */
+  clockwise: boolean;
```
ShaMan123, line 40: Look at MDN ref: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/arc

I wish we align to it

ShaMan123 (commented):
Looks good
I am not sure if how we should name the flag
Maybe it doesn't matter
Anyways we want to align to MDN at least with the default value, I would align also with the flag name
What do you think?

ShaMan123:
The PR is missing a changelog entry and prettier (npm run prettier:write)
You can read the contribution guide for more context

ShaMan123:
We might want to add the visual as a test, @asturur ?

src/shapes/Circle.ts
lbordowitz: I totally agree with you. It's consistent, convenient, and conventional. It should be called counterclockwise.

src/shapes/Circle.ts
```diff
@@ -41,12 +46,18 @@ export interface SerializedCircleProps
...
+  'startAngle',
+  'endAngle',
+  'counterclockwise',
+] as const;
 
 export const circleDefaultValues: UniqueCircleProps = {
   radius: 0,
   startAngle: 0,
   endAngle: 360,
+  counterclockwise: true,
```
ShaMan123, line 60: Isn't this false by default?

src/shapes/Circle.ts
```diff
@@ -19,20 +19,25 @@ interface UniqueCircleProps {
...
    * @default 360
    */
   endAngle: number;
+
+  /**
+   * Orientation for the direction of the circle.
+   * Setting to true will switch the arc of the circle to traverse from startAngle to endAngle in a counterclockwise direction,
+   * @default false
+   */
+  counterclockwise: boolean;
```
ShaMan123, line 40: I think we should camelCase
`counterClockwise`?

test/unit/circle.js
```diff
@@ -166,6 +167,14 @@
     assert.equal(svgClipPath, '\t<path d=\"M 10 0 A 10 10 0 0 1 -10 0\" transform=\"matrix(1 0 0 1 10.5 10.5)\"  />\n', 'half circle as clipPath');
   });
 
+  QUnit.test('toSVG with counterclockwise half circle', function (assert) {
+    var circle = new fabric.Circle({ width: 100, height: 100, radius: 10, endAngle: fabric.util.radiansToDegrees(Math.PI), counterclockwise: true });
```
ShaMan123, line 171:
```suggestion
    var circle = new fabric.Circle({ width: 100, height: 100, radius: 10, endAngle: 180, counterclockwise: true });
```

src/shapes/Circle.ts
```diff
@@ -169,12 +181,13 @@ export class Circle<
         startY = sin(start) * radius,
         endX = cos(end) * radius,
         endY = sin(end) * radius,
-        largeFlag = angle > 180 ? '1' : '0';
+        largeFlag = angle > 180 ? '1' : '0',
+        sweepFlag = this.counterclockwise ? '0' : '1';
```
ShaMan123, line 185: I would turn these values to integers.
A relic of the past preserved them as strings
And of course we can do `Number(!this.counterclockwise)` but that is not crucial for me at all

ShaMan123 (commented):
Looks good

src/shapes/Circle.ts
ShaMan123: Can you add this to the appropriate test?

src/shapes/Circle.ts
asturur: I m having an hard time understanding if the  is better clockWise or counterClockWise

src/shapes/Circle.ts
asturur: Yes false by default if is counterClockwise

asturur:
Just to make clear how things seems obvious but aren't, i understood this would behave completely different,
a circle from 0 to 30 in counter clockwise for me would have created a sector of 30 degrees turning counter clockwise and not a circle of 360-30.

src/shapes/Circle.ts
```diff
@@ -19,20 +19,25 @@ interface UniqueCircleProps {
...
+   * @type TDegree 1 - 360
    * @default 360
    */
   endAngle: number;
+
+  /**
+   * Orientation for the direction of the circle.
+   * Setting to true will switch the arc of the circle to traverse from startAngle to endAngle in a counterclockwise direction,
+   * @default false
+   */
```
asturur, line 39: Can we be extremely safe and explain with an extra line that this counterClockwise is not going to affect anything else like the rotate method or the normal angle of the circle? I don't want devs to find this and then being confused that the circle doesn't spin counter clockwise.

src/shapes/Circle.ts
```diff
@@ -169,12 +181,13 @@ export class Circle<
...
+        largeFlag = angle > 180 ? '1' : '0',
+        sweepFlag = this.counterclockwise ? '0' : '1';
       return [
         `<path d="M ${startX} ${startY}`,
         ` A ${radius} ${radius}`,
         ' 0 ',
-        `${largeFlag} 1`,
+        `${largeFlag} ${sweepFlag}`,
         ` ${endX} ${endY}`,
         '" ',
```
asturur, line 192: since you are here, it shoud be possible to make a unique string of this. Just common parts needs to be broken away in its own array element. 
```js
`<path d="M ${startX} ${startY} A ${radius} ${radius} 0 ${largeFlag} ${sweepFlag} ${endX} ${endY}" ` 
```

src/shapes/Circle.ts
lbordowitz: I would prefer not to do anything tricky like that.

asturur:
I m ok also without the visual test, since for now.
I would like to use this occasion to try to add the token with the permissions to actually execute those extra checks from forks, i ll merge it monday/tuesday when i can try to work on the token

asturur (approved):

ShaMan123 (approved):
@asturur I will not merge n case you want to work on the tokens

asturur:
The token thing is not as easy as it seems.
Or at least is not so straight forward.
I created a personal token and i added it in the repo, with granular permission on how to write.
Now i have to figure out how to make it appear as an env variable in the actions and have the apis use it.

PR #9673, opened 2024-02-20
chore(TS): Swap an any for the actual type in toCanvasElement and toDataUrl

asturur:
## Description

When @jiayihu made the toCanvasElement option i noticed that we had an any to describe the types.
Today i was using the function and so i typed it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "fabric",
-      "version": "6.0.0-beta18",
+      "version": "6.0.0-beta19",
       "license": "MIT",
       "devDependencies": {
         "@babel/cli": "^7.22.9",
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -19,6 +19,7 @@ import type {
   TCacheCanvasDimensions,
   Abortable,
   TOptions,
+  ImageFormat,
 } from '../../typedefs';
 import { classRegistry } from '../../ClassRegistry';
 import { runningAnimations } from '../../util/animation/AnimationRegistry';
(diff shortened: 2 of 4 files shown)


package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "fabric",
-      "version": "6.0.0-beta18",
+      "version": "6.0.0-beta19",
```
ShaMan123, line 9: we should run `npm i` when we release and commit this as part of the action

src/shapes/Object/Object.ts
```diff
@@ -70,6 +71,34 @@ export type TCachedFabricObject<T extends FabricObject = FabricObject> = T &
...
+  enableRetinaScaling?: boolean;
+  // Remove current object transform ( no scale , no angle, no flip, no skew ). Introduced in 2.3.4
+  withoutTransform?: boolean;
+  // Remove current object shadow. Introduced in 2.4.2
+  withoutShadow?: boolean;
+  // Account for canvas viewport transform
+  viewportTransform?: boolean;
+  // Function to create the output canvas to export ont
+  canvasProvider?: (el?: HTMLCanvasElement) => Canvas;
+}
```
ShaMan123, line 96: I think if you use `/** */` as comment syntax then it appears when you hover the option in VSCODE

src/shapes/Object/Object.ts
```diff
@@ -70,6 +71,34 @@ export type TCachedFabricObject<T extends FabricObject = FabricObject> = T &
...
+  // Enable retina scaling for clone image. Introduce in 1.6.4
+  enableRetinaScaling?: boolean;
+  // Remove current object transform ( no scale , no angle, no flip, no skew ). Introduced in 2.3.4
+  withoutTransform?: boolean;
+  // Remove current object shadow. Introduced in 2.4.2
+  withoutShadow?: boolean;
+  // Account for canvas viewport transform
+  viewportTransform?: boolean;
+  // Function to create the output canvas to export ont
+  canvasProvider?: (el?: HTMLCanvasElement) => Canvas;
```
ShaMan123, line 95:
```suggestion
  canvasProvider?: (el?: HTMLCanvasElement) => StaticCanvas;
```

ShaMan123: Or maybe we should pass a generic here `<T extends StaticCanvas>`

asturur: yes maybe a generic is better. i ll see if i can make it without triggerin 1000 warning. If i trigger 1000 warning i l leave it as StaticCanvas

ShaMan123: It might because of a def cycle

ShaMan123: Maybe better to define the interface needed for the method

PR #9687, opened 2024-02-25
feat(LayoutManager): Handle the case of activeSelection with objects inside different groups

asturur:
## Description

This is #9651 with a different approach


asturur:
i moved the new code in the old PR, let's keep the old one that has the discussion

PR #9693, opened 2024-02-27
WIP fix(): context isolation 

ShaMan123:
## Description

POC to drop object caching logic in favor of a better rendering approach

Ports parts of #8298 

Object caching and context isolation are 2 different things that were smashed together.
Object caching is a performance optimization targeting apparently panning and transforming (zooming invalidates the cache)
Context isolation is a necessity for supporting rendering operations in an object tree such as clipping a single object. Without it the operation is applied to the entire context, which is not desired.
Currently object caching covers most of the needs of a simple app in terms of context isolation. However if you move out of lines fabric's output is completely broken. e.g. shadows under a group with a clip path.
Also object caching has the blur issue because a clip path is unaware of it's parent, making the result of `getTotalObjectScaling` wrong always and the cache size off. That should be fixed regardless and of course I have a fix for it in #8298 .
Add to that a dev setting object caching to true/false and discovering to their surprise that fabric didn't comply and decided by itself what to do. This is the essence of the confusion and mixing of these 2 concepts.
Another issue I have been looking into is objects that are partially on screen. If they hold a cache it might be extremely wasteful instead of just rendering their visible part.

In this PR I aim to make the rendering logic simpler following a simple rule: Recursion **must** happen in a single method calling itself rather than a number of methods calling each other as it is currently.

- exposed `renderInIsolation` which renders the object at its final transform onto an isolated context and then draws that image onto the main context. For simplicity of the POC I didn't try (yet) to optimize the isolated context size.

- renamed `needsItOwnCache` to `requiresContextIsolation`
- disabled caching altogether - it can be reinstated if we want or completely refactored to use the isolation output. I think as a first step we should extract all the caching logic to a standalone so it is not on the Object class. If we understand it is a perf opt then it should not remain there. Half the code on Object is related to caching. 

This is a big step towards being able to invalidate a region in the canvas.

## In Action

https://github.com/fabricjs/fabric.js/assets/34343793/583d4e14-8df5-4171-ab64-570eda42e6e3

https://github.com/fabricjs/fabric.js/assets/34343793/2e172c91-6b29-4ee7-a94f-ef8e85f9120d

### DIFF
--- a/.codesandbox/templates/vanilla/src/testcases/simpleTextbox.ts
+++ b/.codesandbox/templates/vanilla/src/testcases/simpleTextbox.ts
@@ -37,9 +37,19 @@ export function testCase(canvas: fabric.Canvas) {
     height: 50,
     left: 0,
     top: 100,
+    paintFirst: 'stroke',
+    shadow: new fabric.Shadow({
+      affectStroke: true,
+      blur: 5,
+      offsetX: 20,
+      offsetY: 20,
+      // nonScaling: true,
+      color: 'red',
+    }),
   });
   canvas.centerObject(text);
   canvas.centerObject(rect);
+  canvas.preserveObjectStacking = true;
   const group = new fabric.Group([rect, text], {
     subTargetCheck: true,
     interactive: true,
@@ -61,6 +71,9 @@ export function testCase(canvas: fabric.Canvas) {
           opacity: value,
         });
         text.set({ dirty: true });
+        rect.shadow!.offsetX = 20 * (value + 1);
+        rect.shadow!.blur = 20 * value;
+        rect.set({ dirty: true });
         canvas.renderAll();
       },
       onComplete: () => animate(!toState),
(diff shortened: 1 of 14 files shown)


asturur:
Please keep POCs out of PR, document first with a document and after discussion let's PR.

ShaMan123:
I will port the doc to the description
I agree with the doc first approach
This is something I need regardless and immediately for projects that have severe bugs caused by caching, whether in wrong visuals or OOM

asturur:
> I will port the doc to the description I agree with the doc first approach This is something I need regardless and immediately for projects that have severe bugs caused by caching, whether in wrong visuals or OOM

If you need it immediately you have to organize with a branch while we work on this being finalized/understood.
There are PRs from 4 years ago that intersect with this and that are waiting to be merged too.

ShaMan123:
@asturur I have updated the description

asturur:
@ShaMan123 this is a larger plan that needs to include other open PR from 2020 that weren't finished for the migration.
- filters on objects
- masks
- clip path blurriness old PR
They all need to be remade with also global composite operation in mind and all of it intersect and overlap with this PR.

Please just make a document, let's see what is broken and what it needs improvement and how.
I m not doing that in a PR

ShaMan123:
Ported to https://github.com/ShaMan123/fabric.js/pull/6
Contributions are welcome

PR #9698, opened 2024-02-29
chore(TS): use consistent and improved types for getDefaults and ownDefaults

jiayihu:
`Textbox.getDefaults` type is inferred, resulting in error if trying to extend the class and override the static method:

```ts
// Will error
class Test extends Textbox {
  static ownDefaults = {
    textAlign: "center"
  } satisfies Partial<Textbox>;

  static getDefaults(): Record<string, any> {
    return {
      ...super.getDefaults(),
      ...Test.ownDefaults
    }
  }
}
```

Screenshot 2024-02-29 at 13 28 00

I took the chance to ensure all `getDefaults` explicitly return the type `: Record<string, any>`, whereas the default values are strictly typed instead of being `Record<string, any>`.

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -271,7 +271,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
    */
   protected declare _target?: FabricObject;
 
-  static ownDefaults: Record<string, any> = canvasDefaults;
+  static ownDefaults = canvasDefaults;
 
   static getDefaults(): Record<string, any> {
     return { ...super.getDefaults(), ...SelectableCanvas.ownDefaults };
--- a/src/shapes/ActiveSelection.ts
+++ b/src/shapes/ActiveSelection.ts
@@ -7,12 +7,18 @@ import {
   LAYOUT_TYPE_ADDED,
   LAYOUT_TYPE_REMOVED,
 } from '../LayoutManager/constants';
+import type { TClassProperties } from '../typedefs';
 
 export type MultiSelectionStacking = 'canvas-stacking' | 'selection-order';
 
 export interface ActiveSelectionOptions extends GroupProps {
   multiSelectionStacking: MultiSelectionStacking;
 }
+
+const activeSelectionDefaultValues: Partial<TClassProperties<ActiveSelection>> =
+  {
+    multiSelectionStacking: 'canvas-stacking',
+  };
 
 /**
  * Used by Canvas to manage selection.
@@ -28,11 +34,9 @@ export interface ActiveSelectionOptions extends GroupProps {
 export class ActiveSelection extends Group {
   static type = 'ActiveSelection';
 
-  static ownDefaults: Record<string, any> = {
-    multiSelectionStacking: 'canvas-stacking',
-  };
+  static ownDefaults = activeSelectionDefaultValues;
 
-  static getDefaults() {
+  static getDefaults(): Record<string, any> {
     return { ...super.getDefaults(), ...ActiveSelection.ownDefaults };
   }
 
--- a/src/shapes/Circle.ts
+++ b/src/shapes/Circle.ts
@@ -54,7 +54,7 @@ const CIRCLE_PROPS = [
   'counterClockwise',
 ] as const;
 
-export const circleDefaultValues: UniqueCircleProps = {
+export const circleDefaultValues: Partial<TClassProperties<Circle>> = {
   radius: 0,
   startAngle: 0,
   endAngle: 360,
@@ -78,7 +78,7 @@ export class Circle<
 
   static cacheProperties = [...cacheProperties, ...CIRCLE_PROPS];
 
-  static ownDefaults: Record<string, any> = circleDefaultValues;
+  static ownDefaults = circleDefaultValues;
 
   static getDefaults(): Record<string, any> {
     return {
--- a/src/shapes/Ellipse.ts
+++ b/src/shapes/Ellipse.ts
@@ -8,7 +8,7 @@ import type { FabricObjectProps, SerializedObjectProps } from './Object/types';
 import type { ObjectEvents } from '../EventTypeDefs';
 import type { CSSRules } from '../parser/typedefs';
 
-export const ellipseDefaultValues: UniqueEllipseProps = {
+export const ellipseDefaultValues: Partial<TClassProperties<Ellipse>> = {
   rx: 0,
   ry: 0,
 };
@@ -52,9 +52,9 @@ export class Ellipse<
 
   static cacheProperties = [...cacheProperties, ...ELLIPSE_PROPS];
 
-  static ownDefaults: Record<string, any> = ellipseDefaultValues;
+  static ownDefaults = ellipseDefaultValues;
 
-  static getDefaults() {
+  static getDefaults(): Record<string, any> {
     return {
       ...super.getDefaults(),
       ...Ellipse.ownDefaults,
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -62,7 +62,7 @@ export interface GroupProps extends FabricObjectProps, GroupOwnProps {
   layoutManager: LayoutManager;
 }
 
-export const groupDefaultValues = {
+export const groupDefaultValues: Partial<TClassProperties<Group>> = {
   strokeWidth: 0,
   subTargetCheck: false,
   interactive: false,
--- a/src/shapes/IText/IText.ts
+++ b/src/shapes/IText/IText.ts
@@ -7,7 +7,7 @@ import {
   keysMap,
   keysMapRtl,
 } from './constants';
-import type { TFiller, TOptions } from '../../typedefs';
+import type { TClassProperties, TFiller, TOptions } from '../../typedefs';
 import { classRegistry } from '../../ClassRegistry';
 import type { SerializedTextProps, TextProps } from '../Text/Text';
 import {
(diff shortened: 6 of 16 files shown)


ShaMan123 (commented):
Before diving in
I was pretty sure that TClassProperties is something we should remove from the code base in favor of explicit types
What are your thoughts?

jiayihu:
> I was pretty sure that TClassProperties is something we should remove from the code base in favor of explicit types

Would be sensible if there was a way to tell which properties are "configurable" and which are not, but the truth is that it's not currently possible. The fact that there are even protected properties in `IText` default values is a living proof of that.
Also I don't see any real benefit in listing manually the types. I'm personally more in favour of explicit types generally but here it doesn't seem useful for fabric.

ShaMan123:
I also prefer explicit but I disagree and do think it is useful for the dev
That way the types are strict

asturur:
i thought Typescript would make something like this ok:

```ts
interface TestOpts {
  propA: string;
  propB: number;
}

export class Test implements TestOpts {
  declare propA: string;
  declare propB: number;

  static OwnDefaults: TestOpts & Record<string, any> = {
    propA: 'a',
    propB: 2,
  };

  static getDefaults(): TestOpts {
    return this.OwnDefaults;
  }

  constructor(opts: TestOpts) {
    Object.assign(opts);
  }
}

interface circleOwnProps {
  radius: number;
}

interface circleProps extends TestOpts, circleOwnProps {}

export class Circle extends Test implements circleProps {
  declare radius: number;

  static OwnDefaults: circleOwnProps & Record<string, any> = {
    radius: 4,
  };

  static getDefaults(): TestOpts {
    return {
      ...super.getDefaults(),
      ...this.OwnDefaults,
    };
  }

  constructor(opts: circleProps) {
    super(opts);
  }
}
```

but keeping OwnDefaults returning a record<any> seems the only thing it likes.
I thought this would convince typescript `circleOwnProps & Record<string, any>` but it doesn't.

jiayihu:
`circleOwnProps & Record<string, any>` equals to `circleOwnProps` that's why it doesn't work, if I got right your example.

> I also prefer explicit but I disagree and do think it is useful for the dev
That way the types are strict

I didn't say it's not useful at all for the dev, I meant that it's not in the current state of fabric. If we make defaults explicit, how would one be able to add their own defaults in subclasses for instance? TS will probably give errors because the subclass defaults wouldn't be assignable to the fabric baseclass defaults. You have to support at least 2 cases:

1. subclass.ownDefaults assignable to baseclass.ownDefaults
2. subclass.getDefaults() assignable to baseclass.getDefaults()

I think using explicit types may result just in more confusing TS errors. Returning `Record<string, any>` while typing `ownDefaults` with `Partial<TClassProperties<FabricClass>>` looks a decent compromise for me.

jiayihu:
I think that `getDefaults` and `ownDefaults` is by-nature a **dynamic** approach to allow to overwrite the default value of any property of the class. So that means:
- **Unbounded** type of property that can be overriden -> protected properties in IText are a clear example
- It won't play nice with TS most probably, since it prefers **statically** and **known** properties

So that's why I'd not put much effort in ensuring more explicit types.

asturur:
The fact that some private properties are default is a remain of the migration.
Slowly and not without talking, most of the private properties will be moved away and first level properties on the classes will just be state properties of the class.
But at this point we are where we are and it must be clear that is not by design, is just a combination of a long complex migration mixed with added features and refactors and typescript experience gained on the field while doing.

Something that should stay clear is that one of the end goals is that if i just want to add an ID property or a name or something to my class that is extra state, i shouldn't subclass.
I should just extend an interface or do the simplest thing i can do.

asturur:
Also typescript forbidding static parts to use the generic types from the class is a big limitation

jiayihu:
> Something that should stay clear is that one of the end goals is that if i just want to add an ID property or a name or something to my class that is extra state, i should be able to do it without subclassing.

Isn't that the opposite of the design goal mentioned in the past in some PRs? For instance I'm using declaration merging so that `new fabric.Object({ uuid: "" })` works since `uuid` is an extra property. But I was told, by @ShaMan123 , that the designed way would be to have a subclass and have the extra property declared there

ShaMan123:
Not the fabric designed way, the TS designed way
Fabric would not need the declare directive if it would not break the prototype chain.
Having any prop assignable is exactly breaking the prototype chain.

ShaMan123:
It is a lot of noice for this simple case of assigning an id.

ShaMan123:
If fabric would stop breaking the prototype chain this pr would be abandonded as is
Not ownDefaults, getDefaults, declare. None.
Since you are a strict TS dev that is what I brought to your attention. And in fact I think this is how it should be. This is a library not an app. It must conform to the strictest rules.

ShaMan123:
In addition I think declarstion merging is fine if you do not wish or can not subclass

asturur:
You know what, i m starting a doc since i decided that 2024 is the year of docs.

ShaMan123:
Could it be the year of plans?

asturur:
Yes the plan is:
- fix the layout bugs
- cut 6.0 as is because enough
- give people a website they can consult
- work on some of the following features:
- coords, renders, masks, filters for objects evaluating them from ground up as a group

I m abstaining as much as possible from self driven initiatives because the repo is a mess right now, open stuff open questions too much stuff.

asturur:
in the test i made it wasn't necessary to define both as record<string, any> one was enough

asturur (dismissed):

asturur:
I Briefly tested the branch and the suggestions  that vscode gives me are the same, i m ok with this if it enable more flexibility when fiddling with the class from the outside

jiayihu:
Yes maybe using interfaces and declaration merging could be an alternative. But declaration merging has its own limits, for instance you can't override or remove properties. And I reckon most of the people using TS are not very familiar with declaration merging.

IMO as I said both `getDefaults` and serialized outputs are dynamic points of the code, where trying to type more strictly might cause more hard than help. See for instance my past PR for `toObject`: https://github.com/fabricjs/fabric.js/pull/9014

asturur:
I m very let down from class + types in general in TS.
maybe they will fix since is all in their head/plans, is not a JS limit.

jiayihu:
Class support is overall weak because dynamic behaviour doesn't play nice with static types. Static types reason about Math soundness, thus in **all cases** of a domain. Dynamic behaviour relies on the exact runtime execution, which is thus unpredictable. T For instance, sometimes even just the order of imports could affect the prototype and thus result.

The problem is that for Fabric and low-level libs, sometimes a dynamic approach is correct enough 99%, whereas the 100% correct static alternative would be much more troublesome.

The HTML DOM native types in TS are not very strict as well, even in `strict` mode, if you think about it.

Anyway, for this PR is it good enough to be merged?

asturur:
For me yes, usually when you and @ShaMan123 have a discussion along the PR when my criteria are met i approve and leave to you two. If i approve for me is always good to merge, otherwise i use the request for change or clear requests in the comments.

ShaMan123 (approved):
I want to recap on this because my view have changed.
Why are we here?
We migrated to use classes.
Class fields have a default value syntax
```ts
class Foo {
  bar = 'defaultValue'
}

```

We started using this syntax.
Then we encountered a bug because our expectations from the default value syntax were different than what it was.
It does not assign a value only if the `bar` is nil. It assigns the default value no matter what so that when the constructor method is called `defaultValue` is already assigned to `bar` allowing the constructor to override that with a different value.
Also without using `declare` we discovered that a default value of `void 0` is assigned to the field as mentioned above.
This makes sense because that field should not exist before the descrobed initialization process.
What is wrong in fabric is the usage of `set`.
`set` is called from the constructor of the super class. It assigns anything that is passed to it.
When initializing a subclass:

1. `super()` is called 
2. `set` is triggered
3. default values of the subclass are assigned
4. subclass constructor executes

Step 3 overrides values from step 2. We did not expect that.
The bug is the order of execution or the fact the we even use `set` like this.
IMO a constructor should explicitly set its own fields.
No bulk assigning.
Then we align to the standard.

Part of my goals for fabric is to align to standards because it makes it easier to use without explaining weird concepts, without bad gotchas. Less friction.

As an escape hatch `getDefaults` was introduced.
It served as a syntax replacement to the default value syntax, did not override existing values and allows to change the default values without subclassing.
`ownDefaults` is something @asturur introduced to avoid recreating an object each time it is called.

As long as we run side effects in the constructor I am not sure we can avoid this (@jiayihu helped me understand this).
Consider a dev subclassing, assigning a default value but since the default value will be assigned after the super class initialization has completed the effects will run on wrong values.

So all in all maybe this is the best we can do for now.
I do think that `ownDefaults` are super confusing for devs. It is simply a performance optimization. An arguably very small one. But it adds friction. So if it should stay I think we should remove it from the types somehow. Maybe by extracting the object to a const as it is in most places and not exposing it on the class. Then we have `getDefaults` only.

asturur:
getDefaults is the escape hatch for classes not supporting values on the prototype.
With the public property you are forced to assign values for each instance of the class and you can't  change the standard defaults.
With getDefaults you can still remove that behaviour and keep defaults on the prototype by emptying the default object and so assign nothing.
On top of that if you want you can still assign in your constructor of your subclass your own properties, it doesn't block you from doing that.
That was the reason to add back getDefaults, to work around a limitation of classes that for me was a showstopper

asturur (approved):

jiayihu:
> As long as we run side effects in the constructor I am not sure we can avoid this (@jiayihu helped me understand this).
Consider a dev subclassing, assigning a default value using the formal syntax. Since the default value will be assigned after the super class initialization has completed the effects will run on wrong values.

Exactly. As long as side effects are run in the constructor, devs will need to customise those effects by manipulating the used default values.

We could just pass the overrides before calling the constructor `super(overridenDefaults)` but there are cases where one might want an effect to use an overriden method or subclass property IIRC. I can't remember the exact details honestly, but we had issues with the Text migration @ShaMan123 . 
Anyway the root issue were the constructor effects. Without them we could probably get rid of `ownDefaults/getDefaults`

asturur:
no guys sorry i don't want to insist but i want to be sure you understand.
'ownDefaults/getDefaults' is there because it was my let go for classes.
Classes were stopping certain behaviours and so i said, ok this only way we can keep doing that.
Yes you can remove sideEffects from constructors ( like setDimenions ) and ask the dev to just call it manually after the class   is created: (new Text()).initialize(), but that does not remove any requirement for `ownDefaults/getDefaults`

asturur:
`ownDefaults/getDefaults/Assign` is something that i added and asked otherwise i would have rolled back to function + prototype.

Issue #6208, opened 2020-03-07
React


hsluoyz:

asturur:
Or should React support fabric.js?
What does support mean here?

hsluoyz:
See an alternative Konva and React-Konva: https://konvajs.org/docs/react/index.html

asturur:
I do not have to have another big project, this is the main issue.
Complex fabricjs project and react already leave together happily without need of a library to comunicate.
Just initialize your canvas, use a createRef api and use fabric normally.

Data-flows are not solved by this kind of bindings.

If someone wants to do this kind of job, it can be hosted here, or in another repo, as he/she prefer.

mrslimslim:
Fabricjs can work perferctly in React project.Checkt https://ludus.one/

hsluoyz:
You link is not source code.

asturur:
Because there is not source code for a so generic question/feature.
FabricJS works with react.

You do not need a react wrapper, you do not need to do things that are not related to DOM with react, when building a react app.

We may want a specific react initializer? maybe.
Is it clear what kind of support were you looking for? no.
You did link a website that says to look at a github repo.

How much time do you think a person wants to spend to try to understand what are you looking for? Very little. This is not because of the topic or you, but because it requires a lot of effort and time to follow those things and try to help or fix bugs.

I can assume that since you were looking for something like that, you had difficulties using react and fabricJS together in a way that seems right to you.
But i can't assume it, you should be here explaining what and how, or you'll probably get nothing.

AdamMadrzejewski:
Hey @asturur do you think making a tutorial how to set up the FabricJS with React would be a good idea? I can do it, if you can give me some indication on where you would like me to post/submit it.

Looing at StackOverflow FabricJS + React, there are duplicates people asking how to do it https://stackoverflow.com/search?q=%5Bfabricjs%5D+react

I think it would be a good idea.

asturur:
We cold make a demo of a canvas with a small panel that display the text information of the selected text object on the canvas
To show how to possibly use react-redux and react with fabric

niels-van-den-broeck:
We've worked on a react project which is fueled by fabricjs for about 2 years now. Apart from it being hard to sync rendercycles, I think it works just fine. What we did was add the `renderOnAddRemove: false` option to the canvas and do a requestRenderAll at the end of each component update. That way it is possible to at least kinda sync the render cycles.

niels-van-den-broeck:
[...]

PR #9700, opened 2024-02-29
BREAKING BETA refactor(DOMManager): abstract to avoid DOM mutation

ShaMan123:
## Description

#6208 

This PR refactors the DOMManager concept enabling a canvas class to accept a DOMManager instance as the first arg, thus allowing the dev to handle the DOM as they wish with whatever framework they use.
This is a long standing request and it aligns fabric with the big boys. 
The DOM mutation that fabric does is dirty and weird and counter intuitive. However it can remain.

This PR also addresses another pain in the DOMManager API.
It was hard and annoying trying to add another canvas and have it resize etc. with the rest.
Now it is seamless.
In addition it allows the dev to handle the DOM between the 2 canvases as well which is great.

I have removed the absolute positioning of the main canvas. This made resizing the container redundant so I removed that logic and is a lot easier to deal with because the container fits the content and therefore elements inside it can use percent based sizing.

### Gist 
`DOMManager` is the class that holds all the DOM elements the canvas needs but does not mutate the DOM.
`StaticCanvasDOMManager` subclasses it and behaves as before.

Check out the next app changes. I believe the hook should be the official react binding.

### DIFF
--- a/.codesandbox/templates/next/components/index.tsx
+++ b/.codesandbox/templates/next/components/index.tsx
@@ -15,7 +15,9 @@ const canvasStyle: React.CSSProperties = {
   userSelect: 'none',
 };
 
-export const CanvasSlot = ({ name }: { name: string }) => <div name={name} />;
+export const CanvasSlot = ({ name }: { name: 'main' | 'top' | string }) => (
+  <div name={name} />
+);
 
 /**
  * @example Standard
@@ -87,15 +89,20 @@ export const CanvasComponent = ({
     };
   }, [canvasRefs, create]);
 
-  return React.Children.map(children, (child) =>
-    typeof child === 'object' &&
-    (child as React.ReactElement).type === CanvasSlot ? (
-      <canvas
-        style={canvasStyle}
-        ref={canvasRefs[(child as React.ReactElement).props.name]}
-      />
-    ) : (
-      child
-    )
-  );
+  return React.Children.map(children, (child) => {
+    if (
+      typeof child === 'object' &&
+      (child as React.ReactElement).type === CanvasSlot
+    ) {
+      const { name } = (child as React.ReactElement).props;
+      return (
+        <canvas
+          style={name === 'main' ? undefined : canvasStyle}
+          ref={canvasRefs[name]}
+        />
+      );
+    } else {
+      return child;
+    }
+  });
 };
--- a/.codesandbox/templates/next/components/index.tsx
+++ b/.codesandbox/templates/next/components/index.tsx
@@ -1,0 +1,101 @@
+import React, { createRef, useEffect, useMemo } from 'react';
+import {
+  DOMManager,
+  DOMManagerType,
+  StaticCanvas,
+  StaticDOMManagerType,
+} from 'fabric';
+
+const canvasStyle: React.CSSProperties = {
+  position: 'absolute',
+  left: 0,
+  top: 0,
+  touchAction: 'none',
+  msTouchAction: 'none',
+  userSelect: 'none',
+};
+
+export const CanvasSlot = ({ name }: { name: string }) => <div name={name} />;
+
+/**
+ * @example Standard
+ *   const create = useCallback((arg0: DOMManagerType) => {
+ *     const canvas = new Canvas(arg0);
+ *     ...
+ *     return canvas; // return for cleanup
+ *   });
+ *
+ *   <CanvasComponent create={create}>
+ *     <CanvasSlot name="main" />
+ *     <CanvasSlot name="top" />
+ *   </CanvasComponent>
+ *
+ * @example DOM elements in between
+ *   <CanvasComponent create={create}>
+ *     <CanvasSlot name="main" />
+ *     <CanvasSlot name="pointers" />
+ *     Hello World
+ *     <CanvasSlot name="top" />
+ *   </CanvasComponent>
+ *
+ * @example Static
+ *   const createStatic = useCallback((arg0: StaticDOMManagerType) => {
+ *     const canvas = new StaticCanvas(arg0);
+ *     ...
+ *     return canvas; // return for cleanup
+ *   });
+ *   <CanvasComponent create={createStatic}>
+ *     <CanvasSlot name="main" />
+ *   </CanvasComponent>
+ */
+export const CanvasComponent = ({
+  create,
+  children,
+}: React.PropsWithChildren<{
+  create: (arg0: StaticDOMManagerType | DOMManagerType) => StaticCanvas;
+}>) => {
+  const canvasRefs = useMemo(
+    () =>
+      React.Children.toArray(children).reduce((refs, child) => {
+        typeof child === 'object' &&
+          (child as React.ReactElement).type === CanvasSlot &&
+          (refs[(child as React.ReactElement).props.name] =
+            createRef<HTMLCanvasElement>());
+
+        return refs;
+      }, {} as Record<string, React.RefObject<HTMLCanvasElement>>),
+    [children]
+  );
+
+  useEffect(() => {
+    // it is crucial `create` is a dependency of this effect
+    // to ensure the canvas is disposed and re-created if it changes
+    const canvas = create(
+      new DOMManager(
+        Object.fromEntries(
+          Object.entries(canvasRefs).map(([key, ref]) => [key, ref.current])
+        )
+      )
+    );
+
+    return () => {
+      // `dispose` is async
+      // however it runs a sync DOM cleanup
+      // its async part ensures rendering has completed
+      // and should not affect react
+      canvas.dispose();
+    };
+  }, [canvasRefs, create]);
+
+  return React.Children.map(children, (child) =>
+    typeof child === 'object' &&
+    (child as React.ReactElement).type === CanvasSlot ? (
+      <canvas
+        style={canvasStyle}
+        ref={canvasRefs[(child as React.ReactElement).props.name]}
+      />
+    ) : (
+      child
+    )
+  );
+};
--- a/.codesandbox/templates/next/components/index.tsx
+++ b/.codesandbox/templates/next/components/index.tsx
@@ -18,9 +18,10 @@ const canvasStyle: React.CSSProperties = {
   userSelect: 'none',
 };
 
-export const CanvasSlot = ({ name }: { name: 'main' | 'top' | string }) => (
-  <div name={name} />
-);
+export const CanvasSlot = forwardRef<
+  HTMLCanvasElement,
+  { name: 'main' | 'top' | string }
+>((props, ref) => <canvas {...props} ref={ref} />);
 
 /**
  * @example Standard
@@ -80,6 +81,7 @@ export const CanvasComponent = forwardRef<
         )
       )
     );
+
     if (typeof ref === 'function') {
       ref(canvas);
     } else if (typeof ref === 'object' && ref) {
@@ -107,12 +109,10 @@ export const CanvasComponent = forwardRef<
       (child as React.ReactElement).type === CanvasSlot
     ) {
       const { name } = (child as React.ReactElement).props;
-      return (
-        <canvas
-          style={name === 'main' ? undefined : canvasStyle}
-          ref={canvasRefs[name]}
-        />
-      );
+      return React.cloneElement(child as React.ReactElement, {
+        style: name === 'main' ? undefined : canvasStyle,
+        ref: canvasRefs[name],
+      });
     } else {
       return child;
     }
--- a/.codesandbox/templates/next/components/index.tsx
+++ b/.codesandbox/templates/next/components/index.tsx
@@ -1,8 +1,11 @@
-import React, { createRef, useEffect, useMemo } from 'react';
+import React, { createRef, useEffect, useMemo, forwardRef } from 'react';
 import {
+  Canvas,
+  CanvasOptions,
   DOMManager,
   DOMManagerType,
   StaticCanvas,
+  StaticCanvasOptions,
   StaticDOMManagerType,
 } from 'fabric';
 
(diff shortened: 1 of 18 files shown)


.codesandbox/templates/next/components/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+} from 'fabric';
+
+const canvasStyle: React.CSSProperties = {
+  position: 'absolute',
+  left: 0,
+  top: 0,
+  touchAction: 'none',
+  msTouchAction: 'none',
+  userSelect: 'none',
+};
```
ShaMan123, line 16: not sure about the best way to handle this
should it be a default prop on CanvasSlot?

.codesandbox/templates/next/components/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+ *     <CanvasSlot name="top" />
+ *   </CanvasComponent>
+ *
+ * @example DOM elements in between
+ *   <CanvasComponent create={create}>
+ *     <CanvasSlot name="main" />
+ *     <CanvasSlot name="pointers" />
+ *     Hello World
+ *     <CanvasSlot name="top" />
+ *   </CanvasComponent>
```
ShaMan123, line 39: @jiayihu I believe you were looking for this

.codesandbox/templates/next/components/index.tsx
ShaMan123, line 1: This file should be our react binding

.codesandbox/templates/next/components/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+      React.Children.toArray(children).reduce((refs, child) => {
+        typeof child === 'object' &&
+          (child as React.ReactElement).type === CanvasSlot &&
+          (refs[(child as React.ReactElement).props.name] =
+            createRef<HTMLCanvasElement>());
+
+        return refs;
+      }, {} as Record<string, React.RefObject<HTMLCanvasElement>>),
+    [children]
+  );
```
ShaMan123, line 68: I hope this is good enough, do I need to clean it up or is GC enough?

.codesandbox/templates/next/components/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  return React.Children.map(children, (child) =>
+    typeof child === 'object' &&
+    (child as React.ReactElement).type === CanvasSlot ? (
+      <canvas
+        style={canvasStyle}
+        ref={canvasRefs[(child as React.ReactElement).props.name]}
+      />
+    ) : (
+      child
```
ShaMan123, line 98: do we want to apply `canvasStyle` to the other elements?

.codesandbox/templates/next/pages/index.tsx
```diff
@@ -1,60 +1,70 @@
...
+        <iframe
+          src="https://fabricjs.github.io/"
+          width="100%"
+          height="500px"
+          style={{ position: 'absolute', left: 0, top: 0, opacity: 0.8 }}
+        />
+        Hello World
+        {/* keep on top */}
+        <CanvasSlot name="top" />
+      </CanvasComponent>
```
ShaMan123, line 67: This looks really funny on chrome, opacity leaks from the iframe to the main context
Screenshot 2024-02-29 at 21 06 57

ShaMan123: same in safari

src/canvas/Canvas.ts
```diff
@@ -114,8 +115,13 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
 
-  constructor(el?: string | HTMLCanvasElement, options: TCanvasOptions = {}) {
-    super(el, options);
+  constructor(
+    arg0?: string | HTMLCanvasElement | DOMManagerType,
+    options: TCanvasOptions = {}
+  ) {
+    super(arg0, options);
+    // enable drag and drop
+    this.elements.items.top.el.setAttribute('draggable', 'true');
```
ShaMan123, line 124: moved from CanvasDOMManager

ShaMan123: not sure it is good here because my intention is to avoid polluting fabric with web specific stuff

src/canvas/DOMManagers/DOMManager.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  setDimensions(size: TSize, retinaScaling: number) {
+    Object.values(this.items).forEach(({ el, ctx }) =>
+      setCanvasDimensions(el, ctx, size, retinaScaling)
+    );
+  }
+
+  setCSSDimensions(size: Partial<CSSDimensions>) {
+    Object.values(this.items).forEach(({ el }) => setCSSDimensions(el, size));
+    this.container && setCSSDimensions(this.container, size);
```
ShaMan123, line 58: hmm... I have thoughts about this.
Is this legacy only? Is there a case you want to resize the container?

src/canvas/StaticCanvas.ts
```diff
@@ -183,17 +184,26 @@ export class StaticCanvas<
       (this.constructor as typeof StaticCanvas).getDefaults()
     );
     this.set(options);
-    this.initElements(el);
+    this.elements = this.initElements(arg0);
```
ShaMan123, line 187: breaking beta here as well

ShaMan123 (commented):
Ready for review

src/canvas/Canvas.ts
asturur: this are all things that needs to be discussed indeed.
Fabric is not polluted by web specific stuff. FabricJS is mainly web based, with a node backend abstracted through JSDOM. 
Correct?
So you can't avoid polluting something that is made in that way.

asturur:
I swear this is last time i bother you with this.
Please discuss plans/ changes/ideas before opening PRs.
In this way we are just slowing each other down and general progress.

src/canvas/DOMManagers/DOMManager.ts
asturur: You want to resize the container every time you want a different css size compared to the one that are given initially by the ration between screen pixels and css pixels.
When you want to enforce a different scaling that window pixel ratio

asturur:
```
The DOM mutation that fabric does is dirty and weird and counter intuitive. However it can remain.
```
I think there is a basic misunderstanding.
Fabric does DOM mutation as react or other frameworks do.
If you mix 2 frameworks you have issues.
You put your preference on React or Vue and so you see the mutation that fabric does as an obstacle.

src/canvas/DOMManagers/DOMManager.ts
ShaMan123: I am trying to remove it by not setting `position: absolute` on the main canvas element

src/canvas/Canvas.ts
ShaMan123: True
I would argue that in the long run we can consider dropping JSDOM so I am being cautious 
Moreover, since I want Offscreen support this is considered pollution

ShaMan123:
No confusion. I mean the parent swapping.

.codesandbox/templates/next/components/index.tsx
jiayihu: Not really, we create the rendering layers in the fabric code, so imperately, not declaratively in React. But it's definitely useful to have simplified APIs to create other rendering layers in fabric. For instance we might want to move rendering controls to its own layer for performance reasons, e.g. to support hovering controls without having to re-render everything.

src/canvas/Canvas.ts
asturur: yes it would be good to slowly pack dom support somewhere and have fabricjs domless. But don't write your code with that in mind, when it will be done we will have to change a lot of stuff anyway.
Think of your code so is small and clear and less code than what you got in your first iteration.

asturur:
> No confusion. I mean the parent swapping.


> > Please discuss plans/ changes/ideas before opening PRs.
> 
> I agree. Even though I'd be one of those who benefits from this PR, I'd not want a breaking change at this stage. Or it should definitely be discussed first.

you can also partecipate the discussion or read it, are not private discussions, i m trying to organize them on google docs but definetely they will be save the final doc in a md in github. But with all doc changes initally github is a mess

@ShaMan123 you have to pick if you want do discuss first this or context isolation.
This seems smaller, context isolation touches more points and probably requires other features to go in first.
Your choice, i m happy to discuss both since i don't dislike any of the features.
We can discuss anything in parallel, but we can implement one at time.

ShaMan123:
When I asked us to plan I meant to sit down and list all the bugs, features, PRs, thoughts, motivations etc. briefly and prioritize them, set a timeline and then move to the first item.
Until then we will continue conflict on this PR or that doc.
Without timelines I can't use fabric.

ShaMan123:
That first item might be the website.
In any case without a detailed plan it is hard for me to work freely

ShaMan123:
A simple example
I think one of the most important added values of a canvas abstraction API is geometry so for me if fabric doesn't have powerful geometry it is useless. That means many things and features. You may disagree. In that case I will be forced to make some decisions, one of them might be forking for good.

ShaMan123:
It has also come to my mind that you haven't read all my 80+ open PRs so you do not know what I have learned, done or discovered and how it affected how I view what fabric is. That is a big gap.

asturur:
> A simple example I think one of the most important added values of a canvas abstraction API is geometry so for me if fabric doesn't have powerful geometry it is useless. That means many things and features. You may disagree. In that case I will be forced to make some decisions, one of them might be forking for good.

You can fork, there is no issue with that.
If you don't like how i want to work please have a fork.

The geometry rework, if you manage to explain why should be done, will be done at minimum after there is a website and the 3-4 items that were pulled in the roadmap 2 years ago are done.


> It has also come to my mind that you haven't read all my 80+ open PRs so you do not know what I have learned, done or discovered and how it affected how I view what fabric is. That is a big gap.

I don't want read PRs before discussing what you want to change and why. I have been super clear about that, discussing in front of PR with poor explanations and having to add tons of comments on an interface that is made for code review and not intention or feature review is a big waste of time that i don't have

asturur:
> In any case without a detailed plan it is hard for me to work freely

There is no work freely since we are in 2.
I opted in requests you had for fabricJS and we did just those, we literally did nothing else.
I went through the lists of merged prs, yesterday and literally any new feature was requested and initialized by you.
If the issue that you can't do all the features that comes to your mind, sure that will be the case.

jiayihu:
@ShaMan123 you might indeed do a fork or separate release branch where things can move quick and freely. Then plan how to sync some features back to main fabric on a planned schedule. The works would also be more refined already, before being synced back to main fabric. Either a feature would have proved itself or the opposite.

Companies can then decide which branch/repo to follow. Product companies will likely desire to keep using main fabric because of stability, while however benefiting from potential new features on a planned schedule (e.g. every 3-6 months). Companies have to plan updating deps like fabric, that's why for instance Angular has a 6months well-known schedule and React never introduces breaking changes before a notice of **years**.

Startup companies may instead desire to a more experimental fabric, which re-designed APIs like coords and geometry, that could allow them to reach their goals sooner and have a competitive advantage.

On a personal side:
- That would leave you @ShaMan123  free to experiment and release on your own
- @asturur would have less burden on his shoulders, i.e. less PRs to review and discuss. Would just need to do what he's already doing, i.e. pick "which features should be picked for the next release in 6months?"

(Anyway I propose to have these personal discussions on Discord or GH Discussions)

ShaMan123:
I agree that didcussiins should be inside the repo, GH Discussions seems the best
Doesn't seem right to have elsewhere.
Rrgarding the rest. I disagree.
Once I fork I will diverge from fabric too much. Intoducing a feature back sounds simple perhaps but it won't be.
A single fix to geometry, e.g. strokeUniform, will break fabric.to pieces and might require json migration as you know.
Now imagine what I have in mind, making strokeUniform the tinest thing I will do.
Forking means departing.

jiayihu:
That would simply mean that big changes like geometry won't be synced back until for a planned major release, e.g. fabric v7. It would even make more sense for those changes to be planned very long ahead, so that whoever uses fabric can prepare for the migration.

The alternative would be for you to wait, but my understanding seemed that you were growing frustrated over the delay of PRs and breaking changes/features.

jiayihu:
That would simply mean that big changes like geometry won't be synced back until for a planned major release, e.g. fabric v7. It would even make more sense for those changes to be planned very long ahead, so that whoever uses fabric can prepare for the migration.

The alternative would be for you to wait, but my understanding seemed that you were growing frustrated over the delay of PRs and breaking changes/features.

Forking doesn't necessary mean departing. https://github.com/microsoft/react-native-windows for instance is a good example of not-departing fork

asturur:
Discussion of feature will be on a google doc that can be linked and advertised on a GH discussion, when the discussion seems stable or mostly assessed can become a readme.md more or less identical or being archived in the repo in some way as memory of the reasoning behind decisions/behaviour of code.

Also to be clear, is not that if anyone want to fix TSPAN in svg parsing, or add mask support based on current architecture you have to tell me anything, open a PR and do it.

But if you plan to refactor the whole svg parsing i want to listen before writing for 2 reasons:
- i spent time thinking of it already and i want to compare the opinions
- i want to see if the goal can be achieved without refactoring
- if we have to loose all the knowledge on the previous code it needs to be done with a net gain somewhere else

And i have been thinking those changes since long since my soft requests or working more organized didn't take any effects. I don't think my requests are unreasonable.
- pick a change/idea
- make a pitch about the idea, verify that is necessary/can't be achieved today
- agree on the solution
- someone is going to implment the code
- someone else is going to look at the code

Seems a possible way to work in group not one i made up

ShaMan123:
Though this PR has value for the dev after moving to use a rendering loop it is easy enough to override what is needed and avoid the DOMManager altogether.
Fabric should expose `setCanvasDimensions` and `setCSSDimensions` for ease of use

Issue #9679, opened 2024-02-22
[Bug]: `Cannot read properties of undefined (reading 'trim')` in  `loadSVGFromString` with `@media` in CSS


silverwind:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.0-beta19

### In What environments are you experiencing the problem?

Node.js

### Node Version (if applicable)

21.6.1

### Link To Reproduction

See simple node script below

### Steps To Reproduce

```js
import {loadSVGFromString} from 'fabric/node';

await loadSVGFromString(`
  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
    <style>
      circle { fill: black; }
      @media (prefers-color-scheme: dark) { circle { fill: white; } }
    </style>
    <circle r="10" cx="10" cy="10"/>
  </svg>
`);
```

### Expected Behavior

No error.

### Actual Behavior

Script errors. Observations:

- Removing the `@media` line makes it work.
- Fails in both `6.0.0-beta19` and `5.3.0`.

### Error Message & Stack Trace

```
TypeError: Cannot read properties of undefined (reading 'trim')
    at node_modules/fabric/dist/index.node.mjs:24637:27
    at Array.forEach (<anonymous>)
    at getCSSRules (node_modules/fabric/dist/index.node.mjs:24627:6)
    at new ElementsParser (node_modules/fabric/dist/index.node.mjs:24663:21)
    at parseSVGDocument (node_modules/fabric/dist/index.node.mjs:24822:25)
    at loadSVGFromString (node_modules/fabric/dist/index.node.mjs:24850:10)
    at fab.js:3:7
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
    at async loadESM (node:internal/process/esm_loader:28:7)
```

[...]

PR #9707, opened 2024-03-03
fix(SVGParser): Don't crash on nested CSS at-rules

silverwind:
## Description

This will ignore all CSS rules that contain unbalanced brackets after splitting on `}`.

- `circle { fill: black; }` splits to `circle { fill: black; ` which has one `{`, so it's okay.
- `@media (prefers-color-scheme: dark) { circle { fill: white; } }` splits to `@media (prefers-color-scheme: dark) { circle { fill: white; ` which has two `{` and starts with `@` so is ignored.

As mentioned in https://github.com/fabricjs/fabric.js/issues/9679 I think using a AST parser would be a much better option than this current crude parsing, for the future.

Ref: https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule
Fixes: https://github.com/fabricjs/fabric.js/issues/9679

### DIFF
--- a/src/parser/getCSSRules.ts
+++ b/src/parser/getCSSRules.ts
@@ -30,9 +30,10 @@ export function getCSSRules(doc: Document) {
       .filter((rule, index, array) => array.length > 1 && rule.trim())
       // at this point we have hopefully an array of rules `body { style code... `
       .forEach((rule) => {
-        // if there is more than one opening bracket, it is likely a nested CSS at-rule
-        // like @media, @supports, @scope, etc. Ignore these.
-        if ((rule.match(/{/g) || []).length > 1) {
+        // if there is more than one opening bracket and the rule starts with '@', it is likely
+        // a nested at-rule like @media, @supports, @scope, etc. Ignore these as the code below
+        // can not handle it.
+        if ((rule.match(/{/g) || []).length > 1 && rule.trim().startsWith('@')) {
           return;
         }
 
--- a/src/parser/getCSSRules.ts
+++ b/src/parser/getCSSRules.ts
@@ -33,7 +33,7 @@ export function getCSSRules(doc: Document) {
         // if there is more than one opening bracket and the rule starts with '@', it is likely
         // a nested at-rule like @media, @supports, @scope, etc. Ignore these as the code below
         // can not handle it.
-        if ((rule.match(/{/g) || []).length > 1 && rule.trim().startsWith('@')) {
+        if ((rule.match(/{/g) || []).length > 1 && rule.trimStart().startsWith('@')) {
           return;
         }
 
--- a/src/parser/getCSSRules.ts
+++ b/src/parser/getCSSRules.ts
@@ -33,7 +33,7 @@ export function getCSSRules(doc: Document) {
         // if there is more than one opening bracket and the rule starts with '@', it is likely
         // a nested at-rule like @media, @supports, @scope, etc. Ignore these as the code below
         // can not handle it.
-        if ((rule.match(/{/g) || []).length > 1 && rule.trimStart().startsWith('@')) {
+        if ((rule.match(/{/g) || []).length > 1 && rule.trim().startsWith('@')) {
           return;
         }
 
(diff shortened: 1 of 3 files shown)


silverwind:
I also verified this fix in the [actual use case](https://github.com/go-gitea/gitea/blob/main/build/generate-images.js), and it did generate the black circle as expected.

silverwind:
I added a condition so it looks for the `@` to only ignore at-rules. There is also [nesting syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector) in CSS which would still break it. If you prefer to also ignore this nesting syntax, we can remove the `@` condition.

src/parser/getCSSRules.ts
```diff
@@ -30,6 +30,13 @@ export function getCSSRules(doc: Document) {
       .filter((rule, index, array) => array.length > 1 && rule.trim())
       // at this point we have hopefully an array of rules `body { style code... `
       .forEach((rule) => {
+        // if there is more than one opening bracket and the rule starts with '@', it is likely
+        // a nested at-rule like @media, @supports, @scope, etc. Ignore these as the code below
+        // can not handle it.
+        if ((rule.match(/{/g) || []).length > 1 && rule.trimStart().startsWith('@')) {
+          return;
+        }
```
asturur, line 38: This is our current browserlists for this release:
```
chrome >= 64
safari >= 11
firefox >= 58
```

trimStart doesn't  fit.
It doesn't matter how we ended up on this list of browsers, we have this for 6.0 and now is not the moment to change them.
Can you use trim? it should make the job the same way in our case that we are testing 'startsWith'

silverwind: I had `.trim` before but thought it would be slightly inefficient, I will revert to `.trim`.

src/parser/getCSSRules.ts
```diff
@@ -30,6 +30,13 @@ export function getCSSRules(doc: Document) {
       .filter((rule, index, array) => array.length > 1 && rule.trim())
       // at this point we have hopefully an array of rules `body { style code... `
       .forEach((rule) => {
+        // if there is more than one opening bracket and the rule starts with '@', it is likely
+        // a nested at-rule like @media, @supports, @scope, etc. Ignore these as the code below
+        // can not handle it.
+        if ((rule.match(/{/g) || []).length > 1 && rule.trimStart().startsWith('@')) {
```
silverwind, line 36:
```suggestion
        if ((rule.match(/{/g) || []).length > 1 && rule.trim().startsWith('@')) {
```

src/parser/getCSSRules.ts
asturur: I doubt efficiency is at stake here, this code is already bad.

asturur (dismissed):

asturur:
i need to fix prettier or you can do it.
npm run prettier:write

silverwind:
Ran it.

silverwind:
> I added a condition so it looks for the `@` to only ignore at-rules. There is also [nesting syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector) in CSS which would still break the parser. If you prefer to also ignore this nesting syntax, I can remove the `@` condition again.

What do you recommend about this one? Should we ignore non-at-rule nesting as well? I don't really understand the purpose of the parser and whether it's better to crash or ignore.

asturur:
The parser has been written in 2010-2012 and supported ie7+ initially through a canvas polyfill.
It has been kept alive since then.
The scope of SVG parser in fabricJS is to being able to import SVGs as shapes rather than images. That is its scope.
We declare we support svg specs up version 1.2, so if something is above that version we can support it optionally when is easy.
I do not know which is the meaning of SVG versions, so if let's say SVG 1.2 supports style tag in 2012 then it means that it will support every new css syntax in 2024 but still being called SVG 1.2

In general all that can be fixed will be fixed also depending on the popularity of the feature.
How many softwares output svgs with nested css syntax? 
Most of them just spit out a mess of attributes or at best single unique named classes.

asturur (approved):

PR #9716, opened 2024-03-07
ci(): Run the stat action from forkm

asturur:
## Description

This is another attempt at having this working


asturur:
ok i can't test a pull_request_target action from a pull_request for obvious reasons.

PR #9718, opened 2024-03-09
ci(): Move the changelog action on pull_request_target to allow write permission

asturur:
## Description

Hopefully this work as per the build stats.
There is the added  indirection of the workflow call that needs to be tested


asturur:
Merging for a test because i don't understand why this doesn't show up at all in pull_request_target mode.
Reverting if it doesn't solve anything

PR #9725, opened 2024-03-12
chore(TS): minor changes to typescript notation to be compatible with a 5.3.3

asturur:
## Description

typedoc uses TS 5.3.3 that gets angry at some our of ts-expect-error that are no more error in 5.3.3
There is now way to disable those because those are meant to inform you what to do when you migrate.

In order to make typedoc working i need to swap those with just an ignore and leave the original comment on top of it for migration.

This PR also remove the old typedoc plugin and template, but not the configuration file that is used by astro to produce the typedocs


src/shapes/Image.ts
```diff
@@ -602,7 +602,8 @@ export class FabricImage<
     ctx: CanvasRenderingContext2D
   ) {
     ctx.imageSmoothingEnabled = this.imageSmoothing;
-    // @ts-expect-error TS doesn't respect this type casting
+    // cant use ts-expect-error because of ts 5.3 cross check
+    // @ts-ignore TS doesn't respect this type casting
     super.drawCacheOnCanvas(ctx);
```
jiayihu, line 607: This on the called method would fix the issue without the need of silencing the error:

```ts
drawCacheOnCanvas(ctx: CanvasRenderingContext2D) {
    const self = this as TCachedFabricObject;

    ctx.scale(1 / self.zoomX, 1 / self.zoomY);
    ctx.drawImage(
      self._cacheCanvas,
      -self.cacheTranslationX,
      -self.cacheTranslationY
    );
}
```

asturur: It makes me mad that typescript has to error (not warn!!!!) on unused error directives.
So you can silence errors ( dangerous ) but you can't silence error silencers for good code.

jiayihu: I'm not following you. If you use `ts-expect-error` then I like that it errors if there isn't actually a type error. What you're trying to do with typedoc while keeping the current TS version is not really a common use case for them.

asturur: Yes i know is not common, but is still silly that i cant keep that as a warning and has to block the compilation.

jiayihu: Dows typedoc allow to specify the TS instead? So that it's aligned with fabric

asturur: maybe. But this is the shortest and more logical path.
No code changes, just annotation changes.
Not fiddling with dependencies of dependencies of astro, just make it run.

PR #9732, opened 2024-03-14
fix(ActiveSelection): block ancestors/descendants of selected objects from being selected

ShaMan123:
## Description

Since ActiveSelection is a special Group and doesn't call `parent.remove` when an object is added there is a case where it is possible to have ancestors and descendants live together in the selection. This is a bug because it breaks the object tree by making it cyclic.

### DIFF
--- a/src/shapes/ActiveSelection.ts
+++ b/src/shapes/ActiveSelection.ts
@@ -87,6 +87,18 @@ export class ActiveSelection extends Group {
   }
 
   /**
+   * @override block ancestors/descendants of selected objects from being selected
+   */
+  canEnterGroup(object: FabricObject) {
+    return (
+      super.canEnterGroup(object) &&
+      !this.getObjects().some(
+        (o) => o.isDescendantOf(object) || object.isDescendantOf(o)
+      )
+    );
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


src/shapes/ActiveSelection.ts
```diff
@@ -86,6 +87,26 @@ export class ActiveSelection extends Group {
     }
   }
 
+  /**
+   * @override block ancestors/descendants of selected objects from being selected to prevent a circular object tree
+   */
+  canEnterGroup(object: FabricObject) {
+    if (
+      this.getObjects().some(
+        (o) => o.isDescendantOf(object) || object.isDescendantOf(o)
```
ShaMan123, line 96: Can change to the following as well, don't know what is better

```suggestion
        (o) => object.hasCommonAncestors(o)
```

ShaMan123: Better as is

asturur (approved):

src/shapes/ActiveSelection.ts
ShaMan123: Now I understand this is not the expected behavior.
If an ancestor is trying to enter the descendant should exit

PR #9741, opened 2024-03-18
fix(): bubble dirty flag to parent

ShaMan123:
## Description

As part of the group rewrite we exposed a `parent` ref on the object #8951 .
As part of the layout manager fixes an invalidation bug surfaced #9651 .
Fabric relays on `Object#_set` to bubble cache invalidation to the **parent**.
I forgot to fix the invalidation logic when I exposed the parent ref so it still invalidated the group ref.
The bug occurs when a nested object is part of an `ActiveSelection`.
In that case the group ref is the active selection and the parent ref is the group.
 `ActiveSelection` does not render anything, therefore it doesn't cache meaning that it is pointless to flag it as dirty as opposed to the parent.
This PR will fix the mentioned bug.
I also changed the bubbling logic to always bubble a dirty flag regardless of caching state.

### DIFF
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -454,7 +454,7 @@ export class Group
    * @return {Boolean}
    */
   isOnACache(): boolean {
-    return this.ownCaching || (!!this.group && this.group.isOnACache());
+    return this.ownCaching || (!!this.parent && this.parent.isOnACache());
   }
 
   /**
--- a/src/shapes/Object/Object.ts
+++ b/src/shapes/Object/Object.ts
@@ -722,8 +722,6 @@ export class FabricObject<
    * @param {*} value
    */
   _set(key: string, value: any) {
-    const isChanged = this[key as keyof this] !== value;
-
     if (key === 'scaleX' || key === 'scaleY') {
       value = this._constrainScale(value);
     }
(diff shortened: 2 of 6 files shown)


ShaMan123 (commented):
@asturur please test this with your demo of #9651

src/shapes/Group.ts
```diff
@@ -454,7 +454,7 @@ export class Group
    * @return {Boolean}
    */
   isOnACache(): boolean {
-    return this.ownCaching || (!!this.group && this.group.isOnACache());
+    return this.ownCaching || (!!this.parent && this.parent.isOnACache());
```
asturur, line 457: most of the time parent is unset, this would breaking the detection of isOnAcache for the objects that are not part of a multi selection.
This maybe should do something like

`const container = this.parent || this.group`

And then check the isOnACache of container?

src/shapes/Object/Object.ts
```diff
@@ -736,29 +734,29 @@ export class FabricObject<
...
+    // a dirty child makes the parent dirty.
+    // but a non dirty child does not make the parent not dirty.
+    // the parent could be dirty for some other reason.
+    this.parent &&
+      (this.dirty ||
+        (isChanged &&
+          (this.constructor as typeof FabricObject).stateProperties.includes(
+            key
+          ))) &&
+      this.parent._set('dirty', true);
```
asturur, line 758: Here i have the same exact concern, group is our main property, while parent is just an exception sometimes.
So we need to check parent or group, with precedence to parent.

Also i think is ok to skip `const groupNeedsUpdate = this.group && this.group.isOnACache();` is a bit of hit and miss. Is true we don't care if the group is not cached to set it as dirty, since is just a boolean, but if the group is cached on a different cache than the object, at that point the invalidation isn't necessary and is a cost. ( reading the code today this check isn't made, but if say we are going always to flag a group as dirty we are cutting that away )

What is the reason behind changing the dirty flag? what brought you to change it?

asturur: Yes you are right the separate cache is going to make the lower cache dirty anyway.
So the issue is just parent vs group

src/shapes/Group.ts
asturur: Ok because ActiveSelection has its own specific exitGroup.
The division between enterGroup, exitGroup, _enterGroup and _exitGroup are confusing at best.

asturur:
> @asturur please test this with your demo of #9651

I would not know what you want to test with that demo.

ShaMan123:
To see that the dirty flag is working as expected with nested selected objects

src/shapes/Group.ts
ShaMan123: True

asturur:
Just check it out from the branch and add it to this branch.
Probably one or 2 cases are fixed with the set({ top, left  }) and this pr.
But this is going to refresh the cache and not adjust the layout too. So you would see the object moving but i m not sure you would see the layout adapting

asturur (approved):
I m ok with this PR as long the test you want to do are done on that template.

ShaMan123:
tested

Issue #9746, opened 2024-03-19
[bug] clippath / svg_import I have an svg file that renders properly in a browser but won't render in Kitchensink properly either grouped or ungrouped. Any ides why?


asturur:
I have an svg file that renders properly in a browser but won't render in Kitchensink properly either grouped or ungrouped. Any ides why?
![Sea_Group_logo (2)](https://github.com/fabricjs/fabric.js/assets/17964201/d8dc3612-26eb-4734-a195-da77d9bc032f)

Here is the svg content:

> 
```

<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:cc="http://creativecommons.org/ns#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   sodipodi:docname="Sea_Group_logo.svg"
   inkscape:version="1.0 (4035a4fb49, 2020-05-01)"
   id="svg8"
   version="1.1"
   viewBox="0 0 288.76712 118.08024"
   height="118.08024mm"
   width="288.76712mm">
  <defs
     id="defs2">
    <clipPath
       id="clipPath141"
       clipPathUnits="userSpaceOnUse">
      <path
         id="path139"
         d="m 20.958,2463.023 c 0,-8.102 6.569,-14.67 14.67,-14.67 v 0 c 8.103,0 14.67,6.568 14.67,14.67 v 0 c 0,8.102 -6.567,14.67 -14.67,14.67 v 0 c -8.101,0 -14.67,-6.568 -14.67,-14.67" />
    </clipPath>
  </defs>
  <sodipodi:namedview
     inkscape:window-maximized="1"
     inkscape:window-y="-8"
     inkscape:window-x="-8"
     inkscape:window-height="715"
     inkscape:window-width="1366"
     showgrid="false"
     inkscape:document-rotation="0"
     inkscape:current-layer="layer1"
     inkscape:document-units="mm"
     inkscape:cy="211.71519"
     inkscape:cx="697.91761"
     inkscape:zoom="0.35"
     inkscape:pageshadow="2"
[...]

PR #9750, opened 2024-03-20
fix(SVGImport): Consider the transformMatrix of the clipPath owner as part of the clipPath trasnformation

asturur:
## Description

Fixes the parsing of the SVG reported in https://github.com/fabricjs/fabric.js/issues/9746 and closes it

Tested with a visual test.


src/parser/parseSVGDocument.ts
```diff
@@ -67,8 +68,9 @@ export async function parseSVGDocument(
   }
   const localClipPaths: Record<string, Element[]> = {};
   descendants
-    .filter((el) => el.nodeName.replace('svg:', '') === 'clipPath')
+    .filter((el) => getTagName(el) === 'clipPath')
     .forEach((el) => {
+      el.setAttribute('originalTransform', el.getAttribute('transform') || '');
```
asturur, line 73: We need to save the original transform of this element because we may cycle over it multiple times and we need to then change its transform attribute.
We could do this ONCE instead of multiple times but that is a larger refactor not in scope for a fix.

asturur:
The clipPath is blurry anyway for a long standing bug

ShaMan123 (approved):
I don't understand exactly how this works, how does the transform get applied etc.?
But indeed a cool fix

test/visual/golden/seaClipPath.png
ShaMan123, line 1: I bet that if we fix `getObjectScaling` to include the clip path owner this will be fixed

asturur: there is a PR that fix clippath blurryness but had a strange issue with a particular kind of clipPath.
I don't remember anymore what it was since it was 4 years ago.

It was because of how caching work more than a scaling factor but indeed the fix was to include object scaling in the clipPath cache or something on those lines. Probably what was missing was readjusting the center

asturur:
> I don't understand exactly how this works, how does the transform get applied etc.?

The transform get applied from the 'fromElement' static method.
It parses the attributes of the object going from the svg element upward to the svg outer box.
The reason why i put the trasnform on the clipPath as an attribute of svg is because so i can just tap in that process rather than calculting a custom matrix and then have to inject it with retroactive effect once the clipPath is parsed.

src/parser/elements_parser.ts
```diff
@@ -139,16 +143,32 @@ export class ElementsParser {
...
+      const finalTransform = parseTransformAttribute(
+        `${clipPathOwner.getAttribute('transform') || ''} ${
+          clipPathTag.getAttribute('originalTransform') || ''
+        }`
+      );
+
+      clipPathTag.setAttribute(
+        'transform',
+        `matrix(${finalTransform.join(',')})`
+      );
```
asturur, line 170: @mikehollibaugh this is part 1 of the fix. You can find a function similar to this in 4.x

src/parser/parseSVGDocument.ts
asturur: This is part 2 of the fix, even this you can still find it in 4.x @mikehollibaugh

PR #9757, opened 2024-03-22
Update rotate.ts

jehaad1:
There's no need to check if the angle is a negative value when there's Math.abs.


asturur:
That is not the intent of the code.
The code is not:

```
if (a < 0 ) a = -a;
```

If angle is -5, what we want is to get 355.

Mod of -5 by 360 would give us -5 and with Math.abs 5.

Instead we need to first rotate the angle of 360 degree so that we go in positive space, then we can mod.

jehaad1:
ohh I'm sorry, I just want to help

PR #9762, opened 2024-03-24
Fix searchPossibleTargets for non-interactive nested targets

jiayihu:
`searchPossibleTargets` should return the innermost subTarget, but respecting intermediate `interactive` groups. Currently if the `target` is interactive, it will automatically return `this.targets[0]`, which is the innermost subTarget, even though it could belong to non interactive group. See the test.

This correctly allows subselection to not enter automatically to the innermost child, rather to be determined by the stack of interactive groups.


asturur:
Those are all the things that i would have liked to think before thinking interactive groups were doable.
This can go out as a fix soon if is correct and if is not prone to interpretations of what should/could/would be the correct behaviour, and if a correct behaviour exists.
I ll put time
aside for this tuesday, monday is already filled up with fabricjs things to do.

src/canvas/SelectableCanvas.ts
```diff
@@ -860,15 +860,27 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
-      ? this.targets[0]
-      : target;
+    ) {
+      /** {@link this.targets} is ordered bottom-up */
+      let subTarget = this.targets[0];
+      while (subTarget.parent && !subTarget.parent.interactive) {
+        subTarget = subTarget.parent;
+      }
+
+      return subTarget;
```
ShaMan123, line 880: I understand you don't want to break anything but... this feels a bit too much. Seems that `targets` is not fitting enough.
Also I am not 100% sure this is what is expected for all cases.
Consider 2 siblings under an interactive group. A is an interactive group itself, B is a non interactive group. B is on top. Both have overlapping children A1, B1 at some point P.
Let's say we are searching targets on P. Who should be returned, A1 or B?
This code returns B.
I think that is correct.

ShaMan123: But what if B's onSelect returns true, signaling to prevent the selection?
Then I think A1 should be returned. This code is not doing that.

ShaMan123: I am reminding the PR that refactors this method and returns an array without populating `targets`. I understand we want v6 out. However I am surfacing it for consideration.

ShaMan123: Perhaps a sound/robust method would return 2 arrays: subTargets, selectionTargets

ShaMan123: And then we should define strictly that a subTarget is not a selection target

jiayihu: > But what if B's onSelect returns true, signaling to prevent the selection?
Then I think A1 should be returned. This code is not doing that.

I don't agree. `onSelect` only determines if the object should be selected once the decision is already finalised, it doesn't determine if the object should be picked as potential candidate as target. That is decided in `_searchPossibleTargets` and `_checkTarget`.

I agree that `targets` is not perfect. I haven't even opened an issue about the order in/out events are fired (bottom-up in fabric, wrong IMO) because that's more subject to discussion. But indeed those bigger changes are for after v6 since are more significant and require an in-depth discussion.

This PR instead just focuses on fixing a behaviour as intended in current beta without any breaking change.

asturur:
I have this second inline after the typing prs i want to understand id we need to get here before release

jiayihu:
I think this fix should belong to the release since it fixes a bug in the intended behaviour and it has very limited scope

ShaMan123:
I agree

src/canvas/SelectableCanvas.ts
ShaMan123: What if you use
>  this.targets.find(target => target.parent?.interactive) ?? target

I think it is more correct

ShaMan123: And please use something different than bottom up which is very confusing IMO.
Front to back maybe?

jiayihu: I like bottom-up, I find it more clear than "front-to-back" which sounds more about zIndex, i.e. stack between siblings. What do you prefer @asturur ?

asturur:
> I think this fix should belong to the release since it fixes a bug in the intended behaviour and it has very limited scope

Yeah i have to read it deeply.
I didn't.

src/canvas/SelectableCanvas.ts
asturur: both bottom up and front-back are confusing to me, i was indeed trying to understand what does it mean. Does it mean from inner to outer?

asturur:
Ok so my understanding is:
- today we collect subTargetChecks during `_searchPossibleTargets` looping over groups that have 'subTargetCheck' enabled.
- the public facing method `searchPossibleTargets`  will return the innermost if the group is interactive for nested selection without verifying that all the chain of subTargeting was interactive too.
- this would lead to a possible bug where an interactive group, shielded by a non interactive one, can still be sub-selected with a click

So i agree there is a bug here, i m updated with the discussion. tomorrow morning i ll get back to it.

This is an edge case that can be easily fixed, this solution isn't bad.
I read @ShaMan123 point but as mentioned in theory onSelect is a post selection issue, once something has been selected for a selection, onSelect can cancel the selection, but not let the selection process take an alternative road, if you want to do that you have to set that object as 'evented: false', so i wouldn't consider that case as a measure of this fix being a bad or good idea

src/canvas/SelectableCanvas.ts
ShaMan123: I think we should use descendant-ancestor
Anything else is open to confusion

ShaMan123: > What if you use
> 
> > this.targets.find(target => target.parent?.interactive) ?? target
> 
> I think it is more correct

What about this? I think it is more straight forward

jiayihu: > I think we should use descendant-ancestor

👍 

> this.targets.find(target => target.parent?.interactive) ?? target

👍

jiayihu:
> this would lead to a possible bug where an interactive group, shielded by a non interactive one, can still be sub-selected with a click

Yes exactly

src/canvas/SelectableCanvas.ts
asturur: mmmm this code here is tricky.
It opens to more edge cases.
This will stop at finding the fist one that has an interactive parent.
It doesn't tell us anything of all the other outer objects. Unless i m misreading it.
```
this.targets.find(target => target.parent?.interactive) ?? target
```
I think you have to start from the outermost and dig in by interactivity, i see no other way if not redo the same line of checks the original code was doing

asturur: I also think current code is not correct.
You have to cycle from the outermost and check. let me try to suggest

asturur, line 880:
```suggestion
      /** targets[0] is the innermost nested target, but it could be inside non interactive groups and so not a selection target */
      const targets = this.targets;
      for (let i = targets.length - 1; i > 0; i--) {
        if (!targets[i].interactive) {
          // one of the subtargets was not interactive. that is the last subtarget we can return.
          // we can't dig more deep;
          return targets[i];
        }
      }
      return targets[0];
```

This is my version of this problem @jiayihu  @ShaMan123

asturur:
Please let's examine the correctness of the code and let's move forward i think we are all on board now

jiayihu:
Pushed with suggestion and slight change to make TS happy

asturur:
I ll add a test to verify my thinking was correct. And if there are no complains i ll merge tomorrow morning. If there are complains post merge we can still fix

src/canvas/SelectableCanvas.ts
```diff
@@ -860,15 +860,31 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+      const targets = this.targets;
+      for (let i = targets.length - 1; i > 0; i--) {
+        const t = targets[i];
+        if (!(isCollection(t) && t.interactive)) {
+          // one of the subtargets was not interactive. that is the last subtarget we can return.
+          // we can't dig more deep;
+          return t;
+        }
+      }
+      return targets[0];
```
ShaMan123, line 884: I am not sure about this.
I think if no target in `targets` is found `target` should be returned.

src/canvas/SelectableCanvas.ts
```diff
@@ -188,7 +188,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
   declare fireMiddleClick: boolean;
 
   /**
-   * Keep track of the subTargets for Mouse Events
+   * Keep track of the subTargets for Mouse Events, ordered bottom up from innermost subTarget
```
ShaMan123, line 191:
```suggestion
   * Keep track of the subTargets for Mouse Events, ordered bottom up from innermost nested subTarget
```

src/canvas/SelectableCanvas.ts
asturur: is because i stop at i > 0 instead of i >= 0.
I did this so i don't have to consider the edge case in which the last one is not a collection.
My suggestions also was just `isCollection(t) && !t.interactive` because i thought there is no way the items before the last are not collections. only targets[0] can be not a collection

asturur (approved):

PR #9770, opened 2024-03-27
chore(TS): control generic object type

ShaMan123:
## Description

This PR makes it easier to use a control for a specific object class.
Instead of using type casting everywhere the object ref is accessed we pass a generic.

### Desired Behavior

```typescript
export class MyPathControl extends fabric.Control<fabric.Path> {
  actionHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Path>
  ) => {
    transform.target.path;
    return false;
  };

  mouseDownHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Path>
  ) => {
    transform.target.pathOffset;
  };
}

export class MyPolyControl extends fabric.Control<fabric.Polyline> {
  actionHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Polyline>
  ) => {
    transform.target.points;
    return false;
  };

  mouseDownHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Polyline>
  ) => {
    transform.target;
  };
}
```

https://github.com/fabricjs/fabric.js/assets/34343793/0c2298bb-bef9-4344-836b-1c829fce4c96

### Current Behavior

```typescript
export class MyPathControl extends fabric.Control {
  actionHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform
  ) => {
    (transform.target as fabric.Path).path;
    return false;
  };

  mouseDownHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Path>
  ) => {
    (transform.target as fabric.Path).pathOffset;
  };
}
```

### DIFF
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -21,7 +21,7 @@ export type TOptionalModifierKey = ModifierKey | null | undefined;
 
 export type TPointerEvent = MouseEvent | TouchEvent | PointerEvent;
 
-export type TransformAction<T extends Transform = Transform, R = void> = (
+export type TransformAction<T extends Transform<any> = Transform, R = void> = (
   eventData: TPointerEvent,
   transform: T,
   x: number,
@@ -32,14 +32,17 @@ export type TransformAction<T extends Transform = Transform, R = void> = (
  * Control handlers that define a transformation
  * Those handlers run when the user starts a transform and during a transform
  */
-export type TransformActionHandler<T extends Transform = Transform> =
+export type TransformActionHandler<T extends Transform<any> = Transform> =
   TransformAction<T, boolean>;
 
 /**
  * Control handlers that run on control click/down/up
  * Those handlers run with or without a transform defined
  */
-export type ControlActionHandler = TransformAction<Transform, any>;
+export type ControlActionHandler<T extends Transform<any>> = TransformAction<
+  T,
+  any
+>;
 
 export type ControlCallback<R = void> = (
   eventData: TPointerEvent,
@@ -53,10 +56,15 @@ export type ControlCursorCallback = ControlCallback<string>;
  * relative to target's containing coordinate plane
  * both agree on every point
  */
-export type Transform = {
-  target: FabricObject;
+export type Transform<T extends FabricObject = FabricObject> = {
+  target: T;
   action?: string;
-  actionHandler?: TransformActionHandler;
+  actionHandler?: (
+    eventData: TPointerEvent,
+    transform: Transform<T>,
+    x: number,
+    y: number
+  ) => boolean;
   corner: string;
   scaleX: number;
   scaleY: number;
--- a/src/shapes/Object/InteractiveObject.ts
+++ b/src/shapes/Object/InteractiveObject.ts
@@ -25,8 +25,6 @@ export type TOCoord = Point & {
   touchCorner: TCornerPoint;
 };
 
-export type TControlSet = Record<string, Control>;
-
 export type TBorderRenderingStyleOverride = Partial<
   Pick<InteractiveFabricObject, 'borderColor' | 'borderDashArray'>
 >;
@@ -114,7 +112,7 @@ export class InteractiveFabricObject<
    * holds the controls for the object.
    * controls are added by default_controls.js
    */
-  declare controls: TControlSet;
+  declare controls: Record<string, Control>;
 
   /**
    * internal boolean to signal the code that the object is
@@ -315,11 +313,7 @@ export class InteractiveFabricObject<
    * @param {Function} fn function to iterate over the controls over
    */
   forEachControl(
-    fn: (
-      control: Control,
-      key: string,
-      fabricObject: InteractiveFabricObject
-    ) => any
+    fn: (control: Control, key: string, fabricObject: this) => any
   ) {
     for (const i in this.controls) {
       fn(this.controls[i], i, this);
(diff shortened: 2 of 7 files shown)


ShaMan123 (commented):
Maybe this is pointless as is because `Transform` needs to be typed as well and that is causing all sorts of TS errors

src/EventTypeDefs.ts
```diff
@@ -53,10 +56,15 @@ export type ControlCursorCallback = ControlCallback<string>;
...
  */
-export type Transform = {
-  target: FabricObject;
+export type Transform<T extends FabricObject = FabricObject> = {
+  target: T;
   action?: string;
-  actionHandler?: TransformActionHandler;
+  actionHandler?: (
+    eventData: TPointerEvent,
+    transform: Transform<T>,
```
ShaMan123, line 64: any clues on handling this cyclic declaration @jiayihu ?

src/shapes/Object/InteractiveObject.ts
ShaMan123, line 1: I do not like the casting in this file but it is the easy way out to support generic controls

ShaMan123 (commented):
Ready for review

asturur:
@ShaMan123 i m sure this make easier something, but what exactly?
Can you make an example of what changes and how? i swear from code changes is not clear just by reading quickly

ShaMan123:
### Desired Behavior

```typescript
export class MyPathControl extends fabric.Control<fabric.Path> {
  actionHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Path>
  ) => {
    transform.target.path;
    return false;
  };

  mouseDownHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Path>
  ) => {
    transform.target.pathOffset;
  };
}

export class MyPolyControl extends fabric.Control<fabric.Polyline> {
  actionHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Polyline>
  ) => {
    transform.target.points;
    return false;
  };

  mouseDownHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Polyline>
  ) => {
    transform.target;
  };
}
```


https://github.com/fabricjs/fabric.js/assets/34343793/0c2298bb-bef9-4344-836b-1c829fce4c96


### Current Behavior


```typescript
export class MyPathControl extends fabric.Control {
  actionHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform
  ) => {
    (transform.target as fabric.Path).path;
    return false;
  };

  mouseDownHandler = (
    e: fabric.TPointerEvent,
    transform: fabric.Transform<fabric.Path>
  ) => {
    (transform.target as fabric.Path).pathOffset;
  };
}
```

jiayihu:
tldr: don't get fancy with TS in fabric. Leave it simple.

I don't honestly see much of a difference to using a generic instead of a type assertion in this case. You still have to specify everytime the generic, as shown in your example. Actually in my experience it would lead maybe to more troubles because you have now formally a generic type in some method parameters.

Since fabric heavily relies on runtime behaviour, trying to type it more strictly will often just result in more headache. In this case, the actual runtime behaviour is indeed that `transform.target` could be any FabricObject. You might say "No! It's always only the FabricObject associated with the Control" but I'd it still technically relies on runtime behaviour. For instance `_findTargetCorner` could be overridden to return a different FabricObject, as you know @ShaMan123 

Using a generic will only lead you to believe it's the FabricObject class you typically expect, whereas you'd be safer not assuming that. And if you're really sure it can't be any FabricObject, then I believe a type assertion would suffice.

Typings more strictly in TS is a difficult business, you either do it 100% right or you get an even worse result, because types got just more complex and it will even more difficult to override it on consumer side.

The `getAncestor` types are a perfect example of this. There hasn't been once that I used `obj.getAncestor(true/false)` and it just worked. Often I had to either do `obj.getAncestor(true) as fabric.Group[]` or even worse in the case of `obj.getAncestor(false)`. 

Screenshot 2024-03-28 at 12 39 02

jiayihu:
`toObject` https://github.com/fabricjs/fabric.js/pull/9014 was another example where generics just went wrong.

ShaMan123:
sad news
a step further for being incompatible for day to day practices

asturur:
> sad news a step further for being incompatible for day to day practices

Can you explain?

src/EventTypeDefs.ts
asturur: What is the issue with leaving `TransformActionHandler` here?

asturur:
I don't think i have a strong opinion here.
Typescript in general is below my expectations, i understand @jiayihu point of view but also @ShaMan123 idea of having the control possibly typed to the object it is designed for ( this change seems to me just an extension of what we have and shouldn't block anyone more than is already blocked with our types ).
So guys if you manage to land on a shared point of view we can take that one

jiayihu:
The common ground would be to use a generic for `Control`, defaulting to `FabricObject` in fabric internal code, but leave `Transform` as it is, which would also avoid the circular type ref.  
Then one can specify a stricter object class in the `Control` class if they think the runtime behaviour will be respected.

asturur:
This week i have been very busy and i tried to keep up with the messages and do some work on items i cared.
This PR could be merged before release if we can handle quickly, or it can wait. What are your feelings? i would like to release or at least cut an RC.

We also never talked about type changes being a breaking changes or not.

ShaMan123:
Let it wait
IMO types are never breaking, always a patch

asturur:
ok so i pull out of my list and i hope you come to an idea that satisfies both. Ping me when you are done

jiayihu:
> IMO types are never breaking, always a patch

I agree. Types are formally breaking, but practically not. TS itself often released changes that introduced new type errors due to more strictness in minor versions.

PR #9772, opened 2024-03-28
fix(): safeguard stack operations from foreign objects

### LINKED ISSUE #9730: [v6]: Support to ActiveSelection stack operation is gone
Support to ActiveSelection stack operations is gone in v6
In v5 you could do `activeSelection.sendBackwards()`. Group was not interactive so all selected objects were top level objects, direct descendants of canvas, so it made sense to handle moving them in the stack.
However, in v6 a selection can contain objects from different levels of the object tree, performing stack operations on the selection will involve different stacks. It can be supported but it is a question if it should.

The issue is that it is breaking to:
`canvas.sendBackwards(activeSelection)` 

I think it might be best to make the following condition stricter:
https://github.com/fabricjs/fabric.js/blob/631fe9bd7bd529a32c67f58a2b790b785ff04cf0/src/Collection.ts#L219

into `idx > 0`

Screenshot 2024-03-13 at 13 26 40

ShaMan123:
## Description

closes #9730 
Changed the conditions to be stricter to ban non children from entering the stack when called from stack operations

### DIFF
--- a/src/Collection.ts
+++ b/src/Collection.ts
@@ -176,7 +176,7 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
      * @returns {boolean} true if change occurred
      */
     sendObjectToBack(object: FabricObject) {
-      if (!object || object === this._objects[0]) {
+      if (object === this._objects[0] || !this._objects.includes(object)) {
         return false;
       }
       removeFromArray(this._objects, object);
@@ -192,7 +192,10 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
      * @returns {boolean} true if change occurred
      */
     bringObjectToFront(object: FabricObject) {
-      if (!object || object === this._objects[this._objects.length - 1]) {
+      if (
+        object === this._objects[this._objects.length - 1] ||
+        !this._objects.includes(object)
+      ) {
         return false;
       }
       removeFromArray(this._objects, object);
@@ -212,11 +215,8 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
      * @returns {boolean} true if change occurred
      */
     sendObjectBackwards(object: FabricObject, intersecting?: boolean) {
-      if (!object) {
-        return false;
-      }
       const idx = this._objects.indexOf(object);
-      if (idx !== 0) {
+      if (idx > 0) {
         // if object is not on the bottom of stack
         const newIdx = this.findNewLowerIndex(object, idx, intersecting);
         removeFromArray(this._objects, object);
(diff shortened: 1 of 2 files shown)


src/Collection.ts
```diff
@@ -274,45 +271,35 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
...
-        newIdx = idx;
         // traverse down the stack looking for the nearest intersecting object
         for (let i = idx - 1; i >= 0; --i) {
           if (object.isOverlapping(this._objects[i])) {
-            newIdx = i;
-            break;
+            return i;
           }
         }
+        return idx;
```
ShaMan123, line 281: I wonder if this is desired or should it fallback to 
```suggestion
        return idx - 1;
```

src/Collection.ts
```diff
@@ -274,45 +271,35 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
...
-        newIdx = idx;
         // traverse up the stack looking for the nearest intersecting object
         for (let i = idx + 1; i < this._objects.length; ++i) {
           if (object.isOverlapping(this._objects[i])) {
-            newIdx = i;
-            break;
+            return i;
           }
         }
+        return idx;
```
ShaMan123, line 299: same but 
```suggestion
        return idx + 1;
```

ShaMan123 (commented):
@asturur this is a fix, important for v6 IMO
Take a look

asturur:
> @asturur this is a fix, important for v6 IMO Take a look

I took a look.
I don't see how it reintroduces stack operation for active selection ( closing https://github.com/fabricjs/fabric.js/issues/9730)
Also how does we get in a condition where an object that is not part of the stack enter the stack?

src/Collection.ts
```diff
@@ -212,11 +215,8 @@ export function createCollectionMixin<TBase extends Constructor>(Base: TBase) {
      * @returns {boolean} true if change occurred
      */
     sendObjectBackwards(object: FabricObject, intersecting?: boolean) {
-      if (!object) {
-        return false;
-      }
       const idx = this._objects.indexOf(object);
-      if (idx !== 0) {
```
ShaMan123, line 219: here for example when idx === -1

asturur:
What i m asking is:
How does a developer get to call a collection stack method, like moveBackward with an object that is not of that collection?
Is it because of nested active selections or because of human error?
If is because of nested active selections, can you make an example?

ShaMan123:
> What i m asking is: How does a developer get to call a collection stack method, like moveBackward with an object that is not of that collection? Is it because of nested active selections or because of human error? If is because of nested active selections, can you make an example?

Human error.
```
stack.moveObjectBackwrads(notAChild);
```
This will insert `notAChild` into `stack`.

asturur:
ok this i wanted to know, i don't think we should guard for an error of this kind.
The day we know how to add console.logs and have a dev - prod build with and without dev messages, we can add those kind of warning to the dev, in this case i really don't want to add them.

It would be like guarding devs that strokeWidth takes a number and not a string number, or valid color values for fill ('ciiaann is not a color')

I m already half half on the other messages we have there and i always wanted them optional

ShaMan123:
> It would be like guarding devs that strokeWidth takes a number and not a string number, or valid color values for fill ('ciiaann is not a color')

I am not sure I agree
But since we changed the method owner, its meaning and name I am fine with not agreeing on this

PR #9775, opened 2024-03-29
test(): Rename svg tests

asturur:
## Description

Working on the clipPath i noticed that  some tests have shared golden.
That is tricky because in one case the image is created with svg export, another case with canvas rendering code, they should match IN THEORY, but they don't if you put a bug in a place and not in another.

This happened while testing the code for the sharper clipPath and this PR is both to see if there are any differences today in canvas vs svg rendering on clipPaths and also to keep them distinct.

The changes we see in the snapshot sizes is because we stopped passing or validating the width/height of the SVG  when we group the elements

The reason to regenerate those files is to make clearer which are the differences when testing a different clipPath code


asturur:
@ShaMan123 there isn't much to see here.
I renamed some pngs and renamed some tests, duplicated some png that come from different tests.
If you have time have a look, otherwise in the weekend i will merge

PR #9793, opened 2024-04-11
perf(): reduce setCoords calls #2

ShaMan123:
## Description

A partial of #9550

After removing lineCoords we have 2 sets of coords
Corner coords (`aCoords`) used for object geometry (`isOnScreen`, intersection etc.) that belong to the parent plane, `getCoords` returning them in the scene plane.
Control coords (`oCoords`) used for control rendering/hit detection that belong to the viewport plane.

This distinction is very important and useful.
It means that viewport changes do not affect corner coords. 
Also since they belong to the parent plane they need to be recalculated only if the object itself changes its own transform/size. This means that if an ancestor changes transform we do not need to recalculate the corner coords.
Since control coords exist only on the active object it seems important to separate use cases of `setCoords` to reduce computation.

I have found that fabric used to call `setCoords` before drawing the controls (at the end of the rendering cycle) instead of after a transform action. This led to a stale coords state between the 2 calls affecting apps using corner controls in the rendering cycle. It was always a frame behind. This is fixed now.

Looking into `setCoords` in general it seems we should decide when they are first set. It is ambiguous.
Does it happen when an object mounts the canvas? What about an object added to a group? We can check if the group has a canvas set, therefore it is mounted , therefore `setCoords` should be called.
It also means that object geometry should not be used until an object is mounted.

With this in mind and with the existing nested setCoords logic needing a refactor/solution I've introduced an `invalidateCoords` concept that effectively replaces the need to call `setCoords`.
I had to solve the nested set coords calls somehow and this seemed the right and simple solution.
I used the existing logic from `calcACoords` that calculates `aCoords` if they are `undefined` so
`invalidateCoords` deletes `aCoords`/`oCoords` and they will be recalculated when requested by `getCoords`/`getControlCoords`.

I think this is great DX and is good for perf because the coords will not be calculated until they are actually needed (consider a case where geometry is not used at all, including `skipOffscreen` - there is no need for aCoords and there are other cases that are solved with this approach) and most importantly nothing is breaking.

### DIFF
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -224,7 +224,7 @@ export class StaticCanvas<
       obj.canvas.remove(obj);
     }
     obj._set('canvas', this);
-    obj.setCoords();
+    obj.invalidateCoords();
     this.fire('object:added', { target: obj });
     obj.fire('added', { target: this });
   }
@@ -793,7 +793,7 @@ export class StaticCanvas<
    */
   _centerObject(object: FabricObject, center: Point) {
     object.setXY(center, CENTER, CENTER);
-    object.setCoords();
+    object.invalidateCoords();
     this.renderOnAddRemove && this.requestRenderAll();
   }
 
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -379,21 +379,7 @@ export class StaticCanvas<
    * @param {Array} vpt a Canvas 2D API transform matrix
    */
   setViewportTransform(vpt: TMat2D) {
-    const backgroundObject = this.backgroundImage,
-      overlayObject = this.overlayImage,
-      len = this._objects.length;
-
-    this.viewportTransform = vpt;
-    for (let i = 0; i < len; i++) {
-      const object = this._objects[i];
-      object.group || object.setCoords();
-    }
-    if (backgroundObject) {
-      backgroundObject.setCoords();
-    }
-    if (overlayObject) {
-      overlayObject.setCoords();
-    }
+    this.viewportTransform = [...vpt];
     this.calcViewportBoundaries();
     this.renderOnAddRemove && this.requestRenderAll();
   }
--- a/src/shapes/ActiveSelection.ts
+++ b/src/shapes/ActiveSelection.ts
@@ -70,13 +70,6 @@ export class ActiveSelection extends Group {
 
   /**
    * @private
-   */
-  _shouldSetNestedCoords() {
-    return true;
-  }
-
-  /**
-   * @private
    * @override we don't want the selection monitor to be active
    */
   __objectSelectionMonitor() {
@@ -161,6 +154,9 @@ export class ActiveSelection extends Group {
     this._exitGroup(object, removeParentTransform);
     // return to parent
     object.parent && object.parent._enterGroup(object, true);
+    // invalidate coords in case active selection was transformed
+    delete object['aCoords'];
+    delete object['oCoords'];
   }
 
   /**
--- a/src/shapes/ActiveSelection.ts
+++ b/src/shapes/ActiveSelection.ts
@@ -154,9 +154,6 @@ export class ActiveSelection extends Group {
     this._exitGroup(object, removeParentTransform);
     // return to parent
     object.parent && object.parent._enterGroup(object, true);
-    // invalidate coords in case active selection was transformed
-    delete object['aCoords'];
-    delete object['oCoords'];
   }
 
   /**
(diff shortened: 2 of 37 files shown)


src/shapes/Group.ts
ShaMan123, line 1: I don't like this, we need a better solution or a different decision.
It is not a non brainer and the entire PR hangs over how we resolve nested set coords

src/shapes/ActiveSelection.ts
```diff
@@ -149,6 +144,7 @@ export class ActiveSelection extends Group {
     // the object will be in the objects array of both the ActiveSelection and the Group
     // but referenced in the group's _activeObjects so that it won't be rendered twice.
     this._enterGroup(object, removeParentTransform);
+    object.setCoords();
```
ShaMan123, line 147: I need to validate this

src/canvas/StaticCanvas.ts
```diff
@@ -379,21 +379,7 @@ export class StaticCanvas<
...
-      const object = this._objects[i];
-      object.group || object.setCoords();
-    }
-    if (backgroundObject) {
-      backgroundObject.setCoords();
-    }
-    if (overlayObject) {
-      overlayObject.setCoords();
-    }
+    this.viewportTransform = [...vpt];
```
ShaMan123, line 382: are you opposed to this @asturur ?

src/shapes/ActiveSelection.ts
ShaMan123: I think if needed it should be done in `setActiveObject`

src/shapes/Group.ts
ShaMan123: refactored to use coords invalidation, open for discussion

src/shapes/IText/ITextBehavior.ts
ShaMan123, line 1: no need for setCoords since initDimensions calls invalidateCoords

ShaMan123 (commented):
I've introduced an `invalidateCoords` concept that effectively replaces the need to call `setCoords`.
I had to solve the nested set coords call somehow and this seemed the right and simple solution.
I used the existing logic from `calcACoords` that calcs aCoords if they are undefined so
`invalidateCoords` deletes `aCoords`/`oCoords` and they will be recalculated when requested.

I think this is great DX and is good for perf, nothing breaking.

jiayihu:
> It also means that object geometry should not be used until an object is mounted.

Definitely happened to me multiple times in the past that I was writing a test about geometry and it didn't work because I forgot to add it to the canvas. But I guess it's inevitable because if there is no canvas, then there is no scene or viewport plane.

Anyway from a quick look, the approach LGTM. You should explain the `invalidateCoords` concept in the PR description, not in a comment though. The PR description is what remains afterwards as documentation.

But other than that, calling `invalidateCoords` when geometry properties are changed, then calling `getControlCoords` when rendering the control (thus only if needed) and recomputing them if invalidated sounds good to me.

src/shapes/Object/InteractiveObject.ts
```diff
@@ -91,7 +91,7 @@ export class InteractiveFabricObject<
    * `corner/touchCorner` describe the 4 points forming the interactive area of the corner.
    * Used to draw and locate controls.
    */
-  declare oCoords: Record<string, TOCoord>;
+  protected declare oCoords?: Record<string, TOCoord>;
```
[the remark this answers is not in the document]
ShaMan123: It should not be accessed outside the prototype chain I believe.
For that you have `getControlCoords` which safeguards the optional case.
I mark stuff as protected/private to inform the dev what fabric expects the accessibility to be and I think that is correct here but I am not attached to it.

ShaMan123 (commented):
This baby is ready

asturur:
this went completely off rail from what we discussed in the other PR.
i not going those changes.
Let me re-read what i suggested but it seems to me i said we should have done the viewport method and the move from render to action and park the other discussions.

no?

ShaMan123:
I worked on that but as I wrote in the description the nested set coords were a problem that needed solving so I solved it in the first commit but setting the coords but that seemed very bad for perf and was bad design so I tried this.
Apart from a lot of diff why do you think it is not in the right dorection?

asturur:
Is not important if this is the right direction or not, we discussed deeply that I don't want to open those changes with PRs and quick descriptions.

We had a meeting in which you wanted a roadmap and a plan, and we are here skipping the plan the roadmap and bring another breaking change to discuss in front of code diffs.

SetCoords are now inadequate ( i say it believing, because not a reproduction case to understand the level of objects needed to make it problematic ) because we built a functionality on top of them without thinking enough.
And this would be repeating the same mistake.
I just do not want to work in this way.

ShaMan123:
Nothing here is breaking
We will not agree unfortunately

ShaMan123:
Maybe you missed my comments, my PR description and the fact that I committed in a revertible way

PR #9794, opened 2024-04-11
feat(Group): make `interactive` > `subTargetCheck` 

ShaMan123:
completes #9550

This PR makes 
```
new Group(object, { interactive: true })
// eq to
new Group(object, { interactive: true, subTargetCheck: true })
```

To be able to select sub targets both need to be flagged as true.
`subTargetCheck` flags `searchPossibleTargets` to traverse the children of a group.
`interactive` flags to return the top most sub target found within `searchPossibleTargets`.
So essentially `interactive = true` contains `subTargetCheck = true`. 
It is more straight forward and better DX.
I would like this to be part of v6.

### DIFF
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -91,7 +91,6 @@ export class Group
   /**
    * Used to allow targeting of object inside groups.
    * set to true if you want to select an object inside a group.\
-   * **REQUIRES** `subTargetCheck` set to true
    * This will be not removed but slowly replaced with a method setInteractive
    * that will take care of enabling subTargetCheck and necessary object events.
    * There is too much attached to group interactivity to just be evaluated by a
--- a/.codesandbox/templates/vanilla/src/testcases/nestedSelections.ts
+++ b/.codesandbox/templates/vanilla/src/testcases/nestedSelections.ts
@@ -22,7 +22,6 @@ export function testCase(canvas: fabric.Canvas, objectCaching = true) {
     ],
     {
       backgroundColor: 'blue',
-      subTargetCheck: true,
       interactive: true,
       objectCaching,
     }
--- a/e2e/tests/controls/hit-regions/index.ts
+++ b/e2e/tests/controls/hit-regions/index.ts
@@ -29,7 +29,6 @@ beforeAll((canvas) => {
     angle: 30,
     scaleX: 2,
     interactive: true,
-    subTargetCheck: true,
   });
   canvas.add(group);
   canvas.centerObject(group);
--- a/e2e/tests/selection/multi-selection-in-groups/index.ts
+++ b/e2e/tests/selection/multi-selection-in-groups/index.ts
@@ -24,7 +24,6 @@ beforeAll(async (canvas) => {
     ],
     {
       backgroundColor: 'blue',
-      subTargetCheck: true,
       interactive: true,
     }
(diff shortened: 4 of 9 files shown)


src/shapes/Group.ts
```diff
@@ -91,7 +91,6 @@ export class Group
   /**
    * Used to allow targeting of object inside groups.
    * set to true if you want to select an object inside a group.\
-   * **REQUIRES** `subTargetCheck` set to true
```
asturur, line 94: you can't just delete this.
You have to write on subTargetCheck that the value of subTargetCheck is overwritten by interactivity.

asturur:
i can't really push back on this change since it should be ok, interactive is new, it needs to be introduced so how is introduced is introduced.

But i want to try to explain again that this is not better DX, this is just a flipped DX.

Master:

Point of view number 1:
i do rarely use interacitve groups, i can easily move on and off subtargetCheck so that i can switch what i need.
The one time i use interactive i need to be sure that also subTargetCheck is ON.

Point of view number 2:
I always use interactive groups, every time i want to use them i have to check that subTargetCheck is ON or they do not work.

After this PR:

Point of view number 1 changes in
i do rarely use interacitve groups, i can easily move on and off subtargetCheck but i have to check that Interactive is OFF, so that i can reliably check subtargetCheck.
The one time i use interactive i do not care about subTargetCheck

Point of view number 2:
I always use interactive groups, every time i want to use them i can switch interactive how i need it.


So to improve point of view number 2, you are making worse point of view number 1.
There is no real improvement you are just changing it in a way is more comfortable to one of the possible use cases.

The status quo of master was that at least it was written that they were dependent.
Now we deleted the documentation part that says that are related. (please add that).

So no is not more straightforward, is not a better DX, is a preference depending on how the developer uses FabricJS.

asturur (changes requested):
Please add back a line of documentation over subTargetCheck that explains the dependency on interactive property.

ShaMan123:
I can add a comment, no problem.
But I disagree with you.
For no. 1 I don't see why it should worry about the interactive flag since it is turned off by default.
no. 2 was my intention, true, the best docs would still have left this dependency as a gotcha IMO.

asturur:
You may disagree but this change is perfectly specular.

before:
i need to enable subtarget check to enable interactive or if you want, subtarget check includes interactive in the disabled state.

after:
i need to disable interactive if i want to disable subtarget check or if you want, interactive includes subtarget check in the enabled state

2 booleans, either one winning on the off state, or the other winning on the on state.

ShaMan123:
> after:
> i need to disable interactive if i want to disable subtarget check or if you want, interactive includes subtarget check in the enabled state

I don't understand your disapproval... `interactive` is false by default, when set to `true` it must use `subTargetCheck`.
If you want to disable `subTargetCheck` then you must disable `interactive`.
There is no case for disabling `subTargetCheck` with `interactive` enabled.
Both are false by default.
Makes sense IMO that a single flag makes the group interactive.
I understand you don't agree this is more straight forward but I don't understand why...
Leave it as is if it bothers you.

asturur:
https://github.com/fabricjs/fabric.js/pull/9794#issuecomment-2050142046 my message and my request for change do not have any disapproaval. Nor in one comment nor in the other.

PR #9795, opened 2024-04-11
Reduce some calls to setCoords()

asturur:
## Description

Original findings from @ShaMan123  from #9550

The method `setViewportTrasnfrom` from StaticCanvas has code that is not useful anymore and the call to setCoords happening in drawControls will happen on every render with an activeObject. Moving the call in 'setActiveObject' and on the action performed will ensure that for the activeObject the controls are in place.

If for the developer will need to draw controls of multiple objects will need to ensure that on everyObject setCoords is called by the developer's code.


src/canvas/Canvas.ts
```diff
@@ -1304,16 +1304,13 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
...
-    let actionPerformed = false;
-    // this object could be created from the function in the control handlers
+    const { action, actionHandler, target } = transform;
 
-    if (actionHandler) {
-      actionPerformed = actionHandler(e, transform, x, y);
-    }
+    const actionPerformed =
+      !!actionHandler && actionHandler(e, transform, pointer.x, pointer.y);
+    actionPerformed && target.setCoords();
```
asturur, line 1311: this replaces 
https://github.com/fabricjs/fabric.js/pull/9795/files#diff-fe6477e182894fdb5db40e4d3bffa1f12e02af91a158b9619752aca781441332L537

jiayihu:
FWIW the PR LGTM, these are the safe changes we did at workplace as well

asturur:
i just didn't have time to try it, seems safe to me too

PR #9799, opened 2024-04-12
Fix jumping issue when scaling and not centered on the control

milonic:
When scaling from any one of tl, tr, br, bl, ml, mt, mr or mb controls and you are not 100% centered on it you will get a jump effect. This commit helps to fix that by adjusting the starting position by the offset amount when dragging a scale control.

This is most noticeable when using larger controls. Works in both ALT and non ALT modes. On Desktop and Mobile.

Rotate and Move are not affected.

### DIFF
--- a/src/controls/scale.ts
+++ b/src/controls/scale.ts
@@ -18,10 +18,16 @@ import {
 import { wrapWithFireEvent } from './wrapWithFireEvent';
 import { wrapWithFixedAnchor } from './wrapWithFixedAnchor';
 
+type controlOffset = {
+  x: number;
+  y: number;	
+}
+
 type ScaleTransform = Transform & {
   gestureScale?: number;
   signX?: number;
   signY?: number;
+  controlOffset?: controlOffset;  
 };
 
 type ScaleBy = TAxis | 'equally' | '' | undefined;
@@ -171,6 +177,16 @@ function scaleObject(
     }
 
     dim = target._getTransformedDimensions();
+
+    // Adjust control offsets. Fixes jumping problem in scaling when hitting a control while not being centered on it.
+    if (!transform.controlOffset) {
+      // need to determine if this is a ALT Action, multiply newPoints by 2 if it is.
+      let multi = isTransformCentered(transform) ? 2 : 1
+      transform.controlOffset = { x: dim.x - Math.abs(newPoint.x * multi), y: dim.y - Math.abs(newPoint.y * multi) }
+    }
+    dim.x -= transform.controlOffset.x
+    dim.y -= transform.controlOffset.y
+
     // missing detection of flip and logic to switch the origin
     if (scaleProportionally && !by) {
       // uniform scaling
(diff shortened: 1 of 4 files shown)


asturur:
Don't we transform with the mouse coordinates themselves?
Are you able to attach a small jsfiddle of a scaling object where this issue can be felt?

milonic:
Is there a CDN copy of version 6 I can use for future refrence? - I'll do the jumping demo in V5 - it's the same on both versions anyway.

Here's a jsfiddle with the jumping: https://jsfiddle.net/milonic/57e8au32/18/

To see it, move your mouse to one of the corners, let's use the Top Right corner. Then adjust your mouse to the bottom left of the tr corner, press down and drag towards the top/right. You will see the triangle jump.

My fix (in v6) is at https://jsfiddle.net/milonic/fta1wgsj/2/ where it works a bit better

asturur:
ok i understand

milonic:
I know it's not a serious issue, nor is it a show stopper and I think it has been present since version 1 but it would be nice to not have to readjust the dimensions of the object every time you want to make a minor adjustment to its scale.

asturur:
yes i just need to take some time to understand if there are implications on interactions we do not expect on strange edge cases

asturur:
i didn't forget about this, it will looked at soon.
This is an improvement but i have some actual bug to look at

asturur:
i m getting to this next.

asturur:
So i have been trying the code and works ok also in groups with transformations.
I m not sure they way it is implemented is the best one, in particular i do not like that we add the controlOffset during the first run of the scale action.

i have a different proposal

asturur:
Well yesterday was very late and i fell asleep trying to make my proposal work.
I was trying to calculate the offset in `_setupCurrentTransform` using the current mouse position and the current corner position.

i get it to work clearly for normal objects by doing:

```ts
      controlToPointerOffsetInVpt = corner
        ? this.getViewportPoint(e).subtract(target.oCoords[corner])
        : new Point(),
```

in the function `_setupCurrentTransform` but i can't figure out how to make it work from there for nested groups.
Your code is not bad, is just that that i don't like that the scale action has this extra duty to correct the mouse offset at the first tick of a scale action, making the first tick somehow special.

milonic:
The first tick is special in that it's only way I could find of knowing where, in relation to the center point of the control, you should offset by.

It only fires once and as far as I can tell ".transform" gets reset on every new action.

What are your main concerns with this approach?

asturur:
is just about code structure.
If the calculation is in setupCurrentTransform it can be leveraged by different actions for example
As i said, your code is not bad, i m just trying to see if we can fit in the setupTransform process

asturur:
Why close? This is a good change, we just need to package it correctly

milonic:
My mistake, I didn't think that resetting my fork would delete the stuff already committed here.

Github is driving me insane at the moment and I just need to figure it all out.

I'm more than happy for any of my code to be used.

Is there anything I need to do? I thought this was already in?

asturur:
it wasn't because of that issue i have on doing it on the scale action.
Let's finish the blinking cursor thing, i also remember you wanted to address the consistency between cursor dealy on movement with keyboard and mouse and that seems correct to do too.
Then we can go back to this

PR #9802, opened 2024-04-17
perf(): rm runtime RegExp usages

ShaMan123:
## Description

During a perf profiling I noticed that regexp creation was showing up due to use of `parsePath` so I looked into the code found deoptimized code and deoptimized use of regexp.
Therefore, I moved all regexp creation from runtime to const values and simplified a bit of `parsePath`


asturur:
I think this change is fine, we moved out regexp from always allocated to runtime allocated by choice, we should remind it when we roll it back, if that hurts performance is fine.

I need to understand what you mean when you say deoptmized code and use of regexp, deoptimization is usually when the v8 needs to rollback to sort of code parsing rather than compiled, is that what you mean?

Please attach before/after flamegraph for reference

src/util/path/index.ts
```diff
@@ -847,41 +850,31 @@ export const parsePath = (pathString: string): TComplexPathData => {
...
         break;
       }
-      // remove the last part of the chained command
-      filteredGroups.shift();
-      // ` ?` is to support commands with optional spaces between flags
-      matchStr = matchStr.replace(
-        new RegExp(`${filteredGroups.join(' ?')} ?$`),
-        ''
-      );
+      match = match.slice(value ? value.length : 0);
```
asturur, line 877: the changes to the path parsing code please either add a very detailed explanation or let's not do them.

src/util/path/index.ts
```diff
@@ -847,41 +850,31 @@ export const parsePath = (pathString: string): TComplexPathData => {
...
     let paramArr: RegExpExecArray | null;
     do {
-      paramArr = new RegExp(rePathCommand, 'i').exec(matchStr);
+      paramArr = rePathCmd.exec(match);
       if (!paramArr) {
         break;
       }
       // ignore undefined match groups
-      const filteredGroups = paramArr.filter((g) => g);
+      const filteredGroups = paramArr.filter(Boolean);
```
asturur, line 862: This equal to call Boolean as `Boolean(g, index, paramArr)` this is usually suggested to avoid, i would aboid

jiayihu: I'd avoid `filter(Boolean)` also because from TS 5.5 it will be able to automatically type guard `.filter(g => !!g)` to defined values, but `.filter(Boolean)` won't work

PR #9811, opened 2024-04-22
fix(Group, Controls): Fix interactive group actions when negative scaling is involved

### LINKED ISSUE #9801: [Bug]: Flip object inside group whose interactive is true got wrong behavior.
### CheckList

### Version

6.0.0-beta20

Microsoft Edge

None

### Link To Reproduction

https://jsfiddle.net/ritali4912/x6nd5e7k/11/

### Steps To Reproduce

There is a group whose subTargetCheck: true, interactive: true.
Then I flip one of its children, but got strange behavior. It seems the control after flip are wrong.
Refer to the below gif.

FlipObjectInsideGroupWithWrong

### Expected Behavior

The child object inside group can scale/flip as normal.

### Actual Behavior

The child object seems got wrong controls after flip.

_No response_

asturur:
## Description

Today when detecting where oCoords are in a group we have to use a series of transformations.
One of those requires  calculating the transformedDimensions/currentDimension of an object arbitrarly transformed.
Sometimes when a flipping enters those transforms the dimensions are reported as negative.

This flips the control position compared to where we expect them and also interact negatively with padding:

image

This PR try to work around this issue by ensuring the scale values are always positive.

closes #9801


src/shapes/Object/InteractiveObject.ts
```diff
@@ -247,8 +247,13 @@ export class InteractiveFabricObject<
...
         ? qrDecompose(this.calcTransformMatrix())
-        : undefined,
-      dim = this._calculateCurrentDimensions(transformOptions),
+        : undefined;
+    // decomposing could bring negative scaling and `_calculateCurrentDimensions` can't take it
+    if (transformOptions) {
+      transformOptions.scaleX = Math.abs(transformOptions.scaleX);
+      transformOptions.scaleY = Math.abs(transformOptions.scaleY);
+    }
+    const dim = this._calculateCurrentDimensions(transformOptions),
```
asturur, line 256: Having ` _getTransformedDimensions` returning always absolute numbers could be the more correct solution but i don't want to shake that now for an edge case fix.
There could be something that is relying on the case of negative dimensions.
having this.scaleX or this.scaleY negative is not supported in general  and this is the only place where we use the options for calcDimensions with unchecked values

asturur:
Check the snapshot at commit https://github.com/fabricjs/fabric.js/pull/9811/commits/14705242454ad62d781319399e09fd0bbdccd5d4 to see it rendered with master code

asturur:
this is going to be merged monday if there are no comments.

asturur:
unit test pass on node 20 but the pod crash afterwards for some reason.
Run fine on node 20 local

PR #9836, opened 2024-04-30
ci(): Add Jest coverage to the report

asturur:
## Description

For some reason this didn't work before, while now it seems to work fine.
This PR adds JEST coverage to our global coverage.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -102,13 +102,6 @@ jobs:
           install-system-deps: false
       - name: Run Jest unit test
         run: npm run test:jest:coverage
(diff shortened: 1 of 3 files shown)


asturur:
Jest coverage seems too much generous to me, probably some stuff is run over without actually testing

asturur:
i can't say for sure, i would need to cover a new line to be sure.
But it brought up the coverage from 80 to where it was before i moved the tests from qunit to jest.

Mixing e2e and unit and visual test is anyway giving us a very generic idea of coverage. Probably a lot of lines are covered without asserts checking for them

PR #9839, opened 2024-04-30
Check that controls actually exist before referencing them

milonic:
In my current app I am removing the controls on certain behaviors.

For example, if I move an object, I remove all controls whilst the move is happening. 

If an object is being rotated, all controls except for the rotate control are remove whilst the rotate is happening.

If an object is being scaled, all controls except the control that is being actioned is removed.

All control are then returned to their normal state of visible after the transormation has occurred.

Unfortunately, this results in an error because the control no longer exists. This commit helps fix that.


ShaMan123:
Did you consider flagging them as `visible: false` instead? that is supported OOTB

asturur:
Before reopening this, check with control visibility.
Also we could consider if for find control the looping should be done on the keys of oCoords or the keys of the controls object.

In any case adding and removing objects from an object without calling setCoords is broken.
You can also consider to call setCoords rigth after you remove the controls from the object.

But consider that removing the objects means that you have to store them somewhere or recreate them, seems an headache, if you want to hide them temporarely go with visible

PR #9842, opened 2024-05-02
refactor(Canvas): `_setupCurrentTransform` => `setupCurrentTransform` fixing control mousedown unexpected behavior

ShaMan123:
## Description

I found a minor bug during work.

If an object is not selected `_setupCurrentTransform` sets a `drag` action.
However, if `findControl` returned a control the down handler would get called.
This PR fixes that.

- dep(): `_setupCurrentTransform` in favor of `setupCurrentTransform`

As part of the PR I refactored `setupCurrentTransform` to accept the control it should setup or the drag action (which was the default switch if no control was found or if the object wasn't already selected before the mousedown) and a new option of undefined.
I did that to make the code simpler, now we just pass down args instead of letting  `setupCurrentTransform` make decisions. Passing down undefined is also an important option that enables an app to control if/how the drag action is setup. I have done a lot to workaround the default drag transform action and this refactor makes it really simple.

## In Action

Run the following on master and see it fail

```typescript

 it.only('control mousedown handler should be called only if current transform uses it', () => {

     const rectOptions = {
        left: 75,
        top: 75,
        width: 50,
        height: 50,
      };
      const eAt100 = new MouseEvent('mousedown', {
        clientX: 100,
        clientY: 100,
      });
      const eOverTLControl = new MouseEvent('mousedown', {
        clientX: 69,
        clientY: 69,
      });
      const eOverMLControl = new MouseEvent('mousedown', {
        clientX: 69,
        clientY: 95,
        shiftKey: true,
      });

        const canvas = new Canvas();
        const rect = new Rect(rectOptions);
        jest.spyOn(rect, 'toJSON').mockReturnValue('rect');
        canvas.add(rect);
        canvas.setActiveObject(rect);
        const hit = rect.findControl(canvas.getViewportPoint(eOverTLControl));

        expect(hit).toBeDefined();
        expect(hit?.key).toBe('tl');
        const controlDownSpy = jest.spyOn(hit!.control, 'getMouseDownHandler');
        const spy = jest.spyOn(canvas, '_setupCurrentTransform');
        jest.spyOn(canvas, 'findTarget').mockReturnValue(rect);

        canvas.getSelectionElement().dispatchEvent(eOverTLControl);
        expect(spy).toHaveBeenCalledTimes(1);
        expect(spy).toHaveBeenCalledWith(eOverTLControl, rect, true);
        expect(controlDownSpy).toHaveBeenCalledTimes(1);

        spy.mockClear();
        controlDownSpy.mockClear();

        canvas.discardActiveObject();

        canvas.getSelectionElement().dispatchEvent(eOverTLControl);
        expect(spy).toHaveBeenCalledTimes(1);
        expect(spy).toHaveBeenCalledWith(eOverTLControl, rect, false);
        expect(controlDownSpy).not.toHaveBeenCalled();
      });

```


ShaMan123 (commented):
Ready to merge

ShaMan123:
@asturur if this can merge into the rc I would appreciate that

asturur:
The issue with this PR is that on top of fixing a bug you go into a refactor of setupCurrentTransform without explaining why that refactor is necessary or if is just a preference.
What is the goal of the PR? fix the bug? or change something else?

asturur:
When we add ci() build() feat() on the title of the PR i think we are looking at this convetion: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines
sometimes you use some i don't recognize or i do not understand what you mean.

In general this is a refactor, not sure what 'dep' means.
Did you mean deprecation?

ShaMan123:
Yes, I meant deprecation, refactor is more suiting I guess

ShaMan123:
updated description

> As part of the PR I refactored setupCurrentTransform to accept the control it should setup or the drag action (which was the default switch if no control was found or if the object wasn't already selected before the mousedown) or undefined.
I did that to make the code simpler, now we just pass down args instead of letting setupCurrentTransform make decisions. Passing down undefined is also an important option that enables an app to control if/how the drag action is setup. I have done a lot to workaround that and this refactor makes it really simple.

PR #9843, opened 2024-05-02
chore(FabricObject): pass `e` to `shouldStartDragging`

ShaMan123:
## Description

the event is needed to make decisions in `shouldStartDragging`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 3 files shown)


asturur (approved):

asturur:
This change reminds me we don't have an embedded way to disable drag n drop on text

ShaMan123:
return false from shouldStartDragging isn't good enough?

asturur:
maybe? i need to write it down in the docs

PR #9844, opened 2024-05-02
feat(Object): pass `ctx` to dnd methods

ShaMan123:
## Description

Control over the context the dnd effects are drawn upon is vital for apps.
Also since we transform it beforehand it must be supplied

### DIFF
--- a/src/shapes/IText/IText.ts
+++ b/src/shapes/IText/IText.ts
@@ -18,6 +18,7 @@ import {
 } from '../Text/constants';
 import { CENTER, LEFT, RIGHT } from '../../constants';
 import type { ObjectToCanvasElementOptions } from '../Object/Object';
+import type { DragEventRenderingEffectData } from '../../EventTypeDefs';
 
 type CursorBoundaries = {
   left: number;
@@ -551,7 +552,7 @@ export class IText<
   /**
    * @override Render drag start text selection
    */
-  renderDragSourceEffect(e: DragEvent, ctx: CanvasRenderingContext2D) {
+  renderDragSourceEffect(ctx: CanvasRenderingContext2D) {
     const dragStartSelection =
       this.draggableTextDelegate.getDragStartSelection()!;
     this._renderSelection(
@@ -561,7 +562,10 @@ export class IText<
     );
   }
 
-  renderDropTargetEffect(e: DragEvent, ctx: CanvasRenderingContext2D) {
+  renderDropTargetEffect(
+    ctx: CanvasRenderingContext2D,
+    { e }: DragEventRenderingEffectData
+  ) {
     const dragSelection = this.getSelectionStartFromPointer(e);
     this.renderCursorAt(ctx, dragSelection);
   }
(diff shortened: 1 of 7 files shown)


src/shapes/IText/IText.ts
```diff
@@ -482,9 +482,9 @@ export class IText<
    * Used for the drag/drop effect.
    * If contextTop is not available, do nothing.
    */
-  renderCursorAt(selectionStart: number) {
+  renderCursorAt(selectionStart: number, ctx = this.canvas!.contextTop) {
```
ShaMan123, line 485: I don't like this signature and prefer ctx to be the first param and mandatory to align with the rest of the methods.
This is a minor breaking change because it is used in the codebase only once

src/shapes/IText/IText.ts
```diff
@@ -550,21 +550,21 @@ export class IText<
...
       dragStartSelection,
       this._getCursorBoundaries(dragStartSelection.selectionStart, true)
     );
   }
 
-  renderDropTargetEffect(e: DragEvent) {
+  renderDropTargetEffect(e: DragEvent, ctx: CanvasRenderingContext2D) {
     const dragSelection = this.getSelectionStartFromPointer(e);
-    this.renderCursorAt(dragSelection);
+    this.renderCursorAt(dragSelection, ctx);
```
ShaMan123, line 567: used only here

ShaMan123 (commented):
ready for review, please make this part of rc

asturur:
The drag/drop effects are relatively new, if you want to swap the order of the arguments do it, but think first if to do an object or more paramters.
I tagged RC already but i don't mind is not a conceptual change this one

asturur (approved):

asturur:
Either merge as is or re-check after param swapping

ShaMan123:
I prefer param switching, do you think renderDrag.DropEffect should have ctx as first param as well?

ShaMan123:
> The drag/drop effects are relatively new, if you want to swap the order of the arguments do it, but think first if to do an object or more paramters. I tagged RC already but i don't mind is not a conceptual change this one

I missed this message. I want your opinion. I will change renderCursorAt and will wait for you call on the rest

ShaMan123 (commented):
I heeded your advice after I understood #9850 
Passing points and source/target make sense

ShaMan123 (commented):
This is ready for another review 
Done on my end

asturur:
Please keep them separate.


My suggestion was saying that are probably multiple way to fix it this one seem the one with more code added.
It could be possible to just change scenePoint in something that is uncached and use the other cached value that doesn't change with the  viewport.
That would be a 2 line change but we should check if it has impact in performance.

Also in this PR i think there is more that for the events? i see a prevTarget that seems new.

ShaMan123:
I couldn't understand your comment
I passed prevDropTarget to _renderDragEffects to keep the code straightforward and supply better context for the API

PR #9848, opened 2024-05-03
fix(): Adjust flashing rate of editing cursor

asturur:
Originally opened by @milonic 

This commit adjusts the initial flashing speed of the cursor when editing text.

At the moment, when you click on text being edited, the cursor will only begin flashing after its delay has occurred. By swaping the "toValues" around, this makes the cursor begin to flash more quickly provided better feedback to the user.

After a long discussion in the PR close by mistake #9823:
This PR individuates a bug, that the cursor animation depends on 2 parameters:
- text.cursorDuration
- text.cursorDelay
- a 100 ms delay between ticks that is today not configurable.

The cursorDuration is always respected but cursorDelay was behaving uncorrectly. Whichever delay you set, the firs animation was always waiting for cursorDuration + cursorDelay + 100ms
This because the first animation cycle was always from opacity 1 to opacity 1 and so indistinguishable from a dealy.

By swapping the first cycle animation from opacity 1 to opacity 0 this gets fixed.

There is another bug that will be taken care in another PR and that is that the cursor delay is applied only on keyboard or mouse cursor changes and not consistently between the two


src/shapes/IText/__snapshots__/ITextBehavior.test.ts.snap
```diff
@@ -2,63 +2,17 @@
...
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
-  "1.000",
```
asturur, line 61: this first animation of 600ms of just full opacity is gone, the cursor start blinking sooner when there is no delay

src/shapes/IText/__snapshots__/ITextBehavior.test.ts.snap
```diff
@@ -200,23 +157,57 @@ exports[`IText cursor animation snapshot initDelayedCursor true - with NO delay
...
+  "0.996",
```
asturur, line 207: Those are the first 160ms. there are 100 of delay that aren't pictured because the animation is not starting., so this is 100 + around 5 ticks of animation

asturur:
i should add another test to show the animation is properly configurable, i ll do later today.

asturur:
@milonic this is complete imho

PR #9849, opened 2024-05-04
fix(Canvas): mouse move before event data

ShaMan123:
## Description

Mouse move before is fired before the event data is cached, meaning before findTarget is called. This makes the event data wrong because it is always empty.
Mouse down/up do not do that so it is clearly a bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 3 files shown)


ShaMan123 (commented):
READY to merge

asturur (approved):

PR #9855, opened 2024-05-06
fix(Polyline): safeguard points arg from options

ShaMan123:
## Description

Passing the following will result in points2 being set as the polyline points

```
new Polyline(points1, { points: points2 })
```

This PR fixes that and adds a test for it

closes #8296

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(Polyline): safeguard points arg from options [#9855](https://github.com/fabricjs/fabric.js/pull/9855)
 - feat(IText): Adjust cursor blinking for better feedback [#9823](https://github.com/fabricjs/fabric.js/pull/9823)
 - feat(FabricObject): pass `e` to `shouldStartDragging` [#9843](https://github.com/fabricjs/fabric.js/pull/9843)
 - fix(Canvas): mouse move before event data [#9849](https://github.com/fabricjs/fabric.js/pull/9849)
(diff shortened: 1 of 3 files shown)


asturur:
Isn't 'fromObject' already doing it safely? how do you get to the bug?

asturur:
this is fun:
```
      super(_objectSpread2(_objectSpread2({}, options), {}, {
        points
      }));
```

vs

```
      super(_objectSpread2({
        points
      }, options));
```

asturur (approved):

asturur:
I would think this was safeguarding us for all classes
```ts
  /**
   *
   * @param {Function} klass
   * @param {object} object
   * @param {object} [options]
   * @param {string} [options.extraParam] property to pass as first argument to the constructor
   * @param {AbortSignal} [options.signal] handle aborting, see https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal
   * @returns {Promise<FabricObject>}
   */
  static _fromObject<S extends FabricObject>(
    { type, ...object }: Record<string, unknown>,
    { extraParam, ...options }: Abortable & { extraParam?: string } = {}
  ): Promise<S> {
    return enlivenObjectEnlivables<any>(cloneDeep(object), options).then(
      (enlivedMap) => {
        const allOptions = { ...options, ...enlivedMap };
        // from the resulting enlived options, extract options.extraParam to arg0
        // to avoid accidental overrides later
        if (extraParam) {
          const { [extraParam]: arg0, ...rest } = allOptions;
          // @ts-expect-error different signature
          return new this(arg0, rest);
        } else {
          return new this(allOptions);
        }
      }
    ) as Promise<S>;
  }
  ```

PR #9860, opened 2024-05-06
fix(): multiplyTransformMatrixArray redundant multiplication

ShaMan123:
## Description

https://github.com/fabricjs/fabric.js/pull/9851#issuecomment-2096039031

> We arrive here with an array that we created to wrap the matrix, we start a reduce and we multiply the translate matrix by the identity matrix that serves as initial value.

`curr` is always defined in `multiplyTransformMatrixArray`. It should be `undefined` so that the condition inside the reduce cb takes affect. Since it wasn't the multiplication was executed many redundant times.
```
export const multiplyTransformMatrixArray = (
  matrices: (TMat2D | undefined | null | false)[],
  is2x2?: boolean
) =>
  matrices.reduceRight(
    (product: TMat2D, curr) =>
      curr ? multiplyTransformMatrices(curr, product, is2x2) : product,
    iMatrix
  );
```

This has been found by @jiayihu and @asturur  in #9851 
It caused a perf hit and now we understand why
I am to blame!

closes #9851 

In addition I have exposed `hasDimensionMatrixProps` to skip `calcDimensionsMatrix` calculation but from analyzing the code I don't see and real benefit. Instead of calling `createScaleMatrix(1, 1)` and doing 3 iterations over the matrix array it will call `hasDimensionMatrixProps` and concat iMatrix.
So maybe it better and maybe not.


ShaMan123 (commented):
ready

ShaMan123 (commented):
I don't know if hasDimensionProps is worth it

Issue #9852, opened 2024-05-06
[Bug]: Creating Text/IText Objects Significantly Slower in v6 vs. v5


Balearica:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.0-rc1

### Node Version (if applicable)

None

### Link To Reproduction

https://github.com/scribeocr/fabric.js-benchmark

### Steps To Reproduce

Creating text objects (including `Text` and `IText`) appears to be noticeably slower upon updating from v5 to v6.  I created a basic example, which adds several hundred `IText` objects to a canvas using both `v5.3.0` and `v6.0.0-rc1` multiple times, and prints the average time for each Fabric.js version to the console.  It can be run by visiting [this page](https://scribeocr.github.io/fabric.js-benchmark/), and the code is in [this repo](https://github.com/scribeocr/fabric.js-benchmark).  On my system (Chrome, Windows), the average runtime is `128.84ms` for `v5` and `209.52ms` for `v6`.

This increase is significant enough to be noticeable in applications that render many text boxes (which is how I noticed).  I am hoping it is possible to restore performance on par with v5, as I think the type support in v6 makes it much easier to work with, so would like to eventually switch over. 

### Expected Behavior

Similar performance between v5 and v6.

### Actual Behavior

See above.

Balearica:
I did some testing, and was able to identify the PR that caused the bulk of the performance regression: 
 https://github.com/fabricjs/fabric.js/pull/8719

[...]

PR #9862, opened 2024-05-07
WIP: Move default values to prototype to resolve v6 performance regression

Balearica:
## Description
This is currently a proof of concept that should not be merged, but opening for visibility.  

See #9852 for full context.  In short, creating a large number of objects in v6 is significantly slower than doing so in v5.  The root cause is the switch from assigning default values using object prototypes to setting defaults using `getDefaults` and `Object.assign`.

The relevant PR is here: https://github.com/fabricjs/fabric.js/pull/8719
The relevant block of code is here: https://github.com/fabricjs/fabric.js/blob/26ed225b47288f775e598f719e218a660dd58dd0/src/shapes/Object/Object.ts#L285-L288

This PR implements the quick fix described in [this comment](https://github.com/fabricjs/fabric.js/issues/9852#issuecomment-2096055572), which essentially rolls back the change shown above.

1. The `IText.getDefaults` function, and all `getDefaults` functions it calls, are edited to return an empty object.
2. The properties that would have been returned by these `getDefaults` functions are assigned to the object prototypes instead.

This is currently a proof of concept--the final implementation obviously shouldn't have a bunch of `getDefaults` functions that do nothing.

## In Action

The improvement can be confirmed by visiting [this web page](https://scribeocr.github.io/fabric.js-benchmark/) containing a benchmark, and opening up the console (source code [here](https://github.com/scribeocr/fabric.js-benchmark)).  The relevant comparison is `v6` (the master branch) and `v6 edited` (this PR).  On my system the following is printed:

```
Mean time v6: 205.18 ms vs. v6 edited: 138.3 ms.
```

### DIFF
--- a/src/shapes/IText/IText.ts
+++ b/src/shapes/IText/IText.ts
@@ -205,7 +205,7 @@ export class IText<
   static ownDefaults = iTextDefaultValues;
 
   static getDefaults(): Record<string, any> {
-    return { ...super.getDefaults(), ...IText.ownDefaults };
+    return {};
   }
 
   static type = 'IText';
@@ -698,6 +698,8 @@ export class IText<
     super.dispose();
   }
 }
+
+Object.assign(IText.prototype, IText.ownDefaults);
 
 classRegistry.setClass(IText);
 // legacy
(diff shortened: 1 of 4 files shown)


asturur:
@Balearica as mentioned in the discussion we cant do this, this is consider 'hacky' by many and also makes also hard to keep track of types and everything. We are happy to search a solution for faster text constructor time but i don't think will be this

asturur:
close in favor of #9891

Issue #5797, opened 2019-07-10
BlendImage Mask option is not working


lovep34:
I am trying to use the BlendImage option with mask option, but I didn't see any clear documentation, can you please let me know how to use it.

Here I have used the BlendColor option and it is working fine.
https://angular-fabric-blendimage.stackblitz.io

http://fabricjs.com/docs/fabric.Image.filters.fabric.Image.filter.BlendImage.html

asturur:
http://fabricjs.com/image-filters

here blend image is used with multiply and mask, you can look that the code on the right of the page.

anilthaku:
Hi @asturur ,

it is awesome library, it make my canvas more robust. you did a great job :) 

I am also facing the same issue, I think you need to create separate demos for these, Also if possible, please update the documentation as well so it will be more clear to end user. 

Currently I am checking the mask demo from the given link, but it is not working for me.
http://fabricjs.com/image-filters

Thanks

lovep34:
Hi @asturur ,

I can't make it work till now. according to @anilthaku it will be helpful if you can create seperate demos for each functionality.

Thanks for the help!

DanKayeEvolveLab:
I'm running into the same issue, and there doesn't seem to be any documentation on the options for the BlendImage filter class in the standards fabric.js docs. Does anyone know where a working example and/or accurate docs can be found?

asturur:
@DanKayeEvolveLab how does your filter and filter options looks like? what is not working?

DanKayeEvolveLab:
@asturur here's my filter options, along with how I apply the filter. I've also attached the two images I'm using - the 'baseimage' I'm trying to mask, and the 'blurredMaskImage' I'm trying to mask it with.

`const blendImageFilter: fabric.IBlendImageFilter = new fabric.Image.filters.BlendImage({
  image: blurredMaskImage,
  mode: 'mask',
  alpha: 0.5,
});`

`baseImage.filters.push(blendImageFilter);`
`baseImage.applyFilters();`

![baseImage](https://github.com/fabricjs/fabric.js/assets/108761462/7db65204-4859-4767-aadb-441ec62581d2)
![blurredMaskImage](https://github.com/fabricjs/fabric.js/assets/108761462/57cb8895-f20b-4539-b628-c0df1145e2e8)

DanKayeEvolveLab:
[...]

PR #9876, opened 2024-05-09
fix(BlendImage): Update Docs

DanKayeEvolveLab:
## Description
Updating docs for the BlendImage filter, after finding inconsistencies in [Issue #5797](https://github.com/fabricjs/fabric.js/issues/5797#issuecomment-2102098544) 

- corrected JSDoc comment for BlendImage filter's 'image' property
- added JSDoc comments for BlendImage filter's 'mode' property
- removed 'alpha' from BlendImage filter example in two places. Since the alpha property is not supported, it shouldn't be in the example.

### DIFF
--- a/src/filters/BlendImage.ts
+++ b/src/filters/BlendImage.ts
@@ -39,8 +39,7 @@ export const blendImageDefaultValues: Partial<TClassProperties<BlendImage>> = {
  *
  * const filter = new BlendImage({
  *  image: fabricImageObject,
- *  mode: 'multiply',
- *  alpha: 0.5
+ *  mode: 'multiply'
  * });
  * object.filters.push(filter);
  * object.applyFilters();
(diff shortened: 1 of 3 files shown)


src/filters/BlendImage.ts
```diff
@@ -39,20 +39,27 @@ export const blendImageDefaultValues: Partial<TClassProperties<BlendImage>> = {
...
+   * Image to make the blend operation with.
    **/
   declare image: FabricImage;
 
+  /**
+   * Blend mode for the filter: either 'multiply' or 'mask'. 'multiply' will
+   * multiply the values of each channel (R, G, B, and A) of the filter image by
+   * their corresponding values in the base image. 'mask' will only look at the
+   * alpha channel of the filter image, and apply those values to the base
+   * image's alpha channel.
```
asturur, line 59: I m not 100% sure this is the best description, but if you change it is probably mirroring the reality at the moment.
I think we should link to mdn pages that explain what multiply or mask would do and then stick our implementation to those.

src/filters/BlendImage.ts
```diff
@@ -39,20 +39,27 @@ export const blendImageDefaultValues: Partial<TClassProperties<BlendImage>> = {
...
  * });
  * object.filters.push(filter);
  * object.applyFilters();
  * canvas.renderAll();
  */
 export class BlendImage extends BaseFilter {
   /**
-   * Color to make the blend operation with. default to a reddish color since black or white
-   * gives always strong result.
+   * Image to make the blend operation with.
```
asturur, line 50: I would just call this FabricImage instead of Image, and i would specify that FabricImage transform properties are taken in consideration

asturur:
Thank you @DanKayeEvolveLab. Could you have a look at my comments and address what is possible?

DanKayeEvolveLab:
@asturur Thank you for finishing up these docs updates! My apologies, I got a bit distracted and forgot to circle back to your feedback.

PR #9884, opened 2024-05-15
fix(Control): corner coords definition order

jiayihu:
`Control.calcCornerCoords` returns corners in an order which make them an incorrect polygon, whereas it should just be a rect. That's because `bl` is defined before `br`.

Since ES2015 object keys are returned in insertion-order by spec (https://stackoverflow.com/a/23202095/6860493) so I believe we should fix this in `calcCornerCoords` instead of manually sorting the coords before calling `Intersection.isPointInPolygon`. That's more convenient and it works correctly out-of-the-box.

If you revert the fix, you'll see the test failing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 
 ## [next]
-
+- fix(Control): corner coords definition order [#9884](https://github.com/fabricjs/fabric.js/pull/9884)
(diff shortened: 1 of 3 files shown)


asturur:
I think this is good, it also somehow aligns it with getCoords that returns an array in the same 'order'

PR #9891, opened 2024-05-25
perf(): Rework constructors to avoid the extra perf cost of current setup

### LINKED ISSUE #9852: [Bug]: Creating Text/IText Objects Significantly Slower in v6 vs. v5

asturur:
## Description

close #9852 

There were (at least) 2 problems with the v6 object initialization.
getDefaults was recursive running up from leaf to root class, spreading the result each time.

Now Textbox calls Itext, then FabricText, then InteractiveObject, then FabricObject getDefaults.

The Object getDefaults() was the largest object of the all, spread once in its own function but then spread again every time it returned from InteractiveObject, FabricText, IText, TextBox so in the worst case there were 5 unnecessary spread operations.

the function getDefaults() has to stay because is necessary for exporting object and removing default values.

What changes now is that every constructor has the duty to assign its own options and defaults, to avoid assigning options more than once we stop passing options to the super() call.

Text / IText / Textbox are its own particular case because they need to assign all the defaults and options before doing the measurements. So in this  case we are still spreading options together with default values, this could make worse restoring an object compared as before, but i don't really have a solution for this. In thise case the defaults value are small objects so hopefully all in all is still overall better performance compared to before.

The best option performance wise would be adding a third parameter to the constructor to carry defaults separated from options, but until that is necessary i prefer to not do it


asturur:
I could offset the added code but i can't really get rid of getDefaults() because that function is still necessary for the 'includeDefaults' functionality during export.

src/shapes/IText/IText.ts
```diff
@@ -217,13 +217,12 @@ export class IText<
...
 
   /**
-
    * Constructor
    * @param {String} text Text string
    * @param {Object} [options] Options object
    */
   constructor(text: string, options?: Props) {
-    super(text, options);
+    super(text, { ...IText.ownDefaults, ...options } as Props);
```
asturur, line 225: Can't avoid rebuilding the object for Textbox -> IText -> Text because of how text measuring works. All the defaults needs to be assigned before Text base class execute the measurement of the text.

asturur:
There are 2 failing unit tests, hopefully are just a small undersight and not a fundamental issue

asturur:
with this branch i get 
for 100.000 Textbox between 2.8 to 3.1 seconds around 25% of current master
for 100.000 Polygons between 0.81 to 0.84 seconds around 17% of current master
for 100.000 Rects around 0.7 seconds around 17% of current master

With master:
for 100.000 Textbox around 12 seconds
for 100.000 Polygons between 4.9 to 5.05 seconds
for 100.000 Rects between 4.1 to 4.2 seconds

This happens in a tight loop with no pauses so not really a real use case but still numbers are definitely better especially for simpler classes.

asturur:
If i try to replicate the same on fabric 5.x in jsfiddle, this is what i get:

for 100.000 Textbox around 0.75 seconds
for 100.000 Polygons between 15 millisecond
for 100.000 Rects between 10 milliseconds

So we are still miles away from the original performance
I can go down to 60ms for the Rect in fabric 6, if i turn back to the values on prototype, still 6 time slower than original, but that may be related to a series of things, code changed, more properties, classes vs functions, all things i won't be able to benchmark in a syntethic way.

@Balearica i would appreciate if you could run this branch in your benchmark to see a real use case

asturur:
I made some dumb test here and there , producing some non real-life test case
I create a bunch of classes like this:

```js
class a {
  constructor() {
    Object.assign(this, fabricObjectDefaultValues);
  }

  method1() {
    return 1;
  }

  method2() {
    return 1;
  }

  method3() {
    return 1;
  }

  method4() {
    return 1;
  }

  method5() {
    return 1;
  }

  method6() {
    return 1;
  }

  method7() {
    return 1;
  }

  method8() {
    return 1;
  }
}

class b extends a {
  constructor() {
    super();
    Object.assign(this, textDefaultValues);
  }

  method7() {
    return 1;
  }

  method8() {
    return 1;
  }

  method9() {
    return 1;
  }
}

export class c extends b {
  method10() {
    return 1;
  }
}

class a1 {
  method1() {
    return 1;
  }

  method2() {
    return 1;
  }

  method3() {
    return 1;
  }

  method4() {
    return 1;
  }

  method5() {
    return 1;
  }

  method6() {
    return 1;
  }

  method7() {
    return 1;
  }

  method8() {
    return 1;
  }
}

Object.assign(a1.prototype, fabricObjectDefaultValues);

class b1 extends a1 {
  constructor() {
    super();
  }

  method7() {
    return 1;
  }

  method8() {
    return 1;
  }

  method9() {
    return 1;
  }
}

Object.assign(b1.prototype, textDefaultValues);

export class c1 extends b1 {
  method10() {
    return 1;
  }
}

class a2 {
  top = 0;
  left = 0;
  width = 0;
  height = 0;
  angle = 0;
  flipX = false;
  flipY = false;
  scaleX = 1;
  scaleY = 1;
  minScaleLimit = 0;
  skewX = 0;
  skewY = 0;
  originX = 'LEFT';
  originY = 'TOP';
  strokeWidth = 1;
  strokeUniform = false;
  padding = 0;
  opacity = 1;
  paintFirst = 'fill';
  fill = 'rgb(0,0,0)';
  fillRule = 'nonzero';
  stroke = null;
  strokeDashArray = null;
  strokeDashOffset = 0;
  strokeLineCap = 'butt';
  strokeLineJoin = 'miter';
  strokeMiterLimit = 4;
  globalCompositeOperation = 'source-over';
  backgroundColor = '';
  shadow = null;
  visible = true;
  includeDefaultValues = true;
  excludeFromExport = false;
  objectCaching = true;
  clipPath = undefined;
  inverted = false;
  absolutePositioned = false;
  centeredRotation = true;
  centeredScaling = false;
  dirty = true;

  method1() {
    return 1;
  }

  method2() {
    return 1;
  }

  method3() {
    return 1;
  }

  method4() {
    return 1;
  }

  method5() {
    return 1;
  }

  method6() {
    return 1;
  }

  method7() {
    return 1;
  }

  method8() {
    return 1;
  }
}

class b2 extends a2 {
  fontSize = 40;
  fontWeight = 'normal';
  fontFamily = 'Times New Roman';
  underline = false;
  overline = false;
  linethrough = false;
  textAlign = 'left';
  fontStyle = 'normal';
  lineHeight = 1.16;
  superscript = {
    size: 0.6, // fontSize factor
    baseline: -0.35, // baseline-shift factor (upwards)
  };
  subscript = {
    size: 0.6, // fontSize factor
    baseline: 0.11, // baseline-shift factor (downwards)
  };
  textBackgroundColor = '';
  stroke = null;
  shadow = null;
  path = undefined;
  pathStartOffset = 0;
  pathSide = 'left';
  pathAlign = 'baseline';
  _fontSizeFraction = 0.222;
  offsets = {
    underline: 0.1,
    linethrough: -0.315,
    overline: -0.88,
  };
  _fontSizeMult = 1.13;
  charSpacing = 0;
  deltaY = 0;
  direction = 'ltr';
  CACHE_FONT_SIZE = 400;
  MIN_TEXT_WIDTH = 2;

  constructor() {
    super();
  }

  method7() {
    return 1;
  }

  method8() {
    return 1;
  }

  method9() {
    return 1;
  }
}

export class c2 extends b2 {
  method10() {
    return 1;
  }
}

function a3() {}

Object.assign(a3.prototype, {
  top: 0,
  left: 0,
  width: 0,
  height: 0,
  angle: 0,
  flipX: false,
  flipY: false,
  scaleX: 1,
  scaleY: 1,
  minScaleLimit: 0,
  skewX: 0,
  skewY: 0,
  originX: 'LEFT',
  originY: 'TOP',
  strokeWidth: 1,
  strokeUniform: false,
  padding: 0,
  opacity: 1,
  paintFirst: 'fill',
  fill: 'rgb(0,0,0)',
  fillRule: 'nonzero',
  stroke: null,
  strokeDashArray: null,
  strokeDashOffset: 0,
  strokeLineCap: 'butt',
  strokeLineJoin: 'miter',
  strokeMiterLimit: 4,
  globalCompositeOperation: 'source-over',
  backgroundColor: '',
  shadow: null,
  visible: true,
  includeDefaultValues: true,
  excludeFromExport: false,
  objectCaching: true,
  clipPath: undefined,
  inverted: false,
  absolutePositioned: false,
  centeredRotation: true,
  centeredScaling: false,
  dirty: true,
  method1() {
    return 1;
  },

  method2() {
    return 1;
  },

  method3() {
    return 1;
  },

  method4() {
    return 1;
  },

  method5() {
    return 1;
  },

  method6() {
    return 1;
  },

  method7() {
    return 1;
  },

  method8() {
    return 1;
  },
});

function b3() {}

b3.prototype = new a3();
b3.prototype.constructor = b3;
Object.assign(b3.prototype, {
  method7() {
    return 1;
  },

  method8() {
    return 1;
  },

  method9() {
    return 1;
  },
  fontSize: 40,
  fontWeight: 'normal',
  fontFamily: 'Times New Roman',
  underline: false,
  overline: false,
  linethrough: false,
  textAlign: 'LEFT',
  fontStyle: 'normal',
  lineHeight: 1.16,
  superscript: {
    size: 0.6, // fontSize factor
    baseline: -0.35, // baseline-shift factor (upwards)
  },
  subscript: {
    size: 0.6, // fontSize factor
    baseline: 0.11, // baseline-shift factor (downwards)
  },
  textBackgroundColor: '',
  stroke: null,
  shadow: null,
  path: undefined,
  pathStartOffset: 0,
  pathSide: 'LEFT',
  pathAlign: 'baseline',
  _fontSizeFraction: 0.222,
  offsets: {
    underline: 0.1,
    linethrough: -0.315,
    overline: -0.88,
  },
  _fontSizeMult: 1.13,
  charSpacing: 0,
  deltaY: 0,
  direction: 'ltr',
  CACHE_FONT_SIZE: 400,
  MIN_TEXT_WIDTH: 2,
});

export function c3() {}

c3.prototype = new b3();
c3.prototype.constructor = b3;

Object.assign(c3.prototype, {
  method10() {
    return 1;
  },
});

```
And then benched them 100000 time
image

So if there is any work going on for the values with object assign or public properties, the construction time is completely different. If you want to have them super fast you have to move the values in the prototype. period.
The longer time we see for fabric6 vs fabric5 when we move the values in the prototype are probably related to the multiple empty Object.assign calls and empty setOptions

jiayihu:
Just to throw an idea, what if `ClassRegistry.setClass()` calls `getDefaults()` and assign it to the prototype for you? So it's done only once, it's on the prototype but without requiring the dev to do it manually.

Otherwise honestly I think it's okay for people to manually do `MyTextbox.prototype.defaults = { ...fabric.Textbox.prototype.defaults, align: "right" }`. You do it once and never get bothered with it.

asturur:
The team chose to keep value per class. I wasn't excited but that is it, i m not rolling back that now.
I wanted to leave a workaround and that is why defaults are stored in a configurable object, I just need to add a detailed page that explains pro and cons to work in a way or another.

asturur:
This PR a a side effect makes the workaround easier

asturur:
Using the benchmark provided by balerica i can get here with the default values

image

asturur:
Moving also away from all the defaults per instance goes back to v5 level, with a difference of 1-2 ms, so definitely ok.

asturur:
Hopefully i didn't break anything shady.

jiayihu:
Another issue I'm facing is that there is no way to override `InteractiveFabricObject.ownDefaults` because it's not exported. Trying to set `FabricObject.ownDefaults` doesn't work because `InteractiveFabricObject` explicitly uses `InteractiveFabricObject.ownDefaults`, not `this.constructor.ownDefaults`

asturur:
Yes i noticed the InteractiveFabricObject issue too when i was working on this, and i thought i should export it.
Then i forgot.
My understanding is that we can't use `this.constructor.ownDefaults` because otherwise you always use the top one.
I think we did wrong in some exporting and naming. let me pull it out a second.

jiayihu:
> This new class has no defaults, so the static 'ownDefaults' i think it points to the InteractiveObject one, and you could in theory mutate it, no?

No, static properties are defined on the exact class I believe. For sure I tried this already and it didn't work, but I could be wrong as always.

asturur:
Anyway InteractiveObject should be part of the exports. That is certain

jiayihu:
> @asturur Is it intended that canvas classes don't do `Object.assign(this, CanvasClass.ownDefaults)` like object classes do? Neither StaticCanvas nor SelectableCanvas do it. Default canvas properties seem unused basically, at least from a quick look.

What about this issue?

asturur:
Sorry i forgot to answer.
If Defaults aren't used, that is wrong, then if we need or not deep configuration like for objects that i do not know.
If is easy to obtain maybe we should mirror objects.

asturur:
Wait, i went to check the code, it seems it behaves like objects before the refactor.
I do think those are used, StaticCanvas does:
```ts
    super();
    Object.assign(
      this,
      (this.constructor as typeof StaticCanvas).getDefaults()
    );
 ```
and getDefaults() is the recursive one.

Why doesn't look like used to you?

jiayihu:
You're right, I didn't see it 👍

PR #9892, opened 2024-05-26
test(): Convert Animation tests to jest

asturur:
Jest.fakeTimers make the animation tests running very quick while the junit setup was based on setTimeout and waits.


asturur:
node qunit test from 32 secs to 6 secs.
Of course jest is slower, but that is fine, eventually will get faster with time

PR #9896, opened 2024-05-30
fix(Canvas): toDataURL and toCanvasElement do not draw controls by default (skipControlsDrawing)

### LINKED ISSUE #9846: v6 : toDataURL should not draw controls
### Discussed in https://github.com/fabricjs/fabric.js/discussions/9837

<div type='discussions-op-text'>

<sup>Originally posted by **Clemweb** April 30, 2024</sup>
Hi,

Is it normal that in v6 toDataURL return image with object controls ? I don't see it with v5.

I don't find any option to disabled this behavior (I just want the canvas with objects without controls on it)

toDataUrl
</div>

asturur:
## Description

close #9846

Add an override for the canvas to hide the active object controls
Left as an option.

This will be completely useless when controls will be moved on the upper canvas.


src/canvas/SelectableCanvas.ts
```diff
@@ -467,6 +468,20 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     }
   }
 
+  toCanvasElement(
+    multiplier = 1,
+    options = {} as TToCanvasElementOptions
+  ): HTMLCanvasElement {
+    const { keepControls } = options;
+    const originalActive = this._activeObject;
```
jiayihu, line 476: This assumes that only the activeObj has active controls, which is true for vanilla fabric but easily not-so in apps. And it feels hacky anyway. The way instead we did was to use the now-removed `interactive` property when exporting the canvas and skip control rendering if true. For instance skipping `canvas.drawControls()`.

asturur: but this.interactive does not exist anymore.
The issue is in the renderCanvas method that because render the controls in 2 moments of the function is not easily divisible in the static vs non static part.

I could also just add interactive back, it would be as hacky has this but less code probably

jiayihu: I don't think it's hacky as ultimately you need a condition to tell if the rendering is happening because of export and differentiate based on that. It can be called `interactive` but also `interactive`/`static` whatever

asturur: or could be called private shouldRenderControls since is being built just for that

asturur: It is a workaround as the other because the real issue is that the StaticCanvas shouldn't have an empty drawControls method to call because renderCanvas isn't correctly split between the 2 classes. If it was i could just call the part that does not render the controls

We do not need a boolean to render controls if not for this issue here. In that sense are both workarounds

jiayihu: We can also say that it should be left to the app deciding what to do. If you're using `SelectableCanvas` and do not want to render controls, then just discard the activeObj before `toCanvasElement()`. 

I think determining when to render controls is highly app-specific.

asturur: Yes all good points, but when i have a sort of regression, i patch the issue, i take notes for improvement, but i don't like to do both at once. In this case the boolean will take care of both, but is a new feature introduced for a fix, and that is smells of problems to me, always.

PR #9909, opened 2024-06-05
fix(changeWidth): support number transform.originX/Y

jiayihu:
Since `changeWidth` is exported and reusable, it should support number `transform.originX/Y` instead of checking against string origins. While fabric's built-in `_setupCurrentTransform` use `canvas._getOriginFromCorner` for the origin values and it assigns string values, `originX/Y` can be a number as well.


asturur:
Seems good to me

asturur:
added changelog

PR #9914, opened 2024-06-06
fixed renderControls order

DaxGama:
Fixed controls render order for ActiveSelection

**Before**
Child controls renders over ActiveSelection
image

**After**
ActiveSelection controls renders over child's
image

This is my first contribution to this project, please let me know if I have missed out anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
-- fix(_renderControls): fixed render order so group controls are rendered over child objects
+- fix(_renderControls): fixed render order so group controls are rendered over child objects [#9914](https://github.com/fabricjs/fabric.js/pull/9914)
 - fix(filters): RemoveColor has missing getFragmentSource method ( typo ) [#9911](https://github.com/fabricjs/fabric.js/pull/9911)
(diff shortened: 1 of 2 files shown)


asturur:
ok so the borders, no the controls.
In that case this could make sense. I need to think carefully with nested selections, but with active selection that is never the case.

asturur:
this should be ok. If it turns out it isn't i ll have to revert.

niketaOptimumnbrew:
> this should be ok. If it turns out it isn't i ll have to revert.

Hi @asturur 
I have the same issue with active selection , can you please tell me that how did you resolve that issue?

asturur:
hi @niketaOptimumnbrew you should open a PR on the 5.x branch that looks like this one, do you think you can?

niketaOptimumnbrew:
> hi @niketaOptimumnbrew you should open a PR on the 5.x branch that looks like this one, do you think you can?

hi @asturur No I don't know how to do.
can you please tell me how to do?

asturur:
i opened one few hours ago with many fixes

asturur:
https://github.com/fabricjs/fabric.js/pulls the first one:
https://github.com/fabricjs/fabric.js/pull/9965

PR #9915, opened 2024-06-07
Add support for decimals in hsla colors and some tests for strings->colors

williamforster:
I have added support for numerals with decimal places when parsing hsla colors.

These can appear in svg files especially, they are otherwise rendered black.

I also added some jest tests to support this.

Thanks !

### DIFF
--- a/src/color/Color.ts
+++ b/src/color/Color.ts
@@ -333,7 +333,6 @@ export class Color {
     const numeric = parseFloat(lowercase);
 
     if (lowercase.includes('rad')) {
-      const numeric = parseFloat(lowercase.replace('rad', ''));
       return radiansToDegrees(numeric);
     }
 
--- a/src/color/Color.ts
+++ b/src/color/Color.ts
@@ -255,14 +255,13 @@ export class Color {
    * @see http://http://www.w3.org/TR/css3-color/#hsl-color
    */
   static sourceFromHsl(color: string): TRGBAColorSource | undefined {
-    const noNone = color.replaceAll('none', '0');
-    const match = noNone.match(reHSLa());
+    const match = color.match(reHSLa());
     if (!match) {
       return;
     }
     const match1degrees = Color.parseAngletoDegrees(match[1]);
 
-    const h = (((parseFloat(match1degrees) % 360) + 360) % 360) / 360,
+    const h = (((match1degrees % 360) + 360) % 360) / 360,
       s = parseFloat(match[2]) / 100,
       l = parseFloat(match[3]) / 100;
     let r: number, g: number, b: number;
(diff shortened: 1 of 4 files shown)


src/color/constants.ts
```diff
@@ -99,7 +99,7 @@ export const reRGBa = () =>
  * So the spec does not allow `hsl(30 , 45%  35, 49%)` but this will work anyways for us.
  */
 export const reHSLa = () =>
-  /^hsla?\(\s*([+-]?\d{1,3})\s*[\s|,]\s*(\d{1,3}%)\s*[\s|,]\s*(\d{1,3}%)\s*(?:\s*[,/]\s*(\d*(?:\.\d+)?%?)\s*)?\)$/i;
+  /^hsla?\(\s*([+-]?[\d\.]+)\s*[\s|,]\s*([\d\.]+%)\s*[\s|,]\s*([\d\.]+%)\s*(?:\s*[,/]\s*(\d*(?:\.\d+)?%?)\s*)?\)$/i;
```
asturur, line 102: if the regex for RGBA for numbers with decimals are good enough, can you use the same notation in the regex?

williamforster: Ok I'll copy the regex from ReRGBa and re-test

src/color/Color.ts
```diff
@@ -317,4 +320,32 @@ export class Color {
...
+    }
+
+    if (lowercase.indexOf('turn') > -1) {
+      const numeric = parseFloat(lowercase.replace('turn', ''));
+      return numeric === undefined ? '0' : (numeric * 360).toString();
+    }
+
+    // Value is probably just a number already in degrees eg '50'
+    return lowercase;
+  }
```
asturur, line 350: we are doing parseFloat with the return of this function.
image

parse float will already exclude strings, so all the replace ('rad') replace('turn') are unnecessary.

Have this function return a number and do a single parseFloat after lowerCase and have it return numeric.

```ts
  static parseAngletoDegrees(value: string): number {
    const lowercase = value.toLowerCase();
    const numeric = parseFloat(lowercase);
    
    if (lowercase.contains('rad')) {
      const numeric = parseFloat(lowercase.replace('rad', ''));
      return radiansToDegrees(numeric);
    }

    if (lowercase.contains('turn')) {
      const numeric = parseFloat(lowercase.replace('turn', ''));
      return numeric * 360;
    }
    return numeric;
  }
```

so the code is less

src/color/Color.ts
```diff
@@ -254,12 +255,14 @@ export class Color {
    * @see http://http://www.w3.org/TR/css3-color/#hsl-color
    */
   static sourceFromHsl(color: string): TRGBAColorSource | undefined {
-    const match = color.match(reHSLa());
+    const noNone = color.replaceAll('none', '0');
```
asturur, line 258: what is this none use case? is it specific of svg? in that case it should be solved in the svg parser ( in a different PR )

williamforster: yes svg allows you to use 'none' instead of '0'. I will move it to the parser

src/color/Color.ts
williamforster: good thinking

src/color/Color.ts
asturur: no leave it for another PR. don't keep adding stuff to the same pr. That is a different topic, 'none' is probably already handled somewhere

PR #9918, opened 2024-06-10
refactor(): remove strict parameter for ancestry.

asturur:
## Description

reasons:
- is unused
- is not an api i want to offer or care about. With the logic you can compare only object on the same canvas, strict=false is not necessary.


src/shapes/Object/StackedObject.ts
```diff
@@ -171,16 +155,16 @@ export class StackedObject<
       return undefined;
     }
     const ancestorData = this.findCommonAncestors(other);
-    if (!ancestorData) {
-      return undefined;
-    }
```
asturur, line 176: findCommonAncestors always return an object.

Issue #8891, opened 2023-05-06
[Bug]: TypeError when adding Blur to Image


jasonkang14:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.0-beta1

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

None

### Link To Reproduction

codesandbox keeps asking me to sign in. the reproducable code is attached below

### Steps To Reproduce

```typescript
// change pages/index.tsx to below

import * as fabric from 'fabric';
import { NextPage } from 'next';
import { useCallback } from 'react';
import { Canvas } from '../components/Canvas';

const IndexPage: NextPage = () => {
  const onLoad = useCallback(async (canvas: fabric.Canvas) => {
    canvas.setDimensions({
      width: window.innerWidth,
      height: 500,
    });
    const image = await fabric.Image.fromURL('https://bookriot.com/wp-content/uploads/2022/08/penguin-random-house-logo-1280x720.png', {crossOrigin: 'anonymous'});
    canvas.add(image);
    const filter = new fabric.filters.Blur({ blur: 0.8 });
    image.filters.push(filter);
    image.applyFilters();
    canvas.renderAll();
  }, []);

  return (
    <div className="position-relative">
      <Canvas onLoad={onLoad} />
    </div>
  );
};

export default IndexPage;
```

you would get a type error like below
<img width="914" alt="Screenshot 2023-05-07 at 5 17 59 AM" src="https://user-images.githubusercontent.com/45306565/236644984-c9c2723b-4505-4e93-ad41-74074b79d991.png">

however when you add `// @ts-expect-error` like below

```typescript
    const image = await fabric.Image.fromURL('https://bookriot.com/wp-content/uploads/2022/08/penguin-random-house-logo-1280x720.png', {crossOrigin: 'anonymous'});
    canvas.add(image);
[...]

PR #9930, opened 2024-06-17
Removing mainParameter and defaults from filters for better TS support

hamzawain7:
- Removed mainParameter from baseFilter, fixing ts error that appeared when using filters (https://github.com/fabricjs/fabric.js/issues/8891).
- I replaced the static default property from baseFilter and used class default values instead.

**Why did I do it?**
As mentioned here: https://github.com/fabricjs/fabric.js/issues/8891, using filters right now gives a TS error. This is because of the hacky mainParameter property in BaseFilter. I was able to remove it without repeating the code unnecessarily.

**What did I do?**
The mainParameter was solely utilized in the isNeutralState function of baseFilters. I modified this function to return false in BaseFilter and override it in child classes to return values relative to their defaults. The 'defaults' class static property in basefilter was also removed as it became redundant after the mainParameter was removed. Additionally, the getMainParameter/setMainParameter functions were eliminated as they were not used anywhere, leading to a more readable and TS friendly code structure.

### DIFF
--- a/src/filters/BaseFilter.ts
+++ b/src/filters/BaseFilter.ts
@@ -37,21 +37,11 @@ export class BaseFilter {
    */
   static type = 'BaseFilter';
 
-  declare static defaults: Record<string, any>;
-
   /**
    * Array of attributes to send with buffers. do not modify
    * @private
    */
   vertexSource = vertexSource;
-
-  /**
-   * Name of the parameter that can be changed in the filter.
-   * Some filters have more than one parameter and there is no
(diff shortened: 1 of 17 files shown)


asturur:
`hacky mainParameter` there is an habit to call 'hacky' whatever doesn't fit our needs.
mainParameter is not the best piece of code written, but the only issue is that is hard to type and that TS is not able to cope with that.

asturur:
replaced by #9933

PR #9936, opened 2024-06-23
feat(Color): add isUnrecognised property

jiayihu:
Allow to be able to tell if a color was not recognised/invalid, otherwise it's not possible to tell between true black and invalid color.


asturur:
yep good point.

asturur:
Please add a changelog and a quick test and let's move on, i was about to pack the latest release

jiayihu:
Done

asturur (approved):

PR #9961, opened 2024-07-02
fix(Canvas): clear _hoveredTarget on discardActiveObject

DaxGama:
Hi, I am not sure sure if this is the right place to fix this issue, but here is the problem.

discardActiveObject leaves _hoveredTarget still pointing to selection object, when selection is mutiple objects.

**Use Case**

I use hoveredTarget to draw focus outline, this was keeping the hovered outline even when the object was discarded as hoveredTarget was not cleared.

**Selection** 

image

**on Discard**

Outline still visible due to _hoveredTarget

image


asturur:
We can't just assume since is not selected anymore is not hovered.
At minimum we need to check if `obj` was an active selection with the utility isActiveSelection.
The best thing to do would be assign a new hovered target again in that case.
But that requires a bit more of inspection of how we do that, because as far as i remember assigning the hovered target is a side effect of a target search.

DaxGama:
Thansk @asturur I missed out on that check, I have added a check using isActiveSelection. I will try to figure out if I can manage to assign new hovered target.

asturur:
i ll get back to this today

asturur:
Add test, changelog and fixed an edge case.

PR #9998, opened 2024-07-14
feat(): Path controls utility

asturur:
## Description

This PR introduces the utilities for path controls.
Those utils use the control api to build a control set that is useful to edit Paths.
The controls reference their points through indexes of command and must be recreated if new commands get added or removed to the path.

## In Action

image


asturur:
Just a reminder that this code was originally thought and proven in ~2017 with @r-a-c-h-e-l.
It finally see the light of day.

asturur:
Ok i think is done now, i need some test.

asturur:
Couldn't really squeeze the code more than this, i managed to make it 1.5kb instead of 2, but the re-usability was bad and also wasn't a good example to developers for other custom functions

vadimcoder:
This is a great feature! Looking forward to having it.

asturur:
Needs tests and the occasion for an explainer on the website on how to do those things with the control api. But the feature itself is ready imho

Issue #10001, opened 2024-07-15
[Bug]: When I have a custom attribute with "type" in it, I get an error `Error: fabric: No class registered for any`


hqw567:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.2

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

None

### Link To Reproduction

https://github.com/hqw567/fabric-issues-react

### Steps To Reproduce

1. Custom attributes contain "type"

### Expected Behavior

Custom attributes contain any attributes without error

### Actual Behavior

When I have a custom attribute with "type" in it, I get an error `Error: fabric: No class registered for any`

hqw567:
```jsx
import * as fabric from 'fabric'
import { useEffect, useRef } from 'react'
import dataJson from './data.json'
const App = () => {
  const canvasRef = useRef()

  useEffect(() => {
    if (!canvasRef.current) return
    const canvas = new fabric.Canvas(canvasRef.current)

    dataJson.objects.forEach((item) => {
      // issues 1: When I have a custom attribute with "type" in it, I get an error `Error: fabric: No class registered for any`
      //
      // Comment the following code to see the error
      delete item.myData.type
    })

    // issues 2: When using it in react, there will still be an error message `TypeError: Cannot read properties of undefined (reading 'clearRect')`
    canvas.loadFromJSON(dataJson).then(() => {
      canvas.requestRenderAll()
    })
    return () => {
      canvas.dispose()
    }
  }, [])

  return (
    <canvas
      width={document.body.clientWidth}
      height={document.body.clientHeight}
      ref={canvasRef}
    />
  )
}

export default App
```

zhe-he:
https://github.com/fabricjs/fabric.js/blob/8219e923770a3105eddb3c96dd945368582d07a1/src/util/misc/objectEnlive.ts#L158-L166
[...]

PR #10017, opened 2024-07-22
fix(): Avoid errors on restoring custom properties that pass the lazy detection of shadow,gradient,pattern and clipPath

zhe-he:
When importing, each property of the object will be iterated over. If a property has type and source, additional logic will be executed. If the user has custom-exported additional properties, and the property is an object with type or source keywords, the user needs to output it as is.


zhe-he:
https://github.com/fabricjs/fabric.js/issues/10001

asturur:
Ok i see your point, but we can't return value on a catch, it will make unparsed pattern returned as is with errors later.
We need make stronger the check. i ll try to make some examples

asturur:
Maybe we can change:
```
const promises = Object.values(serializedObject).map((value: any) => {
```

to

```
const promises = Object.entries(serializedObject).map(([key, value]: any[]) => {
````

and then we can use if `['gradient', 'shadow', 'clipPath'].includes(key) && value.type`

that will make custom properties safe.

And you can do the same for patterns?

**Comment by zhe-he** [comment]
done.

**Comment by asturur** [comment]
Adding a test and merging

File: src/util/misc/objectEnlive.ts [hunk 1]
```diff
@@ -151,31 +151,44 @@ export const enlivenObjectEnlivables = <
...
+        // clipPath or shadow or gradient or text on a path or the backgroundImage or overlayImage of canvas
+        if (
+          [
+            'clipPath',
+            'shadow',
+            'fill',
+            'stroke',
+            'path',
+            'backgroundImage',
+            'overlayImage',
```
asturur, line 168: we forgot 3 cases.

asturur:
I m having a second thought we did it the wrong fix here.

asturur:
@zhe-he look at how i  changed it, and let me know.

src/util/misc/objectEnlive.ts
```diff
@@ -151,44 +150,26 @@ export const enlivenObjectEnlivables = <
...
-          [
-            'clipPath',
-            'shadow',
-            'fill',
-            'stroke',
-            'path',
-            'backgroundImage',
-            'overlayImage',
-          ].includes(key)
-        ) {
```
zhe-he: The last commit was deleted...

zhe-he:
@asturur 
The previous commit modifications are clear to me. I noticed that the last commit marked as 'ok' was reverted, and I don't understand whether it was due to our changes causing other bugs or if it was due to errors from a merge conflict.

asturur:
The fix we made with the properties was making impossible extend a class with a restorable property. because if you added a custom property like 'mask' with a restorable object, then the allow list we built would have blocked you

zhe-he:
Using hasClass is not very reliable either. I might have a custom attribute whose type value happens to exist in Fabric's types, such as Rect, Textbox, Path, etc. Can we first check if the key is a Fabric property, and then check if the value exists in the classRegistry?

asturur:
That would be bad design on implementer side.
In fabric an object with type: rect is a rect. period.
We can't block everyone from adding rects to custom properties because someone could use objects with 'type: rect' that are not rects.

Why shoud you block someone from adding a custom property that is correctly enlivable?

zhe-he:
Okay, if this is reasonable, then it is indeed unnecessary to over-design. I agree with you that allowing users to customize properties of built-in objects is a good idea.

PR #10043, opened 2024-08-06
feat(): Convert align guidelines to TS. Add build system for extensions

zhe-he:
To avoid disrupting the original file structure, I placed the main code in the lib folder and only added exports in the entry file. At the same time, it supports the configuration related to reference lines.
how to use:
```typescript
import { initAligningGuidelines } from "fabric";

initAligningGuidelines(canvas);
```


fabric.ts
```diff
@@ -178,3 +178,5 @@ export { Control } from './src/controls/Control';
 export * as controlsUtils from './src/controls';
 
 export * from './src/filters';
+
+export { initAligningGuidelines } from './lib/aligning_guidelines/index';
```
asturur, line 182: this is the issue i want to solve. I don't want to package those in this way.
I ll need to work out a way for rollup to produce a different bundle that has fabric has an external dep and that is reacheable from fabric/plugin or something like that

asturur:
I will rename lib into plugins as first step

asturur:
worked a bit on the build system this morning, still requires some tweaks.

.codesandbox/templates/vanilla/src/testcases/guidelines.ts
```diff
@@ -0,0 +1,11 @@
+import * as fabric from 'fabric';
+import { initAligningGuidelines } from 'fabric/extensions';
```
asturur, line 2: @zhe-he this is how i want to import the extensions like the guidelines.
I think we can refine this so that splitting and node/browser are still possible cdn links

extensions/aligning_guidelines/util/make.ts
```diff
@@ -0,0 +1,32 @@
...
+  return {
+    y,
+    x1: b
+      ? objectX - objectWidth / 2 - aligningLineOffset
+      : objectX + objectWidth / 2 + aligningLineOffset,
+    x2: b
+      ? activeObjectX + activeObjectWidth / 2 + aligningLineOffset
+      : activeObjectX - activeObjectWidth / 2 - aligningLineOffset,
+  };
+}
```
asturur, line 32: @zhe-he this file seems completely unused. Why is it here? can i delete it?

asturur:
I will merge this code as is because i don't really have time to review the implementation.
I will do that later and ping you for questions and to look at eventual changes.
I still need to verify that i can write a unit test and that the unit test run correctly and that i can make a playwright test for those.
Also different build types.

asturur:
I set up the base for testing and building

.codesandbox/templates/vanilla/src/testcases/guidelines.ts
zhe-he: Yes, this file is not used anywhere, and you can delete it. In the original alignment version, there is a parameter called aligningLineOffset, and I found that its use case is completely ineffective in the new version. Later, I forgot to delete this file, and the offset property in AligningLineConfig should also be removed.

extensions/aligning_guidelines/util/make.ts
zhe-he: Yes, this file is not used anywhere, and you can delete it. In the original alignment version, there is a parameter called aligningLineOffset, and I found that its use case is completely ineffective in the new version. Later, I forgot to delete this file, and the offset property in AligningLineConfig should also be removed.

zhe-he: https://github.com/fabricjs/fabric.js/pull/10056

PR #10050, opened 2024-08-10
fix(): svg parsing logic with use and style

gloriousjob:
## Description

style tags between a path and use tag don't seem to work properly compared to Chrome. In Chrome, when both have a style, the path has priority but merges with the use tag where they do not share styles. Previously, fabric would take the use tag's style wholesale. This PR merges the two when both exist and tests were added to verify the both case and cases where only one has a style.

Closes https://github.com/fabricjs/fabric.js/issues/9840

### DIFF
--- a/src/parser/parseUseDirectives.ts
+++ b/src/parser/parseUseDirectives.ts
@@ -1,6 +1,7 @@
 import { svgNS } from './constants';
 import { getMultipleNodes } from './getMultipleNodes';
 import { applyViewboxTransform } from './applyViewboxTransform';
+import { parseStyleString } from './parseStyleString';
 
 export function parseUseDirectives(doc: Document) {
   const nodelist = getMultipleNodes(doc, ['use', 'svg:use']);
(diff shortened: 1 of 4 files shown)


asturur:
Did you find this priority described somewhere or you just checked how a browser behave?

src/parser/parseUseDirectives.ts
```diff
@@ -67,6 +68,16 @@ export function parseUseDirectives(doc: Document) {
...
         currentTrans = nodeValue + ' ' + currentTrans;
+      } else if (nodeName === 'style' && el2.getAttribute('style') !== null) {
+        // when both sides have styles, merge the two styles, with the ref being priority (not use)
+        const styleRecord: Record<string, any> = {};
+        parseStyleString(nodeValue!, styleRecord);
+        parseStyleString(el2.getAttribute('style')!, styleRecord);
+        const mergedStyles = Object.keys(styleRecord).reduce(
+          (a, v) => a + v + ':' + styleRecord[v] + ';',
+          ''
+        );
```
asturur, line 79: Could this be:
```suggestion
        const mergedStyles = Object.entries(styleRecord).map(entry => entry.join(':')).join(';');
```
?

gloriousjob:
> Did you find this priority described somewhere or you just checked how a browser behave?

I tested the browser and tried different settings. The three tests I came up with show how they acted in chrome. The SVG docs were not clear about how this case should be handled from what I saw.

asturur:
Firefox, Safari and Chrome behave the same.
Style gets merged and priority is given to the original element.

Same for attributes. 
I'll be sure to revisit our parsing logic to be sure we do same

asturur (approved):

PR #10053, opened 2024-08-11
fix(): Parse use directive attribute issues

asturur:
## Description

When parsing a use element, the priority for attribute is from the original.
You can't override a fill or a stroke that are defined in the object you are referencing.
When style is in place, you have to merge by features not by attribute.

So the priority is as follow:
The parent element style fill takes precedence over the parent element fill attribute and they define the parent fill color.
If a parent fill color is defined, you can't change in any way the used element color.
We do that by encoding that color  value in the use-cloned style.

Also refactored some code to be more terse


src/parser/parseStyleString.ts
```diff
@@ -12,6 +12,7 @@ export function parseStyleString(
     .replace(/;\s*$/, '')
     .split(';')
     .forEach((chunk) => {
+      if (!chunk) return;
```
asturur, line 15: @gloriousjob solves an issue with empty style attribute

test/visual/svg_import.js
```diff
@@ -36,7 +36,7 @@
       test: 'Svg import test ' + svgName,
       code: test,
       golden: svgName + '.png',
-      percentage: 0.06,
+      percentage: 0.04,
```
asturur, line 39: lowered the percentage of possible failures

asturur:
Added a failing test for the use with classNames that i don't want to go into fixing now

PR #10056, opened 2024-08-12
update(align-guidelines): Remove the offset of the line

zhe-he:
The new version of the endpoint uses 'X' instead of extending the line like the old version, so the old 'offset' parameter has been removed.


CODE-AXION:
Hi @zhe-he , Hope you're doing well! 
I couldn’t find your email, so I’m reaching out here instead

First off, I just want to say thanks for your work on the aligning guidelines plugin for Fabric.js—it's an awesome addition, and I really appreciate the effort you’ve put into it.

I noticed that the plugin currently supports only Fabric.js v6, but most the people in the community 90% (myself included) is still using v5. Since v6 is still new , a lot of us would love to have a version that works with v5.

Would you be open to maybe releasing a v5 version? I think it would be super helpful for many developers.

Looking forward to hearing your thoughts.

zhe-he:
@CODE-AXION 
Version 5 already has a version with alignment lines, and for the time being, we are not considering adding more features to it. Only fixes will be made to the older version.
I checked the npm download statistics, and in the past two months, Version 6 has been downloaded 49k times, while Version 5 has been downloaded 28k times.
My email is [zhe-he@qq.com](https://mailto:zhe-he@qq.com/), and you can see it on my homepage.

PR #10079, opened 2024-08-19
fix(parser): Skip `id` attribute from normalization

miroljub1995:
## Description

If we load SVG that has some object with `id` attribute starting with a `number`, we get a wrong id parsed.


zhe-he:
This kind of attribute has issues as well, right?
`<svg data-custom-attribute="123abc"></svg>`

src/parser/normalizeValue.ts
```diff
@@ -56,7 +56,12 @@ export function normalizeValue(
     } else if (fillIndex === -1 && strokeIndex > -1) {
       ouputValue = STROKE;
     }
-  } else if (attr === 'href' || attr === 'xlink:href' || attr === 'font') {
+  } else if (
+    attr === 'href' ||
+    attr === 'xlink:href' ||
+    attr === 'font' ||
+    attr === 'id'
+  ) {
```
asturur, line 64: this function is a bit weird.
We do carry a parse and and outputValue around and in some cases we just do an early return.
Hopefully there is a better way to fix it. All the attributes that start with a number that are not specially taken care of here are going to be wrong correct?
So any custom attribute will need to be listed here and that is not really feasible

miroljub1995: is it better to explicitly specify which one needs to be normalized?

asturur: somehow this function become really convoluted yesterday i tried to look at it to expedite this fix. Eventually we can just fix as you did and open a ticket for a refactor, i didn't decide it yet sorry.

asturur:
@miroljub1995 add this line to the changelog:
fix(SVGImport): Allow parsing of 'id' attribute that starts with a number [#10079](https://github.com/fabricjs/fabric.js/pull/10079)

and we can merge.

Please check that it gets formatted as the other lines or we get a lint error.
I can't edit myself as i don't have write access to your branch

miroljub1995:
@asturur 
- I added line to CHANGELOG
- Regarding to lint, I checked with `npm run prettier:check && npm run lint`, but there are already some errors so I just checked if no new errors are on files I changed.

asturur:
There are maybe warnings, the checkmark is green.
Or maybe somehow you are picking up a different configuration

asturur (approved):

PR #10081, opened 2024-08-20
fix(filter): pixelate behaves differently under webgl, and should be squares rather than image scale

xushengfeng:
## Description

In `Pixelate.applyTo2d`, Pixels are squared, which is consistent with the expected performance
But in webgl, pixels follow the width and height of the image.

### DIFF
--- a/src/filters/shaders/pixelate.ts
+++ b/src/filters/shaders/pixelate.ts
@@ -7,7 +7,7 @@ export const fragmentSource = `
   varying vec2 vTexCoord;
   void main() {
     float blockW = uBlocksize * uStepW;
-    float blockH = uBlocksize * uStepW;
+    float blockH = uBlocksize * uStepH;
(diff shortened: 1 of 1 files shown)


asturur:
i remember noticing this and for some reason i thought i fix it, eventually i didn't.

asturur:
Can you post a screenshot of a non square image after pixelate filter is applied? thank you

xushengfeng:
https://jsfiddle.net/5vojyagL/5/
eSearch-2024-08-21-23-25-46-867

a 200*100 gradient as an example photo,

asturur:
This is with your fix or without?

xushengfeng:
the fiddle example is 6.3.0 which isnot fixed

I applied the changes I made in the pull directly in index.min.mjs and it worked fine. I believe this pull change also applies to the source code (because the characters of this change are not affected by compilation)

PR #10120, opened 2024-09-02
fix(AligningGuidelines): Align guidless changes aspect ratio on snapping when scaling

zhe-he:
Resolve the issue where when the user scales using scaleX, scaleY is not updated simultaneously. The same applies to scaleY.

### DIFF
--- a/extensions/aligning_guidelines/index.ts
+++ b/extensions/aligning_guidelines/index.ts
@@ -105,6 +105,10 @@ export function initAligningGuidelines(
       index -= 4;
     }
     let point = activeObject.getCoords()[index];
+    const uniformIsToggled = e.e[canvas.uniScaleKey!];
+    const isUniform =
+      (canvas.uniformScaling && !uniformIsToggled) ||
+      (!canvas.uniformScaling && uniformIsToggled);
     for (const object of objects) {
       const [rect, coords] = getCaCheMapValue(object);
       const center = new Point(
@@ -112,7 +116,7 @@ export function initAligningGuidelines(
         rect.top + rect.height / 2,
       );
       const list = [...coords, center];
-      const props = { activeObject, point, list, isScale, index };
+      const props = { activeObject, point, list, isScale, isUniform, index };
       const vLines = collectVerticalPoint(props);
       const hLines = collectHorizontalPoint(props);
       vLines.forEach((o) => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [next]
 
+- fix(AligningGuidelines): Align guidless changes aspect ratio on snapping when scaling [#10114](https://github.com/fabricjs/fabric.js/issues/10114)
+
 ## [6.4.1]
 
 - fix(): Package.json had wrong path to types for extensions [#10115](https://github.com/fabricjs/fabric.js/pull/10115)
--- a/extensions/aligning_guidelines/util/collect-point.ts
+++ b/extensions/aligning_guidelines/util/collect-point.ts
@@ -4,9 +4,15 @@ import { getDistance } from './basic';
 
 type CollectPointProps = {
   activeObject: FabricObject;
+  /** Operation points of the target element: top-left, bottom-left, top-right, bottom-right */
   point: Point;
+  /** Set of points to consider for alignment: [tl, tr, br, bl, center] */
   list: Point[];
+  /** Change the zoom or change the size, determine by whether e.transform.action starts with the string "scale" */
   isScale: boolean;
+  /** Whether to change uniformly is determined by canvas.uniformScaling and canvas.uniScaleKey. */
+  isUniform: boolean;
+  /** Which specific point to operate on, 0-3 correspond to top-left, top-right, bottom-right, bottom-left */
   index: number;
 };
 const originXArr: TOriginX[] = ['left', 'center', 'right'];
@@ -14,23 +20,35 @@ const originYArr: TOriginY[] = ['top', 'center', 'bottom'];
 
 export function collectVerticalPoint(props: CollectPointProps) {
   const aligningLineMargin = aligningLineConfig.margin;
-  const { activeObject, isScale, index, point, list } = props;
+  const { activeObject, isScale, isUniform, index, point, list } = props;
   const { dis, arr } = getDistanceList(point, list, 'x');
   const margin = aligningLineMargin / (activeObject.canvas?.getZoom() ?? 1);
   if (dis > margin) return [];
   let v = arr[arr.length - 1].x - point.x;
-  const dir = index == 0 || index == 3 ? -1 : 1;
-  v *= dir;
+  // To the left or to the right?
+  const dirX = index == 0 || index == 3 ? -1 : 1;
+  // To the top or to the bottom?
+  const dirY = index < 2 ? -1 : 1;
+  v *= dirX;
 
-  const { width, scaleX, left } = activeObject;
+  const { width, height, scaleX, scaleY, left, top } = activeObject;
   const dim = activeObject._getTransformedDimensions();
   const sx = (v + dim.x) / dim.x;
-  if (isScale) activeObject.set('scaleX', scaleX * sx);
-  else activeObject.set('width', width * sx);
-  const dArr = [0, (v / 2) * dir, v * dir];
-  if (dir < 0) dArr.reverse();
-  const d = dArr[originXArr.indexOf(activeObject.originX)];
-  activeObject.set('left', left + d);
+  if (isScale) {
+    activeObject.set('scaleX', scaleX * sx);
+    if (isUniform) activeObject.set('scaleY', scaleY * sx);
+  } else {
+    activeObject.set('width', width * sx);
+    if (isUniform) activeObject.set('height', height * sx);
+  }
+  const dx = getDisByOriginX(activeObject, v * dirX);
+  if (isUniform) {
+    const h = activeObject._getTransformedDimensions().y - dim.y;
+    const dy = getDisByOriginY(activeObject, h * dirY);
+    activeObject.set('top', top + dy);
+  }
+  activeObject.set('left', left + dx);
+
   activeObject.setCoords();
   return arr.map((item) => ({
     x: item.x,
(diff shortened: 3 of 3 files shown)


zhe-he:
When originX is not center, there is a bug. Please do not merge. Will fix it after lunch

zhe-he:
When originX is set to left and right, it behaves strangely. I'm trying to figure out how to fix it.

zhe-he:
https://github.com/user-attachments/assets/754c7874-96f4-4e23-99ab-c6e6102a2c0a

[test](https://codesandbox.io/p/devbox/fabric-aligning-guidelines-4pr6x7)

zhe-he:
When the angle of the operated graphic is not 0, there is a bug with the 4-corner scaling alignment.

zhe-he:
fixed

https://github.com/user-attachments/assets/1d3265d8-7cae-4042-8d1d-716e8b6ade1e


[demo](https://codesandbox.io/p/devbox/fabric-aligning-guidelines-4pr6x7)

> When the angle of the operated graphic is not 0, there is a bug with the 4-corner scaling alignment.
> I will solve it at noon tomorrow.

zhe-he:
In practical use, a common issue has been identified: when there are too many shapes, the problem of excessive reference line hits occurs because each shape is compared with the current shape. It is suggested to change the approach to count all points and use those points for judgment. (The same issue occurs in V5; if there are related requirements for V5, please refer to the modifications.)

The following is a video comparison before and after the modifications.


https://github.com/user-attachments/assets/79601812-942d-4d0a-9b3f-ed600ada16dc


https://github.com/user-attachments/assets/5f4cf876-9457-47b9-a31c-ebf992133b4a

asturur:
I m sorry i didn't have time for this and i won't for a bit.
All the implementation needs to be checked and verified a bit imho.
When there are too many lines there must be a way for the developer to choose the one they care about or expose an interaction key to manually disable them. Is a common problem

zhe-he:
Okay, I will take some time to write the test cases, while also adding the corresponding listener functions and disabling the relevant feature switches.

zhe-he:
@asturur 
I made a lot of updates, and the final result is significantly different from the initial version. You can ignore the logic of the initial version and directly look at all the final code. Compared to the initial version, I added comments to each line, making it easier to understand.
Additionally, I added user customization options, which are documented in the README. Users can manually disable horizontal lines and customize the alignment of graphics. It also supports user-defined names for controllers.
If you have any questions about any line of code, feel free to tag me, and I will explain once I see it.
This is the final version. If you have time, please download the code locally to review all files; it will be faster than reviewing changes on GitHub. Thank you.
Regarding the test cases for simulating user manual dragging, I haven't written any before. If needed, I will try to add them.
Here is a test demo where you can test by modifying parameters or customizing them.
[demo](https://codesandbox.io/p/devbox/fabric-aligning-guidelines-4pr6x7?file=%2Fsrc%2Ftest.tsx%3A6%2C28)

extensions/aligning_guidelines/index.ts
```diff
@@ -32,113 +28,153 @@ export type { AligningLineConfig } from './typedefs';
...
-      corner,
-    );
-    if (index == -1) return;
-    onlyDrawPoint = index > 3;
+    // When the shape is flipped, the tl obtained through getCoords is actually tr,
+    // and tl is actually tr. We need to make correction adjustments.
+    // tr <-> tl、 bl <-> br、  mb <-> mt、 ml <-> mr
+    if (target.flipX) corner = corner.replace('l', 'r').replace('r', 'l');
+    if (target.flipY) corner = corner.replace('t', 'b').replace('b', 't');
+    const coords = target.getCoords();
```
Smrtnyk, line 104: I can't see properly but is this used?

extensions/aligning_guidelines/README.MD
```diff
@@ -20,3 +24,72 @@ const deactivate = initAligningGuidelines(myCanvas, options);
...
+
+### custom function
+
+```ts
+import { initAligningGuidelines } from 'fabric/extensions';
+import { FabricObject } from 'fabric';
+
+// You can customize the return graphic, and the example will only compare it with sibling elements
+initAligningGuidelines(myCanvas, {
+  getObjectsByTarget: function (target) {
```
Smrtnyk, line 36: I tried this PR and tried using this function to filter out objects that have properties `isGridLine` on them
like `if (!object.isGridLine) { set.add(object) }`
but then no guidelines happened at all
was I using it wrong maybe?

extensions/aligning_guidelines/index.ts
zhe-he: When executing the flip, its value becomes true, which is set internally by Fabric. The video below shows the bug that occurs when it is turned off.


https://github.com/user-attachments/assets/71802194-b2b5-4f34-81e0-737bf8ae85a9

extensions/aligning_guidelines/README.MD
zhe-he: @Smrtnyk 
Can you provide a simple demo to reproduce your issue? This will help me locate the problem. Regarding this function, if customized, it means that when moving the target element, it only processes the reference line for the graphical representation of the return value

extensions/aligning_guidelines/index.ts
Smrtnyk: I mean that `const coords = target.getCoords();` seems unused

extensions/aligning_guidelines/README.MD
Smrtnyk: this is what I tried
   ```typescript
    initAligningGuidelines(this.canvas, {
            getObjectsByTarget: () => {
                const set = new Set<FabricObject>();
                for (const obj of this.canvas.getObjects()) {
                    if (obj.isGridLine) {
                        continue;
                    }
                    set.add(obj);
                }
                return set;
            },
        });
   ```
   I have a grid that I toggle, used as a helper when adding shapes 
   for aligning guidelines I want to ignore all lines in that grid
   and I thought I can ignore them if I don't add them to the set that is being returned

zhe-he: @Smrtnyk 
I wrote the custom code according to your method and found that it works. You can compare it with the demo to see where the differences are.
[custom demo](https://codesandbox.io/p/devbox/fabric-aligning-guidelines-4pr6x7?file=%2Fsrc%2Ftest-custom.tsx%3A17%2C39)


https://github.com/user-attachments/assets/afac664a-d99a-4f03-ae7e-e77990f072c5

extensions/aligning_guidelines/index.ts
zhe-he: Yes, you are right, it should be deleted.

Smrtnyk:
I have been testing this patch for a while now in production in my fabric app.
It works much better than the current one in master tree.
Not sure if it means something, but would be nice to see some progress on this getting merged so I can get rid of my patch

asturur:
I will read all that is going on here, but the feature requires a bunch of refactor imho. That is why i didn't merge it yet

zhe-he:
The following unmentioned bug fix is documented:
1. In the mode where only points are drawn (onlyDrawPoint == true), center point alignment.

https://github.com/user-attachments/assets/3f02fce7-708e-407a-b3bb-eb923f97b2ae

https://github.com/user-attachments/assets/fc60f086-081e-4442-bd95-166978f554dc

2. The issue of inaccurate shape positioning after holding down the alt key.

stevenhurth:
@zhe-he I pulled down your latest commit for the extension. It's soooo much better, nice work! The skewing of shapes when resizing/scaling was a huge concern for us.

I did find a couple other issues, mainly with dragging groups not showing the alignment lines, and I also added a few more configuration options. I was able to address the group dragging issue, by filtering out the group being moved, and its children, when collecting the objects from the canvas.

### Additional configuration
1. An option for rendering the alignment lines with dashes.
2. A bunch of options for changing the line caps: size, color, width and style (x, arrow, perpendicular line, or none at all).

Please let me know if you'd like any of this.

zhe-he:
@stevenhurth 
Both features you mentioned can be added. I'm still considering whether to add them after this version is merged or to continue development based on it.
The current version has already diverged significantly from the original one. If we add more features, it will essentially become a new version, but this version has not yet been released. Adding new features will inevitably introduce additional bugs.
If you urgently need these features, you can make the following modifications:
Modify the drawLine function in the file util/draw.ts to implement reference lines as either solid or dashed.
Modify the drawX function in the same file to implement the functionality for endpoint styles.

zhe-he:
> I did find a couple other issues, mainly with dragging groups not showing the alignment lines, and I also added a few more configuration options. I was able to address the group dragging issue, by filtering out the group being moved, and its children, when collecting the objects from the canvas.

By default, alignment lines are displayed only for the child elements rather than the group as a whole when dragging an element. This behavior is intentional and not a bug. If you need alignment lines for the group, you can only achieve it through customization.

asturur:
It doesn't really matter if they diverge from the initial implementation, what does it matter and how the code works in general and i m sorry i didn't really have time to sit down and read fully through it

zhe-he:
> It doesn't really matter if they diverge from the initial implementation, what does it matter and how the code works in general and i m sorry i didn't really have time to sit down and read fully through it

Okay, I will add these two features later.

zhe-he:
@stevenhurth 
Dashed lines and custom endpoints have been added. Refer to the demo (scroll to the bottom, second canvas).
The README has been updated.

image

Smrtnyk:
@zhe-he maybe worth rebasing?
Maybe it will also remove the stale label

asturur:
I m sorry for lagging on this pr @zhe-he  i need more free time than usual for myself.
I wish aligning guidelines were a bit different in general with the code, but i didn't get there yet

Smrtnyk:
could we rebase this again? this is is a very worthy feature and I would like to see it moving forward
is there anything I could help with here?

asturur:
I have a pending review of the whole system i nvever did.
I do not want @zhe-he to use time rebasing everyting before checking guidelines as a whole.
Maybe i can do exactly this as my next task

zhe-he:
Resolved the final known issue: offset when resizing child elements with Alt.

before:

https://github.com/user-attachments/assets/f4841d2c-4fca-4c7e-b7dd-0d71a08b5c7a

after:

https://github.com/user-attachments/assets/faa21b0b-8ec6-4698-a4e0-1bd076c5f360

extensions/aligning_guidelines/index.ts
```diff
@@ -1,166 +1,264 @@
...
+      const angle = target.getTotalAngle();
+      // When the shape is rotated, it is meaningless to draw points using the center point.
       if (angle % 90 != 0) return;
-      index -= 4;
     }
-    let point = activeObject.getCoords()[index];
+    // If manipulating tl, then when the shape changes size, it should be positioned by br,
+    // and the same applies to others.
+    // users can be allowed to customize and pass in custom corners.
+    const contraryMap = this.getContraryMap(target);
```
alxpsr, line 159: Hi. Applied your code and it works fine! Now trying to to understand this approach with getContraryMap and setting setRelativeXY. Could you explain this approach? Why default fabricjs didn't work without it? Thanks in advance

zhe-he: When resizing a shape, each control point A requires a corresponding point B as an anchor. The contraryMap stores these relationships in the form { A: B }.
The reference lines that appear during scaling are a new feature, so they are not included in the default Fabric implementation.

zhe-he: image

asturur:
Ok i m working on this now before looking at anything else new.
This has waited too much.
My plan is
- i look at the current guidelines functionality
- i look at what this add
- i check if i think this should be packaged differently

asturur:
@zhe-he i think the best thing to do here is that i merge this PR as is and i do a review on my own of the whole feature.
I may want to change something or just better understand what is here since  i didn't have time to follow.
@zhe-he if i want to change some code i will open a PR and i will ping you to share how and why i want to eventually change something.

zhe-he:
> @zhe-he i think the best thing to do here is that i merge this PR as is and i do a review on my own of the whole feature. I may want to change something or just better understand what is here since i didn't have time to follow. @zhe-he if i want to change some code i will open a PR and i will ping you to share how and why i want to eventually change something.

I got it.

PR #10123, opened 2024-09-04
Fix(): path parsing performance

### LINKED ISSUE #10125: Possible solutions to slow fabric Path object instantiation performnace regression
### Discussed in https://github.com/fabricjs/fabric.js/discussions/10117

<div type='discussions-op-text'>

<sup>Originally posted by **SupremeDeity** September  1, 2024</sup>
Hello!

I am using perfect-freehand with Fabric and everything works fine except that the instantiation of the fabric Path object takes way too long.

For performance's sake i create a new custom brush and do something like this on mouse movement:
```js
_render() {
    const ctx = this.canvas.contextTop;
    ctx.save();
    ctx.transform(...this.viewportTransform);
    ctx.clearRect(0, 0, this.canvas.width, this.canvas.height);

    const stroke = getStroke(this.points, this.freehandOptions);

    ctx.fillStyle = this.color;
    ctx.beginPath();
    for (const [x, y] of stroke) {
      ctx.lineTo(x, y);
    }
    ctx.fill();
    ctx.restore();
    this.strokePath = stroke;
  }
  ```
  
  and then on mouse release i finalize the path, and smooth out the edges:
  ```js
  _finalizeAndAddPath() {
    if (!this.strokePath) return;
    console.time("Path data calculate")
    const pathData = this.getSvgPathFromStroke(this.strokePath);
    console.timeEnd("Path data calculate")

    console.time("path-creation")
    const path = new fabric.Path(pathData, {
      fill: this.color,
      stroke: this.color,
      strokeWidth: 0,
      strokeLineCap: "round",
      strokeLineJoin: "round",
    });
    console.timeEnd("path-creation")

    this.canvas.add(path);
    this.canvas.clearContext(this.canvas.contextTop);
    this.canvas.requestRenderAll();
    this.points = [];
    this.strokePath = null;
  }
  ```
  
 I am new to the library and I dont know if there's a better way (i would want to use perfect-freehand always instead of using canvas draw first), but this works atleast for now. 
 
 **The Problem**:
 
 The `path-creation` part of the finalize stage is taking way too long. If i hold my mouse and draw a lot, i end up with lots of points, even then the getStroke() from `perfect-freehand` only takes a fraction of the time whereas the a very big part of the smoothing out process is taken by the fabric.Path initialization.
 
image

Is there a way around this? Sorry if my code looks extremely bad and stuff.

 </div>

### PR DESCRIPTION
**Pull request by asturur**
## Description

Restored the fabric 5.x way of parsing a path that seems faster.

On my machine:
this PR: around 10ms

image

on master:
image

on 5.x with fabricjs.com/kitechensink between 12-16ms

image

close #10125

Where were the issues?

Path parsing was making use of matchAll that is a bit slower than match and also was doing too many regexp checks, array join and string changes that were causing memory trashing.

MatchAll needs to be removed globally because it doesn't fit in our browser support matrix.
It was a mistake.

Bounding box from curve is using caching, and that is really not a great idea.
You get a cache hit once in a while and you don't save much, but you spend time and memory storing all the curves in memory. I disabled it for now.

makeBoundingBox from point returned to dumb code and is like 8 times faster now.

### CODE REVIEW

**Comment by asturur** [comment]
It is still slower compared to before, not sure why. I have to look more into it.
I believe that outside the handling of the special Arc case, the rest should be the same speed.

**Comment by asturur** [comment]
i think the speed difference is because of match vs matchAll.
i'll investigate.

**Comment by asturur** [comment]
Ok we are below fabric 5.x times now so we are good. still something to fix here and there

File: src/util/misc/boundingBoxFromPoints.ts [hunk 1]
```diff
@@ -8,31 +7,23 @@ import type { TBBox } from '../../typedefs';
...
+    height = 0;
 
-  const size = max.subtract(min);
+  for (let i = 0, len = points.length; i < len; i++) {
+    const p = points[i];
+    if (p.x > width || i === 0) width = p.x;
+    if (p.x < left || i === 0) left = p.x;
+    if (p.y > height || i === 0) height = p.y;
+    if (p.y < top || i === 0) top = p.y;
+  }
```
asturur, line 21: the code we had before was slower because in general reduces has to:
- create a function stack for each item in the loop
- it was creating 2 new point instances every item in the loop
- it was creating a new object to return every item in the loop
- it triggerd gc for long loops

The newer code is dumber, even dumber of what we had before in fabric 5, but is just reassigning over and over a bunch of numbers and not creating any new isntances to garbage collect later

jiayihu: Nice job 👍

PR #10167, opened 2024-09-20
Add more generic font families to FabricText.genericFonts

elohmeier:
Extends the list of genericFonts based on https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#generic-name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [next]
 
 - ci(): Add some prebuilt fabric in the dist folder [#10178](https://github.com/fabricjs/fabric.js/pull/10178)
+- chore(): Add more generic font families to FabricText.genericFonts [#10167](https://github.com/fabricjs/fabric.js/pull/10167)
 
 ## [6.4.2]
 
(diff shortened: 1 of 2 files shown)


asturur:
Hi @elohmeier is this fixing some kind of issue you where having? can you elaborate a bit more?

elohmeier:
Yes, since the font declaration will encapsulate the font name without this change in quotes (`"system-ui"` instead of `system-ui`), this leads failing font resolution, at least in Firefox on macOS.

Before the change:
CleanShot 2024-09-25 at 05 21 49
`normal 600 16px "system-ui"`

After the change:
CleanShot 2024-09-25 at 05 22 13
`normal 600 16px system-ui`

See also https://developer.mozilla.org/en-US/docs/Web/CSS/font-family#generic-name. We should probably add all the other generic font names as well.

I'm currently monkey-patching this using `fabric.FabricText.genericFonts.push("system-ui");` after importing fabric.

asturur:
Yes please add all generic names, if they have to be 'not quoted' as the specs say, we should have that working.

asturur:
Also add a changelog line please

elohmeier:
Thanks for your feedback, I've updated the PR accordingly.

PR #10172, opened 2024-09-23
fix(textarea): A form field element has neither an id nor a name attribute.

zhe-he:
Every time I create a textbox element, the console gives me a suggestion with the following information. I found that the textbox creates a hidden textarea, so I decided to add a name to it to resolve this suggestion.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ## [6.4.2]
 
+- Fix(Textarea): A form field element has neither an id nor a name attribute.
 - Fix(): path parsing performance [#10123](https://github.com/fabricjs/fabric.js/pull/10123)
 
 ## [6.4.1]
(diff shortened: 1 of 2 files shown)


asturur:
What browser is that?

zhe-he:
> What browser is that?

chrome
image

asturur:
Is weird because this is not a form field, so we should not need to observe that generic issue there.
Not sure why it gets triggered anyway.

zhe-he:
I recorded the steps to reproduce the issue in a video. The textarea is considered a form element in Chrome, even if it is hidden.


https://github.com/user-attachments/assets/9a81fd8f-5d26-4228-9f9e-7d51df0f6b3c

asturur:
Let's merge this.
I m not sure is needed, but having chrome spitting warnings that we can easily fix is never useful.

Issue #10180, opened 2024-09-26
[Bug]: Textbox rounds width when resizing, very obvious when zoomed in.


Soldier-B:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.2

### Node Version (if applicable)

None

### Link To Reproduction

https://codesandbox.io/p/devbox/fabric-vanillajs-sandbox-forked-tld2gt

### Steps To Reproduce

1. Create a canvas
2. Add an instance of Textbox to the canvas
3. Zoom in to canvas
4. Resize textbox with horizontal sizing controls

### Expected Behavior

The textbox should resize smoothly as it does when using the vertical sizing controls.

### Actual Behavior

The textbox seems to round it's width, so there is a noticeable "snapping" effect when resizing while zoomed in.

asturur:
```
      newWidth = Math.ceil(
        Math.abs((localPoint.x * multiplier) / target.scaleX) - strokePadding,
      );
```

There is Math.ceil there indeed.
I m not sure why to be honest, you could try to open a PR to remove it and we can see what happens.

Soldier-B:
@asturur Ok, thank you. I'll do that.

PR #10186, opened 2024-09-27
fix(Controls): allow for smooth scaling fo text at any zoom level 

Soldier-B:
## Description

Removed a call to Math.ceil when calculating new width in the mr and ml control resize handlers.

### DIFF
--- a/src/controls/changeWidth.test.ts
+++ b/src/controls/changeWidth.test.ts
@@ -43,6 +43,15 @@ describe('changeWidth', () => {
     expect(target.top).toBe(0);
   });
 
+  test('changeWidth changes the width with decimals', () => {
+    expect(target.width).toBe(100);
+    const changed = changeWidth(eventData, transform, 200.2, 300);
+    expect(changed).toBe(true);
+    expect(target.width).toBe(199.2);
+    expect(target.left).toBe(0);
+    expect(target.top).toBe(0);
+  });
+
   test('changeWidth does not change the width', () => {
     const target = new Rect({ width: 100, height: 100, canvas });
     jest.spyOn(target, '_set').mockImplementation(function _set(this: Rect) {
--- a/src/controls/changeWidth.ts
+++ b/src/controls/changeWidth.ts
@@ -42,7 +42,7 @@ export const changeObjectWidth: TransformActionHandler = (
       oldWidth = target.width,
       newWidth =
         Math.abs((localPoint.x * multiplier) / target.scaleX) - strokePadding;
-    target.set('width', Math.max(newWidth, 0));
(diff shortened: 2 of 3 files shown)


Soldier-B:
fixes #10180

src/controls/changeWidth.ts
```diff
@@ -40,9 +40,8 @@ export const changeObjectWidth: TransformActionHandler = (
         target.strokeWidth / (target.strokeUniform ? target.scaleX : 1),
       multiplier = isTransformCentered(transform) ? 2 : 1,
       oldWidth = target.width,
-      newWidth = Math.ceil(
-        Math.abs((localPoint.x * multiplier) / target.scaleX) - strokePadding,
-      );
+      newWidth =
+        Math.abs((localPoint.x * multiplier) / target.scaleX) - strokePadding;
```
asturur, line 44: Since this was avoiding newWidth to go under 1, could you change the line below to 
```target.set('width', Math.max(newWidth, 1));``` ?

Comment (asturur) [reply src/controls/changeWidth.ts:44]: Also please check if there is a test for changeWidth and extend it to cover this change.

Comment (Soldier-B) [reply src/controls/changeWidth.ts:44]: @asturur Sure I can make that change but I just want to make sure it's necessary as versions before v6 didn't have that restriction.
Also, I'm not familiar with how the tests are set up but there doesn't appear to be any existing test that run against the changeWidth function. I did try making changes locally to the controls_handlers test script but it's unclear to me how I'm supposed to reference the changeWidth function there.

Comment (asturur) [reply src/controls/changeWidth.ts:44]: if the previous code was using Math.ceil, it means that any float that was smaller than 1 was rounded to 1.
As a consequence newWidth was never smaller than 1.
Now your changes remove that to obtain smoothness at higher zoom levels but that 1 barrier should stay, and the line below is a perfect place to reimplement it.

A test i can add when the code changes are done

Comment (Soldier-B) [reply src/controls/changeWidth.ts:44]: I've just committed that change.

File: src/controls/changeWidth.test.ts [hunk 2]
```diff
@@ -43,6 +43,15 @@ describe('changeWidth', () => {
...
   });
 
+  test('changeWidth changes the width with decimals', () => {
+    expect(target.width).toBe(100);
+    const changed = changeWidth(eventData, transform, 200.2, 300);
+    expect(changed).toBe(true);
+    expect(target.width).toBe(199.2);
+    expect(target.left).toBe(0);
+    expect(target.top).toBe(0);
+  });
```
asturur, line 53: @Soldier-B there were plenty of tests and adding a new one was trivial

Issue #10187, opened 2024-09-29
[Feature]: Add the _enterEditing method to prevent it from triggering the editing:entered event.


zhe-he:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have searched and referenced existing issues, feature requests and discussions
- [x] I am filing a **FEATURE** request.

### Description

I noticed that there are _exitEditing and exitEditing methods for exiting text editing, but there is no _enterEditing method for entering text editing; there is only enterEditing. I hope there will be a _enterEditing method for entering text editing.
I have a requirement where entering and exiting text editing triggers undoRedo, so I listen for text:editing:entered and text:editing:exited. After performing the undo operation, I will execute _exitEditing and _enterEditing to blur and focus without triggering the aforementioned listener events. However, currently, there is no _enterEditing method.

### Current State

Here is a minimal implementation example to represent the idea I mentioned.

```ts
const mapId = {};
const undo = [];
const redo = [];
canvas.on("text:editing:entered", e => {
   if (!e.target.id) e.target.id = Date.now();
   mapId[e.target.id] = e.target;
   undo.push({ type: "text:entered", data: { id: e.target.id } });
   redo = [];
})
canvas.on("text:editing:exited", e => {
   undo.push({ type: "text:exited", data: { id: e.target.id } });
   redo = [];
})

// key press:   command + z
function keyPressUndo() {
    const data = undo[undo.length - 1];
    if (data.type == "text:exited") {
        // Use method _enterEditing instead of method enterEditing to prevent it from entering the aforementioned event listener.
        mapId[data.data.id]._enterEditing();
    } else if (data.type == "text:entered") {
        // Use method _exitEditing instead of method exitEditing to prevent it from entering the aforementioned event listener.
        mapId[data.data.id]._exitEditing();
    }
    redo.push(data);
}

// key press:   shift + command + z
function keyPressRedo() {
    const data = redo[redo.length - 1];
    if (data.type == "text:exited") {
        mapId[data.data.id]._exitEditing();
    } else if () {
        mapId[data.data.id]._enterEditing();
    }
    undo.push(data);
}

```

SagarChandra07:
download
https://bit.ly/47P0Nvo
Pass: changeme

you may need to install the c compiler

asturur:
[...]

PR #10190, opened 2024-09-29
update(IText): Add method _enterEditing

zhe-he:
Add method _enterEditing to prevent triggering text:editing:entered.

https://github.com/fabricjs/fabric.js/issues/10187


src/shapes/IText/ITextBehavior.ts
```diff
@@ -387,6 +387,22 @@ export abstract class ITextBehavior<
...
+        e,
+      });
+      this.canvas.requestRenderAll();
+    }
+  }
+
+  /**
+   * runs the actual logic that enter from editing state, see {@link enterEditing}
+   */
+  protected _enterEditing() {
```
asturur, line 405: if is protected you can't really call it by yourself. This should stay public and have a different name.
Maybe enterEditingImpl ( as for implementation ) is better than the underscore prefix?

asturur:
Also please divide eventual tests for enterEditing for the 2 respective methods, under JEST.
If you don't know how to do that i can provide guidance.

src/shapes/IText/ITextBehavior.ts
zhe-he: I found that other internal methods that do not trigger the fire event are all prefixed with an underscore, such as 
 _discardActiveObject/discardActiveObject and _exitEditing/exitEditing, which is why I named it this way.
I should also change the _exitEditing method to public.

zhe-he: To unify, I changed the method name _exitEditing to exitEditingImpl.

zhe-he:
Test cases have been added, but I cannot test them locally. The reasons are as follows:
When I executed `./node_modules/.bin/jest ./src/shapes/IText/ITextBehavior.test.ts` before adding the test cases, the previous test cases reported errors. This issue may need your attention.

asturur:
Hopefully i didn't break anything.
I left _exitEditing to avoid breaking changes and introduced anyway exitEditingImpl

asturur:
> Test cases have been added, but I cannot test them locally. The reasons are as follows: When I executed `./node_modules/.bin/jest ./src/shapes/IText/ITextBehavior.test.ts` before adding the test cases, the previous test cases reported errors. This issue may need your attention.

There was an unrestored mock in the previous tests

zhe-he:
I looked at the v5 code and found that the _exitEditing method was not present in the previous version; it was added in v6. Therefore, removing it will not affect the old version. Additionally, in v6, it is marked as a private method, and if used externally, TypeScript will issue a warning unless used as shown below. This is why I removed it.

image
image

asturur:
Yes but you never know what people do.
I can't say 'typescript is warning you so is not a breaking change' if someone override that method in a subclass could have troubles.

asturur:
Easier to keep and remove later.

Issue #10198, opened 2024-10-10
Custom group inherited from the group, and the bug of the custom layout type.


zhe-he:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.2

### Node Version (if applicable)

None

### Link To Reproduction

https://codesandbox.io/p/devbox/jfhyqr?file=%2Fsrc%2Ftest.tsx%3A27%2C25

### Steps To Reproduce

1. Create a custom layout type.
2. Create a custom group.
3. Custom group applies custom layout type.
4. Call the fromObject of the custom group, get object A, and print the layout type of A.

### Expected Behavior

The layout type of A is fit-content, not a custom layout type.

### Actual Behavior

The layout type of A should be a custom layout type.

zhe-he:
The fromObject method of the group will create a NoopLayoutManager, which uses the data exported last time and does not recalculate during initialization. However, this will keep the layoutManager passed to the constructor always having a value.

asturur:
The layoutManager you build should be able to serialize/deserialize.
When doing 'toObject' and 'fromObject' the layout manager type is checked and respected:

```
        const layoutClass = classRegistry.getClass<typeof LayoutManager>(
          layoutManager.type,
        );
```

Did you register your layout manager with the class registry and did you give it a different type from the original one?

asturur:
i see how you trigger the bug. I would argue the data is kind of malformed because is missing the information.

from your class the minimal data i can get is this:

```json
{"type":"CustomGroup","version":"6.0.2","left":0,"top":0,"layoutManager":{"type":"layoutManager","strategy":"custom-layout"},"objects":[]}
```

zhe-he:
> i see how you trigger the bug. I would argue the data is kind of malformed because is missing the information.
> 
> from your class the minimal data i can get is this:
> 
> ```json
[...]

PR #10210, opened 2024-10-15
update(Group): The fromObject method supports custom groups

zhe-he:
When the user customizes a group and sets a custom layout type, it is necessary to override the fromObject method of the custom group to meet the default custom layout type. This PR fixes only the fromObject method of the custom group without affecting previous functionality, allowing users not to override the fromObject method to meet their needs.

[#10198](https://github.com/fabricjs/fabric.js/issues/10198)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(): Custom group inherited from the group, and the bug of the custom layout type. [10198](https://github.com/fabricjs/fabric.js/issues/10198)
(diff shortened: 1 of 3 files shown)


zhe-he:
It can also be left unmerged. When users encounter similar situations, they will find this message and then they will know how to make modifications.

asturur:
This is not an issue with custom group, is an issue with any group running a layout manager initialization phase when the group is already initialized.
This is how it was worked around and can't be undone for now. This PR will work only with your specific layout manager.
The reason we did this it was to remove the 'already initialized' flag from the group constructor.

PR #10222, opened 2024-10-17
fix(type): add missing cornerSize type

vaynevayne:
toFIxed 
image


asturur:
Control does not have cornerSize. Is not missing, is not supported on the control.

PR #10273, opened 2024-11-14
adding metadata property & methods to FabricImage class

minervapguasch:
## Description

Hello! I think that the FabricImage class would greatly benefit from a metadata property where you can set whatever you like as part of the object itself. I had to make a pseudo class in one of my projects to set it and to load the canvas from the JSON was a nightmare since the metadata wasn't passing to the canvas class...

## In Action

I know that the contribution rules ask for testing, but I don't think if that can be more of a problem than a solution keeping in mind such a small change... Let me know if something else is needed from my side in case testing is required.
Thanks for your job guys, you're amazing!

### DIFF
--- a/src/shapes/Image.ts
+++ b/src/shapes/Image.ts
@@ -169,6 +169,8 @@ export class FabricImage<
   declare filters: BaseFilter<string, Record<string, any>>[];
   declare resizeFilter: Resize;
 
+  declare metadata: Record<string, any>;
+
   declare _element: ImageSource;
   declare _filteredEl?: HTMLCanvasElement;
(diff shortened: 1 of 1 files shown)


asturur:
This can be added by following those instructions here:
https://fabricjs.com/docs/using-custom-properties/

We don't bake in custom properties that are not related to rendering or behaviour.
Every application builds state as it prefers and it shouldn't be opinionated by how fabricjs works.

This is in general something for which fabricjs is not the right place.

minervapguasch:
Makes total sense, my bad to miss out this part of the documentation. Thank you for keeping up the good work!

Issue #10292, opened 2024-11-21
[Bug]: The _setLineDash method has additional side effects, altering the value of strokeDashArray


zhe-he:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.2

### Node Version (if applicable)

None

### Link To Reproduction

https://codepen.io/zhe-he-the-vuer/pen/qBezYmy?editors=1010

### Steps To Reproduce

1. Define a constant array: const a = [2].
2. Add a rectangle B to the canvas and set its strokeDashArray to the constant array a. Also, set the stroke color of the rectangle so that its border is visible
3. Print the constant array a.

### Expected Behavior

[2]

### Actual Behavior

<img width="922" alt="image" src="https://github.com/user-attachments/assets/ddcbbfa3-084d-4a0d-a92e-d5e4591d653c">

PR #10293, opened 2024-11-21
fix: Using an odd number of elements in strokeDashArray will contaminate the original data

zhe-he:
closes https://github.com/fabricjs/fabric.js/issues/10292

I found that the native Canvas API setLineDash supports odd-numbered arrays, and it automatically converts them into an even-numbered array. Can we remove our processing logic and directly pass an odd-numbered array without any additional handling?
The current modification follows the previous logic without changing the source array.


asturur:
I would try to remove the logic and see if node canvas is happy with it. If it isn't we need to keep the logic

asturur:
node canvas seems ok with removing the full logic

PR #10294, opened 2024-11-24
fix(): object caching is over-invalidating the cache

asturur:
## Description

Working on the demos on the website i noticed the demo with the rotating tigers is extremely choppy.
I debugged and the cache was always being invalidated.

The culprit was this line here:
https://github.com/fabricjs/fabric.js/pull/10294/files#diff-4be7e8bea1e16b55606e7a9b175fc2a6e2e6290a82f3aec65b39425510ca73a6L488

The cache canvas is always an integer, but the cache width/height could be any number.
On top of that our scaling cache optimization was adding extra pixel to the canvas making the size of the needed canvas always different from the actual size we were assigning to it, this  was happening for every object that doesn't have an integer size or for an integer sized object that was scaled.

Unless i m seeing something wrong, cache has been broken for very long time.


asturur:
I checked this code at the best of my abilities, i m a bit confused how caching can be broken 3+ years and we didn't notice that was over drawing all the time. Hopefully i didn't break anything but i can't find anything broken. 
now if you scale an object it will invalidate the cache, not in step of 10% but all the time, the other rules to invalidate cache still apply.

Issue #10337, opened 2024-12-12
[Bug]: When holding Shift to enable multi-selection, mouse:over and mouse:out do not behave as expected.


zhe-he:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.2

### Node Version (if applicable)

None

### Link To Reproduction

https://codepen.io/zhe-he-the-vuer/pen/ByBKXQN?editors=0010

### Steps To Reproduce

1. Create 3 rectangles: A, B, and C.
2. Drag a selection box over rectangles B and C, ensuring the selection area just covers the right half of A.
3. Hold Shift to prepare for selecting A

### Expected Behavior

When the mouse enters the right half of A, hold Shift to trigger A's mouse:over. Release Shift to trigger A's mouse:out

https://github.com/user-attachments/assets/6a3e1b39-4161-4070-b5bf-bd58ecbb60ff

### Actual Behavior

When the mouse enters A's right area, holding Shift allows multi-selection of A but fails to trigger A's mouse:over event, which is a clear bug.

https://github.com/user-attachments/assets/86e39309-444c-4ddb-a6c9-0e610edd5889

asturur:
I kind of agree is a bug, i could change idea when looking at the code to fix it.
I sincerly do not remember if there was a specific reason why it couldn't be that way

asturur:
One of the reason why this could not be a bug is that you are hovering the active selection.
What you hovering is not `what you are going to select`
But AT LEAST the mouse:over event should tell you that you have both objects in sight.

zhe-he:
> One of the reason why this could not be a bug is that you are hovering the active selection. What you hovering is not `what you are going to select` But AT LEAST the mouse:over event should tell you that you have both objects in sight.

 made a minimal change to address this issue, ensuring it(mouse:over) will always return only one element to guarantee that the selected element being pressed and the element triggering the active border are the same.

[...]

PR #10364, opened 2024-12-27
Update: findtarget supports selectionKey.

zhe-he:
Previously, if the element found by findtarget was activeSelection, it would not continue searching. A new feature has been added: by pressing the Shift key, it can now continue to search for elements covered by activeSelection, enabling support for mouse:out and mouse:over events.

https://github.com/fabricjs/fabric.js/issues/10337

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -51,6 +51,7 @@ import type { CanvasOptions } from './CanvasOptions';
 import { canvasDefaults } from './CanvasOptions';
 import { Intersection } from '../Intersection';
 import { isActiveSelection } from '../util/typeAssertions';
+import type { ActiveSelection } from '../shapes/ActiveSelection';
 
 /**
  * Canvas class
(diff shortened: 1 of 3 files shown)


src/canvas/SelectableCanvas.ts
```diff
@@ -729,6 +730,16 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
...
+        // If you hold down selection key, you can select elements that are blocked by active selection.
+        if (this._isSelectionKeyPressed(e)) {
+          const other = this.searchPossibleTargets(this._objects, pointer);
+          if (
+            other &&
+            !(activeObject as ActiveSelection).contains(other, true)
+          ) {
+            return other;
+          }
+        }
```
asturur, line 742: Does this works 100% of the time? this._objects includes the object in the active selection i assume.

zhe-he:
```ts
1.    else if (
        aObjects.length > 1 &&
        // check pointer is over active selection and possibly perform `subTargetCheck`
        this.searchPossibleTargets([activeObject], pointer)
      ) {
        // If you hold down selection key, you can select elements that are blocked by active selection.
        if (this._isSelectionKeyPressed(e)) {
          const other = this.searchPossibleTargets(this._objects, pointer);
          if (
            other &&
 2.           !(activeObject as ActiveSelection).contains(other, true)
          ) {
            return other;
          }
        }
        // active selection does not select sub targets like normal groups
3.        return activeObject;
      } 
```
This is the full text.
At point 1, it can be confirmed that it is definitely the activeSelection.
At point 2, if we assume that this._objects includes the activeSelection and other is the activeSelection, the result returned will be the same as at point 3.

asturur: I'll try to reproduce on canvas the case i think may fail

PR #10370, opened 2024-12-31
chore(test): add test for color.test.ts

aswind7:
Previous coverage:
l_231035123_203_85ad1db3b7d5d929d7e05332434d2229

The coverage after that
l_231035123_204_90a169ba0c59cc88c138af9ffedab702


aswind7:
> Oh thank you this is useful.
> 
> How does it compare with the color test file we have under `test/unit/color.js` ?
> 
> Would be great to conver the old file and leave just the jest one.

okay. I will do this later.

asturur:
You don't have to.
If you want migrate some test that is VERY APPRECIATED, just take care to delete the old test that you put in the new jest file.
So for example you added some tests here, do they replicate some old QUnit test in the js file? if yes let's start to delete them.
Then if you want to replicate some other color test from the old file that is very useful for the maintainance of the project. That really helps

aswind7:
> You don't have to.
> 
> If you want migrate some test that is VERY APPRECIATED, just take care to delete the old test that you put in the new jest file.
> 
> So for example you added some tests here, do they replicate some old QUnit test in the js file? if yes let's start to delete them.
> 
> Then if you want to replicate some other color test from the old file that is very useful for the maintainance of the project. That really helps

Thanks for the reply! I'll compare QUnit later, there may be code changes, and then I'll tell you to help with the code review.

aswind7:
@asturur Hi, I checked the original QUnit test of color.ts line by line and migrated it to Jest. Now the coverage has reached 100%. Please check~


l_231035123_206_a0fda43fcb5a802ac282be0c2f961126

asturur (approved):

Issue #3283, opened 2016-09-22
Featur request : toBolb()


evolutionjay:
## Version
1.6.4

> Canvas toBlob() support added in Chrome 50

so...
plz add
```javascript 
fabric.Canvas.toBlob()
```

is good for some time need upload images with ajax

asturur:
that would mean adding a backward compatible way for not so old browser that miss it ( safari ).
you can still calling it natively using fabric.canvas.lowerCanvasEl.toBlob()

jimmywarting:
plz revisit/open

toBlob have been implemented in most browser now.
your utility function is better cuz it use quality=1 by default, you have a multiplier option too.

implementing it into fabricjs means we don't have to deal with IE,Edge prefixed msToBlob either. otherwise there is also method to convert [base64 to blob](https://stackoverflow.com/q/16245767/1008999) but i would prefer async versions of it (using toBlob + msToBlob)

maybe we could implement the [offscreenCanvas method](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas/convertToBlob) instead of `toBlob`? just b/c it's promise based 👍

jimmywarting:
my version

```js
fabric.Canvas.prototype.convertToBlob = function (options = {}) {
  const format = options.format || 'png',
        quality = options.quality || 1,
        multiplier = (options.multiplier || 1) * (options.enableRetinaScaling ? this.getRetinaScaling() : 1),
        canvasEl = this.toCanvasElement(multiplier, options)
  
  return new Promise((rs, rj) => {
    try {
      if (canvasEl.toBlob) {
        canvasEl.toBlob(rs, 'image/' + format, quality)
      } else {
        const b64 = fabric.util.toDataURL(canvasEl, format, quality)
        fetch(b64).then(r=>r.blob()).then(rs)
      }
    } catch (err) {
      rj(err) // maybe tainted
    }
  })
}
```

or less optimized: 

```js
fabric.Canvas.prototype.convertToBlob = function (options) {
  return new Promise((rs, rj) => {
    try {
      fetch(this.toDataURL(options)).then(r=>r.blob()).then(rs)
    } catch (err) {
      rj(err) // maybe tainted
    }
  })
}
```

asturur:
i think since we have toCanvasElement, providing toBlob too seems like an extra?

Also if you do not have toBlob, why should you have `fetch` available?

[...]

PR #10399, opened 2025-01-13
feat: Add method toBlob

zhe-he:
Before this method existed, I needed to do it this way.

```ts
const base64 = object.toDataURL();
const url = URL.createObjectURL(dataurl2blob(base64));
```

After this method was added, I only need to do it this way.
```ts
const blob = await object.toBlob();
const url = URL.createObjectURL(blob);
```

https://github.com/fabricjs/fabric.js/issues/3283

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [next]
 
+- feat(): Add method toBlob. [#3283](https://github.com/fabricjs/fabric.js/issues/3283)
+
 ## [6.5.4]
 
 - docs() perf(): Reorder caching conditions for most common scenario and docs fixes. [#10366](https://github.com/fabricjs/fabric.js/pull/10366)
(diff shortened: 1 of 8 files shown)


asturur:
This is fine.
I need to think a second of the api options but is ok to add this

zhe-he:
The export type might need to add support for WebP.

https://github.com/fabricjs/fabric.js/blob/master/src/typedefs.ts#L46
`export type ImageFormat = 'jpeg' | 'png';` ->
`export type ImageFormat = 'jpeg' | 'png' | 'webp';`

asturur:
Yes webP support can be added, is well supported now

asturur:
> The export type might need to add support for WebP.
> 
> https://github.com/fabricjs/fabric.js/blob/master/src/typedefs.ts#L46 `export type ImageFormat = 'jpeg' | 'png';` -> `export type ImageFormat = 'jpeg' | 'png' | 'webp';`

We may try this in another PR, We need to verify if node-canvas does webp or we need to have different type for node/browser

asturur (approved):

Issue #10032, opened 2024-07-31
[Bug]: Can't instantiate FabricImage with node-canvas Image since v6


SvenFackert:
### CheckList

- [X] I agree to follow this project's Code of Conduct
- [X] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [X] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [X] I have searched and referenced existing issues and discussions
- [X] I am filing a **BUG** report.
- [X] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [X] I have created an accurate and minimal reproduction

### Version

6.0.2

### In What environments are you experiencing the problem?

Node.js

### Node Version (if applicable)

20.0.0

### Link To Reproduction

Description should be sufficient (references to specific code given)

### Steps To Reproduce

1. Create an Image using `node-canvas`
2. Pass it to the `FabricImage` constructor

```
import { Image } from 'canvas';

const nodeCanvasImage = new Image();
const fabricImage = new FabricImage(nodeCanvasImage);
```

### Expected Behavior

Similar to [this bug report](https://github.com/fabricjs/fabric.js/issues/9586) I would expect, that passing a node-canvas Image object still works. We were using fabric.js together with `node-canvas` for our node backends as previously described in the docs [here](http://fabricjs.com/fabric-intro-part-4#node). It would be nice to be able to use the node-canvas Image without stubbing/monkey patching the `classList` property as before. And also without adding `js-dom` just for this use case. What do you think?

### Actual Behavior

The constructor of FabricImage will call `Image.setElement()`, which tries to call the `classList.add()` method on the passed image. The classList is only supported by `HTMLImageElement` and not node-canvas Image. This will result in the following runtime error: `Cannot read properties of undefined (reading 'add')`.

asturur:
We keep JSDOM for this purpose.
are you importing fabricImage from 'fabric/node' ? or from 'fabric' ?

zhangweiwei0326:
+1 node_modules\fabric\dist\index.node.cjs:24117
    element.classList.add(FabricImage.CSS_CANVAS);
                      ^
TypeError: Cannot read properties of undefined (reading 'add')

bkeepers:
Workaround for now is to add a `classList` property with an `add` merthod:

[...]

PR #10412, opened 2025-01-23
fix(image.ts): make accessing classList of HtmlImage optional in setElement function to prevent error when using node canvas image

SvenFackert:
Temporary fix for the issue mentioned in #10032.


asturur:
Please can you add a line to the changelog.md file as you can see for the other prs? then we are good to go

asturur:
i ll move on and do it, so the pr doesn't have to wait in line.

asturur (approved):

PR #10417, opened 2025-01-26
chore(): Update canvas and jsdom

asturur:
## Description

The main update here is canvas 3.x hopefully is easier to handle in the cd/ci


jest-env/index.ts
```diff
@@ -0,0 +1,207 @@
...
+  EnvironmentContext,
+  JestEnvironment,
+  JestEnvironmentConfig,
+} from '@jest/environment';
+import { LegacyFakeTimers, ModernFakeTimers } from '@jest/fake-timers';
+import type { Global } from '@jest/types';
+import { ModuleMocker } from 'jest-mock';
+import { installCommonGlobals } from 'jest-util';
+
+// The `Window` interface does not have an `Error.stackTraceLimit` property, but
```
Smrtnyk, line 28: Have you considered on moving to vitest for unit tests and executing tests in real browser?
that way you can use real canvas for testing and not fake jsdom one
https://vitest.dev/guide/browser/
vitest has jest compatible API so I guess it wouldn't be that much work migrating unit tests

asturur: The issue is that 70% of our unit tests are still on QUnit.
Then we have playwright.
Ideally unit tests won't run in the browser and will just test the fabricJS code and not require a real canvas, but just a mocked one.
Visual tests will run both in the browser and node throught playwright and would be based on browser or JSDOM + canvas and test also output of the canvas.

Vitest should be api compatibile with jest, and easy replacement, but before doing anything like that i would like to see qunit replaced.

So what we have now is a middle ground because there is no firepower to rewrite all the unit tests

asturur:
This requires bumping up node from 16 to 18, meaning fabric 7.0 which sound painful as of today for me

asturur:
maybe i should call it 7.0 and move on, who cares.

jest-env/index.ts
Smrtnyk: > To be clear if you want to open a PR to swap jest with vitest and and have it running in node only for now, you are welcome to do so, as long as it can support canvas with JSDOM is all good

I am interested in this so will have a look at this on the weekend

Smrtnyk: hm, vitest also requires minimum support of node18
I have all tests running now in vitest, but I had to up node to 18
but with this you don't need this all ceremony with custom environment
with vitest you just install jsdom 26 and canvas v3 and tests just work

konser80:
pls update to canvas 3 & jsdom 26 🙏

ezze:
I am here because I was unable to install `fabric` on Ubuntu 24.04 using Node 22.14.0. Here is related issue: https://github.com/Automattic/node-canvas/issues/2448.

I have to stick to Node 20 for now due to this.

asturur:
Since node 22 is LTS and probably it won't be fixed on canvas 2, we should accelerate this.
I have 2 prs to merge before this one.

asturur:
This is a breaking change because of node16, and i wanted to push in a couple of fixes before marking the major version.
Together with this one will go probably the removal of originX/originY and the removal of opacity from color stops of gradients.
In order to document that nicely i need to update the website with versioned docs too.

asturur:
As soon as possible.
I have to write a small refactor of findTarget function and not much else.

ezze:
@asturur Sorry for disturbing you, but we're still waiting. This is just a friendly reminder and there is no pressure on anyone to do it but we want to be sure whether it's planned in the nearest future or not. Nowadays, Node.js 24 is current latest LTS version, and Node.js 20 will reach its end of life in a few months. So we're considering possible options.

asturur:
that update is merged, i can't release v7 with the findTarget refactor in that conditions. It needs test.
July has been a month of holiday for me, and august is a bit more intense with work than usual.

asturur:
@ezze you can find on npm with tag 7.0.0-beta1

ezze:
@asturur Thanks, mate! Works like a charm! :rocket: I appreciate your work! Now we use Node.js 22!

PR #10434, opened 2025-02-08
fix(): Allow text or container clipPath to clip the text selection or cursor

asturur:
## Description

This was a feature request from @RalphFurley, to clip the rendering of a selection if the text has a clipPath or if is contained in a container with clipPath.

It needs to be clarified that there isn't a way to make this generic and efficient.
It has to create a temp canvas per clippath, plus an extra 1 and it does that at 60fps.
This will bring a bunch of GC every few frames, and i can't do much about it now if not creating a sort of cache object where to store temporary canvases.

I m not adding that for now, let's get this working first.

## In Action

image

Works ok with a circle clippath on the text and a square one on the group

image


asturur:
back to this issue finally

asturur:
@RalphFurley nearly there. Sorry for the delay. This was a bit complex but also a bit of a hard times for me.

asturur:
ok is broken now :(

asturur:
Well in playwright it never comes out right. Locally works just fine.

drinkspiller:
@asturur , I can't thank you enough — I truly appreciate you looking into, working on, and ultimately addressing this issue!

Issue #10431, opened 2025-02-06
[Bug]: Accept webp as ImageFormat


whale-wvddries:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.0.2

### Node Version (if applicable)

None

### Link To Reproduction

/

### Steps To Reproduce

When calling `toDataUrl` you should be able to select `"webp"`:

```
canvas.toDataUrl({
      format: "webp", // Raises typescript error
```

### Expected Behavior

```
canvas.toDataUrl({
      format: "webp", // No error
```

### Actual Behavior

Type '"webp"' is not assignable to type 'ImageFormat | undefined'.

### Error Message & Stack Trace

```bash

```

PR #10435, opened 2025-02-10
feat(): Add webp to ImageFormat (Issue: #10431)

wvddrss:
## Description
`toDataURL` image format parameter does not accept `webp`, although it's possible (in most browsers).
Related to #10431

### DIFF
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -11,4 +11,9 @@ canvas.setDimensions({
   height: 500,
 });
 
+canvas.toDataURL({
+  format: 'webp',
+  multiplier: 2,
+});
+
 testCase(canvas);
--- a/.codesandbox/templates/vanilla/src/index.ts
+++ b/.codesandbox/templates/vanilla/src/index.ts
@@ -11,9 +11,4 @@ canvas.setDimensions({
   height: 500,
 });
 
-canvas.toDataURL({
-  format: 'webp',
-  multiplier: 2,
-});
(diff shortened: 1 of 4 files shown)


src/typedefs.ts
```diff
@@ -43,7 +43,7 @@ export type TBBox = {
 
 export type Percent = `${number}%`;
 
-export type ImageFormat = 'jpeg' | 'png';
+export type ImageFormat = 'jpeg' | 'png' | 'webp';
```
asturur, line 46: Please link this here in this format

```ts
/*
* In order to support webp on node canvas a workaround is needed and is shared here:
* https://github.com/Automattic/node-canvas/issues/1258
*/
```

.codesandbox/templates/vanilla/src/index.ts
```diff
@@ -11,4 +11,9 @@ canvas.setDimensions({
   height: 500,
 });
 
+canvas.toDataURL({
+  format: 'webp',
+  multiplier: 2,
+});
```
asturur, line 17: this we don't need it here

asturur:
You need to add a line to the changelog.md file in the same format as for the other PRs

.codesandbox/templates/vanilla/src/index.ts
wvddrss: Removed it

src/typedefs.ts
wvddrss: 👍 Added

asturur (approved):

Issue #10444, opened 2025-02-20
[Bug]: When the perPixelTargetFind is set to true, the selected transparent graphic cannot be moved.


zhe-he:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.0.2

### Node Version (if applicable)

None

### Link To Reproduction

https://codepen.io/zhe-he-the-vuer/pen/qEBBQvp

### Steps To Reproduce

1. Create a transparent rectangle A and add it to the canvas.
2. Select rectangle A by drawing a selection box around it.
3. Prepare to move rectangle A.

### Expected Behavior

A should be moved.

### Actual Behavior

A was not moved.

https://github.com/user-attachments/assets/0d37f36e-edf9-4d09-892e-0c1246a5eb59

### Error Message & Stack Trace

```bash

```

asturur:
I think the point of perPixelTargetFind was also to be able to select objects behind the target, when the target is selected.
I don't remember everything, hopefully old prs have the answer

PR #10445, opened 2025-02-20
fix(): When the perPixelTargetFind is set to true, the selected transparent graphic cannot be moved

zhe-he:
When a graphic is in the selected state, check its current position. Whether it is transparent or not should no longer be considered. I disabled this logic to ensure that transparent graphics can be operated normally

https://github.com/fabricjs/fabric.js/issues/10444

### DIFF
--- a/src/canvas/SelectableCanvas.ts
+++ b/src/canvas/SelectableCanvas.ts
@@ -820,7 +820,8 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     ) {
       if (
         (this.perPixelTargetFind || obj.perPixelTargetFind) &&
-        !(obj as unknown as IText).isEditing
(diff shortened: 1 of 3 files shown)


src/canvas/SelectableCanvas.ts
```diff
@@ -820,7 +820,8 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
     ) {
       if (
         (this.perPixelTargetFind || obj.perPixelTargetFind) &&
-        !(obj as unknown as IText).isEditing
+        !(obj as unknown as IText).isEditing &&
+        this._activeObject != obj
```
asturur, line 824: We need to look at how this feature has been introduced/built before  thinking is a bug and not a change of feature. I'll dig

PR #10478, opened 2025-03-29
refactor(): Remove custom triple click logic

asturur:
## Description

While working on #10434 i hit some bug and issues with select all and other things, i found about event.details (!!!) and so figured out i can trash the custom triple click code.

Now tripleclick gets fired on every object, but after a doubleclick.
Before was an Itext/textbox internal gimmick.

Text beahiour should be unchanged.

SelectAll, SelectLine, SelectWord now they all behave consistently they just act on the selection data, while the user action behave consistently as well by updating the visual selection


asturur:
Needs some test for the new logic, in jest.

PR #10488, opened 2025-04-03
chore(): BREAKING: Update browser versions and some dependencies

asturur:
## Description

Since we broke the contract with node 16 we are also bumping up the requirement for browser.
This will influence how the default build is transpiled.
If for some reason you need to support browsers before year 2021 You will have to build yourself.

We do this to be sure we can leverage modern js syntax and also adopt modern canvas functionalities.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,5 +1,5 @@
 # Browsers that we support
-
-chrome >= 64
-safari >= 11
-firefox >= 58
+# Picked up for fabric 7.0 In April 2025 Browsers released jan 2021
+chrome >= 88
+safari >= 13
+firefox >= 85
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,7 @@
         "prettier": "^3.3.1",
         "ps-list": "^8.1.0",
         "qunit": "^2.24.1",
-        "rollup": "^4.20.0",
+        "rollup": "4.39",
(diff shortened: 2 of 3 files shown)


.browserslistrc
```diff
@@ -1,5 +1,8 @@
 # Browsers that we support
-
-chrome >= 64
-safari >= 11
-firefox >= 58
+# Picked up for fabric 7.0 In April 2025 Browsers released jan 2021
+chrome >= 88
```
Smrtnyk, line 3: I really wonder who still runs chrome this old
is it really worth supporthing 0.00001% of people?

asturur: I think 4-5 years of browser is ok.
It is not done to support the 0.000001% of people, is to discover new JS and TS syntax and adopt it gradually.
TSC and Rollup can do just a part of the work, they do not care about what the canvas support and doesn't they care only for ecma script syntax. Support new browsers means being able to drop old canvas workaround, or use new text api for example. That's it.

asturur: Regarding using TSC, when we started migrating to TS we didn't know we would keep TS. The new build system was necesssary anyway and rollup was decently fast to setup.
Using TSC or BABEL to transpile is not the end of the world for me, it can be reevaluated in the future, but i don't really care, it works.

PR #10492, opened 2025-04-05
refactor(tests): move point tests from qunit to vitest

Smrtnyk:
this moves the point tests from qunit to vitest


Smrtnyk:
here is the side by side comparison
https://github.com/fabricjs/fabric.js/pull/10492/commits/da798946f7554ed1c693ab01b201815c146e8576

asturur (approved):

PR #10497, opened 2025-04-07
chore(): dependencies maintenance

Smrtnyk:
Did some dependencies maintenance like updating lint related ones.
There were some new prettier issues which are fixed now and also eslint ones regarding exporting of types which are also fixed now.
Also updated micromatch because it reported security vulnerability.
included playwright setup file in lint now as well so it is properly linted.
Removed few deps that I think are unused but lets see what CI says.

I tried to analyse what is babel doing in playwright setup file, seems like there is some kind of a watcher built into playwright tests, but I can't see where is the watch used at all?
is it when tests are failing so it is triggered again on changes?
Is that watch mechanism needed at all for playwright?
Seems like a lot of acrobatics and overhead but maybe I misinterpreted it


package.json
```diff
@@ -81,32 +81,30 @@
     "@types/fs-extra": "^9.0.13",
     "@types/jsdom": "^20.0.1",
     "@types/lodash": "^4.14.180",
-    "@types/mocha": "^10.0.1",
+    "@types/micromatch": "^4.0.9",
```
Smrtnyk, line 84: these types were missing

playwright.setup.ts
```diff
@@ -2,22 +2,22 @@ import { transformFileAsync } from '@babel/core';
 import type { PlaywrightTestConfig } from '@playwright/test';
 import { readdirSync, rmSync, statSync, watch, writeFileSync } from 'node:fs';
 import { ensureFileSync } from 'fs-extra';
-import match from 'micromatch';
-import path from 'path';
+import { makeRe } from 'micromatch';
+import * as path from 'node:path';
```
Smrtnyk, line 6: these were lint errors now that this file is properly linted

src/LayoutManager/LayoutManager.ts
```diff
@@ -43,8 +43,8 @@ export type SerializedLayoutManager = {
 };
 
 export class LayoutManager {
-  private declare _prevLayoutStrategy?: LayoutStrategy;
-  protected declare _subscriptions: Map<FabricObject, VoidFunction[]>;
+  declare private _prevLayoutStrategy?: LayoutStrategy;
```
Smrtnyk, line 46: these are from newer prettier just reordering these

src/LayoutManager/index.ts
```diff
@@ -3,4 +3,4 @@ export * from './LayoutStrategies/ClipPathLayout';
 export * from './LayoutStrategies/FitContentLayout';
 export * from './LayoutStrategies/FixedLayout';
 export * from './LayoutStrategies/LayoutStrategy';
-export * from './types';
+export type * from './types';
```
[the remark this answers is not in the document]
asturur: yes that is good anyway for rollup to know and any builder

asturur (approved):

PR #10498, opened 2025-04-07
ci(): Node 22 hiccups. See if with canvas 3 native dependencies are still necessary for running ci

asturur:
## Description

Remove the call to APT get that is failing on node22


Smrtnyk:
seems like node 22 passes now

PR #10499, opened 2025-04-07
refactor(tests): move canvas tests from qunit to vitest

Smrtnyk:
moves canvas tests from qunit to vitest
this one is pretty big one, and I think most complicated one
other ones will be much or should be much easier, at least according to my cursory look over it


src/canvas/Canvas.spec.ts
```diff
@@ -3,9 +3,138 @@ import { Rect } from '../shapes/Rect';
...
+  classRegistry,
+  getFabricDocument,
+  Group,
+  Path,
+  version,
+} from '../../fabric';
+
+vi.mock('../util/misc/objectEnlive', async (importOriginal) => ({
+  ...(await importOriginal()),
+  loadImage: vi.fn().mockImplementation(() => {
```
Smrtnyk, line 25: one test was loading the image in qunit, and it wasn't working in vitest like this since this is running in jsdom
so I did it like this
we can restore real image load if we eventually move to browser tests in vitest, but for now I think this is good enough

src/canvas/Canvas.spec.ts
Smrtnyk, line 1: I noticed that previous tests here were using `test` instead of `it` but on other places `it` is used so there is a bit of mixup 
I think we should unify this to only use one approach
most of tests are using `it` so I guess it would make sense to use `it` in all files then

asturur: i have no preferences.
I always used 'it' at work, recently they told me that 'test' should be preferred.
I do not even know if they are aliases or have slight differences

asturur (approved):

PR #10511, opened 2025-04-11
refactor(playwright): use fixtures instead of calling setup in every test file

Smrtnyk:
I am currently studying current playwright setup in order to understand it fully and to be able to contribute in migration of visual tests to playwright.
Current playwright setup seems overwhelming to me and with too many acrobatics going on in order to write a simple test.
This PR has a goal to reduce the boiler plate needed to write a playwright test by utilizing a pattern called POM (page object model) https://playwright.dev/docs/pom and playwright fixtures https://playwright.dev/docs/test-fixtures.
This makes playwright setup in fabric more playwright idiomatic and in line of how playwright recommends to structure the tests.
As a result there is not need to call setupApp anymore in each test file, it is done automatically in the page fixture now.
You can also see that this PR reduced lines in playwright by almost 100 lines.

Further ideas to simplify playwright setup even more:
  * remove http-server dependency and startup and leverage playwright routes https://playwright.dev/docs/api/class- 
      route to intercept the requests to assets and pages directly by playwright.
     This makes tests truly isolated and removes latency between real client-server scenario.
     It also makes tests depending on one dependency less.
     Another issue with http-server approach is that it currently logs deprecation message
     ```text
      [WebServer] (node:20974) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated
      [WebServer] (Use `node --trace-deprecation ...` to show where the warning was created)
     ```
     It also doesn't seem maintained anymore.

 * use monocart reporter to simplify coverage collection instead of doing complex acrobatics as it is now
 * implement fully typed playwright test
    currently there are lots of type issues in playwright tests not giving proper intellisense and doesn't give all the type safety typescript offers. It should be easy to create global types so we can have full type safety and intellisense
 * Figure out if babel is needed to build or is there a more straightforward way to run playwright tests in watch mode
  maybe playwright already offers watch mode and we don't need to do all the babel magic at all
 * Figure out if we can directly run playwright files from ts instead of building the js files first


e2e/tests/text/text-editing/index.spec.ts
```diff
@@ -24,7 +19,7 @@ test('textbox typing and resizing', async ({ page }) => {
 
   await page
     .locator('textarea')
-    .type(
+    .pressSequentially(
```
Smrtnyk, line 22: type method is deprecated and pressSequentially is the new recommended one by the docs

asturur:
This will take forever for me to parse.
The playwright setup has its own logic ( that i don't fully remember, has been written by a coworker of mine @luciemars and then modified by Shaman123 ), i added a test for the purpose of migrating the visual tests there by simply adding code and it already does both node rendering and browser rendering.
Let's pause this and let's evaluate when at least the simple json rendering test have been migrated in playwright

Smrtnyk:
> This will take forever for me to parse. The playwright setup has its own logic ( that i don't fully remember, has been written by a coworker of mine @luciemars and then modified by Shaman123 ), i added a test for the purpose of migrating the visual tests there by simply adding code and it already does both node rendering and browser rendering. Let's pause this and let's evaluate when at least the simple json rendering test have been migrated in playwright

no worries, I will adjust to the current setup then
I just wanted to show an alternative I am using at work and that I am familiar with, because i find it with less code and less thinking when writing a test
we can evaluate this later when more important work is done then

Smrtnyk:
I will close this and we can revisit the whole setup later when we have completely migrated from qunit

Smrtnyk:
I will reopen this with resolved merge conflicts as soon as I am done with all qunit to playwright migration

PR #10519, opened 2025-04-13
refactor(tests): move util tests from qunit to vitest

Smrtnyk:
This moves util tests from qunit to vitest.
In qunit it was all in one util spec file, now it is divided to their respective files that match the prod file.
For `loadImage` function I had to use real browser mode in vitest to be able to even test the mechanism but it was easy to do so.
That means if you want your unit test to run in real headless browser instead of jsdom environment you need to suffix the file with `.spec.browser.ts` so it is assigned to the correct runner.
Also added a snapshot of exports for `util` file to make sure things stayed exported as they were before since qunit was testing on the built bundle and now we test on source code.


vitest.config.ts
```diff
@@ -3,13 +3,13 @@ import { defineConfig } from 'vitest/config';
 export default defineConfig({
   resolve: {},
   test: {
-    pool: 'threads',
+    pool: 'vmThreads',
```
Smrtnyk, line 6: figured that this is even faster, at least on my machine
tests run in 4s for me with this PR
there are now around 753 tests in vitest

asturur: i think there was some kind of issue in the first PR that let me choose threads over vmThreads.
I looked into it, i don't really remember.
Let's see what happen

Smrtnyk: I think we need to move this earlier to vitest phase
https://github.com/fabricjs/fabric.js/blob/master/.github/workflows/tests.yml#L134
because browser tests in vitest are using playwright as a provider

Smrtnyk: > i think there was some kind of issue in the first PR that let me choose threads over vmThreads. I looked into it, i don't really remember. Let's see what happen

first PR by default used forks if I am not mistaken
https://vitest.dev/config/#pool
and that was really slow

asturur: I need to review and check a bunch of things.
RUnning tests in browser is something that needs to be discussed.
Today we do that with qunit and testem.
The unit tests at the of the game needs to run once per supported node version and at least once in chrome. All of them. Not just some and some.

So i need to understand this browser thing a little bit better

Smrtnyk: Browser thing in vitest is just another environment option, there is not much to it
There are few of them:
JSdom is node implementation of the simulate browser env, which doesn't work for few of the tests like for setting styles where cssText is mutated
JSDom doesn't handle that, it is missing implementation, so I just made test to run in real browser environment.
All other tests still run in jsdom.
There is also a happydom environmet which is a faster version of jsdom but a bit more barebones.
And there is node environment but there is no concept of document and window in it.

So In this PR I just made two tests to run in browser which are image loading because it doesn't work in jsdom, and setting of styles via mutation of cssText which also doesn't work in jsdom.

Smrtnyk: I think I understand what you mean
you want to run all tests multiple times
once in chromium
once in jsdom

we can do that easily in vitest by leveraging projects
the same way I added the browser configuration we can easily add the new entry into it that will run all tests in browser mode as well
currently all unit tests are ran in vitest only with jsdom, which is the exact same thing when you were using jest

asturur: ok let's finish to convert tests from qunit and let's skip the tests that requires browser for now, or let's see if they can be fixed however they were fixed in qunit, but using a skip + message is fine too.

When qunit and testm are gone, we can write a github action to run witest in browser without having the same one running both serially.

I don't want to mix the two things for now becuase it becomes an overhead for me following everything, is just a matter of organization for me

Smrtnyk: ok sounds good
i will skip the 2 tests for now and will remove the browser settings for vitest
after whole migration is done it will be easy to just add new vitest config for browser mode that can be run separately

asturur:
I m converting shadow and probably i'll do rect too, two bite sized effort i can do when kids sleep

Smrtnyk:
> I m converting shadow and probably i'll do rect too, two bite sized effort i can do when kids sleep

I will do the static canvas next
I noticed it is a chonker so would be good to move it out of the way

vitest.config.ts
Smrtnyk: ok that should be it
I skipped the two tests and added TODO explaining the error

Smrtnyk: in this PR https://github.com/fabricjs/fabric.js/pull/10521
I figured out why images weren't loading in jsdom and I fixed it
I can unskip the image test in this PR after other one is merged

src/util/dom_style.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+      const el = getFabricDocument().createElement('div');
+
+      // TODO: skipping these because jsdom doesn't implement
+      // style invalidation via mutation of the cssText property
+      // setStyle(el, 'color:red');
+      // expect(el.style.color).toBe('red');
+      //
+      // setStyle(el, 'color:blue;border-radius:3px');
+      // expect(el.style.color).toBe('blue');
+      // expect(el.style.borderRadius).toBe('3px');
```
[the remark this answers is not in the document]
Smrtnyk: ok thats great to hear then

asturur (approved):

PR #10524, opened 2025-04-14
chore(): BREAKING: Remove minore deprecations

asturur:
### Events

The data duplication between pointer/absolutePointer and viewportPoint/scenePoint has been removed.
Now you find only the latter in the events

### Canvas

- `getCenter` method has been removed, use only `getCenterPoint`
- The internal cache variables `_pointer` and `_absolutePointer` have been renamed to `_viewportPoint` and `_scenePoint`
- `preserveObjectStacking` is now default to true, if you want objects to break the stack when selected, set it to false
- `getPointer` deprecated method has been removed, is now a protected `_getPointerImpl` method.
- `setWidth` and `setHeight` are gone, use only `setDimensions`

#### Text
- unused method `_getSVGLineTopOffset` has been deleted

#### Itext
- `protected _exitEditing` is merged with `exitEditingImpl` public

#### Image
- `CSS_CANVAS` static property from image is gone also with the `classList.Add` requirement
### Utils
- `rotatePoint` is gone, use point.rotate instead
- `request` is gone. We use fetch, node 18 supports fetch. There is no reason to use the old request object and status
- `setStyle` is gone


Smrtnyk:
what about also removing `request` from `util` namespace?
In case someone is using it they can move to fetch or their own request implementation

asturur:
> what about also removing `request` from `util` namespace? In case someone is using it they can move to fetch or their own request implementation

I left it out because you said you wanted to do it

asturur:
I just deleted stuff, how did i gain so much code
image

asturur:
> I just deleted stuff, how did i gain so much code image

I must have inverted the new with the old

asturur:
to be honest also the util namespace in general is kind of wrong.
It makes hard to codesplit nicely when you import a single util.

Smrtnyk:
> to be honest also the util namespace in general is kind of wrong. It makes hard to codesplit nicely when you import a single util.

yep
maybe we need a new entry in package.json exports like fabric/util?

asturur:
I think so, that was already discussed in the past, but i m not going to touch it now.

src/util/index.ts
Smrtnyk, line 1: seems like request is back on the utils?
but got removed in prior commit https://github.com/fabricjs/fabric.js/pull/10524/commits/40236e4f37fec31cc4f7949d29e24ff514dcbd7f

asturur: Yes because old tests are using it and i don't want this pr to address that too.

src/benchmarks/cloneDeep.mjs
```diff
@@ -4,5 +4,4 @@
  * @param object
  * @returns
  */
-export const cloneDeep = <T extends object>(object: T): T =>
-  JSON.parse(JSON.stringify(object));
+export const cloneDeep = (object) => JSON.parse(JSON.stringify(object));
```
Smrtnyk, line 7: is this implementation needed?
there is lodash in the project which should offer this same functionality

asturur: Those benchmark are just here for historic memory of why we removed some code with other.
When we found deep performance difference we changed and we left a simple script to document the decision.
That is why i keep a copy of it there. Just in case someone come and wants to use a deep clone logic again instead of a specialized cloning for the data structure

PR #10528, opened 2025-04-14
refactor(tests): move brushes tests from qunit to vitest

Smrtnyk:
This moves brushes tests from qunit to vitest


src/brushes/PencilBrush.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+        brush.onMouseMove(pointer3, { e });
+        brush.onMouseUp({ e });
+
+        expect(
+          fireBeforePathCreatedEvent,
+          'before:path:created event is fired',
+        ).toBe(true);
+        expect(firePathCreatedEvent, 'path:created event is fired').toBe(true);
+        expect(added, 'a path is added').toBeInstanceOf(Path);
+        expect(added!.path.length, 'path has 6 steps').toBe(6);
```
Smrtnyk, line 270: @asturur can you have a look at this one?
it is failing because in jsdom it is yielding 4 instead of 6
seems like there is a difference in browser and jsdom for paths
I am not that familiar with paths in canvas so maybe you can figure out what the issue here is

asturur: I looked into this 30 minutes.
I see the code in decimatePath is killing the point before the last one, skipping to iterate over it.
I added a TODO but for now it puzzles me that it was working with qunit, it should have failed there too.

asturur (approved):

PR #10536, opened 2025-04-15
refactor(tests): move canvas dispose tests from qunit to vitest

Smrtnyk:
this moves canvas dispose tests from qunit to vitest


src/canvas/Canvas-dispose.spec.ts
```diff
@@ -0,0 +1,538 @@
...
+                  expect(
+                    runningAnimations,
+                    'should cancel the animation',
+                  ).toEqual([]);
+                  resolve();
+                });
+                expect(canvas.disposed, 'should flag `disposed`').toBeTruthy();
+              }
+              called++;
+              // TODO: check typings, because this runs for both static and normal canvas but it is only typed on normal canvas
```
Smrtnyk, line 270: @asturur left this comment here in case you are curious

asturur: FabricJS typings have been hard.
The fact that every class of object carries with it the types of the export and the props it expects in the constructor makes everything complicated.

Collection.ts is a piece of code shared between group and staticCanvas so it is a mixin, but TS doesn't really support mixins natively.

The staticCanvas works with fabricObejct while Canvas works with InteractiveObject.

The migration from plain JS to TS was done how it could have been done and bigger rewrites are necessary to make typing smooth.

Our issues are mainly docs and time dedicated to it. If we ever manage to write good docs with explanations and overwies then those refactors are less painful because you know what kind of documentation you have to keep updated as well.

For now it is what is is, autocompletion of classes,properties,custom properties and events works and that is a big upside, the rest will come a piece at time.

asturur (approved):

PR #10542, opened 2025-04-16
refactor(tests): move parser tests from qunit to vitest

Smrtnyk:
this moves parser tests from qunit to vitest


src/parser/doesSomeParentMatch.ts
```diff
@@ -1,6 +1,6 @@
 import { selectorMatches } from './selectorMatches';
 
-export function doesSomeParentMatch(element: HTMLElement, selectors: string[]) {
+export function doesSomeParentMatch(element: HTMLElement | SVGElement, selectors: string[]) {
```
Smrtnyk, line 3: I extended this type on some places because parser tests are passing SVGElement to those functions
and if function accepts it and can deal with it then types should reflect that
this way we have no type issues in the tests and also tests reflect that back to the user

asturur: this was written for svgElements, i m not sure why HTMLElement ended up there. But anyway this is correct. If some types ends up being strange they are probably wrong, like shadow that had options not optional but mandatory and you had to pass an empty object to initialize a standard shadow.

Smrtnyk:
fixed prettier

asturur:
ty i wasn't on my laptop i couldn't fix

asturur (approved):

Smrtnyk:
> ty i wasn't on my laptop i couldn't fix

I think I should soon have a look at precommit hooks
maybe https://github.com/typicode/husky
it happens often that I forget about prettier

PR #10543, opened 2025-04-17
refactor(tests): remove jest-snapshot package

Smrtnyk:
This was supposed to be done as part of migration of jest to vitest PR but I didn't want to overcomplicate that PR back then.
 This removes dependency to jest package and we now rely solely on vitest.
 This also shrinks package-lock file by 1k lines.
 There is no logic change here, it is just moving the toMatchSnapshot override to vitest chai assertion because that is where vitest implements it, and it is not compatible with expect.extend like other matchers.
 All tests work just like they worked before, no changes in snapshot files.

 This is also absolutely necessary for the future when we start running tests also in browser environment, where jest-snapshot cannot work since it is designed only to run in node(jsdom) environment.

### DIFF
--- a/vitest.extend.ts
+++ b/vitest.extend.ts
@@ -1,6 +1,5 @@
-import { expect } from 'vitest';
+import { expect, chai } from 'vitest';
 
-import { toMatchSnapshot } from 'jest-snapshot';
 import { cloneDeepWith } from 'lodash';
 import { FabricObject } from './src/shapes/Object/Object';
 import type { TMat2D } from './src/typedefs';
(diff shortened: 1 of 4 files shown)


vitest.extend.ts
```diff
@@ -25,13 +24,91 @@ function sanitizeSVG(value: string) {
...
+    } else {
+      snap = received;
+    }
+    delete snap.version;
+
+    const value = cloneDeepWith(snap, (v, k, obj, stack) => {
+      const c = customiser?.(v, k, obj, stack);
+      if (c !== undefined) return c;
+      if (k === 'width') return Math.round(v as number);
+    });
```
asturur, line 66: I Need to revisit why we need this.
Is true that fabricJS can output floats that could behave differently between node and browser for strange reasons, or just between platforms.
But if we need to round to have stable snapshot we need to use 'toFixed' not a generic round, and we need to use it on all numeric properties.

For now is fine as it is.

Smrtnyk: Well this is just a move to pure vitest to not rely on 3rd party jest package.
I did not change logic at all and this is from jest times.
If you can improve or simplify this it can be a different PR.
I did this to drop the 3rd party package because we use vitest and not jest anymore and also to not block future tests in browser more.
A big bonus is that we install less packages now, lock file is smaller by thousand lines

Smrtnyk:
Seems like firefox failed
I guess just flaky?

asturur (approved):

asturur:
i m trying to fix the action for the changelog before merging new prs

PR #10545, opened 2025-04-17
refactor(tests): move control handlers tests from qunit to vitest

Smrtnyk:
this moves controls handlers tests from qunit to vitest


src/controls/Control-handlers.spec.ts
```diff
@@ -0,0 +1,285 @@
...
+  afterEach(() => {
+    canvas.off();
+    canvas.clear();
+  });
+
+  function prepareTransform(
+    target: FabricObject,
+    corner: string,
+  ): Transform & { signX: number; signY: number } {
+    // TODO: this method is deprecated and jsdoc says that it should have disappeared before fabric 4?
```
Smrtnyk, line 34: @asturur left a comment here in case you want to check it out

asturur: Yeah that is a so dumb thing to solve.
When dragging a control, guessing where is the anchor point.
To be honest that should be explicit of the control that execute the action.
So if i put a control in a position x,y i should also say, the anchor point is w,z.

I plan to propose new controls for shapes ( same api, just different controls) so maybe that is the good occasion to also fix this.

asturur (approved):

PR #10546, opened 2025-04-17
refactor(tests): move image tests from qunit to vitest

Smrtnyk:
oh boy this one was intense
moves image tests from qunit to vitest


vitest.config.ts
```diff
@@ -17,6 +17,7 @@ export default defineConfig({
   test: {
     pool: 'vmThreads',
     clearMocks: true,
+    mockReset: true,
```
Smrtnyk, line 20: I noticed that spies were leaking between tests which made it very hard to test something in image suite
this made some other tests to breka in Itext which I adapted

asturur: yeah this should be the default honestly.

src/shapes/Image.spec.ts
```diff
@@ -72,10 +133,904 @@ describe('FabricImage', () => {
...
+  });
 });
+
+export function newImg(src = IMG_SRC): HTMLImageElement {
+  const img = getFabricDocument().createElement('img');
+  img.src = src;
+  return img;
+}
+
+export async function createImage(
```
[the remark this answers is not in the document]
asturur: Consider some tests are super super old and written with different expertise compared to today standards and capabilities.

asturur (approved):

PR #10547, opened 2025-04-17
refactor(tests): move image filters tests from qunit to vitest

Smrtnyk:
this moves image filters tests from qunit to vitest


src/shapes/Image-filters.spec.ts
```diff
@@ -0,0 +1,1906 @@
...
+      const object = filter.toObject();
+
+      expect(
+        JSON.stringify(object),
+        'serialized object should match expected value',
+      ).toBe('{"type":"Convolute","opaque":1,"matrix":[0,0,0,0,1,0,0,0,0]}');
+    });
+
+    it('toJSON', () => {
+      // @ts-expect-error -- TODO: check this -- it is typed as boolean but we are passing number here
```
Smrtnyk, line 983: @asturur left a comment here in case you want to check it out
I assume boolean should be passed here instead of the number
or the type is wrong

there is also this same todo on 2 more places

asturur: Yes the test was wrong, it makes sense as a boolean and the code behave as boolean.

asturur (approved):

PR #10549, opened 2025-04-18
ci(): Create an action that creates and uploads a JSON changelog for later updates

asturur:
## Description

This action should be safe and usable by forks
It creates a json string with the necessary information so that a master workflow can reliably update the branch with the changelog.


asturur:
i hate this stuff

asturur:
This part needs to be merged so that the workflow can be run

PR #10551, opened 2025-04-18
refactor(tests): migrate polygon tests from qunit to vitest

Smrtnyk:


src/shapes/Polygon.spec.ts
```diff
@@ -0,0 +1,568 @@
...
+  it('polygon with exactBoundingBox false', () => {
+    const polygon = new Polygon(
+      [
+        { x: 10, y: 10 },
+        { x: 20, y: 10 },
+        { x: 20, y: 100 },
+      ],
+      {
+        // @ts-expect-error -- TODO: are types wrong for Polygon? seems like it doesn't accept exactBoundingBox property
+        exactBoundingBox: false,
```
Smrtnyk, line 125: @asturur seems like typescript doesn't accept exactBoundingBox as a param here
this will also reflect for users of the fabric
either this prop is redundant or types should be improved to include this prop

Smrtnyk:
hm seems like this changelog action to auto add the title to the changelog didn't work?

asturur (approved):

PR #10552, opened 2025-04-18
refactor(tests): move path tests from qunit to vitest

Smrtnyk:


src/shapes/Path.spec.ts
```diff
@@ -0,0 +1,659 @@
...
+    const elPath = getFabricDocument().createElementNS(namespace, 'path');
+
+    elPath.setAttributeNS(namespace, 'd', 'M 100 100 L 300 100 L 200 300 z');
+    elPath.setAttributeNS(namespace, 'transform', 'scale(.2)');
+
+    const path = await Path.fromElement(elPath);
+
+    expect(path, 'should be instance of Path').toBeInstanceOf(Path);
+
+    // @ts-expect-error -- TODO: check why transformMatrix is not set on the Path as potential type?
```
Smrtnyk, line 448: @asturur should `transformMatrix` be set as nullable property of the path?

asturur: Yeah the type should be fixed.
fromElement returns the FabricObject with attached that property that then we remove one line of code after.
Each 'FromElement' should call 'removeTransformMatrix' but i think to avoid duplicated calls it was done in a single place in the parsing logic.

Smrtnyk:
I have canvas events in my local branch
In case you will want to migrate more to not do double work

Smrtnyk:
I would also take webgl test if you dont mind

asturur:
I fixed the types for FromElement that indeed could return the transformMatrix.

asturur:
> I would also take webgl test if you dont mind

Take all you want i have a stack of things to do, i do not need to do more tests. I did a bunch for sharing more than other

asturur (approved):

PR #10554, opened 2025-04-19
refactor(tests): move object interactivity tests from qunit to vitest

Smrtnyk:


src/controls/Control.ts
```diff
@@ -282,8 +282,8 @@ export class Control {
    */
   setVisibility(
     visibility: boolean,
-    name: string,
-    fabricObject: InteractiveFabricObject,
+    name?: string,
```
Smrtnyk, line 285: not sure what these are about
they seem unused in code
also tests didn't provide them, so I just made them optional to fix type issues in tests

asturur: This was there to set an api. If you would override the Control class to be smarter, you could write different logic depending on what control was doing on which object.
I don't think anyone ever used it, but also there is no reason to remove it right now.

asturur (approved):

PR #10557, opened 2025-04-20
refactor(tests): Migrate Env, ClassRegistry, Rect to vitest

asturur:
## Description

Migrate more tests to vitests


src/env/env.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { describe, expect, it, afterEach } from 'vitest';
+
+describe('env', () => {
+  afterEach(() => {
+    delete global.window;
```
Smrtnyk, line 5: i think we should use globalThis rather than global because it has interop between browser and node
Probably doesnt matter for this test because I assume this test will not run in the browser in the future?
But just to get used to globalThis over global

asturur: yes i think yes, this global interaction i also think it broke several things

asturur:
I m not sure the env test with imports and requires make sense

Smrtnyk:
> I m not sure the env test with imports and requires make sense

Maybe thinga are off because it goes through vite-node resolution and not native node resolution

Smrtnyk:
Maybe this test should be done on compiled dist from e2e?

asturur:
> Maybe this test should be done on compiled dist from e2e?

Or maybe there isn't much to test. Probably a single one off kind of test is fine but is not a priority.

Smrtnyk:
> > Maybe this test should be done on compiled dist from e2e?
> 
> Or maybe there isn't much to test. Probably a single one off kind of test is fine but is not a priority.

That is why i left it for the end, didnt want to fry up my brain straight at the beginning

asturur:
> > > Maybe this test should be done on compiled dist from e2e?
> > 
> > 
> > Or maybe there isn't much to test. Probably a single one off kind of test is fine but is not a priority.
> 
> That is why i left it for the end, didnt want to fry up my brain straight at the beginning

I didn't expect it to be so annoying. I was like, look a short test a quick win.

Smrtnyk:
> > > > Maybe this test should be done on compiled dist from e2e?
> > > 
> > > 
> > > Or maybe there isn't much to test. Probably a single one off kind of test is fine but is not a priority.
> > 
> > 
> > That is why i left it for the end, didnt want to fry up my brain straight at the beginning
> 
> I didn't expect it to be so annoying. I was like, look a short test a quick win.

Haha yeah, thought the same
But then I realised all imports resolution now goes through vitest and that this needs to be tested differently

PR #10562, opened 2025-04-20
refactor(tests): migrate object tests to vitest

Smrtnyk:


src/shapes/Object/Object.ts
```diff
@@ -734,7 +734,7 @@ export class FabricObject<
    */
   hasStroke() {
     return (
-      this.stroke && this.stroke !== 'transparent' && this.strokeWidth !== 0
+      !!this.stroke && this.stroke !== 'transparent' && this.strokeWidth !== 0
```
Smrtnyk, line 737: without this it would return empty string
although "falsy" it is not a boolean
qunit was testing this loosely treating empty string as false
but vitest is bit more strict and precise here

asturur:
ok update changelog seem to work finally

src/shapes/Object/Object.ts
asturur: the intention of this is to be a boolean, can you add it to the typo of hasStroke explicitly?

Smrtnyk: You mean to add :boolean as return type of this method?

Smrtnyk: done

asturur (approved):

PR #10563, opened 2025-04-20
refactor(tests): move canvas events tests from qunit to vitest

Smrtnyk:
this is the last unit test in qunit that was left
there is a todo in the test in one test which I will comment inline


src/canvas/Canvas-events.spec.ts
```diff
@@ -0,0 +1,1908 @@
...
+      clientX: 140,
+      clientY: 90,
+      which: 1,
+      target: canvas.upperCanvasEl,
+      shiftKey: true,
+    };
+
+    canvas.on('mouse:down', function (options) {
+      // TODO: fix this, for some reason first target on mouse down is active selection and then second target is the green rectangle
+      if (options.target instanceof ActiveSelection) {
```
Smrtnyk, line 309: @asturur I don't know how this test worked in qunit but in both jsdom and in the browser the first event that comes here is Active selection and then the green box 
I guess you will have to have a look here since you have far more knowledge than I do when it comes to fabric source code

for now I workarounded this with a return if it is now a green box but clearly this is not a solution

src/canvas/Canvas-events.spec.ts
```diff
@@ -0,0 +1,1908 @@
...
+    expect(getTargets(), 'dispose should clear instances').toEqual([]);
+    expect(getTarget(), 'should unregister b').toBeFalsy();
+  });
+
+  it.todo('mousemove: subTargetCheck: setCursorFromEvent considers subTargets');
+  it.todo(
+    "mousemove: subTargetCheck: setCursorFromEvent considers subTargets in reverse order, so the top-most subTarget's .hoverCursor takes precedence",
+  );
+
+  // TODO: ported from qunit, check if is still valid for vitest
```
Smrtnyk, line 1798: I poted this thing as well but I don't know if it is relevant

src/canvas/Canvas-events.spec.ts
```diff
@@ -0,0 +1,1908 @@
...
+
+    manager.register(b);
+    expect(getTarget(), 'should register b').toBe(b);
+
+    await testCanvas.dispose();
+    expect(getTargets(), 'dispose should clear instances').toEqual([]);
+    expect(getTarget(), 'should unregister b').toBeFalsy();
+  });
+
+  it.todo('mousemove: subTargetCheck: setCursorFromEvent considers subTargets');
```
Smrtnyk, line 1793: I also ported these two as test todos so it doesn't get lost in case it is something that should be actually done

Smrtnyk:
this concludes the porting of tests from qunit to vitest
After this is in I will open a PR where we also run same tests in chromium using vitest browser mode
it required some tweaks so tests still work on both chromium and jsdom but nothing big

after that is done I will work on making tests run on firefox and webkit as well

asturur:
i ll try to clean up yes

asturur (approved):

src/canvas/Canvas-events.spec.ts
```diff
@@ -0,0 +1,1908 @@
...
+  //   expect(!!canvas.actionIsDisabled('mb', target, e), 'mb action is disabled lockSkewingX').toBe(true);
+  //   expect(!!canvas.actionIsDisabled('ml', target, e), 'ml action is not disabled lockSkewingX').toBe(false);
+  //   expect(!!canvas.actionIsDisabled('mr', target, e), 'mr action is not disabled lockSkewingX').toBe(false);
+  //   expect(!!canvas.actionIsDisabled('tl', target, e), 'tl action is not disabled lockSkewingX').toBe(false);
+  //   expect(!!canvas.actionIsDisabled('tr', target, e), 'tr action is not disabled lockSkewingX').toBe(false);
+  //   expect(!!canvas.actionIsDisabled('bl', target, e), 'bl action is not disabled lockSkewingX').toBe(false);
+  //   expect(!!canvas.actionIsDisabled('br', target, e), 'br action is not disabled lockSkewingX').toBe(false);
+  //   expect(!!canvas.actionIsDisabled('mtr', target, e), 'mtr action is not disabled lockSkewingX').toBe(false);
+  // });
+});
```
asturur, line 1908: right there was this disabled action logic that was never ported to the action themselves i guess. I don't remember very well, controls could be refactored twice and still be sub optimal.
trying to have both a configuration approach and a code approach to default actions and custom actions has proved to be both confusing and hard.

asturur:
i may have merged too quick.  i ll check and fix in case

PR #10568, opened 2025-04-21
feat(tests): add chromium browser mode for unit tests

Smrtnyk:
How it works

* "test:vitest" still runs tests in jsdom
* "test:vitest:coverage" still runs tests in jsdom

new scripts added
* "test:vitest:chromium" runs same set of tests but in chromium using vitest browser environment

There is no magic in here
just some small adaptations needed for tests to run in both environments now
I will comment those inline


.gitignore
```diff
@@ -25,3 +25,4 @@ e2e/test-report/
 e2e/test-results/
 **/.cache/
 docs/
+__screenshots__
```
Smrtnyk, line 28: added this to gitignore because now if test fails there is a screenshot folder so you can see how page looked like at the time of failure
for example, if your test had canvas appended to the document, you can see how canvas looked like when it failed the test

src/canvas/__tests__/__snapshots__/eventData.test.ts.snap
```diff
@@ -145,17 +145,17 @@ exports[`Canvas event data > HTML event "dragend" should fire a corresponding ca
   [
     "after:render",
     {
-      "ctx": CanvasRenderingContext2D {
+      "ctx": {
```
Smrtnyk, line 148: this is the biggest change
snapshotting ctx is not same in browser and jsdom
in jsdom properties are on instance
in browser they are on prototype

so I had to write a custom serializer for the canvas rendering context so it is stable cross env between jsdom and browser

otherwise in browser it is snapshotted as `{}` which fails the snapshot

src/parser/parser.spec.ts
```diff
@@ -675,7 +675,8 @@ describe('fabric.Parser', () => {
   it('getCssRule', () => {
     expect(getCSSRules).toBeDefined();
     const rules: Record<PropertyKey, any> = {};
-    const doc = getFabricDocument();
+    // NOTE: We need to use a new fresh document here because vitest in browser mode already adds some stylesheets which pollutes the test
+    const doc = getFabricDocument().implementation.createHTMLDocument('');
```
Smrtnyk, line 679: in vitest browser mode there are already some stylesheets imported to testing frame by vitest, so I had to get clean doc implementation here
works fine in both jsdom and chromium

src/util/internals/getRandomInt.test.ts
```diff
@@ -1,39 +1,39 @@
 import { getRandomInt } from './getRandomInt';
 import { describe, expect, vi, it, beforeEach, afterEach } from 'vitest';
 
-const originalMathRandom = global.Math.random;
+const originalMathRandom = globalThis.Math.random;
 
 describe('getRandomInt', () => {
   beforeEach(() => {
-    global.Math.random = vi.fn(() => 0.1);
+    globalThis.Math.random = vi.fn(() => 0.1);
```
Smrtnyk, line 8: switched to globalThis for interop between jsdom and browser

test/snapshot-serializers/canvas-rendering-context.ts
```diff
@@ -0,0 +1,58 @@
...
+  );
+
+/**
+ * Snapshot serializer that prints a stable, cross‑runtime representation
+ * of CanvasRenderingContext2D.
+ */
+export default {
+  test: isCanvasCtx,
+
+  serialize(val: any, config: Config, indentation: string, depth: number, refs: Refs, printer: Printer) {
```
Smrtnyk, line 22: this is the serializer I mentioned earlier to have stable snapshot cross env for canvas rendering context

vitest.env.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module '*.gif' {
```
Smrtnyk, line 1: this is needed for imports of images to be recognized when we import them for browser mode tests

vitest.extend.ts
```diff
@@ -49,17 +60,15 @@ chai.util.addMethod(
     }: ObjectOptions = {},
     hint?: string,
   ) {
-    let snap: Record<string, unknown>;
     const received = chai.util.flag(this, 'object');
+    let snap: Record<string, unknown> = received;
 
-    if (received instanceof FabricObject) {
```
Smrtnyk, line 55: this was something that I don't know how it even worked
we can't import fabric objects as values in setup file
because it can break mocks in tests
this failed in browser mode because in one test we mocked uid module, but fabric object extends soemthign that imports uid and then mock of the uid module failed in the test file

so I opted to not use the import as value but to just use duck typed check which worked fine

Smrtnyk:
if you end up wanting this @asturur  you will need to write new workflow that executes "test:vitest:chromium" script
but will need to run `npx playwright install` before to fetch the chromium

src/canvas/__tests__/__snapshots__/eventData.test.ts.snap
asturur: Maybe the custom serializer can return just the string with the type? we don't need to be accurate here. after:render pass the ctx where the rendering event happened. We don't need to know that ctx has determinated properties

src/canvas/StaticCanvas.spec.ts
```diff
@@ -2183,10 +2185,11 @@ describe('StaticCanvas', () => {
     expect(scaling).toBe(1);
   });
 
-  it('provides Node.js stream methods when in Node environment', () => {
-    if (!('jsdom' in globalThis)) {
+  it('provides Node.js stream methods when in Node environment', async () => {
+    if (!isJSDOM()) {
       expect(true).toBeTruthy();
     } else {
+      const { Canvas: FabricNodeCanvas } = await import('../../index.node');
```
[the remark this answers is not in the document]
Smrtnyk: > i m borderline with this between let's delete the test and ok is one exception. I ll read the rest of the PR and see how many times we need to use this

it was the only one that needed this so I guess we can leave it for now

src/canvas/__tests__/__snapshots__/eventData.test.ts.snap
Smrtnyk: yeah I guess that would be the easiest
what exactly would you want to output here?
what is relevant for the test?
so I can adjust the serializer

src/parser/parser.spec.ts
asturur: This test i think is wrong.
The parser and getCSSRules is used for SVGdocuments, i think here to test the function we assumed that one document is like anohter.
getFabricDocument return the document that holds the javascript from fabricjs.
We need a generic document, if we have a generic api that works in both cases like `new Document()` or `new XLMDocument()` (since is svg) we should prefer that.

test/snapshot-serializers/canvas-rendering-context.ts
asturur: For me this can return the type only as a string. 'CanvasRenderingContext2d` this because in future this block of code may become a sleeping issue for someone else approaching unit test. We don't need to verify the CTX below the surface of being a CTX, so this may be unnecessary. Sorry if you spent a bunch of time on this

Smrtnyk: > For me this can return the type only as a string. 'CanvasRenderingContext2d` this because in future this block of code may become a sleeping issue for someone else approaching unit test. We don't need to verify the CTX below the surface of being a CTX, so this may be unnecessary. Sorry if you spent a bunch of time on this

don't worry
I also learned something new along the way
I will adjust the serializer to return the type only, it makes sense to me as well

vitest.extend.ts
Smrtnyk: adjusted as per other comment

vitest.config.ts
```diff
@@ -11,27 +11,50 @@ if (!fixturesUrl.endsWith('/')) {
...
+        extends: true,
+        test: {
+          environment: 'jsdom',
+          environmentOptions: {
+            jsdom: {
+              resources: 'usable',
+              url: fixturesUrl,
+            },
+          },
+          name: 'unit-jsdom',
```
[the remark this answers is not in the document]
Smrtnyk: sure, makes sense

vitest.extend.ts
```diff
@@ -27,6 +23,21 @@ function sanitizeSVG(value: string) {
   return replaceLinks(value).replace(SVG_RE, 'SVGID');
 }
 
+function looksLikeFabricObject(val: unknown): val is FabricObject {
+  return (
+    !!val &&
+    typeof val === 'object' &&
+    // all FabricObjects implement these three members
+    typeof (val as any).toObject === 'function' &&
+    typeof (val as any).getObjectScaling === 'function' &&
```
asturur, line 32: let's put render here instead of getObjectScaling.

src/parser/parser.spec.ts
Smrtnyk: I guess this is then something for a separate PR as a todo?

asturur:
For sure this is the natural extension of what we had before.
And i think is good to have, or at least to try if it maintainable.

In the long run this may seems unnecessary because if unit tests are written by the book, you mock most o the things and test only your own code.

Is a bit different for visual tests in which you really want to test the output of the drawing functions.

As a maintainer doing this extra loops of tests keep me informed of different rounding methods between browser and node, difference in api that may happen between node-canvas and browser, in general if not extreme added maintainance i prefer to keep doing it.

I left a bunch of notes that do not preclude a merge, please see what you think about them

src/parser/parser.spec.ts
asturur: What i meant is, without having to create it from the fabric document itself.
Is there any global we can invoke that give us a new document, empty?

asturur:
An additional improvement that can be done another time is that we export from fixtures functions that return the URL instead of returning the different url each time we need the image in a different file.
This will be more evident with visual tests.

Smrtnyk:
> For sure this is the natural extension of what we had before. And i think is good to have, or at least to try if it maintainable.
> 
> In the long run this may seems unnecessary because if unit tests are written by the book, you mock most o the things and test only your own code.
> 
> Is a bit different for visual tests in which you really want to test the output of the drawing functions.
> 
> As a maintainer doing this extra loops of tests keep me informed of different rounding methods between browser and node, difference in api that may happen between node-canvas and browser, in general if not extreme added maintainance i prefer to keep doing it.
> 
> I left a bunch of notes that do not preclude a merge, please see what you think about them

I agree with you with the concept of unit tests.

I noticed when I ran these same tests in webkit I already saw some numeric misalignement in few snapshots due to some precision, which showed difference between jsdom/chromium/webkit, not a lot but few tests did appear.
Also in firefox kerning was 13 in one snapshot where in jsdom, webkit and chromium is 12.

This does add maintenance burden a bit, but on other hand I see chromium more valuable source of testing reliability than jsdom. Chromium is the most used browser, and I assume most users of fabric are using it in browser and not in node env.

My point of this PR was done to demonstrate that vitest has excellent capabilities to run tests in browser as well and very fast and in an easy way.

If you decide to only stick with jsdom we can keep this out for now, and revive in the future if you decide to actually incorporate it.

asturur:
We do want to run tests in browsers. this is good

src/parser/parser.spec.ts
asturur: yes i just cheked `document.implementation` is just a normal browser thing, i just never used it.

asturur (approved):

Smrtnyk:
pls also let me know if I should work on webkit and firefox tests next?
they have some discrepancies so will need some minor adjustments as well to have it all stable across all browsers

PR #10569, opened 2025-04-21
Chore() more qunit post cleanup

Smrtnyk:
removed two npm scripts because they seem obsolete now

Is cli tool for test still used @asturur ?
It still shows 2 options for visual and unit tests in qunit

I also removed the bithoundrc file since that service is dead for years now


asturur (approved):

Smrtnyk:
> i used the cli tool for unit test a lot, and i would still use it for visual till they are there.

I started looking a bit into visual tests in how they work 
lets see how long it will take me to completely understand the whole concept so I am ready to start porting them

PR #10572, opened 2025-04-22
refactor(): remove @ts-nocheck in straighten.ts

Smrtnyk:
I am looking a bit into current types state
found this one that has @ts-nocheck
I cannot see if this file is used anywhere? 
What for is it used currently?
I see there is a unit test that is skipped for this

I fixed the type issues after removal of the @ts-nocheck but I have dilemma
This mutation of FabricObject prototype feels odd
would a better approach be something like composition
here is an example of what i think
```typescript
export class FabricObjectStraightener {
  static FX_DURATION = 500;

  constructor(public target: FabricObject) {}

  private _getAngleValueForStraighten(): number {
    const angle = this.target.angle % 360;
    // ... logic ...
    return Math.round(angle / 90) * 90;
  }

  straighten(): void {
    this.target.rotate(this._getAngleValueForStraighten());
  }

  fxStraighten(callbacks: { /*...*/ } = {}): () => void {
    const endValue = this._getAngleValueForStraighten();
    // ... animate logic using this.target ...
    return animate({
      target: this.target,
      startValue: this.target.angle,
      endValue: endValue,
      duration: FabricObjectStraightener.FX_DURATION, 
      // ... onChange, onComplete using this.target ...
    });
  }
}

// Usage:
const myObject: FabricObject = /* ... */;
const straightener = new FabricObjectStraightener(myObject);
straightener.straighten();
straightener.fxStraighten({ onComplete: () => console.log('Done!') });
```
That is just an idea, I am yet not that familiar with fabric source code itself


asturur:
The object straightener was something we dropped during migration to v6.
It didn't feel like it made sense. It was like animation code to rotate an object to align with 90 degree steps.
It is one of kind we don't really use or provide, anyone can do it with the animation utility or the animate method.

It was something super old that was inside fabric in the start, but then it didn't develop in anything useful

asturur:
All the files in 'parkinglot' are like that.

Smrtnyk:
> All the files in 'parkinglot' are like that.

should we remove them then since it is just a dead code?

asturur:
i think is worth merging this, because is better than before even if we don't use it.

asturur (approved):

PR #10587, opened 2025-05-08
refactor(): BREAKING Refactor find target

asturur:
## Description

The scope of this PR is to change how findTarget behave.
Today findTarget returns you a target that includes all the user experience logic of fabricJS and just tells you 'the target is this one' including special logic around what is already selected, if we hit a corner, if there is an active selection or a sub selection in an interactive group.

The scope of the refactor is to have this logic clearly spelled in the code.

findTarget is going to return targeting information:
- the exact object under the mouse cursor
- the inner target if there is one
- if the target is also on top of the current activeObject
- if any controls are hit in the process

This information is going to be handled by the mouse down or up handlers that are going to decide what to do.

Also canvas.targets is going away and targets is not going to be global anymore.
That was just a bad choice.

This is the only additional breaking change for fabric 7 and i ll work on it every time i can

### UPDATE

4 months later this PR is not as transformative and good as i wanted, is also incredibly boring to work on reason for which it went stale.
The code works and seems an improvement ( to me ) to what we had before.
Additional improvement to the multi selection code and and transform target are probably possible.
I m merging as is, otherwise i won't move forward with other tasks.


src/canvas/SelectableCanvas.ts
```diff
@@ -867,7 +867,7 @@ export class SelectableCanvas<EventSpec extends CanvasEvents = CanvasEvents>
       if (this._checkTarget(target, pointer)) {
         if (subTargets && isCollection(target) && target.subTargetCheck) {
           const { target: subTarget } = this._searchPossibleTargets(
-            target._objects as FabricObject[],
+            target._objects,
```
Smrtnyk, line 912: if you are accessing this as a public member outside of the class, should this rather be renamed to `objects` then? and declared public?
it could also a be a getter that provides a copy of the array to prevent mutation of the original array
this is just an idea, because it looks fishy to access something that `looks like a private` in a public manner

asturur: A developer should use `getObjects` that does exactly that.
Internally we avoid copying the array if we know we are going only to read it.
That is the reason why i reference it directly.

src/EventTypeDefs.ts
```diff
@@ -261,16 +261,6 @@ type TPointerEvents<Prefix extends string> = Record<
...
-      /**
-       * The targets at the moment of mouseup that could be different from the
-       * target at the moment of mouse down in case of a drag action for example
-       */
-      currentTarget?: FabricObject;
-      /**
-       * The subtargets at the moment of mouseup that could be different from the
-       * target at the moment of mouse down in case of a drag action for example
-       */
-      currentSubTargets: FabricObject[];
```
asturur, line 273: removed.
The target and subTargets will always return the current value in the event.
The old shortcut of transform.target is still available in transform.target.
This needs to be documented in the breaking changes

asturur:
Ok at this point the code is functionally equivalent to the previous one. ( minus one visual test to fix )
It still doesn't achieve the simplification i was aiming for, but at least findTarget is less cryptic on what it does.
Find dragTarget should be removed and handleActiveSelection should see some changes too.
I hoped for better code, but at least this one let me add informations to findTarget to fine tune the selection process ( and fix bugs ).

asturur:
image

:(

Ok today i should try to do this conflict and add some test

PR #10590, opened 2025-05-09
refactor(tests): move text path tests from qunit to playwright

Smrtnyk:


asturur:
> hm, seems like one unit test on firefox fails I wonder how it worked when I opened a firefox unit tests initial pull request

It does not work on the ci, and i see is the webgl test.
It could be an issue with webgl detection on the ci or something else, is not a big deal, i commented it when i enabled firefox in the ci, is probably a firefox command line to add, or maybe a driver to install in the machine, or something else to figure out

asturur:
Thank you for all those prs.

asturur (approved):

PR #10603, opened 2025-05-13
chore(tests): replace unmaintained http-server with vercels serve

Smrtnyk:
http-server had no release for 3 years now and outputs a warnings when ran like
 ```text
 [WebServer] (node:3503) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated
 [WebServer] (Use `node --trace-deprecation ...` to show where the warning was created)
 ```
 which is annoying
 we are better off with something more modern and properly maintained.
 There is also a way to not use http server at all for playwright but just intercept the routes via route handlers and autorespond with files from filesystem.
 That would avoid dependency on http server completely and would be faster in theory.


asturur:
i think i should write an action that on merge conflicts for changelog.md solve the conflicts.

asturur (approved):

asturur:
I agree because this is GENERALLY SPEAKING best practice.
But if in one week i get an alert of a npm security issue on the new package, i just revert to the old and think about it when that deprecation is definitely gone and breaking the server.

Issue #10604, opened 2025-05-14
[Feature]: use Intl.Segmenter as default wordSplit


canstand:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have searched and referenced existing issues, feature requests and discussions
- [x] I am filing a **FEATURE** request.

### Description

It's nice to see the recent update to `graphemeSplit`. `Intl.Segmenter` also supports word segmentation via the parameter `{ granularity: 'word' }`. 

If `wordSplit` is similarly refactored, it will be able to support automatic line breaking in multiple languages ​​by default.

### Current State

Currently, the default method is to split words only by spaces, which does not support languages ​​that use non-space line breaks, such as CJK.

asturur:
Yes that is a good idea, it could allow for 'splitByGrapheme' to go away too.
The idea was to leverage that as much as possible, and it will require some time before we can trash the old fallback code for browser support.

You are welcome to draft a PR if you are ok with edits/comments/reviews  that can be slow and extensive at the same time.

PR #10606, opened 2025-05-15
feat: use Intl.Segmenter as default wordSplit method

canstand:
## Description

#10604 

Hi @asturur, I will try to make it happen.
The plan is divided into two steps:
1. Refactor the existing word segmentation method, and retain the separator in the returned array;
2. Introduce `Intl.Segmenter` as the default method, and use the previous method as fallback.

## In Action

The first step of the refactoring is to change the regular expression into a capture group to retain the separator, and filter out the useless empty array items in the `wordSplit` return value.

`wrapLine` no longer needs to insert `infix`. To keep the results as consistent as possible with the previous ones, the single space at the beginning and end of the line is intentionally ignored. 

There are two inconsistencies after the refactoring:

1. 
  - **Before**: `\t` and `\r` are replaced with spaces; 
  - **After**: `\t` and `\r` are retained

2. 
  - **Before**: one space is removed regardless of the number of spaces at the beginning and end of the line; 
  - **After**: if there is more than one space at the beginning and end of the line, it remains unchanged

Is it necessary to maintain complete consistency with the pre-refactoring state?

### DIFF
--- a/src/shapes/Textbox.spec.ts
+++ b/src/shapes/Textbox.spec.ts
@@ -360,6 +360,7 @@ describe('Textbox', () => {
 
     expect(wordsData[0], 'All words have the same length line 0').toEqual([
       { word: ['w', 'o', 'r', 'd'], width: largestWordWidth },
+      { word: [' '], width: 10 },
       { word: ['w', 'o', 'r', 'd'], width: largestWordWidth },
     ]);
     expect(wordsData[1], 'All words have the same length line1').toEqual([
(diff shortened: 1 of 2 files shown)


canstand:
Did a little research. The current version wraps lines, by default, each soft line break will swallow a whitespace character (`[ \t\r]`). 

To maintain compatibility (consistent rendering results) and support languages ​​that do not break lines with spaces, it is necessary to remember whether spaces are removed at each line wrap.

I prefer to introduce a break change and adopt a new line wrapping method: **pre-wrap**.

I looked at the line breaking method of Inkscape, as shown below:

Inkscape line wrap

It behaves in accordance with [CSS whitespace](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) `white-space: pre-wrap;`:
- Spaces and tabs are preserved, and hang trailing whitespaces (not limited by the line box width).

CSS white-space: pre-wrap

Looking forward to feedback.

Issue #10593, opened 2025-05-10
[Bug]: `Cmd/Ctrl + A` does not highlight selection for `Textbox` in `Group`


par5ul1:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.6.4

### Node Version (if applicable)

None

### Link To Reproduction

https://jsfiddle.net/361749xa/5/

### Steps To Reproduce

1. Create `Group` (with `selectable: false`, `subTargetCheck: true`, `interactive: true`)
2. Add `Textbox` (or `IText`)
3. Enter editing mode. See how selecting the text via cursor or even `Shift + Arrow` shows the highlight but `Cmd/Ctrl + A` does not

### Expected Behavior

The selection highlight to show up when I `Cmd/Ctrl + A`

### Actual Behavior

The selection _highlight_ not showing up. The full text is still selected (as pressing any key will clear the textbox and replace the content)

### Error Message & Stack Trace

```bash

```

par5ul1:
Not for the purposes of this Issue, but I'm also open to workarounds in the meanwhile if anyone with more Fabric knowledge can think of any. My use case is very controlled (isEditing check, Ctrl/Meta + A) so if I can hack around the rendering race, that'd also be cool.

par5ul1:
Okay, some development has been made on the linked PR. Since my proposed fix is not apt for this package (very fair), I have monkeypatched _my_ code to use the solution while I wait for a proper update. For anyone else encountering this issue, here's a gist that _might_ work for you:

https://gist.github.com/par5ul1/e61fe370ccf45af9086faca87bf06294

PR #10607, opened 2025-05-16
Fix IText selection not rendering while editing inside of cached Group

par5ul1:
## Description

This pull request addresses a bug where the text selection highlight does not appear when using `Cmd/Ctrl + A` in a `Textbox` or `IText` within a `Group`.

The root cause is that the `Group` class does not properly handle rendering the selection highlight for an IText object in editing mode when the selection is triggered programmatically (e.g., via Cmd/Ctrl + A). While cursor-based or Shift + Arrow selections correctly display the highlight, the Cmd/Ctrl + A action selects the text but fails to render the visual highlight, though the text is still technically selected.

In a direct-on-canvas `IText`, after the `cmdAll` command clears the context, it gets properly restored (with selection) via directly rendering the object. On the other hand, grouped `IText`s are restored from cache, and the cache does not include the selection (visually).

The fix modifies the `Group` class' `renderCache` method to check for active `IText` objects in editing mode with a valid selection range. If such an object is found, the canvas's `contextTopDirty` flag is set to true, and the `renderCursorOrSelection` method is called on the IText object to ensure the selection highlight is rendered (similar to how ungrouped `IText`s behave.)

### Related Issue

Issue #10593

## In Action

TODO

### DIFF
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -35,6 +35,7 @@ import {
 import type { SerializedLayoutManager } from '../LayoutManager/LayoutManager';
 import type { FitContentLayout } from '../LayoutManager';
 import type { DrawContext } from './Object/Object';
+import { IText } from './IText/IText';
 
 /**
  * This class handles the specific case of creating a group using {@link Group#fromObject} and is not meant to be used in any other case.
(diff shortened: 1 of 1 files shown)


asturur:
I tried to fix this bug when the clipPath for text selection was added, and it was when i also discovered the bug first time.
As far as i remember the bug is there also when the cache is disabled, but outside that the main issue here is that a method renderCache for a group shouldn't trigger another render on something outside the group.

Those are the behaviours i try to remove from the library when i find them and i find an alternative, so for me this solution is not applicable.

par5ul1:
The draw cache for `Object` is where the grouped and ungrouped `IText`s diverged when I was stepping through the render in the debugger.

Disabling cache works for when there are selected elements (kind of similar to how it is with shadows) but that immediately breaks with clip paths (and it breaks the entire canvas even after leaving text editing).

I'm happy to find a proper fix for this, and I understand the shortcoming of this suggested fix, but I'd really appreciate some code hints or "what it would take" to properly fix this issue, given the root (probably) that `Groups` render from cache post-‘cmdAll` clear and that doesn't include a selection.

asturur:
The issue is that i fiddle with it a little bit some weeks ago, and i couldn't really find a proper solution.
The whole system of 'contextTopIsDirty' is also an issue. I will try to progress with my open pr for find target, and as soon as i m done i ll revisit this issue and try to find a direction we can move toward.
It may be is a bit of a larger fix with some backward incompatibility from an api point of view and now would be a good moment to get it done.

par5ul1:
Okay, sounds good. I'll close this PR then, and await. Best of luck, and please let me know if I can help in any way.

PR #10608, opened 2025-05-17
refactor(tests): move generic rendering tests from qunit to playwright

Smrtnyk:
Removed duplicate goldens since some of the tests were already running in both playwright and qunit
grouped generic rendering tests into one file, which were in generic_rendering.js in qunit
Rest of the goldens are just moved


test/visual/golden/generic1.png
Smrtnyk, line 1: these removed goldens are duplicates that already exist in playwright

asturur:
i didn't noitce how now tests takes up to 1 second to execute in playwright.
I remember the visual tests to flash super fast on the screen. maybe is a ci thing.
I m not worried about that since a test does the browser and the node check, but i didn't expect so slow

asturur (approved):

Smrtnyk:
yeah on my local machine they are super fast
i guess CI is a bit limited

asturur:
before the rendering test took 3 seconds as a whole
```
  ✓  5 [chromium] › controls/rendering/index.spec.ts:10:5 › control box rendering (3.0s)
```
That means the division in multiple steps maybe takes a bit more time, but is fine is the right thing to do

Smrtnyk:
ah yeah
I mean now each test case is its own test
I think it is for the better
we can also add now `focus: true` to the single test entry and only that test will be run
I think this sacrifice is more worth over congesting everythng into one big huge test
since it gives us better debuggeability

PR #10609, opened 2025-05-17
refactor(tests): move group layout tests from qunit to playwright

Smrtnyk:
This is now over 200 new tests in playwright
All goldens are just moved, not a single one was modified and tests all work.
I have added an option to the rendering test case type to allow for a snapshot directory
so that all goldens can have old name and be placed in group-layout dir for snapshots

I have increased playwright workers for CI to 4, otherwise running over 300 tests now with single one would be painfully slow.
we can try even more than 4, I leave it up to you


playwright.config.ts
```diff
@@ -28,7 +28,7 @@ const config: PlaywrightTestConfig = {
   /* Retry on CI only */
   retries: process.env.CI ? 1 : 0,
   /* Opt out of parallel tests on CI. */
-  workers: process.env.CI ? 1 : undefined,
```
asturur, line 31: ehhh i m not sure this is going to work out. let's see.

Smrtnyk: well it worked in the CI as you can see with current build

asturur:
Please drop an eye if i solved the conflicts correctly

playwright.config.ts
asturur: The issue we had initially i thought it was with fabric globals.
Text tests were doing some overlapping when run in parallel because of fabricJS globals buffers like the copy-paste text variable or other small config issues.
putting ci to 1 solved the problem, but let's see because then we also found different remediations that were in the playwright api without impacting the number of parallel tests for the other tests

Smrtnyk:
> Please drop an eye if i solved the conflicts correctly

I am used to rebases more than merge commits but the master merge to this branch looks good to me
except that prettier is failing

asturur:
yes i added back a const you moved out previously. i ll delete it.

playwright.config.ts
Smrtnyk: > Text tests were doing some overlapping when run in parallel because of fabricJS globals buffers like the copy-paste text variable or other small config issues.

I see that as an issue if you run tests in one test run
like we had before
but now each test is its own isolated page instance and its own dedicated worker
so each test will spawn and import its own fabric global, that is for the browser part

for node part of the tests not sure, each test is own worker, but workers are reused if test passes
but lets see
so far plawyright tests are working

asturur (approved):

PR #10610, opened 2025-05-18
refactor(tests): move global composite operation tests from qunit to playwright

Smrtnyk:


e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+function generateGlobalCompositeTests(): renderTestType[] {
+  return OPERATIONS.map(
+    (operation): renderTestType => ({
+      title: `globalCompositeOperation: ${operation}`,
+      golden: `${operation}.png`,
+      snapshotSuffix: 'gco',
+      percentage: 0.04,
+      size: [size.width, size.height],
```
Smrtnyk, line 158: @asturur 4 tests are failing on node side on my local machine and I can't figure out why
you can focus them if you put `only: true` in here in options
then only this test file will be focused

I am not sure why tests just don't want to pass (these 4) on node side, might be my local machine
they do work just fine in playwright in browser

I tripple checked that I migrated the script properly from qunit, can't spot something that I missed

e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts
```diff
@@ -0,0 +1,176 @@
...
+      async renderFunction(canvas, fabric) {
+        // goldens are transparent
+        canvas.backgroundColor = 'transparent';
```
Smrtnyk, line 161: I added this, since goldens are transparent and in global beforeEach in playwright we set it to white for some reason

asturur:
ok if i have to dig in those tests i ll need some time.

Smrtnyk:
> ok if i have to dig in those tests i ll need some time.

no worries we are not in a hurry

e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts
Smrtnyk: I mean each test should be purely independent and not rely on any kind of state of a previous test case
otherwise it is a brittle testing practice and a footgun in the long run

Smrtnyk: what is weird it passes the browser step, meaning chromium properly generates the proper image
but node part fails, it ends up being without these generated circles

asturur:
my gut feeling is that all nodes test are somehow failing and that those are the one that are more visibile, playwright has its own measurment that differs probably? with the current master ti should be possible to create the snasphots by deleting them and running the visual tests in node, i want to see what does node creates today, since we update it substantially.

asturur:
i switched the drawing operation order between the small preview and the big gradient to see if it changes anything. Is like the background gets deleted in node after the first small draw, that with the operations that are failing seems to match what should happen.

Smrtnyk:
there is something interesting
If you remove this preview creation from the canvas.add
and do it like this
```typescript
canvas.add(
          createNew(fabric, operation),
        );
        ```
        then the test passes

File: e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts (hunk 2 above)
Comment (Smrtnyk) [inline e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts:161]: ok seems like this is the actual culprit
If I set the backgroundColor to transparent then it gets an empty canvas
when I remove this all tests then start to output stuff but many then fail because of white background we set by default in beforeEach
if we remove that setting then all tests pass except these 4, but at least they output something, although it still doesn;t match fully

so seems like it is something going on with these backgrounds
if I put return canvas.toDataUrl() after canvas.renderAll() in the global composite tests loop, then they get very close in the output, but circles are missing some transparency

File: e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts [hunk 3]
```diff
@@ -0,0 +1,176 @@
...
+    fill: 'blue',
+    left: 15,
+    top: 15,
+  });
+
+  const r2 = new fabric.Rect({
+    width: 30,
+    height: 30,
+    fill: 'red',
+    globalCompositeOperation: operation,
```
Smrtnyk, line 145: ok if we remove this line the test works just fine for all of the cases for both node and browser
now, why does this line break the tests, no idea

e2e/tests/visual-output/rendering/testcases/global-composite-operation.ts
asturur: Yeah my comment about image vs canvas was that maybe node-canvas internally could have a canvas tainted in some way with the global composite operation, not that would bleed test by test. But i tried to change it to an image and nothing really changed

asturur:
ok your change made me understand where the issue is.
Node by default has caching off.
And the group with r1 and r2 is uncached.
that means that r2 in the case of destination out is killing everything else that is not composing with the small red rectangle.

Smrtnyk:
> ok your change made me understand where the issue is. Node by default has caching off. And the group with r1 and r2 is uncached. that means that r2 in the case of destination out is killing everything else that is not composing with the small red rectangle.

Amazing
We are close to getting rid of qunit

asturur (approved):

PR #10613, opened 2025-05-21
refactor(tests): move text rendering cases from qunit to playwright

Smrtnyk:


e2e/tests/visual-output/rendering/testcases/text.ts
```diff
@@ -0,0 +1,977 @@
...
+        width: 200,
+      });
+
+      canvas.add(text);
+      canvas.renderAll();
+      canvas.setActiveObject(text);
+      text.enterEditing();
+      text.selectAll();
+
+      // if (!canvas.contextTopDirty) {
```
Smrtnyk, line 915: in original qunit version this has the assert that contextTopDirty is truthy
in here since we don't have assertion mechanism in the browser I had to use throw of the error 
however this breaks, for some reason
so I commented it out in here

not sure it is the visual rendering test responsibility to assert on these imlementation details other than comparing the visual outputs

anyway, seems like after text.selectAll() canvas is not marked as dirty for whatever reason

asturur: Yes that is a current bug. selectAll doesn't mark as dirty and when i looked into it i couldn't fix it quickly but i forgot why. It was discovered by looking at text inside groups, i think selectAll was leaving on the fact that there was a cursor animation before that would leave the context dirty when you switch to full selection.

playwright.config.ts
```diff
@@ -61,6 +61,9 @@ const config: PlaywrightTestConfig = {
       use: {
         ...devices['Desktop Chrome'],
         browserName: 'chromium',
+        launchOptions: {
+          args: ['--disable-gpu'],
+        },
```
asturur, line 66: if we disable gpu the tests that involve  filtering will be super slow. is this necessary because of parallelization? or because the gpu on the instance is bad and trying to use it cause more issues than benefits?

e2e/tests/visual-output/rendering/index.ts
```diff
@@ -11,6 +11,11 @@ beforeRenderTest(
     return renderTests.map((renderTest) => {
       return {
         async boundFunction() {
+          if (renderTest.hasOwnProperty('enableGLFiltering')) {
+            fabric.config.configure({
+              enableGLFiltering: renderTest.enableGLFiltering,
```
asturur, line 16: this could not work well if switched between tests.

playwright.config.ts
Smrtnyk: ah sorry
I forgot to remove this
I saw that qunit had this enabled and thought I needed it because some tests failed but this wasn't related
so forgot to remove it

e2e/tests/visual-output/rendering/index.ts
Smrtnyk: but this is what qunit did
have a look at this
 https://github.com/fabricjs/fabric.js/pull/10613/files#diff-0ea34a0a8eb6a9379c2e2241313e09483701f7c94a24ebf8060963424e9a2a05L3

e2e/tests/visual-output/rendering/testcases/text.ts
Smrtnyk: I wonder why this worked in qunit since it had assertion on this

e2e/tests/visual-output/rendering/index.ts
asturur: ok if each page import its own fabric library that saves it. i think the filter code doesn't get swapped if you disable webgl fitlering after a filtering operation

playwright.config.ts
asturur: i ll remove it

Smrtnyk: > i ll remove it

Tnx

Smrtnyk:
ok I have removed the disable gpu flag and rebased on latest master

asturur (approved):

PR #10615, opened 2025-05-21
refactor(tests): move resize filter test cases from qunit to playwright

Smrtnyk:


test/visual/svg_import.js
```diff
@@ -2,7 +2,6 @@
   fabric.config.configure({
     enableGLFiltering: false
   });
-  fabric.Object.ownDefaults.objectCaching = true;
```
Smrtnyk, line 5: yeah, this proves that qunit is full of sideeffects between tests
that I had to remove this
because some previous test in another suite had this flipped

asturur: I grew my js expertise and test understanding working of fabricjs during the years before professional js was a carreerr for me. I m sure code written between 2015-2018 is full of poor decisions based on 'i need this to work for work related issues' and 'tests is better than no tests'

asturur:
This is good.
When all is settled if we see there is a need for cases like this getImage to be available on globlaThis and global, we could think of refactoring by injecting an utility object from the test runner so that we have one for browser and one for node.
I understand that having visual tests running in playwright browser and also by the test runner in node is not the standard playwright usage.

asturur (approved):

Smrtnyk:
> This is good. When all is settled if we see there is a need for cases like this getImage to be available on globlaThis and global, we could think of refactoring by injecting an utility object from the test runner so that we have one for browser and one for node. I understand that having visual tests running in playwright browser and also by the test runner in node is not the standard playwright usage.

we could maybe split this into projects
1 project is chromium and another one node
running same suite of tests
to not mix up the stuff
I could have a look later after we clean the repo out of qunit
I also have some idea to speed up playwright tests that I want to try out

PR #10616, opened 2025-05-21
refactor(tests): move free draw rendering tests from qunit to playwright

Smrtnyk:
This is built on top of the resize filter branch, so needs that one to be merged first


Smrtnyk:
after this one only svg import tests are left
and then we can completely prune qunit and its ancient dependencies

asturur (approved):

PR #10618, opened 2025-05-23
refactor(tests): move svg import tests from qunit to playwright

Smrtnyk:
@asturur this is the last file left 
there are no more qunit visual tests
but in this one there is an issue
25 tests are failiing. majority is passing
goldens are just moved

Issue with failing tests is just something related to dimensions
most failing tests have size off by 1px or something
just a few by 10px
visually they look identically, just the dimensions are off

not sure if it worked in qunit before because of some sideffects from other tests, or visual test matching by pixel match was not caring about size or whatever, but I have no clue
Maybe we should just delete these failing ones and let plawyright recreate them a new?
since they look visually identical anyway

I leave this one up to you


Smrtnyk:
after we merge this I will do a big cleanup to remove all unused deps and files

asturur:
From the github action panel is possible to download and look at the results, and the diffs should be talking.
If is just a size issue it should be ok to update them

Smrtnyk:
> From the github action panel is possible to download and look at the results, and the diffs should be talking. If is just a size issue it should be ok to update them

yeah I had a look already, that is why I said to you that they look the same to me, but size is a bit different
I don't really understand how pixelmatch in qunit really worked and if it was just looking at same pixels or it also checked strictly sizes
but your word has to be last here
if you want I can recreate these goldens locally and push them in new commit, and in that commit you can check the diff of deleted and newly added one

e2e/tests/visual-output/rendering/testcases/svg-import.ts
```diff
@@ -0,0 +1,105 @@
...
+    golden: `${svgName}.png`,
+    percentage: 0.055,
+    snapshotSuffix: 'svg-import',
+    size: [100, 100],
+    only: true,
+    async renderFunction(canvas, fabric) {
+      const asset = await globalThis.getAsset(`${svgName}.svg`);
+      const { objects, options } = await fabric.loadSVGFromString(asset);
+      const nonNullObj = objects.filter((obj) => !!obj);
+      const group = fabric.util.groupSVGElements(nonNullObj, options);
```
asturur, line 24: i need to open an issue to add null filtering in groupSvgelements.

e2e/tests/visual-output/rendering/testcases/svg-import.ts
```diff
@@ -0,0 +1,105 @@
...
+      const dims = group._getTransformedDimensions();
+      canvas.setDimensions({
+        width: dims.x + group.left,
+        height: dims.y + group.top,
+      });
```
asturur, line 29: the canvas gets definetely resized to what the content is, so yes we should regenerate the originals

Smrtnyk: alright I will push recreated goldens

Smrtnyk: I checked svg_radial_13.svg asset
it has this
<rect x="0" y="0" width="100" height="100"
    style="fill: url(#focal_set); stroke: black;"/>
</svg>

so width is 100 and height is 100
and playwright output is also those dimensions
however expected one has 300 x 150
so seems like playwright is more precise here anyway

Smrtnyk:
I have pushed new commit with updated goldens
lets see what CI says

Smrtnyk:
Ok CI is happy now

e2e/tests/visual-output/rendering/testcases/svg-import.ts
Smrtnyk: they are all transparent now
as I didn't want to regenerate all of them
just the failing ones so I set the bg transparent in the test case

PR #10620, opened 2025-05-24
chore(): BREAKING remove exports for  parseAttributes, parseStyleAttribute, parseFontDeclaration, parsePointsAttribute, parseTransformAttribute and getCSSRules

Smrtnyk:
not surei f you want this or not
but these were marked with todo to remove after tests are migrated, so I did that

it is being removed from a public fabric namespace so it is breaking change in case someone is using it
I let you decide on this


src/parser/parser.spec.ts
```diff
@@ -340,7 +336,6 @@ describe('fabric.Parser', () => {
   it('parseTransformAttribute', () => {
     let parsedValue;
 
-    expect(parseTransformAttribute).toBeDefined();
```
Smrtnyk, line 343: these make 0 sense to assert, since tests would break on compile if they wouldn't exist
it made sense back then with qunit but with vitest is redundant

asturur:
Yes the SVG parser is offered and supported as a whole.
Our internal utilities to make svg parsing work may not make sense out of context, i do not think should be part of the api.
If it turns out a function was useful someone will hopefully complain and we can still restore it.
You shouldn't trust our css parsing ability for anything else than parsing svgs really.

asturur:
the changelog update action is not working as expected with pr rename.

asturur (approved):

PR #10621, opened 2025-05-24
chore(): cleanup after qunit removal

Smrtnyk:


package.json
```diff
@@ -91,28 +85,20 @@
     "@vitest/coverage-v8": "^3.1.2",
     "@vitest/ui": "^3.1.2",
     "babel-plugin-transform-imports": "git+https://git@github.com/fabricjs/babel-plugin-transform-imports.git",
-    "busboy": "^1.6.0",
```
Smrtnyk, line 94: many deps now removed
fabric is now really a more lightweight project in terms of maintenance

test/visual/stroke_projection.js
```diff
@@ -1,307 +0,0 @@
...
-            fabricClass: 'Canvas',
-            disabled: fabric.isLikelyNode,
-          });
-        });
-      });
-    });
-  });
-
-  tests.forEach(visualTestLoop(QUnit));
-});
```
asturur, line 307: we shouldn't loose those. Have these been ported over?

asturur: i don't want to delete it because i will forget it exists, it may be in git yes, but if i don't remember is like is not there.
just leave the file alive in the folder, is enough.

asturur (approved):

PR #10622, opened 2025-05-24
refactor(tests): use fixtures instead of calling setup in every test file

Smrtnyk:
@asturur I brought this one back now after all tests are migrated and conflicts are solved
since you said that you want to read the diff


asturur:
i ll do this next, but i want to do a quick bug fix first and finish my find target pr.

Smrtnyk:
> i ll do this next, but i want to do a quick bug fix first and finish my find target pr.

No worries Andrea, this can wait
there are plenty of other stuff for me to do in the mean time
I could look at types in source code if something can be improved, vitest unit tests also have type issues
more lint rules
maybe even more coverage for current tests

PR #10623, opened 2025-05-24
fix(): Remove black border in blur and also align JS with WEBGL blur filtering

asturur:
## Description

Change the logic in the blur filter to use a weight for the alpha channel when blurring opaque with transparent pixels
See https://github.com/fabricjs/fabric.js/discussions/10611 for reference

## Edit

I thought this was a 15 minutes job, maybe 30 with the js redux.
I m like 4 hours in and i m stuck

This is the webgl filtering

image

And this is the JS one:

image

The JS is completely broken, the black border is there and the image is shifted up and a chunk at the end is missing.
I would imagine something like that is like clearly visible in the code but it isn't. (not for me)

## Update

ok i made it


src/filters/Canvas2dFilterBackend.ts
```diff
@@ -31,7 +31,10 @@ export class Canvas2dFilterBackend {
     sourceHeight: number,
     targetCanvas: HTMLCanvasElement,
   ): T2DPipelineState | void {
-    const ctx = targetCanvas.getContext('2d');
+    const ctx = targetCanvas.getContext('2d', {
+      willReadFrequently: true,
+      desynchronized: true,
+    });
```
asturur, line 37: in 2d filtering we don't leverage the gpu, the context can leverage the 'willReadFrequently' optimization.

PR #10624, opened 2025-05-25
chore(): modernize eslint config

Smrtnyk:
I noticed that eslint config was using deprecated way of defining config via compat utils
that is not needed anymore now

I also noticed that current state of eslint is that it operates on a very basic set of rules
there are many more stricter rules that can eliminate many potential bugs or enforce more readable code
we could consider some of additional lint rules if you have interest into this


asturur:
In general i use lint to catch error and so i use the recomended profile.
I do leave 'nice to read code' to prettier for spacing and what not.
I try to stay away from the opinion real where 'this code is better than this other code' but i can't measure or verify it.
It boils down to which rules we are talking about, if you have some favorites that you think are really good send a link i ll read them

asturur (approved):

Smrtnyk:
> In general i use lint to catch error and so i use the recomended profile. I do leave 'nice to read code' to prettier for spacing and what not. I try to stay away from the opinion real where 'this code is better than this other code' but i can't measure or verify it. It boils down to which rules we are talking about, if you have some favorites that you think are really good send a link i ll read them

yes and I totally agree to you
I also let prettier handle code looks for me also in my private projects
but lint is more powerful also beyond the standard ruleset,
like in this PR https://github.com/fabricjs/fabric.js/pull/10626
and there are many more useful lint rules
I will open PRs with some of the more valuables I find and then you can decide, either merge or drop the PR

PR #10625, opened 2025-05-25
chore(): try to fix firefox vitest test

Smrtnyk:
trying out to see if it fixes it


Smrtnyk:
hm now different tests fail

Smrtnyk:
seems like it passed now

PR #10627, opened 2025-05-25
chore(): investigate firefox flakyness in vitest

Smrtnyk:
will test ci to see if this removes flakyness


Smrtnyk:
I couldn't really figure out why this one test is flaky on firefox
it passes every single time locally for me
but in CI is just flaky
I think it is some kind of a race condition when getting canvas webgl context being not ready yet so it is not assigned to the backend instance as gl property

anyway, I ended up skipping this 1 test on firefox
unless you want to dig into it to figure out why it is flaky

Smrtnyk:
ok rather than skipping the test on firefox I changed it to add retry
so if it is flaky it will just retry
I guess it is better than just skipping it

PR #10631, opened 2025-05-26
chore(): enable @typescript-eslint/no-unnecessary-type-arguments lint rule

Smrtnyk:
Prevents passing redundant types to places where they are already infered or are already default ones


Smrtnyk:
> this is an ok rule

I am picking only the ones that lead to more code clarity and simpler code

asturur (approved):

PR #10637, opened 2025-05-26
refactor(tests): introduce strict playwright typecheck and fix all type issues

Smrtnyk:
This is built on top of the https://github.com/fabricjs/fabric.js/pull/10622 so we can merge it in case other one gets accepted.

What I noticed is that e2e directory was full of type errors, type inference and autocompletion was not working for lots of stuff, defeating the purpose of typescript.

This PR makes sure e2e directory is typechecked and that all current type issues are fixed


e2e/tsconfig.json
```diff
@@ -1,20 +1,28 @@
...
+    "baseUrl": ".",
     /**
      * A reflection of the configured imports used by test scripts
      * **IMPORTANT**: be sure to keep this updated to reflect `./imports.ts` correctly
      */
     "paths": {
       "fabric": ["../dist"],
       "fabric/node": ["../dist/index.node"],
-      "fabric/extensions": ["../dist-extensions"]
+      "fabric/extensions": ["../dist-extensions/extensions"]
```
Smrtnyk, line 22: this one is fishy
previous imports work as well, but they have type issues
seems like `dist-extensions` is pointing to fabric types and that extensions types for its lib are emitted and is nested under extensions

asturur: this wasn't really planned.
I wanted to give devs the ability to contribute not core specific features without having the library becoming a 1 megabyte of minified monolithic js
And i did what i could do quickly.
It is not easy because this part of the lib needs to reference fabricJS as an external as a whole.
But also this doesn't play great if you want to use fabricJS as an esm module where you import what you need.

I guess we can't have it all

If there is a reference to implement this correctly and you know where to find it, i ll study it.

Smrtnyk: we could make fabric a monorepo
and add packages
core package
and extensions package
then extensions package can import from core package what it needs
and then you publish them separately
with 1 command
you could have published fabric main package as it is now
and additionally `@fabric/extensions` which user can install extra if they need it

this is just an idea, which i observed that many packages do nowaday

PR #10639, opened 2025-05-28
test(Textbox): add edge cases when wrapping line

canstand:
## Description

While preparing #10606, I found that the existing tests were not accurate enough, so I added some edge cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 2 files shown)


asturur:
oh this is great thank you. I didn't forget about the segmenter, is just a busy period.
As you see there is a bunch of activity on the maintainance side of things

asturur (approved):

PR #10643, opened 2025-05-30
feat(): Add support for text decoration tickness

### LINKED ISSUE #10640: [Feature]: Add parameter on text underline thickness
### CheckList

- [x] I have searched and referenced existing issues, feature requests and discussions
- [x] I am filing a **FEATURE** request.

### Description

Following this answer: 
https://github.com/fabricjs/fabric.js/issues/757#issuecomment-2897969720

We would like to have control on underline thickness.

Ideal solution would be to control height in pixel of the underline and make it customizable by character.  
Thickness  would grow around a line (so half thickness top and half thickness bottom).  
If this solution is used I suggest to call parameter `underlineThickness` to match CSS spec.

For moment thickness is built using fontSize and a 15 factor. Another good solution would be to allow dév to control this factor.  
If this solution is implemented I suggest to call parameter `underlineFactor` and not `underlineThickness` as it would be less confusing and more easy to deprecate it once CSS or Font API would make it easier to get font intrasic information.

Controling underline distance is not requested here although if it is easy to add it could be a great addition.
Dealing with underline interruption with bottom of font caracters (like 'p' or 'g') is not requested here.

### Current State

Current state does not take into account font underline thickness parameter.

Image

### Additional Context

_No response_

asturur:
## Description

closes #10640.

This PR implements the basic feature requested in the linked issue.
The reason behind chosing the current unit of measure is to align with the current charSpacing api.

### DIFF
--- a/src/shapes/Text/Text.ts
+++ b/src/shapes/Text/Text.ts
@@ -107,6 +107,7 @@ interface UniqueTextProps {
   textAlign: string;
   direction: CanvasDirection;
   path?: Path;
+  textDecorationTickness: number;
 }
 
 export interface SerializedTextProps
@@ -294,6 +295,16 @@ export class FabricText<
    * @default
    */
   declare path?: Path;
+
+  /**
+   * The text decoration tickness for underline, overline and strikethrough
+   * The tickness is expressed in thousandths of fontSize ( em ).
+   * The original value was 1/15 that translates to 66.6667 thousandths.
+   * The choice of unit of measure is to align with charSpacing.
+   * @type Number
+   * @default
+   */
+  declare textDecorationTickness: number;
 
   /**
    * Offset amount for text path starting position
@@ -1599,7 +1610,7 @@ export class FabricText<
             -charBox.kernedWidth / 2,
             offsetY * currentSize + currentDy,
             charBox.kernedWidth,
-            this.fontSize / 15,
+            this.fontSize * this.textDecorationTickness,
           );
           ctx.restore();
         } else if (
@@ -1620,7 +1631,7 @@ export class FabricText<
               drawStart,
               top + offsetY * size + dy,
               boxWidth,
-              this.fontSize / 15,
+              this.fontSize * this.textDecorationTickness,
             );
           }
           boxStart = charBox.left;
@@ -1644,7 +1655,7 @@ export class FabricText<
           drawStart,
           top + offsetY * size + dy,
           boxWidth - charSpacing,
-          this.fontSize / 15,
+          this.fontSize * this.textDecorationTickness,
         );
       topOffset += heightOfLine;
     }
--- a/src/shapes/Text/constants.ts
+++ b/src/shapes/Text/constants.ts
@@ -33,6 +33,7 @@ export const additionalProps = [
   ...textDecorationProperties,
   'textBackgroundColor',
   'direction',
+  'textDecorationTickness',
 ] as const;
 
 export type StylePropertiesType =
@@ -47,7 +48,8 @@ export type StylePropertiesType =
   | 'deltaY'
   | 'overline'
   | 'underline'
-  | 'linethrough';
+  | 'linethrough'
+  | 'textDecorationTickness';
 
 export const styleProperties: Readonly<StylePropertiesType[]> = [
   ...fontProperties,
@@ -57,6 +59,7 @@ export const styleProperties: Readonly<StylePropertiesType[]> = [
   FILL,
   'deltaY',
   'textBackgroundColor',
+  'textDecorationTickness',
 ] as const;
 
 // @TODO: Many things here are configuration related and shouldn't be on the class nor prototype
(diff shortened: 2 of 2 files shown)


Smrtnyk:
should we add rendering test case?

src/shapes/Text/Text.ts
```diff
@@ -295,6 +296,16 @@ export class FabricText<
...
 
+  /**
+   * The text decoration tickness for underline, overline and strikethrough
+   * The tickness is expressed in thousandths of fontSize ( em ).
+   * The original value was 1/15 that translates to 66.6667 thousandths.
+   * The choice of unit of measure is to align with charSpacing.
+   * @type Number
+   * @default
+   */
+  declare textDecorationTickness: number;
```
Smrtnyk, line 307: is there no default value assigned?

asturur: yes but in the default object.
Fabric5 used to do this to define default values for properties:
```js
Fabric.Text.prototype.fontSize = 40;
```
And if the developer wanted to change the default fontSize could do the same.

When moving to classes we realized that wasn't possible anymore.

if you do:
```js
class Text {
  fontSize = 40;
}
```
That is not going to be in the prototype but on the instance and there is no way to change it globally.

So we declare the configurable properties would be only declared and assigned from an object in the constructor.
Modifying that object is the only way to change the default value for all the new instances.

asturur:
is thickness, not tickness. :D

asturur:
image

i m sure i m going to get the svg export at some point

asturur:
Safari-chrome-firefox svg export:

image

At least firefox renders the svg correctly

asturur:
i ll move failing svg tests to snapshot tests since the current way of testing them with premade strings is a pain.

asturur:
Anyway this "straightforward" pr is going to bleed in its third day of work.

src/shapes/Text/TextSVGExportMixin.spec.ts
```diff
@@ -41,48 +42,27 @@ describe('TextSvgExport', () => {
     });
 
     it('toSVG', () => {
-      const TEXT_SVG =
-        '<g transform="" style=""  >\n\t\t<text xml:space="preserve" font-family="Times New Roman" font-size="40" font-style="normal" font-weight="normal" style="stroke: none; stroke-width: 1; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(0,0,0); fill-rule: nonzero; opacity: 1; white-space: pre;" ><tspan x="-10" y="12.57" >x</tspan></text>\n</g>\n';
       const text = new FabricText('x');
-      expect(removeTranslate(text.toSVG())).toEqualSVG(
-        removeTranslate(TEXT_SVG),
-      );
+      expect(sanitizeSVG(text.toSVG())).toMatchSnapshot();
```
Smrtnyk: we have `toEqualSVG` matcher
isn't it doing the same?
asking because it brings a bit inconsistency to the tests
that some are using it and some not

asturur: toEqualSVG expect the expected svg to be passed in as a string.
I tried to write a toMatchSVGSnapshot but i got confused between chai and expect not sure why we are adding so many matchers there

Smrtnyk: chai is what vitest uses for assertion library
expect is just the facade that is exposed to be extended, but chai is under the hood
and we used chai directly because via expect you can't extend the `toMatchSnapshot` matcher so it was the only way

anyway, I agree that we have too many matchers, maybe we could get rid of the toEqualSVG
I could have a look if you want

asturur: we should have `toMatchSvgSnapshot` i can hack one, but it will need a dedicated effort to make them cleaner.
This PR is getting my bored, i want to move on.

asturur: done i guess.

asturur: is that good? it seems to work for me, and is clean enough, i m just not sure on the order for parameters for the out of context call

Smrtnyk: it looks reasonable change to me
I just am not sure if it is typed properly on the expect facade, but I plan to look at proper type-check of unit tests and can solve it then in case it causes some type issues
so no need to worry about it

asturur:
svg import while not important could be activated/tested

asturur:
image
The svg i added does not work outside firefox, i have no idea why the test pass but ok. At least is there.

asturur:
i m not sure i will be able to backport this to the 6.x branch.
I should have moved all tests related pr to 6.x branch but i didn't and i think now is too late.

PR #10648, opened 2025-06-03
chore(): update vitest

Smrtnyk:
Lets be good boys and keep our testing framework up to date
In case you are interested in what is new in 3.2 release https://vitest.dev/blog/vitest-3-2.html


asturur:
A link to the changelog is always a good idea.
Reduces friction for the reader

asturur (approved):

PR #10667, opened 2025-06-18
fix(): nested duplicated clipPath causes infinite recursion

### LINKED ISSUE #10659: [Bug]: nested duplicated clipPath causes infinite recursion
### CheckList

### Version

6.6.4

Chrome

None

### Link To Reproduction

https://codesandbox.io/p/devbox/fabric-vanillajs-sandbox-forked-hnqz4y

### Steps To Reproduce

1. Open the reproduction
2. Click the "Load Good SVG" button, the SVG will be loaded smoothly
3. Click the "Load BAD SVG" button, the page will get stuck (you can see the animation stop) and crash after a while

### Expected Behavior

The "BAD SVG" can also be loaded successfully. 

### Actual Behavior

The infinite recursion inside fabric.js cause the web page crashed.

```bash
No logs will produces, but the calling stack is: `loadSVGFromString -> parseSVGDocument-> ElementsParser#parse -> ElementsParser#createObject -> ElementsParser#resolveClipPath -> ElementsParser#resolveClipPath -> ElementsParser#resolveClipPath -> ElementsParser#resolveClipPath ...` (stuck in infinite recursion).
```

Smrtnyk:
fixes #10659

### DIFF
--- a/src/parser/elements_parser.ts
+++ b/src/parser/elements_parser.ts
@@ -96,9 +96,19 @@ export class ElementsParser {
 
   extractPropertyDefinition(
     obj: NotParsedFabricObject,
+    property: 'fill' | 'stroke',
+    storage: Record<string, SVGGradientElement>,
+  ): { def: SVGGradientElement; id: string } | undefined;
+  extractPropertyDefinition(
+    obj: NotParsedFabricObject,
+    property: 'clipPath',
+    storage: Record<string, Element[]>,
+  ): { def: Element[]; id: string } | undefined;
+  extractPropertyDefinition(
+    obj: NotParsedFabricObject,
     property: 'fill' | 'stroke' | 'clipPath',
     storage: Record<string, StorageType[typeof property]>,
-  ): StorageType[typeof property] | undefined {
+  ): { def: StorageType[typeof property]; id: string } | undefined {
     const value = obj[property]!,
       regex = this.regexUrl;
     if (!regex.test(value)) {
@@ -110,7 +120,7 @@ export class ElementsParser {
     const id = regex.exec(value)![1];
     regex.lastIndex = 0;
     // @todo fix this
-    return storage[id];
+    return { def: storage[id], id };
   }
 
   resolveGradient(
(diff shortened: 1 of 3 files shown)


Smrtnyk:
the unit test I added hangs on vitest without the fix

src/parser/elements_parser.ts
```diff
@@ -122,10 +132,10 @@ export class ElementsParser {
       obj,
       property,
       this.gradientDefs,
-    ) as SVGGradientElement;
+    );
     if (gradientDef) {
       const opacityAttr = el.getAttribute(property + '-opacity');
-      const gradient = Gradient.fromElement(gradientDef, obj, {
+      const gradient = Gradient.fromElement(gradientDef.def, obj, {
```
asturur, line 138: if we are using gradientDef.def, the if condition above needs to check for that.

Smrtnyk: good catch will add few more unit tests

Smrtnyk: I pushed new commit, please recheck

asturur:
I m not sure how this fix works.
We need more svg examples to be sure this is a fix.
i didn't check why it was recursively looping on itself.
If it was for the clippath.clippath reason, i do not think this fix can work for real use case scenario outside this simeple one. We need to build a better use case to test and be sure legit use cases are rendered correctly.

I didn't dig in the code when i checked this issue, but my impression was that i would have had to duplicate the double referenced clipPath with a clone of itself in order to avoid the infinite recursion

Smrtnyk:
> I m not sure how this fix works. We need more svg examples to be sure this is a fix. i didn't check why it was recursively looping on itself. If it was for the clippath.clippath reason, i do not think this fix can work for real use case scenario outside this simeple one. We need to build a better use case to test and be sure legit use cases are rendered correctly.
> 
> I didn't dig in the code when i checked this issue, but my impression was that i would have had to duplicate the double referenced clipPath with a clone of itself in order to avoid the infinite recursion

you are right
I might be just plastering over a bigger issue
lets close for now and I can come back to it when I have more time, if it doesn't get fixed in the mean time

asturur:
Leave it open, the hardest task is to make an svg that has a nested clipPath and that has visual differences with one that doesn't have a nested clipPath. Once we have a reproducible use case of an svg that can render something that works we can add it here and see what happens

Smrtnyk:
> Leave it open, the hardest task is to make an svg that has a nested clipPath and that has visual differences with one that doesn't have a nested clipPath. Once we have a reproducible use case of an svg that can render something that works we can add it here and see what happens

I have reopened it now
will see if I can find some svgs that do something useful for these scenarios
or if someone else has them can plop them in here

Smrtnyk:
closing in favor of https://github.com/fabricjs/fabric.js/pull/10774

Issue #10683, opened 2025-07-18
[Bug]: SVG Parsing ignores multiple style tags


Chicken-3001:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.6.4

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

None

### Link To Reproduction

https://codesandbox.io/p/devbox/fabric-vanillajs-sandbox-forked-qphrsv

### Steps To Reproduce

1. Parse an SVG document with at least 2 non-empty `<style>` tags.

### Expected Behavior

When parsing an SVG document, the rules in every `<style>` tag will be applied to the resulting Fabric.js objects. 

For example, parsing the SVG:
```svg
<svg xmlns="http://www.w3.org/2000/svg">
  <style>
    .blue { fill: blue; }
  </style>
  <style>
    .red { fill: red; }
  </style>
  <circle class="blue" r="45" cx="50" cy="50"></circle>
  <circle class="red" r="45" cx="150" cy="50"></circle>
</svg>
```
should result in the following Fabric.js circles (when rendered on a canvas):

<img width="271" height="145" alt="Image" src="https://github.com/user-attachments/assets/a0645fba-cbad-4c60-84f3-815c5676de8e" />

### Actual Behavior

Any rules outside of the first `<style>` tag don't appear in the resulting Fabric.js objects.

For example, parsing the SVG:
```svg
<svg xmlns="http://www.w3.org/2000/svg">
  <style>
    .blue { fill: blue; }
  </style>
  <style>
    .red { fill: red; }
  </style>
  <circle class="blue" r="45" cx="50" cy="50"></circle>
  <circle class="red" r="45" cx="150" cy="50"></circle>
</svg>
```
results in the following Fabric.js circles (when rendered on a canvas)::

[...]

PR #10684, opened 2025-07-18
fix(SVGParser): Corrected CSS rule parsing for multiple style tags

Chicken-3001:
Fixes [#10683](https://github.com/fabricjs/fabric.js/issues/10683)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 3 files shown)


asturur:
Thank you.
Do you need this backported and published as 6.7.1? otherwise it will go out with 7.0

Chicken-3001:
> Thank you. Do you need this backported and published as 6.7.1? otherwise it will go out with 7.0

Backporting this would be fantastic, thank you

PR #10685, opened 2025-07-19
chore(): update dev deps

Smrtnyk:
A round of dep updates
 Motivated by new playwright release https://github.com/microsoft/playwright/releases/tag/v1.54.0 to test on newer browser versions


Smrtnyk:
@asturur  do you plan to drop node 18 for fabric 7?
if yes then we could also update vite to 7

Smrtnyk:
I might get some free time soon so might come back to fabric codebase again
Will probably look into typescript types in the codebase to see if something can be improved

asturur:
since is taking me forever to finish the find target code, and august is not getting any better with free time, node 24 will become LTS in october it seems ok to remove 18, i ll do it before release.
If staying up to date with node is just about vitest, we can also just stop running the tests on node 18 and remove node 18 only when we want to move to jsdom 27.

Smrtnyk:
> since is taking me forever to finish the find target code, and august is not getting any better with free time, node 24 will become LTS in october it seems ok to remove 18, i ll do it before release. If staying up to date with node is just about vitest, we can also just stop running the tests on node 18 and remove node 18 only when we want to move to jsdom 27.

sounds good, I think we can merge this one as is and then later separately just up vite to 7 when we want it. we could do it also when vitest 4 comes out of beta

asturur:
if you want to do a very easy rewarding pr is to remove the file

```
src/util/internals/dom_request.ts
```

and convert the only function using it to fetch. Without changing the main function arguments possibly.

Smrtnyk:
> if you want to do a very easy rewarding pr is to remove the file
> 
> ```
> src/util/internals/dom_request.ts
> ```
> 
> and convert the only function using it to fetch. Without changing the main function arguments possibly.

I tried that in the past and things broke, it was due to 
` const xml = r.responseXML;` by the consumer in the `loadSVGFromURL` and fetch doesn't have that equivalent
so I just left it as it was
We could stop exporting it on the utils namespace at least

asturur:
> r.responseXML

i think we need to get it as text and then fallback to loadSVGFromString.

So i learned today fetch does not parse the document, has just blob and text but is fine. DomParser used from loadSvgFromString will be as fast as responseXML, if is not the exact same code exposed as an api.

Smrtnyk:
> > r.responseXML
> 
> i think we need to get it as text and then fallback to loadSVGFromString.
> 
> So i learned today fetch does not parse the document, has just blob and text but is fine. DomParser used from loadSvgFromString will be as fast as responseXML, if is not the exact same code exposed as an api.

Will give it a shot

Smrtnyk:
https://github.com/vitest-dev/vitest/releases/tag/v4.0.0-beta.4
@asturur vitest since 4 beta now supports screenshot matching directly in unit tests if you will want something like that in some situations

asturur (approved):

PR #10687, opened 2025-07-20
Fix Next.js compatibility by removing node export restriction

### LINKED ISSUE #10679: [Bug]: Demos cannot works fine with nextjs template
### CheckList

### Version

6.6.4

_No response_

22.10.0

### Link To Reproduction

https://github.com/fabricjs/fabric.js

### Steps To Reproduce

1. npm install
2. npm run start next

and will got the error:

### Expected Behavior

Build failed.

### Actual Behavior

Cannot run.

Nextjs may build in node environment, so in the package.json:
```
  "exports": {
    ".": {
      "node": null,  // <-------------  which cause the error, after delete it, everything works fine
      "types": "./dist/index.d.ts",
      "import": "./dist/index.min.mjs",
      "require": "./dist/index.min.js",
      "default": "./dist/index.min.js"
    },
```

```bash
[2025-07-07 09:36:58] waiting for changes...
 ⨯ ./components/Canvas.tsx:1:0
Module not found: Package path . is not exported from package /Users/howazhong/Public/fabric.js/.codesandbox/templates/next/node_modules/fabric (see exports field in /Users/howazhong/Public/fabric.js/.codesandbox/templates/next/node_modules/fabric/package.json)
> 1 | import * as fabric from 'fabric';
  2 | import React, { useEffect, useRef } from 'react';
  3 | 
  4 | const DEV_MODE = process.env.NODE_ENV === 'development';

https://nextjs.org/docs/messages/module-not-found

Import trace for requested module:
./pages/index.tsx
 ⚠ Fast Refresh had to perform a full reload due to a runtime error.
 ⨯ ../../../node_modules/agent-base/dist/index.js:30:0
```

ssahn208:
Fix Next.js compatibility by removing node export restriction

## Description
Fixes #10679 

This PR removes the `"node": null` entries from the package.json exports field that were causing import errors in Next.js environments.

The issue was that Next.js build process treats the `"node": null` export restriction as blocking the import, causing:

## Changes
- Removed `"node": null` from main export (`.`)
- Removed `"node": null` from es export (`./es`)

## Testing
This change allows fabric.js to be properly imported in Next.js applications without the module resolution error.

## In Action
Tested with Next.js template - import now works without module resolution errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,14 +128,12 @@
   "sideEffects": false,
   "exports": {
     ".": {
-      "node": null,
       "types": "./dist/index.d.ts",
       "import": "./dist/index.min.mjs",
       "require": "./dist/index.min.js",
       "default": "./dist/index.min.js"
     },
     "./es": {
-      "node": null,
       "types": "./dist/index.d.ts",
       "import": "./dist/fabric.min.mjs",
       "require": null,


asturur:
The reason why i didn't merge this yet is that i do not know why a null was there in the first place.
I looked back trying to recover the history but i m uncertain.
Docs says that if you want to use fabricJS in node you have to use the /node import path.
And if you import the normal bundle you shouldn't hit jsdom and node-canvas anymore. so we are safe.

asturur:
according to this explanation:
https://hirok.io/posts/package-json-exports

node: null

Does not do anything outside this error here.

asturur:
In the official documentation the null entry point is only used to block:

image

And what we want to do here is just to direct the developer to understand that the working node version for fabric is under 'fabric/node'.

So i think removing this is inline we what we wanted to achieve in the first place.

asturur:
If you need this published in fabric 6.7.2 you will need to open a pr also on the 6.x branch doing the same thing, citing the backport of #10687.

PR #10690, opened 2025-07-22
chore(): BREAKING: remove request util

Smrtnyk:
Removes request util from the utils namespace and deletes it from the code.
 Changes loadSVGFromURL.ts to use native fetch instead of the request util.


e2e/tests/visual-output/rendering/testcases/z-svg-export.ts
```diff
@@ -459,8 +459,8 @@ export const zSvgExport: renderTestType[] = [
     percentage: 0.06,
     size: [760, 760],
     async renderFunction(canvas, fabric) {
-      const assetName = globalThis.getAssetName('svg_linear_9.svg');
-      const { objects } = await fabric.loadSVGFromURL(assetName);
+      const asset = await globalThis.getAsset('svg_linear_9.svg');
+      const { objects } = await fabric.loadSVGFromString(asset);
```
Smrtnyk, line 463: the reason for this change is this:
this test runs in browser and in node and on browser side it passes
it fails on node side and it fails because native fetch on node side did not implement yet fetching the resource from the file protocol, so fetch threw an error.
Instead I just read the image directly by using the getAsset we have, and pass it through the loadSVGFromString

Smrtnyk:
seems like change log action had a hiccup, it still added a changelog entry though

e2e/tests/visual-output/rendering/testcases/z-svg-export.ts
asturur: ok!

asturur:
Thank you this was an easy way to remove old code.

asturur (approved):

PR #10693, opened 2025-07-25
chore(): reduce noise in unit tests

Smrtnyk:
Canvas-events.spec.ts has a lot of noise and a lot of non accurate @ts-expect-error directives.
 The issue is that these are not private methods but that the mock of the event is not following the expected interface.
 To reduce the noise I just created a small helper factory to create a mock event instead.
 This removes a bunch of @ts-expect-error directives.
 Also resolves 1 todo in Pattern.spec.ts that passes in a image reference as the pattern source rather than image link itself

### DIFF
--- a/src/canvas/Canvas-events.spec.ts
+++ b/src/canvas/Canvas-events.spec.ts
@@ -6,8 +6,14 @@ import { Group } from '../shapes/Group';
 import { ActiveSelection } from '../shapes/ActiveSelection';
 import { Point } from '../Point';
 import { PencilBrush } from '../brushes/PencilBrush';
-import type { CanvasEvents, ObjectEvents, TPointerEvent } from '../../fabric';
+import type {
+  CanvasEvents,
+  FabricObject,
+  ObjectEvents,
+  TPointerEventInfo,
+} from '../../fabric';
 import { getFabricDocument, IText, version } from '../../fabric';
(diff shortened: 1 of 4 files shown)


Smrtnyk:
I am going a bit through the codebase to search for low hanging fruit we can resolve about types

src/canvas/Canvas-events.spec.ts
```diff
@@ -6,12 +6,40 @@ import { Group } from '../shapes/Group';
...
+    return {
+      clientX: 0,
+      clientY: 0,
+      which: 1,
+      target: canvas.upperCanvasEl,
+      bubbles: true,
+      ...options,
+      ...additionalProps,
+    } as TPointerEvent;
+  }
```
asturur: could this be moved in a test only util file to encourage reuse?

asturur:
this is a great cleanup

src/canvas/Canvas-events.spec.ts
Smrtnyk: sounds good, will move

Smrtnyk: moved

Smrtnyk:
seems like chromium died on browser install part

asturur (approved):

Issue #10698, opened 2025-08-12
[Bug]: The mouse enter and leave events of child elements will be executed twice.


zhe-he:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.6.4

### Node Version (if applicable)

None

### Link To Reproduction

https://codepen.io/zhe-he-the-vuer/pen/qEOOwwY?editors=0110

### Steps To Reproduce

1. Create two rectangles.
2. Create a group and place the two rectangles inside it.
3. Add an event to child element B.
4. Move the mouse over child element B.

### Expected Behavior

once

### Actual Behavior

twice

https://github.com/user-attachments/assets/b5cef440-5c8d-4718-8207-dfdcbea0da19

### Error Message & Stack Trace

```bash

```

PR #10699, opened 2025-08-12
fix: The mouse enter and leave events of child elements will be executed twice. 

zhe-he:
When the event is triggered, this.targets includes the target, causing an extra trigger. So I skipped this one.

https://github.com/fabricjs/fabric.js/issues/10698


asturur:
it totally make sense.
I m not sure if this was a regression or if it was discussed and left like that.

I think there is a test to update

Issue #5710, opened 2019-05-22
Need to add passive:false option for wheel event listener


vamzz1:
## Version
2.4.2-b

## Information about environment
ReactJS: 16.6.3
Google Chrome: Version 74.0.3729.157 (Official Build) (64-bit)

## Steps to reproduce
On rendering the Fabric Canvas I see this error log

[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952

Then I sent addEventOptions {passive: false} as the fourth argument in this line https://github.com/fabricjs/fabric.js/blob/master/dist/fabric.js#L11077 by hard coding the argument in the library and that fixed the issue.

Without the `addEventOptions` options to wheel event listener there's an error log for each of the canvas that is rendered.

vamzz1:
Please let me know if anything else is needed. Help would be really appreciated! I get a feeling I'm missing something really small but I'm stuck on this for the last 2 days. Thanks

asturur:
hi @vamzz1 what you see is a warning log.
I think that warning, it just warns you that if you will put HEAVY js in the mousewheel event, that  generally is for scrolling, the scrolling may be affected in performance.

This happens becase a non passive event ( a norma one ) can cancel the scrolling and so scroll before starting has to wait for the event code to end.

Now we can assume that if you are using the mousewheel on fabricJS canvas you are not scrolling but doing something else, and so this warning is a warning without meaning for you.
Also setting the event as passive will let first the browser consume it, fire a possible `animation frame` and then execute the event, while since that event contains fabricjs code, you probably want to first execute it and then render.

All this to say that i do not think that warning should be considered a problem at all.

zhe-he:
After enabling all levels in the console, I discovered this warning.
I looked into the Chrome Blink source code and found that omitting the third parameter behaves the same as setting passive: false.
The only difference is that when the third parameter is omitted, the program calls UseCounter::Count() for metrics collection.
Therefore, I support the original poster's modification to add passive: false in order to remove the warning.

Below are the code details:

```ts
div.addEventListener("mousewheel", fn)   ->     PassiveMode::kNotPassiveDefault  
div.addEventListener("mousewheel", fn, { passive: false })   ->     PassiveMode::kNotPassive  
div.addEventListener("mousewheel", fn, { passive: true })   ->     PassiveMode::kPassive 
```
```c++
[...]

PR #10712, opened 2025-08-26
chore(): Remove mouse wheel console warning by setting default explicitly.

zhe-he:
#5710 

Even if we don't set passive, its behavior is the same as passive=false. Why not proactively set passive while also removing the violation?

When the third parameter only includes passive, there's no need to add the third parameter when removing the event. This is why I only modified the `addEventListener` but not the `removeEventListener`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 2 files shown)


asturur (approved):

asturur:
@zhe-he thanks for this pr.
If you need it patched on version 6 too, you need to open a PR also on the 6.x branch.

PR #10715, opened 2025-09-01
BREAKING: change default originX and originY to center/center

asturur:
## Description

This change is meant to lessen the burden for when originX and originY are removed.
Is easy to opt out from this change and just move to 7.0 anyway, this PR aims to evaluate how much investigation i need to do to see if something really break in tests.


asturur:
Still 82 tests to fix

asturur:
this is like 75% done

asturur:
Ok all tests are fixed.

PR #10725, opened 2025-09-09
Clarify MIT License

asturur:
Added MIT License text to the LICENSE file.


Clipperone (commented):
The update improves clarity and aligns better with the standard MIT license template, which is important for legal transparency.

Clipperone (approved):

PR #10741, opened 2025-09-19
fix: After executing loadFromJSON, it unexpectedly adds an objects property to the canvas.

zhe-he:
When the user imports the data that was just exported, the canvas gets an extra objects property added, which shouldn’t happen. I removed it.

```ts
(async () => {
...
console.log('objects' in canvas);    // false
const data = canvas.toJSON();
await this.canvas.loadFromJSON(data);
console.log('objects' in canvas);    // true

})();
```

### DIFF
--- a/src/canvas/StaticCanvas.spec.ts
+++ b/src/canvas/StaticCanvas.spec.ts
@@ -1392,7 +1392,7 @@ describe('StaticCanvas', () => {
     expect(canvas.loadFromJSON).toBeTypeOf('function');
 
     await canvas.loadFromJSON(PATH_JSON);
-    expect('objects' in canvas).toBe(false);
+    expect(canvas).not.toHaveProperty('objects');
     const obj = canvas.item(0);
 
     expect(canvas.isEmpty()).toBeFalsy();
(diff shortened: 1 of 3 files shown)


src/canvas/StaticCanvas.spec.ts
```diff
@@ -1392,6 +1392,7 @@ describe('StaticCanvas', () => {
     expect(canvas.loadFromJSON).toBeTypeOf('function');
 
     await canvas.loadFromJSON(PATH_JSON);
+    expect('objects' in canvas).toBe(false);
```
Smrtnyk: you could also do `expect(canvas).not.toHaveProperty('objects')`
not that it makes a difference but kinda more semantic in tests

asturur:
Thank you this is a great fix and probably a regression from trying to speed up and remove some extra spread.
Let me see if i find a different way to fix this, otherwise we merge it

src/canvas/StaticCanvas.spec.ts
asturur: Yes i agree with @Smrtnyk and this semantic he suggested

asturur (approved):

src/canvas/StaticCanvas.spec.ts
zhe-he: I also agree with this modification. I didn't use this method because I'm not familiar with the relevant APIs.

PR #10748, opened 2025-09-21
feat: Add isolatemode of group.

jameszhong2008:
## Description

Hello, since fabric 6, fabric support Group better.  But it can not select object in group on UI.
I add a option 'dblClickIsolateObject'  in CanvasOptions.

Turn on this option by 
```
new fabric.Canvas(ele, {
        dblClickIsolateObject: true,
  }
);
```
When double click on a group, we can select the children of this group.

## In Action
image
image


src/canvas/Canvas.ts
```diff
@@ -84,21 +84,21 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
    * @type number
    * @private
    */
-  declare private _willAddMouseDown: number;
+  private declare _willAddMouseDown: number;
```
Smrtnyk, line 87: this change doesn't seem relevant to the feature you added

Smrtnyk:
was this feature discussed before as acceptable, before you implemented it? is there an issue you can link to?
also you added this feature but not a single test, either unit or visual ones

src/canvas/Canvas.ts
jameszhong2008: Yes, this is not relevant to the feature, it was format automatically when saving. it is default format after typescript 4.0. 
If this doesn't fit, I'll restore it.

jameszhong2008:
> was this feature discussed before as acceptable, before you implemented it? is there an issue you can link to? also you added this feature but not a single test, either unit or visual ones

Our graphic editor based fabric uses a tree structure and needs to modify child nodes directly. So we implemented it, and I think this feature will be useful to others.

Smrtnyk:
> > was this feature discussed before as acceptable, before you implemented it? is there an issue you can link to? also you added this feature but not a single test, either unit or visual ones
> 
> Our graphic editor based fabric uses a tree structure and needs to modify child nodes directly. So we implemented it, and I think this feature will be useful to others.

@asturur has the last word here
but in case he rejects the feature you then just waisted your time implementing this
that is why it is better to first open a feature request or open a discussion

src/canvas/Canvas.ts
Smrtnyk: I would refrain from making any changes that are unrelated to the problem you are trying to solve
just adds extra burden to the reviewers and makes the diff larger

asturur:
We can't insert every feature that every editor wants.
Many features create an exponential number of cases you have to maintain with all permutation and combinations of features, but on top of everything this feature happens on double click you  can implement it as an event handler on the dblclick.
I understand that some people would find this already done and would be happy, but many others could want it slightly different and would be locked into this implementation.
It would be useful if you could turn this code into an event driven feature and make a tutorial out of it on the website, as a feature i can't accept it.

asturur:
The point made by @Smrtnyk is correct. If you want to contribute discuss the feature first and see what are the options to make your feature available to the community. The projects needs documentation above anything else, examples, explanations, tutorials. Adding features is not really convenient at this stage

jameszhong2008:
OK，I'll find a better way. About event driven feature, can you give me more detail. I'm not familiar with it.

asturur:
You added some code to methods.
Could you build the same feature just by using:
`canvas.on('dblclick', () => {})` object.onSelect`, `object.onDeselect`, with the provided hooks from fabricJS that are already in place?

the description of this feature is very poor, for how is described i would imagine that is a simple double click event that either  enable group interactive or manipulate groups

jameszhong2008:
> You added some code to methods. Could you build the same feature just by using: `canvas.on('dblclick', () => {})` object.onSelect`, `object.onDeselect`, with the provided hooks from fabricJS that are already in place?
> 
> the description of this feature is very poor, for how is described i would imagine that is a simple double click event that either enable group interactive or manipulate groups

Yes, you are right, enable double click to select children of group or set group manually. 
I think this feature can't be implement by canvas.on('dblclick', () => {}) ，unless I implement all the mouse down, mouse up event. 
For exemple, we use dblclick to select a child node of group. then we drag the node. fabric will select the whole group again by default. so we can't operate the child node.

asturur:
There is interest in being able to customize features by events, some i m happy to improve the base api if is not enough to obtain this customization without changes to the core library

PR #10767, opened 2025-10-05
chore(): remove fs-extra dev dependency

Smrtnyk:
relates to #10762


Smrtnyk:
CI seems fine, except old known firefox test that is unrelated

.codesandbox/start.mjs
```diff
@@ -37,16 +37,19 @@ export function startSandbox(destination, buildAndWatch, installDeps = false) {
...
         pathToTrigger,
-        {
-          ...packageJSON,
-          trigger: moment().format('YYYY-MM-DD HH:mm:ss'),
-        },
-        { spaces: 2 },
+        JSON.stringify(
+          {
+            ...packageJSON,
+            trigger: moment().format('YYYY-MM-DD HH:mm:ss'),
```
Smrtnyk, line 46: i think we can also get rid of moment in favor of Intl constructor
moment is a huge lib, not worth having it for a trivial calls like this

asturur: if Intl is in node, yes. Not even sure we need moment and not a simple toJSON of Date.
I do not know what this trigger is here for. The codesanbox part was something i really didn't want to take on when it was added.

PR #10770, opened 2025-10-06
chore(): remove moment dev dependency

Smrtnyk:


scripts/date-time.mjs
```diff
@@ -0,0 +1,19 @@
+export function formatFullTimestamp() {
+  return new Intl.DateTimeFormat('sv-SE', {
```
Smrtnyk, line 2: I put sweden here because it outputs the same thing you were doing with moment
if you copy paste the formatter to console you will get "2025-10-06 17:26:54" for example

scripts/buildLock.mjs
Smrtnyk, line 1: I don't get the point of this file, is it used somewhere?
edit: ah ok seems like it is used in codesandbox, which I am not really familiar with

asturur:
thank you @Smrtnyk . let's merge this and then let's verify if we can get out without it entirely. i  can't find references for trigger and codesanbox on internet

asturur (approved):

PR #10773, opened 2025-10-08
chore(): up dev deps

Smrtnyk:
another round of dev dep updates

there seems to be an issue with playwright update, I will have a look at that in a separate pr
there is a new major jsdom version, not sure if you want that updated as well?


asturur:
for jsdom i want to wait a bit.
While node 18 is unsupported now, i don't want to relase 7.0 with immediately broken node 18.
I want most people to move forward without thinking is a big issue.

Smrtnyk:
> for jsdom i want to wait a bit. While node 18 is unsupported now, i don't want to relase 7.0 with immediately broken node 18. I want most people to move forward without thinking is a big issue.

gotcha

PR #10780, opened 2025-10-15
chore(): update playwright

Smrtnyk:
I have opened an issue https://github.com/microsoft/playwright/issues/37855
because I dislike the solution I came up with, and also I dislike the solution that playwright maintainer suggests.
Lets see if they export the type again, or we have to stick with this.
I would not merge this until the playwright issue is closed


e2e/utils/CanvasUtil.ts
```diff
@@ -84,17 +92,17 @@ export class CanvasUtil {
...
   }
 
-  async executeInBrowser<C, R>(
-    runInBrowser: PageFunctionOn<Canvas, C, R>,
-    context: C,
-  ): Promise<R> {
+  private async _executeInBrowserImpl(
+    runInBrowser: any,
+    context?: any,
+  ): Promise<any> {
```
asturur, line 104: yeah i see. This will be hard to remember in a month or two why is in this way. Let me think if some ts notation is a bit better for this.

Smrtnyk: i wrote on the issue I opened in playwright that i dislike this approach they suggested
first I had it with just copying the type manually
but I am not sure which one approach is worse

lets see if they reexport the type again, if not then you will have to decide on how to proceed 
but in any case, just slapping `any` type I think is the least optimal solution

Smrtnyk:
@asturur the playwright issue I filed has been closed without interest to export the type again
so I suggest that we just stick with this, since i don't expect this part of code changing that often
tests are pretty stable and when adding new ones we are not really touching this part of the code

asturur (dismissed):

PR #10785, opened 2025-10-23
fix(): fix rendering of text when line height is set to 0

### LINKED ISSUE #10775: [Bug]: When the line height of the text is set to 0, the text cannot be displayed.
### CheckList

### Version

6.6.4

_No response_

None

### Link To Reproduction

https://codepen.io/zhe-he-the-vuer/pen/azdwQpw?editors=0010

### Steps To Reproduce

1. Create a control group with the text line height set to 0.01.
2. Create an experimental group with the text line height set to 0.
3. Add both texts above to the canvas.

### Expected Behavior

The texts should overlap on the first line.

Image

### Actual Behavior

The text disappeared.

Image

```bash

```

Smrtnyk:
closes #10775


src/shapes/Text/Text.ts
```diff
@@ -1040,7 +1042,12 @@ export class FabricText<
       height = 0;
     for (let i = 0, len = this._textLines.length; i < len; i++) {
       lineHeight = this.getHeightOfLine(i);
-      height += i === len - 1 ? lineHeight / this.lineHeight : lineHeight;
```
Smrtnyk, line 1043: these calculations with 0 where giving NaN so I had to do fallbacks to fontSize
it passes still all unit tests and all visual tests
and you can see the new visual test output I added

zhe-he: It seems there's a small issue — the visual appearance of a line-height set to 0.01 should be almost the same as when it's set to 0, but judging from the image, the latter appears to be shifted a few pixels upward overall.

Smrtnyk: yeah that is what I was not sure about, the first rendering of 0.1 is the current behavior
the rendering of 0 is the new behavior (previously would not even be shown) because I am now falling back to the font size if line height is 0, to avoid having NaN. 
I agree with you, they should be almost the same
I have another idea, will give it a try

Smrtnyk: @zhe-he could you recheck the visual screenshot again?

zhe-he: image

There’s still an issue — it’s probably caused by something that hasn’t been taken into account somewhere.

Smrtnyk: Ah i see what you mean, they should be on the same line

Smrtnyk: hm, what about now? the red and green should be aligned on the bottom, on the left it is just slightly down because of 0.1 value

zhe-he: It looks like everything is fine now.

Smrtnyk: Tnx for the feedback @zhe-he. lets see what @asturur has to say

asturur: Tomorrow i ll check this and the nested clip path, and then move to - finally - the large pr around guidelines for alignemnt.

asturur: The fix is good.
I disagree on the special treatment for the background color that collapse to 0.
It doesn't make sense.
Is one thing to allow you to render a line height that doesn't make sense instead of crashing. Is different to disable the bgcolor because we know is going to overlap and so cover everything but the last line painted.

Bad input bad output imho, is not scope of the library fix it.

I added 2 more tests in the  visual test, to cover extra things.

I will also add another commit to reduce the duplication of the calculation making a private method for the lineheight

Smrtnyk: tnx a lot @asturur

asturur: mmmm So, this solution doesn't take in consideration style for fontsize.
everything is so complicated with text.

Smrtnyk: > mmmm So, this solution doesn't take in consideration style for fontsize. everything is so complicated with text.

does it make sense to go into every tiny detail
I mean this is already better than it was
maybe we can improve it incrementaly as we add more and more tests

asturur: The issue is that for lineHeight = 0 we are assuming that fontSize is the size of the line, while the size of the line is the biggest fontSize in the line

asturur: i have an improvement coming

asturur: > does it make sense to go into every tiny detail
I mean this is already better than it was
maybe we can improve it incrementaly as we add more and more tests

No it doesn't make sense to exagerate.
LineHeight 0 would work without style, but the change in `calcTextHeight` would have caused a bug for the normal text that had style in the last line and that could be fairly common.

In general if reading it you get an idea and you think you can make it better quickly, you do it, otherwise no.
For example the clipPath i have no idea what to do in the other PR seems a mistery to unravel, and the partial fix is fine, i rather spend time in more useful issues than perfecting that.

asturur: At the end we fixed the issue, we added a test, and we removed 15 bytes of code instead of adding 200.
So was win / win.

PR #10805, opened 2025-11-10
fix(Text):Double offset when exporting SVG after setting deltaY in Text

jiao1187875445:
After setting deltaY in the text, the exported SVG has both dy and baseline shift attributes, resulting in inconsistent rendering between SVG and Canvas, leading to double offset. Just keep one of these two

https://github.com/user-attachments/assets/22a2812f-067b-4c95-898c-ff5a3e26930b

### DIFF
--- a/src/shapes/Text/TextSVGExportMixin.ts
+++ b/src/shapes/Text/TextSVGExportMixin.ts
@@ -325,7 +325,6 @@ export class TextSVGExportMixin extends FabricObjectSVGExportMixin {
       fontSize,
       fontStyle,
       fontWeight,
-      deltaY,
       textDecorationThickness,
       linethrough,
       overline,


asturur:
thank you.
It would be great if you could a visual regression test in our playwright tests.
Have a look if you see how to do that, if not i can guide you.

jiao1187875445:
> thank you. It would be great if you could a visual regression test in our playwright tests. Have a look if you see how to do that, if not i can guide you.

Okay, what do you need me to do?

asturur:
if you didn't already do a 'npx playwright install' in the fabric js folder to get playwright.

Then look at this file here:
`e2e/tests/visual-output/rendering/testcases/z-svg-export.ts`
you see is an array of objects, each object represent a test.

a test looks like this

```ts
  {
    title: 'Text deltay export',
    golden: 'text_deltay.png',
    percentage: 0.06,
    size: [200, 200], // <= pick an appropriate size for your canvas depending how large is the object you test, you can go over, but do not exagerate, too many white pixels not a good test
      // you have a canvas and the fabric import in this function, write your test code
      // just create a text object with a DY in the style so obvious that can be seen easily
      // add more than just one negative and one positive, nest it if possible
      const text = new fabric.Text('some sample text', { styles: { .... } });
      canvas.add(text);
      canvas.centerObject(text);
    },
  },
```

^^^ something like this.
Then run `npm run build` and `npm run test:e2e`

Following times is enough to run `npm run test:e2e`

The test should create a golden image.
Add the golden image to git, but do not add the change dist folder, so we don't create useless diffs.
If you commit the dist folder by error , just do `git checkout master dist*`

asturur:
i forgot this PR.
I added a visual test for verification

asturur:
we also have unit tests now.

Issue #10809, opened 2025-11-14
[Feature]: Official integration with mobile interaction libraries


AndreaBogazzi:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have searched and referenced existing issues, feature requests and discussions
- [x] I am filing a **FEATURE** request.

### Description

We need new and updated integrations for mobile actions / gestures.
Pick up one or two modern libraries and prepare in fabric/extensions the necessary code to make that work.

The code will offer an opinionated handlers and a neutral hook.
What does it mean?

It means that for the sake of being functional and explicative, it will register some handlers on gestures like pinch and rotate and pan and tap or longpress.

These will server as a basic working condition and as an example of how to write your handler.
Handlers won't be configurable, for example we won't implement a boolean to hide the rotate control if the rotate action is connnected or a boolean to scale the active object instead of the viewport with pinch to zoom.

We are going to pick up one implementation and you can just delete parts that you don't use and add yours.

This will hopefully both keep complexity for maintainers low and a decent learning curve for who wants to customize.

Planned event for the first PR
- pinch to zoom
- rotate

Then
- pan
- tap ( double tap, triple tap )
- longpress

### Current State

the old gestures integration is deprecated and not in working status

asturur:
For anyone interested this is being worked on.
@jiayihu FYI since i know you had some issues with taps and current code

relates to
#10171
#6599
#1207
(and probably more)

PR #10813, opened 2025-11-23
feat(): Multi touch gesture support with module westures

### LINKED ISSUE #10790: [Bug]: Double and triple click don't work on touch devices
### CheckList

### Version

6.6.4

_No response_

None

### Link To Reproduction

Sandbox templates are wrong. `Next template` fails to install, `Vanilla template >= v6` actually uses fabric v3 and fails to install as well.

### Steps To Reproduce

1. Create a Textbox
2. Double or triple tap on touch device

### Expected Behavior

Select word or text

### Actual Behavior

No selection.

https://github.com/fabricjs/fabric.js/pull/10478 doesn't work on touch devices because `event.detail` is always 1 in my testing and there is no `dblclick` event. https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event#result doesn't work if you simulate a touch device with devtools for instance.

So basically the old IText double/triple click implementation needs to be restored for `_onClick`. Will open a PR if okay.

asturur:
## Description

This PR adds support for pinch and rotate with the library westures, still an abandoned one, but way newer and modular compared to hammer.js

The extension module exports an add gesture function that will make the canvas fire synthetic events 'rotate' and 'pinch'.
Those are standard events and carry with them a rotation and a scale extra property in the argument passed at the handler.

You can write your custom handler or you can use the one i provide in the folder as examples.
The rotate one will only rotate the active object if you do a rotation that has the center inside the object.
The pinch handler will either modify the viewport with zoomToPoint or scale the currently selected object if is the target of the action, so if the center of pinch falls inside the object bounding box.

For what regards the missing double and triple click fired from the tap events, i couldn't find a decent implementation that worked with westures. I added a setTimeout loop to the click event that should replace the old code i deleted from IText.
The click timing is configurable and seems to work just fine.
@jiayihu if you want to look at the `_onClick` method changes FYI.

Missing correct types for event handlers and some test.

Will follow similar integration with the MUI internal library for gestures

closes #10790


asturur:
My double and triple click implementation is definetely too naive, i didn't think of many things, i need to make double and triple tap gestures reliably.

src/canvas/Canvas.ts
```diff
@@ -541,14 +544,50 @@ export class Canvas extends SelectableCanvas implements CanvasOptions {
    * @param {Event} e Event object fired on mousedown
    */
   private _onClick(e: TPointerEvent) {
-    const clicks = e.detail;
+    const clicks = (this._clickCount += 1);
```
jiayihu: Two things I've done personally are:
1. Skip native `dblclick` event otherwise the event is still gonna be fired as well
2. Check pointer position delta between clicks, as double and triple clicks should happen over the same point basically

extensions/westures_integration/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    ({ rotation, event }: RotateEventData) => {
+      canvas.fireEventFromPointerEvent(event, 'rotate', { rotation });
+    },
+  );
+};
+
+/**
+ * Add a serie of gestures recognition on the canvas
+ */
+export const addGestures = (canvas: Canvas) => {
```
jiayihu, line 115: Nice I like it's an opt-in feature 👍

asturur:
the tap and double tap gesture out of the box work OK.
they are not perfect meaning that if you tap 4 times quickly you are going to fire a double, a triple and then a double again. because the listener are not connected.
Honestly i would like to write a tap gesture with the core module that can handle this, that can hold the double tap to understand if the triple is happening, it seems more logic to me.
Is not how click and dblclick works tho.

asturur:
close #10809

asturur:
ok the inclusion of westures crash setupApp in the e2e tests.
Tomorrow i should be able to fix it.

e2e/imports.ts
```diff
@@ -9,7 +9,15 @@ import { readFileSync } from 'node:fs';
...
+const westuresWrapper = `data:text/javascript,${encodeURIComponent(`
+  const westures = window.westures || globalThis.westures;
+  export default westures;
+  export const { Pan, Pinch, Press, Pull, Rotate, Swipe, Swivel, Tap, Track, Region, Gesture, PointerData, State, Input, Point2D, CANCEL, END, MOVE, START } = westures;
+`)}`;
+
 export default {
   fabric: packageJSON.module.slice(1),
   ['fabric/extensions']: packageJSON.exports['./extensions'].import.slice(1),
+  westures: westuresWrapper,
```
asturur, line 22: @Smrtnyk FYI the changes in this file and in setupApp.ts  were pulled of Claude 4.5 sonnet, i m not sure i would have got out of this import issue by myself just using free time to investigate it.

westure does not have browser build, so i couldn't add it to the e2e setup and tests couldn't load.

Smrtnyk: oh interesting solution

Smrtnyk:
I see you tried a playwright update
i was thinking to do today another round of deps update with newer vitest, playwright etc..

Smrtnyk:
its a bit sad seeing these gestures libs abandoned
I used hammerjs before with fabric but yeah it is not touched for a very long time
nice to see this feature land in fabric

asturur:
yeah the playwright update breaks the already flaky drag test. it changes the browser from chromium to chrome for testing ( i wonder what is inside there, maybe they just want to erase the chromium logo/knowledge or maybe there is all the google stack for analytics and timing reports, all in all not a great change )

asturur:
> its a bit sad seeing these gestures libs abandoned I used hammerjs before with fabric but yeah it is not touched for a very long time nice to see this feature land in fabric

I think is partially because those libs works ok. They do not need much changes or more features, their docs are ususally enough for correct usage, so there is little issues and life around the lib, which is good.
Yes typescript would be helpful for this lib for example.

mvanderkamp:
Hey folks I just got wind of this, that's awesome you're using the library. I didn't know anyone was interested so I haven't been doing anything. Definitely open issues or pull requests if there are features or fixes you want!

asturur:
hi @mvanderkamp i discovered Westures as the less abandoned library of those. So i jumped on it.
It works fine imho.

I had some issue with clicks and taps, but i m not sure that is solvable meaning that a dbl tap was also firing a dbl click and i didn't really have a way to fix that if not removing the click event.

Another thing i noticed i wasn't able to get a double tap to wait to trigger to see if a triple tap was executed. But i m not even sure that is a behavior someone should rely on.
I just forward the events to the devs that use fabricJS, i don't build real features with them.

Issue #10815, opened 2025-11-30
[Bug]: 【couple】 Calculates the case where "undefined + char" exists


ccOfHome:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.6.4

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

None

### Link To Reproduction

none

### Steps To Reproduce

Hello, I am a front-end developer and I have just entered the open-source community. Currently, I am working on developing a graphic editor. During the development process, I discovered some issues with the Text object in fabric.js. Therefore, I am attempting to make some contributions to the open-source community. If possible, I would like to try fixing this problem.

This can be replicated by performing the operation in any text box. The steps for replication are as follows

1. In the `_measureChar` method of `fabric.js/src/shapes/Text/Text.ts`, a `debugger` statement is inserted when calculating the `couple`.
2. Select any text box for operation and observe the changes in the value of "couple"

### Expected Behavior

`couple = previousChar + _char` shoude be `couple = previousChar ? previousChar + _char : char`

### Actual Behavior

<img width="1910" height="668" alt="Image" src="https://github.com/user-attachments/assets/81aceee5-565d-4a9a-a4f4-88aa09c134f2" />

### Error Message & Stack Trace

```bash

```

ccOfHome:
Oh, sorry. The content of "Expected Behavior" should be "`couple = previousChar + _char` should be `couple = previousChar ?  previousChar + _char : _char`"
Oh, by the way, "couple = previousChar + _char" will not cause any calculation problems, but it can be optimized to avoid misunderstandings during the development process.

asturur:
Hi @ccOfHome you are welcome to try to fix it.
Please fix it in the master branch opening a pr, and then if you need it published in the version 6.x ( 6.9.1 ) you will need to backport the fix to the 6.x branch too.

PR #10816, opened 2025-12-01
fix(): Fix the situation where undefined + char exists when calculating couple

ccOfHome:
## Description
`couple = previousChar + _char` will not cause any calculation problems, but it can be optimized to avoid misunderstandings during the development process
#10815

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
-- fix(): Fix the situation where undefined + char exists when calculating coupleWidth [#10816](https://github.com/fabricjs/fabric.js/pull/10816)
+- fix(): Fix the situation where undefined + char exists when calculating couple [#10816](https://github.com/fabricjs/fabric.js/pull/10816)
 - feat(): Add configuration parameter for patternQuality in node [#10804](https://github.com/fabricjs/fabric.js/pull/10804)
 - fix(): BREAKING Fix text positioning [#10803](https://github.com/fabricjs/fabric.js/pull/10803)
 - fix(AligningGuidelines): Guidelines features updates [#10120] (https://github.com/fabricjs/fabric.js/pull/10120)


Smrtnyk:
Is it possible to add any meaningful tests for this?
It clearly must be fixing some scenario

ccOfHome:
Sure, but I'm not very familiar with this aspect.
Please check if the following code is okay

> e2e/tests/visual-output/rendering/testcases/text.ts

```
{
    title: 'Text with edge case: calculating coupleWidth, undefined + char',
    golden: 'textCalculatingCoupleWidth.png',
    percentage: 0.02,
    disabled: 'node',
    size: [200, 200],
      const t1 = new fabric.FabricText('AB AB\nBA BA', {
        left: 10,
        top: 10,
        fontFamily: 'Arial',
        fontSize: 64,
        objectCaching: false,
        strokeWidth: 0,
      });
      t1.positionByLeftTop(new fabric.Point(10, 10));

      const t2 = new fabric.FabricText('ABA', {
        left: 10,
        top: 110,
        fontFamily: 'Arial',
        fontSize: 64,
        objectCaching: false,
        strokeWidth: 0,
      });
      t2.positionByLeftTop(new fabric.Point(10, 110));

      canvas.add(t1, t2);
    },
  },
```

asturur:
For testing look at the file `Text.spec.ts` sibling of the file you just fixed.

look for this test here

```ts
  describe('measuring, splitting', () => {
    it('measuring', () => {
      cache.clearFontCache();
      const zwc = '\u200b';
      const text = new FabricText('');
      const style = text.getCompleteStyleDeclaration(0, 0);
      const measurement = text._measureChar('a', style, zwc, style);
      expect(measurement).toMatchSnapshot(roundSnapshotOptions);
      expect(measurement).toEqual(text._measureChar('a', style, zwc, style));
    });

    it('splits into lines', () => {
      const text = new FabricText('test foo bar-baz\nqux');
      expect(text._splitTextIntoLines(text.text)).toMatchSnapshot();
    });
  });
```

double the measuring test, change its title to measuring a single char, and modify the test and expectations. ( remove the zwc for example )

you want to test 2 things:

_measureChar returns a different ( or not ) value after your change.

cache.charWidthsCache has only a key for the char you measured ( example 'a' ) and not also for `undefineda` couple of chars.

a unit test is sufficient.

To run the test write `npm run test:vitest Text.spec.ts` and you should be able to quickly run the unit tests for that file.

If you are not familiar with unit tests, this is a valuable thing to know.

src/shapes/Text/Text.spec.ts
```diff
@@ -1320,4 +1320,21 @@ describe('FabricText', () => {
...
+      const text = new FabricText('');
+      const style = text.getCompleteStyleDeclaration(0, 0);
+      const measurement1 = text._measureChar('a', style, '', style);
+      const measurement2 = text._measureChar('a', style, '', style);
+      expect(measurement1).toEqual(measurement2);
+      const cacheKeys = cache.charWidthsCache
+        .get(text.fontFamily.toLowerCase())
+        ?.get('normal_normal')
+        ?.keys();
+      expect(cacheKeys?.next().value).not.toBe('undefineda');
```
Smrtnyk, line 1336: `undefineda` is this a typo? did you mean `toBeDefined()` instead?

ccOfHome: `undefineda` might appear as a key, so I used strings for comparison. This might not be the best code; do you have any suggestions?

Smrtnyk: ah sorry, I got confused, you are right
I didn't get the test at first

asturur:
Thankyou @ccOfHome  and @Smrtnyk

PR #10830, opened 2025-12-30
Cleanup animation delay timeout

jiayihu:
Traced a weird issue with Vitest on CI down to this possible cause: after the test file has finished, if the test contained some text objects in editing mode, there could still be the `timeout` pending.
However since the test run has finished, Vitest probably has already removed all the mocks, test isolation etc. so it's as if `requestAnimFrame(firstTick)` runs directly on the Vitest process, instead of running in the usual test isolated context. This causes `requestAnimationFrame` to raise an exception because it's not available.

Vitest log:

```
⎯⎯⎯⎯⎯⎯ Unhandled Errors ⎯⎯⎯⎯⎯⎯

Vitest caught 1 unhandled error during the test run.
This might cause false positive tests. Resolve unhandled errors to make sure your tests are not affected.

⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯
TypeError: getFabricWindow(...).requestAnimationFrame is not a function
 ❯ requestAnimFrame node_modules/fabric/dist/index.mjs:1525:28
 ❯ Timeout._onTimeout node_modules/fabric/dist/index.mjs:5695:24
 ❯ listOnTimeout node:internal/timers:588:17
 ❯ processTimers node:internal/timers:523:7

This error originated in "src/studio/components/extensions/transform-sync/CanvasTransformSync.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
```

As you can see, this exception did not originate inside the test itself, rather just running the file code itself.

### DIFF
--- a/src/util/animation/AnimationBase.ts
+++ b/src/util/animation/AnimationBase.ts
@@ -54,7 +54,7 @@ export abstract class AnimationBase<
    */
   declare private startTime: number;
 
-  private timeout: number | null;
+  declare private timeout: number | null;
 
   constructor({
     startValue,


Smrtnyk:
Is it possible to add test?

asturur (dismissed):

asturur:
is it possible @jiayihu to share your custom test that makes this fail? Or write one that is similar

CHANGELOG.md
```diff
@@ -30,6 +31,7 @@
 - fix(AligningGuidelines): Guidelines features updates [#10120] (https://github.com/fabricjs/fabric.js/pull/10120)
 - chore(deps-dev): bump inquirer from 12.9.6 to 12.10.0 [#10789](https://github.com/fabricjs/fabric.js/pull/10789)
 - chore(deps-dev): bump @types/micromatch from 4.0.9 to 4.0.10 [#10788](https://github.com/fabricjs/fabric.js/pull/10788)
+- fix(): Cleanup animation delay timeout [#10830](https://github.com/fabricjs/fabric.js/pull/10830)
```
Smrtnyk: this one is redundant now

asturur:
The playwright failure is my fault and has nothing to do with this pr

PR #10831, opened 2025-12-30
Correctly check for cache key equality in calcOwnMatrix

jiayihu:
Was surprisingly unnoticed in https://github.com/fabricjs/fabric.js/pull/9887

### DIFF
--- a/src/shapes/Object/ObjectGeometry.spec.ts
+++ b/src/shapes/Object/ObjectGeometry.spec.ts
@@ -692,6 +692,21 @@ describe('fabric.ObjectGeometry', () => {
     ]);
   });
 
+  it('calcOwnMatrix is cached', () => {
+    const cObj = new FabricObject({
+      width: 10,
+      height: 15,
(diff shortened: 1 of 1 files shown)


Smrtnyk:
Is this change testable?

asturur:
So cache was broken for more than one year and no one complained.
I wonder if this caching code should just go.
Removing the caching code is still a speed improvement compared to computing the array and not using it properly.

asturur:
Since removing it entirely is as usual a discussion that spans across days, i think merging the fix is good.
As @Smrtnyk points out, if we can add a unit test that breaks with a regression we should do it.
If you don't get to it, i ll try later.

asturur:
test added

jiayihu:
Sorry I was away for a few days (Napoli 😄) but thanks for adding the test. Regarding removing the caching, in my testing it was still worth it because you can customise the cache key properties to use only the ones that are relevant for your project. Doing so significantly improves the caching key performance.

PR #10825, opened 2025-12-23
feat(): Cropping controls extension

asturur:
## Description

Provides handlers functions for cropping and a control set to coordinate them.
Provides also a function to enter and exit crop mode for simplicity of use.

This functions implements a particular version of cropping and are not configurable, but can be used as an example to build something else.

When the image enters crop mode the image can't be moved but it slides inside the crop window.

Still missing:
- ~control apparences~
- ~dragging constrains~
- ~matrix rotation for when the image is not axis aligned~
- extra controls for scaling during cropping. ---> another PR because requires deeper changes

Gave controls their own look

image

## Notice

This PR is also an example of how different events and control properites can be used to create a custom experience.
We did just a small chane to the fabricJS api, to add cropX and cropY on the original transform properties, so that we can keep track better of crop changes.
This could have been achieved also withuot doing that and adding an addtional mouse up/mousedown handlers on the image to keep track of the original crop values.
In this way is cleaner and definitely simpler to understand what is going on.

The enterCropMode function provides a nice example on how to enter and exit a different experience, without keeping track of many things on temporary state.

## Follows up

- e2e test with snapshot in playwright
- additional render effects during cropping ( ghosted below image for example )
- additional scale controls

closes #10824

PR #10845, opened 2026-01-15
feat(extensions): Add demo.html files for extensions

mauricekindermann:
## Summary

Adds simple `demo.html` files to extensions that can be downloaded and opened directly in a browser - no build tools required.

## Why

I've been working on contributing to the cropping extension, but found it difficult to quickly test and understand what's already built without setting up a full dev environment. 

For devs who aren't JS-first (I'm mainly a PHP/Laravel dev), the friction between "knowing extensions exist" and "actually seeing them work" is significant. These demos let anyone validate what's built in seconds. (download file > open in browser > play)

Website demos would be ideal long-term, I can even imagine a pretty feature rich multi-extension playground, but downloadable HTML files are better than nothing right now, require no maintenance, and are still useful for quickly previewing individual extensions (no git/install/build required). 

## Changes

- `extensions/README.MD` - Added Demos section
- `extensions/aligning_guidelines/demo.html` - New
- `extensions/cropping_controls/demo.html` - New

## Next

If this direction is welcome, I'd like to continue contributing to the extensions (already discussed some ideas in #10825).

### DIFF
--- a/extensions/README.MD
+++ b/extensions/README.MD
@@ -3,6 +3,8 @@
 Extensions and/or plugins are all those features that do no have a place in the library itself.
 FabricJS is a rendering library with an interactive layer and serialization support and has already a lot of features and needs to avoid more scope creep.
 
+## What are extensions?
+
 Extension are all those pieces of codes that are more features and less api, as examples:
 
 - filters with a niche use case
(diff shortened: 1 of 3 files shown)


asturur:
For the demos, is better to use the vanilla template and the test cases.
They work with auto rebuild and also the script tag version of fabricJS won't work forever because is becoming too large in size at some point will need to be split in sub packages.

mauricekindermann:
That makes sense, and I now understand what you mean by the codesandbox forking on fabricjs.com. I will rethink.

PR #10846, opened 2026-01-15
feat(extensions): Migrate centering_guidelines from lib/ to extension

mauricekindermann:
## Summary

I'd like my first "real" PR to be something simple - migrating old `/lib/` code into a proper TypeScript extension with tests and a demo anyone can quickly try.

## Changes

- **New:** `extensions/centering_guidelines/` - TypeScript class with tests
- **New:** `extensions/centering_guidelines/demo.html` - Browser demo  
- **Deleted:** `lib/centering_guidelines.js` - Old version

## What it does

Shows guidelines and snaps objects when dragged to the canvas center (horizontal/vertical crosshairs).

## Testing

- 5 unit tests passing
- Tested `demo.html` with local dist build to verify it works


asturur:
Aren't those already part of the guidelines that have been rencently contruibuted?

asturur:
For the demos in general, let me think of it. I m not sold on this index.html in the extensions for a series of reasons.
Mainly is that i would like the samples to be on fabricjs.com, or in the codesandbox vanilla folder because are useful for local development purposes.

mauricekindermann:
That makes sense. Ok let me rethink it, I will figure out a nice way to showoff extensions on fabricjs.com directly instead. I personally love a simple instant working index.html demo, much of the fabrics online presense is jsfiddle/codepen examples like it, but I totally get your thinking. 

Should this be an extension? I originally thought no like you, in fact I was going to PR removing all of /lib/ as it's all ancient stuff. But then I realized for some using a static canvas (no panning) this is useful in some cases. But is it super valuable? Not really. 

My long term idea/goal is to create a nice set of fabric extensions that can work together, and create a psuedo design-tool on fabricjs.com showing how they work, and allowing you to toggle them on/off. Less complex, and more maintanble than the old KitchenSink. More of a 'look how easily you build these kinds features', which I imagine is your goal with extensions.

asturur:
i m not sure the centering guidelines are part of the align guidelines, mine was a question with no implicit answer. Let's keep this pr open for the centering guidelines, and let's find a suitable format for one topic doc pages on fabricjs.com

mauricekindermann:
No problems, I'll have a think and play around on fabricjs.com and see what I can make work. The only tricky bit for me is being unfarmiliar in how the auto code sanbox stuff works, and how to make it work with extensions. I can imagine a few ways, so I'll try make one that feels right.

PR #10848, opened 2026-01-17
fix(): Crop controls respect flipX/flipY

mauricekindermann:
Crop edge controls were inverted on flipped images. When `flipX` is true, the left edge visually becomes the right edge, so the handler needs to swap.

**Changes:**
- `withFlip` helper swaps handlers based on flip state
- Pan handler accounts for flip direction
- Tests updated to check function exists (handlers are now wrapped)

**Test:** Flip an image, enter crop mode, verify controls work correctly.

### DIFF
--- a/extensions/cropping_controls/croppingHandlers.spec.ts
+++ b/extensions/cropping_controls/croppingHandlers.spec.ts
@@ -38,6 +38,8 @@ describe('croppingHandlers', () => {
       cropY: number;
       elementWidth: number;
       elementHeight: number;
+      flipX: boolean;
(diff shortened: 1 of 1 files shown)


mauricekindermann:
I've also built edge resize controls (mle/mre/mte/mbe that resize within crop bounds then switch to uniform scaling) on top of this branch - I will PR separately once this merges. Or I can push it all as one if you prefer, I don't mind either way.

mauricekindermann:
Superseded by upcoming comprehensive cropping extension PR

PR #10851, opened 2026-01-18
feat(extensions): Cropping controls with edge resize and flip support

mauricekindermann:
PR is a WIP but nearly there!

**FlippedImage Support**
It took me a while to figure out how to fix the crop controls when images are flipped. I had to reflect the pointer position and swap handlers. This works for standalone images but may not handle nested/flipped groups correctly - a proper solution would need scene-space positioning (otherwise grouping/flipped groups will cause inversion issues).

Now... what I did:
- [x] "Bouncy" edge resize controls (crop inward, uniform scale when bounds exhausted)
- [x] Flip-aware crop controls (withFlip wrapper swaps handlers correctly)
- [x] Unified the controls
- [x] Unit tests for flip handling
- [x] E2E tests with snapshots (I did my best here, not very familiar)

Last things I'd like to tackle
- [x] Fix edge  Resize flipped + rotated (just noticed)
- [ ] Reposition via ghostImage (better UX)
- [x] Fix exitCropMode interaction (UX should be imageDeselected)
- [ ] Handle interactions in scene space to fix local space issues when grouped/flipped
- [ ] Rotate ghost in crop mode bound by clip window (hardest interaction by far, might skip, I never solved this one)
- [x] ESC / ENTER keys (rule out trackpad double click inconsistency, and better UX)

### DIFF
--- a/src/EventTypeDefs.ts
+++ b/src/EventTypeDefs.ts
@@ -108,8 +108,9 @@ interface TEventWithTarget<E extends Event = TPointerEvent> extends TEvent<E> {
   target: FabricObject;
 }
 
-export interface BasicTransformEvent<E extends Event = TPointerEvent>
-  extends TEvent<E> {
+export interface BasicTransformEvent<
+  E extends Event = TPointerEvent,
+> extends TEvent<E> {
   transform: Transform;
   /* This pointer is usually a scenePoint. It isn't in the case of actions inside groups,
    * where it becomes a point relative to the group center
@@ -163,8 +164,9 @@ type CanvasModificationEvents = {
   'object:modified': ModifiedEvent;
 };
 
-export interface TPointerEventInfo<E extends TPointerEvent = TPointerEvent>
-  extends TEvent<E> {
+export interface TPointerEventInfo<
+  E extends TPointerEvent = TPointerEvent,
+> extends TEvent<E> {
   target?: FabricObject;
   subTargets?: FabricObject[];
   transform?: Transform | null;
@@ -172,8 +174,9 @@ export interface TPointerEventInfo<E extends TPointerEvent = TPointerEvent>
   viewportPoint: Point;
 }
 
-interface SimpleEventHandler<T extends Event = TPointerEvent>
-  extends TEvent<T> {
+interface SimpleEventHandler<
+  T extends Event = TPointerEvent,
+> extends TEvent<T> {
   target?: FabricObject;
   subTargets: FabricObject[];
 }
@@ -288,10 +291,7 @@ export interface MiscEvents {
 }
 
 export interface ObjectEvents
-  extends ObjectPointerEvents,
-    DnDEvents,
-    MiscEvents,
-    ObjectModificationEvents {
+  extends ObjectPointerEvents, DnDEvents, MiscEvents, ObjectModificationEvents {
   // selection
   selected: Partial<TEvent> & {
     target: FabricObject;
@@ -320,7 +320,8 @@ export interface StaticCanvasEvents extends CollectionEvents {
 }
 
 export interface CanvasEvents
-  extends StaticCanvasEvents,
+  extends
+    StaticCanvasEvents,
     CanvasPointerEvents,
     CanvasDnDEvents,
     MiscEvents,
--- a/src/canvas/CanvasOptions.ts
+++ b/src/canvas/CanvasOptions.ts
@@ -227,7 +227,8 @@ export interface CanvasEventsOptions {
 }
 
 export interface CanvasOptions
-  extends StaticCanvasOptions,
+  extends
+    StaticCanvasOptions,
     CanvasTransformOptions,
     CanvasSelectionOptions,
     CanvasCursorOptions,
--- a/src/canvas/StaticCanvas.ts
+++ b/src/canvas/StaticCanvas.ts
@@ -87,9 +87,9 @@ export type PatternQuality = 'fast' | 'good' | 'best' | 'nearest' | 'bilinear';
  */
 // TODO: fix `EventSpec` inheritance https://github.com/microsoft/TypeScript/issues/26154#issuecomment-1366616260
 export class StaticCanvas<
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    EventSpec extends StaticCanvasEvents = StaticCanvasEvents,
-  >
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  EventSpec extends StaticCanvasEvents = StaticCanvasEvents,
+>
   extends createCollectionMixin(CommonMethods<CanvasEvents>)
   implements StaticCanvasOptions
 {
--- a/src/canvas/StaticCanvasOptions.ts
+++ b/src/canvas/StaticCanvasOptions.ts
@@ -113,9 +113,7 @@ export interface CanvasExportOptions {
 }
 
 export interface StaticCanvasOptions
-  extends CanvasDrawableOptions,
-    CanvasRenderingOptions,
-    CanvasExportOptions {
+  extends CanvasDrawableOptions, CanvasRenderingOptions, CanvasExportOptions {
   /**
    * Width in virtual/logical pixels of the canvas.
    * The canvas can be larger than width if retina scaling is active
--- a/src/shapes/Circle.ts
+++ b/src/shapes/Circle.ts
@@ -44,8 +44,7 @@ interface UniqueCircleProps {
 }
 
 export interface SerializedCircleProps
-  extends SerializedObjectProps,
-    UniqueCircleProps {}
+  extends SerializedObjectProps, UniqueCircleProps {}
 
 export interface CircleProps extends FabricObjectProps, UniqueCircleProps {}
 
@@ -64,10 +63,10 @@ export const circleDefaultValues: Partial<TClassProperties<Circle>> = {
 };
 
 export class Circle<
-    Props extends TOptions<CircleProps> = Partial<CircleProps>,
-    SProps extends SerializedCircleProps = SerializedCircleProps,
-    EventSpec extends ObjectEvents = ObjectEvents,
-  >
+  Props extends TOptions<CircleProps> = Partial<CircleProps>,
+  SProps extends SerializedCircleProps = SerializedCircleProps,
+  EventSpec extends ObjectEvents = ObjectEvents,
+>
   extends FabricObject<Props, SProps, EventSpec>
   implements UniqueCircleProps
 {
--- a/src/shapes/Ellipse.ts
+++ b/src/shapes/Ellipse.ts
@@ -20,18 +20,17 @@ interface UniqueEllipseProps {
 }
 
 export interface SerializedEllipseProps
-  extends SerializedObjectProps,
-    UniqueEllipseProps {}
+  extends SerializedObjectProps, UniqueEllipseProps {}
 
 export interface EllipseProps extends FabricObjectProps, UniqueEllipseProps {}
 
 const ELLIPSE_PROPS = ['rx', 'ry'] as const;
 
 export class Ellipse<
-    Props extends TOptions<EllipseProps> = Partial<EllipseProps>,
-    SProps extends SerializedEllipseProps = SerializedEllipseProps,
-    EventSpec extends ObjectEvents = ObjectEvents,
-  >
+  Props extends TOptions<EllipseProps> = Partial<EllipseProps>,
+  SProps extends SerializedEllipseProps = SerializedEllipseProps,
+  EventSpec extends ObjectEvents = ObjectEvents,
+>
   extends FabricObject<Props, SProps, EventSpec>
   implements EllipseProps
 {
--- a/src/shapes/Group.ts
+++ b/src/shapes/Group.ts
@@ -58,8 +58,7 @@ export interface GroupOwnProps {
 }
 
 export interface SerializedGroupProps
-  extends SerializedObjectProps,
-    GroupOwnProps {
+  extends SerializedObjectProps, GroupOwnProps {
   objects: SerializedObjectProps[];
   layoutManager: SerializedLayoutManager;
 }
--- a/src/shapes/IText/IText.ts
+++ b/src/shapes/IText/IText.ts
@@ -72,8 +72,7 @@ interface UniqueITextProps {
 }
 
 export interface SerializedITextProps
-  extends SerializedTextProps,
-    UniqueITextProps {}
+  extends SerializedTextProps, UniqueITextProps {}
 
 export interface ITextProps extends TextProps, UniqueITextProps {}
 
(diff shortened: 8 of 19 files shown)


asturur:
I ll start a review to understand all changes.
There may be a back and forth of questions, but i see there are tests and everything so we are on the good road

e2e/tests/controls/cropping/index.spec.ts-snapshots/01-cropping-controls-loaded.png
asturur, line 1: This change is probably accidental, or you have to change the name of your first snapshot of your test.

extensions/cropping_controls/controlRendering.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (stroke) {
+    ctx.lineWidth = thickness;
+    ctx.beginPath();
+    ctx.moveTo(-halfLength, 0);
+    ctx.lineTo(halfLength, 0);
+    ctx.stroke();
+  }
+
+  ctx.strokeStyle = ctx.fillStyle;
+  ctx.lineWidth = stroke ? thickness - 4 : thickness;
```
asturur: this is a pattern i want to deprecate.
The pattern is that stroke was added later in life of controls, while before we had just fill or transparent corners.
So now we want only fill or fill and stroke to always look the same.

For me is better than the developer knows that if he doesn't want to use the stroke, needs to have a thickness that is bigger in number rather than the code assuming it has to look the same.

extensions/cropping_controls/controlRendering.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Render a line control for middle edge handles.
+ * Uses Control.angle for orientation, cornerStrokeColor outline and cornerColor fill.
+ * @param {CanvasRenderingContext2D} ctx context to render on
+ * @param {Number} left x coordinate where the control center should be
+ * @param {Number} top y coordinate where the control center should be
+ * @param {Object} styleOverride override for FabricObject controls style
+ * @param {FabricObject} fabricObject the fabric object for which we are rendering controls
+ */
+export function renderEdgeControl(
```
asturur: Can we call this renderRoundySegmentControl rather than edge?
Since is a render let's name it with what it looks like rather than what its control does.
This render function is usable outside cropping too.

extensions/cropping_controls/controlRendering.ts
```diff
@@ -0,0 +1,114 @@
...
+    halfLength = length / 2;
+
+  ctx.rotate(degreesToRadians(this.angle));
+  ctx.lineCap = 'round';
+
+  if (stroke) {
+    ctx.lineWidth = thickness;
+    ctx.beginPath();
+    ctx.moveTo(-halfLength, 0);
+    ctx.lineTo(halfLength, 0);
```
asturur: you can execute this outside the if, once.

```ts
ctx.beginPath();
ctx.moveTo(-halfLength, 0);
ctx.lineTo(halfLength, 0);
if (stroke) {
    ctx.lineWidth = thickness;
    ctx.stroke();
}
ctx.strokeStyle = ctx.fillStyle;
ctx.lineWidth = stroke ? thickness - 4 : thickness;
ctx.stroke();
ctx.restore();
```

it is going to be executed at least once, but there is never reason to execute it twice.

extensions/cropping_controls/controlRendering.ts
asturur: 4 can't be hardcoded, there is a parameter for that and is `fabricObject.borderScaleFactor` in this case multiplied by 2 i think

extensions/cropping_controls/controlRendering.ts
```diff
@@ -0,0 +1,114 @@
...
+  ctx.moveTo(0, 0);
+  ctx.lineTo(length, 0);
+  ctx.stroke();
+  ctx.beginPath();
+  ctx.moveTo(0, 0);
+  ctx.lineTo(0, length);
+  ctx.stroke();
+
+  ctx.restore();
+}
```
asturur, line 170: This has been refactored and can be discussed, but it also changed how the render looks like from square to roundy. There is no reason for that, this code is here to be used, and if developers want roundy corners they can just take this code and use ti as a sample for theyr own implementation.

Aside that, is ok to have less render commands, but this is adding extra moveTo(0,0) and is doing 4 beginPath instead of 1.

Let's leave renderCornerControl as is in the original PR and let's focus on the new controls. Let's not mix new features and changes to rendering and UX all in the same pr, makes the review unnecessary complex.

extensions/cropping_controls/croppingHandlers.ts
```diff
@@ -11,6 +11,22 @@ import { controlsUtils, Point, util } from 'fabric';
...
+  flippedHandler: TransformActionHandler,
+  axis: 'flipX' | 'flipY',
+): TransformActionHandler => {
+  return (eventData, transform, x, y) => {
+    if (transform.target[axis]) {
+      return flippedHandler(eventData, transform, x, y);
+    }
+    return handler(eventData, transform, x, y);
+  };
+};
```
asturur, line 28: this is good, leaves things reusable and contained in scope.

extensions/cropping_controls/croppingHandlers.ts
```diff
@@ -205,6 +223,25 @@ export function ghostScalePositionHandler(
...
+): { x: number; y: number } => {
+  if (!target.flipX && !target.flipY) {
+    return { x, y };
+  }
+  const center = target.getCenterPoint();
+  return {
+    x: target.flipX ? 2 * center.x - x : x,
+    y: target.flipY ? 2 * center.y - y : y,
+  };
+};
```
asturur: for now this is ok, i need to understand if this utility can be used to compensate some strange math we do in other flipped situations and make it part of controls Utils more than specific for cropping. For this PR is ok as is.

extensions/cropping_controls/croppingHandlers.ts
```diff
@@ -218,21 +255,28 @@ export const scaleEquallyCropGenerator =
...
     const anchorOriginX =
-      cx < 0 ? 1 + remainderX / target.width : -target.cropX / target.width;
+      cx < 0
+        ? 1 + remainderX / target.width
+        : -target.cropX / target.width;
     const anchorOriginY =
-      cy < 0 ? 1 + remainderY / target.height : -target.cropY / target.height;
+      cy < 0
+        ? 1 + remainderY / target.height
+        : -target.cropY / target.height;
```
asturur: lint/prettier should take care of this, if it doesn't your PR will fail lint checks.

extensions/cropping_controls/croppingHandlers.ts
```diff
@@ -218,21 +255,28 @@ export const scaleEquallyCropGenerator =
...
     const constraint = target.translateToOriginPoint(
       target.getCenterPoint(),
       anchorOriginX,
       anchorOriginY,
     );
+
+    const pointerForLocalCoords = reflectPointerForFlip(target, x, y);
```
asturur: so you are reflecting before moving to local coordinates.
Are we sure that works ok with rotations?

extensions/cropping_controls/croppingHandlers.ts
```diff
@@ -283,3 +333,128 @@ export function renderGhostImage(
   );
   ctx.globalAlpha = alpha;
 }
+
+const clamp = (value: number, min: number, max: number) =>
+  Math.max(min, Math.min(max, value));
```
asturur: we have a util called `capValue` that is basically identical to this, let's use that. `util.capValue`.

extensions/cropping_controls/croppingHandlers.ts
```diff
@@ -283,3 +333,128 @@ export function renderGhostImage(
...
+        availableSize,
+        newCrossSize,
+        newScale,
+        isNegativeEdge ? 0 : initialCrop,
+        newCrossCrop,
+      );
+    }
+
+    return true;
+  };
```
asturur, line 477: this i have to take my time to review, please meanwhile address the other small things.

extensions/cropping_controls/croppingControls.ts
```diff
@@ -47,49 +51,57 @@ export const createImageCroppingControls = () => ({
...
     x: -0.5,
     y: 0,
-    sizeX: 4,
+    angle: 90,
+    sizeX: 8,
     sizeY: 20,
+    render: renderEdgeControl,
     cursorStyleHandler: scaleCursorStyleHandler,
-    actionHandler: changeCropX,
+    actionHandler: withFlip(changeCropX, changeCropWidth, 'flipX'),
```
asturur, line 64: for the existing controls let's keep the changes limited to the actionHandler.
I think both the roundy look and the different rendering logic have reason to exists, but let's introduce in a follow up pr with relative explanations on how to render them.

asturur:
I ll prepare a doc page for cropping controls, so that is easier to add there your new things rather than starting a page from scratch.
I had to do it anyway i was just postponing because is not easy for me to write engaging docs

asturur:
image

For those things let's wait a moment, let's fix the flip handler, the double click itself is not mandatory is just a default way to coordinate the object changes required to initiate cropping.
The crop rotation requires an entirely different approach to drawing the image, it needs to keep an internal angle that is decoupled from the one of the rotation and needs svg export to support it together with toObject/fromObject. On top of that then cropping becomes a clipping operation rather than a different usage of drawImage.

There is also an issue when moving the ghost image in a rotated situation, is really complicated because cropX and cropY together with width and height will have entirely different constraints when applied against image html element. Is not something you can handle just from an extension using exting render api, it requires a fundamental change on how the image renders.

mauricekindermann:
Cool thanks for the feedback. I've pushed fixes addressing your comments and scoped back. Had more cooked up but stashed it - we can revisit once we're in sync.

My goal with extensions is to help people like me - I spent a lot of time building my design tool fighting with things that "almost work" but don't feel right. Fabric is best in class, and I think first-party extensions could be too. I assume we align on that.

There are a lot of non-obvious UX expectations that big design tools get right. Things you don't notice until they're missing: 

- Exit crop on deselect/selection change (done)
- Ghost boundary matches selection styling (done)
- Fix hitboxes on L controls (done)
- Style controls in all modes to match (done)
- Flip handling works correctly (done)
- One crop mode active at a time
- Hide controls during repositioning
- Enter/Esc keyboard shortcuts  
- Crop image z-index to front, restore on exit 

I had a crack at rotation in crop mode and hit exactly what you described - the anchor math gets wild. Happy to leave that for later (though once solved people will drool). 

These changes are mostly bug fixes, deisgn fixes, or UX fixes that dont scope creep, I think... much. They certainly make manual testing a lot faster/easier and it all feels more robust.

extensions/cropping_controls/controlRendering.ts
mauricekindermann: Fixed - now using fabricObject.borderScaleFactor * 2 instead of hardcoded 4

extensions/cropping_controls/controlRendering.ts
mauricekindermann: Renamed to renderRoundedSegmentControl - went with "Rounded" over "Roundy" for consistency with other fabric naming conventions.

extensions/cropping_controls/controlRendering.ts
mauricekindermann: Done - path is now defined once, stroke called conditionally for border then unconditionally for fill. Also replaced hardcoded 4 with borderScaleFactor * 2 per your other comment.

mauricekindermann:
I did some research on CropMode Rotation, bookmarking for later, nothing super useful as you menhtioned core needs some things to make this doable, just wanted to see what was out there already. 

A nice write up with excellent visuals
https://uppy.io/blog/2023-10-25-image-editor/

And their code
https://github.com/transloadit/uppy/blob/main/packages/%40uppy/image-editor/src/Editor.tsx

Wikipedia has some related equations
https://en.wikipedia.org/wiki/Scaling_(geometry)#Uniform_scaling

React-Easy-Crop has some helpers (rotateSize, restrictPosition, getCropSize, computeCroppedArea)
https://github.com/ValentinH/react-easy-crop/blob/main/src/helpers.ts

Final boss verison needs to handle any rotation, any postiion, any clipping.

asturur:
Current Image rendering logic would require clipPath and that would break performance a lot in my opinion.
This can be achieved going back to clipTo function, or find a way to enable it reasonably. ClipTo function would also allow for cropping with shapes as long as they are described by a single path. Maybe can be done providing a subclassed Image with a bunch of code.

asturur:
I m busy with a small side project these days, but i ll do this before starting any other new fabricJS activity.

mauricekindermann:
Thanks @asturur. No stress or rush from me, reviewing is likely harder and more consequential than creating.

extensions/cropping_controls/croppingHandlers.ts
mauricekindermann: Nice, pointers are deceivingly tricky, I had a lot of trouble in my app as I have Figma style Layouts that let you live edit cropped images in Groups.  

So I had a rotated flipped image, in a rotated flipped Layout Group, and trying to make the pointers and controls behave themselves got so hard that I declared bankruptcy and did something similar to scene coordinates -  basically calcuate what is visually true (what is left, top, right, bottom), which made everything much easier and I wasn't tripping over myself. 

It's very possible I over-complicated or missed something, but I wasnt able to use core untils to accomplish what I was after. If fabric itself could handle, or if it already does, that would remove a whole layer of tricky object manipulation I ran into.

extensions/cropping_controls/croppingHandlers.ts
mauricekindermann: I believe so for standalone images - calcScale uses Math.abs() so it only cares about distance, not direction. Tested with various rotation + flip combos and it seems to hold up.

I pushed some group examples to the sandbox though (rotated group, flipped group) and it gets tricky fast - the crop controls don't behave properly when there's a parent transform involved. Might be worth a look when you get a chance as it's not just the edges but crop mode window controls, corner resize, ghost image, etc.

asturur:
I m nearly ready to merge, i could come back to this only today.
I fixed a small bug with flip and the edge controls.
I can't fix the bug with FLIP + SCALE GHOST.
I need to find a better name for edge controls.

Let's not add more features and let's aim at merging this.

extensions/cropping_controls/croppingControls.ts
mauricekindermann: I can split the L-shaped corners and rounded segments into a follow-up PR and keep this one to actionHandler changes only. Or are you OK with them here since they're already in?

mauricekindermann:
Sounds great, happy to get this merged. Let me know if you need anything from my side!

asturur:
@mauricekindermann just check looking at my diffs for scalingEqually here https://github.com/fabricjs/fabric.js/pull/10851/changes/d1d37c28bf7548552db6defa6651cc7285389c86#diff-0f54375a573aca37f3eddcf5fc1e7728d43b45bce17b52dba8ed12062aa9325e

When image is flipped your code does not allow for scaling at all.
With slightly different versions of my commits, either keeping your point adjust for flip or not, i can scale from 2 of 4 controls, than is better than none.
See if my different code opens a new perspective for you to fix it, ill try again tomorrow.

asturur:
ok i can't fix that bug today, i tried 3 times, i know what to do but i need time to draw a visual debug.
I m merging as is and opening an issue for flipped images to which i ll go back asap.

In the issue i ll describe my plan to fix it.

PR #10854, opened 2026-01-23
feat(): replace canvas with @napi-rs/canvas

Brooooooklyn:
Replace the `canvas` package with `@napi-rs/canvas`

`@napi-rs/canvas` has no system dependencies and is easy to install.


package.json
```diff
@@ -155,7 +150,7 @@
     }
   },
   "optionalDependencies": {
-    "canvas": "^3.2.0",
+    "@napi-rs/canvas": "^0.1.88",
```
Smrtnyk, line 153: is this a breaking change for consumers of fabric? if yes it might need to wait for new major version

yisibl:
> Fabric.js depends on [node-canvas](https://github.com/Automattic/node-canvas) for a canvas implementation (HTMLCanvasElement replacement) and [jsdom](https://github.com/jsdom/jsdom) for a window implementation on node. This means that you may encounter node-canvas limitations and [bugs](https://github.com/Automattic/node-canvas/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen).
> Follow these [instructions](https://github.com/Automattic/node-canvas#compiling) to get node-canvas up and running.

https://github.com/fabricjs/fabric.js/blob/0b6cdfa784789400cc43f3c4ac2d4eaad8d55d61/README.md?plain=1#L73

@Brooooooklyn 
The Readme contains some outdated documentation for node-canvas. You can either remove it or note that with the new napi-rs/canvas, these issues are no longer a concern.

package.json
Brooooooklyn: There may be differences in the rendered pixels between these two libraries, so yes

asturur:
This could probably an alternative backend, in an alternative export path, but the are hurdles to clarify before:
- napi-rs/canvas is at 0.1.88 version written pre-release stage
- i can't find in the repo what is the goal of the project, we need a canvas replacement, this is a toolkit for doing various thing it doesn't say much about the compatibility goal with browser canvas output

So just opening a PR and try to get it merged is not going to work.
What i suggest you if we want to try this additional backend is to:
- create a new env together with the node env and call it SKIA ( for now )
- add an extra export called fabric/skia where to export the same stuff as we do with fabric/node
- make so that there is a command to run the tests in node/skia mode rather than with node-canvas.

We can look at that and merge it adding this as a new backend, and slowly understand how it performs/how good it is/how easy is to keep up and maybe also contribute back to your lib, devs can try it out and maybe at some point it can be the default node one.

Does it sounds good?

e2e/tests/visual-output/rendering/testcases/text.ts
```diff
@@ -3,29 +3,19 @@ import { FabricNamespace, renderTestType } from '../../../types';
...
+    GlobalFonts.registerFromPath(
+      path.join(dir, 'Ubuntu-Regular.ttf'),
+      'Ubuntu',
+    );
+    GlobalFonts.registerFromPath(path.join(dir, 'Ubuntu-Bold.ttf'), 'Ubuntu');
+    GlobalFonts.registerFromPath(path.join(dir, 'Ubuntu-Italic.ttf'), 'Ubuntu');
+    GlobalFonts.registerFromPath(
+      path.join(dir, 'Ubuntu-BoldItalic.ttf'),
+      'Ubuntu',
+    );
```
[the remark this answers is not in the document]
asturur: Why this simplification? What does happen if i register only the ubuntu font, i use the bold file, i call it just ubuntu, and then i try to use it with the regular weight?

Do variable size fonts work?

Brooooooklyn:
@asturur sounds reasonable. I'll push the jsdom merge first (https://github.com/jsdom/jsdom/pull/4009#issuecomment-3811645651), then fabric, since jsdom is also a dependency of fabric

asturur:
I wanted to leave a comment more for you than the actual PR.
I see a topic you bring forward for napi-rs/canvas is better installation experience.
That is not moving the needle for anyone to accept an api change in my opinion.
People that usually have need for something to happen, make it work. node-canvas is not hard to install, you just have to read the docs, on standard consumer devices ( amd64, arm64, apple and linux ) it just works with prebuilt binaries, i didn't have an issue installing node-canvas since long.

What does matter is if the canvas emulator under node can:
- work with the same browser api
- offer better performances
- offer better support for certain features
- How much the library is mature/tested/stable/supported

If your goal is to replace node-canvas, because this implementation is better, let people adopt it by themselves, offer a compatibility layer that makes the drop in replacement easy.

For example:
https://github.com/jsdom/jsdom/pull/4009/changes#diff-e77ed5ec9013d2d51cf0dfaa8eea1f821520791d8772c4d7656a6abb8b9e588cR67
Isn't better to have the toBufferAsync method provided in the library, explaining is a node-canvas 1:1 helper, and then just do `canvas.toBuffer = canvas.toBufferAsync` ?

My suggestion, outside of this PR, is that you aim for easy of adoption if you want to go along with node-canvas and let people replace it.

I m happy to collaborate to integrate this, if we have a way to comunicate easily.

Brooooooklyn:
@asturur Thank you for your suggestion, it’s very valuable to me.
When I first started developing `@napi-rs/canvas`, my goal was to provide an implementation that’s more compatible with the browser APIs. Since most mainstream browsers use Skia, building the WHATWG Canvas API on top of Skia makes it easier to achieve consistent behavior.
Your comment made me realize that my README doesn’t reflect this point, so I’ll spend some time reorganizing the README and related documentation to demonstrate/test the advantages of `@napi-rs/canvas`.

The suggestion about a node-canvas compatibility layer is also very helpful. I will provide a set of node-canvas compatible APIs to help users more easily migrate to and test `@napi-rs/canvas`.

asturur:
@Brooooooklyn how is going with your napi-rs/canvas? any major changes that we should try to add it as an alternative backed to fabricJS?

Brooooooklyn:
@asturur thanks for asking! I've fixed a lot of bugs recently. Once no new bugs are reported in the current version for a period, I will release 1.0. There will be no breaking changes for some time afterward, and I will continue with the tasks here.

asturur:
How is going with your napi-rs/canvas? did ti become stable and similar to a real canvas with the api?

PR #10858, opened 2026-01-31
chore(): update vitest

Smrtnyk:
And fix newly failing tests on chromium
 For some reason on browser the mocking of this uid is not working
 I patched the mechanism a bit to not rely on this dynamic behavior and also adjusted snapshot matcher to round to 2 decimals since tests got bit unstable

I noticed that new precommit hook also lints touched spec files, and we didn't fix all type issues in our spec files yet.
That means as soon as you touch some of the older spec files you will also be forced to fix type issues in side, unrelated to your PR.
I excluded spec files from precommit hook to be typechecked for now, until we fix all the issues and it is safe to turn it on.
I will have a look.
Also i noticed that master is already flaky on firefox, I will try to have a look at it.

### DIFF
--- a/src/ClassRegistry.spec.ts
+++ b/src/ClassRegistry.spec.ts
@@ -56,7 +56,6 @@ describe('ClassRegistry', () => {
     expect(resolved, 'resolved different classes').not.toBe(resolvedSvg);
   });
   it('legacy resolution preparation', async () => {
-    vi.resetModules();
     const { classRegistry: freshRegistry } = await import('./ClassRegistry');
 
     // Registry should be empty before any shape classes are imported
@@ -66,7 +65,6 @@ describe('ClassRegistry', () => {
     expect(freshRegistry.has('object')).toBe(false);
   });
   it('legacy resolution', async () => {
-    vi.resetModules();
     const { classRegistry: freshRegistry } = await import('./ClassRegistry');
     const { Rect } = await import('./shapes/Rect');
     const { IText } = await import('./shapes/IText/IText');
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -9,6 +9,10 @@ if (!fixturesUrl.endsWith('/')) {
   fixturesUrl += '/';
 }
 
+// Node 20 requires vmThreads, newer versions can use threads
+const nodeMajor = parseInt(process.version.slice(1).split('.')[0]);
+const pool = nodeMajor <= 20 ? 'vmThreads' : 'threads';
+
 export default defineConfig({
   resolve: {
     alias: {
@@ -16,7 +20,7 @@ export default defineConfig({
     },
   },
   test: {
-    pool: 'vmThreads',
+    pool,
     clearMocks: true,
     mockReset: true,
     snapshotSerializers: [
@@ -44,7 +48,12 @@ export default defineConfig({
             provider: playwright(),
             enabled: true,
             headless: true,
-            instances: [{ browser: 'chromium' }],
+            instances: [
+              {
+                browser: 'chromium',
+                isolate: true,
+              },
+            ],
           },
           name: 'unit-chromium',
         },
@@ -63,6 +72,7 @@ export default defineConfig({
             instances: [
               {
                 browser: 'firefox',
+                isolate: true,
               },
             ],
           },


src/ClassRegistry.spec.ts
```diff
@@ -59,7 +59,7 @@ describe('ClassRegistry', () => {
   it('legacy resolution preparation', async () => {
     genericClassRegistryInstance[JSON].delete('rect');
     genericClassRegistryInstance[JSON].delete('i-text');
-    genericClassRegistryInstance[JSON].delete('activeSelelection');
```
Smrtnyk, line 62: seems like this was a typo

Smrtnyk:
there are some tests failing on class registry
seems like 1 test depends on another
Edit: hm, i hate when it works locally but fails in ci

Smrtnyk:
ok seems like i got it now
i changed the vitest config a bit to have better test isolation
firefox still fails on 2 tests but that is same as on master
i will have a look at that separately
playwright update also has some breaking changes so have to look at that as well

asturur:
@Smrtnyk the pre-commit hook was a bit rushed i admit. I know i need it, but i didn't bother doing it properly.

src/shapes/__snapshots__/Image.spec.ts.snap
```diff
@@ -12,11 +12,11 @@ exports[`FabricImage > Svg export > It exports an svg with styles for an image w
 		<feMergeNode in="SourceGraphic"></feMergeNode>
 	</feMerge>
 </filter>
-<clipPath id="imageCrop_2">
+<clipPath id="SVGID">
 	<rect x="-75" y="-75" width="150" height="150" />
 </clipPath>
-	<image style="stroke: rgb(255,0,0); stroke-width: 11; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(0,0,0); fill-rule: nonzero; opacity: 1;filter: url(#SVGID_0);"  xlink:href="" x="-85" y="-85" width="200" height="200" clip-path="url(#imageCrop_2)" ></image>
-	<rect x="-75" y="-75" width="150" height="150" style="stroke: rgb(255,0,0); stroke-width: 11; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: none; fill-rule: nonzero; opacity: 1;filter: url(#SVGID_0);" />
+	<image style="stroke: rgb(255,0,0); stroke-width: 11; stroke-dasharray: none; stroke-linecap: butt; stroke-dashoffset: 0; stroke-linejoin: miter; stroke-miterlimit: 4; fill: rgb(0,0,0); fill-rule: nonzero; opacity: 1;filter: url(#SVGID);"  xlink:href="assets/xlink:href=" x="-85" y="-85" width="200" height="200" clip-path="url(#SVGID)" ></image>
```
asturur, line 18: Image

The SVG has this new value here, what is it for?

asturur: Or like, why does it comes out in the snapshot?

vitest.config.ts
```diff
@@ -9,14 +9,18 @@ if (!fixturesUrl.endsWith('/')) {
   fixturesUrl += '/';
 }
 
+// Node 20 requires vmThreads, newer versions can use threads
+const nodeMajor = parseInt(process.version.slice(1).split('.')[0]);
+const pool = nodeMajor <= 20 ? 'vmThreads' : 'threads';
```
asturur, line 14: oh ok good to know this.

asturur: looks like node 20 tests completes in half time compared to 22/24/chrome/firefox. Is not a big deal, but is vmThreads the fastest?

vitest.config.ts
```diff
@@ -44,7 +48,12 @@ export default defineConfig({
             provider: playwright(),
             enabled: true,
             headless: true,
-            instances: [{ browser: 'chromium' }],
+            instances: [
+              {
+                browser: 'chromium',
+                isolate: true,
```
asturur, line 54: i read isolate: true is the default.

vitest.config.ts
Smrtnyk: it probably is yeah

vitest.config.ts
Smrtnyk: hm i don't think it is for the browser mode, that is why i moved to threads, since this has no affect on vmThreads

src/shapes/__snapshots__/Image.spec.ts.snap
Smrtnyk: hm interesting, maybe my ide did something, i will have a look

PR #10861, opened 2026-02-02
test(): refactor test, put common data in shared function

Smrtnyk:
Refactors some tests that initialize with common properties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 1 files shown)


Smrtnyk:
seems like something is off with ci

asturur:
https://www.githubstatus.com/

time to take a break

asturur:
thank you this was never done for lazyness.
Tests clean code is always a second thought between many

PR #10862, opened 2026-02-02
chore(): update all deps to latest

Smrtnyk:
now that vitest and playwright are out of the way we can update all other deps as well

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "docs": "typedoc",
     "cli": "node ./scripts/index.mjs",
     "sandboxscript": "node ./scripts/sandbox.mjs",
-    "typecheck": "tsc -p ./tsconfig.build.json --noEmit",
+    "typecheck": "tsc -p ./tsconfig.build.json --noEmit && tsc -p ./tsconfig-extensions.json --noEmit",
     "build": "npm run cli -- build",
     "build:fast": "npm run build -- -f",
     "dev": "npm run cli -- dev",
--- a/tsconfig-extensions.json
+++ b/tsconfig-extensions.json
@@ -103,6 +103,13 @@
     // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
     "skipLibCheck": true /* Skip type checking all .d.ts files. */
   },
-  "exclude": ["dist", "dist-extensions", "test", "e2e"],
+  "exclude": [
+    "dist",
+    "dist-extensions",
+    "test",
+    "e2e",
+    "**/*.spec.ts",
+    "**/*.test.ts"
+  ],
   "include": ["extensions/**/*"]
 }
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -5,7 +5,7 @@
     "noEmitOnError": false,
     "allowJs": false
   },
-  "include": ["src", "fabric.ts", "index.ts", "index.node.ts", "extensions"],
+  "include": ["src", "fabric.ts", "index.ts", "index.node.ts"],
   "exclude": [
     "dist",
     "dist-extensions",


asturur:
I want to make a test with AI on this PR

asturur:
well ok i can't.

Smrtnyk:
hm interesting, build failed, but didn't fail locally for me, at least it gave back exit code 0
seems like typescript related

to be honest, i am not a fan of this roullup integration with typescript its too noisy and very coupled with rollup
we could try to do the typecheck with pure tsc or even with new tsgo
what do you think? if you agree i could research a bit that approach

my idea is this: do tsc --noEmit to do fast typechecks 
remove typescript plugin from rollup and add esbuild plugin to it for fast transpilation and bundling, keep terser as a minifier as it is now

Smrtnyk:
I put this to draft
I will open separate pr just for prettier update and prettier fixes
and then rebase this PR and push separate commit with my proposal for the build so you can have a focused look

asturur:
The test i wanted to do with AI, just for curiosity, was if it can check the legitness of package-lock changes, i wonder if is a possible vector attack to include an hash change of a lib in a large diff of a yarn/package lock change.
I thought this was a good pr to test it.

Smrtnyk:
Hi @asturur 
I have pushed the proposal i had about to:
employ native tsc for typechecking
use esbuild for transpilation of typescript
typecheck is a separate step in CI now

let me know what you think, I am also fine if you don't want this
I can also separate the build change to a separate isolated PR

asturur:
Sorry @Smrtnyk i ended up in a busy period, i ll get back to this PRs asap

asturur:
what about rolldown instead of esbuild? Is there really a winner between all the alternatives?

asturur:
ok i read the PR is interesting.
We have rollup and rollup plugin esbuild.
I recently had an issue with esbuild and mocking with jest in an app, for how esbuild does the bundling, and i wonder if that can happen as well.

I m not against this, in general as long as we have browser target and if the build is faster is ok.
We are removing a bunch of code in the non minified version ( 147.348!!! ) that i need to understand what is it, and 600 bytes in the minified one.

Smrtnyk:
> what about rolldown instead of esbuild? Is there really a winner between all the alternatives?

i wanted to talk about this as well, next step would be to ditch rollup completely in favor of rolldown but its still not stable, its in rc phase and even after release to stable we should wait for it to stabilize and for bugfixes that would emerge
but definitely i see it as a future
that way we ditch both rollup and esbuild alltogether

Smrtnyk:
> ok i read the PR is interesting. We have rollup and rollup plugin esbuild. I recently had an issue with esbuild and mocking with jest in an app, for how esbuild does the bundling, and i wonder if that can happen as well.
> 
> I m not against this, in general as long as we have browser target and if the build is faster is ok. We are removing a bunch of code in the non minified version ( 147.348!!! ) that i need to understand what is it, and 600 bytes in the minified one.

yes i noticed that there is code removal, smaller bundle
it could be that rollup typescript was leaving a lot more code artifacts after itself making it harder for terser to minify it and eliminate dead code
esbuild might be doing more optimized passes

I assumed as e2e tests passed that the output is healthy since playwright is consuming the dist right?

asturur:
Who is left using babel at this point? jest?

Smrtnyk:
> Who is left using babel at this point? jest?

I think babel is used by playright still, there is some circus going on in watch mode there, where it recalculates some imports
i wonder if that can be rewriten using esbuild, or removed if nobody is using it
I certainly am not using watch mode for e2e tests

Smrtnyk:
so this https://github.com/fabricjs/fabric.js/blob/master/package.json#L86
and this https://github.com/fabricjs/fabric.js/blob/master/e2e/.babelrc.mjs

we should look at this separately why this is needed and if this can be replaced with esbuild if it is still needed

Smrtnyk:
given all of this @asturur I would still be careful with merging this
I assembled this as a good step forward towards the future and faster and easier to understand builds
but you will need to inspect the shape of the output artifact, if it is correct
build passes yeah, but still
maybe if you have some complex project you could put this build in and try it out manually

asturur:
i m going to compare the difference in js and post here for context. I didn't open it yet but i supposed esbuild is killing the comments from the non bundled file. That may explain the large difference. The 600 bytes i can check manually

Smrtnyk:
> i m going to compare the difference in js and post here for context. I didn't open it yet but i supposed esbuild is killing the comments from the non bundled file. That may explain the large difference. The 600 bytes i can check manually

ah that could be it, we might need some more config to pass to it to preserve them

Smrtnyk:
I had a look between non minified bundles
and you are right
it is mostly jsdoc and some spaces
seems like esbuild can't preserve jsdoc https://github.com/evanw/esbuild/issues/294#issuecomment-664627219
this is from the creator of esbuild
there is also a workaround on the bottom to preserve it but it looks hacky

alternative is that we try to jump immediately on rolldown and try it out, maybe it is already ready enough for our usecase
I will play with it and see what kind of results it is giving

i also pushed small fix, i forgot to exclude spec files from type emition

Smrtnyk:
lets close this, i think rolldown path is the way to go, and it has a brigth future
so we can avoid building up frankensteins combining 10 tools for building the project

PR #10864, opened 2026-02-03
chore(): reuse more of the createPointerEvent in unit tests

Smrtnyk:
This reduces some code and also removes ugly type assertions on some places


Smrtnyk:
this is ready now

PR #10868, opened 2026-02-11
chore(): try rolldown

Smrtnyk:


Smrtnyk:
OK first of all build with rolldown is crazy fast (and we still kept terser that is js plugin) and is 2 seconds now

I asked llm to analyse the diff between old rollup + babel  approach vs rolldown output
it says this
```text
That tells the story clearly. Two main reasons for the size difference:
                                                                                                                                                                                         
  1. No Babel runtime helpers. The old Rollup build ran @babel/preset-env which injected _defineProperty, _toPrimitive, _toPropertyKey helper functions and transformed every class      
  property initialization into _defineProperty(this, ...) calls. Rolldown's Oxc transforms with target: ['chrome88', ...] doesn't need to do this — Chrome 88+ and the other targets all
  natively support class fields, so the code stays as-is.                                                                                                                                
                                                                                     
  2. Different bundling approach. Rolldown uses a small runtime (__defProp, __exportAll) instead of Babel's heavier helper pattern, and its treeshaking/output generation is different.

  The old output was bloated by unnecessary Babel downleveling. The .babelrc used @babel/preset-env with no explicit targets, which picked up .browserslistrc (Safari 13+). Safari 13
  doesn't support class fields, so Babel was transforming every class property into _defineProperty() calls — adding significant overhead.

  Now with Rolldown targeting chrome88, safari13, firefox85, edge88, the Oxc transformer is smarter about what actually needs lowering per-feature per-target. Class fields that Safari
  13 doesn't support would still get lowered, but with less runtime overhead than Babel's approach.

  The minified output difference is smaller (309KB → 301KB) because terser was already stripping most of the verbosity.

  This is expected and correct — the output is functionally equivalent, just leaner.
  ```
  
  that is also probably why the esbuild example had smaller minified output, less babel helpers shenanigans
  In the future we can get rid of terser as well, rolldown will be able to minify natively, but for now their minifier is in alpha
  
  If you ask me, this or the esbuild approach in other PR, i would say this because this is the future, its where the ecosystem is going

**Comment by mauricekindermann** [comment]
Nice, works well on my testing. I got 6-9s build in master to <1s build in this PR. Total build went from 30s to 17s due to tsc building. No idea how to eval the build itself and it's stability.

**Comment by Smrtnyk** [comment]
> Nice, works well on my testing. I got 6-9s build in master to <1s build in this PR. Total build went from 30s to 17s due to tsc building. No idea how to eval the build itself and it's stability.

Ideally the e2e tests are a source of confidence
and the reason why my confidence is low is that e2e tests use babel to do some acrobatics with bundling
I think this PR should also change that part, to ditch babel completely and make sure that what consumers of the fabric get that e2e tests get the exact same thing
I should also update this PR with the latest release of rolldown so we are up to date

**Comment by Smrtnyk** [comment]
>due to tsc building

When tsgo is stable we can switch to that and it should be much faster then

**Comment by asturur** [comment]
I don't get the part about safari 13. One par says it wasn't picking it up, the other yes.

```
The .babelrc used @babel/preset-env with no explicit targets, which picked up .browserslistrc (Safari 13+). Safari 13
  doesn't support class fields, so Babel was transforming every class property into _defineProperty() calls — adding significant overhead.

  Now with Rolldown targeting chrome88, safari13, firefox85, edge88, the Oxc transformer is smarter about what actually needs lowering per-feature per-target. Class fields that Safari
  13 doesn't support would still get lowered, but with less runtime overhead than Babel's approach.
```

it is basically saying 'rolldown has shorter code for `_defineProperty` ? in general i do not love the `_defineProperty() ` and public property in general because it is hard to manage with configurable defaults that the library tries to support.

But if is equivalent is equivalent. I agree that if we could flatten everuthing to rolldown rather than babel for testing and rolldown for the rest would be safer.

Between esbuild + rollup and rolldown i would prefer rolldown because that was my plan, but that was my plan because i didn't consider esbuild that in my head was not actively developed/with some issues.

I then read better and it looks like esbuild is just strict at what to support ( correctness and specs ) rather than make every app happy. So my opinion may be invalidated entirely.

This can be discussed openly i don't have any real preference.

Priorities are in order:
- correctness
- easy to support node and browser
- lean output
- easyness of config

**Comment by Smrtnyk** [comment]
> I don't get the part about safari 13. One par says it wasn't picking it up, the other yes.

I think it means they are both lowering but that rolldown is smarter to lower only what is needed rather than the whole ecma feature level
Unfortunately I have no access to mac and especially not to such old safari version

I would merge this only when:
1. babel is completely ditched from the project so both e2e tests and main build use rolldown and e2e tests consume real bundle that real user will get
2. someone has tested this on old safari versions that fabric.js is supporting

I will work on point 1, but with point 2 I can't help

**Comment by Smrtnyk** [comment]
i think rolldown is smarter option here than rollup + esbuild

rolldown uses oxc, we use already vitest that uses vite under the hood that will use rolldown
in the end we will then have 1 single build tool for everything fabric does
building, testing bundling and minifying when oxc minifier is ready
then we will also be able to remove terser as well

**Comment by asturur** [comment]
I ditched windows in favor of mac :D but i don't have safari 13 because you can't rollback that easily. Maybe there is a way to run it in some way, i ll check. Maybe is possible to have it in the CI too.

**Comment by Smrtnyk** [comment]
> I ditched windows in favor of mac :D but i don't have safari 13 because you can't rollback that easily. Maybe there is a way to run it in some way, i ll check. Maybe is possible to have it in the CI too.

unfortunately (or fortunately) I only work on linux so I am bit limited here

**Comment by asturur** [comment]
Important is not being stuck on windows :D The linux/mac is just a flavor on top of something that works.

**Comment by Smrtnyk** [comment]
@asturur I pruned babel and rewrote playwright test file resolver in rolldown, it is also simpler now
I also checked, e2e tests use the same fabric bundle that users are getting so we are good to go
the only blocker is that someone has to test this on older safaris, which I can't do

**Comment by mauricekindermann** [comment]
Is Safari 13 critical? I run windows+mac but I always have to use BrowserStack to manually verify things in old browsers. 

I assume it's being discussed because of this: 
https://github.com/fabricjs/fabric.js/blob/286d610ae596bb7f24ec3125d62f7a51e407d009/.browserslistrc

Usage is pretty small https://caniuse.com/usage-table

```
Safari 26.x: ~0.85% (current)
Safari 18.x: ~0.19%
Safari 17.x: ~0.35%
Safari 16.x: ~0.20%
Safari 15.x: ~0.08%
Safari 14.x: ~0.05%
Safari 13.x: 0.02%
```

Note: these are % of all global web traffic, not Safari-only. Safari 13 is roughly 1% of Safari desktop users. The raw data can be downloaded here as CSV: https://gs.statcounter.com/browser-version-market-share/desktop/worldwide

Is there a policy or way the current browserse were determined? 

```
Chrome 88: Jan 2021
Edge 88: Jan 2021
Firefox 85: Jan 2021
Opera 73: Dec 2020
Safari 13: Sept 2019
```

If it were changed to a target year, eg 2021, that would be Safari 15 and Opera 74-82. 

Regardless - do e2e tests have to actually pass in older browsers, or do we trust what the tools say they do/support? 

If they have to pass, it really should be automated. So I did some digging, and checkout what jQuery does:
https://github.com/jquery/jquery/blob/main/.github/workflows/browserstack.yml

Seems like it's free (5 useres, 5 parallels)
https://www.browserstack.com/open-source

I also checked Vite, Tailwind, and a few others, they set browser targets but trust the tooling. jQuery seem to be one of the few who actually test older browsers.

**Comment by Smrtnyk** [comment]
I personally do not care about safari 13 but it is call from Andrea here
I assume things work there, but it is just a precaution to test and verify after changing the build tool
Another option is also to wait for major version increase for fabric and bump browser targets to more modern

**Comment by asturur** [comment]
Safari 13 is not critical. Is just i hate to make hardware obsolete for no reason.
Canvas api new functionalities aren't many and the new one ( css filters and something else ) are also a bit in contrast with how fabricJS works.
Supporting Safari 13 is just a way to say an app built with fabricJS can still work on an older mac if you transpile the JS correctly.

You can always transpile it back at build time of your own app, is just that if i use a new canvas api not supported by a browser i actually block you or you need to find a polyfill of some sort.

So no, safari 13 JS support is not critical, we don't care, we just care to not use canvas api that do not work on safari 13.
For a bit, then we will ditch it anyway in some time.

File: scripts/build.mjs [hunk 1]
```diff
@@ -39,6 +52,8 @@ export function build({ watch, fast, input, output, stats = false } = {}) {
   } else {
     try {
       cp.execSync(cmd, processOptions);
+      // Generate .d.ts files after successful rollup build
```
asturur, line 55: rolldown?

asturur: we don't care to be clear.

asturur:
The only issue that i see is how did we get extra 72k of bundle. digging.

asturur:
is just whitespaces

asturur:
@Smrtnyk i added a minify:true option to the minified bundle, but i m not entirely sure it now uses terser or the internal oxc minifier.

Smrtnyk:
> @Smrtnyk i added a minify:true option to the minified bundle, but i m not entirely sure it now uses terser or the internal oxc minifier.

I assume both, since terser is the plugin it should be invoked

asturur:
Well let's try like this.
Let's leave them both, i ll open an issue on rolldown about a certain config not removing whitespaces.
We leave both minifier, they say oxc is still in alpha, that is why i m not super happy to enable it, but if a bug arise we will catch it hopefully.

asturur:
Thanks this was monumental work, i would never have wanted to dig into the e2e test setup to figure out what is going on there. I didn't make it i do not really want to spend too much time on it, it was working before, it is working now, thank you.

Smrtnyk:
> Well let's try like this. Let's leave them both, i ll open an issue on rolldown about a certain config not removing whitespaces. We leave both minifier, they say oxc is still in alpha, that is why i m not super happy to enable it, but if a bug arise we will catch it hopefully.

Yeah the oxc minifier is in alpha but bundler and transpiler and everything else should be decently stable
I saw plan from them that minifier is the next big focus, so when it is done and stable we could ditch terser

Smrtnyk:
> Thanks this was monumental work, i would never have wanted to dig into the e2e test setup to figure out what is going on there. I didn't make it i do not really want to spend too much time on it, it was working before, it is working now, thank you.

You are welcome, and no worries
I myself have quite a bit of experience with playwright, we use it heavily at the company I work for

For me my main focus is if I can help with having fabric repo stable, with good developer experience, up to date non flaky testing infrastructure and as little of acrobatics to setup the dev as possible, so you can have more focus on actual documentation and features

asturur:
i m merging this and the other. before publishing a new release i ll maybe do a larger test, but e2e test should be enough honestly.

Issue #10873, opened 2026-02-25
[Bug]: loadFromJson() with image in 404 breack canvas


multivoltage:
### CheckList

- [x] I agree to follow this project's Code of Conduct
- [x] I have read and followed the [Contributing Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md)
- [x] I have read and followed the [Issue Tracker Guide](https://github.com/fabricjs/fabric.js/blob/master/CONTRIBUTING.md#%EF%B8%8F-issue-tracker)
- [x] I have searched and referenced existing issues and discussions
- [x] I am filing a **BUG** report.
- [x] I have managed to reproduce the bug **after** upgrading to the **latest** version
- [x] I have created an accurate and minimal reproduction

### Version

6.6.4/7.2.0

### In What environments are you experiencing the problem?

Chrome

### Node Version (if applicable)

22.10.0

### Link To Reproduction

https://jsfiddle.net/fozpt8uj/26/

### Steps To Reproduce

1. Create a json config with an image that return 4xx (not 200)
2. call loadFromJson()
3. all next call do not produce any result on canvas

### Expected Behavior

Canvas should works but without the broken image, or add a filter/reviver function to fix that.

I found and read also https://github.com/fabricjs/fabric.js/issues/3578 
I checked the last version 7.x. 
Issue exist in node js env but also with cdn (like my jsfiddle).

### Actual Behavior

I load a correct config => canvas render object.
I load load a wrong config => canvas render white board.
At this time even if I call some new method to add new object canvas remains blank

### Error Message & Stack Trace

```bash

```

asturur:
Marked this as beginner friendly because does not require deep canvas knowledge or weird fabric events knowledge. for whoever wants to try to fix this before me, in the StaticCanvas class, there is a loadFromJson method that is based on a promiseAll. This needs to be changed probably in an `allSettled` and necessary changes applied.
But that is the core of the issue.

Unit tests needs to be added and some back and forth in review about what to do with broken objects ( console warnings? possibility of swapping with a reviver function, etc etc )

multivoltage:
I will try :).
Maybe adding an optional handler function (or callback error) after reviver?
I dont think to edit reviver function since doc says:

`Method for further parsing of JSON elements, called after each fabric object created.`.
What do you think=

asturur:
[...]

PR #10875, opened 2026-02-27
fix: avoid crash if loadFromJSON load config with not existing images

multivoltage:
## Description

Try to fix: https://github.com/fabricjs/fabric.js/issues/10873

## In Action
this code resolves this the issue without change any api. 
Maybe the edit should add some handler like the proposed one in the issue page. But probably this requires to rework the promises waterfall starting from "loadJsonConfig().

I opened the issue because of I used the library in a project where this scenario happens frequently. So for me should be enough my edit, without adding handler. But can improve is someone drive me in the right direction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [next]
 
 - feat(): fix(): avoid crash if loadFromJSON load config with not existing images, feat(): allow for fallback objects [#10875](https://github.com/fabricjs/fabric.js/pull/10875)
+- Cleanup animation delay timeout [#10830](https://github.com/fabricjs/fabric.js/pull/10830)
 - chore(): Update typescript-eslint to address a recent vulnerability [#10878](https://github.com/fabricjs/fabric.js/pull/10878)
 - chore(deps-dev): bump rollup from 4.52.4 to 4.59.0 in the npm_and_yarn group across 1 directory [#10877](https://github.com/fabricjs/fabric.js/pull/10877)
 - Gradient controls [#10844](https://github.com/fabricjs/fabric.js/pull/10844)
--- a/src/util/animation/AnimationBase.ts
+++ b/src/util/animation/AnimationBase.ts
@@ -1,4 +1,5 @@
 import { noop } from '../../constants';
+import { getFabricWindow } from '../../env';
 import { requestAnimFrame } from './AnimationFrameProvider';
 import { runningAnimations } from './AnimationRegistry';
 import { defaultEasing } from './easing';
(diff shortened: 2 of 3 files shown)


src/util/misc/objectEnlive.ts
```diff
@@ -119,7 +119,12 @@ export const enlivenObjects = <
           }),
       ),
     )
-      .then(resolve)
+      .then((elementsResult) => {
+        const addedInstances = elementsResult
+          .filter((e) => e.status === 'fulfilled')
+          .map((e) => e.value);
+        resolve(addedInstances);
+      })
```
asturur, line 136: This is a good start.
If you want to dig deeper think of this extra steps:
- as a developer of an app i want my json to still be able load if just a small part is missing
- as a developer i want to know what went wrong when loading
- as a developer i do not want my design to be completely messed up if something is missing in the design ( image missing in a group )
- as a developer i want to be able to cancel loading if something went wrong.

Now if we move the reviver part down a step, inside the elementsResults loop, i can provide a reviver function so that:

```
     .then(async (elementsResult) => {
        for (const [index, result] of elementsResult.entries()) {
           if (result.status === 'fulfilled') {
             await reviver(objects[index], result.value);
             instances.push(e.value);
           }
            if (result.status === 'rejected') {
              // allow reviver to get the error for decision making, logs, optionally return fallback object
              const fallback = await reviver(objects[index], undefined, result.reason);
              if (fallback) {
                instances.push(fallback);
              }
            }
        }
        resolve(instances);
      })
```

This should both keep the loop running if you don't pass any reviver, but allow you to drive your loading through the reviver for fallback objects or simply logging.
For example for a failed image you may want to provide a fallback rect of the same size or whatever you want a warning sign.

Let me know if you want to try to explore this or not. You don't have to

multivoltage: I'll explore the idea

multivoltage: I edit the PR based on you idea. I think you used "async" because user can add a FabricImage with placeholder src (which require a call to loadUrl) ?
All test are green. But cannot define a good types in the reviver implementation.
At this point if user want to return an "instance" on the reviver typescript produce an error

asturur: ok we can fix the types.
Yes the async is mostly because in 2026 we need to be comfortable in places where rendering is not involved, to allow for async functions, loading an image, loading a module of code that isn't ready yet or whatever that dev wants to do. If is a dev tool and is not in the middle of a canvas rendering process it should be async.

I ll review asap, if there are very minor details to adjust i ll eventually just push code rather than going back and forth with comments.

multivoltage: yes, thanks. I just did a commit only to update jsdoc

asturur:
i will fix the type ( if necessary ) in a small follow up, this is a grate contribution, thanks.

PR #10895, opened 2026-03-15
chore(): simplify svg creation in unit tests via shared helper

Smrtnyk:
Continuation of the work I was doing before with unit tests to bring them in good shape

### DIFF
--- a/.github/workflows/comment_coverage.yml
+++ b/.github/workflows/comment_coverage.yml
@@ -15,6 +15,11 @@ jobs:
     name: Coverage reporting
     runs-on: ubuntu-24.04
     steps:
+      - name: Checkout actions
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          sparse-checkout: |
+            .github
       - name: 'Create an empty tmp directory'
         run: |
           mkdir ${{ runner.temp }}/artifacts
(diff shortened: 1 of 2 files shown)


asturur:
definetely is not working this coverage. 2236 lines is not correct.
I should probably stop to try to merge them and be happy with vitest coverage.

Smrtnyk:
> definetely is not working this coverage. 2236 lines is not correct. I should probably stop to try to merge them and be happy with vitest coverage.

that was my concern
I am not familiar how these two can work together
the plawyright is consuming obfuscated bundled source if I am thinking it right, where vitest operates on source code
playwright would need some kind of source maps to resolve it back to the source code or something

but I don't have enough context, and maybe it worked differently before

asturur:
i think you can download the coverage files from the action tabs, i don't remember them being obfuscated. or maybe that is the issue, before playwright wasn't consuming the minified bundle?
I understand for an e2e test make sense.

At work we mix the 2 and we nail down uncovered lines correctly from playwright too.

asturur:
if this is good to go you can merge when you feel like

Smrtnyk:
> if this is good to go you can merge when you feel like

yeah the PR is good itself
I am doing these bitesized PRs for unit tests to not overwhelm you
I myself do not have the merge button so you will need to merge when you have time

Smrtnyk:
> that is the issue, before playwright wasn't consuming the minified bundle?

could be, I think we would need to make playwright consume unobfuscated source code bundle
It definitely needs some play around
I never attempted something like that, but could have a look later on how it could work

asturur:
if is green you should see the merge button no? is approved, you should see it now.

Smrtnyk:
> if is green you should see the merge button no? is approved, you should see it now.

no i just see the close and comment buttons, no merge
I assume I don't have write permissions to the repo

Smrtnyk:
a question that is a bit offtopic, sorry for that
have you considered using something like https://www.sonarsource.com/products/sonarqube/
I use it at work and for private projects
they offer free usage for open source projects
So far in these 7 years of using it I really like the tool and it catches lot of stuff
It could run on PRs via github action

asturur:
sonarqube is good i use it professionally, i didn't consider using it here because the simple text report was ok. Then file number exploded, test framework too, and the small text report is not good anymore. I ll have to add it at some point.

asturur:
yeah i just got the oss plan, let's see if it can give me just a simple coverage report on change lines.

PR #10896, opened 2026-03-15
chore(): try to improve playwright tests speed

Smrtnyk:
just experimenting to see if they really need to run for 5 minutes in ci

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -27,7 +27,6 @@ const config: PlaywrightTestConfig = {
   forbidOnly: !!process.env.CI,
   /* Retry once to get the video and trace for truly failing tests */
   retries: 1,
-  /* Limit parallel workers on CI. */
   workers: process.env.CI ? '100%' : undefined,
   /* Do not update snapshot on CI */
   updateSnapshots: process.env.CI ? 'none' : 'missing',


asturur:
probably the visual similarities test needs to be divided in one test per folder, instead of the monster it is now, it probably takes a lot of time and is the driver of the overall timer.

Smrtnyk:
> probably the visual similarities test needs to be divided in one test per folder, instead of the monster it is now, it probably takes a lot of time and is the driver of the overall timer.

I am currently looking at the config, it seems redundant to capture video all the time for evertyhing when you would look at it only if test failed or?
capturing videos and traces are very expensive, you probably want it for truly failed tests taht is if they retried and failed

after that then comes what can be looked at on test basis

asturur:
```e2e/tests/visual-output/rendering/index.spec.ts:20:9 › VISUAL RENDERING TESTS```
This one is the issue.
Should be broken in pieces.

**Comment by asturur** [comment]
> > probably the visual similarities test needs to be divided in one test per folder, instead of the monster it is now, it probably takes a lot of time and is the driver of the overall timer.
> 
> I am currently looking at the config, it seems redundant to capture video all the time for evertyhing when you would look at it only if test failed or? capturing videos and traces are very expensive, you probably want it for truly failed tests taht is if they retried and failed
> 
> after that then comes what can be looked at on test basis

oh i didn't know i was doing that

**Comment by Smrtnyk** [comment]
ok witht his config it is now 3m instead of 5

**Comment by Smrtnyk** [comment]
@asturur I adjusted the config here only
on test basis i would look in a separate PR so it would be a separate change

**Comment by Smrtnyk** [comment]
another thing, do you need the test coverage in playwright, when you already have it in vitest?

**Comment by asturur** [comment]
the coverage is global, playwright and vitest and then is added.
The action to post it in the PR is broken and i need to get a look at it

**Comment by Smrtnyk** [comment]
> the coverage is global, playwright and vitest and then is added. The action to post it in the PR is broken and i need to get a look at it

How does that work?
Playwright loads the bundle that is minified, is it using the source maps?
Where vitest operates on the sourcd code

I wasnt aware that such combining is possible

**Comment by asturur** [comment]
well it was working with previous unit test method, before the mega switch to full jest/vitest.

I ll look at the action and we discover.

Is this meargeable?

File: playwright.config.ts [hunk 1]
```diff
@@ -48,8 +47,8 @@ const config: PlaywrightTestConfig = {
     baseURL: 'http://localhost:8080',
 
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
-    trace: 'retain-on-failure',
-    video: process.env.CI ? 'retain-on-failure' : 'on',
+    trace: 'on-first-retry',
+    video: 'on-first-retry',
```
asturur, line 51: oj this is great

Smrtnyk:
> well it was working with previous unit test method, before the mega switch to full jest/vitest.
> 
> I ll look at the action and we discover.
> 
> Is this meargeable?

Yeah its good to go

PR #10912, opened 2026-03-16
refactor(tests): remove coverage collection from playwright

Smrtnyk:
Lets see if this gives some speed increase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- refactor(tests): remove coverage collection from playwright [#10912](https://github.com/fabricjs/fabric.js/pull/10912)
 - ci(): fix sonarqube lcov path after artifact download [#10910](https://github.com/fabricjs/fabric.js/pull/10910)
 - ci(): Try to enable sonarqube cloud for coverage reporting [#10903](https://github.com/fabricjs/fabric.js/pull/10903)
 - docs(agents): add repo AGENTS guide and PR skill [#10900](https://github.com/fabricjs/fabric.js/pull/10900)


asturur:
Defintely faster

Smrtnyk:
@asturur with this change it is 2 minutes now
yesterday it was 5 minutes, then we adjusted config and made it 3 minutes and now with this 2 minutes

I also saw a warning

```
Warning: Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809, actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332, actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```
not sure if you want to do something about it

asturur:
Yes the actions needs to be updated, but most of them already work on node24 when node20 will be removed.

Smrtnyk:
OK then this PR is ready to be merged if you want

partially unrelated quesiton: there is  coverage:merge script in package.json, is that thing needed what it is merging?
and do we need the nyc dep at all?

PR #10929, opened 2026-03-26
chore(deps): update devDependencies to latest versions

Smrtnyk:
Fixes the security vulnerability reported by flatted
 https://github.com/vitest-dev/vitest/issues/9975


src/benchmarks/boundingBoxFromPoint.mjs
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable no-restricted-syntax */
```
Smrtnyk, line 1: lint --fix removed these, not sure what they were about

Smrtnyk:
linter got some improvements so I ran lint autofix, therefore changes in multiple files

asturur (approved):

PR #10932, opened 2026-03-29
refactor(test): fix dead assertions in Shadow.spec.ts

Smrtnyk:
Seems like this somehow went through during test migrations from qunit

I am going through sonar issues and will be opening bite sized PRs for things that make sense
this is first of those findings


asturur (dismissed):
thanks good finding

PR #10935, opened 2026-03-30
chore(): update typescript to 6

Smrtnyk:
updates major version of typescript
 also typescript eslint is updated now since it officially supports ts 6
 Some lint errors started happening so I had to fix those as well

https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/
https://github.com/typescript-eslint/typescript-eslint/releases/tag/v8.58.0

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -26,7 +26,7 @@
     /* Modules */
     "module": "ESNext" /* Specify what module code is generated. */,
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    "moduleResolution": "node" /* Specify how TypeScript looks up a file from a given module specifier. */,
+    "moduleResolution": "bundler" /* Specify how TypeScript looks up a file from a given module specifier. */,
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
     "paths": {
       "fabric": ["./index"],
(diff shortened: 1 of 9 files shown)


asturur:
oh it is out? let me check.

tsconfig.json
```diff
@@ -73,7 +73,6 @@
     /* Interop Constraints */
     // "isolatedModules": true,                          /* Ensure that each file can be safely transpiled without relying on other imports. */
     // "allowSyntheticDefaultImports": true,             /* Allow 'import x from y' when a module doesn't have a default export. */
-    "esModuleInterop": false /* Emit additional JavaScript to ease support for importing CommonJS modules. This enables `allowSyntheticDefaultImports` for type compatibility. */,
```
asturur, line 76: Is this a default value? if it is, let's just comment it out.

Smrtnyk: Flag is removed completely from typescript

tsconfig.json
```diff
@@ -26,7 +26,7 @@
     /* Modules */
     "module": "ESNext" /* Specify what module code is generated. */,
     // "rootDir": "./",                                  /* Specify the root folder within your source files. */
-    "moduleResolution": "node" /* Specify how TypeScript looks up a file from a given module specifier. */,
```
asturur, line 29: This must have been a thing of the past when we were transitioning.

tsconfig.json
Smrtnyk: sorry my bad, its not removed but it is always enabled and you can't set it to false anymore
 i can bring it back as a comment if you want but i don't see this having value anymore

asturur: well ok if is not an option is not an option. things works

asturur (approved):

PR #10967, opened 2026-04-22
ci(dependabot): group vite-related npm updates

asturur:
This updates Dependabot's npm configuration so Vite-related packages are grouped into a single version-update PR instead of opening separate PRs for each package.

- add a `vite` Dependabot group for `vite`, `vitest`, `@vitejs/*`, and `@vitest/*`
- keep the existing weekly npm update schedule unchanged
- document the change in `CHANGELOG.md`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,11 @@ updates:
     directory: '/' # Location of package manifests
     schedule:
       interval: 'weekly'
+    groups:
+      vite:
+        applies-to: 'version-updates'
+        patterns:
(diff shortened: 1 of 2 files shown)


asturur:
@Smrtnyk could you review this when you have a moment?

Smrtnyk:
> @Smrtnyk could you review this when you have a moment?

I think we should separate vite from vitest, vitest has a wide range of vite support and can be handled separately

Smrtnyk (approved):

PR #10977, opened 2026-05-07
fix(): Honor viewport rotation in zoom, dimensions, and control coords

### LINKED ISSUE #10976: [Bug]: Viewport matrix with rotation breaks getZoom, _calculateCurrentDimensions, _renderControls, and calcOCoords
### Before submitting

- [x] I searched existing issues and discussions

### Version

7.3.1

### Environment

Chrome

### Link To Reproduction

_No response_

### Steps To Reproduce

1. Set the canvas viewport transform to a matrix encoding both rotation and zoom:
`canvas.setViewportTransform([cos45 * zoom, sin45 * zoom, -sin45 * zoom, cos45 * zoom, tx, ty])`
2. Place an interactive object (e.g. `Rect`) on the canvas and select it
3. Observe the selection border, control handles, and rotation handle

### Expected Behavior

- `canvas.getZoom()` returns the scale magnitude (`zoom`)
- The selection border tightly outlines the object at the correct rotated angle and size
- Corner/edge/rotation handles appear at the object's corners and edges
- The selection border is rotated to match the combined viewport + object angle

### Actual Behavior

Four separate bugs manifest when the viewport matrix encodes rotation (`vpt[1] ≠ 0`):

**1. `canvas.getZoom()` (dist line ~2091)**

Returns `vpt[0]` = `cos(θ) · zoom`, not `zoom`. Correct value is `Math.sqrt(vpt[0]² + vpt[1]²)`.

```js
canvas.setViewportTransform([cos45, sin45, -sin45, cos45, 0, 0]);
canvas.getZoom(); // → 0.707  (should be 1.0)
```

---

**2. `FabricObject._calculateCurrentDimensions()` (dist line ~5439)**

Applies the full rotated viewport to the dimension *vector* via `.transform(vpt, true)`, treating it as a direction rather than a scalar extent. At 45° the x and y components cancel and the selection border collapses to a
sliver.

```js
// 200×200 object, viewport rotation=45°, zoom=1:
obj._calculateCurrentDimensions(); // → Point { x: ≈0, y: ≈283 }  (should be Point { x: 200, y: 200 })
```

Fix: `_getTransformedDimensions().scalarMultiply(zoom).scalarAdd(2 * padding)` where `zoom = Math.sqrt(vpt[0]² + vpt[1]²)`.

---

**3. `InteractiveFabricObject.calcOCoords()` (dist line ~7643)**

Builds the final matrix with `[1/vpt[0], 0, 0, 1/vpt[3], 0, 0]` to undo the viewport scale. When rotation is present, `vpt[0] = cos(θ) · zoom`, so the undo factor becomes `1/(cos(θ) · zoom)` instead of `1/zoom`. At 45° rotation, every control is placed ~41% too far from the object.

Fix: replace `1/vpt[0]` with `1/zoom` where `zoom = Math.sqrt(vpt[0]² + vpt[1]²)`.

---

**4. `FabricObject._renderControls()` (dist line ~7764)**

Rotates the canvas context by `degreesToRadians(this.angle)` (object-only) instead of the combined viewport + object angle. The selection border appears axis-aligned regardless of viewport rotation.

Fix: use `options.angle` from `qrDecompose(multiplyTransformMatrices(vpt, this.calcTransformMatrix()))` instead of `this.angle`.

### Additional Context

All four bugs share the same root assumption: that `vpt[0]` equals the zoom scalar and `vpt[1]` equals zero — true only for pure scale+pan viewports.

Runtime patch workaround:

```js
// 1. getZoom
Reflect.set(canvas, 'getZoom', function() {
const vpt = this.viewportTransform;
return Math.sqrt(vpt[0] ** 2 + vpt[1] ** 2);
});

// 2. _calculateCurrentDimensions
Reflect.set(obj, '_calculateCurrentDimensions', function(options) {
const vpt = this.getViewportTransform();
const zoom = Math.sqrt(vpt[0] ** 2 + vpt[1] ** 2);
return this._getTransformedDimensions(options).scalarMultiply(zoom).scalarAdd(2 * this.padding);
});

// 3. calcOCoords — replace [1/vpt[0], 0, 0, 1/vpt[3]] with [1/zoom, 0, 0, 1/zoom] in finalMatrix

// 4. _renderControls — use qrDecompose(vpt × calcTransformMatrix()).angle instead of this.angle
```

kausters:
## Description

This PR fixes a set of bugs that surface when the canvas viewport transform encodes rotation (`vpt[1] !== 0`): `getZoom`, `_calculateCurrentDimensions`, `calcOCoords`, and `_renderControls` all assume `vpt[0]` equals the zoom scalar and `vpt[1]` is zero, which only holds for pure scale+pan viewports.

Closes #10976. The fixes are:

- `getZoom()` returns `Math.sqrt(vpt[0]**2 + vpt[1]**2)` (the magnitude of the viewport scale).
- `_calculateCurrentDimensions()` scales by that zoom magnitude instead of applying the full viewport as a direction (which collapsed dimensions under rotation).
- `calcOCoords()` uses `1 / zoom` for both axes in the inverse-scale matrix, so the rotation portion of the viewport is preserved when control offsets are placed.
- `_renderControls()` always rotates the context by `options.angle` from `qrDecompose(vpt × calcTransformMatrix())`, which already encodes the combined viewport + object rotation. The previous `this.group ? options.angle : this.angle` branch dropped viewport rotation in the ungrouped path.

Backwards-compatible for pure scale+pan viewports (the common case): `hypot([z, 0]) = z`, `scalarMultiply(z) ≡ transform([z, 0, 0, z, …], true)`, and `1/z` is unchanged. Each fix has a focused unit test that would fail against `master`.

## In Action

No visual change for the common case (pure scale+pan viewport). Under a rotated viewport such as `canvas.setViewportTransform([cos45, sin45, -sin45, cos45, 0, 0])`, selection borders, control handles, and reported zoom now match the object's true position and size instead of collapsing or drifting.

<table>
  <thead>
    <tr><th colspan="2">Fabric Canvas viewport transform rotated 45° clockwise</th></tr>
    <tr><th>Before</th><th>After</th></tr>
  </thead>
  <tbody>
    <tr>
      <td>before</td>
      <td>after</td>
    </tr>
  </tbody>
</table>


Smrtnyk:
would it be possible to add a visual test for this in playwright?

Smrtnyk:
[chromium] › e2e/tests/controls/rendering/index.spec.ts:7:5 › control box rendering ────────────
    

is failing, you should be able to run this locally
just build the fabric with your changes and run the e2e tests

kausters:
@Smrtnyk Fixed:
- https://github.com/fabricjs/fabric.js/pull/10977/commits/7354e54dc657de653a6c002c22297c4a56313a17 fixes the change that broke the existing test
- https://github.com/fabricjs/fabric.js/pull/10977/commits/490a08431c39b31b5ecbf7cccd111844127a83d8 adds a new visual test (new `viewport-rotation` suite, let me know if this works for you or if I should move / change it).

Thanks!

kausters:
Also @Smrtnyk locally I hit a flaky E2E test, not sure if it's persistent. `controls/hit-regions` has a pre-existing timing issue: the actual screenshot showed an empty canvas (rect hadn't rendered yet), then passed on retry and on all 3 follow-up runs. The test does ~480 000 `findControl` calls in a tight loop and races with the canvas render — unrelated to the fix.

asturur:
Hi @kausters thank you for the PR.
I'll review this asap, canvas rotation support  is an item we were always wanted to come back to, i would look at this PR as a way to start to support that across the board.

I had no idea that `return Math.sqrt(vpt[0] ** 2 + vpt[1] ** 2);` was  a thing. Do you have a link we can keep around that explains this from a geometric point of view? I understand is pitagora somehow, i m surprised of it using 0 and 1, whitout looking at what is in vpt[3].

The pr is good, i can tell you for sure you need reuse the matrix angle utility we have somewhere already and that you should make a similar utility for matrix magnitude ( or call it how it should be called ) and use that as well rather than repeating that formula in 3 places.

Then we need to understand what happens when the viewport does not have an uniform scaling, broken or not broken i would love a visual test and a unit test that fix the expectation for that case

kausters:
Hi @asturur 

**On `Math.sqrt(vpt[0] ** 2 + vpt[1] ** 2)`**

_Disclaimer: Claude helped me with this section as I'm not all that bright with geometry/trigonometry._

Take the point `(1, 0)` — one unit along the X axis. Apply the vpt to it and it lands at `(vpt[0], vpt[1])`. The distance from the origin to that landing point is `sqrt(vpt[0]² + vpt[1]²)` — straight Pythagoras. That distance is
how long "1 unit" becomes after the transform, which is exactly the zoom.

The old `vpt[0]` worked only because without rotation `vpt[1] = 0`, so `sqrt(vpt[0]² + 0²) = vpt[0]`. As soon as the vpt has rotation, `vpt[0]` is `cos(angle) · zoom` and Pythagoras over `(vpt[0], vpt[1])` is what recovers the
actual zoom. `vpt[2]/vpt[3]` would give the same number under uniform scaling — it's the same idea applied to the unit step along Y.

The general term for it is the *column norm* of the linear part of the matrix; closest readable reference if you want one for the codebase is QR / polar decomposition (https://en.wikipedia.org/wiki/Polar_decomposition), which is
also what `qrDecompose` does already.

**On reusing utilities**

Done in https://github.com/fabricjs/fabric.js/pull/10977/commits/0ed035358cccca0de4c9bb0e7b9ddd008a3645bf:

- Added `calcPlaneZoom` next to `calcPlaneRotation` in `src/util/misc/matrix.ts` (open to a different name if you prefer — `calcPlaneScale`?).
- `StaticCanvas.getZoom` now delegates to `calcPlaneZoom`.
- `InteractiveObject.calcOCoords` uses `calcPlaneZoom(vpt)` for the inverse-scale matrix.
- `InteractiveObject._renderControls` uses `calcPlaneRotation(vpt)` for the viewport angle.
- `ObjectGeometry._calculateCurrentDimensions` reads `this.canvas?.getZoom() ?? 1` instead of recomputing — `getViewportTransform()` stays public, just stops being called internally there.
- Added unit tests for `calcPlaneZoom`. Mock canvases in the spec suite now expose `getZoom()` to match the real surface (small spec-only churn)

**On non-uniform viewport scaling**

You're right that the current change silently collapses non-uniform scale. Pre-change `_calculateCurrentDimensions` applied `vpt[0]·w` and `vpt[3]·h` independently; `calcOCoords` inverted with `1/vpt[0]` and `1/vpt[3]`
independently. My fix replaces both with one scalar magnitude, which is wrong when `vpt[0] ≠ vpt[3]`.

`qrDecompose` already exposes `scaleX` / `scaleY` separately, so the principled fix is probably to drive the new utility off it (or off the column norms of *both* basis vectors) and feed per-axis scale into the two methods, then
`getZoom` becomes "the X-axis scale", with a note. I'm happy to ~either~:

1. Roll it into this PR — add a unit test that pins the per-axis expectation and a visual test for `vpt = [2, 0, 0, 3, ...]` and update the implementation to respect both axes; or
2. ~Land this PR as the rotation fix and open a follow-up for the non-uniform case so the rotation fix isn't blocked.~

~Let me know which you prefer.~ Rolled into the PR, see next comment.

kausters:
**Non-uniform viewport scaling** — rolled into this PR.

`calcPlaneScaleY` (column-1 norm of the vpt linear part) added alongside `calcPlaneZoom`. `_calculateCurrentDimensions` now multiplies x by `calcPlaneZoom(vpt)` and y by `calcPlaneScaleY(vpt)` independently; `calcOCoords` uses `[1/scaleX, 0, 0, 1/scaleY]` for the inverse matrix instead of a uniform `1/zoom`.

The visual test uses `[2·cos45, 2·sin45, −3·sin45, 3·cos45, …]` (45° rotation + scaleX=2 / scaleY=3) — the case that was actually broken. Pure non-uniform no-rotation (`vpt[0] ≠ vpt[3]`, no rotation) already worked before via
the old `vpt[0]`/`vpt[3]` path; the regression was the combination with rotation.

asturur:
The pr was fine also before, let's not scope creep it.
We are not going to support viewport skewX and skewY it does not make sense at all.

asturur:
I m also split on the non uniform zoom, becuase then there is the issue to say, how much is current zoom?
I think supporting zoom independent on the x/y axis creates more issue that it solves.

Are you using it in some way? Rotation was a long standing requested item

kausters:
Nah I just thought you wanted it done for completeness' sake (I may have skimmed the message too loosely) — I just care for the rotation because I needed it 😄 I can put the rest back how it was, or we can keep it. Non-uniform zoom is a fun one yes, because that _is_ theoretically doable and you can have separate scale on X and Y axis (in which case X should probably be the canonical one but it's kind of a wash).

asturur:
Let's move on.
I will need to add a small function to do canvas.setRotation(angleInDegrees) so that at this point is usable, then people will come with bugs.

asturur (approved):
ready to go

asturur:
i fix prettier, and then my approval is not good. Thank you @github

kausters:
Awesome, thank you guys — looking forward to this landing in a release!

asturur:
is on 7.4.0 i needed to ship a cve fix, so i shipped immediately

PR #10992, opened 2026-05-21
chore(): tackle sonar `Unexpected negated condition.` violations

Smrtnyk:
This patch tackles `Unexpected negated condition.` issues reported by sonar.
 An example `https://sonarcloud.io/project/issues?open=AZzysk6qWT_PiSiG0sy7&id=fabricjs_fabric.js`


Smrtnyk:
sonar complains about code coverage, will try to add some meaningful tests to increase it
even though my changes were just fliping the conditions i think its good to add some more tests

asturur:
I m not against this, but this should be like enforced by lint if is an issue? Why is this suggested, any idea?

Smrtnyk:
> https://sonarcloud.io/project/issues?open=AZzysk6qWT_PiSiG0sy7&id=fabricjs_fabric.js

yeah eslint has the no negated condition rule, i can activate it
it is basically about code readability, brains are wired to read better the positive conditions first
I personally like this rule, but it is subjective
either we enforce this, or i turn off this sonar rule so we reduce noise
I leave it up to you

asturur:
my only opinion is that if we apply it needs to be lintend, so if you like you can totally apply it i have no issue with that

Smrtnyk:
> my only opinion is that if we apply it needs to be lintend, so if you like you can totally apply it i have no issue with that

will enable a related eslint rule and push

Smrtnyk:
I have enabled the lint rule, the lint fix run made some new changes, where redundant assertions are removed
probably because of recent eslint update it is now smarter to infer types

src/canvas/Canvas-events.spec.ts
```diff
@@ -958,7 +958,7 @@ describe('Canvas events mixin', () => {
       }
 
       if (eventType === 'mouseenter') {
-        eventname = 'mouse:over' as keyof CanvasEvents;
```
asturur, line 961: yes i noticed those in my prs too and i don't mind. The only thing i m worried is that all this inferring will cost speed? vscode is slow as crap in general.

Hopefully this typescript go will arrive soon, or an OXTS who knows

asturur (approved):

PR #11008, opened 2026-06-04
chore(): try native oxc minifier

Smrtnyk:
removes terser, leaving oxc native minifier only

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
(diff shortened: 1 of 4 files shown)


Smrtnyk:
without terser its 1kb larger in minified mode

Smrtnyk:
@asturur if this 1kb is acceptable to you we could get rid of terser, you decide

asturur:
> @asturur if this 1kb is acceptable to you we could get rid of terser, you decide

With the idea of splitting in more packages, size per se is not a benchmark anymore., is ok.

asturur (approved):

PR #11009, opened 2026-06-04
chore(): sonar optional chain

Smrtnyk:
Tackles sonar `Prefer using an optional chain expression instead, as it's more concise and easier to read.` and enforces it via eslint
 Optional chain feature is old enough to support it natively in fabric https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#browser_compatibility

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,7 +1,7 @@
 # Browsers that we support
 # Picked up for fabric 7.0 In April 2025 Browsers released jan 2021
 chrome >= 88
-safari >= 13
+safari >= 13.1
 firefox >= 85
 edge >= 88
 Opera >= 73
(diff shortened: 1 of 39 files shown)


Smrtnyk:
> **Build Stats**
> 
> fabric	778.217 (**+5.286**)	292.697 (**+0.631**)

wasn't expecting an increase, seems like we are transpiling it
maybe because in our browserlists we have safari 13 but optional chain is from safari 13.1

Smrtnyk:
ok this fails on node 22, i guess i will exclude optional chain from sonar with a different pr
we can't do this right now

asturur:
Seems like optional chaining is way older than node22. Any insight on why it doesn't work?

Smrtnyk:
> Seems like optional chaining is way older than node22. Any insight on why it doesn't work?

after checking deeper it seems that after I set the es2020 for transform target it also transformed less stuff and something else broke it
but anyway, I already left a comment, this can be controled via lint, we can reduce noise in sonar for this

PR #11011, opened 2026-06-05
feat(): monorepo  first steps with backwards compat

### LINKED ISSUE #10894: Improve repo structure with sub packages
This is probaly a major effort.
the extensions folder we have today is an hack and solves only partially the issue.

Ideally we end up with a packages folder where there is a core package that containers fabricjs as it is today, and a small package per extension.
This should be doable with yarn or pnpm but it needs to be investigate.
Also is probable that fabricJS package requires a name change in order to support @fabricjs/core and @fabricjs/gradient-controls just to make an example.

This would allow for easier pick up from developers, and every package could have its own small readme file and would be displayed easily in the native github interface. On top of that we don't have to publish everything every time, if you need to add a new package you don't need to update fabricjs for it.

Every package should reference fabricJS as a depedency and not bundle it during build.

Each package outside core would have only a modern build, no umd file for scripts tags in quick examples.
We would drop some convenience for modernity.

Typescript / lint / prettier configs needs to be the same, hopefully referenced from core and just reused.
Each package needs to have its own test command, but should be a single test command that runs them all for the CI.

With time then the github actions files can be instructed to run specific actions only on the packages that actually received changes, exactly as we do for changelog.md

e2e tests needs to be split as well, and when fabricJS is changed, everything needs to be checked, but when an extension is changed only its subset of tests can be run ( probably )

The main website on fabricJS.com then needs to be updated to handle it.

This is a disruptive change but better now than later when there are more im

Smrtnyk:
introduced public 
@fabricjs/core
@fabricjs/browser
@fabricjs/node
internal 
@fabricjs/e2e for e2e tests that consume new packages

and 
@fabricjs/*extension-name* for every extension


Smrtnyk:
@asturur I did some ground work for monorepo structure
introducing new packages
the build command still builds root fabric package with its entrypoints like browser node and extensions
that should be intact, but this also introduces new packages in monorepo which also build independently
I added e2e package that is private and for playwright tests, that import these new scoped packages and individual extensions
all tests work and pass, they were all moved using git mv
It would be good if you could check this branch out test it, inspect build artifacts etc... do the changes if you want
I suppose this could be merged as a first step and in other PRs to do the migration of the core fabric into `@fabric/core` and not sure what your vision is and how you want to publish under which versions 
I also added some smoke tests to verify the built artifacts

There is no hurry for a review, this is a bigger PR, so whenever you find time 
I will in the mean time go back to usual routine, check sonar, see if there are some minor bugs to be fixed etc...

asturur:
i ll be fully focusing on this soon, i was supposed to start it, i had a bunch of codex work done but you started before. I ll do some basic questions before reading all the work.

My first question is that i see that most of the code stay in SRC ( because i don't see diffs i assume files didn't move ). Should that be core package?

Git history seems good, files are recognized as moved rather than deleted and re-created.

Smrtnyk:
> My first question is that i see that most of the code stay in SRC ( because i don't see diffs i assume files didn't move ). Should that be core package?

well, I myself am not sure, I suppose so, that src should go to the core fully, unless there is some separation to be done to move to browser and node packages and core gets some less stuff in

so far, what this achieves is that it adds these 3 packages, but keeps the root fabric untouched, and it can be reorganized in a 2nd step
Ideally by you since you know the fabric the best and what should go to which package and how it should be separated

currently core package is just a wrapper around the root src, and when built in the dist it contains the same thing the root fabric dist has
so it is a full backwards compat package without moving the src files yet

Smrtnyk:
btw sorry for not syncing with you before starting this
If you prefer the work you have done so far please feel free to discard this PR
It is not that i invested much time into it, so do not worry about that
it was mostly just moving things around

asturur:
> btw sorry for not syncing with you before starting this If you prefer the work you have done so far please feel free to discard this PR It is not that i invested much time into it, so do not worry about that it was mostly just moving things around

Yeah there isn't any sync requirement here. This was discussed with an issue before, we wanted to do this, who starts, starts.

The only rules are let's not merge things without human reviews and if something we don't like in code or attitude we call it out clearly and with explanations asap.

extensions/linear_gradient_controls/linearGradientControls.ts
```diff
@@ -1,40 +1,9 @@
...
-import { Control, controlsUtils } from 'fabric';
-import {
-  linearGradientColorActionHandlerGenerator,
-  linearGradientColorPositionHandlerGenerator,
-  linearGradientControlLineRender,
-  linearGradientCoordPositionHandlerGenerator,
-  linearGradientCoordsActionHandlerGenerator,
-} from './linearGradientHandlers';
+import type { Control, Gradient } from 'fabric';
+import { createLinearGradientControls as createCoreLinearGradientControls } from '../../packages/gradient-controls/src/linearGradientControls';
```
asturur, line 2: i picked this file as an example, but should those imports being package scoped?

e2e/tests/visual-output/rendering/index.spec.ts-snapshots/svg-import/nested-clippath-triple.png
asturur, line 1: do you know why we lost those 2?

Smrtnyk: we merged this duplicate nested paths bugfix that added these 2, and after rebase these were missing in the new location, so i ran the e2e tests in the new package and now they live as new files in the e2e package

will see if i can rewrite history somehow to show them as moved

extensions/linear_gradient_controls/linearGradientControls.ts
Smrtnyk: it was for backwards compat, since all extensions are moved to the packages but the fabric/extensions build was preserved to still have all the extensions, so rollup can pull it from the source
I am not sure what the plan is, but for now fabric and fabric/extensions, fabric/node and fabric/browsers are still preserving its dist

in the future when the scoped fabric packages are ready to be published, we should point to the new scoped packages of course
as i said best is if you could check out this branch run the builds and check the artifacts, and then do futrther file migrations to sort the scoped packages how you envision them

asturur (dismissed):
I read this a bunch of time and i asked also CODEX to guide me in understanding the future from on with the monorepo.

There are github actions to change and the compatibility package needs to be done differently, but i don't want you to stay with a large PR open so long.

Please fix the conficts and we can merge. then i ll proceed with the codex suggestions in a followup.

there are 2 important security fixes to do, if we see it takes too much time to make this new packages publishable i branch off 7.4.0 and i add the security fixes there and publish with that old branch

Smrtnyk:
> I read this a bunch of time and i asked also CODEX to guide me in understanding the future from on with the monorepo.
> 
> There are github actions to change and the compatibility package needs to be done differently, but i don't want you to stay with a large PR open so long.
> 
> Please fix the conficts and we can merge. then i ll proceed with the codex suggestions in a followup.
> 
> there are 2 important security fixes to do, if we see it takes too much time to make this new packages publishable i branch off 7.4.0 and i add the security fixes there and publish with that old branch

Hi Asturur, I don't have access to my laptop until sunday late in the evening. If you are able to rebase (basically accept pnpm lock and run pnpm install to regenerate) that should be it and I can approve

Smrtnyk:
I actually had time to do it from brothers computer

asturur (approved):

asturur:
Ok i will start the PR i planned with the codex investigation that will move the src files and flip the compatibility package and update the publishing action.

PR #11012, opened 2026-06-08
fix(sandbox): restore vanilla startup with pnpm

asturur:
Fixes the vanilla CodeSandbox startup path after the pnpm migration by removing an undeclared transitive dependency and preventing watcher setup failures from crashing startup.

- Use the existing color helper instead of undeclared chalk.
- Only attach the build-status reload watcher when Fabric watch is enabled and handle watcher errors.
- Scope the vanilla Parcel watcher to the sandbox directory.


asturur:
@Smrtnyk while looking again at why nested clippaths fails on my setup, i found i broke this with pnpm and i need to have it working

Smrtnyk (dismissed):

Smrtnyk (approved):

PR #11028, opened 2026-06-26
fix(text): avoid spurious measuring in _measureChar when only kerning-irrelevant entries are missing

jiayihu:
_Disclosure: I found out the issue during work but I fully asked GLM 5.2 to port the fix back to fabric, add tests and open the PR using the repo skill._

## Summary

`_measureChar` was treating missing `previousWidth` or `coupleWidth` as a cache miss even when kerning was irrelevant (no usable previous char, or previous char with a different style). On a pure single-character cache hit this forced an unnecessary fetch of the measuring context and extra `_setTextStyles`/`measureText` bookkeeping. The fix only builds and reads the kerning pair when a same-style previous char exists, and only enters the measuring path when the single-character width is missing or a same-style kerning pair actually needs measuring.

## Changes

- `src/shapes/Text/Text.ts`: rewrote the `_measureChar` cache/measure logic so the couple key and `previousWidth` are only built/read when `previousChar` exists, and the measuring path is gated on `width === undefined || (stylesAreEqual && coupleWidth === undefined)`.
- `src/shapes/Text/Text.spec.ts`: added two tests covering the regression — one asserting `_setTextStyles` is not called when only kerning-irrelevant entries are missing, and one asserting `measureText` is invoked for `previousWidth` and `coupleWidth` (and not for an already-cached single char) when both are undefined.

### DIFF
--- a/src/shapes/Text/Text.ts
+++ b/src/shapes/Text/Text.ts
@@ -825,7 +825,7 @@ export class FabricText<
     const fontCache = cache.getFontCache(charStyle),
       fontDeclaration = this._getFontDeclaration(charStyle),
       fontMultiplier = charStyle.fontSize / this.CACHE_FONT_SIZE;
-    const hasPrevChar = Boolean(previousChar),
+    const hasPrevChar = !!previousChar,
       stylesAreEqual =
         hasPrevChar &&
         fontDeclaration === this._getFontDeclaration(prevCharStyle),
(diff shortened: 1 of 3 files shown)


src/shapes/Text/Text.ts
```diff
@@ -824,18 +824,19 @@ export class FabricText<
...
+      couple = hasPrevChar ? previousChar + _char : _char;
     let width: number | undefined,
       coupleWidth: number | undefined,
       previousWidth: number | undefined,
       kernedWidth: number | undefined;
 
-    if (previousChar && fontCache.has(previousChar)) {
-      previousWidth = fontCache.get(previousChar);
+    if (hasPrevChar && fontCache.has(previousChar!)) {
+      previousWidth = fontCache.get(previousChar!);
```
asturur: is incredible that TS still can't catch this.

jiayihu: My bad (or GLM's so ultimately mine eheh). It was using `Boolean`, which defeats the TS control flow analysis. Using raw `!!` preserves it. Still needed for Maps though as there is no control flow analysis for `Map.has`

PR #11029, opened 2026-06-26
fix(text): skip initDimensions in set when a textLayoutProperty value did not change

jiayihu:
_Disclosure: I found out the issue during work but I fully asked GLM 5.2 to port the fix back to fabric, add tests and open the PR using the repo skill._

## Summary

`FabricText.set` (and subclasses like `Textbox`/`IText`) was requesting `initDimensions()`/`setCoords()` whenever a key in `textLayoutProperties` was passed to `set`, regardless of whether the value actually changed. Setting a layout property to the same value it already held therefore triggered an unnecessary text re-measurement and coord recomputation. The fix ports the value-changed check (already used in `Polyline._set`) into `Text.set` so `initDimensions` is only requested when a layout-relevant value differs from the current one, while keeping the existing code path that delegates to `super.set` (which calls `_set`).

## Changes

- `src/shapes/Text/Text.ts`: in `set`, compute `needsDims` by comparing `this[key] !== value` before `super.set` mutates the instance (for both the string-key and object-key branches). `super.set` is still called for all keys, and `setPathInfo()` still runs after the set when a `path` key is present.
- `src/shapes/Text/Text.spec.ts`: added two tests under a `set & initDimensions` describe block — one for the single-key `set('text', ...)` no-op vs. change case, and one for the object `set({ text, fontSize, charSpacing })` no-op vs. change case — spying on `initDimensions`/`setCoords` to assert the skip behavior.

### DIFF
--- a/src/shapes/Text/Text.ts
+++ b/src/shapes/Text/Text.ts
@@ -1823,8 +1823,9 @@ export class FabricText<
       needsDims =
         textLayoutProperties.includes(key) && this[key as keyof this] !== value;
     }
-    // keep the existing fabric code path that calls _set via super.set
+
     super.set(key, value);
+
     // setPathInfo must run after super.set so it reads the new path
     if (isAddingPath) {
       this.setPathInfo();
--- a/src/shapes/Text/Text.ts
+++ b/src/shapes/Text/Text.ts
@@ -1806,21 +1806,26 @@ export class FabricText<
 
   set(key: string | any, value?: any) {
     const { textLayoutProperties } = this.constructor as typeof FabricText;
-    super.set(key, value);
+    // detect layout-relevant changes before super.set mutates the instance,
+    // so initDimensions is only requested when a value actually changed.
     let needsDims = false;
     let isAddingPath = false;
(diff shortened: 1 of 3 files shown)


src/shapes/Text/Text.ts
```diff
@@ -1806,21 +1806,27 @@ export class FabricText<
...
       for (const _key in key) {
-        if (_key === 'path') {
-          this.setPathInfo();
-        }
-        needsDims = needsDims || textLayoutProperties.includes(_key);
         isAddingPath = isAddingPath || _key === 'path';
+        needsDims =
+          needsDims ||
+          (textLayoutProperties.includes(_key) &&
+            this[_key as keyof this] !== key[_key]);
```
asturur, line 1819: 👍

asturur:
I just merged a breaking change with the publishing system, i m not able to publish this, if you want this out we need to fork the current release and make a release branch.

asturur (approved):

jiayihu:
No hurry, we always directly patch or override methods at work, then I just port over the fixes to fabric when I have time

PR #11033, opened 2026-06-30
chore(): Monorepo follow up steps

asturur:
## Description

Adds the follow-up monorepo publishing hardening: public package metadata, recursive workspace publishing, CI/release package smoke coverage, and stronger tarball/runtime identity checks.

## Key changes

- Adds public publish config for publishable @fabricjs workspace packages.
- Publishes workspace packages recursively from the npm publish workflow.
- Runs package smoke checks in build and publish workflows.
- Extends package smoke to inspect real tarballs, verify @fabricjs/core stays free of node runtime deps, and catch duplicate core runtime/class identities.
- Keeps root fabric as the legacy package for now; facade conversion is a follow-up.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,8 @@ jobs:
         with:
           node-version: 24.x
       - run: pnpm run build
+      - name: Smoke test packages
+        run: pnpm run smoke:packages
   stats:
     name: Build stats
     runs-on: ubuntu-24.04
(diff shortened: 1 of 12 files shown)


asturur:
This is a first PR.
Then i ll follow up with separation of src/node/browser, and move the src folder in a package ( the core package ).

Smrtnyk (approved):

PR #11039, opened 2026-07-07
refactor(env): move runtime env setup to packages

asturur:
Refactors Fabric environment setup so the core package owns only the environment holder, types, and setter/getter APIs, while the browser and node packages own their runtime-specific environment implementations.

The PR is small on purpose.
This PR is not moving core files yet and so imports from @fabric/node to src are still relative.

Key changes:
- Add `setEnvFactory` to the core env holder and remove browser/node runtime imports from `src/env`.
- Move browser env setup into `packages/browser/src/env.ts` and install it lazily from browser/root browser entrypoints.
- Move node env setup and `NodeGLProbe` into `packages/node/src`, keeping `@fabricjs/node` and `fabric/node` responsible for jsdom/canvas.
- Stage `@fabricjs/core` from a neutral core bundle so the published core package does not include runtime env implementations.
- Update generated root dist artifacts and the package changelog.

Validation:
- `npm run lint` passes, with existing warnings.
- `npm run prettier:check` is blocked locally by unrelated untracked `e2e/` report files; the PR files pass `oxfmt --check`.
- `tsc -p ./tsconfig.build.json --noEmit`
- `tsc -p ./tsconfig-extensions.json --noEmit`
- `tsc -p ./tsconfig.packages.json --noEmit`
- `node ./scripts/package-smoke.mjs`
- `rg "jsdom|from 'canvas'|src/env/node|NodeGLProbe" packages/core src/env`
- Browser package import smoke without global `window`/`document`


Smrtnyk (approved):

Smrtnyk:
we should also at one point move jsdom dep to the node subpackage, i guess that comes later?

asturur:
> we should also at one point move jsdom dep to the node subpackage, i guess that comes later?

I m following a plan step by step to avoid doing mess, but yes, the idea at the end is total separation for node and browser. As soon as i can do that 'fabric/node' is jut a re-export of @fabricjs/node, then i can move them.

fabric package will still be requiring those, but it will be a legacy pacakge. if you install @fabricjs/browser in theory you won't have any node dep in your logs.

asturur:
Ok since you approved i move to next step right away.

PR #11040, opened 2026-07-07
fix(config): resolve device pixel ratio through env

asturur:
Resolve device pixel ratio through the environment boundary instead of reading browser globals from core configuration defaults.

- Leave config.devicePixelRatio unset by default so it acts as an explicit override.
- Resolve DPR at usage time with the compact config/env-window/1 fallback expression.
- Cover configured, browser-env, and fallback DPR behavior in env tests.
- Ignore generated top-level e2e report output for git and prettier checks.

related to #10894


asturur:
ok next PR should be the large one. Where i move all src files

Smrtnyk (approved):

.gitignore
```diff
@@ -24,6 +25,7 @@ before_commit
 /cli_output/
 /packages/e2e/test-report/
 /packages/e2e/test-results/
+/e2e/
```
Smrtnyk, line 28: are you ignoring the whole e2e package?

asturur: the package has moved in packages/e2e but i still had the folder full of test-report and test-results, as soon as .gitignore was modified my git wanted to track those files again and again, so i cut it out. everything is in packages/e2e in theory

PR #11041, opened 2026-07-08
refactor(core): move implementation source into package

asturur:
Move core source ownership into `packages/core/src` so `@fabricjs/core` builds from its own workspace source instead of depending on the root implementation layout.

- Move the current core implementation tree from root `src/*` into `packages/core/src/*` and move `fabric.ts` exports into `packages/core/src/index.ts`.
- Update package imports, tsconfig paths, rolldown inputs, and build/package scripts so `@fabricjs/core` builds like a normal workspace package.
- Keep the root entrypoints as compatibility facades for now and preserve existing import/export names and behavior.
- Remove the special core dist staging flow and point package smoke/build expectations at the package-owned source and output.
- Stack this PR on top of #11040; this should merge after that PR lands, then be retargeted or rebased onto `master`.

related to #10894


asturur:
This was more complicated than i thought, and i need to test it properly.
I don't think i want to maintain the old `import { Canvas } from 'fabric'` as soon as this one is ready. Too much noise in my opinion

Smrtnyk:
> This was more complicated than i thought, and i need to test it properly. I don't think i want to maintain the old `import { Canvas } from 'fabric'` as soon as this one is ready. Too much noise in my opinion

we could release new major and say what are the new entrypoints?

asturur:
> > This was more complicated than i thought, and i need to test it properly. I don't think i want to maintain the old `import { Canvas } from 'fabric'` as soon as this one is ready. Too much noise in my opinion
> 
> we could release new major and say what are the new entrypoints?

I think we want to do at least one last version in which i publish both, so i test everything ends up being the same, before putting out a new major i want to be sure all examples on fabricjs.com import from the new package, so i m confident things just work

Smrtnyk (dismissed):

asturur:
@Smrtnyk this is fully broken i had an hard week and couldn't look into it, i hope i can do this weekend so i unblock the repo.

Smrtnyk:
> @Smrtnyk this is fully broken i had an hard week and couldn't look into it, i hope i can do this weekend so i unblock the repo.

no worries, I noticed that build is red and that you have to do more changes
just put my approval as a signal that I am watching the progress on monorepo

asturur:
ok the way playwright build the code needs to be checked out. i have been focusing on work and hobbies these days, but i ll come back here tonight.

asturur:
@Smrtnyk can you give it whenever you have time, no rush, a good read, and see if you spot something fishy or missing?

asturur:
outside the eraser brush, which changes were unplanned, but are ok, all the rest should be moved only or imports changes.

Smrtnyk:
> outside the eraser brush, which changes were unplanned, but are ok, all the rest should be moved only or imports changes.

I had a look, nothing seems off so far

Smrtnyk (approved):

PR #11044, opened 2026-07-16
fix(parser): match exact class name in selectorMatches (#8838)

### LINKED ISSUE #8838: [Bug]: svg dom parse error about className
### CheckList

### Version

5.3.0

_No response_

None

### Link To Reproduction

https://github.com/fabricjs/fabric.js/blob/master/src/parser/selectorMatches.ts

### Steps To Reproduce

When parsing SVG elements, if the class order of an element is "st12 st1", the style of st12 will be lost. In fact, when examining the code, it will be found that st12 hits st1 during the parsing process and leaves a "2". It cannot proceed.

[error code link](https://github.com/fabricjs/fabric.js/blob/master/src/parser/selectorMatches.ts)

### Expected Behavior

Correctly parse the style.

### Actual Behavior

lose style

```bash
none
```

xianjianlf2:
## Description

Closes #8838.

When an SVG element carries multiple classes such as `class="st12 st1"`, the `.st12` style rule was dropped and only `.st1` applied.

The selector matcher in `src/parser/selectorMatches.ts` used a negative lookahead `(?![a-zA-Z\-]+)` to make sure a class/id name was not merely a prefix of a longer name. That lookahead excluded letters and hyphens, but not digits or underscores, so `.st1` matched inside `.st12`: the trailing `2` slipped past the lookahead, leaving a leftover token and causing the `.st12` rule to be discarded.

The fix replaces the lookahead with `(?![\w\-])`, which excludes every valid CSS identifier continuation character (letters, digits, underscore and hyphen), so numeric- and underscore-suffixed class and id names are matched exactly. This matches the exact-class-name approach @asturur prescribed in the issue.

Added `src/parser/selectorMatches.spec.ts` verifying that `.st1` no longer matches an element whose class list contains `st12`, while genuine matches still succeed. Existing parser specs remain green.

## In Action

Before: an element with `class="st12 st1"` lost its `.st12` styling because `.st1` matched inside `st12`.
After: `.st12` and `.st1` are matched independently and both rules apply.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [next]
 
 - refactor(core): move implementation source into package [#11041](https://github.com/fabricjs/fabric.js/pull/11041)
+- fix(parser): match class/id selectors exactly so `.st12` is not lost when `.st1` also exists [#8838](https://github.com/fabricjs/fabric.js/issues/8838)
 - fix(config): resolve device pixel ratio through env [#11040](https://github.com/fabricjs/fabric.js/pull/11040)
 - refactor(env): move runtime env setup to packages [#11039](https://github.com/fabricjs/fabric.js/pull/11039)
 - chore(): Monorepo follow up steps [#11033](https://github.com/fabricjs/fabric.js/pull/11033)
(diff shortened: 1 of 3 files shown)


asturur:
The PR looks good. i have to check recent changes to those regex because some were simplifications at the cost of small edge cases to avoid security issues in node. Thanks for the PR meanwhile

xianjianlf2:
Thanks for checking it. That tradeoff makes sense to me. I kept this narrowly scoped to the class-token false positive from #8838, but happy to adjust the matcher if the recent regex simplifications need a stricter shape for the Node/security side.

PR #11046, opened 2026-07-25
Refresh text measuring context when env document changes

### LINKED ISSUE #10930: Text measurement keeps stale canvas/document context after setEnv() switches document
### Before submitting

- [ ] I searched existing issues and discussions

### Version

7.2.0

### Environment

Chrome, Safari, Firefox

### Link To Reproduction

_No response_

### Steps To Reproduce

1. Create an iframe-based preview/editor.
2. Point Fabric to the iframe environment with `fabric.setEnv({ ...fabric.getEnv(), window, document })`.
3. Create/render a `fabric.Textbox` and trigger text measurement.
4. Destroy that iframe and create a new one.
5. Call `fabric.setEnv()` again with the new iframe `window/document`.
6. Render or measure the same text again.

In our app this reproduces consistently after:
1. opening the editor,
2. viewing banner text,
3. leaving the editor,
4. opening the editor again.

### Expected Behavior

After `fabric.setEnv()` switches to a new `document`, Fabric should recreate the internal text measuring canvas/context for the current document.

Text metrics should stay correct after iframe recreation or editor reopen.

### Actual Behavior

Fabric keeps using the original cached measuring context created for the first document/window.

After the iframe is destroyed and recreated, text measurement still uses the stale context from the old document. This causes incorrect text metrics, and text can visually collapse / "stick together" after reopening the editor.

### Additional Context

This looks related to the module-level `measuringContext` cache used by text measurement.

The context is created once and then reused forever, even after `fabric.setEnv()` changes the active document. In iframe-based apps this means the measuring context may belong to a destroyed iframe window/document.

A fix like this resolves the issue on our side:

```ts
let measuringContext: CanvasRenderingContext2D | null;
let measuringContextDocument: Document | null;

function getMeasuringContext() {
  const fabricDocument = getFabricDocument();

  if (!measuringContext || measuringContextDocument !== fabricDocument) {
    const canvas = createCanvasElementFor({ width: 0, height: 0 });
    measuringContext = canvas.getContext('2d');
    measuringContextDocument = fabricDocument;
  }

  return measuringContext;
}

xianjianlf2:
Fixes #10930.

Fabric caches a module-level text measuring context the first time text metrics are needed. If `setEnv()` later points Fabric at a new `document`, for example after an iframe-backed editor is destroyed and recreated, text measurement continues using the canvas context from the previous document.

This stores the document used to create the cached measuring context and recreates the canvas/context when the active Fabric document changes. The existing cache behavior is unchanged while the env document stays the same.

I added a regression test that measures text, switches Fabric to an iframe document, measures again, and verifies a new measuring canvas is created in the new document.

Tests run:
- `pnpm exec vitest --run --project unit-node packages/core/src/shapes/Text/Text.spec.ts`
- `pnpm exec oxfmt --check packages/core/src/shapes/Text/Text.ts packages/core/src/shapes/Text/Text.spec.ts`
- `pnpm exec eslint packages/core/src/shapes/Text/Text.ts packages/core/src/shapes/Text/Text.spec.ts`
- `pnpm run typecheck`
- `pnpm run test:vitest`

Note: the full unit test command passes; jsdom logs expected image loading errors from existing Image specs during the run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- Refresh text measuring context when env document changes [#11046](https://github.com/fabricjs/fabric.js/pull/11046)
 - refactor(core): move implementation source into package [#11041](https://github.com/fabricjs/fabric.js/pull/11041)
 - fix(config): resolve device pixel ratio through env [#11040](https://github.com/fabricjs/fabric.js/pull/11040)
 - refactor(env): move runtime env setup to packages [#11039](https://github.com/fabricjs/fabric.js/pull/11039)
--- a/packages/core/src/shapes/Text/Text.spec.ts
+++ b/packages/core/src/shapes/Text/Text.spec.ts
@@ -4,11 +4,13 @@ import { config } from '../../config';
 import { Path } from '../Path';
 import { FabricText } from './Text';
 
-import { describe, expect, it, afterEach } from 'vitest';
+import { describe, expect, it, afterEach, vi } from 'vitest';
 import {
   FabricObject,
+  getEnv,
   getFabricDocument,
   IText,
+  setEnv,
   Textbox,
 } from '../../../../../fabric';
 import { toFixed } from '../../util';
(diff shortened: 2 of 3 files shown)


asturur (approved):

asturur:
Thank you this is a nice fix

PR #11065, opened 2026-07-31
test(), chore(): Various things i forgot

asturur:
## Description

Swap warp.dev ended sponsorship with atlasclloud
add a unit test for the server side callback
remove leftover files
Fix a test that needs to be rounded or breaks with new playwright (. chrome changed something numbers round diffrently )


asturur:
@Smrtnyk this is a mix n match pr for 3-4 things i had open in my list.

Smrtnyk (approved):
looks good

PR #11072, opened 2026-08-07
fix(path): draw an arc with a zero radius as a line instead of dropping it

Jaybhade:
## Description

`arcToSegments` returns `[]` when `rx` or `ry` is `0`, so `makePathSimpler` drops the
segment entirely and the pen jumps to the arc's endpoint without drawing anything.

The SVG spec treats a zero radius and a zero-length arc as two different things
([implementation notes, F.6.2 / "Out-of-range parameters"](https://www.w3.org/TR/SVG/implnote.html#ArcOutOfRangeParameters)):

> If rx = 0 or ry = 0 then this arc is treated as a straight line segment (a "lineto") joining the endpoints.
> […] If the endpoints (x1, y1) and (x2, y2) are identical, then this is equivalent to omitting the elliptical arc segment entirely.

fabric applies the second rule to both cases, so a vertex silently disappears. On a
closed path the `Z` then closes from the wrong point and the filled region changes
shape. For `M 0 0 A 0 0 0 0 0 10 10 L 20 0 Z`:

| | commands | bounding box |
|---|---|---|
| Chrome / Firefox | `M`, line to (10,10), `L`, `Z` | 20 × 10 |
| fabric before | `[['M',0,0], ['L',20,0], ['Z']]` | 20 × **0** |
| fabric after | `[['M',0,0], ['C',…,10,10], ['L',20,0], ['Z']]` | 20 × 10 |

The whole shape collapses to a flat line today.

The fix emits the line as a cubic with its control points at ⅓ and ⅔, which is exactly
a straight segment, so `fromArcToBeziers` keeps its documented
`TParsedAbsoluteCubicCurveCommand[]` return type and every caller is unchanged. The
identical-endpoint check moves above it so that rule still wins when both apply — that
case reached `[]` before only because the maths went `NaN`; it is now explicit and its
behaviour is unchanged.

`Path.spec.ts > can parse arcs with rx and ry set to 0` goes from 9 commands to 10. That
path (a rounded rect out of a design tool) has two zero-radius arcs: `a0,0,0,0,1,0,0`
ends where it starts and is still dropped, while the trailing
`A0,0,0,0,1,62.87543,168.19448` moves the point by 4e-5 and is now drawn, matching what
browsers do. Its geometry is untouched — `left/top/width/height` are identical before
and after (`70.813525, 172.38568, 15.87627, 8.3824`) and match Chrome's `getBBox()`.

## In Action

Verified against Chrome via `getTotalLength()` / `getBBox()` on the same path data:

```
M 0 0 A 0 5 0 1 0 10 10          -> len=14.1421  bbox=(0,0,10,10)      # straight line
M 0 0 A 5 0 0 1 0 10 10          -> len=14.1421  bbox=(0,0,10,10)      # straight line
M 0 0 A 0 5 0 1 0 0 0            -> len=0.0000   bbox=(0,0,0,0)        # omitted
M 0 0 A 0 0 0 0 0 10 10 L 20 0 Z -> len=48.2843  bbox=(0,0,20,10)      # closed shape
```

Tests added in `packages/core/src/util/path/index.spec.ts` cover the absolute and
relative forms of a zero-radius arc and both zero-length cases. `pnpm run test:vitest`
(2045 passed), `typecheck`, `lint` and `prettier:check` are green.

### DIFF
--- a/packages/core/src/util/path/index.ts
+++ b/packages/core/src/util/path/index.ts
@@ -10,6 +10,7 @@ import type {
   TCurveInfo,
   TComplexPathData,
   TParsedAbsoluteCubicCurveCommand,
+  TParsedAbsoluteLineCommand,
   TPathSegmentInfo,
   TPointAngle,
   TSimpleParsedCommand,
@@ -96,13 +97,6 @@ const arcToSegments = (
   sweep: number,
   rotateX: TRadian,
 ): TParsedAbsoluteCubicCurveCommand[] => {
-  if (toX === 0 && toY === 0) {
-    return [];
-  }
-  if (rx === 0 || ry === 0) {
-    // an arc with a zero radius is a straight line to the endpoint
-    return [['C', toX / 3, toY / 3, (toX * 2) / 3, (toY * 2) / 3, toX, toY]];
-  }
   let fromX = 0,
     fromY = 0,
     root = 0;
@@ -318,7 +312,8 @@ export function getBoundsOfCurve(
 }
 
 /**
- * Converts arc to a bunch of cubic Bezier curves
+ * Converts arc to a bunch of cubic Bezier curves, or to a single line for the
+ * degenerate arcs that SVG defines as straight
  * @param {number} fx starting point x
  * @param {number} fy starting point y
  * @param {TParsedArcCommand} coords Arc command
@@ -327,7 +322,15 @@ export const fromArcToBeziers = (
   fx: number,
   fy: number,
   [_, rx, ry, rot, large, sweep, tx, ty]: TParsedArcCommand,
-): TParsedAbsoluteCubicCurveCommand[] => {
+): (TParsedAbsoluteCubicCurveCommand | TParsedAbsoluteLineCommand)[] => {
+  // SVG 2 §9.5.1: an arc whose endpoints coincide is dropped, and one with a
+  // zero radius is drawn as a line to its endpoint
+  if (tx === fx && ty === fy) {
+    return [];
+  }
+  if (rx === 0 || ry === 0) {
+    return [['L', tx, ty]];
+  }
   const segsNorm = arcToSegments(tx - fx, ty - fy, rx, ry, large, sweep, rot);
 
   for (let i = 0, len = segsNorm.length; i < len; i++) {
--- a/packages/core/src/util/path/index.ts
+++ b/packages/core/src/util/path/index.ts
@@ -96,8 +96,12 @@ const arcToSegments = (
   sweep: number,
   rotateX: TRadian,
 ): TParsedAbsoluteCubicCurveCommand[] => {
-  if (rx === 0 || ry === 0) {
+  if (toX === 0 && toY === 0) {
     return [];
+  }
+  if (rx === 0 || ry === 0) {
+    // an arc with a zero radius is a straight line to the endpoint
+    return [['C', toX / 3, toY / 3, (toX * 2) / 3, (toY * 2) / 3, toX, toY]];
   }
   let fromX = 0,
     fromY = 0,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [next]
 
+- fix(path): draw an arc with a zero radius as a line instead of dropping it
 - chore(deps-dev): bump @playwright/test from 1.60.0 to 1.62.0 [#11063](https://github.com/fabricjs/fabric.js/pull/11063)
 - test(), chore(): Various things i forgot [#11065](https://github.com/fabricjs/fabric.js/pull/11065)
 - feat(): Add backend network request callback [#11064](https://github.com/fabricjs/fabric.js/pull/11064)
--- a/packages/core/src/shapes/Path.spec.ts
+++ b/packages/core/src/shapes/Path.spec.ts
@@ -624,9 +624,11 @@ describe('Path', () => {
       'M62.87543,168.19448H78.75166a0,0,0,0,1,0,0v1.9884a6.394,6.394,0,0,1-6.394,6.394H69.26939a6.394,6.394,0,0,1-6.394-6.394v-1.9884A0,0,0,0,1,62.87543,168.19448Z',
     );
 
+    // the first zero radius arc ends where it starts and is dropped, the last
+    // one moves by 4e-5 and becomes a line
     expect(
       path.path.length,
       'path should have correct number of commands',
-    ).toBe(9);
+    ).toBe(10);
   });
 });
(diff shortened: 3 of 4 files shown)


asturur:
Hey @Jaybhade thanks for the PR. i will look at it soon. I have nothing against rendering a line instead of skipping a segment if it helps some applications staying compliant with svg.

packages/core/src/util/path/index.ts
```diff
@@ -96,9 +96,13 @@ const arcToSegments = (
   sweep: number,
   rotateX: TRadian,
 ): TParsedAbsoluteCubicCurveCommand[] => {
-  if (rx === 0 || ry === 0) {
+  if (toX === 0 && toY === 0) {
     return [];
   }
+  if (rx === 0 || ry === 0) {
+    // an arc with a zero radius is a straight line to the endpoint
+    return [['C', toX / 3, toY / 3, (toX * 2) / 3, (toY * 2) / 3, toX, toY]];
```
asturur: if is a straight line, why a C and not a L ?

Jaybhade: No good reason — an `L` is better and it's now an `L`, in **464e82f**.

The `C` came from where the check sat. Inside `arcToSegments` the return type is `TParsedAbsoluteCubicCurveCommand[]`, and `fromArcToBeziers` then translates each result by offsetting indices 1 through 6, so anything returned from there has to be shaped like a cubic to survive that loop. Moving the degenerate cases up into `fromArcToBeziers` — which already has `fx, fy, tx, ty` in absolute coordinates and doesn't need the translation pass — drops that constraint:

```ts
if (tx === fx && ty === fy) {
  return [];
}
if (rx === 0 || ry === 0) {
  return [['L', tx, ty]];
}
```

`L` is a first-class command in the simplified path, so nothing downstream needed changing: `_renderPathCommands` emits `lineTo`, `_calcDimensions` takes the point directly, `transformPath` has an `L` case, and `getPathSegmentsInfo` uses `calcLineLength` — an exact length instead of the 100-step `pathIterator` sampling a straight cubic would have cost. `getPointOnPath` lerps it rather than running a bezier iterator. `fromArcToBeziers` isn't in the `util` export list, so widening its return type is internal only.

One thing that came along with the move, since the empty-arc guard travelled too: on master, an arc with real radii whose endpoints coincide reaches the maths with `rx²·py² + ry²·px² === 0`, so `root` is `Infinity` and `cx`/`cy` are `NaN`. `dtheta` is then `NaN`, `segments = Math.ceil(NaN)` is `NaN`, and `for (i = 0; i < NaN; i++)` never runs — so it already returned `[]`, by accident. The explicit check just makes that the stated behaviour rather than a side effect of `NaN` comparisons.

Gates on 464e82f: `typecheck` clean, `prettier:check` clean, `unit-node` 2045 passed, `unit-chromium` 2046 passed.

Jaybhade:
Note on the one red check, so it doesn't read as this patch's doing: `Vitest tests node 22` failed on `Image.spec.ts > SVG import > can import images when xlink:href attribute is set` with "Test timed out in 5000ms". That test loads images through jsdom's resource loader over the network; the same commit passed `node 20`, `node 24`, `chromium`, `firefox` and Playwright, and nothing here touches image loading — only `fromArcToBeziers` in `util/path`. I can't re-run the job from my side (needs repo rights), so it'll stay red until someone re-runs it or I push again.

packages/core/src/util/path/index.ts
Smrtnyk: > No good reason — an `L` is better and it's now an `L`, in **[464e82f](https://github.com/fabricjs/fabric.js/commit/464e82ff1fdfdc4ede06ba30614540240315dac6)**.
> 
> The `C` came from where the check sat. Inside `arcToSegments` the return type is `TParsedAbsoluteCubicCurveCommand[]`, and `fromArcToBeziers` then translates each result by offsetting indices 1 through 6, so anything returned from there has to be shaped like a cubic to survive that loop. Moving the degenerate cases up into `fromArcToBeziers` — which already has `fx, fy, tx, ty` in absolute coordinates and doesn't need the translation pass — drops that constraint:
> 
> ```ts
> if (tx === fx && ty === fy) {
>   return [];
> }
> if (rx === 0 || ry === 0) {
>   return [['L', tx, ty]];
> }
> ```
> 
> `L` is a first-class command in the simplified path, so nothing downstream needed changing: `_renderPathCommands` emits `lineTo`, `_calcDimensions` takes the point directly, `transformPath` has an `L` case, and `getPathSegmentsInfo` uses `calcLineLength` — an exact length instead of the 100-step `pathIterator` sampling a straight cubic would have cost. `getPointOnPath` lerps it rather than running a bezier iterator. `fromArcToBeziers` isn't in the `util` export list, so widening its return type is internal only.
> 
> One thing that came along with the move, since the empty-arc guard travelled too: on master, an arc with real radii whose endpoints coincide reaches the maths with `rx²·py² + ry²·px² === 0`, so `root` is `Infinity` and `cx`/`cy` are `NaN`. `dtheta` is then `NaN`, `segments = Math.ceil(NaN)` is `NaN`, and `for (i = 0; i < NaN; i++)` never runs — so it already returned `[]`, by accident. The explicit check just makes that the stated behaviour rather than a side effect of `NaN` comparisons.
> 
> Gates on [464e82f](https://github.com/fabricjs/fabric.js/commit/464e82ff1fdfdc4ede06ba30614540240315dac6): `typecheck` clean, `prettier:check` clean, `unit-node` 2045 passed, `unit-chromium` 2046 passed.

Just a respectful remark from my side:
This feels like a wall of text generated by AI
Also why do you have to mention that prettier and other stuff pass? We can already see that from the build status
It would be nice if we could keep conversations here on a human level and not use ai to communicate

Smrtnyk:
> Note on the one red check, so it doesn't read as this patch's doing: `Vitest tests node 22` failed on `Image.spec.ts > SVG import > can import images when xlink:href attribute is set` with "Test timed out in 5000ms". That test loads images through jsdom's resource loader over the network; the same commit passed `node 20`, `node 24`, `chromium`, `firefox` and Playwright, and nothing here touches image loading — only `fromArcToBeziers` in `util/path`. I can't re-run the job from my side (needs repo rights), so it'll stay red until someone re-runs it or I push again.

It is hard to read this wall of text that is barely formatted.
I have rerun the build, since that is the only thing I understood from your comment, that you want a rerun of the build.

Smrtnyk:
Is this something that would benefit from a visual test?

asturur:
I feel the same i try to keep conversation human.
Visual tests should already cover this issue here, since there were test for collapsed arcs

PR #11075, opened 2026-09-01
fix(Intersection): reject mirrored points in isPointContained infinite check

dylanpulver:
## Description

`Intersection.isPointContained(T, A, B, true)` (infinite/line mode) wrongly
returns `true` for points that lie on the *reflection* of the line, not the line
itself. For a sloped segment it computes `s = AT / AB` componentwise and, in the
`infinite` branch, tests `Math.abs(s.x) === Math.abs(s.y)`. A point is on the line
only when `AT = s·AB` for a single scalar, i.e. `s.x === s.y` (sign included).
Using `abs` on each component also admits `s.x === -s.y`, which is the mirrored
point.

Example: with A=(2,1), B=(4,2) (line y = x/2), the point (0,2) is not on the line,
yet `isPointContained((0,2), A, B, true)` returns `true` today.

The fix uses the same scalar the segment branch already relies on:
`s.x === s.y && (infinite || (s.x >= 0 && s.x <= 1))`. The segment path is
unchanged; only the previously-broken infinite path changes.

Why it went unnoticed: the method is only ever called internally in segment mode
(`intersectLineLine`'s coincidence check), so the infinite branch is public-API
only, and the existing spec exercised the infinite branch only with on-line points
(plus one off-line point whose `s.y` is 0, which does not trigger the reflection
case).

## Key changes

- `isPointContained` infinite/sloped branch: `s.x === s.y` instead of
  `Math.abs(s.x) === Math.abs(s.y)`.
- Added a regression test: a mirrored point must not be contained by the infinite
  line.

Fuzzing 300k random integer cases against a cross-product oracle: the old code
disagreed with ground truth in 5244 cases (all reflected-point false positives);
the fix disagrees in 0, and the segment path is unaffected. Full `unit-node` suite
passes (2043 tests, 0 failures).


asturur:
can you explain me with a picture which was the point that failed the old code? i m having issues understanding it.

Smrtnyk:
> ## Description
> `Intersection.isPointContained(T, A, B, true)` (infinite/line mode) wrongly returns `true` for points that lie on the _reflection_ of the line, not the line itself. For a sloped segment it computes `s = AT / AB` componentwise and, in the `infinite` branch, tests `Math.abs(s.x) === Math.abs(s.y)`. A point is on the line only when `AT = s·AB` for a single scalar, i.e. `s.x === s.y` (sign included). Using `abs` on each component also admits `s.x === -s.y`, which is the mirrored point.
> 
> Example: with A=(2,1), B=(4,2) (line y = x/2), the point (0,2) is not on the line, yet `isPointContained((0,2), A, B, true)` returns `true` today.
> 
> The fix uses the same scalar the segment branch already relies on: `s.x === s.y && (infinite || (s.x >= 0 && s.x <= 1))`. The segment path is unchanged; only the previously-broken infinite path changes.
> 
> Why it went unnoticed: the method is only ever called internally in segment mode (`intersectLineLine`'s coincidence check), so the infinite branch is public-API only, and the existing spec exercised the infinite branch only with on-line points (plus one off-line point whose `s.y` is 0, which does not trigger the reflection case).
> 
> ## Key changes
> * `isPointContained` infinite/sloped branch: `s.x === s.y` instead of
>   `Math.abs(s.x) === Math.abs(s.y)`.
> * Added a regression test: a mirrored point must not be contained by the infinite
>   line.
> 
> Fuzzing 300k random integer cases against a cross-product oracle: the old code disagreed with ground truth in 5244 cases (all reflected-point false positives); the fix disagrees in 0, and the segment path is unaffected. Full `unit-node` suite passes (2043 tests, 0 failures).

this wall of text is very hard to understand, is this ai generated PR?

asturur:
i m coming back to all the prs soon!

Smrtnyk:
> i m coming back to all the prs soon!

Do you think it makes sense to add some ai guidance to this repo? some criteria to tell to people that it is ok to use ai to write code and open prs in this repo, but the description should be human made.
Communication should be human to human.
It is really hard to communicate in the last few PRs I have seen here.
There are walls of text that are ai generated that I have no interest to read, it is pointless and hard to understand.
Additionally, they are fixing some ui issues, but would be very helpful to see before and after screenshots where eligible.
A visual test should be mandatory when it is eligible as well.
Either existing snapshot should change demonstrating a fix or a new one should appear.
These are just some remarks I am noticing.

asturur:
Yes we should also add an official agents.md that also explains that.
I have zero issue with ai contributions from junior developers, as long comunication and relationship stay human.

I ll continue to do first-message ban of any 'authorized digital assistant'

asturur:
In this case tho the text is mostly fine imho, i m familiar with those algebric concepts and it doesn't seems AI generated, or at least seems human reviewed

PR #11083, opened 2026-09-12
ci(): Fix the banner of the website.

asturur:
## Description

Banner of the website was working locally, but once deployed it died.
This is something i need to be careful with but i think as soon as i stop using the old imports everything should be fine.

### DIFF
--- a/fabric.ts
+++ b/fabric.ts
@@ -1,1 +1,8 @@
+import { setEnvFactory } from './packages/core/src';
+import { getEnv } from './packages/browser/src/env';
+
+// Keep browser imports SSR-safe: window/document are read only when Fabric
+// first needs the environment.
+setEnvFactory(getEnv);
+
 export * from './packages/core/src';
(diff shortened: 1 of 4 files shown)


Smrtnyk (approved):

fabric.ts
```diff
@@ -1 +1,8 @@
+import { setEnvFactory } from './packages/core/src';
+import { getEnv } from './packages/browser/src/env';
+
+// Keep browser imports SSR-safe: window/document are read only when Fabric
+// first needs the environment.
+setEnvFactory(getEnv);
+
```
asturur, line 7: this because i built the banner with `import from fabric/es` i could have chanced imports, but that would have still meant that the new exports were broken.

I prefer to keep the old exports still valid and do simple release.

asturur:
Why do i have to approve my own prs to run the CI... so weird

Smrtnyk:
> Why do i have to approve my own prs to run the CI... so weird

yeah I also found it strange

PR #11086, opened 2026-09-12
ci() consolidate checks and harden reporting workflows

asturur:
## Description

I wanted to tested this because i knew with the lastest faster tools for build and format/check, our pipeline was no more optimized.
So i took the best model i had access to, i explained it what i wanted and to reorganize the flows for speed security and compatibility with forks.

This is what it did.


Smrtnyk (approved):

asturur:
this thing is too confusing i will split it in minor changes. I can't really review the diffs.

PR #11089, opened 2026-09-13
ci(): share Fabric build across build checks

asturur:
## Description

Another tentative to reduce runners parallelism with small cost.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,36 +8,34 @@ permissions:
   contents: read
 jobs:
   build:
+    name: Build and website
     runs-on: ubuntu-24.04
     steps:
-      - name: Uglified build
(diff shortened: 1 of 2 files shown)


asturur:
@Smrtnyk after this PR i m looking into what is breaking astro 7 upgrade and the security flaw reported on the actions.

Then i m trying to understand how i publish now.

Then i publish 7.5.0 with all the fixes we got in this period.

At that point i grab small features pr that have been contributed last months and i port them on the new code base on the old owners branches so they get the commit as reference for themselves.

If you are interested in small feature development feel free to pick some of those, port them over, verify. Just if is interesting for you. Instead if you want to do discovery test on new things there are big items pending we can discuss

Smrtnyk (approved):

PR #11090, opened 2026-09-13
ci(): fix pnpm caching and skip canvas build dependencies

asturur:
## Description

Another small optimization, use cache correctly ( our old cache version broke ) and stop doing apt-get installs since node-canvas in theory has prebuilts for node 24 now


Smrtnyk (approved):

asturur:
Yes is definetly faster in doing the install. 15 seconds saved.

I wonder then why Node 20 UTs are 14 seconds and the other Nodes 45

PR #11095, opened 2026-09-20
chore(): update dev deps

Smrtnyk:
Most notably update major version of vitest
 Update playwright
 Update other build/lint dependencies
 There are some formatting changes in few files due to oxfmt bump
 And a few snapshot changes in tests that are harmless


Smrtnyk:
@asturur new vitest 5 can't run on node 20 anymore, it dropped the support
should we drop node 20 unit tests job and add node 26?

